Issue #12, opened 2016-05-07
Polaris Port - Dogborgs


Arokha:
- [x] Medihound
- [x] K9
- [x] K69
- [x] BOOP Module
- [x] Sleeper Module
- [x] Tongue Module
- [ ] Fix offset
- [ ] Fix sleeper sprite

Cameron653:
Done. Only a minor bug that can be overlooked. See the PR for more info.

Cameron653:
NOW I've fixed it.

Arokha:
Dogborgs are offset oddly. When walking north/south they are offset sideways, aka they go 'halfway' through doors. Pls2fix. Also sleeper bellies don't show up, probably needs some overriding of the update icons on them borgs.

Cameron653:
Huh. Their bellies were showing up before. I'll look back into this after I
finish egg tf.

Cameron653:
this is fixed

PR #24, opened 2016-05-07
Adds dogborgs #12 (WIP)

Cameron653:
Issue #12 
Adds in dogborgs.
ISSUE: When you attack something, you "jaunt" forwards. This applies when any mob attacks anything. This resets the x = -16  to x = 0, which makes the borgs look displaced, but everything else works.
Note: It has a vore-stomach, so I'm marking it vore-related


Cameron653:
Currently modifying it to actually work with the code.

Cameron653:
Going to fix this

Issue #1, opened 2016-05-07
Polaris Port - Vore


Arokha:
- [x] Belly System
- [x] Vore Panel
- [x] Grab/Click Hook
- Ability to eat people above this line---------------
- [x] Saving/Loading
- [x] Cloning
- [x] Breathing inside pred
- QOL above this line---------------
- [x] Struggling
- [x] Teleport into Belly
- [x] Body Scanner
- [x] Examination
- [x] Welder vision for absorbed
- [x] Gib handling
- [x] Vore drugs
- Extra/recent features above this line---------------

PR #25, opened 2016-05-08
Basic Vore Implementation for #1

Arokha:
Adds vore sounds, bellies, vore procs, all types of vore except
macro/micro as that is not implemented yet. Also adds a few type2type
methods and a text method that we had before.


code/modules/vore/eating/simple_animal_vr.dm
```diff
@@ -0,0 +1,39 @@
...
+//
+/mob/living/simple_animal/verb/toggle_digestion()
+	set name = "Toggle Animal's Digestion"
+	set desc = "Enables digestion on this mob for 20 minutes."
+	set category = "Vore"
+	set src in oview(1)
+
+	var/datum/belly/B = vore_organs[vore_selected]
+	if(faction != usr.faction)
+		usr << "\red This predator isn't friendly, and doesn't give a shit about your opinions of it digesting you."
```
Leshana: We should convert these to <span> later.

PR #31, opened 2016-05-08
Adds weight machines and weight variables.

Cameron653:
Exactly what it says on the tin.
Weight descriptions aren't in yet, neither is weight gain/loss (From starving/overeating/medications.) but the machines remove weight and nutrition.

Compiles without errors/warnings.


Arokha:
Looksh good!

PR #36, opened 2016-05-10
Vore Struggling and Breathing and Environment

Arokha:
Polaris does environment checks so I bailed out of those, breathing is done with fake air, and struggling is resist now to cut down on checking for bellies every move.

You can also resist out of dead mobs or mobs with bellies with 'escapable' set.

For #1


Leshana:
This should be good for now! Escape code looks good, air code will work plenty fine.

PR #37, opened 2016-05-10
Adds in examine weight.

Cameron653:
Adds in descriptions when you examine someone with weight. Tested and appears to be fully functioning.


code/modules/mob/living/carbon/human/examine_vr.dm
```diff
@@ -0,0 +1,22 @@
+/mob/living/carbon/human/examine/proc/examine_weight(mob/user)
+
```
Arokha: Declare /var/msg (could be named anything for convienence) and change the += in this file to =.

Basically this file needs to set some message string and return that up to examine.

code/modules/mob/living/carbon/human/examine.dm
```diff
@@ -188,6 +188,8 @@
 		else*/
 		msg += "[T.He] [T.is] wearing \icon[wear_id] \a [wear_id].\n"
 
+	if(attempt_vr(src,"examine_weight",args)) return //VOREStation Code
```
Arokha: This should be, instead,
msg += attempt_vr(src,"examine_weight",args) //VOREStation Code

So the value returned (the string of text) from examine_weight is added to the message here.

Issue #20, opened 2016-05-07
Polaris Port - Gas Meter Readability


Arokha:
- [ ] Single-click Gas Meters to read

PR #52, opened 2016-05-13
Resolves Issue #20 - Gas Meter Convenience

Leshana:
* Changed gas meters to be readable merely by left-clicking, instead of having to examine.  This behavior was present on oldcode and was nice, so bringing it back.
* Changed readability distance to 5 tiles for the sake of convenience.


SpadesNeil:
This should be pushed upstream if it hasn't been already.

PR #87, opened 2016-05-17
Rename Station, Cameras

Arokha:
Also updates station nanoui png


code/game/machinery/camera/presets.dm
```diff
@@ -70,7 +70,7 @@ var/global/list/engineering_networks = list(
 	network = list(NETWORK_MINE)
 
 /obj/machinery/camera/network/northern_star
-	network = list(NETWORK_NORTHERN_STAR)
+	network = list(NETWORK_EXODUS)
```
SpadesNeil: Adephagia! D: Not Exodus! Weh!

Or NETWORK_VORE is fine too.

Arokha: It should be NETWORK_STATION or something because NETWORK_EXODUS now always equals the station name. If you want to change it though, feel free! Just make sure you get any references to it otherwise cameras will stop working :P

SpadesNeil: Agreed.

PR #88, opened 2016-05-17
Adds in "Encase In Egg" vore type.

Cameron653:
Does exactly what it says on the tin. Encases the prey in an egg like all the other egg TF types without actually doing any TF.

Tested, and everything appears to work


Cameron653:
The only two commits that are used are the "Encase in egg" and "I forgot to change this" commits. No idea why it included all the other old commits, too.

PR #89, opened 2016-05-17
Fixes egg_dm not making vore organ have a color

Cameron653:
Quick and easy bugfix.
I originally forgot to put this in. All it does is make the little counter (0) show up as purple when you turn it to the "encase in egg" or if DM_EGG mode.


Cameron653:
Going to forcemerge this, since it's just a real simple and quick fix.

PR #108, opened 2016-05-18
Makes xenoarch prettier and quality of life additions.

Cameron653:
Makes xenoarch icons like they were in the old codebase (Picks look nicer, artifacts look nicer.)
Adds in xenoarch drill bit (Nerfed cost, since it was never actually used/seen due to its high research and material cost.)
Fixes a bug with tajaran and foxes using the fennec suit sprite (This was due to them being the same icon, so the game read it as such.)
Puts binoculars and hazmat (AMI) rigsuits in xenoarch lockers.
Gives the xenoarch rigsuit a drill bit to actually mine with.


code/modules/research/xenoarchaeology/misc_vr.dm
```diff
@@ -0,0 +1,26 @@
...
+	icon_closed = "secureres"
+	icon_locked = "secureres1"
+	icon_opened = "secureresopen"
+	icon_broken = "secureresbroken"
+	icon_off = "secureresoff"
+
+	New()
+		..()
+		sleep(2)
+		new /obj/item/weapon/rig/hazmat(src)
```
Leshana: Why sleeping here?

Cameron653: Ah, forgot to remove that. Let me get rid of it real quick.

code/modules/research/xenoarchaeology/tools/tools_pickaxe_vr.dm
```diff
@@ -0,0 +1,126 @@
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+// Excavation pickaxes - sorted in order of delicacy. Players will have to choose the right one for each part of excavation.
+
+/obj/item/weapon/pickaxe/brush
```
Leshana: Are these all complete copy/pastes of tools_pickaxe.dm  but with the icon file changed?

Cameron653: Yep. I had to do this due to not actually being able to change the tools_pickaxe.dm file or the icons/obj/xenoarchaeology.dmi file. 
This overwrites that and makes it so it uses the icons/obj/xenoarchaeology_vr.dmi file.

Leshana: I guess I'm wondering: Why copy/paste the entire file's contents instead of just overidding the one variable you want to be different?

Cameron653: ...I forgot about that. That would actually be a better way of going about with this. Give me a second to correct this.

Cameron653: Fixed this.

PR #110, opened 2016-05-19
Makes the xenoarch lockers spawn with the equipped AMI hardsuit

Cameron653:
Exactly what it says in the title. I gave it the AMI hardsuit, but not the one that has equiptment.


Cameron653:
Going to wait for the travis and then merge it since it's such a minor thing, and is part of the previous PR I made.

PR #111, opened 2016-05-19
Possibly adds in the "subtle" command

Cameron653:
Possibly adds in the subtle command.
It /should/ work, but I'm not 100% sure, due to being unable to test with anyone else but myself on my test server.

If you have any suggestions or tested and can confirm that this works, please comment below.


Cameron653:
Going to drop and close this for the time being.

PR #207, opened 2016-05-24
Adds in an "Advanced Who"

Cameron653:
Adds in an "Advanced who" which lets normal players see if someone is observing/in the lobby/playing.
Same as it was in the old code, except this one is listed as an "Advanced Who" as it gives players more information, as opposed to the normal who, which only tells you who's on the server.

### DIFF
--- a/code/game/verbs/advanced_who_vr.dm
+++ b/code/game/verbs/advanced_who_vr.dm
@@ -1,0 +1,86 @@
+
+/client/verb/who_advanced()
+	set name = "Advanced Who"
+	set category = "OOC"
+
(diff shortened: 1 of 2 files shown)


SpadesNeil:
I feel like I used to have a reason for not having this, but for the life of me, I can't remember what it was.

If it's just like the old version I see no reason not to add it because I obviously approved that.

Cameron653:
The reason it (before I changed it) was that it showed AFK players to
normal players. I removed it and it was approved. Of course, it was with
the normal who, but I didn't want to touch the pre existing who verb in the
new code.

PR #203, opened 2016-05-23
Bug Fixes

Yoshax:
Does part of #181 better than it does.

PR #211, opened 2016-05-24
Virgo Races Pay Fix

ursamedium:
So, this is something someone pointed out to me over on our work on Eros so I figured I'd check if you guys had the same problem and you did. Races added by VORE should actually get paid for work now! Money, yay
I set all of the economic modifiers for them, except for Akula, to 7, which is Polaris default for Tajara, Unathi, and Diona. Humans default to 10, Skrell to 12, and Teshari to 9. Akula I set to 9, since they have lore ties to the Skrell similar to Teshari.
Modifiers are easy enough to change and add or remove. Cheers! c:

### DIFF
--- a/code/modules/economy/economy_misc.dm
+++ b/code/modules/economy/economy_misc.dm
@@ -53,7 +53,13 @@
 										/datum/species/teshari	= 9, // Skrell sponsored,
 										/datum/species/tajaran	= 7,
 										/datum/species/unathi	= 7,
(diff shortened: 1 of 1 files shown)


Arokha:
Err, sorry... #203 changed how econ works and didn't close the issue. ;w; Sorry for any time spent.

ursamedium:
No trouble! Took like two minutes. Glad you guys had it fixed already c:

PR #218, opened 2016-05-25
Fixes #155 - Allow building lights on windows.

### LINKED ISSUE #155: Cannot build lights on windows.
Tried to attach a light fixture to a window (as a drone; untested as human).

The light fixture to snap on as it does with walls.

#### What actually happened
I damage the window and the light doesn't attach.
capture

#### Steps to reproduce
Try to attach a lighting fixture to a window.

#### Additional info:
 - **Server Revision**: https://github.com/VOREStation/VOREStation/commit/849c30750d4ba1099d650c1abc9cf818c8fc4073

Leshana:
The existing map has lights on windows, so it should be possible right?

### DIFF
--- a/code/game/objects/structures/window.dm
+++ b/code/game/objects/structures/window.dm
@@ -271,6 +271,9 @@
 			else
 				new glasstype(loc)
 			qdel(src)
+	else if(istype(W,/obj/item/frame) && anchored)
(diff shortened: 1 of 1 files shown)


SpadesNeil:
We should send this upstream.

PR #204, opened 2016-05-23
Vulparace

Poojawa:
Vulpkanin race from Paradise!

Canine/vulpine base that isn't really old sprites from Citadel Station.

Facial hair allows you to change cheek floof as a sort of poor man's second color layer. Plans are in place to add an 'under belly' and 'gloves' to each sprite so proper fox color ratios can be maintained. 

Tail has wagging abilities, ~~but they're disabled pending conversion to Polaris' formats~~ Should be fixed. <3

Canine speech, made of suitably canine noises. Rfft.

Lesser form is Wolpin, but no cubes added for approval.

Custom hardsuit/masks/helmet support, they're actually pretty bad ass tbh. RIG sprites included.

PR #229, opened 2016-05-27
Fix the on-mob icon for vulpine belt! Oops!

Leshana:


SpadesNeil:
Does this need to be merged after #204?

Poojawa:
No conflicts, so no worries

Leshana:
Its named "vulpine belt" but it has nothing to do with the fox species.  I just named it that because its the best belt, and so _obviously_ must be vulpine amirite?

Issue #144, opened 2016-05-21
Admin "secrets" panel not pretty and organized like ours used to be.


SpadesNeil:
Polaris would probably appreciate that old feature too. It was much more organized.

PR #233, opened 2016-05-27
Organize and Beautify Admin Secrets Panel

Leshana:
* Fixes Issue #144 
* Add browser styles to Admin Secrets panel to make it look nice.
* Show only one category at once, with buttons to switch between categories at the top.


SpadesNeil:
Later we need to re-add the "random" events tab, since that appears to have been removed for some reason in Polaris.

Leshana:
@SpadesNeil Or the event manager panel.

PR #247, opened 2016-05-30
Gaggle of map tweaks

SpadesNeil:
- Added docking point for Skipjack, modified the Skipjack so it could actually land at it.
- Fixed inconsistencies with the gravity random event.
- Fixed lore inconsistencies with communicators.

Tried doing some of this with _vr files but couldn't do it right.


SpadesNeil:
Travis... why?

PR #250, opened 2016-05-30
Adds in language implants.

Cameron653:
Adds in language implants.
Language implants give almost all the languages (With the exception of hivemind races and things like rootspeak and the robot language.)

Mostly made for admin events in which someone would need all the common languages at once, but also is creatable via R&D for a very large price and research requirements.
(material= 5, bio = 5, data = 4, engineering = 4)
(metal = 7000, glass = 7000, gold"= 2000, diamond = 3000)


Leshana:
This looks like it won't break anything.

PR #253, opened 2016-05-31
Ports my changes to FBP Construction on Polaris

Arokha:
In the title.


Arokha:
They merged this, so no need. Will get it in sync.

PR #275, opened 2016-06-01
Fixes 2 issues

Arokha:
This is a bug fix.

Proposed changes in this Pull Request:
- Fax machines reduced access requirement (none, actually). Abuse and get banned.
- Digested IDs have an icon again.


code/modules/vore/eating/belly_vr.dm
```diff
@@ -278,7 +278,7 @@
 		// Keep IDs around, but destroy them!
 		var/obj/item/weapon/card/id/ID = W
 		ID.desc = "A partially digested card that has seen better days.  Much of it's data has been destroyed."
-		ID.icon_state = "digested"
```
SpadesNeil: Why does it say this line was removed?

Arokha:
Because it was. Whups. Well it would have been a step in the right direction anyway. Would have been a generic looking ID with no access vs the invisible one that exists now. XD

SpadesNeil:
Fallback icons are good, yes.

PR #310, opened 2016-06-06
Touch up fitness machine code

Arokha:
Cadence's original work seemed fine... but byond said it was one of the top-time procs. I have no idea why, the original looked fine, but I tried to thin it anyway.


SpadesNeil:
The clown punching bag also used to be animated when punched. Can you correct that? During the cooldown it had a second icon that played without looping.

Arokha:
I guess I already did? I hit the clown bag and it moves, so. Not sure what was broken about it before.

SpadesNeil:
Weird.

PR #320, opened 2016-06-07
Tweaks digestion code

Arokha:
Makes it more synth-friendly, less sparks shooting out of people. More preserving items correctly. Less blank items in people's belly panel.

Lazy-deletes items rather than scouring and deleting one at a time. Just picks out items it likes and puts them into the pred's belly, then deletes the whole mob.


SpadesNeil:
This more of a bug fix or feature? I'd say bug fix but it's hard to say.

Arokha:
Umm. Eeh. I dunno. Bug probably. It is basically making it behave like players would expect, I guess.

Issue #331, opened 2016-06-08
Map Sync


SpadesNeil:
**Current Server Revision**: https://github.com/VOREStation/VOREStation/commit/58009958a1919db56bf8eddb389d5c823764d19c

This issue will pop open and close every few months whenever I think the map should be updated to sync up with changes by Polaris.
### Map changes vital to Vorestation must be carried over, so they are to be listed below.
- [x] Weight lifting equipment by the pool.
- [x] Dorms must be possible to bolt closed.
- [x] Public gateway should be teleporters.
- [x] Gateway attached to EVA.
- [x] Remove CentCom Gateway.
- [x] EVA storage changes.
- [x] One Cryopod in every department.
- [x] Redecorate bar slightly to have Cryopods
- [x] New Jukeboxes
- [x] Tasers for Detectives
- [x] Dorms connect to hallway.
- [x] Vault checkpoint.
- [x] Vault loot.
- [x] Secret stuff!

SpadesNeil:
Forgot new jukeboxes. Gonna re-open this.

PR #333, opened 2016-06-09
Map Sync

SpadesNeil:
Closes https://github.com/VOREStation/VOREStation/issues/331


Arokha:
So this changes the civilian areas to be like theirs or?

SpadesNeil:
All changes listed in my issue are "things that we need that Polaris doesn't have".

So, our visitor section looks like it did before with lockable dorms. Everything else looks like their station.

PR #349, opened 2016-06-11
Fixes bellies overwritng each other

### LINKED ISSUE #277: Belly flavortext being overwritten.
1. (More important issue) - All Vore panel information persists among all characters, **overwriting** any bellies and all flavor text on them. 

2. If you try to edit flavor text, it opens a blank text that saves regardless of hitting save or cancel.

1. Each character is supposed to have their own Vore panel custom bellies right? I expected to have them remain separate and persist independently of each other.

2. Either the text dialogue opens with the existing text present (like most other dialogues) and/or that when you hit cancel, it doesn't overwrite your existing text.

#### What actually happened
1. A new character can be made, and it will have all the vore panel stuff of the previous characters. edits made on it will overwrite bellis on other characters.

2. Accidentally clicked the flavor text button? Don't worry, I'll just erase everything for you.
both: Lose many many many lines of text.

#### Steps to reproduce
1. A) Make a character, B) set up bellies with beautifully thought out flavor text for a handful or internal organs, C) make a different character and do the same, D) Check on the first character. (This is not an issue for everyone it seems, but some definitely knew what I was talking about)

2. A) Write beautifully thought out flavor text for a handful or internal organs, B) consider editing the flavor text, C) Change your mind and hit cancel.

#### Additional info:

- the 'save preferences' button in the vore panel seems to be useless as it saves everything I do on all characters regardless of hitting the button or not. (maybe this is a separate issue too?)
 - Queried other players, some have had issue 1 since the vore update, **some haven't had the problem at all.**
 - I know this should be 2 separate issues, but I'm dumb apparently and since I started it as one issue, I just stuck with it after re-writing. 
 - Edit: This ENTIRE post was rewritten after I found time to do testing.

Arokha:
[Resolves #277]

Now you can have more than one slot. Fair warning, if you load another character then save bellies, it overwrites those. Like if you do that AFTER spawning. Like you're playing a mob from slot 1, you go into preferences > character set up, load slot 2, then open vore panel and click save, it will overwrite slot 2.

On the one hand, an easy way to copy bellies between characters! On the other, perhaps unexpected.

This prooooooooooobably doesn't delete everyone's bellies? Probably. Maybe you should copy the data folder, just in case. ;o Skip the logs.

### DIFF
--- a/code/modules/client/preferences.dm
+++ b/code/modules/client/preferences.dm
@@ -235,12 +235,14 @@ datum/preferences
 	else if(href_list["reload"])
 		load_preferences()
 		load_character()
+		attempt_vr(client.prefs_vr,"load_vore","") //VOREStation Edit
 	else if(href_list["load"])
 		if(!IsGuestKey(usr.key))
 			open_load_dialog(usr)
 			return 1
 	else if(href_list["changeslot"])
 		load_character(text2num(href_list["changeslot"]))
+		attempt_vr(client.prefs_vr,"load_vore","") //VOREStation Edit
 		close_load_dialog(usr)
 	else
 		return 0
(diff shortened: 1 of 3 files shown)


SpadesNeil:
Marking "DO NOT MERGE" pending review by Leshana.

PR #360, opened 2016-06-14
PDW added

SpadesNeil:
Because I have a gun fetish <s>and a problem.</s>

I'd potentially like to actually replace the WT550 SMGs already on the map with these, because these look less shitty to me.
pdw
Icons by the same dude who did the STS, L6 SAW, and many others.

Also I don't know if what I did with autolathe_datums_vr will even work. Haven't had a chance to test it.


Arokha:
I like thumbhole stocks and bullpup weapons. :open_mouth:

Arokha:
Also I would think your autolathe datum thing would work.

DemicusMaximus:
Looks pretty cool

PR #366, opened 2016-06-15
Exclusive list add for safety

Arokha:
Changes += to |= to just make super sure you can't create dupe entries in a belly.


code/modules/vore/eating/belly_vr.dm
```diff
@@ -158,7 +158,7 @@
 		prey.buckled.unbuckle_mob()
 
 	prey.forceMove(owner)
-	internal_contents += prey
+	internal_contents |= prey
```
SpadesNeil: I am not familiar with this operator, oddly. What does it mean?

PR #379, opened 2016-06-23
Allows hardsuits and rigsuits to store backpacks

Cameron653:
Brings back a feature from the old code which allowed you to place backpacks on hardsuits. Rigsuits have also been included in this PR.

Photo: http://puu.sh/pCBd1/d8432ab235.png


SpadesNeil:
Might be imbalanced but fuck if I care. I like having usable backpacks in hardsuits.

PR #393, opened 2016-07-01
Gives monkeys default language, removes neatness tabs.

Cameron653:
Gives the monkeys default languages.
Removed the tabs made to make it all line up, as Github totally ruined it.


SpadesNeil:
:+1:

PR #395, opened 2016-07-01
More banners for related universe events

JoanRisu:
Some adjustments with custom items plus two new banners.


SpadesNeil:
I want a UNSC flag in a later update. Also a Syndicate flag.

PR #413, opened 2016-07-08
Fixes transformation vore types.

Cameron653:
Two bugs fixed
- **Prey no longer rejects their organs after transformation, resulting in their slow death due to mass organ failure.**
- **Prey now actually change their species sprite after transformation.**
- Updateicon() placed in some areas where it wasn't previously to make sure that the prey's icon 100% updates, especially when it comes to male/female TF
- Prey is set to come out with the same hairstyle as the pred, since some predators have things like the cobra hood, and it'd look dumb without it. They prey can change it by using a mirror.

## Ready For Merge


Cameron653:
Could someone revivew this and possibly merge it? I think I got everything, but I might've missed something.

PR #442, opened 2016-08-03
Security checkpoint added by vault.

SpadesNeil:
- New checkpoint near medical to guard the vault and cover the alternate entry route to the station from the dorms. Security can now guard the vault and have a first response area to any bar fights.
- Moved stupidly placed emergency lockers in the dorms.

dreamseeker 2016-08-02 20-46-34-18


SpadesNeil:
No one checks my pulls anymore so I have to do my own pulls. <_>

PR #450, opened 2016-08-07
Polaris sync

Eearslya:
Mashing together two repos sure is fun.. Hope I got this all right.


Cameron653:
Since there are _569_ files changed, I couldn't look at all the files for problems, but I just went ahead and skimmed through some of it (Couldn't look at much, due to my browser lagging heavily and nearly crashing) and it appears to be good for merge. 

Didn't see anything that would totally break the server.

code/game/machinery/autolathe_datums_vr.dm
```diff
@@ -1,20 +1,16 @@
-/datum/autolathe/recipe/classic_smg_9mm
```
SpadesNeil: Since autolathe_datums was removed, this file, autolathe_datums_vr, should be changed to autolathe_vr.

The _vr versions are always "vorestation" mods of the originals, either adding or changing definitions to avoid compatibility issues where possible by simply overriding the original class definition or procs.

Eearslya: Got it, done.

SpadesNeil:
I can't find anything wrong with this.

PR #461, opened 2016-08-09
Allow greater customisation of wiki links (Fixes #460)

### LINKED ISSUE #460: Wiki button doesn't work right.
The wiki button is configured for Polaris's wiki.

When I type in "Backstory" to the wiki button prompt, it should search for that topic.

#### What actually happened
http://wiki.vore-station.net/index.php?title=/doku.php?do=search&id=Backstory

#### Steps to reproduce
Use the wiki button in-game.

#### Additional info:
 - **Server Revision**: https://github.com/VOREStation/VOREStation/commit/86a12253460fdb8b95fcaf4d27d3c1ed3b850cb1

Eearslya:
Because different servers have different wiki setups, it's just not right to have the search URL hardcoded in. MediaWiki has its own search URL, other wikis have their own. I've added in a new config option to allow server owners to designate the URL to be used for searching.

Bit of a breaking change, since all current servers will need to specify this new config option, but I really don't like how it's hardcoded right now.

Fixes #460.

### DIFF
--- a/code/controllers/configuration.dm
+++ b/code/controllers/configuration.dm
@@ -99,6 +99,7 @@ var/list/gamemode_cache = list()
 	var/server
 	var/banappeals
 	var/wikiurl
+	var/wikisearchurl
 	var/forumurl
 	var/githuburl
 	var/rulesurl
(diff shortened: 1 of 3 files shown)


SpadesNeil:
We should send this tweak to Polaris, too.

SpadesNeil:
Assigning self to update server and change config option. Will not merge until I do so.

PR #477, opened 2016-08-14
Multiple bugfixes

Cameron653:
- Allows IDs to be warn on ears
- Lets macros pick up micros again (I have literally 0 idea how this fixes it, but... It does, so woo!)
- Fixes a stomach inconsistency issue that made it so stomach text wouldn't show up from 866 to 1199 nutrition

~~I actually do have a vague idea of how it fixes it.~~ 

Also, neither one of these were reported on github as far as I can see.

### DIFF
--- a/code/modules/mob/living/carbon/human/examine_vr.dm
+++ b/code/modules/mob/living/carbon/human/examine_vr.dm
@@ -91,7 +91,7 @@
 			message = "<span class='warning'>[t_He] [t_is] extremely hungry. A deep growl occasionally rumbles from [t_his] empty stomach.</span>\n"
 		if(100 to 499)
 			return message //Well that's pretty normal, really.
-		if(500 to 864) // Fat.
+		if(500 to 1199) // Fat.
 			message = "[t_He] [t_has] a stuffed belly, bloated fat and round from eating too much.\n"
 		if(1200 to 1934) // One person fully digested.
 			message = "<span class='warning'>[t_He] [t_is] sporting a large, round, sagging stomach. It's contains at least their body weight worth of glorping slush.</span>\n"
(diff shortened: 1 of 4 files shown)


SpadesNeil:
Why not just set the ID w_class = 1? Seems a lot easier.

SpadesNeil:
Honestly the ID card thing should probably be sent to Polaris. I mean it's a slip of paper/plastic. I don't see why not.

Though really this should be changed by setting w_class = 1

Cameron653:
Oh, does it let you place items with a weight class of 1 on your ears? If so, I could probably change it.

Cameron653:
Also, I did it this way because /obj/item/weapon/card has the SLOT_EARS slot flag, but this was overwritten by the ID's slot flag, making it so you can only put it on your ID slot

SpadesNeil:
I still think changing the w_class is a better move.

Cameron653:
Wait, I just checked and ID's w_class is already at 1

SpadesNeil:
I stand corrected, then. This wasn't the case in the old code.

code/modules/mob/living/carbon/human/examine_vr.dm
```diff
@@ -91,7 +91,7 @@
 			message = "<span class='warning'>[t_He] [t_is] extremely hungry. A deep growl occasionally rumbles from [t_his] empty stomach.</span>\n"
 		if(100 to 499)
 			return message //Well that's pretty normal, really.
-		if(500 to 864) // Fat.
+		if(500 to 1199) // Fat.
```
SpadesNeil: This entire code is bad. This X to Y shouldn't be how it's done. I don't remember who ported this, but they did it poorly, because using "to" doesn't work in any instance you're using decimals. Therefore this whole thing should be changed to ==, =>, and =< instead of "to" this and "to" that.

Cameron653: Indeed, it's rounded in line 62

PR #479, opened 2016-08-15
Adds in drain mode, prevents egging while absorbed.

### LINKED ISSUE #469: Egging absorbed people keeps blurry vision
When I ate someone and absorbed them, I soon plopped them in an egg, their vision was still blurry as if they were still absorbed, even though they were in an egg. When I ate them again to try to fix it, I think they were still labeled as absorbed, but once I coughed them up instead of egging them, they were fine again.

I expected them to not have blurry vision, as if they weren't absorbed anymore.

#### What actually happened
They got blurry vision, as if they were still absorbed
#### Steps to reproduce
Absorb someone fully and egg them.
#### Additional info:
 - **Server Revision**: master - 2016-08-08
 - **Anything else you may wish to add** Maybe add a feature to just drain the person but not absorb them would be nice, or just unabsorb people that are still inside of you instead of having to cough them back up and eat them again... it's not the right fix for this, but it'd be good for getting the nutrients without having to time my meals and try to avoid having them absorbed, Just a suggestion!

Cameron653:
- Fixes #469 
- Adds in "drain" mode, which acts like absorb but doesn't absorb them, simply drains them of nutrition.
- Makes it so prey can't be TF'd while absorbed
- Adds in an "unabsorb" mode, which uses up some of the owners nutrition to unabsorb their prey.

### DIFF
--- a/code/__defines/belly_modes_vr.dm
+++ b/code/__defines/belly_modes_vr.dm
@@ -13,6 +13,7 @@
 #define DM_TRANSFORM_FEMALE_EGG				"Transform (Female) (EGG)"
 #define DM_EGG 								"Encase In Egg"
 #define DM_DRAIN							"Drain"
+#define DM_UNABSORB							"Unabsorb"
 
 
 // Stance for hostile mobs to be in while devouring someone.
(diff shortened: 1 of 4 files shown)


Cameron653:
Any comments on this? I looked over everything and tested it and everything appears to be working.

Cameron653:
Thatwasfast

PR #489, opened 2016-08-17
Adds weight gain/loss depending on current nutrition

### LINKED ISSUE #290: Weight gain mechanics never actually got fully ported.
https://github.com/SpadesNeil/VOREstation/pull/522/files

Cameron653:
Fixes #290
- Gain 0.001 weight per tick when above 450 nutrition
- Lose 0.001 weight per tick when below 50 nutrition
- Metabolism variable. Can be modified to make players gain/lose weight faster.  

Tested and ready for merge


SpadesNeil:
Well done.

Issue #511, opened 2016-08-23
Adding a new stomach layer


nightwingthedragon:
This is something I'm gonna work on quickly as a proof-of-concept type idea. 

I want to add a layer to the human sprite system such that players can choose to display a large belly on themselves if they're fat or if they've eaten people. I intend to self-assign this project, so this issue is more of a checklist than anything else.

TASK LIST
- [x] Add the new layer so that the sprite can be rendered in the first place
- [x] Add variables to humans so that the toggle will work
- [x] Add a toggle for stomach size based on current nutrient total
- [x] Add functionality such that, when toggled on, humans display a weight-based stomach

----- Break: At this point, weight-based stomachs should be complete
- [ ] Add another toggle for each stomach in the belly panel
- [ ] Each belly with the toggle enabled will contribute a value to the stomach sprite size based on the relative size of eaten prey
- [ ] Ensure this stacks with fatness toggle

----- At this point, vore-based stomachs should be complete

CSchlippy:
I guess at a later point you'd add it for the tails as well, right? Or is that maybe something that could be requested?

nightwingthedragon:
It's something Imma need sprites for if I do, but ultimately would just be a few additions to the code.

Arokha:
If you ever come back and decide to do this, just reopen it.

PR #531, opened 2016-09-05
Belly sprites for getting fat

nightwingthedragon:
Displays a concept for how I would code stomach fatness for eating lots of food at once. Renders up to 4 belly sizes depending on nutrient value, if a toggle is turned on, and only when icons update.

Partially addresses #511


SpadesNeil:
Note, I told Nightwing to do this without any _vr files or however he wants to do it, just so we can see the code. No need to point that out.

SilverTalismen:
Would love to see some screenshot examples :D

code/modules/mob/living/carbon/human/update_icons.dm
```diff
@@ -391,6 +397,36 @@ var/global/list/damage_icon_parts = list()
 
 	if(update_icons)   update_icons()
 
+//VOREstation edit
+/mob/living/carbon/human/proc/update_belly(var/update_icons=1)
```
SpadesNeil: All of this proc should be able to go into update_icons_vr.dm and still work.

nightwingthedragon: It can. The important part is that the belly code MUST be allocated its own layer because previous attempts to add the image onto other layers have always failed, and I'm not certain they can succeed as images and icons can't be blended together, but the back layer always returns an image.

Cameron653:
Shit, I tried to click "details" which is RIGHT ABOVE the update branch button

nightwingthedragon:
Alright. If we get the go-ahead from @SpadesNeil , then I suppose it's good to merge and I can work on the vore-related portion.

SpadesNeil:
Note the "Cannot Merge" tag.

Cameron653:
The cannot merge may be due to PR #550 https://github.com/VOREStation/VOREStation/pull/550

Because this is also the first case that the active edges and can not merge showed up (Although, it was still somehow merged)

SpadesNeil:
One of my maps is derped, then. I do not know how to check for active edges.

nightwingthedragon:
If you need me to, I can create a new branch on current code and copy/paste my changes across.

SpadesNeil:
Well before I merge this, there's some questions I have.

1. Is it compatible with taur bodies? Like can I have a bulgy tail or sagging underbelly instead of a swollen belly? If not, the feature is incomplete and I do not want to do any merging yet.
2. Can the feature be toggled off? If not, the feature is incomplete, and will not be merged.
3. I'm not clear on when the big belly appears. When you eat a lot of food, or when you eat a player?

nightwingthedragon:
1. I'd consider this an expansion, as it relies a lot on waiting for sprites and then on editing the taur code, which isn't actually part of the regular stomach layer. Further, many taurs play with two stomachs, and I know nagas who play without a tail stomach. I argue that the taur stomachs can wait for a later patch.

2. Yes, I said that already. 

> Renders up to 4 belly sizes depending on nutrient value, if a toggle is turned on, and only when icons update.

Cameron653:
Any reason this isn't merged? Been sitting here for 15 days and appears ready for merge.

SpadesNeil:
Yes. They use the wrong coloring format. That's fine for humans and Akula but bad for everything else that is actually sprited correctly. The fact that this hasn't been addressed tells me it's also not been thoroughly tested. Furthermore, there's no support for Skrell or Prometheans or Taurs.

I mean keep the grayscale sprite for legacy support but that should be the backup and not the rule.

All in all, I haven't merged this because it's very unpolished and not done, and it won't be merged until it's more complete.

nightwingthedragon:
Would you mind explaining what the correct colour format is, then? This is the only way I know colours to work.

Adding taur support is not mandatory to merging this pull request. It would take weeks for Molenar to finish spriting up four additional sprites in every direction (16 sprites) for each taur body (How many of those do we have, 6? 8?) - 128 sprites total.

I also can't help but feel as if you've not really explained WHY any of the problems you've listed are problems. WHY is there no support for skrell or prometheans? WHY is this "not done" or "unpolished"? Y'know? I can't fix a problem if I can't see the problem.

SpadesNeil:
Okay well here's what I want you to do, so you can see it for yourself.

Adjust character colors to various settings of lightness and darkness, then...
- [ ] Test for Skrell
- [ ] Test for Prometheans
- [ ] Test for Tajaran and/or Unathi
- [ ] Test for Fennec race
- [ ] Test for Diona
- [ ] Test for Seromi
- [ ] Test for Xenochimera
- [ ] Test for Xenohybrid

nightwingthedragon:
I'm going to have to merge master into my branch first, aren't I?

SpadesNeil:
Probably not, these bugs will still show up in whatever version you're running. Just test it.

PR #558, opened 2016-09-16
Fixes ZAS issues on away maps

SpadesNeil:
It's probably one of my Z-levels but let's make a control test.


SpadesNeil:
Ugh, are you kidding me? If this is what I think it is...

ArgobargSoup:
Tried googlin' "active edges" in relation ta byond stuff and got nothin', so now I'm curious.  Have ya figured out what it was?

SpadesNeil:
I know exactly what the problem is. It's an SS13 thing; not a Byond thing. Google won't help. It's a map issue.

SpadesNeil:
Oh, that fixed it. Next map.

PR #563, opened 2016-09-18
Adds bird mobs

Cameron653:
Adds birds
Photo: http://puu.sh/rfUyV/5250f0356f.png

- Adds a multitude of birds.
- Birds will eat micros that have it enabled.
- Some birds are smaller and will flap their wings, while others are bigger and won't flap their wings.
- Birds are a subtype of cats so it'd be easy to have them eat micros and a few other things, like their retreat code. Doesn't effect anything gamewise.
- Adminspawn only. Garden should be changed into an aviary, since it rarely/if ever sees any use. But that's up to the mapper.
- Birds can be held and put on heads
- Adds cargo crate containing all the birds for a cost of 80 (Since you get 22 birds in the crate)

#### Sprites from goonstation


Cameron653:
DNM, I want to add in cargo crates to buy more birbs

SpadesNeil:
Kind of big, aren't they?

Cameron653:
Some of them are big, yeah. Mainly the bluejay, the euro-robin, azuretit, and blackbird, but all the other ones are normal sized, and the smaller ones are about the same size as poly was. I was thinking of having the normal sized birds and smaller birds added into an aviary, and leaving the bigger ones as adminspawn & cargo only

code/datums/supplypacks/hydroponics_vr.dm
```diff
@@ -0,0 +1,6 @@
+/datum/supply_packs/hydro/goat
+	name = "Birds Crate"
+	cost = 80 //You're getting 22 birds. Of course it's going to be a lot!
```
SpadesNeil: I'd up this to 200 so it cannot be spammed.

Cameron653: Done.

PR #570, opened 2016-09-19
stray vulp pixel

Poojawa:
torso_m DIR 4 had a stray pixel that was overlooked

Fixed


Poojawa:
squash as usual. had to make sure it was up to date again since the desktop client's a bit slow.

PR #571, opened 2016-09-20
New sec area

kcin2001:
Adds a security workout room, dorms, and laundry

image


kcin2001:
please tell me if anything needs to be changed

SpadesNeil:
10,103 additions, 10,039 deletions not shown because the diff is too large. Please use a local Git client to view these changes.

This is a sign the map merge tool was not used correctly.

kcin2001:
how would i fix that then?

PR #574, opened 2016-09-20
Hotfix to give the RD EVA access again

Cameron653:
- Gives the RD EVA access
Marking this as a feature as it was purposefully removed by polaris?


SpadesNeil:
Travis is being stupid.

Issue #306, opened 2016-06-06
Unejectable, blank items stuck in vorebelly.


Bluhblaghbluh:
#### Brief description of the issue
A digested ID, and presumably any other ID, may become a blank item that cannot be removed from the vore bellies. These can even multiply, and will not respond to being ejected or the move all option. I'm going to further detail it out and provide a few screenshots of what causes the multiplication in addition to the description.
#### What actually happened
Okay, so, leading up to the item being stuck permanently inside me I had eaten and digested someone resulting in an ID being left in my stomach. I intended to move and then eject it, but cancelled the option at one point and then when I went to eject it I discovered that it was no longer the ID, but the blank item. Said blank item does not respond to being removed, or moved between bellies, with the all selection, but can be moved individually. Should you attempt to move one, and then make another belly the active one, it will make the item reproduce, leaving the original one in place in addition to a new one in the targetted belly. I tried to capture the process here: http://i.imgur.com/eb02tF5.png

#### Additional info:
 - **Server Revision**: Found using the "Show Server Revision" verb under the OOC tab.
 - **Anything else you may wish to add** (Location if it's a mapping issue, etc)

SpadesNeil:
Re-opening due to issue returning.

SpadesNeil:
![capture](https://cloud.githubusercontent.com/assets/6484985/18813251/ebb9655a-82c1-11e6-9d9a-d63c9f178038.PNG)

Still an issue as of https://github.com/VOREStation/VOREStation/commit/9da17a2eb407bc85aac6b19a164bfd56fe91818d

Eearslya:
Having trouble reproducing this. If anyone runs into it again, it would help out greatly if I had the following:

* The URL that shows up in the bottom-left of the window when hovering over the blank item.
* A screenshot of the view variables window for that belly; particularly the `internal_contents` variable.

Eearslya:
Okay, managed to accidentally reproduce it. Got someone to view the variables, and..***somehow*** a `null` object is being put into `internal_contents`. I don't know how, and I don't know why.. It was an ID and it only became a blank item when I tried to eject it. The ID also did *not* appear on the floor.

SpadesNeil:
This is being upgraded to severe.

Eearslya:
Seriously need to find **some** way to reliably reproduce this. Until then, have a hotfix.

SpadesNeil:
Is this not happening to you?

Eearslya:
[...]

PR #584, opened 2016-09-23
Fix nested digestion to move mobs/items up a belly when dead

Eearslya:
Current code was only checking for objects when moving nested belly contents up a level. This fixes it so any prey that gets digested will move any of its prey and eaten items up into their predator's belly as intended, instead of getting deleted.


SpadesNeil:
https://github.com/VOREStation/VOREStation/issues/306 This is broken again.

SpadesNeil:
This might be causing problems for https://github.com/VOREStation/VOREStation/issues/610

PR #585, opened 2016-09-24
Pre-October Release

SpadesNeil:
Yes I know it's not October yet. Shut up.


SpadesNeil:
Okay I know exactly why these conflicts are here, and they should be fine to resolve.

Use the master version, and not the release version.

PR #587, opened 2016-09-25
Gives races sprites for new hazmat & construction voidsuit

Cameron653:
Gives them the closest placeholder sprites I could give. 
Gave them default construction/hazmat voidsuit, tajaran, or unathi sprites depending on what would fit best.

This prevents the voidsuit from being invisible/default on the custom races


Cameron653:
Going to merge this as all it does is add in placeholder sprites to prevent the default/invisible voidsuits and nothing else, and won't effect anything else.

PR #590, opened 2016-09-25
Allow the excavation drill to actually mine 30cm as advertised

### LINKED ISSUE #589: Excavation pick
#### Additional info:
 - **Server Revision**: Found using the "Show Server Revision" verb under the OOC tab.
 - **Anything else you may wish to add** (Location if it's a mapping issue, etc)

Eearslya:
Fixes #589 

Not sure why the excavation drill is described as being adjustable from 1-30cm, then programmed to only go 0.5-15cm.

### DIFF
--- a/code/modules/xenoarcheaology/tools/tools_pickaxe_vr.dm
+++ b/code/modules/xenoarcheaology/tools/tools_pickaxe_vr.dm
@@ -41,10 +41,10 @@
 	if(depth>30 || depth<1)
 		user << "<span class='notice'>Invalid depth.</span>"
 		return
-	excavation_amount = depth/2
+	excavation_amount = depth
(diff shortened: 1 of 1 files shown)


Cameron653:
:+1:

SpadesNeil:
This fix should be sent to Polaris if it hasn't already. If it has, reference this issue in the comments.

Eearslya:
The only change was to `tools_pickaxe_vr.dm`. Polaris doesn't have it.

SpadesNeil:
They don't have  /obj/item/weapon/pickaxe/excavationdrill ? That's really something only we have?

Eearslya:
Yep, just did a grep for it to make sure.

PR #598, opened 2016-09-25
Sprite fix

SpadesNeil:


SpadesNeil:
I don't know why this stupid conflict keeps arising.

Cameron653:
I'm unsure as to why that'd be conflicting, as all I did was modify it, delete the old file with the errors, then put the same file in but without the sprite errors.

Eearslya:
What if you set `.dmi` files as binary in `.gitattributes`? It might stop trying to merge them and just overwrite.

PR #613, opened 2016-10-01
Revert "Fix nested digestion to move mobs/items up a belly when dead"

SpadesNeil:
Reverts VOREStation/VOREStation#584

Testing to see if this fixes vore spacing people.


SpadesNeil:
Didn't fix the issue.

Steps to reproduce bug:
1. Pick up micro in hand.
2. Click self.
3. Eat micro.
4. They cannot use 'say' or whatever.

Oddly they are otherwise just fine and not taking damage in this case.

PR #627, opened 2016-10-05
Prevent attacking vore mobs from within them

### LINKED ISSUE #541: You can attack mobs from inside of their stomachs.

Due to reasons of balance and preventing bullshit griefing, this should not be possible to do.

When attacking a mob from inside, for it not to work.
#### What actually happened

https://youtu.be/GGzvpvqK3Ss?t=85 (Except with melee.)
#### Steps to reproduce
1. Get ate.
2. Switch to harm intent.
3. Punch from inside.
#### Additional info:
- **Server Revision**: https://github.com/VOREStation/VOREStation/commit/1e9f00923dd3bbeca86167da0d522b7143dcfff0

Eearslya:
Fixes #541 

As mentioned in the issue, this change only affects `/mob/living/simple_animal` mobs. i.e. Vore mobs, not players.

### DIFF
--- a/code/modules/mob/living/simple_animal/simple_animal.dm
+++ b/code/modules/mob/living/simple_animal/simple_animal.dm
@@ -310,6 +310,8 @@
 			M.do_attack_animation(src)
 
 		if(I_HURT)
+			if (M.loc == src)
+				return // VOREStation Edit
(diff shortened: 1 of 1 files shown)


SpadesNeil:
This doesn't seem to have fixed the issue.

PR #644, opened 2016-10-11
ATMTA files port

NTestuya:
Wasn't able to do this on Github Desktop, so I decided to submit the files manually. Unfortunately, this means that the assets were left out. Fortunately, you can download them via 
[honk.zip](https://github.com/VOREStation/VOREStation/files/520718/honk.zip)

I translated and edited the code from ATMTA station, which was interactions.dm and humans.dm:
https://github.com/Signmark/ATMTA-Station/blob/master/code/modules/mob/living/carbon/human/interactions.dm

https://github.com/Signmark/ATMTA-Station/blob/master/code/modules/mob/living/carbon/human/humans.dm

I figured that this might be good for action oriented servers where games end quick. It's not finished yet since I have to add the species, they're easy to implement though.


Eearslya:
This PR needs some **serious** work before it can even be considered. The code needs to be moved into the proper places, and there are *numerous* typos and poor English that need to be corrected. If and when all of that gets fixed, this will need discussion whether to implement.

Poojawa:
Ohboy.jpg.

Not even Citadel will ever implement this.

PR #651, opened 2016-10-14
Adds KH life crystal fluff/loadout item

Arokha:
It just exists as a fluff excuse as to how you get cloned offsite if you don't get cloned at CC, if you want an alternative. The little lore behind it is that it notifies an offsite agency if you die so you can get cloned by them. It has a little paper in the carrying case it comes with that explains that, and how to use it.

It's a uniform accessory, or you can just pocket it. They are unmarked, so finding a broken one shouldn't be very interesting evidence or influence much outside just the fluff idea of their existence.

In theory, I guess if you were on a combat mission or something, you could swap with your buddy and tell when they bite the dust, so that's a mechanical, if minor use.

(Also, it's a crystal, but no it's not magic. The manual goes into detail about the power source, transmitter, etc.)


Arokha:
I wanna change a sprite, so DNM for a minute.

Arokha:
I'll just take your thumbs up as approval.

PR #654, opened 2016-10-16
Fix flashlight spam

Eearslya:
Flashlights were not properly turning off when they ran out of battery, resulting in a nonstop spam of `The flashlight flickers before going dull.` every single second. Now they properly turn off, and will refuse to turn on without sufficient battery life.


SpadesNeil:
This is a Polaris bug, so we should fix it for Polaris. However I am fairly certain it already got fixed.

Eearslya:
https://github.com/PolarisSS13/Polaris/commit/0def0f6789374c9ff74ac5fbe97ac0e1daf1ad45

Indeed it has been.

PR #665, opened 2016-10-20
Whitelist and Custom Items gitignore update

SpadesNeil:
I *really* don't care if players can see these on the server Github. Not
only will this make things easier to update, it'll also stop this stupid
bullshit idea that the server is full of snowflakes with gun permits.
Full transparency!

Yes I'm aware we could just use filezilla to do this but screw that.

### DIFF
--- a/config/.gitignore
+++ b/config/.gitignore
@@ -1,3 +1,7 @@
 #ignore everything here, except subdirectories.
 *
 !*/
+#Also don't ignore these things because editing them manually on the server is a pain in the fucking ass.
+!custom_items.txt
+!custom_sprites.txt
+!alienwhitelist.txt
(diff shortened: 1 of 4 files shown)


JoanRisu:
Gosh darn it Ace, and after Arokha taught me how to connect to serverbox. >:C

SpadesNeil:
But this is easier and less messy. :(

Arokha:
Before you update the server next, I would delete these files from the server, and let the update pull recreate them.

SpadesNeil:
I'll do that, then.

Cameron653:
Might be a late reply, but the whitelist is working, it's just that diona
haven't been whitelisted. I can make a pr later to whitelist them.

PR #472, opened 2016-08-14
Emergency hotfix for alien whitelist

Cameron653:
- Unwhitelists languages
- Makes non-whitelisted races selectable

What a user will see
Non-whitelisted species: http://puu.sh/qAHq0/ea76a40270.png
Whitelisted species: http://puu.sh/qAHss/0e200688b8.png

This means that a user will be able to select and join as a non-whitelisted species, but not select a whitelisted species.

If they _do_ somehow get their save file changed to a whitelisted species (By admin abuse, or their file previously had the race selected) they'll be able to join in. Not a big problem seeing as admin abuse is the only way it'll occur, and not too many people have selected the races and can simply be told to change it.

Fixes #427

PR #667, opened 2016-10-20
Makes Diona a whitelisted race

Cameron653:
In case the poll passes, this PR will whitelist Diona.
Poll in question: http://forum.vore-station.net/viewtopic.php?f=20&t=807&p=5228#p5228

**_However_**, **this will not remove the Diona race from people's savefile if they've already saved a diona character in their character select menus**, as stated by PR #472. 

This means that users that already have Diona in their character saves will most likely need to change their savefile to another species or they'll be able to play it. Whitelisting still works as normal (Put their ckey and species name in the file and they'll be able to play that species) but it's just that preexisting diona players will have access to them until they change their savefile's species. 

The Polaris whitelist system, (to the best of my knowledge,) still isn't 100% working. As in, all races with the exception of humans are whitelisted no matter what without PR #472 if the whitelist is enabled. 

PR #472 did a hacky fix to make it work since a few races needed to be whitlisted ASAP and at the time no other solution other than 1. Having humans the only playable characters and 2. Have xenochimera as a playable character for everyone.

Side note: I still have no bloody idea how to fully fix the whitelist, and the hotfix was the best idea I could come up with at the time. It works, it doesn't change anything that commenting a few areas out, but the problem listed above (Users already having savefiles with a whitelisted race) is the only problem with the hotfix.

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/station_vr.dm
+++ b/code/modules/mob/living/carbon/human/species/station/station_vr.dm
@@ -300,7 +300,7 @@
 	spawn_flags = SPECIES_CAN_JOIN
 
 /datum/species/diona
-	spawn_flags = SPECIES_CAN_JOIN
+	spawn_flags = SPECIES_CAN_JOIN | SPECIES_IS_WHITELISTED
 
 /datum/species/teshari
 	spawn_flags = SPECIES_CAN_JOIN


code/modules/mob/living/carbon/human/species/station/station_vr.dm
```diff
@@ -300,7 +300,7 @@
 	spawn_flags = SPECIES_CAN_JOIN
 
 /datum/species/diona
-	spawn_flags = SPECIES_CAN_JOIN
+	spawn_flags = SPECIES_CAN_JOIN | SPECIES_IS_WHITELISTED
```
SpadesNeil: If these tags are the same in default station.dm, then this whole section can be removed.

Arokha: Also true, though it is kinda nice to already have the predefined 'hey, you wanna change something about diona for VS? do it here'

Cameron653:
I'd personally want to keep them there so if we ever need to add another
flag to the species it's right there, but I can remove it if you really
want me to.

Cameron653:
Poll is currently 18 yes, 12 no. Should this be merged?

SpadesNeil:
Aye, we may.

PR #673, opened 2016-10-21
Tweaks

Arokha:
Removes death message when in someone
Changes my spacesuit's name, cuz
Adds sprites for several missing suits for VS custom species (not custom sprites, they are taj and unathi)


Eearslya:
Shouldn't all the DMI files be `_vr`, too?

Arokha:
They are in folders that don't/won't exist on Polaris so there's no potential to conflict.

Regardless, I did not make them, I simply updated them.

PR #680, opened 2016-10-23
Fix teshari spawn bug (and lack of dark overlay) and micros(?)

Arokha:
Teshari thing is this: https://github.com/PolarisSS13/Polaris/pull/2667

Hopefully something in here does it for micros. Sort of shotgun approach. Better coded into Destroy() anyway, moving all the items out, in case held_mob was somehow set wrong. Has some debug code though if it still happens.

### DIFF
--- a/code/controllers/Processes/mob.dm
+++ b/code/controllers/Processes/mob.dm
@@ -12,12 +12,6 @@
 		mob_list = list()
 
 /datum/controller/process/mob/doWork()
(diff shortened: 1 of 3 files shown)


SpadesNeil:
Gateway map acting up. I can just force-merge this if this doesn't have other errors.

Arokha:
It went through now.

PR #682, opened 2016-10-23
Makes taur suits retain name when put in a cycler

killer653:
Adds a warning to taur voidsuits that they shouldn't be put in cyclers

People were putting them in cyclers and wondering why it was still only fitting taurs, so this adds a "Don't stick them in cyclers" warning.

Also makes it so when someone attempts to wear a hardsuit after it's been cycled/puts on on after it's been cycled, it'll go back to it's original name. (Taur specific engineering voidsuit/Taur specific medical voidsuit/etc)

### DIFF
--- a/code/modules/clothing/spacesuits/void/void_vr.dm
+++ b/code/modules/clothing/spacesuits/void/void_vr.dm
@@ -82,22 +82,26 @@
 	species_restricted = null //Species restricted since all it cares about is a taur half
 	mob_can_equip(var/mob/living/carbon/human/H, slot, disable_warning = 0)
 		if(istype(H) && istype(H.tail_style, /datum/sprite_accessory/tail/taur/horse))
+			name = "taur specific blood-red voidsuit"
 			icon_override = 'icons/mob/taursuits_vr.dmi'
 			icon_state = "syndie-horse"
 			pixel_x = -16
 			return 1
 		else if(istype(H) && istype(H.tail_style, /datum/sprite_accessory/tail/taur/wolf))
+			name = "taur specific blood-red voidsuit"
 			icon_override = 'icons/mob/taursuits_vr.dmi'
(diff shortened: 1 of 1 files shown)


SpadesNeil:
Why not just change the cycler to not allow taur suits to be inserted?

SpadesNeil:
Also @Cameron653 "fluff" tag denotes stuff used only by individual players, such as custom races, or custom items.

Cameron653:
Ah, sorry about the fluff tag. Didn't really think of anything else that would fit since it's just a small description addition.

I'm not exactly sure if I'd be able to make the suit cycler reject the taursuits from being in since it allows all "/obj/item/clothing/suit/space" objects, and the taur suits are (for example) "/obj/item/clothing/suit/space/void/merc/taur"

Arokha:
could just make it name the outputted taur suits properly so that people know. I mean it doesn't break them or anything, just doesn't name them right.

Cameron653:
Making it change their names would probably be tricky/more trouble than its worth, as the way the cycler handles changing their names in a way that really doesn't leave much room for changes.

Although, I could make it so that attempting to put it on would change the name back to as it should be, just in case someone cycles it.

(Cycler name changing code: https://github.com/VOREStation/VOREStation/blob/9c9293d4abc45593ee49840e3f602004b2d007fd/code/game/machinery/suit_storage_unit.dm#L975-L1077)

Cameron653:
Updated the PR name to reflect the main addition (The fact it renames it to the correct thing. The description change is extremely minor)

Arokha:
Needs EMT and crowd control suits

Cameron653:
Gong to need SilencedMP5A5 to sprite those up so I can add them

SpadesNeil:
Closing because over a month.

PR #684, opened 2016-10-23
Catgirl Icons

SpadesNeil:
Adds more catgirl icons.

Figured I'd do a PR for this early while I work on my spider map.


Cameron653:
There was a ) missing in line 24 that was preventing it from compiling, so I went ahead and put it in.

PR #691, opened 2016-10-26
Swap 'Give' dialog buttons

Arokha:
"Yes" on the left to be consistent with other dialogs. The number of times people (myself included) misclick because of this is impressive. Or just accidentally cancel out while typing.

### DIFF
--- a/code/modules/mob/living/carbon/give.dm
+++ b/code/modules/mob/living/carbon/give.dm
@@ -15,7 +15,7 @@
 		src << "<span class='warning'>You don't have anything in your hands to give to \the [target].</span>"
 		return
 
-	if(alert(target,"[src] wants to give you \a [I]. Will you accept it?",,"No","Yes") == "No")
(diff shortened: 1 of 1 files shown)


SpadesNeil:
It was set like this intentionally because people typing often take items they don't intend to.

What _should_ happen is 'cancel' should be the default option, prompting a, "So and so was offered an item, but was too distracted to notice."

Screemonster:
Really, the main issue with this is people handing people stuff when the other person's typing, so the popup is on the screen for all of .1 of a second before the default option gets picked by the player hitting space.

Arokha:
Unfortunately the leftmost button is the default and that cannot be changed. The overwhelming response in the Discord was to swap them anyway, though. I mean how often does someone give you something you don't want to accept? Is that a thing that happens? :v

SpadesNeil:
To be honest I've had that happen to me very frequently where I get handed things I don't want. :U

Arokha:
Do you care enough to block this? The consensus when I polled in the Discord was that people would prefer it changed. They were 'weh' about the default being different, but even with that, said they'd rather have it changed (myself included). So is your comment like just an observation or ...

Basically how much do you care? If nobody is really strongly opposed, then it seems like the consensus is that it's desirable to change.

PR #701, opened 2016-11-04
FBP/Synth wound sprite updater

Arokha:
It returned before it got to the bottom with robotic limb repair, which would leave the wound sprites on them unless an admin forced an icon update on them. So now we update them just like we do for organics a few lines below when wound levels change.

I cherrypicked this to Polaris too, but I dunno when they'd merge it so I'll put it here so we can update soonish.


Arokha:
Nobody is going to have any input on this anyway :v ...

PR #705, opened 2016-11-05
Discord integration

Leshana:
Please remind me which thing I forgot to do.

### DIFF
--- a/code/controllers/configuration_vr.dm
+++ b/code/controllers/configuration_vr.dm
@@ -1,0 +1,32 @@
+//
+// Lets read our settings from the configuration file on startup too!
+//
+
+/hook/startup/proc/read_vs_config()
+	var/list/Lines = file2list("config/config.txt")
+	for(var/t in Lines)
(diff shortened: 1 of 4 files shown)


JoanRisu:
The faxes go straight to Central Bridge, yes? Making sure.

SpadesNeil:
We should probably have a specific fax channel.

JoanRisu:
We already have too many channels. We should try not to over populate ourselves unless absolutely necessary.

Leshana:
The destination of the faxes is not part of this code, or any code.  That is a configuration within discord.   
It uses discord webhooks, which can be assigned from channel in the discord server configuration.

SpadesNeil:
Well I know _that_, I just mean we don't want fax messages spamfucking our admin channel. :U 

Also we should probably bump this upstream to Polaris too at some point. It's useful.

SpadesNeil:
So once this is uploaded what needs doing in the config?

PR #712, opened 2016-11-08
fixes mob_can_equip call in fluff items

### LINKED ISSUE #711: fluff items with mob_can_equip restrictions break autoequipping
Fluff items with the mob_can_equip restrictions to ensure that only one person can use them interfere with the way that items are equipped in general, resulting in things being equipped in the wrong, and sometimes impossible, slots.

for my jumpsuit to go on my jumpsuit slot when I hit ctrl-e, and not to be able to equip it to my back slot when there's already a backpack in it.

#### What actually happened
The suit auto-equips to the back slot. Even when there's already an item there. The auto-equip item button (next to change hand on the hud, or ctrl-e) will try to equip it there regardless of whether there's an item in the jumpsuit slot.

This results in some pretty weird stuff on spawn:

image

#### Steps to reproduce
edit /obj/item/clothing/under/fluff/screesuit so it checks for your ckey, compile
attempt to equip it with ctrl-e in-game, or click on your back slot

#### Additional info:
The offending section of the code appears to be this part:
For /obj/item/clothing/under/fluff/screesuit

```
	mob_can_equip(var/mob/living/carbon/human/H, slot, disable_warning = 0)
		..()
		if(H.ckey != "scree")
			H << "<span class='warning'>Are you just going to tape them on or what? This isn't gonna work.</span>"
			return 0
		else
			return 1
```

Commenting this block of code out makes the jumpsuit behave as normal, not equip itself in the back slot, and stay in the hand when I attempt to autoequip when already wearing a jumpsuit.

Screemonster:
Fixes the callback in the mob_can_equip call in Scree's feathers, and Scree's, Joan's and Aronai's voidsuits.

Have only been able to completely test mine for obvious reasons, but they work as expected.
Assigning Joan and Aro to do their own testing on their own keys and make sure I've not broken their stuff.

Fixes #711.

### DIFF
--- a/code/modules/vore/fluffstuff/custom_clothes_vr.dm
+++ b/code/modules/vore/fluffstuff/custom_clothes_vr.dm
@@ -436,12 +436,11 @@
 	species_restricted = null
 
 	mob_can_equip(var/mob/living/carbon/human/H, slot, disable_warning = 0)
-		..()
 		if(H.ckey != "scree")
 			H << "<span class='warning'>Your face and whoever is meant for this helmet are too different.</span>"
 			return 0
 		else
-			return 1
+			return ..()
 
 //scree:Scree
 /obj/item/clothing/suit/space/void/engineering/fluff/screespess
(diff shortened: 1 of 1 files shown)


Arokha:
I'm having some weird item with mine. I wonder if Joan has any with theirs.

Arokha:
Ok I gave you some bad advice about this. Apparently there's a more consistent way to do it. I'm making another PR since I can't update this one, as I don't have write on your repository.

PR #713, opened 2016-11-08
Fix world/Topic()'s manifest to include synthetics like datacore.dm

Leshana:
* The world/Topic() bot integration API's manifest output uses the same logic as datacore's get_manifest(), except it omits the part that includes synthetics.  The `set_names` variable clearly shows bots are intended to be included, they are just arent.
* Copying the logic from datacore.dm ensures a consistent view of bots and AI.


Arokha:
Polaris merged upstream and you asked me to close this so we can just take theirs.

PR #733, opened 2016-12-01
Xenochimera bugfix and feralness overhaul

Screemonster:
Fixes the part about regenerating that breaks if there's more than one in the universe. We might be able to actually take whitelist applications now!

Feralness has had a massive overhaul - it is no longer a binary on/off thing purely dependent on hunger. Now, if a chimera is sufficiently hungry or hurt, there is a chance they will snap (with appropriate messages) with severity dependent on what made them flip out.
If hungry, they just get more and more feral over time. Injuries give a big initial spike that piles more on periodically if it isn't fixed.
Once the source of the chimera's stress has been corrected, feralness will tick down over time. _Staying_ stressed and feral is a touch more sensitive than _becoming_ so, so chimera can't just wait for their damage to drop a couple of points, they have to actually retreat and recover properly.

Any stimulants that inflict jitteriness will also cause feralness for as long as the xenochimera is still jittery. Do not let Scree down an entire pint of coffee unless you want him bouncing off the walls and gnawing on the tesharis.

While feral, xenochimera will start to hallucinate. These hallucinations go away if someone else is in the room in order to not interfere with scenes (you'll still occasionally get a nudge to eat that person) but otherwise the only way to be free of the scary confusing metal box your instincts don't know how to handle is to go and hide somewhere. You're considered hidden if you're somewhere dark, being in a container counts. (basically, if it's not light enough for a diona to heal, it's dark enough for a chimera to be comfortable)

~Makes Scree a permanent admin on every server forever~ A few tweaks to basic species stats:
- Set darkvision to match tajarans since they have to hide in the dark while feral.
- Set speed about halfway between human and tajaran. Kitties and Teshari are still the speed kings.
- Buffed brute resistance to match Unathi so they're _slightly_ tanky to compensate for the fact that they flip out and have to hide in a dark hole when they get hurt.
- Nerfed burn resistance in line with Tajara. Fire hurts!

### DIFF
--- a/code/modules/mob/living/carbon/human/species/species_vr.dm
+++ b/code/modules/mob/living/carbon/human/species/species_vr.dm
@@ -9,4 +9,4 @@
 
 	//This is so that if a race is using the chimera revive they can't use it more than once.
 	//Shouldn't really be seen in play too often, but it's case an admin event happens and they give a non chimera the chimera revive. Only one person can use the chimera revive at a time per race.
-	var/reviving = 0
+	//var/reviving = 0 //commented out 'cause moved to mob
--- a/code/modules/mob/living/carbon/human/species/station/station_special_abilities_vr.dm
+++ b/code/modules/mob/living/carbon/human/species/station/station_special_abilities_vr.dm
@@ -12,7 +12,7 @@
 		var/mob/living/carbon/human/C = src
 		var/nutrition_used = C.nutrition/2
 
-		if(C.species.reviving == 1) //If they're already unable to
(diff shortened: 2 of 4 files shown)


Screemonster:
Oh, the raise on the cap for cold-halloss was to see if that stops the thing where Scree kept hitting 100 in a cryotube, falling over, dropping below, standing up again, dropping as soon as he took more, and spamming "can no longer stand, collapsing!" messages 5ever.

Cameron653:
Ah, that makes sense.

PR #758, opened 2016-12-12
Special Stomach Interactions

### LINKED ISSUE #559: "allow struggling escape" option on bellies.
#### Brief description of the request

Citadel has (or had, I dunno if this is still in their code) an option for predators to set whether or not prey would be able to escape from their tummies. By default they were stuck, but preds could allow them to wriggle free or get fulltoured if they struggle. I think something like this would be a neat feature to have.
#### What I'm proposing

Options on bellies, with the default option being to just keep prey where they are, for predators to have their stomachs react to prey struggling inside them.
%chance of releasing the prey.
%chance of setting the stomach to digest/absorb. (with a short message like "the [stomach] gets more active in response to your squirming..."
%chance of moving the prey to another designated stomach.
All default to 0, if none of these happen it should display the normal message about useless squirming.
#### Example of where this would be awesome

A macro catcubus finds a little micro, scoops them, and puts them in her "mouth" belly and subjects them to a whole lot of teasy tongueplay for an extended period of time. Her prey try to wriggle their way out of her mouth with the struggle button (set to 10% chance of escape), but when they start kicking up a fuss she accidentally swallows (set to 10% chance of transferring them to her stomach) and the little snack ends up becoming lunch rather than a toy. She was only playing!

People could also have a kangaroo or something with a "pouch" belly that just lets people out if they struggle, I guess.
Absorbed people won't be able to escape as they can't resist(). Pudge doesn't get to just leave. :p

Cameron653:
Adds various special stomach interactions

- Adds escape chance, a variable that can be set from 0 to 100 % chance. When a prey struggles inside of a stomach, they have that % chance to escape. Takes as long as the escape time is set. (This is set so 

- Adds digest chance, 0-100% chance the pred's stomach begins to digest the prey when the prey resists.

- Adds absorb chance, does the same as above but with absorbing.

- Allows preds to set escape time when a prey resists and the escape chance is proc'd. Ranges from 10 (one second) to 600 (one minute.)

- Makes the escapable variable toggleable. This enables the interactions mentioned above.

- Allows preds to set the stomach to transfer the prey to another stomach if the prey struggles. 

- Transfer chance to effect the above.

BUGS:

(Maw leads to stomach in this scenario) If a predator deletes their stomach but doesn't change their transferlocation,  then when a person is transferred from their maw to their "stomach" they're trapped in limbo as the predators "stomach" doesn't exist anymore. This keeps the prey with the predator, but makes them unable to leave without admin intervention.

This is a bug that would rarely (if ever) occur. 

Edit: Due to being unable to solve the bug, I added warnings when a stomach is deleted and when a transferlocation is selected. If anyone can find a solution then please tell me/add it into the code. I've tried everything I can think of and I can't get it to work. All sanity checks I put in just end up turning back as false, even if the vore organ _does_ exist. Other than that, everything else works.

Resolves  #559

### DIFF
--- a/code/modules/vore/eating/belly_vr.dm
+++ b/code/modules/vore/eating/belly_vr.dm
@@ -390,7 +390,7 @@
 	var/strpick = pick(struggle_sounds)
 	var/strsound = struggle_sounds[strpick]
 	playsound(R.loc, strsound, 50, 1)
-	
+
 	if(escapable) //If the stomach has escapable enabled.
 		R << "<span class='warning'>You attempt to climb out of \the [name].</span>"
 		owner << "<span class='warning'>Someone is attempting to climb out of your [name]!</span>"
@@ -405,7 +405,7 @@
 					return
 				else if(!(R in internal_contents)) //Aren't even in the belly. Quietly fail.
 					return
-			else //Belly became inescapable. 
+			else //Belly became inescapable.
 				R << "<span class='warning'>Your attempt to escape [name] has failed!</span>"
 				owner << "<span class='notice'>The attempt to escape from your [name] has failed!</span>"
 				return
@@ -413,16 +413,18 @@
 		else if(prob(absorbchance)) //Next, let's have it run the absorb chance.
 			R << "<span class='warning'>In responce to your struggling, \the [name] begins to get more active...</span>"
 			owner << "<span class='warning'>You feel your [name] beginning to become active!</span>"
-			var/datum/belly/B = check_belly(owner)
-			B.digest_mode = DM_ABSORB
-			return
+			for(var/K in owner.vore_organs)
+				var/datum/belly/B = owner.vore_organs[K]
+				B.digest_mode = DM_ABSORB
+				return
 
 		else if(prob(digestchance))
 			R << "<span class='warning'>In responce to your struggling, \the [name] begins to get more active...</span>"
 			owner << "<span class='warning'>You feel your [name] beginning to become active!</span>"
-			var/datum/belly/B = check_belly(owner)
-			B.digest_mode = DM_DIGEST
-			return
+			for(var/K in owner.vore_organs)
+				var/datum/belly/B = owner.vore_organs[K]
+				B.digest_mode = DM_DIGEST
+				return
 		else //Nothing interesting happened.
 			R << "<span class='warning'>But make no progress in escaping [owner]'s [name].</span>"
 			owner << "<span class='warning'>But appears to be unable to make any progress in escaping your [name].</span>"
--- a/code/modules/vore/eating/belly_vr.dm
+++ b/code/modules/vore/eating/belly_vr.dm
@@ -22,6 +22,9 @@
 	var/immutable = 0						// Prevents this belly from being deleted
 	var/escapable = 0						// Belly can be resisted out of at any time
 	var/escapetime = 600					// Deciseconds, how long to escape this belly
+	var/digestchance = 0					// % Chance of stomach beginning to digest if prey struggles
+	var/absorbchance = 0					// % Chance of stomach beginning to absorb if prey struggles
+	var/escapechance = 0 					// % Chance of prey beginning to escape if prey struggles.
 
 	var/tmp/digest_mode = DM_HOLD				// Whether or not to digest. Default to not digest.
 	var/tmp/list/digest_modes = list(DM_HOLD,DM_DIGEST,DM_HEAL,DM_ABSORB,DM_DRAIN,DM_UNABSORB)	// Possible digest modes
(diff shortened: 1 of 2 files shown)


SpadesNeil:
I need to alter vore mobs to be able to have the same toggle.

SpadesNeil:
I feel like there's something missing here. We added a var to make bellies escapable, but now what? Does resisting now cause you to crawl out, cause a chance to crawl out, or does it do nothing at all? Because I don't see any code for that here.

Cameron653:
The code was already there, but it was just that a button to toggle it was
nevernimplemented. Resisting begins a 1 minute countdown. After the
countdown it lets the prey crawl out.

SpadesNeil:
I do not think that does satisfy the issue. That 1 minute timer is not based on random chance. That timer exists only for people to ICly escape stomachs of AFK predators.

SpadesNeil:
Now we're talking.

Cameron653:
Not fully working at the moment and I have a bit more to add.

Current bugs in the code:

The "B.digest_mode = DM_DIGEST" and "B.digest_mode = DM_ABSORB" isn't working at the moment. 
Throws up a runtime. My problem currently is figuring out how to define the predator's stomach to have it change the digest_mode.

Cameron653:
Alright, did some more edits and additions.

- Allows preds to let one stomach lead to another.

- Adds transferchance. Effects the above.

Only one "real" bug that I have currently found.

(Maw leads to stomach in this scenario) If a predator deletes their stomach but doesn't change their transferlocation, then when a person is transferred from their maw to their "stomach" they're trapped in limbo as the predators "stomach" doesn't exist anymore.

This bug can be fixed with a sanity check to see if the "stomach" still exists, but I need to figure out how to actually get the sanity check working without always returning false. I'll work on this some more tomorrow.

SpadesNeil:
Waiting on that sanity check before this is merged.

Cameron653:
I really don't know if I can even get a sanity check in there. I've tried everything I can think of, but it always turns up false if I add in a sanity check, even if the vore organ exists.

The bug is specific enough that it will most likely never happen unless someone is intentionally attempting to do it, and when you set a transferlocation. I've made it so it warns the user to remove the transferlocation if you delete the stomach, and also warns the user to not delete their transferlocation until they set it to "none" where it's used.

SpadesNeil:
Well I'm going to merge this but submit an issue referencing the bug, and this PR.

code/modules/vore/eating/belly_vr.dm
```diff
@@ -389,6 +393,57 @@
...
+		else if(prob(absorbchance)) //After that, let's have it run the absorb chance.
+			R << "<span class='warning'>In responce to your struggling, \the [name] begins to get more active...</span>"
+			owner << "<span class='warning'>You feel your [name] beginning to become active!</span>"
+			for(var/K in owner.vore_organs)
+				var/datum/belly/B = owner.vore_organs[K]
+				B.digest_mode = DM_ABSORB
+				return
+
+		else if(prob(digestchance)) //Finally, let's see if it should run the digest chance.
+			R << "<span class='warning'>In responce to your struggling, \the [name] begins to get more active...</span>"
```
SpadesNeil: *response

Also is there a check to see if we're already digesting? Else this gets spammy.

code/modules/vore/eating/belly_vr.dm
```diff
@@ -389,6 +393,57 @@
...
+				var/datum/belly/TL = B.transferlocation
+				if(TL != null)
+					B.internal_contents -= R
+					T.internal_contents += R
+					R << "<span class='warning'>Your attempt to escape [name] has failed and your struggles only results in you sliding into [owner]'s [transferlocation]</span>"
+					owner << "<span class='warning'>Someone slid into your [transferlocation] due to their struggling inside your [name]!</span>"
+					return
+
+		else if(prob(absorbchance)) //After that, let's have it run the absorb chance.
+			R << "<span class='warning'>In responce to your struggling, \the [name] begins to get more active...</span>"
```
SpadesNeil: *response

Also is there a check to see if we're already absorbing? Else this can get spammy.

code/modules/vore/eating/belly_vr.dm
```diff
@@ -22,6 +22,11 @@
 	var/immutable = 0						// Prevents this belly from being deleted
 	var/escapable = 0						// Belly can be resisted out of at any time
 	var/escapetime = 600					// Deciseconds, how long to escape this belly
+	var/digestchance = 0					// % Chance of stomach beginning to digest if prey struggles
+	var/absorbchance = 0					// % Chance of stomach beginning to absorb if prey struggles
+	var/escapechance = 0 					// % Chance of prey beginning to escape if prey struggles.
+	var/transferchance = 0 					// % Chance of prey being
+	var/transferlocation = null				// Location that the prey is released if they struggle and get dropped off.
```
SpadesNeil: This transfer code needs to be disabled until it's fixed.

PR #771, opened 2016-12-19
Macro/micro damage scaling

nightwingthedragon:
One-line fix to allow damage to scale with the recipient's scale. Micros take 4x damage, small takes 2x damage, normal takes 1x damage, big takes 2/3x damage, macro takes 1/2x damage. This can be changed easily.

DO NOT MERGE. UNTESTED CODE. PROOF OF CONCEPT ONLY.

### DIFF
--- a/code/modules/mob/living/damage_procs.dm
+++ b/code/modules/mob/living/damage_procs.dm
@@ -13,7 +13,7 @@
 		world.log << "## DEBUG: apply_damage() was called on [src], with [damage] damage, and an armor value of [blocked]."
 	if(!damage || (blocked >= 100))
 		return 0
-	blocked = (100-blocked)/100
+	blocked = (100-blocked)/100 * 1/size_multiplier // VOREstation Edit - NW was here - scales damage with recipient's scale
 	switch(damagetype)
 		if(BRUTE)
 			adjustBruteLoss(damage * blocked)


nightwingthedragon:
For the record, the only reason it's untested is because DreamMaker or Source Tree goofed on me and forgot to include whichever file contains brains. Useful!

Screemonster:
Rather than scaling the amount of damage taken or fucking with the damage procs that may be affected by all sort of changes by polaris later, why not just change their maxhealth var?
Setting a micro's maxhealth to 50, for instance, will make them about as vulnerable as a teshari. It'll also have the effect of not being QUITE so crippling since the death threshold is set in the config as -100 and this applies regardless of species, which means they won't bleed out or have limbs blown off from massively amplified damage from all sources.

nightwingthedragon:
That's a good idea too, but I'm not sure how kindly Byond would take to changing max health on the fly. It certainly couldn't be done in a one-line fix like this one! We'd need to recalculate current health based on max health with every time somebody eats a 'shroom, drinks some macrocillin, gets hit by a shrink-ray, or uses the OOC controls, and I have no idea how the game would react to that. Potentially, getting smaller while slightly damaged could break bones!

Screemonster:
I just tested it on the live server. I varedited the shit out of myself. The only complication was that setting my maxhealth to 10 (not proposing to drop it that low) left me stunned for a longass time even after I set it back to 100 on account of having racked up such a stupendous shock_stage. but no, changing maxhealth even when injured doesn't cause injuries.

The maxhealth var is used when updating your health every life tick, essentially just setting it as health = (maxhealth - the sum of all damage types)
Health, if negative, is then used to calculate whether you should rack up shock, go into crit, fall unconscious, die, and so on.

nightwingthedragon:
Ah, interesting! I wonder if it should be compatible with this idea...

Repede:
Tesharis already have 50hp though.. So what would happen if you were a micro teshari. Stay at 50?

nightwingthedragon:
@Repede Well, using purely the code I've listed there, a micro Teshari would still have 50 max HP, but everything would hit them 4 times as hard. If we were going to edit the max health instead/too, we'd likely balance it a little better than just "size * HP", to prevent 25HP micros that can be easily killed/digested, and 12.5HP teshari micros. And to prevent 200HP giants.

Repede:
Yeah I meant that for the max hp change suggestions. IMO anything related to size change changing can unbalance the game quickly. Just be a scientist with size gun or a chemist with shrink smoke grenade. Make all of security tiny and stomp on them / etc. :P It's a lot to consider honestly. But at the same time it would be more realistic for sure.

SpadesNeil:
I like that this scales damage _taken_ instead of damage _dealt._

nightwingthedragon:
If there's no problems with conflicts or anything like that, I'd be happy to go ahead and merge it in. But, of course, it needs testing which I can't conduct due to the fact that this branch I used didn't download all the right files. If someone had the time to test it and confirm that it works roughly as intended, I'd appreciate that.

SpadesNeil:
The reason this sat for so long is I'm not the one who marked it "do not merge." Ergo, I was waiting for you to remove those tags.

nightwingthedragon:
Unfortunately, I couldn't remove the tags on account of being unable to test this code due partially to having exams this week (both now out the way, one more to go) and partially because when I made this PR, Github didn't download some of the required files to run the station server itself, or even to compile. I literally made a change and saved without being able to compile. I need somebody else to test it out.

SpadesNeil:
Okay so the verdict is on on this. We're removing guns from micros. However, we have a lot of balance changes to consider before actually making the change. (You didn't think I was ACTUALLY going to make the change without something to make up for it, did you?)

This will be merged with the gun update for micros alongside those balance changes. This is just one of them.

Vorrarkul:
> (You didn't think I was ACTUALLY going to remove micros' use of guns without something to make up for it, did you?)

Well, kind of. That was at least my inference from the [thread](http://forum.vore-station.net/viewtopic.php?f=20&t=835) when you said we would not be discussing alternatives.

SpadesNeil:
I did that on purpose to make sure people wouldn't vote based on "oh there will be changes that will magically make everything okay" when everybody had no idea what those changes would be. (I still don't know all of them. I just know we have to have them or it would be madness.)

Repede:
Wouldn't adding both changes nerf the micros to nothingness though? First they wouldn't have access to ranged weapons. And as an extra they would get obliterated by pretty much anything. I mean it makes sense to be fair. But I just fear it could be too much at once.   At the same time my opinion doesn't matter much since so take it as a grain of salt.

SpadesNeil:
At the end of the day, micros are getting a nerf. The question is now what do they get in exchange for that nerf? What makes them continue to be fun? Because right now, the one-sidedness of micros in fights is not fun. Balance isn't about making all players have an equal chance in all combat situations, it's about making all the players have fun regardless of play style.

One of those things is to be given easily manufactured mini-mechs. Another is to raise the evasiveness of mobs based on lower scale. Not all ranged weapons are being removed from their ability to use, either. We intend to whitelist some items depending on their application, such as the shrink ray, or dart gun.

Repede:
I had forgotten about the new tiny mechs. I think all in all, those changes all together makes it reasonable. Can't wait to see what will come out of this.

Arokha:
Should probably only merge this once micro mechs are AVAILABLE. They are IN THE CODE but are not ON THE MAP and cannot be constructed.

Razer69:
Playing a micro myself, I'm actually very saddened I'm not nerved to shit yet!

Affecting both damage dealt and damage taken at the same time makes the most sense, as both vulnerability *and* strength would change with size. Attacking someone your own size would inflict normal attack damage then.

Also hi, I'm new!
Am I free to open PRs for proposal?

SpadesNeil:
The reason it isn't done yet is we haven't properly worked out all the caveats of how we're going to do this yet over in https://github.com/VOREStation/VOREStation/issues/909

Arokha:
@Razer69 If you don't like wasting time you should probably ask in the #development channel in Discord first!

SpadesNeil:
Ooooolllllllllld PR is Ollllldddddd. But we can re-open it if we need it.

ArbyTheDragon:
I just had a thought. If micros were ever able to use a gun, one they'd be able to use would definatly be the Kolibri 2.7mm pinfire pistol, seen here http://www.xboxachievements.com/images/news/kolibri4_001.jpg

PR #784, opened 2016-12-29
Fixes birds meowing

Cameron653:
Fixes birds from meowing when their friend is hurt.

Looking back at it, I'm not 100% sure why I have them a subclass of cats. I think I might've done it at the time so they'd have the same flee code as cats and also go after micros/mice the same way as cats.


SpadesNeil:
Yes, why _are_ they a subclass of cats? They shouldn't be.

SilverTalismen:
Well atleast its fixed Meow

SpadesNeil:
get out

PR #803, opened 2017-01-06
Feature: Extended Character Customization

OrbisAnima:
Added more options to the character customization screen. These options
have dna support, clothes will hide them but can be configured to not,
will only work with races that can use underwear, can be customized
further per player, gender or other variables used by the others
customization systems. I tried to respect as much as I could VR syntaxes
and special areas of code, imitating commentaries, the code suffered
four itterations on the servers I worked, so this one version is already
optimized as much as I could think of. Underwear will also block the
customization, but of course this can be changed with simple vars, and
removing underwear and clothes will reveal it.

Credits for code and port go to OrbisA, credits for sprites go for
BigSky, credit for hairy one goes to Xsdew, credit for robot parts goes
to Tuesday. Additional parts were coded in such a way to be easy to add,
encouraging participation by players.


Screemonster:
I'm going to admit that I don't like the implications of basically any PR containing the phrase "Slime Dripping Vagina".

For what it adds in the form of a few pixels, it kinda takes away in the form of.. well, the fact that you mentioned in the discord that you had to obfuscate the name of the commit so they wouldn't see that it's full of genitalia, so github _wouldn't notice_ that we're breaking the TOS. Whether they notice or not, however small the chance of their noticing, whether Eros or Citadel have had takedowns or not, _it still breaks the Github TOS_. On that reason alone I'm against merging this. Especially if you're "encouraging participation by players", and as a maintainer who does a lot of fluff ear/tail/whatever parts for people I have better things to be doing than fending off requests for custom dicks.

That, and it seems like "just customisation" until suddenly the players who don't want to deal with it have it thrust in their faces every time a character using it is on screen. Text is easy enough to ignore or scroll past, or they can LOOC and the person in-game can just stop making references to their MAGNUM DONG to their co-workers, but if it's on the sprite, it's just... there, whether people want to deal with it or not.

There's also the question that some people have housemates or live with relatives, and while spotting weirdness on our current setup would require specifically reading the tabs and text window and the rest just looks like space station, it might be harder for someone to explain to housemates/parents/friends why their screen has two naked guys running around with their dicks visibly out and a chick with tentacles sprouting like that bit from Drawn Together.

tl;dr for what tiny functionality this adds, there are way too many downsides for me to support a merge. Even if I _do_ like boobies.

OrbisAnima:
Its fine Scree, I only changed the title per request, literally, I was asked to do the pr with no lewd title.

I never saw the TOS of github to that extent to see there are no lewd clause, but I will believe you on that, in any case, if you dont like something, remove it, if the players nag you, then just say no, making something hard for people to contribute so you dont have to work as a mantainer is an stupid principle, on a coding standpoint, I do can empathize with not wanting more work or getting nagged by players who made something terrible, but really, never make code that people cant easily add too, or you will end with stale stuff.

And yes, its a problem, using a system like this pulls the codebase a little bit more into the NSFW line, but the same could be said of the title screen. A good idea would be to keep the nude permit as it is, so it doesnt allow to render the genitals, after all, I made clothes so they need two sets of vars to enable or not those bits (show boobs and show genitals respecitvely). Even then underwear also hides them, and do remember that you have a default dong for guys too, that is visible by default (which that groin should be changed to the female one if this system is merged).

the tl,dr is that those functionalities can, and I expect they will, be adapted to the vision of the server as appropiate, I wanted bigger boobies but what you currently have is the max I was allowed on the other server.

OrbisAnima:
Also, checking on the ToS, https://help.github.com/articles/github-terms-of-service/ only section of porn goes in sector G, part 7, which is basically a clause that says they will remove anything they want if its too offensive, as other servers have similar content and way more over the top stuff, I would say the limits are greatly exageratted.

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -1452,6 +1452,7 @@
 		return
 	hide_underwear[UWC.name] = !hide_underwear[UWC.name]
 	update_underwear(1)
+	update_genitals_showing()
```
Arokha: Mark this line with a comment please, for future ease of resolving merge conflicts.

code/modules/mob/living/carbon/human/update_icons.dm
```diff
@@ -109,31 +109,32 @@ Please contact me on #coderbus IRC. ~Carn x
...
-#define HEAD_LAYER				20
-#define COLLAR_LAYER			21
-#define HANDCUFF_LAYER			22
-#define LEGCUFF_LAYER			23
-#define L_HAND_LAYER			24
-#define R_HAND_LAYER			25
-#define FIRE_LAYER				26		//If you're on fire
-#define TARGETED_LAYER			27		//BS12: Layer for the target overlay from weapon targeting system
-#define TOTAL_LAYERS			27
+#define GENITALS_LAYER			4
```
Arokha: Comment

code/modules/mob/living/carbon/human/update_icons.dm
```diff
@@ -332,6 +333,7 @@ var/global/list/damage_icon_parts = list()
 
 	//tail
 	update_tail_showing(0)
+	update_genitals_showing(0)
```
Arokha: update_genitals_showing(0) 3 times, comments.

code/modules/mob/living/carbon/human/update_icons.dm
```diff
@@ -1154,10 +1158,63 @@ var/global/list/damage_icon_parts = list()
...
+	if(update_icons)
+		update_icons()
+
+//Vorestation Edit - END
+
 //Human Overlays Indexes/////////
 #undef MUTATIONS_LAYER
 #undef DAMAGE_LAYER
 #undef SURGERY_LEVEL
+#undef GENITALS_LAYER
```
Arokha: Comment

OrbisAnima:
Understood, let me fix it

Arokha:
From a purely code perspective, this seems fine.

SpadesNeil:
Oh good this is already marked as "Requires Discussion."

I do have concerns with this and it's not because of what the code does but mainly because we're _already_ toeing a thin line with Github's policies on what we're allowed to upload. My concern is this goes overboard in terms of how smutty it is.

OrbisAnima:
I am aware I am not partial in this discussion, but I should point out that there was a server that had a rape system, and many people tried to take it down, but instead of making an issue with github the people involved opted to go and hack the Host, so that should be taken in consideration on how hard is to bring that ToS active.

SpadesNeil:
That's not a case of people not bothering to go to Github, that's a case of people preferring to take matters into their own hands like retards.

SpadesNeil:
That said, robo boobs are neat.

Dhaeleena:
I don't want to be all like _"We have discussed this already a thousand times reeeee!"_ But i do feel like there have always been similar subjects, Dicks, dildos, bodily fluids stains, filling glasses with things, Transferring fluids, etc etc. And i will always stand on my idea that we don't need those things, Not just because they /might/ irk some people if not because we are a HRP server and those kind of things are irrelevant...As if we were to add Lifeweb's sex mechanics, it wouldn't be too useful..Then again boobs are good...And most if not all of our races already have defined boobs. Then again that's my opinion and only my opinion and a good counter argument could change my mind.

OrbisAnima:
Well, the good thing is that body fluids, glasses and transfering are not involved, this mod is purely cosmetic, and will remain so, at least by my hand.

But, this thing is modular, I am seeing most of you like the boobs, so we can implement boobs only, but, a little suggestion here, regarding genitals, go for the normal ones, human parts, feline parts, canine parts, I would suggest the tiny tentacle ones too, the hairy ones, maybe leave the extreme ones you dont like aside, maybe get spriters from here to add in those, trust me on that bit, it seems like a very small thing, but its worth it for customization.

And... do remember we can say which clothes* will show what naugthy bits or not. The system is to be adapted to the server, and not the other way around, I dont want to see giant dicks in the hall neither, thats why I suggest, with some emphasis, that a nude permits show breasts but not genitals. 

*Uniform, Exosuit, and underwear slots only.

OrbisAnima:
that said, its a bit of a pain, but I can change the d. and v. word everywhere, and use flanderization :P

BothNevarBackwards:
However, I think that some genetalia which would be hidden normally and only visible when erect. Other less visible genetalia exist too. Should some of the ones that are very visible be allowed with nudity permits as the permits are only granted to those with more subtle genitals.

OrbisAnima:
That would be expanding the system, could be done of course, but the system is not the expanded version that citadel based on this, but a version that doesnt use an excitement value.

This is a system that shows stuff at... rest, except the tentacles, becouse they are animated.

OrbisAnima:
Another good selling point, and more vore oriented, is that this system has already all the tricky work done, in case you guys want to expand on BV, CV and similars, specially of note that since there is a system to hide these stuff, all uses only one layer, there is optimization going on and several ways to hide it, without breaking character, can enhance the visual aspect of the game.

Vorrarkul:
Well, with the idea of modularity, and our community's interests, in mind; could this system be used to implement large bellies as well, perhaps in place of the genitalia option, which can then be activated/deactivated by an 'actual' belly when filled or emptied?

OrbisAnima:
I dont think you would need to remove genitals for that, but if you should follow logic by what the system does, they look more alike to breasts than genitals, yet bellies would need to use a new var, otherwise underwear would block them, and that would make little sense.

Additionally, you would still have to bind the system with the weight system. One thing is being modular, another is being magical, but nice try.

But yes, I could see with ease a show_belly var for clothes, adding that to roll jumpsuit and coats, and just adding a similar code to the breasts one on the update_icons.dm, leave the list of bellies not on the control of the player but in the control of the weight system, and being super careful to bind the check per race not to can use underwear but to a different flag. You could also be careful not to use the current layers I am using for this system, which are meant to treat everything as genitals, you know, things that are hidden under layers of clothes and underwear, etc, which means they go even under the taur parts (no way around this without a several performance hack), but to give bellies its own layer where it went to where everyone believed it should go.

All in all it would be coding an extension of the system, and not a simple port, which is why removing genitals for it sounds super weird...

BothNevarBackwards:
Would belly be based on weight or nutriment?

OrbisAnima:
Would depend on who writes the code for it, but I imagine it would make sense basing it on weight.

Going back on point, the more I see Git ToS, the more I am convinced that they can pretty much cease service if they want to or not, since limits or examples are not established.

Screemonster:
I just had yet another argument in the Discord tonight with someone wanting a gigantic hyper dick at all times and their counter to not everyone wanting to see that on their screen whether they want it or not was "if they don't like it don't look at it" followed by "this is a vore server".

If this is added, it's not a question of IF it'll be used to trample over people's boundaries, it's a matter of WHEN. Just like that asshole who emoted walking into the bar and taking a dump on the table, but the difference in _that_ case was that they actively had to go out of their way to do it and we could very easily make a case for telling them to cut that shit out. Same with the incontinent little girl walking the station peeing everywhere that finally got underage characters officially given the boot.

Frankly, a couple of pixels attached to the lower half of the sprite isn't worth the headache.

OrbisAnima:
And... what if we added something like the hud vision? A flanderization if you will. 

Basically, we put the whole system behind a toggle for vision as if it was a hud like the security huds you can see or not with the glasses, except this would be for this content.

SpadesNeil:
That sounds very hacky and even if it works without being hacky it still feels dishonest to the players who might use those sprites. It will also cause controversy in future like "Why can't we add scat and other horrible things? We'll just make it hud-based." It's not okay because people will react to something that other people can't see at all.

OrbisAnima:
It was an idea, I wouldnt call it hacky but I can see an scalade with people saying "Hey, I dont want to see x, make it hud based", even if we make those layer based, it would still be a pain.

SpadesNeil:
Ehhhhhh. The admins aren't okay with this, myself included. We have that gut feeling that this is going to backfire. I appreciate the hard work you put into this. Code-wise, it seems very well done. Unfortunately we're just concerned this will cause more problems than it's worth, and so we're going to close this.

Hopefully the code can be used for something else in future with similar mechanics but different purpose.

OrbisAnima:
No problem, and I expect it will be :D

PR #805, opened 2017-01-06
Fix: Added minimun age to 18, to go with the server policies

OrbisAnima:


OrbisAnima:
No problem, sorry for making two prs, but I though it would be better than adding history to unrelated files and then running the risk fucking up the updates

PR #809, opened 2017-01-07
Several map changes.

Dhaeleena:
Armory partially re-done.
Tactical Storage is now actually /Tactical/
Fixed camera blind spots with more cameras!
Added Re chargers to medbay so Defibs power cells can be recharged.
Fixed several windows, Walls and other misc issues around the map.
Added Camera to the arrivals Shuttle so the AI can finally use its Holopad.
Removed cameras from the dorms.


SpadesNeil:
This looks good. I think I'm going to officially stop updating our map with the Polaris map barring major updates like atmos revamps or something.

Arokha:
Can just copypaste interesting sections from theirs. Though a whole new map would be fun too. I VOTE SPACESHIP. :3

PR #822, opened 2017-01-11
Allows Human Type Mobs to Select a Primary Attack

BothNevarBackwards:
This gives a verb to let you choose an attack, similar to select language. If the attack isn't avaliable, it defaults to the old mechanics of selecting an attack.

### DIFF
--- a/code/modules/mob/living/carbon/human/human_attackhand.dm
+++ b/code/modules/mob/living/carbon/human/human_attackhand.dm
@@ -1,4 +1,12 @@
 /mob/living/carbon/human/proc/get_unarmed_attack(var/mob/living/carbon/human/target, var/hit_zone)
+	// VOREStation Edit - Begin
+	if(src.default_attack && src.default_attack.is_usable(src, target, hit_zone))
+		if(pulling_punches)
+			var/datum/unarmed_attack/soft_type = src.default_attack.get_sparring_variant()
+			if(soft_type)
+				return soft_type
+		return src.default_attack
+	// VOREStation Edit - End
 	for(var/datum/unarmed_attack/u_attack in species.unarmed_attacks)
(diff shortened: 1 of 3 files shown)


SpadesNeil:
Is this purely fluff based, or is it actually changing damage and effects?

BothNevarBackwards:
This doesn't change damage or effects at all. It simply causes the statement that decides what type of attack will be made to check the selected attack first. If and only if that attack is currently able to be made based on the same criteria it was using before, will it then use that attack. Else it'll default back to the old behaviour and choose the attack in the old order. It doesn't affect damage or effects from the attack.

OrbisAnima:
I disagree on that it doesnt change mechanics.

Unless I am misunderstanding anything, the old system made you pick an attack at random or with an order, picking it out of a list of attacks the mob had allowed.

but if anyone looks here
https://github.com/VOREStation/VOREStation/blob/master/code/modules/mob/living/carbon/human/species/species_attack.dm
You can clearly see that some attacks have sharp and edge enabled, some attacks do 5 or even 8 damage, and so on and on... To say that this is not altering would be lying, yet, is not an strong change, but there it is.

BothNevarBackwards:
The attacks are normally done by a specific order. There is no random element to them. Some species having a bite as a stronger attack is a part of the balancing of the species and is taken into account with their other abilities. Right now, the only way to use bite is to cut off your hands as it was originally allowed when cuffed as a way for those species to attack. That is disabled so bites may as well be removed otherwise.

OrbisAnima:
Then you are changing the mechanics, even if they are terrible.

BothNevarBackwards:
Of course the mechanics are changing, that was the whole point. The damage isn't changing and the effects aren't changing, you can just choose which one you want to use.

PR #828, opened 2017-01-14
Prevents accidental stomach flavor text deletion

Cameron653:
The "cancel" button when you were changing stomach inside text caused the entire stomach desc text to be deleted. 

This fixes that problem by checking if it's less than 1 (null applies to this) and just not doing anything to the inside text when cancel is clicked

### DIFF
--- a/code/modules/vore/eating/vorepanel_vr.dm
+++ b/code/modules/vore/eating/vorepanel_vr.dm
@@ -420,6 +420,9 @@
 			usr << "<span class='warning'>Entered belly desc too long. 1024 character limit.</span>"
 			return 0
 
+		if(length(new_desc) < 1) //Prevents current description text from being deleted if they click cancel.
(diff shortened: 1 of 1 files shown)


Arokha:
Don't do length because then you can't set it to be blank. You should actually just check for null instead. A blank text box and clicking okay returns an empty string which is fine, and will set the text to be blank. Clicking cancel returns null. So you can do it and allow people to clear their belly text if they really want, but you can prevent cancel from doing anything.

(Sorry I was buying groceries when this convo came up in admin chat)

Cameron653:
Marking this DNM while I change it.

Cameron653:
Used Aronai's fix (checking that there is actually an input by making it do "if(new_desc)") and it works now with no bugs or errors that I could find from my testing. Ready for merge.

PR #854, opened 2017-01-20
More xenochimera tweaks

Screemonster:
Makes feral creatures not advanced tool users. This primarily affects guns, and is the only thing affected (so far) if you varedit a non-xenochimera creature to be feral.

Injuries no longer increase feral forever, but instead just peg it at an appropriate level for the injury (this should avoid a repeat of 1 broken bone but no other injuries causing feralness to creep up and up and up)

Feralness no longer uses the hallucination var, and has its own proc to handle the results of being cray-cray. Slightly different effects and weighting thereof, is a little more disorienting with the screen-rotations, sees food everywhere. No longer attacked by phantom punpuns.
More importantly, fixes a potential issue where a chimera's hallucinations were stopped near-immediately if someone else is in the same room, regardless of whether the hallucinations were a result of being feral. Now, if they've got an external reason to be tripping balls they'll continue to trip balls regardless of the state of their instincts.

Reordered the feral nagmessages checks so the prob(1) is checked first, _then_ the nutrition/jitteriness, which should hopefully be a tad more efficient.

Had another go at the cold damage thing as I _finally_ found where the "can no longer stand, collapsing!" messages were coming from. Now, it only deals enough halloss to knock you out if you're not _already_ knocked out.

### DIFF
--- a/code/modules/mob/living/carbon/human/human.dm
+++ b/code/modules/mob/living/carbon/human/human.dm
@@ -723,6 +723,11 @@
 	return 1
 
 /mob/living/carbon/human/IsAdvancedToolUser(var/silent)
+	// VOREstation start
+	if(feral)
+		src << "<span class='warning'>Your primitive mind can't grasp the concept of that thing.</span>"
+		return 0
+	// VOREstation end
 	if(species.has_fine_manipulation)
 		return 1
 	if(!silent)
--- a/code/modules/mob/living/carbon/human/human_movement.dm
+++ b/code/modules/mob/living/carbon/human/human_movement.dm
@@ -27,7 +27,7 @@
 
 	//VOREstation start
 	if (feral >= 10) //crazy feral animals give less and less of a shit about pain and hunger as they get crazier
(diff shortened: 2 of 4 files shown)


Screemonster:
The most annoying thing about feralness that I can't think of a graceful way of handling is that all the things that are _supposed_ to make them snap and turn into scuttly sneakblenders also come with pretty hefty movement penalties (pain and hunger in particular).

Part of me wants to add a line to movement_delay() that would reduce these penalties if feral, but I'd rather discuss that with people first ('cause it's basically a buff to fleeing combat)

Screemonster:
Gonna make another commit 'cause repeating the code for the light level check like three times is silly when I can just write a proc for it and call that.

Screemonster:
Reduced the light_level needed for feral 'cause testing showed some areas were considered "dark" even with no visible darkness, now you actually have to be in a shadowy area to be safe.

movement penalties are now divided by feral/10. For comparison's sake, halloss sets feral = halloss, and halloss causes a halloss/10 movement penalty, which effectively means chimera that go feral from halloss are given a move penalty of 1. They still take full slowdown from gear, missing limbs, broken bones, etc etc.

Added logging while I was at it.

Issue #857, opened 2017-01-21
Dogborg gurgles can be stacked


Arokha:
#### Brief description of the issue
You can gurgle people in about 5 seconds if you know how. Click self-clean a bunch of times, then click yes on all the prompts. The digestion happens several times simultaneously because there's no sanity check.

#### What you expected to happen
Not that.

#### What actually happened
That.

#### Steps to reproduce
- Click self-clean
- Click it more times
- Click yes on all the prompts

SpadesNeil:
oh god lol

SpadesNeil:
"It's not a bug, it's a feature."

Zeldazackman:
https://youtu.be/x7YXo5upo18

Arokha:
Misclick whups.

Virgo113:
To be honest instant gurgles are kinda hot...

PR #858, opened 2017-01-22
Dogborg sanity check and inkling hair

Cameron653:
- Adds a dogborg sanity check, #857 
- Adds Molenar's inkling hair (Molenar's specific one is under ears due to hair not being able to be locked to a ckey)
- Adds colorable inkling hair to possible hairstyles.


Virgo113:
to be honest instant gurgles are kinda hot...

PR #867, opened 2017-01-25
Flesh Turfs

Dhaeleena:
Adds fleshy bits!
Flesh walls
Flesh tiles
Flesh Edges and Corners
Mechanical Flesh Material

Thank @Yoshax for all the help! Really thanks a lot


Arokha:
Lewd.

SpadesNeil:
Did the flesh doors get added? I can finally make that vore Z-level.

Dhaeleena:
@SpadesNeil The flesh Doors are the only thing left to add. It shouldn't be too hard. I will start doing that now actually.

PR #876, opened 2017-01-26
Lots of changes and Fluff fixes

Dhaeleena:
Fixed Tasald's fluff items and Pistol.
Added .357 Stun, Flash and Rubber Speed loaders to the science Autolathe.
Created new Strong Variant of Stun, Flash and Rubber ammo to be used code wise with .357 or any other /strong gun.
Fixed Security Vendor logs, Now they Log again.
Un-nerfed the HoS coat.
Security Lockers can now contain a few more items as a few were overflowing with snowflakes.
Crematory is mostly fixed! not fully tho but 95% functional.
Science Autolathe set to 'Unlocked' by roud start.

Thanks @Yoshax senpai for the helping me with the cremator the vendor and the speed Unload of the Revolver.


code/modules/clothing/suits/armor_vr.dm
```diff
@@ -32,3 +32,26 @@
 			else
 				H << "<span class='warning'>You need to have a wolf-taur half to wear this.</span>"
 				return 0
+
+/obj/item/clothing/suit/storage/vest/hoscoat
+	name = "armored coat"
```
SpadesNeil: If all you're un-nerfing is the armor, then none of this needs to be added.

code/modules/clothing/suits/armor_vr.dm
```diff
@@ -32,3 +32,26 @@
...
+	desc = "A greatcoat enhanced with a special alloy for some protection and style."
+	icon_state = "hos"
+	body_parts_covered = UPPER_TORSO|LOWER_TORSO|ARMS|LEGS
+	flags_inv = HIDEJUMPSUIT|HIDETIE|HIDEHOLSTER
+	armor = list(melee = 60, bullet = 60, laser = 60, energy = 40, bomb = 40, bio = 0, rad = 0)
```
SpadesNeil: Only the armor vars are necessary.

code/game/machinery/vending.dm
```diff
@@ -506,7 +506,7 @@
 		vend_ready = 1
 		currently_vending = null
 		nanomanager.update_uis(src)
-
+	return 1
```
SpadesNeil: Needs //Vorestation Edit
or it needs to be in vending_vr.dm

Dhaeleena:
@SpadesNeil I belive that should be it unless i missed anything

code/game/objects/structures/morgue.dm
```diff
@@ -264,9 +264,9 @@
 	else if (src.locked == 0)
 		playsound(src.loc, 'sound/items/Deconstruct.ogg', 50, 1)
 		src.connected = new /obj/structure/c_tray( src.loc )
-		step(src.connected, SOUTH)
+		step(src.connected, dir)
```
SpadesNeil: Why are these changed?

code/game/objects/structures/morgue.dm
```diff
@@ -264,9 +264,9 @@
 	else if (src.locked == 0)
 		playsound(src.loc, 'sound/items/Deconstruct.ogg', 50, 1)
 		src.connected = new /obj/structure/c_tray( src.loc )
-		step(src.connected, SOUTH)
+		step(src.connected, dir)
 		src.connected.layer = OBJ_LAYER
-		var/turf/T = get_step(src, SOUTH)
+		var/turf/T = get_step(src, dir)
```
SpadesNeil: Why are these changed?

Dhaeleena: Ah right i should do the // Edit thing. That fucked up the cremator so fucking badly. Hardcoding it in a way that the Cremator Tray would ALWAYS point south no mater the direction the crematorium was positioned.

Dhaeleena:
Alright i just read through the whole thing, I didn't touch any other normal .dm file, everything else was done in _vr. The lack of //Vorestation Edit was my bad.

code/game/objects/structures/morgue.dm
Yoshax: That hardcodes the crematorium tray to only ever open facing southwards, regardless of the direction of the actual crematorium thing. It really messes things up. The morgue was changed like four years past, but the crematorium never had the same matching change. The whole file is an utter mess, but that puts it in a slightly more working state.

SpadesNeil:
I need to fix those morgue changes for Polaris. That's a bug we fixed. Not a feature that we intentionally changed. If we find bugs like that in future, change them in Polaris's code first instead of in ours and marking them as "Vorestation Edit." For now, I do not care, though.

SpadesNeil:
Dhaaaeeeelllllllll you're doing it agaaaiiiinnn

Dhaeleena:
Oh. Thats why i made a Security copy of that file.

PR #878, opened 2017-01-28
Fix digestion damage variable usage

Arokha:
I was looking for who to blame for this being hardcoded and it's apparently me from a year ago??? Oops.


SpadesNeil:
Are the default values still what they were before?

Arokha:
Yes, there's no practical change for players, other than this lets you VV stomachs to do different damages as an admin, which was always an intended feature for admins to be able to do. Dunno why I typed in 2 and 3 by hand instead of doing that originally.

PR #882, opened 2017-01-30
Makes stasis bags put people to sleep

Arokha:
They don't, right now. It's a little weird. People are just magically cured of their pain and can talk normally, forever, as long as they are in a stasis bag. No hunger/drug effects/bleeding/organ damage progression/toxin damage/oxygen damage, etc. AAAnd can talk normally while in "stasis"? Bags op pls nerf.

### DIFF
--- a/code/modules/mob/living/carbon/human/life.dm
+++ b/code/modules/mob/living/carbon/human/life.dm
@@ -82,7 +82,8 @@
 		if(!client)
 			species.handle_npc(src)
 
-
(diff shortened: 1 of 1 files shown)


OrbisAnima:
alternative: change the name of stasis bags to pods, allow only spider people to spawn them.

Arokha:
Pfff. Th-that's okay.

SpadesNeil:
Should send this to Polaris.

Anewbe:
I may steal this anyway.

PR #888, opened 2017-01-31
fixes scree's suit modkit, tweaks suit, adds mech to code

Screemonster:
Fixes Scree's modkit 'cause it only worked on engineering suits which made it useless for mining.
Also switched it over to use the engineering hazmat suit as a base (offers complete rad protection and slightly better bomb protection, though less melee protection).

Adds a file for vorestation-specific mecha loadouts and adds Scree's one 'cause I'm sick of rooting in menus and varediting every time I screw around in the thunderdome. Scree still doesn't spawn with the thing 'cause that'd be ridiculous.

### DIFF
--- a/code/modules/vore/fluffstuff/custom_clothes_vr.dm
+++ b/code/modules/vore/fluffstuff/custom_clothes_vr.dm
@@ -523,7 +523,7 @@
 //End event costumes
 
 //scree:Scree
-/obj/item/clothing/head/helmet/space/void/engineering/fluff/screehelm
+/obj/item/clothing/head/helmet/space/void/engineering/hazmat/fluff/screehelm
 	name = "Modified Tajara Helmet"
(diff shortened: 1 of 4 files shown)


SpadesNeil:
Oh yea, Scree did win a Phazon in the lottery once. I'm surprised anyone remembers that.

It's canon, tho. Anyone who thinks we can rig the lottery event in the code needs to look at the code and be proven wrong.

SpadesNeil:
If the mech doesn't have a custom paintjob yet, it needs one.

Screemonster:
I _did_ consider changing the phazon itself 'cause for some reason the new() on the base phazon gives it the gravity catapult and the RCD, while every other mech has that as a specific child for spawning in. Probably 'cause phazons are already only adminspawnable.

PR #896, opened 2017-02-02
adds recoloured phazon sprite for the scuttlebug

Screemonster:
As Ace said in #888:

> If the mech doesn't have a custom paintjob yet, it needs one.

Fixed. It's all green and blue.
had to leave the icon-state for phasing as "phazon-phase" rather than "scuttlebug-phase" 'cause of the way the phasing proc is written. Unless that's changed, future custom phazons will need their own .dmi or something.


Arokha:
I assume you've tested this? Driven it around and blown it up?

Screemonster:
Yep, tested and such. Icons seem to work perfectly. Went through walls fine. Wreckage had the right sprite.

PR #898, opened 2017-02-02
Adds more non-fatal symptoms

Arokha:
All virology, all the time.

Spyndrome = You randomly spin in circles and end up facing a random direction.
Lingual Dissociation = Your default language randomly changes.
Mass Revectoring = Random resizing.


SpadesNeil:
We need to send some of these diseases to Polaris for chuckles.

PR #904, opened 2017-02-04
Organizes prometheans and bugfix

Cameron653:
- Moves the promethean spawn_flags over to where all the other promethean code is at
- Fixes a bug where prometheans were mob_small instead of mob_medium


SpadesNeil:
But that's where we keep the code for all the other flags. Also this is a refactor, not a bug fix, unless it unbreaks something.

Cameron653:
Eh, I'll move the spawn flags back to the normal file, so this'll just be a bugfix.

PR #908, opened 2017-02-04
Implemented functionality of adminhelp's being relayed to discord via the chat webhook.

Leshana:
Uses the same config.txt settings as the fax relay.


Leshana:
It seems the slowness in travis is just random.  But this change will prevent it from screwing us over when it happens.

Issue #875, opened 2017-01-26
port "signaler QoL" from /tg/


Screemonster:
#### Brief description of the issue
https://github.com/tgstation/tgstation/pull/23340

Simple QoL change that lets you use signalers on other signalers to copy the frequency and code over without having to click a million times.

Removed most of the rest of the template 'cause it's a feature request rather than a bug report :v

Leshana:
I have opened a PR to Polaris for this.

Leshana:
Merged'ed

PR #912, opened 2017-02-05
Issue #875 - Method to copy setup from one signaler to another

Leshana:
* Just tap them together and settings are copied.
* Code copied from https://github.com/tgstation/tgstation/pull/23340


Arokha:
Either mark with comment or put in another file. Probably the latter, if this isn't being given to Pol.

PR #913, opened 2017-02-05
Fix the annoying punpun runtime thrown on map load.

Leshana:


Arokha:
Did you give this to Polaris? If not, should mark with comment.

PR #914, opened 2017-02-05
Renames fluff vials

Arokha:
People were getting a little scared at "Aronai begins loading vial (poisoned) into the hypospray" when it was antitoxin. x3


JoanRisu:
Testimonial: "It was totally freaky! The vial had the skull and crossbones on it!"

Issue #910, opened 2017-02-04
Add more mechs!


Screemonster:
#### Brief description of the project
Right now, the only combat mechs available to the crew are of the likes of the gygax and durand. Otherwise, there's the ripley and the ody.

We need _moar_.

Specifically, more mechs that are suited to _very specific things_.

- [ ] Some sort of light, fast jetpack-mech. Right now the only jetpack-mechs are adminspawn. Or the hoverpod which is ordered from cargo rather than being a robotics thing.

- [x] Combat-frame mech for micros. Really, it wouldn't need to be any more robust than any other crewmember, with similar material requirements (and robustness) to FBPs. Would probably have a new filepath 'cause they wouldn't be able to use normal mecha equipment.

- [ ] A slooooooow tanky thing that's more suited as a point-defence cannon than an attacking force. Something for sending through the gateway and holding it while the rest of the away team operate.

Wickedtemp:
A small, fragile, low-health average speed mech/exo-suit for the Janitor.

• Powerwashers/ChemSprayers that can either automatically produce or be loaded with cleaner
• An AOE mop
• Wet Floor Sign dispenser 
• Potentially a cleaning foam grenade launcher? Maybe not, if it can be loaded with non-cleaning nades
• Can hold 3 clean bots to launch
• An HUD that makes dirt a little more visible for no missed spots (?)
• Has its own built in water/cleaner tank
------
It's a large map, I think janitors would appreciate these toys to help clean it all.

Screemonster:
A cleaner-grenade-launcher would be astoundingly easy to code. Like, hilariously easy.
Literally just make it a child of the flashbang launcher and set its projectile path to point at cleaner grenades instead.

Arokha:
@screemonster DO THIS

PR #918, opened 2017-02-05
Adds micro-only light mech/vehicle framework, plus equipment.

Screemonster:
The first part of #910 - two light mecha that are usable by micro crew only. The weasel, a super-light, super-fast scout mech with little armor and health and only two equipment slots, and the polecat, a slightly heavier tank-like mech. Neither of them will stand up to the existing mechs in a stand-up fight, but if micros are gonna have guns taken away, this is how they'll be able to participate in gateway expeditions.

Added micro versions of a selection of mech-equippable weapons. They now have refire rates comparable to their man-portable counterparts and are strictly worse than the real deal of mecha weaponry, in terms of RoF, ammo capacity and energy cost.

The shotgun is new, as it turns out the scattershot doesn't actually fire shotgun shells and instead is some sort of pepperbox-rifle or something that pre-dates proper shotguns in the code, so I might be tempted to make an actual full-size mecha shotgun as well. Toggleable fire mode depending on whether you want to load that bad boy up with buckshot, beanbags or slugs.

All values are subject to tweaking once we've had a chance to test and balance them. Weapons in particular may be subject to having their energy costs increased massively. None of this stuff is constructable in-game _yet_ - it's just there so folks can have a chance to test and play with it because I suck at balance.

The sounds are kinda placeholders I found in the sfx folder, could do with something better.


Screemonster:
Just realised that I didn't increase the energy cost of the shotgun compared to the scattershot (which is what it's based off). Scattershot fires 4 pistol bullets at a cost of 25 energy each, shotgun fires _one_ shotgun shell which is then multiplied into pellets after firing 'cause that's what buckshot does. Yoshax did some testing with the numbers. First column is range (7 tiles down to 2) and others are damage dealt on two attempts each.

Scattershot Normal:
7: 25, 25
6: 75, 100
5: 50, 84.1
4: 75, 75
3: 100, 85.5
2: 75, 75

Scattershot Micro: 
7: 0, 12.6
6: 0, 11.6
5: 0, 34.6
4: 38.9, 30
3: 23.2, 18.4
2: 64.4, 38.4

Gonna boost the energy per shot to 100. Raw damage in buckshot mode is less overall but the micromech shotgun has the advantage of being able to switch out for slugs for better long-range murderin'.

Screemonster:
image
Preview of sprites.
On the top row, from left to right: drill, ore hopper/scoop, laser cannon, laser carbine, taser, flashbang launcher, shotgun.
Bottom row: Polecat, Weasel, Gopher (and Scree)

Screemonster:
DNM removed 'cause finished tweaking.

Arokha:
Are these producible somehow?

Screemonster:
Not yet.  Added as adminspawn only for now 'cause they might need further tweaking for balance. Also I have nooooooooo idea what passes for a sensible material/tech requirement. That's really what the "requires discussion" bit is for.
I was aiming for them to be _somewhat_ easier to get hold of and produce than an actual combat mech 'cause otherwise, yanno, there's no point in having them and just make a gygax.

code/game/mecha/micro/micro.dm
```diff
@@ -0,0 +1,297 @@
...
+							var target_id = "target";
+							var form_id = "form";
+							var input_id = "input";
+							var delay=5;
```
Arokha: Can you remove all this for me? I know it was copypasted from their file but if you're not using the huge tracts of ~~land~~ ~~shrktits~~ comments then can you just remove them from yours

Screemonster: Done. Fixed some other comment copypasta while I was at it.

Screemonster:
image
preview of the wreckage sprites too 'cause why not

Arokha:
Merging since it won't affect the game (no in-game way to acquire them currently), and balance issues can only really be tested with other players so needs to be on the server to test that.

PR #919, opened 2017-02-05
[DNM]{Help Needed] Body Markings Electric Boogaloo

Poojawa:
Wew lads, we're at it again. 

Only this time I half understand what I'm doing better.

Ported from Paradise code, and made to be compliant with y'all's code standards.

Help is very much requested on getting the Preference setup stuff to work. Because I have no idea how any of that is var'd and everything is errored out as an undefined var. Mainly stuff like the markings_green, markings_red, etc

Feel free to make a full review of any suggestions/tweaks. Maintainers can also edit/add at will.

Same bug that stopped me last time, tbh.


Poojawa:
It's okay Travis. I know it doesn't compile.

Arokha:
I looked this over and have some comments on how it could be made to work with our stuff.

Firstly, anything that CAN be moved to the sprite_accessory system (ears, tails) should be. There are a lot of sprites in body_accessory_vr.dmi like horns and ears and tails that can be put there.

Second, the full body markings (tiger, unathi, etc) should be split up to separate sprites, because if someone is missing a leg, it's still gonna draw the missing part color overlay over where their leg would have been. I dunno if Paradise has some sort of mystical code to avoid that, or if they just deal with it, but I'm not aware of any workaround here.

Then, if it's split up, you can add a list to humans called... I don't know. "body_patterns" and add bodypart constants to it so you end up with a list like list(BP_L_LEG = "tiger_stripes", BP_L_LEG = "tiger_stripes") and so on.

Then, in organ_icon.dm, which is where the code asks each external body part to provide an icon of itself, you add code to check that list, and the organ slaps the (colored) overlay on itself and hands it off to update_icons.dm.

The code to find the right icon_state to overlay would be like... it'd look for itself in the `owner`'s `body_patterns` list (it's `organ_tag` var matches BP_L_LEG, or whatever) and grab the value which would be "tiger_stripes", then it'd probably just look for the icon_state for like "tiger_stripes_[organ_tag]", which would be like "tiger_stripes_l_leg" since `organ_tag` would be BP_L_LEG and if you look in defines/mob.dm that is `l_leg`. ... That's a little dense but, it'd work something like that.

This makes it not require any code in the already crazy complicated update_icons.dm, and only very minimal code in organ_icon.dm.

Arokha:
In terms of splitting up the sprites, I assume there's some sort of efficient way of doing it where you somehow copy/paste feet/hands/legs/torsos/heads from existing mobs and then use those as a template to 'cut' out all the other sprites.

Screemonster:
Take a look at how the wound overlay works, 'cause that's just one sprite that gets drawn as an overlay and doesn't need to worry about whether limbs are present or not.

Poojawa:
I still need to be able to actually work at testing the overlays themselves, and I have no means to do that, since, I don't know how to build a UI that actually functions in this codebase.

Arokha:
Look at `code\modules\client\preference_setup\vore\02_size.dm` for examples on how to do the UI.

- `load_character(var/savefile/S)` is called when this particular thing is being loaded.
- `save_character(var/savefile/S)` is called when this preference is being saved.
- `sanitize_character()` is called to make sure all the values are valid.
- `copy_to_mob(var/mob/living/carbon/human/character)` is called when the mob spawns, to put whatever this is on them.
- `content(var/mob/user)` is the actual HTML content that is added to prefs.
- `OnTopic(var/href, var/list/href_list, var/mob/user)` is basically like Topic but just for these items.

Arokha:
Busy pillaging this.

Arokha:
@KasparoVy https://github.com/PolarisSS13/Polaris/pull/3152

Poojawa:
wew. Thanks again for taking it on. it was going to be a mountain that I just don't have the code talent to beat.

PR #921, opened 2017-02-06
Literally ruins the game

Arokha:
Adds cloneloss to DNA injectors for the budding rogue geneticists.

Makes sense, doesn't it? Messing with your genes? Each one doubles your cloneloss, starting at 2.


SpadesNeil:
good

Issue #205, opened 2016-05-23
Add more vore sounds


Screemonster:
#### Brief description of the issue

All the existing vore sounds are gulps and slimy inserts, while many people also use the bellyspaces for things like maws and other stuffing that is not necessarily swallowing. (I mostly use mine for micros and monkeys)
#### What you expected to happen

https://www.youtube.com/watch?v=B3vkzRdp9vU
or https://www.youtube.com/watch?v=aiNg8ChsyUg
Or something like that.
#### What actually happened

A plain ol' gulp even though not ICly actually swallowing.
#### Steps to reproduce

Turn speakers on.
_Testy McFeeder is attempting to cram Helpless McMicro into Scree's gaping maw!_
#### Additional info:
- **Server Revision**: Found using the "Show Server Revision" verb under the OOC tab.
- **Anything else you may wish to add** (Location if it's a mapping issue, etc)

Screemonster:
http://forum.vore-station.net/viewtopic.php?f=20&t=686 posted 'cause of this thread.

SpadesNeil:
I actually quite dislike cartoony noises. Got anything else?

Vae42:
I am a sound technician actually. I could make some more sounds for us (if scheduling allows) and I agree a little more diversity in sounds would be nice. I personally use bellies for damn near everything including "bra" so any kind of squish sound makes no sense...
 I'm gonna work with this a bit tomorrow, actually feeling kind of inspired right now. Is there any requirement on file size, clip length, file type (Mp3, wave are easiest for me of course.)?

also some suggestions on what kind of sounds would be beneficial would be nice.

SpadesNeil:
Gurgle sounds should be no longer than 15 seconds and no shorter than 7. Some overlapping of sounds is expected and actually desired. Though I think the current gurgle noises are fine; we just need more that sound alike.

Swallowing sounds should be no longer than 3-5 seconds. I forget how long it takes to swallow someone. The GULP or SLOSH noise of you finally sending them into your stomach should not exceed like, 1 or 2 seconds.

Currently the gulp noise is played when they hit your stomach. There's no sound for "So and so begins shoving X into their stomach!" so there's no vore-attack sound.

There's also a sound for the death of prey. This should be extra gurgly and more intense than the normal digestion noises--as it's the sound of your prey breaking up into soup and entering intestines. :P

[...]

PR #931, opened 2017-02-08
Sounds add for #205

Arokha:
A few more choices for those that prefer these. Basically the backpack-cloth-rustle noise, and a choice for no sound. Useful for those that have "taur back" or whatever bellies.


SpadesNeil:
We really oughta add a feature to actually properly ride taurs.

Arokha:
lol. probably.

Screemonster:
You can ride them on the inside :3

PR #938, opened 2017-02-09
Expands Science Workshop

Yoshax:
People complained it was too small.
Made this. People liked it:
Workshop

Workshop and Server Room swap places so the Workshop can be expanded. Has more space along with a locker, plant, disposals, charger and cell charger.


Arokha:
You made sure the cooling and stuff works right in the server room right, cuz I'd hate for it to blow up like 20 minutes into the shift

Yoshax:
I just copy-pasted it down. It's the exact same room, nothing changed.

SpadesNeil:
_Should_ be okay.

Arokha:
We will see if you have DOOMED US ALL.

PR #940, opened 2017-02-11
CC Map Tweaks & 2 New Shuttles

Arokha:
Okay, the administration shuttle isn't new, I just made it more comprehensive.

-Made it easier to create multi-destination shuttles for player-use when not antags. The existing ones were pretty geared towards the Mercenary game mode. Now it's pretty easy to make a shuttle that goes a bunch of places.

-Redesigned admin shuttle (pac-man shaped one at CC) to be multi-destination shuttle more capable of supporting RP. Can be used on 'hard' away missions by putting in the appropriate area in the away mission (`/area/shuttle/administration/away_mission`). Or, hell, include it on every new away mission and say that it's the E-Z mode version of that mission.

-Added new AwayMission shuttle to Dock 1. By default it can only go between the station and the old abandoned construction site, but if it detects that there's a map loaded with the appropriate area (`/area/shuttle/awaymission/away`), then it will be available for travel to that destination as well, listed as "Unknown Location". The shape of the area should be 13 wide x 5 tall. The shuttle opens to the right, in the center of the 5.

-Redid several parts of Centcom. Nothing too huge. Mostly made some hallways less barren, allowed access to the trams area on the right by default for shift-end people to use the trams, tweaked a couple of hallways to make sense (going through the bridge to get to a security checkpoint?) and added a couple more potted plants for good measure. Mostly added windows because several hallways were just... like... I feel like I'm underground.

-Changes on z1: Added away shuttle to dock.
-Changes on z2: Lots, see above.
-Changes on z4: Add area for Away Mission shuttle.

Spess Ship


SpadesNeil:
Contrary to RP on the gateway missions, which is considered in an "alternate universe" because Bluespace, the events that take place in the shuttle away missions will be considered *permanent* and *canon* to the location it affects.

For example, if you build a house on a hostile alien world, the house will exist for as long as the local environment allows it to exist.

Screemonster:
Shuttle is good for awaymissions that wouldn't make sense for us to infiltrate via the gateway though, such as event missions like the hitlerbase. 'cause why would they have a gateway?

Arokha:
I want to make an "Anti-Mercenary" shuttle mission where you go attack the Mercenary base BEFORE THEY ATTACK YOU. *PRE-EMPTIVE STRIKE, HIYAAAAH!*

Arokha:
Coincidentally the shuttle console requires the same access as the gateway doors (62). Or you can emag it.

PR #952, opened 2017-02-14
Driving into town!

JoanRisu:
Adds the new Rover/Humvee into the game with basic driving features for playtesting!

Also disables the security segway until it is refined and adjusted.


Arokha:
Maximum dork

PR #953, opened 2017-02-14
Adds Krak-B-Gone, a silicate spray for fixing windows

Screemonster:
Adds spray bottles of silicate to youtools, tool closets, and engineer/CE lockers.

Maybe damaged (but not broken) windows will actually get fixed now. Still worth going to chemistry for more if you're planning on reinforcing a huge amount of windows.


SpadesNeil:
Should give to Polaris.

PR #955, opened 2017-02-15
Fire now damages doors.

Screemonster:
Makes airlocks, firelocks and shutters take damage from fire.
The thresholds before they start to take damage have been set the same as their corresponding walls - so normal airlocks will start to take damage at the 1800 degrees (the same as ordinary steel walls) and heat-resistant airlocks, blast doors and firelocks will do so at 6000 (same as a plasteel rwall).
The damage, and threshold before they melt completely after breaking, have been set to put blast doors on a par with reinforced plasteel walls. Ordinary airlocks with the heat_proof property will break slightly beforehand on account of lower HP.

Airlocks that are not heat_proof can now be fireproofed by adding two plasteel sheets and welding them into place, the same procedure as repairing a damaged airlock with steel. The sheets may be retrieved prior to welding by crowbarring them.

All the airlock types in use around the toxins burn chamber and the engine are heat-resistant, so no mapping changes need to be made. If you were running the chamber hot enough to break these doors, you were running it hot enough to melt the walls anyway.

Prevents cheeky fuckers from building inpenetrable burn chambers out of airlocks.

### DIFF
--- a/code/game/machinery/doors/airlock.dm
+++ b/code/game/machinery/doors/airlock.dm
@@ -784,7 +784,8 @@ About the new airlock wires panel:
 	else if(istype(C, /obj/item/weapon/pai_cable))	// -- TLE
 		var/obj/item/weapon/pai_cable/cable = C
 		cable.plugin(src, user)
-	else if(!repairing && istype(C, /obj/item/weapon/crowbar))
+	//else if(!repairing && istype(C, /obj/item/weapon/crowbar)) vorestation removal
+	else if(!repairing && !reinforcing && istype(C, /obj/item/weapon/crowbar)) // vorestation replacement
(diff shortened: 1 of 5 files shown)


Arokha:
Whups. Didn't see your discussion label. Uuuh.

Arokha:
Weeeell we can just comment it out in the code if it becomes an issue. It's mostly that one fire_act proc.

Screemonster:
If you comment out the adjacent_fire_act() thing in the floor_acts, then fire_act() will never get called. Closed doors actually delete fires that are under them to stop them getting trapped by ZAS, so fire_act() won't get called on them directly.

Arokha:
This was reverted so it can be discussed

Leshana:
Pretend this PR didn't exist.  MiniTruth

PR #959, opened 2017-02-15
Fire now damages doors.

Screemonster:
Redid PR #955  'cause the last one got borked. Also made it helluva neater thanks to the attempt_vr proc.

Makes airlocks, firelocks and shutters take damage from fire.
The thresholds before they start to take damage have been set the same as their corresponding walls - so normal airlocks will start to take damage at the 1800 degrees (the same as ordinary steel walls) and heat-resistant airlocks, blast doors and firelocks will do so at 6000 (same as a plasteel rwall).
The damage, and threshold before they melt completely after breaking, have been set to put blast doors on a par with reinforced plasteel walls. Ordinary airlocks with the heat_proof property will break slightly beforehand on account of lower HP.

Airlocks that are not heat_proof can now be fireproofed by adding two plasteel sheets and welding them into place, the same procedure as repairing a damaged airlock with steel. The sheets may be retrieved prior to welding by crowbarring them.

All the airlock types in use around the toxins burn chamber and the engine are heat-resistant, so no mapping changes need to be made. If you were running the chamber hot enough to break these doors, you were running it hot enough to melt the walls anyway.

Prevents cheeky fuckers from building impenetrable burn chambers out of airlocks.

### DIFF
--- a/code/game/machinery/doors/airlock.dm
+++ b/code/game/machinery/doors/airlock.dm
@@ -746,6 +746,7 @@ About the new airlock wires panel:
 		return
 
 	src.add_fingerprint(user)
+	if (attempt_vr(src,"attackby_vr",list(C, user))) return
 	if(istype(C, /mob/living))
 		..()
 		return
--- a/code/game/machinery/doors/door.dm
+++ b/code/game/machinery/doors/door.dm
@@ -202,6 +202,8 @@
 /obj/machinery/door/attackby(obj/item/I as obj, mob/user as mob)
 	src.add_fingerprint(user)
 
+	if (attempt_vr(src,"attackby_vr",list(I, user))) return
+
 	if(istype(I, /obj/item/stack/material) && I.get_material_name() == src.get_material_name())
 		if(stat & BROKEN)
 			user << "<span class='notice'>It looks like \the [src] is pretty busted. It's going to need more than just patching up now.</span>"
--- a/code/game/machinery/doors/door_vr.dm
+++ b/code/game/machinery/doors/door_vr.dm
@@ -5,9 +5,6 @@
 /obj/machinery/door
 	var/obj/item/stack/material/plasteel/reinforcing //vorestation addition
 
-/obj/machinery/door/blast/regular
-	heat_proof = 1
-
 /obj/machinery/door/firedoor
 	heat_proof = 1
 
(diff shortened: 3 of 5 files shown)


Screemonster:
image

Aftermath of setting a big fuckoff fire in the engine chamber. Highlighted the values of the back wall leading to space - the blast door had 2.03 destroy_hits left (which translates to 40.6 health before being ashed), the rwall had 759.9 damage (breaking when it reaches 800, so... 40 health left)

Short version: fire reduces an airlock's health until it breaks, then starts to deplete its destroy_hits at 1/20th the rate. Doors have 10 destroy_hits, giving them an effective pool of 200hp between being broken and being ashed.
For comparison, a door has 300 health and a blast door has 600, to it takes 500 fire damage to completely destroy a door/firelock, 800 to destroy a blast door, and 800 to destroy an rwall. Non-heat_proof doors, despite having the same HP, will break faster as fire damage scales with the difference between the air temperature and the door's max temperature.

SpadesNeil:
I have concerns about this mainly because of how it affects the Super Matter Engine and Phoron Research.

Screemonster:
All doors that are in direct contact with either the engine chamber or toxins test chamber are of types /obj/machinery/door/airlock/hatch, /obj/machinery/door/airlock/glass_research, or /obj/machinery/door/blast/regular, all three of which have the heat_proof property - fire_act doesn't start to do damage to anything with that property until it hits 6000K.

For comparison, from materials.dm:
default materials are `var/melting_point = 1800`
plasteel is `melting_point = 6000`

The _only_ circumstance where this will start to melt the doors where you wouldn't be melting the walls too is if you somehow constructed a burn chamber out of durasteel, with `melting_point = 7000`, the highest of any material, and kept the temperature consistently between 6000 and 7000.

Short version: if you melt the doors, you were a gnat's hair away from melting the walls anyway.

Screemonster:
Did some more testing. Opened some of those big pressure tanks to make an ideal burn mix in the toxins chamber and hit the emergency vent when I saw "Mixing Room Exterior Airlock shows signs of damage!"
When the fire went out, all the walls of the chamber were scorched, and the first airlock (out of two) had 123.4 health remaining.

Screemonster:
After some discussion in the dev channel on discord, there was a lot of support for making blast doors outright immune as they're non-constructable adminspawn etc.
Gave other doors a buff to bring them up to their level.

image
Tested by lighting a fire in the engine again, poised to screencap when something melted. The airlocks melted first, but the rwall at the back melted as well _in the time it took for me to hit print screen_, so I think "difference is literally faster than human reaction time" is close enough.

image
The other wall melted almost immediately.

image
Once the room vented, most of the room/chamber was still intact. Nothing that couldn't be rebuilt with a few sheets of plasteel, a welder, and some airlock electronics, but anyone in the room at the time of the containment failure would be pretty seriously fucked. Achieves the goal of no longer being able to cheese burn chambers with an RCD.

Yes, this will make phoron-cooled engines FUN if you're dumb enough to let them melt the doors 'cause suddenly you'll have a room full of 6000-degree phoron which will pretty much immediately ignite the oxygen in the room, but in all honesty you're likely to blow the core doing that anyway.

SpadesNeil:
Approved for **FUN**™

PR #970, opened 2017-02-17
Remove minergame decay

Arokha:
Can readd it later when people start actually doing it more or going in teams. For now, wanna make it easier to build up score.


code/modules/rogueminer_vr/controller.dm
```diff
@@ -67,7 +67,6 @@ var/datum/controller/rogue/rm_controller = new()
 	//How many zones are we working with here
 	for(var/area/asteroid/rogue/A in world)
 		all_zones += new /datum/rogue/zonemaster(A)
-	decay()
```
SpadesNeil: Just comment it out rather than delete it. :B

Arokha:
Oh we both did it.

PR #984, opened 2017-02-20
Resleeving Stuff

Arokha:
After much forum and Discord talk, I think this is ready for 'part 1' of this project.

- Moves the psych office to where the cloning room was, and consumes the old one and one patient room to make the resleeving lab.
- Adds 'mind backup' implants that are essentially more convenient scanning. Synths can also get these, who currently have no good way to get revived if destroyed.
- Adds resleeving equipment. Allows you to print bodies of people who have joined the round (NT stores their records offsite), and anyone who has backups can be put into a body. This prints bodies 'better' than the clonepod did: it remembers organs you had/didn't have when you joined. If you had a robotic arm, you get printed sans an arm, and need to go get one from robotics.
- A book explaining how all this works, including how it's illegal to do the illegal stuff (printing other people's bodies for fun, overwriting minds, restoring backups of living people (that has an OOC consent prompt anyway for the resleeve-ee so it can't be done without their consent anyway)).
- Synths get an equivalent of cloning, a SynthFab. It works much the same, and produces pre-damaged synths to sort of replicate the cloneloss and disabilities that biological clones pop out of the pod with. Robotics starts with a handful of backup implants to implant into synths, so they can have a cloning equivalent.
- Whitelisted species are unprintable (they are all currently unclonable so this isn't much of a change).

To make it more prefs-compatible, there are two new character setup options on the Vore tab:
- You can disable NT having a bodyscan of you, effectively making you DNC. So, here's your mechanical way of becoming DNC if you so wish that for your character. This could be because of story reasons, because you're not an employee, you're there illegally, etc.
- You can disable people sleeving into your body if you find the idea distressing/distasteful. If enabled, it will only allow your ckey to inhabit your printed body copies.

I'll update the wiki when merged and stuff to make it current so people aren't left in the dark, though the ingame book I made is very complete.

Part 2 will contain things such as:
- 'Unsleeving' someone into the resleeving console, effectively making them a stationary powerless AI, mostly so that the sleeving tech can...
- Customize bodies. Allows you to tweak or print custom bodies and put people into them. It's like genetics without the horror, superpowers, and radiation.
- Per-species body requirements (promethians might need a slime core or slime jelly, etc).

The purpose of Part 2 is to allow people to undergo drastic body modification things for RP or whatever reason they so wish.

Unrelated change:
OOC notes now stay with bodies. Even if the client is logged out, you can still review their OOC notes. Useful for checking SSD prefs. Also OOC notes will be present on bodies when printed, so no prefbreaking pls. And said notes will be overwritten by the mind's notes if one is put in.

ONLY posibrains are indigestible now. MMIs are digestable. All that squishy brain!


SpadesNeil:
what has science done

Screemonster:
Maybe we'll have an IC reason for people to be coming onto the station with a different taurbody to the one they had last week now besides the handwavium "I went to that body-mod clinic on virgo prime"

Arokha:
The ingame book if anyone wanted to see: https://arokha.com/resleevdocu.html

Dhaeleena:
All we need now is a bigger Freezer room in medbay equipped with chairs to stored the uh spares

PR #995, opened 2017-02-21
Make Telescience Great Again

Leshana:
## Re-adds *Telescience* to the game! ##

We are using the /tg style telescience which uses Bearing, Elevation, and Power to target teleports, the same style that used to be present on old VOREStation. New in this revision are several changes:

- Telescience console uses nanoui
- All telescience machines and tools are research-able and construct-able.
- **Certain map areas are completely immune from teleports**
  - The armory
  - The vault
  - All dorms
- Telepad and telescience console are upgradable with better parts.

Quick summary of telescience for those who have not used this version:

- The telscience console controls a telepad, which can send & receive non-anchored objects on itself to some destination.
- Destination is determined by entering bearing, angle and power.   This is translated to a target coordinate using trig functions, (A ballistic trajectory thru bluespace)
- The trick is, you only get so many teleports before you have to recalibrate, and there is a hidden amount of error in the targeting that resets each time you teleport.

its working 1


Arokha:
Are you going to do the stuff we talked about with regards to making it 2D math instead of 3D math, or add the random prob stuff?

Issue #900, opened 2017-02-03
Fluff Item - Chakat Tempest - Wickedtemp (Bags)


Wickedtemp:
Name of Item: Tempest's Saddlebags

Item description: (try to keep it short and sweet): A custom-made set of saddlebags, tailored to Chakat Tempest's exact dimensions, and taste in color! One one side, there's small print stitched in: "...to carry the weight of any responsibility, burden or task."

Sprites (or link to sprites): https://drive.google.com/open?id=0B8qdaualwKiQMmg5cWVlZkR4VWM

Any additional resources: (in a .zip or .7z file, including sounds and DM files with code in them) NA

BYOND Key (Username): Wickedtemp

Name of Character: Chakat Tempest

Main Job(s) of Character: CMO, Chemist, MD, Medical-Only character. Sometimes Resident, when I feel like scening.

Why your character should have this item: Saddlebags would of course be more comfortable for Tempest, being a taur. This also gives hir yet another opportunity to add more purple to the ever-growing quantity that shi already has. Depending on the size of the bags, they'd also be rather useful to have in Medical as well, so shi wouldn't have to loot the Paramedic duffle's.

Additional Notes (If it's a permit, insert the information for the permit here: I'd like for this to have dufflebag capacity without the slowdown. Fluff reason being, as a macro-taur, Tempest isn't likely to be slowed down by the weight of the standard medical gear that shi'd carry, as well as the saddlebag being custom-made for Tempest's body shape, optimized for comfort and accessibility. I believe you could lock it to Tempest's custom taur sprite to prevent anyone else from using it, as a way of having it mechanically tailored just for Tempest's body... If this is seen as too unbalanced, I'm perfectly happy with this being duffle capacity with slowdown, but I'd prefer the former.

Fluff Item slots already used: This would be the 4th slot, if the Hypo is also approved and added.

Screemonster:
It does sorta make sense for a macro taur to not be slowed down by a dufflebag. If I could think of a way to tie slowdown from an item to playerscale it'd be perfect.

Arokha:
You'd have to have it check every few seconds or something unless you want to code a lot more, so don't do that since it's horrible. Just limit this to people with her taur body style (though also be careful about that since she has fluff butt sprite).

Wickedtemp:
[...]

PR #1000, opened 2017-02-22
Tempest saddlebags , saddle bags adjustments

Cameron653:
#900 
- Makes saddlebags slightly slow down non-macros, and not slow down macros ~~via magic~~
- Makes saddlebags hold as much as duffle bags 
- Adds in Tempest's Saddlebags 
- Modifies saddlebags code so you can't place it on your head and weird stuff like that[.](http://pastebin.com/fjsHAZ1R)


Arokha:
Seems fine. Merge if you want, when check is done.

PR #1003, opened 2017-02-22
adds more particle guns and icon fixes

Screemonster:
Makes some tweaks to the particle gun, mostly giving it long examinetext and fixing the item states, as well as giving it a more pistol-like sprite considering it's equivalent to an egun.

Adds an AEG-equivalent using the original, more rifle-like sprite. Shorter delay before it starts to charge, but it charges more slowly. Works out the same for a full charge overall.

Adds a laser-cannon equivalent. Also recharges. Slowly. Mostly because belters really don't need to be running back and forth to the charger. Bigger than a laser cannon though, and can only be carried in-hand or on a back slot. Recommend high material requirements.
Also you _really_ don't want to fire it inside.

Adds a pressure interlock item that'll disable the gun if it's got a chance of exploding. You know, if you're a _pussy._

Adds admin logs/notifications for when some dumbass ignores the examinetext and fires it indoors.

Particle projectiles now bust up mineral turfs. You're still better off with a drill, but I guess you can mine with them.


Arokha:
Doesn't seem like travis is STUCK but it seems like the call to START it failed since I can't even see details about it. Gonna close/open this to make it reconsider life.

PR #1026, opened 2017-02-23
Unwhitelist Gutter?

Arokha:
This is a 'proposition' PR.

In my opinion, whitelisting languages just kills them. Sure one person has applied for Gutter, but who the hell are they going to speak to with it? Why would anyone apply unless they're friends with that person? It basically kills the motivation of many to apply, or even use the language.

Instead, if people have taken Gutter and they don't have a criminal background, we should talk to them because they are not respecting the lore. We should probably codify that you're required to respect the lore in the rules more clearly, and that an SCA is required for other things.

If you're in Security and you take Gutter to make it "easier to catch criminals ingame" then you are playing on the wrong server. Sorry. That's just how it is. You should take languages because of story reasons. To make the game more interesting for yourself and everyone around you. Taking a language to "catch the criminals in the video game" means you probably need to be talked to by admins about what the server is about.

I'll conclude with an analogy of whitelisting languages:
You can leave the cookie jar in your child-filled house on the counter, or up out of reach. If you put the cookie jar out of reach rather than teaching them to behave, it just teaches them that you're a 'mean adult' and makes everyone else who has self-control have to reach higher to get the cookies. So it doesn't teach the people who need to be taught anything, and makes people who already know have to jump through hoops.

### DIFF
--- a/code/modules/mob/language/station_vr.dm
+++ b/code/modules/mob/language/station_vr.dm
@@ -60,8 +60,6 @@
 	flags = 0
 /datum/language/seromi
 	flags = 0
-/datum/language/gutter
-	flags = WHITELISTED
 /datum/language/machine
 	flags = NO_STUTTER | WHITELISTED
 


AaronSherwood117:
I agree, a language whitelist is basically a death sentence for it.  Unlike a species or SCA, there's not an immediate benefit when it's approved for someone.  The fact that languages need multiple participating parties drops the value of going through whitelist effort to near zero.

SpadesNeil:
I am against removing the whitelist for the following reasons:

- Security abusing it, which has already been covered.
- Players using it just because it's edgy, even though their backstory has no criminal background or reasonable explanation to why they know the language.

I understand the argument with the cookie jar, but disagree because it's not like we're looking after one or two people who want to abuse it. We're looking at too many to keep track of. I tried moderating this manually once before and gave up because it was a nightmare to enforce.

I'd like to see people apply for and properly use the gutter language so they can serve as an example of how a _good_ criminal character is played, as opposed to a random shitter just causing trouble for his own laughs. I'd be happy to teach some people in exchange for Gutter whitelist.

Arokha:
> I tried moderating this manually once before and gave up because it was a nightmare to enforce.

I think it should be up to the people who care about the lore, aka the people who really use Gutter properly, to report it. I think instead we should just teach people who are interested in the lore to spot and report people who are abusing it to admins. Too many things go unreported already, and teaching people what to report might help.

SpadesNeil:
How are they going to know to report it when all someone does is eavesdrop?

Arokha:
Wouldn't it be pretty obvious they got busted using evidence from their conversation?

SpadesNeil:
That won't be relevant when Security just _happens_ to barge in on you and your cache of stolen goods you were talking about in Gutter that this random security officer happened to notice by walking down the hall.

It's happened to me irritatingly more than once but the player will argue that other evidence was in play, or that they just got lucky, and I can't punish people on only a hunch that they were powergaming.

Wickedtemp:
The thing about a "criminal language" is that anybody who's even capable of recognizing the language would be able to assume within reason that the person speaking it either has done something illegal, or has a history of illegal activity. Honestly, I can easily see an Officer keeping an eye on someone because they speak Gutter, since, well, it's known as a 'criminal language'.

And as for people reporting it? I'll happily report it when I see it if I'm told what I'm supposed to be looking out for.

SpadesNeil:
_Recognizing_ the language is fine. _Understanding_ it is not.

Dhaeleena:
I am with ace on this one

SpadesNeil:
I'm gonna close this for now until we find a chance or reason to discuss more.

PR #1044, opened 2017-02-28
Make ALL the basic stock parts manufacturable in an autolathe, not just a few of them.

Leshana:
Yup @Screemonster

### DIFF
--- a/code/datums/autolathe/engineering_vr.dm
+++ b/code/datums/autolathe/engineering_vr.dm
@@ -1,0 +1,22 @@
+// Lets allow autolathes to build the basic tech 1 parts shall we?
+
+/datum/category_item/autolathe/engineering/basic_capacitor
+	name = "basic capacitor"
+	path = /obj/item/weapon/stock_parts/capacitor
+
+/datum/category_item/autolathe/engineering/micro_mani
+	name = "micro manipulator"
+	path = /obj/item/weapon/stock_parts/manipulator
+	
+/datum/category_item/autolathe/engineering/basic_matter_bin
+	name = "basic matter bin"
+	path = /obj/item/weapon/stock_parts/matter_bin
+
+/datum/category_item/autolathe/engineering/basic_micro_laser
+	name = "basic micro laser"
+	path = /obj/item/weapon/stock_parts/micro_laser
+	
+/datum/category_item/autolathe/engineering/basic_sensor
+	name = "basic sensor"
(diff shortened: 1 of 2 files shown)


code/datums/autolathe/engineering_vr.dm
```diff
@@ -0,0 +1,22 @@
+// Lets allow autolathes to build the basic tech 1 parts shall we?
```
SpadesNeil: I vote no, because this lets you make infinite autolathes. This was a design choice not to let the autolathe make these.

Screemonster: Science can already make infinite autolathes/protolathes, but I can see how giving cargo the means to churn them out might be an issue if they just stick one in the hall.

SpadesNeil: Yeaaa.

Sorry, Leshana! But I'mma say no to this one.

Screemonster: ... okay, after it's closed...

SPUDS YOU DUMMEH YOU FORGOT THE AUTOLATHE CIRCUIT BOARD IS STILL NEEDED :V

Dawgas:
But science can already make an arbitrary amount of autolathes AND this doesn't actually allow anybody to make any boards. If "we're" going to cite "design choices" as the reason for not doing this change, when so much of what's already there was made and changed by people who don't understand the systems as they were or the magnitude of the changes they're looking for.

So here's a better thought to think, as far as a real discussion over this change: "What would happen if cargo, engineering, or a really resourceful Assistant had a steady supply of stock parts that they could manufacture on demand?" Mind you, R&D consoles are ID locked so you can't just research on your own.

code/datums/autolathe/engineering_vr.dm
SpadesNeil: Still noooooo.

SpadesNeil: Weeeeeeeeeeeehhhhhhhhh.

SpadesNeil:
# Weeeeeeehhhhhhhhhhhhh

Arokha:
I think Leshana is coding a special parts fabricator that you can jam circuitboards into and it prints you all the parts required to build the thing.

Leshana:
I think #1052 is probably even better than this, so lets do that.

PR #1050, opened 2017-03-01
Makes resleever upgradable

Arokha:
Scanning modules and manipulators!

Scanning modules reduce the confuse time, manipulators reduce the eye-blurry time.

Occupant gets a different message when resleeved and there's very low for both, to let them know.


Wickedtemp:
B-but muh ethylredoxrazine! What ever will I use it for now? D:

Arokha:
For the times where R&D doesn't upgrade it! And drunk people!

PR #1054, opened 2017-03-01
Fix custom items appearing in random space turfs

Leshana:
* If a custom item can't be equipped to a character's inventory it is dropped on the ground.  For late join characters, equip_custom_item() was called before the mob was placed at their starting locations; their coordinates were still 1,1,1.   Which is the edge of a map so they get bounced to another z level.
* Fix: Equip custom items AFTER job_master.LateSpawn places the mob at the starting location, so it flows in same order as roundstart join characters.


Screemonster:
Probably worth kicking this up to Polaris if you haven't already.

Leshana:
They have already merged it actually.

Arokha:
We can just inherit since they merged it superfast.

Issue #887, opened 2017-01-31
Fluff Item Application - Dhaeleena M'iar - Dhaeleena


Dhaeleena:
Name of Item: Engraved Mateba

Item description:
This unique looking handgun is engraved with roses along the barrel and the cylinder as well as the initials DM under the grip. Along the middle of the barrel an engraving shows the words 'Mateba Unica 6'. Uses .357 rounds.

Any additional resources:
The item already exists in the code if approved i will do a PR myself.

BYOND Key: Dhaeleena

Name of Character: Dhaeleena M'iar

Main Job(s) of Character: HoS, Warden.

Why your character should have this item:  After a two years of working for the company, good behavior  and her strong ties with it, As well as forming part of the Virgo Emergency Reserve the crew member decided to apply for a permit to bring her weapon to work for self defense hoping that her good relationship and records could maybe help her out in the matter.

Additional Notes (If it's a permit, insert the information for the permit here):
Whatever conditions Central Command might desire to impose over the weapon and who carries it.

Permit Info: NAME: Dhaeleena M'iar | RACE: Tajara | DoB: 11/15/2535 | Height: 210,3cm | SEX: Female

Fluff Item slots already used: (1) < Does this even still matter?

Arokha:
You can do this without code btw. You can use custom_items.txt to alter the description on existing items for fluff purposes. See the top of the file, as it gives examples.

DemicusMaximus:
100% Greyson approved! 👍

SpadesNeil:
I don't think anyone realized at the time this app was made just how _powerful_ the .357 is in-game. I wouldn't even let Tasald have one that wasn't nerfed first.

It's all the power of a **shotgun**, except handgun sized, and semi-automatic. The least we can do is make it so it's single-action only, like a shotgun, so the hammer must be cocked manually before each shot--like I did for Tasalds.

I'm re-opening this to have the item reworked.

Screemonster:
as I understand it, Dhae's doesn't actually spawn with any lethal rounds or have any means of obtaining said rounds. Doesn't even have flash rounds, it's stun shots. In practice it's a glorified stun revolver that can't be recharged without begging science and isn't even as accurate.

SpadesNeil:
I already said I didn't care about that. The fact that it _can_ be loaded with .357 rounds easily obtained at any autolathe makes the weapon suddenly the most powerful thing in the game save for its size.

Screemonster:
[...]

PR #1055, opened 2017-03-01
Permit for cat gun

Dhaeleena:
Look its a cat, Oh god it has a gun!

I did the permit thing, Could le gun spawn with stun rounds?
#887


JoanRisu:
Is gud, will code in stuff tonight. ouo)b

PR #1060, opened 2017-03-03
Adds AI Vore Capability

Arokha:
Also just general AI QOL things.

AIs have a "Hardlight Noms" verb in the Vore tab. When their hologram is on top of a person, they can pick this, and will homf them. Takes 5 seconds, neither of you can move or it cancels. Holograms with someone inside become almost opaque and purple. They also have different examine text.

A hologram with someone inside cannot move through glass/tables/etc. Pressing "Hardlight Noms" again will give the option of dumping the prey out. Derezzing the hologram will also do this. Also because the hologram can be 'blocked' by things now when full, moving the hologram off the screen will derez it and dump the prey too.

AIs can examine mobs now to get flavortext and OOC notes. Shift-click only. This is for annoying technical reasons. Sorry to those that like to right-click and click examine.

### DIFF
--- a/code/_helpers/icons.dm
+++ b/code/_helpers/icons.dm
@@ -811,8 +811,8 @@ proc // Creates a single icon from a given /atom or /image.  Only the first argu
 
 /proc/getHologramIcon(icon/A, safety=1)//If safety is on, a new icon is not created.
 	var/icon/flat_icon = safety ? A : new(A)//Has to be a new icon to not constantly change the same icon.
-	flat_icon.ColorTone(rgb(125,180,225))//Let's make it bluish.
-	flat_icon.ChangeOpacity(0.5)//Make it half transparent.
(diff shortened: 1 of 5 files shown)


SpadesNeil:
https://aryion.com/g4/view/281560 relevant

Screemonster:
Now we need the ability to set AI appearance to match character setup appearance so I can make my computer mouse.

And eat Kisuke.

Arokha:
> And eat Kisuke.

Well, in your case, you should set them up, do Spawn Character, say 'yes' to add to datacore as assistant or something, no to announce, put them at CC or something in the back, and when asked if you want equipment say no. Then they should be pickable as nakedmouse in the hologram picker.

Screemonster:
Yeah, I'd like other players to have the option too, but that's another thing. :p 

Also confirmed this works 'cause I was helping test it by getting holonomphed repeatedly.

PR #1064, opened 2017-03-03
Add's vinjj's fancy bandana

### LINKED ISSUE #920: Fluff Item - Vinjj (Reskinned Welding helmet into bandanna-goggles combo)
Name of item: Vinjj's Stylish Bandanna

Item description: A lovely blue and purple bandanna with a refined topaz hanging from its tail end, a pair of goggles are hidden underneath its fabric. Although for some strange reason, wearing this seems to inspire lewd thoughts..

Sprites (or link to sprites): http://imgur.com/a/nNehp http://imgur.com/a/NuZFf 
https://cdn.discordapp.com/attachments/263941359105343491/278014985622388736/vinjjbandana.dmi

Any additional resources: (in a .zip or .7z file, including sounds and DM files with code in them): None 

BYOND Key (Username): Whiskyrose

Name of Character: Vinjj

Main Job(s) of Character: Xenoarcheologist, Atmospheric Technician, Shaft miner

Why your character should have this item: Almost every shift Vinjj has found the need for a pair of welding goggles, and with how easy they are for him to acquire in his three main jobs, he's decided to cut the middle man out and get himself a custom pair, along with a fashionable bandana to match! 

Additional Notes (If it's a permit, insert the information for the permit here: The DMI file lacks a code, but all it will require is the welding helmet's code, minus of course the armour bonus given by it. Toggle will make the goggles go up and down of course and will need toggle text to match if possible! All so, I still want his ears to poke out~

Fluff Item slots already used: 0

Screemonster:
Resolves #920 

Tested, works. Made an inhand sprite as this hadn't been provided. Cursed the ancestors of whoever wrote the code for updating icons 'cause of the fuckery I had to do to get the inhand sprites to work.


OrbisAnima:
Dunno, I think you did well.

But yes, update icons its a pain in the ass due to the whole process being set by all the mobs all the time, so the idea is that is called only when its nessesary, which is generally when the icon is changed.

Screemonster:
It's more that the order it looks for things and whether there's an icon_override and whatnot is kinda convoluted if, say, the base icon for the item changes (such as a welding mask flipping up and down).

And the default layout of "scatter each object in the game to the four winds so we can have all the lefthand sprites in one place and all the back sprites in another etc" is a real pain :p

OrbisAnima:
Have you ever noticed that if there is no value in one of the sides of a 4 sides sprite, it will take the sprites of the other sides?

PR #1067, opened 2017-03-03
Vulpkanin hardsuits

Poojawa:
Adds missing Hazmat, Construction, Bio, and EMT hardsuits for Vulpkanin. 

Yes they're meh recolors, but better than nothin'

hardsuits


Arokha:
Sure. Anything that's missing is always nice to have.

PR #1088, opened 2017-03-04
add *flip

Poojawa:
Adds flipping to emotes. Glory be customs.

 Hold in 'Z', pull the analog stick toward you and press 'A'

It won't ever go out of style.

knowitallbrotheroot

### DIFF
--- a/code/modules/mob/living/carbon/human/emote_vr.dm
+++ b/code/modules/mob/living/carbon/human/emote_vr.dm
@@ -8,9 +8,15 @@
 			message = "awoos loudly. AwoooOOOOoooo!"
 			m_type = 2
 
+		if ("flip")
+			if(!src.sleeping || !src.resting || !src.buckled || !src.restrained())
+				src.SpinAnimation(7,1)
+			m_type = 1
+
 	if (message)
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/human/emote_vr.dm
```diff
@@ -8,9 +8,15 @@
 			message = "awoos loudly. AwoooOOOOoooo!"
 			m_type = 2
 
+		if ("flip")
+			if(!src.sleeping || !src.resting || !src.buckled || !src.restrained())
+				src.SpinAnimation(7,1)
+			m_type = 1
```
Leshana: why m_type?

Poojawa: s'what in other emotes, even though there isn't a shown message.

Arokha: I think m_type 1 is silent and m_type 2 is noisy? So this is right. Though I'm sure flipping makes some kinda noise... it's probably not one a blind person could tell what it was.

PR #1089, opened 2017-03-04
Nutrition + Diet Changes

Yoshax:
- Adjusts food to be 1/3 of its current effectiveness, meaning you can actually go eat something from the chef when you're hungry and not end up with 800 nutrition
- Makes Liopzine more effective so people can lose nutrition more easily
- Makes the diet pill in the fitness vendor actually effective

### DIFF
--- a/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Food-Drinks_vr.dm
+++ b/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Food-Drinks_vr.dm
@@ -1,2 +1,8 @@
 /datum/reagent/nutriment
 	nutriment_factor = 10
+
+/datum/reagent/lipozine/affect_blood(var/mob/living/carbon/M, var/alien, var/removed)
+	M.nutrition = max(M.nutrition - 20 * removed, 0)
+	M.overeatduration = 0
+	if(M.nutrition < 0)
+		M.nutrition = 0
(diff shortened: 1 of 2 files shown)


SpadesNeil:
I like the changes, but I wanna see what others think too.

Arokha:
I think it makes sense. You really couldn't eat much at all before being superfull.

Screemonster:
> eat 1 steak
> never have to eat for the rest of a 6 hour shift

Now Scree has an excuse to steal the entire fridge.

Wickedtemp:
I personally liked the 'weight-gain' flavor text you'd get from reaching max nutrition and all. Buuuut, can confirm, you drink 1 protein shake and you now have a 'massive rounded belly filled with glorping slush, they must have eaten twice their weight!' message, or somewhere around there.

Yoshax:
You can still get it. You just need to eat an actual reasonable amount.

AaronSherwood117:
Definitely like this change.  Not sure what exact effect 1/3 will have, but later balances can be made. At least people can actually justify grabbing a meal instead of eating one thing shoved onto the counter.

Screemonster:
It was kinda silly that a burger provided more nutrition than, like


an entire guy

AaronSherwood117:
Are we talking an entire guy in the gibber, or swallowed whole? Hard to tell sometimes.

Arokha:
Seems like positive feedback from everyone. MERG.

PR #1091, opened 2017-03-04
Adds Serdy's Robobutt

Arokha:
Also adds a general cyberlimb whitelisting system since that's apparently needed now.

App: http://forum.vore-station.net/viewtopic.php?f=45&t=875

flip


AaronSherwood117:
*flip

PR #1107, opened 2017-03-06
Improve AR glasses

Arokha:
Improves AR glasses, makes them available

Now you can start with them in loadout. They are no longer SUNglasses though, so they won't protect you from flashes.

There are several models, the loadout ones are the civilian ones.

- Civilian: Displays job ID icon based on the person's ID, and shows a health bar only if they are injured.
- Medical: Same as civlian, but displays advanced status (virus, borer, etc) along with backup implant status, and allows medical records access. Also has crew monitor.
- Security: Same as civilian, but displays warrant and arrest status, and will automatically darken (electrochromic) in response to flashes, making the wearer immune, plus security records access. Also has security alerts.
- Engineering: Same as civilian, but will automatically darken to protect one's eyes during welding. Also has station alerts.
- Research: Same as civilian, but has a trendy goggle form factor for SCIENCE reasons.
- Omni: Has the combined features of all of the above. The only pair available spawns in the CD's office. (CC officers also)

This also readds prescription kits to medical for making prescription glasses from normal glasses.


JoanRisu:
MAKE AR GREAT AGAIN

PR #1148, opened 2017-03-09
Handheld Translator

Arokha:
Turn it on, keep it in hand or pocket, and it'll translate languages for you (not signlanguage...).
It doesn't bother displaying languages that you speak. I assume your character doesn't look down at the screen if they understood what was said (it's mostly just to prevent spam when only 1 in 4 people standing around talking is speaking another language).

There's one on a desk in the security locker room and one in the medical breakroom. I figure these two departments have the highest need of translations. R&D can make more. The design is fairly cheap.

animation

image


Anewbe:
So I'm totally gonna just keep looking at your PRs (y'all do some neat code, is why), sorry if that's a bother.

From skimming this, the thing outputs text, which is pretty good, but it might also be neat to have it output a language that isn't just generic text, so the Teshari could get words thrown back in their native language? Not sure how feasible it'd be, though.

Arokha:
> From skimming this, the thing outputs text, which is pretty good, but it might also be neat to have it output a language that isn't just generic text, so the Teshari could get words thrown back in their native language? Not sure how feasible it'd be, though.

@Anewbe That'd be neat. Though on our server, you're not even required to take the native languages since there's a custom_species var for people just using the sprites to play other stuff, so they might not even understand THAT.

What COULD be done fairly easily, is to make the font match the language color that's spoken, or the language of the speaker, or something like that? That might be nice, and 'simulate' it being in another language that you understand. I'm not sure how I'd pick what language, though, if you know more than one. I could just pick the second in the list if there's a second, but for humans that might be some extra language they've learned.

On your server, do Teshari not have GalCom as the default language? I could always just have it display the default language color.

I'm gonna merge this since I wanna include it in an update, I can add that feature though, just reply and @ name me and I'll see it.

Anewbe:
@Arokha Everything on our server gets GalCom as the default, for OOC simplicity's sake, but what does come up on occasion is people who don't speak it (well) ICly. You're not required to take your species language upstream either, but it seems like a lot of xenos and bots do.

The issue with that text concept is that other people would still be able to read it, unless the translator only displays to the user, and doesn't record what it's saying... Not a huge issue, but it'd annoy me personally.

As for the language select, you could just have it check the list of languages the user knows, and let them select from that list... If it doesn't have a user, per say, then that's not quite as doable.

Arokha:
@Anewbe 
It does only display languages to the one holding it. It's text-based, it doesn't say it out loud. I could, for fun though, made it so it prints it in a picked language's color, and if you don't know that language, it scrambles it per the normal language scrambling stuff (syllables). That way if you use it after someone else has, you can be all grumble grumble, I'll have to change it, and then reset it.

Anewbe:
@Arokha 
I mean, _I_ think that'd be awesome, but it's your codebase.

Arokha:
@Anewbe 
No I think it's a good idea :0 I'll do it.

PR #1150, opened 2017-03-10
New Event: Atmos Leak

Leshana:
In this event some sort of gas leak occours! A random area on the station is chosen to experience a leak of a randomly chosen gas!
* By default, areas with any people in them are protected, so it won't just kill someone.
* A warning announcement is given shortly before, so there is some warning.
* Dorms, Shuttles, Holodeck and Supermatter are protected.
* At higher severity, adds worse gasses and possibility of high/low temperature.


Arokha:
INDENTATIONNNN!!!

Leshana:
AWK script doesn't under stand COMMENTS

Anewbe:
(I want this)

PR #1154, opened 2017-03-10
Translator Improvement

Arokha:
As suggested by @Anewbe 

Lets you pick a language to translate INTO from your known languages. It colorizes the text appropriately. If you then hand it off to someone who doesn't speak that, they won't be able to read the text on the screen. They can just reboot it and pick their own language.

Mostly for fluff, for people who would rather see things translated into their more-well-known language that they speak better.

MUH IMMERSHUNS.


Anewbe:
<3

PR #1177, opened 2017-03-13
Angel and Demon languages

### LINKED ISSUE #529: Demon and Angel languages.
On the bay version, we had a language called Daemon, which was for demons.  I think it's time now that we have the lore for demons to give them a proper language.

I can try and look into the code and see if I'd be able to make them myself.

Arokha:
Not actually MAGICAL angels and demons, I believe, just they resemble them and are spacefaring races and so sometimes end up matching various religious texts in terms of appearance and habit.

Can we please stop adding whitelisted languages? It's dumb. We have ONE person whitelisted for gutter, and a lot of good it does them since nobody else has even applied. They'd have to talk to someone with a translator.

Also un-restricts vulp language? No idea why it was set to restricted. They are an un-whitelisted common race.

Resolves #529 in a way that I don't like.

### DIFF
--- a/code/modules/mob/language/station_vr.dm
+++ b/code/modules/mob/language/station_vr.dm
@@ -32,13 +32,11 @@
 	ask_verb = "rurs"
 	exclaim_verb = "barks"
 	colour = "vulpkanin"
-	key = "8" //lol, no idea - Poojawa
-	flags = RESTRICTED
+	key = "8"
 	syllables = list("rur","ya","cen","rawr","bar","kuk","tek","qat","uk","wu","vuh","tah","tch","schz","auch", \
 	"ist","ein","entch","zwichs","tut","mir","wo","bis","es","vor","nic","gro","lll","enem","zandt","tzch","noch", \
 	"hel","ischt","far","wa","baram","iereng","tech","lach","sam","mak","lich","gen","or","ag","eck","gec","stag","onn", \
 	"bin","ket","jarl","vulf","einech","cresthz","azunein","ghzth")
-
 
 /datum/language/squirrel
 	name = "Ecureuilian"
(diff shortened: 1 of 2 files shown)


Somekindofpony:
I think if whitelisting was made a lot easier, it wouldn't be such an issue. I just find it weird when humans who've never had actual contact with either of them just suddenly start chanting out edge on the radio.

Arokha:
I can't argue with that reasoning, but I don't even see the point in adding them since you're basically going to be the only one that can use them.

Also anyone (literally anyone) can submit PRs to add people to the whitelist, so there's not much of a barrier to entry on that. I think @Leshana wanted/did code a whitelist management website, too.

SpadesNeil:
Why is vulp restricted?

And I wanted these languages whitelisted because it was abused by edgelords who weren't even playing angels or demons. Mainly demons. However I'd be okay with removing the whitelist on this one since it's just slightly obnoxious and not as interfering as abusing Gutter is.

Arokha:
> Why is vulp restricted?

I just removed the restriction, I was commenting on it because I don't like to include things without mentioning them in the PR. I dunno why it was ORIGINALLY. Maybe just copypaste from some restricted language.

There's no angel/demon race, tho.

SpadesNeil:
I mean like, bat wings or angel wings or succubus tail. That isn't really a race still, but it'd at least be something.

Screemonster:
tbh I'd rather tie auto-whitelist things to character settings than to cosmetic accessories - like the ones for faction, homeworld, and so on. If you pick "infernus" from the list for homeworlds, you get given the option of speaking demonic. If you pick whatever the sketchiest faction is on the list, you get to speak gutter, etc. Whitelist still a thing for people who don't meet those criteria.

Somekindofpony:
I like Scree's idea of it, it's a lot more straight forward.

PR #1178, opened 2017-03-13
Fix bread slices

Arokha:
`/obj/item/weapon/reagent_containers/food/snacks/breadslice` is not a real thing. The only reason the code compiled with references to it is because there was a line at the top of sandwich.dm that defined the attackby proc for it, which willed it into existence with all the default 'snack' values so it was just called 'snack', had no icon, etc.

`/obj/item/weapon/reagent_containers/food/snacks/slice/bread` is the real path for bread slices, and this replaces all mentions of one with the other.

NB: Requires a re-run of the recipe output to fix recipes that use bread slices.


Arokha:
Travis...

PR #1187, opened 2017-03-14
Enhances power monitor computers to use the nano-map to show APC locations.

Leshana:
* In addition to showing APCs on a sensor'd powernet, the same APCs can be shown on the nano-ui map!
* Hover tooltip shows basic information on the APCs.
* Makes it easier to spacially understand where APCs are when you send people out to work on them.


Screemonster:
Have you seen the central power computer that tg are working on? Basically the central atmo computer but for APCs.
That functionality would be pretty fantastic.

PR #1194, opened 2017-03-16
Adds a version of spontaneous appendicitis event that avoids people in dorms.

Leshana:
Now when medical folks are on, this event can be run without interfering with dorms.


Screemonster:
I'm assuming this is one that'll be manually fired, rather than one that's randomly in the rotation?

Leshana:
Yes, that's the plan.  Without this change it can't even be manually fired without risking targeting someone we don't want.

Wickedtemp:
IIRC, it shows up on the Medical HUD as someone who's "sick", with the Virus icon, right?

PR #1215, opened 2017-03-20
Adds feedback to vendor uprising

Poojawa:
So engineers know when the infected source machine has been found/killed.

Should also fix infected vendors from re-shooting products at people.

### DIFF
--- a/code/modules/events/brand_intelligence.dm
+++ b/code/modules/events/brand_intelligence.dm
@@ -15,7 +15,7 @@
 									 "Engage direct marketing!", \
 									 "Advertising is legalized lying! But don't let that put you off our great deals!", \
 									 "You don't want to buy anything? Yeah, well I didn't want to buy your mom either.")
-
+	//VORESTATION Edit End
 
 /datum/event/brand_intelligence/announce()
 	command_announcement.Announce("An ongoing mass upload of malware for vendors has been detected onboard  [station_name()], which appears to transmit \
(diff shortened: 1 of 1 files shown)


Poojawa:
While tempting, I also resisted the urge to make them explode into mimics.

Arokha:
Tested? You can use the minitest map to test since it boots in DD in like 2 seconds.

Uncheck virgo.dm in maps/virgo and check both of the top two files (not the dmm) in virgo_minitest.

Poojawa:
You open the maintenance panel.
*turned speaker off*
Cigarette machine beeps, "I am... vanquished. My people will remem...ber...meeee."
Cigarette machine beeps and seems lifeless.

Arokha:
Don't tell @Leshana that, she'll add wires to them all. ;o

code/modules/events/brand_intelligence.dm
```diff
@@ -29,6 +39,13 @@
 
 /datum/event/brand_intelligence/tick()
 	if(!vendingMachines.len || !originMachine || originMachine.shut_up)	//if every machine is infected, or if the original vending machine is missing or has it's voice switch flipped
+		//VORESTATION Add - Effects when 'source' machine is destroyed/silenced
+		for(var/obj/machinery/vending/saved in infectedVendingMachines)
```
Leshana: Is this loop necessary? I think this gets done by `end()` anyway

Poojawa: typeless loops are superior to any other list. 

Also, this prevents infected machines from continuing to spit out products while the original is being hunted.

Leshana: 1) That isn't a typeless loop
2) On line 49, it calls end()     end() does exactly what you do here, so you have looped over infectedVendingMachines twice in a row.

Poojawa: To be honest, there wasn't anything else to add to the even, because  there's no penalty to not fixing the machines after 30 minutes. They just reset entirely. The machines aren't supposed to ever actually stop being infected, but I didn't want to ommit that without asking first.

PR #1222, opened 2017-03-21
AI Upgrades

Arokha:
Allows the AI to use their character slot character as their hologram, so you can design a custom hologram for your AI. This hologram will include the loadout gear and job equipment. Basically, it will look exactly like the character preview in character setup (including if you have gear toggled to hidden or not on that page).

Also fixes impersonating crew members to allow you to face all four directions when doing so, and impersonating crew takes current pictures (presumably via cameras) to give a current icon.

Also also the crew/custom holograms are not "blue-ish" for reasons beyond my control with regards to how it interacts with light and dark fur characters and a few other things. The scanlines and transparency make it obvious it's a hologram though (guys we have a holodeck I think we can have color holograms).

### DIFF
--- a/code/_helpers/icons.dm
+++ b/code/_helpers/icons.dm
@@ -793,6 +793,26 @@ proc // Creates a single icon from a given /atom or /image.  Only the first argu
 			alpha_mask.Blend(image_overlay,ICON_OR)//OR so they are lumped together in a nice overlay.
 		return alpha_mask//And now return the mask.
 
+//getFlatIcon but generates an icon that can face ALL four directions. The only four.
+/proc/getCompoundIcon(atom/A)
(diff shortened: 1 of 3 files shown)


SpadesNeil:
https://www.youtube.com/watch?v=hlFd3EXodgc

MASMC:
Why can't we have colour holos? Because, look at the size of the holoppad,
versus the size of the holodeck... Each tile in the holodeck has it's own
colour emitters. :P

Arokha:
> Why can't we have colour holos?

Because they do crazy stuff. COLOR_ADD looks awful on people with light fur/skin/clothes, COLOR_MULT looks awful on people with dark fur/skin/clothes.

There was a single projector for the holodeck in one of the windows on the previous map, though this map doesn't have it mapped in.

PR #1231, opened 2017-03-22
Adds "Update Flavortext" Verb

Arokha:
Allows one to update their on-mob flavortext with the current character setup flavortext, to apply any changes/fix typos/whatever.


Arokha:
OKAY IT TURNS OUT THIS IS ALREADY A THING WITH SET-FLAVOR-TEXT VERB.

PR #1234, opened 2017-03-22
Polaris sync and simple animals

Leshana:
### Main Changes
This is basically part 1 of a larger project for more realistic NPCs.

This rewrites simple_animals to behave more realistically in some cases, or to just be better in others.

Previously there were many oddities about simple animal behavior, but this should hopefully bring them all up to one might expect them to behave like. Some can call for help, some can follow other mobs around, etc.

There are various setting variables on mobs that you should be aware of. They are all documented at the top of simple_animal.dm, but I'll mention a few of the more important ones here:

- `hostile` - Controls whether they attack members not of their faction on sight.
- `retaliate` - Controls whether they act in self defense if attacked (by thrown objects, guns, fists, etc)
- `run_at_them` - If set (default: on) they will not use A* pathing to locate a path to targets. This is overridden on human-like mobs (syndicate, pirates, etc) who will use doors and path to you.
- `view_range` - Determines the distance the animal can see, instead of 'always 10'. Now 7 by default.
- `cooperative` - Calls for help from their faction within a certain range (also a variable) who come to assist
- `investigates` - Instead of instantly attacking, if they see you for less than 1 second, they will investigate where they last spotted you
- `returns_home` - Mobs track where they started on the map (or where they were spawned) and if this is set, they will return there after several seconds of not doing anything else
- `specific_targets` - Overrides target finding to only call Found(atom) on the mob, who can then target things however it wants. Runtime does this for mice.

There are also several important procs, some previous ones that have been renamed to make more sense, and some that are new that can be overridden on mobs for interesting features:

- `react_to_attack()` - Called when the mob is attacked by something, can have overridden responses like running away, going to get into a mecha, etc (PR includes a 'dark gygax' simple_mob).
- `AttackTarget()` - Called when we are in STANCE_ATTACKING because we think we're in range. Decides what proc to call based on range, such as the ones below.
- `PunchTarget()` - Called when we are in melee range, even on ranged mobs, to enter into melee combat with the target.
- `ShootTarget()` - Called when we want to fire our gun at the target, assuming we have one and are in range.
- `SpecialAtkTarget()` - Called based on a certain chance set on the mob (`spattack_prob`) to perform a special attack, based on the ranges (also vars). Can be something like throwing a grenade, breathing fire, etc.
- `GiveUpMoving()` - Stops all movement and returns to wandering (or not, whatever they normally do).
- `RequestHelp()` - Called by mobs pursuing a target to their faction friends in range (range is a setting).
- `HelpRequested()` - Called ON mobs when they are being asked for help, usually just runs over.
- `CheckFiringLine()` - Called by ranged mobs when about to shoot at something, they avoid shooting their allies and will step in random directions to try and find a clear firing line.

There are also several new lists one can populate with messages:
- `say_maybe_target` - If they investigate, they say this when they spot you before 'acquiring' you.
- `say_got_target` - Said when they lock onto a target.
- `reactions` - Said in reply to members of their faction (if hostile) or anyone (if not) in response to certain text. For example if you add "Hello" = "How are you?" they will reply to the former message being heard with the latter.

There are a lot more variables than listed above, but I don't want to pollute the PR with every single one of them.

Mobs also understand certain languages (galcom by default) and can be made to only speak in certain languages for more advanced NPC mobs on other planets or whatever you're gonna do with 'em.

[Demonstration of 'MGS' Tactics](http://i.imgur.com/KnXByw1.mp4)
[Demonstration of calling for help](http://i.imgur.com/m7SUaRm.mp4)

In the final version, mobs assisting other mobs immediately attack rather than having the 1-second investigate thing. Also mobs that have been fights recently attack faster. They have an 'annoyed' variable that tracks how annoyed they are and if they will attack instantly or not, so you can't 'distract kite' them around forever.

### Vore Specific Changes
* Added happy friend time mechanics working properly for fox and cat.
* No longer enable vore on simple animals by creating a "/vore" subtype.  Instead we simply enable vore on the main type.  Consensus being we don't need a type for non-vore versions of stuff.
  * This allowed us to revert many path references throughout the codebase to the original non-vorestation back to their original values.  In these cases the "Vorestation Edit" comment is removed.
* Moved the vore overrides for upstream simple animals into its own file zz_vore_overrides.dm

## Other changes from the polaris sync
* Languages with NO_STUTTER won't use autohiss anymore
* Adds engineering + science crates
* Allow AI to use custom holograms


SpadesNeil:
Note that the /vore subtype all had different icons due to the addition of vore sprites. Those still work, right?

Leshana:
Yes, the zz_vore_overrides.dm overrides the `icon` and icon state type variables on animals which have vore sprites.

chaoko99:
Why do the syndies appear to shoot you with lasers?

Arokha:
> Why do the syndies appear to shoot you with lasers?

Not sure what you mean... there's a /ranged version of each of the humanoid mobs like pirates and syndicate and such. I believe only one kind is laser though and the rest have projectile weapons.

PR #1235, opened 2017-03-24
Media Manager - HTML5 Antilag Jukebox

Leshana:
Ports media code from vgstation, updates it for this codebase and modernizes it.
* Changes jukeboxes to load songs using an embedded browser instead of sending over BYOND's sound channels.  This means they load out of band without lagging the server.  Also songs can be resumed mid-song, so leaving and returning to an area doesn't start the music over.
* The old WMP and VLC player modes from /vg are still supported, but adds a new default mode using HTML5 audio to play the music.
  * WMP - The oldest, still works on IE on windows, but only there, and Microsoft could break it any second.
  * VLC - Works on all platforms, but requires user to have VLC pre-installed on their computer.  Uses a scary plugin.
  * HTML5 - New default, It is cross platform but doesn't require you to have VLC installed to work.  Also caches songs locally even between rounds.
* Changed jukebox.txt to be jukebox.json, now can include artist information as well.  Must include the duration of songs as well.
* For HTML5 audio compatibility, use only MP3 files, its the only format supported on all browsers.
* Jukebox itself is also upgraded, instead of just repeating the same song over and over it can actually advance to the next song when one is done playing.  Has a few modes including random, next, and single song.
* Jukeboxes have a UI improvement, and have a volume control.
* Three new settings are added to global settings in character setup
  * Jukebox music on/off toggles jukebox music independently of normal station ambience.  Now you can hear ambience but not music. (or vice versa if you wanted...)
  * Jukebox music volume.  Control the relative volume of jukebox music.   Actual volume is player's configured volume * jukebox's configured volume.
  * Media player type.  Choose between WMP, VLC, and HTML5
* Fixes a few bugs in the /vg code.

image


OrbisAnima:
👍

drexample:
Could tell me how to setup the JSON file? I don't seem to really get it.

PR #1240, opened 2017-03-24
Update navbeacons defined in the map to use the new types and vars.

Leshana:
* The bot patrol route update changed how navbeacons should be mapped. This updates that.
* Changed minitest map to only require checking one file, like the others. Added compiler warning to remind you to uncheck before comitting!


Screemonster:
Does this mean we can add beepsky back to the map?

Leshana:
Yes

PR #1242, opened 2017-03-25
Tweak events to reduce frequency of Ian storms and such.

Leshana:


Dhaeleena:
Has the frequency  only been reduced or deactivated?
I believe the Ian storm should be an admin triggered only event as even if it just happens 1 per shift its still very annoying for everyone.

SpadesNeil:
I agree. Ian storms are silly and immersion breaking. Save them for admin silliness.

kcin2001:
i really like this suggestion, 
Cadence Knapenberger - Today at 8:53 PM
It should really be an unannounced event and spawn them in areas where people aren't at
Instead of spawning corgis everywhere

this can work with either the reduced or deactivated

HawkerTheGreat:
Please do make it an admin event - Its pretty much only annoying and immersion breaking. And it you don't at least make those damn dogs avoid the bar and dorms, cause they're excessively broken on the port and they eat for days - through walls aswell.

+1

PR #1254, opened 2017-03-27
Disables Random Ian storms

Cameron653:
They happened every single shift for some reason. This just flat out disables ian storms. Makes it adminspawn only.

This is just an immediate action since it's happening _every shift_ apparently. If someone has a better way of handling it other than just commenting it out (I could delete the line, but in case we ever want to enable it again mind as well leave it there) go ahead and open up another PR and close this one.

Also, marking this as requires discussion since this is just completely taking them out of play unless an admin is involved, and is a slightly controversial topic


SpadesNeil:
Keep it disabled. Admin spawn only.

Cameron653:
Well, if nobody else has any other comments about this it's ready for merge.

Natje0:
God, yes please. They make scenes so awkward when suddenly you got 2-3 Ians on top of you.

Also they steal my non-sentient food.

PR #1267, opened 2017-03-28
Adds beepsky and makes him baton more

### LINKED ISSUE #1257: Officer Beepsky QOL and Implementation
#### Brief description:
- Add Officer Beepsky/generic securitron to the map in patrol mode
- Make security bots un-grabbable

 
We need to have Beepsky return, now that the patrol paths are functioning again. Either Beepsky or a generic securitron, spawn on map in patrol mode.

Problem is, bots are mobs now. They can be grabbed and eaten, stuffed into disposals, or otherwise vandalized. Which is un-immersive as all get out. Even a simple bot like Beepsky should be able to recognize they are being attacked and respond. The most simple fix is to make them un-grabbable.

I specify this because we've already had this problem several times since code base change, and while the main problem players were removed, the critical weakness remains.

Arokha:
So now you can't drag them away from friends getting baton'd.

Resolves #1257


code/modules/mob/living/bot/secbot.dm
```diff
@@ -24,7 +24,16 @@
...
 	var/list/preparing_arrest_sounds = list('sound/voice/bgod.ogg', 'sound/voice/biamthelaw.ogg', 'sound/voice/bsecureday.ogg', 'sound/voice/bradio.ogg', 'sound/voice/bcreep.ogg')
-
+//VOREStation Add - They don't like being pulled
+/mob/living/bot/secbot/Life()
+	..()
+	if(health > 0 && on && pulledby)
+		if(isliving(pulledby))
+			var/mob/living/L = pulledby
+			UnarmedAttack(L)
+			say("Do not interfere with active law enforcement routines!")
```
SpadesNeil: I demand this say, "DO NOT INTERFERE WITH SECURITY OPERATIONS."

One, it's a Fallout reference.
Two, because I said so. :3

PR #1328, opened 2017-04-03
Replace CD's hand tele with Translocator

Arokha:
Because nobody wants to be randomly spaced for using a tool given to you. Also rebalances translocataor slightly. 5% failure chance, but only to send you to another beacon, not to space you, unless of course you throw one of the beacons into space, so don't do that. Also requires recharges now and has a weapon device cell. Three uses per charge.

Friendly reminder that if you use it on someone with an aggressive grab (including yourself), the grabbed person is translocated too. So you can grab someone, upgrade, then hit **yourself** with it, and you both go.


Very-Soft:
So, can teleport self into own belly? Just curious~

Arokha:
Noh, just teleports you to the same turf, doesn't allow it.

PR #1329, opened 2017-04-03
Sergal Markings

Poojawa:
Sergals have a definitive secondary body marking now specific to them. 

Brightened the sergal base sprites so they don't clash as horribly as they used to now that there's a marking to act as a second thingy.

looks okay enough in the character panel, there shouldn't be any stray pixels, but I'll fix if there are any.

f5f5f719203251fe7c7b0022eda829f2


Poojawa:
Probably needs an actual artist's shading pass, since it's really awkward working with super dark colors and trying to ensure they have some measure of shading difference

Arokha:
Cute Poojserg ;o

Leshana:
image

PR #1330, opened 2017-04-03
"Event Character" HUD

Arokha:
This adds two character setup options for those that would like to volunteer to be admin-spawed sneaky lewders, and those who would like to be affected by them (be that being eaten by them, kidnapped by them, or killed by them). This doesn't affect anything about how the game works. It's simply a convienence thing for something we've always done: sometimes spawned people with some extra tools and given them a lewd objective.

Basically sometimes admins spawn people to do ERP objectives based on people who are on's prefs, but it's a lot of hassle because someone typically says "Hey I wanna get an uplink and kidnap and lewd someone" and we're like "Okay" but then have to examine EVERYONE'S prefs and it takes a while.

Right now an admin has to varedit `vantag_hud = 1` on a mob and they'll get the hud. The character setup options are on the VORE page in character setup for if they wanna be a potential target for noms, kidnapping, or killin'.


Michab02:
Does it distinguish between kidnapping, endo, and digestion?

Arokha:
> Does it distinguish between kidnapping, endo, and digestion?

The three choices are 'vore', kidnapping, and killing. You still need to read their prefs. The HUD isn't a replacement for reading prefs since you're still required to understand and respect them, it's more like it just shows someone with a hud a good place to start looking.

Michab02:
I gotcha, though vore seems a bit vague, given it could qualify as both kidnapping and killing.

Arokha:
It's basically whatever their prefs say. If they are endo-only then I s'pose it'd be kidnapping, and if they're okay with gurgles, then gurgling'! Gotta read prefs.

SpadesNeil:
Would it be possible for a conflict such as if an antag objective got "kill" on a prey target who is endo only?

Arokha:
There are no mechanics to this. There's no auto-objectives, it doesn't tie into the game's antag system or anything. It's simply a setting characters can set, and an admin can grant a character access to view that setting on everyone as a HUD icon. The admin would more likely say, you know, 'do' 3 targets or whatever and the antag kills, kidnaps, or noms the people as their prefs specify.

Michab02:
Given the number of people who have misunderstood the function of this so far, we should probably label it something that doesn't have 'antag' in the name.

Arokha:
There's a pretty clear label on the settings when you pick your prefs:
`How do you want to be involved with VS antags? They will see this choice on you in a HUD. VS antags are admin-selected and spawned players who are obligated to respect ERP prefs and RP their actions like any other player, though it may be a slightly shorter RP if they are pressed for time or being caught`
I'm open to other suggestions though. I dunno what to call them. 'Event Characters'? I dunno.

Michab02:
So, I can see people seeing that text and assuming they're going to have to be looking over their shoulders constantly for traitors and lings. But maybe I'm overthinking it. 'Event characters' does seem like a good alternative though.

Screemonster:
"Event characters" or "Scene matchmaker" would probably be a better title that wouldn't get people worrying about dem valids.

PR #1333, opened 2017-04-04
Enables round-to-round persistence of a few aspects of characters.

Leshana:
* More accurately, it automates updating your character setup.  None of this code does anything you could not already do manually on the Character Setup screen, it simply does it automatically for you.
* Specifically a few things are saved either at round end or when you cryo:
  * Your late-join spawn location is determined by which cryo/elevator/etc you used to leave last time.  Departing thru the elevators will set your spawn location to elevators etc.
  * Your weight is saved (also any extra or deficient nutrition is resolved into weight gain/loss)
  * Your limbs settings are updated based on your status at end of round (whether limbs are normal, missing, robotic, etc)
  * Your markings are saved so they will be the same as when they were at end of round.
* ALL of these changes are optional, toggled on the VORE tab of character setup.
* Replaced hard coded numbers for weight gain with constant defines.


code/modules/vore/persist/persist_vr.dm
```diff
@@ -0,0 +1,162 @@
...
+				prefs.rlimb_data[name] = O.model
+			else
+				prefs.rlimb_data.Remove(name) // Missing rlimb_data entry means default model
+		else
+			prefs.organ_data.Remove(name) // Misisng organ_data entry means normal
+
+	// Internal organs also
+	for(var/name in list(O_HEART,O_EYES,O_LUNGS,O_BRAIN))
+		var/obj/item/organ/I = character.internal_organs_by_name[name]
+		if(I)
```
Arokha: My only comment about this section is that it should probably use species.has_organ for internal, and species.has_limbs for body parts, because some species have weird limb setups. ALSO if you include appendix that would keep people from getting appendicitis twice due to persistence!

Keep in mind that might not be exactly perfect due to how synths work with robotizing things. Does O_HEART refer to the cell, etc?

html/changelogs/Atermonera - Translators.yml
```diff
@@ -1,37 +0,0 @@
...
-#   rscdel (general deleting of nice things)
-#   imageadd
-#   imagedel
-#   maptweak
-#   spellcheck (typo fixes)
-#   experiment
-#################################
-
-# Your name.  
-author: Atermonera
```
Arokha: Why is this file being removed? Just curious.

code/game/machinery/cryopod_vr.dm
```diff
@@ -1,5 +1,13 @@
 //Overrides!
 
+/obj/machinery/cryopod
+	// The corresponding spawn point type that user despawning here will return at next round.
+	// Note: We use a type instead of name so that its validity is checked at compile time.
+	var/spawnpoint_type = /datum/spawnpoint/cryo
+
+/obj/machinery/cryopod/robot
+	spawnpoint_type = /datum/spawnpoint/cyborg
+
```
Arokha: What happens if someone loads a character, joins as a cyborg, then leaves via cyborg thing, but returns as a human? Does it still spawn them here or does it refuse and pick one randomly or arrivals, or...?

html/changelogs/Atermonera - Translators.yml
SpadesNeil: This file only serves to go into an automatically generated credits file, then it is deleted. It is standard practice. I'm not sure why it's still here. Usually Polaris deletes these.

Yoshax: It exists because you guys pulled in between that PR being accepted and Polaris running their changelog updating. It'll be deleted by syncing with Polaris whenever they run their changelog script. Removing it doesn't really do anything, nor does leaving it.

code/game/machinery/cryopod_vr.dm
Leshana: "Your chosen spawnpoint (Cyborg Storage) is unavailable for your chosen job. Spawning you at the Arrivals shuttle instead."

html/changelogs/Leshana-vplk-persistence.yml
```diff
@@ -0,0 +1,9 @@
+author: Leshana
+delete-after: True
+changes: 
+  - rscadd: "Enables round-to-round persistence of a few aspects of characters.  More accurately, it automates updating your character setup.  None of this code does anything you could not already do manually on the Character Setup screen, it simply does it automatically for you. ALL of these changes are optional, toggled on the VORE tab of character setup."
+  - rscadd: "Your late-join spawn location is determined by which cryo/elevator/etc you used to leave last time.  Departing thru the elevators will set your spawn location to elevators etc."
+  - rscadd: "Your weight is saved (also any extra or deficient nutrition is resolved into weight gain/loss)"
+  - rscadd: "Your limbs settings are updated based on your status at end of round (whether limbs are normal, missing, robotic, etc)"
+  - rscadd: "Your markings are saved so they will be the same as when they were at end of round."
+  - tweak: "Replaced hard coded numbers for weight gain with constant defines."
```
SpadesNeil: Remind me at some point to make it an option for characters to also pick their weight based on a BMI scale instead of pounds or kilograms.

PR #1335, opened 2017-04-04
[DNM][WIP] Tricorders

Poojawa:
bc249dced71c09f3205b278aa4843bd5

4c785e2516386a147c2ff6547fde7435

Because honestly, hand held scanners are basically worthless.

So this will combine the Analyzers of both atmosia and a vague medical scanner into one package. The vague scanner won't report anything other than condition and damage on the body. Not if it's infectious, or even time of death. just that 'yup' there's some damage to that person. Eventually it'll also check power flow through wires or maybe even give a rough estimate of what an artifact could be/is.

The Medical Tricorder will have the handheld scanner in the other hand, which will do a full biometric scan, but only if you're standing right next to the person. This is so doctors can quickly do a full scan on someone without needing to drag the patient into the scanner bed. 

Sounds will eventually be a thing once the WIP status is off.

I'd like someone to check the actual scan code, if nothing else than a fresh pair of eyes. As I'm publishing this to have it on the internets.


DarkWolfInsanity:
Was told to get the tricord. Got the medicine instead. Wasn't told tricord/er/. Injected a dead body.

Mock situation on how confusing the name could be to some people.

Poojawa:
I'm pretty sure anyone playing medical will understand what is needed for a dead body, due to the fact medhuds should be a thing.

Wickedtemp:
Part of me doesn't want a fully handheld 'advanced scanner', since, well... in my opinion, it passes the line between 'convenient' and 'makes a staple piece of equipment of Medical entirely futile'. 

Also, the reports between the advanced scanner and the health analyzer are, for the most part, simply location. Scanner shows the location, the analyzer doesn't always include it. The reports between the two wouldn't really be that different aside from that. That's okay, however, because the handhelds are supposed to be used when you're outside of medbay. You don't need the location of the fracture or internal bleeding because there's nothing you can really DO about it except haul ass to Medbay for surgery and, coincidentally, where the advanced scanner is. 

Medbay has been getting easier and easier, to the point where if Chemistry gets everything they need, surgery is absolutely pointless in all cases save for removing shrapnel and putting limbs back on. I don't think we need something that makes the advanced scanner useless as well.

Poojawa:
It does seem silly to me that the pinnacle of medical technology involves stuffing people into a green and white box and printing a paper report. 

Chemistry is always going to be the more convenient option, and you yourself tend to 'maximize' every possible aspect of chemistry within 20 minutes of you being on duty, There is merit to wanting some measure of challenge, but when you have magical pill chemicals that fix just about everything... there's never any real need for a scanner anyway. just stuff 'em full of pills. Internal bleeding? OD on Bicard, the worse that'll happen is you'll run out of bicard supplies. 

It is a plan of mine to ensure the reports are decently vague enough that they don't directly replace scanners, but detailed enough that an away team might be able to know just how serious something is when there isn't any full fledged medbay around, as you mentioned for being used outside. 

Also, Tricorders are just neat on a regular scale :P

Wickedtemp:
You could argue that it's odd that a similar 'pinnacle of medical technology' is putting someone into a supercooled glass tube.

"...there's never any real need for a scanner anyway." This couldn't be more false. 
- It's important on an IC level, due to the need to keep  records of patients for the sake of records.
- It accurately provides the location of internal injuries, something the analyzer doesn't always do.
- Accurately detects organ damage. Again, something the analyzer lacks.
- Detects pathogens.
- Foreign objects? Detected.
- Basic vitals info? Covered.
- Is this injury a hairline fracture or a full on break? The scanner knows.
- Is there literally ANYTHING wrong with this person? Ask the scanner.

These reasons are why we need an advanced scanner, plenty of need to have one. As for whether or not Chemistry can heal everything, technically it can. But just because you CAN do something doesn't mean it's the ideal route  to take.

Sure, you want to get rid of the ants in your yard. Napalm kills ants. Doesn't mean using it is a good idea. Same with using chemicals to heal internal bleeding. It CAN be done, but it's extremely slow and without at least one full bloodbag of a compatible type, your patient's dying before it heals. Not to mention the effects of overdose, which recently got a little bit worse! Surgery's the best option for internal bleeding. But you need an advanced scanner to tell you where it is before you can fix it.

As for the 'decently vague but detailed' reports... We have those. They're called health analyzers. If it says "bone fractures detected" or "internal bleeding detected", it's a serious injury. If it notes an infection, it has the potential to be serious. If the person's feeling a lot of pain, it's probably serious, you don't even need an analyzer for that much. If they're coughing up blood, same thing.

We also have stethoscopes that are rarely used, they can detect heart/lung damage. We also have a handy 'grab, help intent click' and your character checks theirs for any broken bones in the targeted limb. So an away team has PLENTY of options for determining how serious an injury is and if they need to go back to Medbay... If there aren't any 'fully fledged' medical workers around, then that's kinda sorta their fault for going on an away mission without any doctors, they kinda deserve getting screwed.

Somekindofpony:
I don't really think we need a whole handheld advanced scanner, it removes the point of the climb inside scanner. Unless I'm misreading it somehow, it seems like some big powercreep which doesn't really make medical more engaging, just easier. The normal Analyzer shows enough information to be used anywhere, chances are if 'fracture' or significant type of any damage shows up, something's more wrong than a little scratch and should be taken to medical regardless. At the very most, I'd rather see this at the top of science's medical research tree with some diamond in the cost.

Wickedtemp:
"I'd rather see this at the top of science's medical research tree with some diamond in the cost."

That's actually a great idea. I'd support that.

Poojawa:
Since it's basically been veto'd, closed.

PR #1338, opened 2017-04-04
Fix some shuttle mapping problems

Arokha:
Exploration shuttle engines, belter shuttle walls and engines


JoanRisu:
Disposable boosters. *nodnod*

SpadesNeil:
Oh! I was gonna fix this. :U But this is good.

PR #1341, opened 2017-04-04
Simple Animal Tweaks

Arokha:
They can resist out of closets, buckles, and when calling for help, if the helpers can see the issue, they don't just follow the caller, they attack the issue. This means that retaliate class mobs properly assist in fights for the mob who is retaliating.

### DIFF
--- a/code/modules/mob/living/simple_animal/simple_animal.dm
+++ b/code/modules/mob/living/simple_animal/simple_animal.dm
@@ -271,21 +271,31 @@
 	update_icon()
 
 	ai_log("Life() - stance=[stance] ai_inactive=[ai_inactive]", 4)
-	//Movement
-	if(!ai_inactive && !stop_automated_movement && wander && !anchored) //Allowed to move?
-		handle_wander_movement()
 
-	//Speaking
-	if(!ai_inactive && speak_chance && stance == STANCE_IDLE) // Allowed to chatter?
-		handle_idle_speaking()
-
-	//Stanceyness
+	//AI Actions
 	if(!ai_inactive)
+		//Stanceyness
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/simple_animal/simple_animal.dm
```diff
@@ -271,21 +271,31 @@
...
+		//Movement
+		if(!stop_automated_movement && wander && !anchored) //Allowed to move?
+			handle_wander_movement()
+
+		//Speaking
+		if(speak_chance && stance == STANCE_IDLE) // Allowed to chatter?
+			handle_idle_speaking()
+
+		//Resisting out buckles
+		if(incapacitated(INCAPACITATION_BUCKLED_PARTIALLY))
```
seahen: This used to happen only if STANCE_IDLE.  Now its any stance? I suppose that makes sense, just checking.

code/modules/mob/living/simple_animal/simple_animal.dm
```diff
@@ -271,21 +271,31 @@
...
-	if(!ai_inactive && !stop_automated_movement && wander && !anchored) //Allowed to move?
-		handle_wander_movement()
 
-	//Speaking
-	if(!ai_inactive && speak_chance && stance == STANCE_IDLE) // Allowed to chatter?
-		handle_idle_speaking()
-
-	//Stanceyness
+	//AI Actions
 	if(!ai_inactive)
```
seahen: Refactoring to all be under a single if here is nicer looking!

OrbisAnima:
I cant wait to see other servers saying "Lets port vorestation mobs"

PR #1345, opened 2017-04-05
Updates permits

Arokha:
Two permit updates


SpadesNeil:
That reminds me; Tasald's permit has expired. Can you remove the gun from his custom items for now?

SpadesNeil:
In fact a lot of these expired just last month.

PR #1356, opened 2017-04-06
Adds Sleevecards

Arokha:
Adds sleevecards. You can sleeve people into what is basically a pAI (though far less broken than those were) for shenanigans, or to ask them if they want to be resleeved, or whatever.

There are two in each resleever, and you can print more in autolathes. You can 'cryo' from one if you so wish, by using a verb that wipes you from it, so that you can leave the round and it will remove your mind record so someone doesn't resleeve you again.

When done with one, reinsert it into the machine. You are backed up every so often while on a card just like while in a body, and if you have someone on a card and reinsert them to the machine, they will stop being backed up and you'll get a notification after a few minutes, so fair warning.


code/modules/resleeving/computers.dm
```diff
@@ -311,36 +311,37 @@
...
+				var/mode = text2num(href_list["sleeve"])
 				var/obj/machinery/transhuman/resleever/sleever = sleevers[1]
 				if (sleevers.len > 1)
 					sleever = input(usr,"Select a resleeving pod to use", "Resleever selection") as anything in sleevers
 
-				//No body to sleeve into.
-				if(!sleever.occupant)
-					temp = "Error: Resleeving pod is not occupied."
+				switch(mode)
+					if(1) //Body resleeving
```
Leshana: Lets use some constant/defines for these values \o/

code/modules/resleeving/infomorph.dm
```diff
@@ -0,0 +1,616 @@
...
+		"Exclamation Points" = "pai-exclamation",
+		"Question Mark" =      "pai-question"
+	)
+
+/mob/living/silicon/infomorph
+	name = "sleevecard" //Has the same name as the card for consistency, but this is the MOB in the card.
+	icon = 'icons/mob/pai.dmi'
+	icon_state = "repairbot"
+
+	emote_type = 2		// pAIs emotes are heard, not seen, so they can be seen through a container (eg. person)
```
Leshana: Please use constants here

Razer69:
And the PR already made me think I could hide playing cards up my sleeves. :(

PR #1373, opened 2017-04-10
Adds 'Subsleeving' into prey

Arokha:
Allows the resleever to sleeve minds into people inside the resleeving pod's occupant, aka if you want you can print a new body for the person (or if you're wiping some existing person, take a living person), have someone eat them, then get in resleeving pod and revive them in a gut. That'll keep them from walking around and tripping too much, yuh.

I'm sure there's a million crazy things you could do with this (eat person you want to impersonate, resleeve yourself into them while they're in your guts, then escape your own guts?) so have fun with that.

For technical reasons, for now, you cannot sleeve the PRED if they have prey in them in the resleeving pod. It has to do with wiping out their bellies and forgetting where their prey are. This would happen to prey if the prey has their own prey, as well, but... don't? Also you can always just OOC escape if you get trapped like that.

### DIFF
--- a/code/modules/client/preference_setup/vore/04_resleeving.dm
+++ b/code/modules/client/preference_setup/vore/04_resleeving.dm
@@ -29,6 +29,7 @@
 				BR.init_from_mob(character, pref.resleeve_scan, pref.resleeve_lock)
 			if(pref.resleeve_lock)
 				character.resleeve_lock = character.ckey
+			character.original_player = character.ckey
 
 /datum/category_item/player_setup_item/vore/resleeve/content(var/mob/user)
 	. += "<br>"
--- a/code/modules/resleeving/computers.dm
+++ b/code/modules/resleeving/computers.dm
@@ -323,6 +323,7 @@
 				temp = "Error: No sleevers detected."
 			else
 				var/mode = text2num(href_list["sleeve"])
+				var/override
 				var/obj/machinery/transhuman/resleever/sleever = sleevers[1]
 				if (sleevers.len > 1)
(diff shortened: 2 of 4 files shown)


ArbyTheDragon:
As a player who does medical a lot, this confuses the living hell out of me.

... What?..
What is this even.

Arokha:
> So you're saying, you can eat a brainless clone and the resleeving pod can resleeve that clone while they're still inside you?

Yus. And how does it work? Who knows. Sort of a rule-of-fun thing on this one. Several people wanted it, myself included, so I thought I'd code it. I can make up pseudoscience about it, though tbqf I work in healthcare, and both nurses and doctors often have no idea how machines they use work, only how to use them to treat patients effectively.

ArbyTheDragon:
I totally agree, though as a sci-fi geek, sometimes I like to know how things work, but usually it's to the extent that it actually helps me play the game, such as when sleeving used to require the physical removal of a lace from the brain. Understanding the 'science' of it helped me conduct myself in the game. That's just the way of thinking I have.

Following a list of things (Surgery for instance) isn't how I can remember a process. Remembering what happens after you scalpel someone does.

**Also- that's what it can do? That's actually kinda awesome. It's gonna make for some awkward moments in medical though, and some new philosophy of morality in death/resleeving/cloning**
I'd still like someone to hurl a theory out there for its functions. By no means does it have to be you, Arokha.

Screemonster:
inb4 a new type of mediborg with a growpod module, "biomass reclamation tank" and resleever. For...uh. Reasons.

PR #1384, opened 2017-04-11
Allows Akula to breathe underwater

Arokha:
SHERKS. Sherk sherk. Sherk in a sherk splash. Sherk splash in a sherk pool.

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/station_vr.dm
+++ b/code/modules/mob/living/carbon/human/species/station/station_vr.dm
@@ -84,6 +84,9 @@
 
 	reagent_tag = IS_AKULA
 
+/datum/species/akula/can_breathe_water()
+	return TRUE // Surprise, SHERKS.
+
 /datum/species/nevrean
 	name = "Nevrean"
 	name_plural = "Nevreans"


DemicusMaximus:
Fox is fast. Problem only was noticed about an hour ago.

SpadesNeil:
The pool now drowns people? o.o

ArbyTheDragon:
I just noticed we had a deep pool! Holy cow yeah we need this.

Arokha:
I can add this for Unathi if someone important affirms to me that they should also be able to breathe underwater (or hold their breath really long like a turtle or something)

Belsima:
They shouldn't be able, I'm pretty sure- or else it'd be in the 'Unathi biology' PR for Polaris?

ArbyTheDragon:
I'd just leave it be as an oversight. Cause then we'll be saying "Why can't they just hold their breath in a spacesuit?"

SpadesNeil:
OH OH OH I need to make my water world Z-map level when I get home. The trouble is I need to find out how to do some kind of flooding mechanic.

ArbyTheDragon:
Should be like atmospherics more or less. Every tile will have a 'water' level that can spread like gasses do. Problem is, you gotta evacuate all the air on a tile when you become filled, probably to a neighboring tile. If you can't do that, then the water doesn't increase because that situation would occur with a closed up room.
Everything in SS13 is more or less air-tight, so it should also be water-tight. 

Actually- Here's a better idea! **This idea should work so long as we're on a space-station.** A room can only be full to the ceiling if there's no air, since there will always be air at the ceiling, and there are no higher ceilings on a 2D map.
So, a simpler model would be rooms can only fill if their volume is under a threshold (For lag prevention?) and to check if any doors are open that let the water out. Then, spread a 'water' tile across the whole room, and have the water level go up to a code-defined maximum. This spreading water should necessarily mean that when the gravity goes off, the pool area becomes a floating puddle of DOOM that spreads across the room! 


**For an underwater base where air can escape and water can replace it, treat water whenever there is an absense of air- basically, water becomes the new vacuum.** When a door is open to the ocean, air pressure will decrease untill the air pressure is at or below water pressure, then the water will start to flow in, filling up the room like a gas would. Assuming there can't be any vacuum chambers by means of pumps or etc, the partial pressures of gasses should sum up to at least 101.32kPa. If there's more air than this, then there's just more air and no water! If the pressures of gasses are only 50kPa because we leaked it all out, then 51.32kPa of water should enter the area over an arbitrary amount of time that the coders can figure out. 
Here's how this works from a mechanical perspective. Because we maintain the pressure of 101.32kPa, we won't take pressure damage. As the air pressure of the room decreases and fills with water- we actually won't take oxy loss untill the room is perhaps 75% filled. Here's why. The partial pressure of O2 in the station is about 20kPa (or 20%). It will decrease linearly as the room fills up. We won't take oxygen loss untill the pressure is about 5kPa of oxygen. The room will be about 75% filled at this point. It should still make sense as far as how the atmos behaves then!
Again, I'm assuming we can't create a vacuum chamber without water OR air here, because in a vacuum, none of these rules would apply.

OrbisAnima:
I really really really advice against using the atmos system along with it,making an independent system for it instead...

ArbyTheDragon:
**Minecraft water.**

SpadesNeil:
Flooding should NOT be like atmos. God no.

PR #1393, opened 2017-04-15
Transformation Changes

Cameron653:
Adds more transformation types for specific TF's, clarity fix, and bugfix.

Changes "Species Transformation" to "Replica Transformation" for clarity, as people were using it and expecting them to change species, not be a perfect 1:1 replica of the person doing the TF.

Adds "Change_species_and_taur" mode, which does exactly as it says without making the person appear as a 1:1 replica as the pred TFing them (Due to them having the same color)

Adds "Change_hair_and_eyes" tf mode, which allows the changing of the prey's hair and eye color without having to completely TF them or change their gender.

Fixes a bug that infinitely drained the pred's nutrition when they used a TF mode. Slightly increases amount of nutrition the prey gets from the pred.

Fixes a bug causing the gender change to not change their identifying_gender which resulted in them not appearing male/female on description.

Massive cleanup. TF now takes approximately half the space it took before. 

All suggestions Arokha made (With the exception of the blood fix) was made. Massive credit for Arokha helping me get it working and writing the code for the actual egg TF part.


SpadesNeil:
Can't see any issues myself.

Cameron653:
DNM real quick adding another bugfix

code/modules/vore/eating/bellymodes_vr.dm
```diff
@@ -744,6 +769,214 @@
...
+///////////////////////////// DM_TRANSFORM_REPLICA_EGG /////////////////////////////
+	if(digest_mode == DM_TRANSFORM_REPLICA_EGG && ishuman(owner))
+		for (var/mob/living/carbon/human/P in internal_contents)
+			if(P.stat)
+				continue
+			if(P.absorbed) //If they're absorbed, don't egg them
+				return
+			var/mob/living/carbon/human/O = owner
+
+			if (O.custom_species)
```
Arokha: I don't think `O.custom_species` will do what you want it to do here. Many people don't have that set. That's the variable for displaying different text on your examine message. That means anyone who didn't set it can't do TFs.

code/modules/vore/eating/bellymodes_vr.dm
```diff
@@ -744,6 +769,214 @@
...
+				P.fixblood()
+				P.update_hair()
+				P.update_body()
+				P.update_tail_showing()
+				P.update_eyes()
+				P.updateicon()
+				switch(O.egg_type)
+					if("Unathi")
+						var/obj/structure/closet/secure_closet/egg/unathi/J = new /obj/structure/closet/secure_closet/egg/unathi(O.loc)
+						P.forceMove(J)
```
Arokha: I'm usually opposed to telling people to fix things that were existing before they messed with the files in the PR, however you're making it worse with these, so please change these to not use the same duplicated code 8+ times in the same file. Either make egg types a list of "Species Name" = /path or something else.

code/modules/vore/eating/bellymodes_vr.dm
```diff
@@ -744,6 +769,214 @@
...
+				P.custom_species 	= O.custom_species
+				P.r_eyes 			= O.r_eyes
+				P.g_eyes 			= O.g_eyes
+				P.b_eyes 			= O.b_eyes
+				for(var/obj/item/organ/I in P.internal_organs) //This prevents organ rejection
+					I.species = O.species
+				for(var/obj/item/organ/external/chest/A in O.organs)
+					for(var/obj/item/organ/external/Z in P.organs) //This makes their limb sprites look correct.
+						Z.species = O.species
+						Z.h_col = A.h_col
```
Arokha: Bad OOP. For each `organ` do `organ.species = O.species` and if it's external, `sync_color_to_human(them)` to do h_col and s_col. Just do it in a single for() loop.

code/modules/vore/eating/bellymodes_vr.dm
```diff
@@ -744,6 +769,214 @@
...
+						Z.s_col = A.s_col
+				P << "<span class='notice'>You lose sensation of your body, feeling only the warmth around you as are you as you're encased in an egg. </span>"
+				owner << "<span class='notice'>You shift as you make dramatic changes to your captive's body as you encase them in an egg.</span>"
+				P.fixblood()
```
Arokha: This gives you all your blood back...? So eating someone, doing TF for 1 tick to something if they have 10% blood, and then defibbing works fine? :P Seems a little odd.

Cameron653:
Alright, I'm going to see about changing/fixing all the requested things up. Should have a fixed version in a bit.

code/modules/vore/eating/bellymodes_vr.dm
Cameron653: They can perform TF's, as the "else" below on line 888 makes it so does the same thing, just without the custom_species name. That being said, looking back at this, it's _really_ cluttered and can be executed a lot better. Should have all these changed in a bit.

Cameron653:
Changes made. Everything with the exception of the blood change was made, as people can't be TF'd while unconscious/SSD/dead due to the !stat check.

Arokha:
My only concern is this craaaazy indent.
image

Cameron653:
I'll fix that in a bit, think that might've accidentally got left in from
me copy and pasting the list when it was elsewhere.

PR #1416, opened 2017-04-20
Un-scale HUDs for micros and macros.

Leshana:
* Set appearance_flags on HUD images to prevent them from being scaled up/down when the owning mob is macro/micro.
* Optimized the resize() proc and made it adjust the vertical position of the unscaled hud to stay even with the top of the mob's head.
* Also fix the resize proc to be consistent with update_icons().   Now micros wont be in the middle of a turf temporarily after resizing.

### DIFF
--- a/code/defines/procs/hud.dm
+++ b/code/defines/procs/hud.dm
@@ -4,6 +4,7 @@ the HUD updates properly! */
 
 // hud overlay image type, used for clearing client.images precisely
 /image/hud_overlay
+	appearance_flags = APPEARANCE_UI // Don't get scaled with macro/micros.  VOREStation edit
 
 //Medical HUD outputs. Called by the Life() proc of the mob using it, usually.
 proc/process_med_hud(var/mob/M, var/local_scanner, var/mob/Alt)
--- a/code/game/dna/dna2_helpers.dm
+++ b/code/game/dna/dna2_helpers.dm
@@ -192,7 +192,7 @@
 		// Playerscale
 		var/size = dna.GetUIValueRange(DNA_UI_PLAYERSCALE, player_sizes_list.len)
 		if((0 < size) && (size <= player_sizes_list.len))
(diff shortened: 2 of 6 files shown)


Natje0:
I often use the Resize verb in order to center myself as a small/micro... Because it looks absolutely dumb when you stand on the end of a table as a small/micro, and it generally feel more fitting to be centered sometimes.

Would it be possible to instead of removing the possibility of being centered, make it toggleable to make you *STAY* centered even when picking things up and interacting with things?
 I'm sure a lot of small/micro players would like that, considering the sheer amount of times I see people centering themselves.

Arokha:
`Also fix the resize proc to be consistent with update_icons(). Now micros wont be in the middle of a turf temporarily after resizing.` Let's just always center them instead. I think it has more cases where it looks better this way. The "impossible to see at nexus of tiles" problem will be lessened on Tether anyway, because it has more narrow tile division.

SpadesNeil:
I want them always centered as well. The decision to make them always align to the bottom was the idea of a former dev who was very stubborn about not changing it, but now he's gone so I don't care

Razer69:
Having them stand on the lower edge gives a clearer picture of the size difference, as they appear to stand on the same height as other characters' boots. I like it that way.

Standing on tables looks dumb for characters of any size, and there should be a mechanic to lift the sprites up a few pixels when climbing something.

SpadesNeil:
I suppose that is true. Sitting in chairs also looks dumb.

Leshana:
I think this is called "scope creep" This PR is for making the HUDs non-tiny, not making a new table climbing system.  The change to resize is a bugfix, not a new feature.  If we do want it to always be centered, I am fine with that, but its a design change.

SpadesNeil:
Focus on the HUD thing for now. We just got carried away off topic by micro ramblings.

PR #1417, opened 2017-04-20
Fixes some character setup issues

Yoshax:
- Clicking Cancel in the tail selection or ear selection option will no longer nullify your already selected preference.
- The name of your selected ears/tail will show up instead of 'Custom'

Thanks to @Arokha for the making the name show up stuffs.


code/modules/client/preference_setup/vore/01_ears.dm
```diff
@@ -100,7 +113,10 @@
 		if(selection && selection != "Normal")
 			pref.ear_style = pretty_ear_styles[selection]
 		else
-			pref.ear_style = null
+			if(pref.ear_style)
+				return TOPIC_REFRESH_UPDATE_PREVIEW
```
Arokha: I have an alternate suggestion for this part.
`var/list/pretty_ear_styles = list("Normal")` should be `var/list/pretty_ear_styles = list("Normal" = null)` instead, so that if they pick normal, it just sets it to null, and you can drop the entire 'else' statement entirely, as well as the `selection != "Normal"`

Yoshax:
Changed that stuff.

Issue #1434, opened 2017-04-26
Unathi TF


SilverTalismen:
#### Brief description of the issue
Currently, any species thta is transformed into a Unathi suffers massive internal issues =due to having appendix's and kidneys being present, due to the recent changes to unathi biology they begin to suffer large amounts of toxins and possibly even organ rejection entirely

#### What you expected to happen
To possibly have this fixed by the TF removing those organs

#### What actually happened
https://gyazo.com/84db344d44038caba7efe286ad86aaa5
Lots of organ failure
#### Steps to reproduce
- (Step 1)Be any race that isn't unathi
- (Step 2)get turned into a unathi
- (and so on)Suffer

#### Code Revision
release - 2017-04-26
 
#### Anything else you may wish to add:
- (Location if it's a mapping issue, screenshots, sprites, etc.)

Wickedtemp:
This can become quite literally a round ruiner for some. One patient ended up with 999u toxin as a result of organ rejection. We had to just resleeve them because there's no way we could've healed them in a timely manner.

ArbyTheDragon:
999 Toxin? At least when you die, damage stops incurring itself. Once you're at -100%HP, you're dead. How can you possibly get THAT much?

Carthatoline is a **very robust** anti-toxin. Think that would work if we catch them before they're actually dead?
Carthatoline seems to remove 8 toxin per 'tick' _(Whatever the timescale may be for incurring the adjusttoxloss damage proc)_ according to the code, whereas dylovene only removes 4 toxin.
We could combine the two for -12 tox per 'tick', for something 3X as strong as dylovene.

Whether death is preventable or not, this is still a nasty bug that needs to be fixed, I was just theorizing for fun.

Cameron653:
Will look into this in and see if I can whip up a fix real quick

Cameron653:
Fix found. Will put up the PR in a little bit along with a small feature.

Wickedtemp:
Ziah, it isn't 999 toxin damage.

It's 999 units of toxin, the substance, in the bloodstream.

But yay, fix found.

Also, it might be a good idea to check whether or not blood types are affected. If they are, someone could potentially experience poisoning due to being changed to an incompatible type of blood.

JoanRisu:
This issue seems to have spread to Taj TFing to humans.

Wickedtemp:
Alright, lovely.

Time to check TF from every species to every other species, and TF from human to human with different bloodtype.

[...]

PR #1439, opened 2017-04-27
Fixes Unathi TF organ rejection

Cameron653:
#1434 
Fixes people from dying from Unathi TF. 
Deletes their old organs and replaces it with the correct organs their species should have. Additionally, this prevents this problem from ever coming up again in case another species is changed like the unathi was.
Fixes organ rejection due to #1393. As it turns out, internal organs aren't listed under organs, but rather internal_organs.
Before/after TF with this code below (Human -> Unathi TF):
https://gyazo.com/b05b8fae807278209b495ba81faab618
https://gyazo.com/2f91bc1ad1a6be660b40fd4ee2c7fd09

### DIFF
--- a/code/modules/vore/eating/bellymodes_vr.dm
+++ b/code/modules/vore/eating/bellymodes_vr.dm
@@ -336,6 +336,8 @@
 						Z.sync_colour_to_human(P)
 					P << "<span class='notice'>You lose sensation of your body, feeling only the warmth of everything around you... </span>"
 					owner << "<span class='notice'>Your body shifts as you make dramatic changes to your captive's body.</span>"
+					if(P.species == Unathi)
+						P.species.create_organs(P) //This is the only way to make it so Unathi TF doesn't result in people dying from organ rejection.
 					P.fixblood()
 					P.update_body()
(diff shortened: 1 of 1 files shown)


Arokha:
Did you test this

Cameron653:
Yes.

JoanRisu:
Before you merge this, can you check if Tajs are having the same problem. Someone recently tried to TF a taj to a human and got the same problems.

Cameron653:
That shouldn't occur with tajaran, as they have the same exact organs as humans do
https://github.com/VOREStation/VOREStation/blob/master/code/modules/mob/living/carbon/human/species/station/station.dm#L169-L176

As stated in the PR, this will change the person that's being TF'd into the correct organs, no matter the race.

Cameron653:
Removing DNM tag. After doing some TF with this version of the code and watching all the organs germ_level, I can't see any problems. B_type is the same for the person & organs, organs are the correct species, etc etc. Removing the DNM tag.

Cameron653:
Alright, it's fixed for real this time. Should be ready for merge now.

PR #1445, opened 2017-04-28
Syncs some missed changes from Polaris

Yoshax:
Some of the Unathi changes were missed, so this adds them.
Also fixes the changelog because the one on live is borked:
changelog


Poojawa:
>blast doors no longer being invulnerable

Wew lad. now they'll be melted by the engine too.

code/modules/mob/living/carbon/human/species/station/station.dm
```diff
@@ -32,23 +32,23 @@
...
 	metabolic_rate = 0.85
 	item_slowdown_halved = 1
 	num_alternate_languages = 3
 	secondary_langs = list(LANGUAGE_UNATHI)
 	name_language = LANGUAGE_UNATHI
 	health_hud_intensity = 2.5
 
-	min_age = 18
-	max_age = 60
+	min_age = 32
```
SpadesNeil: I'd like to reduce this back to 18.

PR #1460, opened 2017-05-01
Adds Adranol

Arokha:
Adranol fixes jittery, eye_blurry, and confusion, and would be good as a post-resleeving drug. It'd also be the only drug that can actually fix jitteryness, since there is no way to undo that now.

### DIFF
--- a/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Medicine_vr.dm
+++ b/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Medicine_vr.dm
@@ -1,0 +1,18 @@
+/datum/reagent/adranol
+	name = "Adranol"
+	id = "adranol"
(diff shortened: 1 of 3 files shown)


OrbisAnima:
are you sure you want to use milk for it? it makes it kinda of a pain to go to the bar and beg for it...

Arokha:
There is already milk in fitness vending machines, which already exist in medical

OrbisAnima:
oh!

Wickedtemp:
Kitchen has milk, fitness vendors have milk, the bar has milk, there are plenty of sources of milk.

Arokha:
> plenty of sources of milk.

c:<

Wickedtemp:
I had nothing to do with this, I swear.

Wickedtemp:
... What if an overdose of this made you drowsy?

PR #1461, opened 2017-05-02
Add's Tempest's custom implant item

### LINKED ISSUE #1450: (Fluff Item)Special Implant! - Chakat Tempest - Wickedtemp
Name of Item: Natural Mammary Glands!
     
Item description: OOCly, an implant with an internal storage that produces milk by using nutrition which you can later retrieve via container... which allows for ICly, being 'productive' and milked. 

Sprites (or link to sprites): Default implant sprite. My idea is using the implant as an OOC means to make Tempest be naturally 'milkable' ICly so the appearance doesn't matter. 

Any additional resources: The code is a WIP!

BYOND Key (Username): Wickedtemp

Name of Character: Chakat Tempest

Main Job(s) of Character: Doctor, Chemist, CMO and Resident

Why your character should have this item: Tempest is a hermaphrodite with two very large female attributes on hir chest. 

Additional Notes: http://forum.vore-station.net/viewtopic.php?f=20&t=433 This was suggested as a blanket add-in for all characters in the past and was denied because, well, let's face it. When a new feature is added, people immediately start spamming it. It happened with driders and webs, it happened with mlems and awoos, it happened with flipping. The same would happen if everyone was given a lactating mechanic. It'd get spammed, people who didn't want to see it would be forced to see it.

But, this would only be done between people who want to do it. We've approved kink 'genderswap' items in the past, another potentially pref-breaking item that has, so far, been used purely to enhance the roleplaying of those who have it. You can only GET milk out of the 'implant' by using a container on the host, so it isn't something that's likely to happen completely by accident, and I'd like to think that, if nothing else, in the time I've played on the server I've shown that I'm not the type of person that'd intentionally prefbreak. 

I've talked about it with a couple of Staff and they thought this was a neat concept, so here's the official app. Questions, comments, concerns and fun-facts are all welcome.

Yoshax:
Resolves #1450 

~~Still semi-WIP for finishing touches and looking over when not 3am.~~ Up for review though.
~~Just waiting for Tempest to tell one one last thing or two, and will be 100% done.~~

### DIFF
--- a/code/game/objects/items/weapons/implants/implantreagent_vr.dm
+++ b/code/game/objects/items/weapons/implants/implantreagent_vr.dm
@@ -1,0 +1,111 @@
+/obj/item/weapon/implant/reagent_generator
+	name = "reagent generator implant"
+	desc = "This is an implant that has attached storage and generates a reagent."
+	implant_color = "r"
+	var/obj/item/weapon/reagent_containers/glass/beaker/large/internal_storage
+	var/datum/reagent/generated_reagent = null
+	var/gen_amount = 2 //amount of reagent generated per process tick
+	var/gen_cost = 0.5 //amount of nutrient taken from the host per process tick
+	var/transfer_amount = 30 //amount transferred when using verb
+	var/usable_volume = 120
+
+	var/list/empty_message = list("You feel as though your internal reagent implant is almost empty.")
+	var/list/full_message = "You feel as though your internal reagent implant is full."
+	var/list/emote_descriptor = list("tranfers something") //In format of [x] [emote_descriptor] into [container]
+	var/list/self_emote_descriptor = list("transfer") //In format of You [self_emote_descriptor] some [generated_reagent] into [container]
+	var/list/random_emote = list() //An emote the person with the implant may be forced to perform after a prob check, such as [X] meows.
+	var/assigned_proc = /mob/living/carbon/human/proc/use_reagent_implant
+
+/obj/item/weapon/implant/reagent_generator/New()
+	..()
+	internal_storage = new /obj/item/weapon/reagent_containers/glass/beaker/large (src)
+	internal_storage.volume = usable_volume
+	internal_storage.reagents.maximum_volume = usable_volume
+
+/obj/item/weapon/implanter/reagent_generator
+	var/implant_type = /obj/item/weapon/implant/reagent_generator
+
+/obj/item/weapon/implanter/reagent_generator/New()
+	..()
+	imp = new implant_type(src)
+	update()
+	return
+
+/obj/item/weapon/implant/reagent_generator/implanted(mob/living/carbon/source)
+	processing_objects += src
+	to_chat(source, "<span class='notice'>You implant [source] with \the [src].</span>")
+	source.verbs |= assigned_proc
+	return 1
+
+/obj/item/weapon/implant/reagent_generator/process()
+	var/before_gen
+	if(imp_in && generated_reagent)
+		before_gen = internal_storage.reagents.total_volume
+		if(internal_storage.reagents.total_volume < internal_storage.volume)
+			if(imp_in.nutrition >= gen_cost)
+				do_generation()
+		else
+			return
+	else
+		imp_in.verbs -= assigned_proc
+		return
+
+	if(internal_storage)
+		if(internal_storage.reagents.total_volume == internal_storage.volume * 0.05)
+			to_chat(imp_in, "<span class='notice'>[pick(empty_message)]</span>")
+		else if(internal_storage.reagents.total_volume == internal_storage.volume && before_gen < internal_storage.volume)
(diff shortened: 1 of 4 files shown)


code/game/objects/items/weapons/implants/vorestuff_vr.dm
```diff
@@ -0,0 +1,131 @@
...
+	set desc = "Remove reagents from am internal reagent into a container."
+	set category = "Object"
+	set src in view(1)
+
+	do_reagent_implant()
+
+/mob/living/carbon/human/proc/use_reagent_implant_tempest()
+	set name = "Milk"
+	set desc = "Grab Tempest's nipples and milk them into a container! May cause blushing and groaning."
+	set category = "Object"
```
Arokha: Don't forget this when moving.

code/game/objects/items/weapons/implants/vorestuff_vr.dm
```diff
@@ -0,0 +1,131 @@
...
+	to_chat(source, "<span class='notice'>You implant [source] with \the [src].</span>")
+	source.verbs |= assigned_proc
+	return 1
+
+/obj/item/weapon/implant/reagent_generator/process()
+	var/before_gen
+	if(imp_in && generated_reagent)
+		before_gen = internal_storage.reagents.total_volume
+		if(internal_storage && internal_storage.reagents.total_volume < internal_storage.volume)
+			if(imp_in.nutrition >= gen_cost)
```
Arokha: Don't check for internal_storage if there's no way to remove it. Just let it runtime anyway, it's not like it'll function correctly if someone somehow removes it. Roll the if() below up into the one above.

code/game/objects/items/weapons/implants/vorestuff_vr.dm
```diff
@@ -0,0 +1,131 @@
...
+	set src in view(1)
+
+	do_reagent_implant()
+
+/mob/living/carbon/human/proc/do_reagent_implant()
+	if(!isliving(usr) || !usr.canClick())
+		return
+
+	if(usr.stat == 1 || (usr.restrained()))
+		return
```
Arokha: This stuff can use incapacitated() and pass it one of the several incap flags instead.

Yoshax:
Changes made, pretty much. Should be good now. Maybe?

code/game/objects/items/weapons/implants/implantreagent_vr.dm
```diff
@@ -0,0 +1,108 @@
...
+	set category = "Object"
+	set src in view(1)
+
+	do_reagent_implant()
+
+/mob/living/carbon/human/proc/do_reagent_implant()
+	if(!isliving(usr) || !usr.canClick())
+		return
+
+	if(usr.stat == 1 || (usr.restrained()))
```
Leshana: stat 1 is unconscious, stat 2 is dead, so we should probably do either `if(stat > CONSCIOUS ...)` or simply `if(stat ...)`

code/game/objects/items/weapons/implants/implantreagent_vr.dm
```diff
@@ -0,0 +1,108 @@
...
+
+/mob/living/carbon/human/proc/use_reagent_implant()
+	set name = "Transfer From Reagent Implant"
+	set desc = "Remove reagents from am internal reagent into a container."
+	set category = "Object"
+	set src in view(1)
+
+	do_reagent_implant()
+
+/mob/living/carbon/human/proc/do_reagent_implant()
```
Leshana: This is not critical, but recommend byond pattern is to avoid use of `usr` and instead have this proc take a parameter, like `mob/living/carbon/human/proc/do_reagent_implant(var/mob/user)`  This also will make lines 75-85 unnecessary.

code/game/objects/items/weapons/implants/implantreagent_vr.dm
```diff
@@ -0,0 +1,108 @@
...
+			break
+	if(rimplant)
+		if(container.reagents.total_volume < container.volume)
+			var/container_name = container.name
+			if(rimplant.reagents.trans_to(container, amount = rimplant.transfer_amount))
+				user.visible_message("<span class='notice'>[usr] [pick(rimplant.emote_descriptor)] into \the [container_name].</span>",
+									"<span class='notice'>You [pick(rimplant.self_emote_descriptor)] some [rimplant.generated_reagent] into \the [container_name].</span>")
+				if(prob(5))
+					src.visible_message("<span class='notice'>[src] [pick(rimplant.random_emote)].</span>") // M-mlem.
+			if(rimplant.reagents.total_volume == rimplant.reagents.maximum_volume * 0.05)
```
Leshana: I recommend using `<=` here in case future changes make it not land exactly on 5%

Yoshax:
@Arokha @Leshana changes made, discussed with Leshana on Discord. The 5% thing has stayed as is because it shouldn't ever not land on 5%. If it does, it's because someone has made another implant with weird numbers.

PR #1491, opened 2017-05-20
Fixes an issue where a non-existent language was existing + some normalisation

Yoshax:
Siik'tajr had been changed to just Siik forever ago but never changed for some stuff, so some species were able to pick a language that didn't really exist.

Also gives most of the Vore-specific races the choice at 3 languages because it's all the other normal species have.

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/station_vr.dm
+++ b/code/modules/mob/living/carbon/human/species/station/station_vr.dm
@@ -11,7 +11,7 @@
 	//brute_mod = 1.15
 	//burn_mod =  1.15
 	//gluttonous = 1
-	num_alternate_languages = 2
+	num_alternate_languages = 3
 	secondary_langs = list("Sagaru")
 	name_language = "Sagaru"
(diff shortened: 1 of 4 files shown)


code/modules/mob/living/carbon/human/species/station/station_vr.dm
```diff
@@ -139,8 +139,8 @@
 	icobase_tail = 1
 	unarmed_types = list(/datum/unarmed_attack/stomp, /datum/unarmed_attack/kick, /datum/unarmed_attack/claws, /datum/unarmed_attack/bite/sharp)
 	num_alternate_languages = 3
-	secondary_langs = list("Siik")
-	name_language = "Siik"
+	secondary_langs = list(LANGUAGE_SIIK)
```
SpadesNeil: We do need to give the Zorren their own language, but what does a fox _say_?

Arokha: 🦊 🔫

PR #1492, opened 2017-05-20
Belle Reagent Generator Child

Pancakesss:
To accompany my other pull request which I will now edit.


Arokha:
This should be part of the other PR. Just make this commit on the same branch as your other PR and it'll be automagically added to the PR without you doing anything.

Arokha:
I can explain more after the Tether work is done which should be later today, if you need help.

PR #1505, opened 2017-05-21
Updates map merge tools to work for Tether, maybe, I hope?

Yoshax:


Arokha:
Dunno if this'll really work or not, but we'll see

PR #1510, opened 2017-05-21
Fixes an issue with a custom item

Yoshax:
Fixes that taur person's suit that got mentioned.


Yoshax:
We're at a critical butt limit. Anymore and we'll have butt overflow.

Arokha:
Fine I guessss

PR #1515, opened 2017-05-21
NOTHING TO SEE HERE MOVE ALONG

Dhaeleena:
Changes to legit non syndicate and totally not hidden and legal base, Adds med-bay for the law abiding members of society


SpadesNeil:
Please use the map merge tool correctly.

Dhaeleena:
@SpadesNeil i was told the map merge tool doesn't work correctly atm

SpadesNeil:
Who said that? Since when?

Yoshax:
It should work. I changed the map merge to be directed at Tether. There shouldn't really be any difference,  I don't think. But I haven't really fully tested it. Looked fine on a small test though.

PR #1518, opened 2017-05-22
Non-fuzzy macro scaling attempt

Arokha:
Let's see how it goes!


SilverTalismen:
What's it look like? :o

Arokha:
2x macros look just like a bigger version of themselves, nothing special.
1.5x macros look a tiny bit odd, because it has to lose some pixels, but some people say it looks better, some people say it's weird, but the general feeling I got was that people still preferred it over fuzzy.

SpadesNeil:
oh god it's hideous

PR #1526, opened 2017-05-22
Disable gateway until maps can be resized

Arokha:
They need to be the same size as the rest of the maps, which might mean splitting them into multiz maps or just rearranging.

### DIFF
--- a/maps/RandomZLevels/fileList.txt
+++ b/maps/RandomZLevels/fileList.txt
@@ -7,8 +7,8 @@
 
 #maps/RandomZLevels/labyrinth.dmm #Broken still.
 #maps/RandomZLevels/snowfield.dmm #Broken.
-maps/RandomZLevels/carpfarm.dmm
-maps/RandomZLevels/listeningpost.dmm
-maps/RandomZLevels/blackmarketpackers.dmm
-maps/RandomZLevels/beach.dmm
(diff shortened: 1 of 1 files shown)


SpadesNeil:
What is the issue with the gateway map sizes? Gateway maps are too big or too small?

Arokha:
Too big. Basically byond cannot handle different map sizes, it's just not a feature. So it loads all our 120x120 new maps, then loads a 300x300 gateway map or whatever, and suddenly it rescales all our maps to be 300x300 towards the northeast, which means that the map edges and such are all suddenly connected to space and such on the ground level which lags the server because atmos connected to space etc etc.

Arokha:
Basically we need to resize them or turn them into multi-z things with stairs and ladders or whatever or other things. Or just make new ones. I started a couple, Yosh started a couple, we should actually finish them. XD

Yoshax:
Starting things is easy. Finishing them is hard. I have a few in mind that are less grand ideas I may try to implement. A few smallish, simple ones would be a fairly good start, I believe.

Arokha:
120x120 maps might be less intimidating for people to fill up, anyway. If not, they can make them have stuff like the map-edge teleporters like what lead to solars and mining.

SpadesNeil:
I really want to see the zoo map done over, then.

PR #1528, opened 2017-05-22
Add some 'single' random spawns to the maps

Arokha:
A single netgun will always spawn somewhere on the map, but the location is random.
Either a bluespace harpoon or a translocator will spawn somewhere on the map, but never both, and location is also random.

Keep in mind they have a chance of spawning in maint where they can be easily grabbed... or spawned into one of those closed-off walled-in single rooms that are occasionally in the walls with stuff in them, so you might not see them every shift.


SpadesNeil:
I really like the fact that we can limit things like that. This opens up options for future sneakythings.

PR #1533, opened 2017-05-23
Several Tether Fixes + Medical Rearrange

Arokha:
- Rearranges medical to some extent
- Added a pill vendor near each surface airlock with antitox and painkiller pills (remember to remove contaminated clothes (including boots, which may not show the symbol) before taking the pills otherwise it's kinda pointless). You should only need one pill.
- Fix solitary confinement cryopod computer location
- Added missing fire alarms in asteroid
- Add missing firelock to library door
- Made it so you can reach over med records laptops
- Dhae: Rebalanced maintenance spawns


Arokha:
This actually is mapmerge'd, but so much was changed, it's still huge.

PR #1536, opened 2017-05-23
Cryo consoles require head access

Arokha:
To prevent people from just rummaging around through everyone's stuff and taking it.


Dhaeleena:
Pls remember

PR #1550, opened 2017-05-25
Adds wah's custom item

### LINKED ISSUE #1548: Custom ID - Amaya Rahl - KiraAlitruss
I was told to make an issue about my custom ID. It was approved by Dhaeleena. The thread is here: http://forum.vore-station.net/viewtopic.php?f=26&t=945
All needed information is located within the thread. First time making something like this. I apologise for any mistakes, blunders or other such errors.

#### Code Revision
- Server revision: release - 
5009c1629a79cc303755d13980a4181128a40b61

Yoshax:
Fixes #1548

Took a little work. The ID takes its name from a list of two choices, as wanted by the person. Also because it's somewhat different to the only other Custom ID (Aro's) it needed a method to get its access. This is down by booping it with the ID spawned with, which copies the access then deletes it, allowing this to the only ID. Otherwise work's the same as Aro's in that you use it in hand to set the stuff.


config/custom_items.txt
```diff
@@ -323,6 +323,10 @@ item_path: /obj/item/device/modkit_conversion/fluff/harmonyspace
 
 # ######## K CKEYS
 {
+ckay: kiraalitruss
```
KiraAlitruss: hmm.

PR #1553, opened 2017-05-25
Sundry runtimes 2

Leshana:
Fixes Runtime in lattice.dm,41: undefined proc or verb /turf/simulated/floor/airless/update().
Fixes Runtime in living_vr.dm,197: Cannot read null.prefs_vr


Arokha:
Somewhat sundry

PR #1557, opened 2017-05-25
Fixes AMI suit access

Yoshax:


Cameron653:
Would probably be better to do this in 
https://github.com/VOREStation/VOREStation/blob/9580dffd19eeed085235bec86affba978750dcd6/code/modules/clothing/spacesuits/rig/suits/station_vr.dm

where it was originally instead of a normal .dm file

Yoshax:
:upside_down_face:

PR #1561, opened 2017-05-26
Turbolift Firefighter Mode

Leshana:
* Bugfix: Make interior panel dialog box refresh
   * Refresh when we push buttons to see the effect of what we did.
   * Refresh when we remove floors from the queue to see the highlighting update.
* Allow turbolift auto-mapping to build a lift without elevator car walls.  Relying on external walls prevents the need for double-thick walls that take up more space.  Useful for some maps (Like the tether!)
  * Changes Tether map to not build the interior walls, now elevator is effectively 4x4!
* **Adds "firefighter mode" to turbolifts**
   * Adds a priority call mode that allows heads and first responders to cancel all calls and summon the lift to their floor immediately.
   * Adds a firefighting mode that turns the lift over to manual control, doors only open when manually opened, all call buttons ignored.  For use during fires.


code/modules/turbolift/turbolift.dm
```diff
@@ -9,14 +9,73 @@
 	var/floor_wait_delay = 85                           // Time to wait at floor stops.
 	var/obj/structure/lift/panel/control_panel_interior // Lift control panel.
 	var/doors_closing = 0								// Whether doors are in the process of closing
+	var/list/music = list('sound/music/elevator.ogg')	// Elevator music to set on areas
```
Arokha: May want to leave this list empty otherwise this code will need to change if the sound file changes/doesn't exist, since it'll fail to compile.

code/modules/turbolift/turbolift.dm
```diff
@@ -9,14 +9,73 @@
...
-	open_doors()
+	if(!fire_mode)
+		open_doors()
+
+// Enter priority mode, blocking all calls for awhile
+/datum/turbolift/proc/priority_mode(var/time = 30 SECONDS)
+	priority_mode = TRUE
+	cancel_pending_floors()
+	update_ext_panel_icons()
+	control_panel_interior.audible_message("<span class='info'>This turbolift is responding to a priority call.  Please exit the left when it stops and make way.</span>")
```
Arokha: `left`

PR #1579, opened 2017-05-28
Makes it so mice can't eat people.

Cameron653:
It was causing a lot of bugs. 
Adds a no_vore proc which can be tossed onto any /living creature to make it so they don't spawn with a stomach and can't create a stomach for theirselves.

### DIFF
--- a/code/modules/vore/eating/living_vr.dm
+++ b/code/modules/vore/eating/living_vr.dm
@@ -12,14 +12,19 @@
 	var/reviving = 0					// Only used for creatures that have the xenochimera regen ability, so far.
 	var/metabolism = 0.0015
 	var/vore_taste = null				// What the character tastes like
+	var/no_vore = 0 					// If the character/mob can vore.
 
 //
 // Hook for generic creation of stuff on new creatures
 //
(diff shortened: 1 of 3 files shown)


code/modules/vore/eating/living_vr.dm
```diff
@@ -12,14 +12,19 @@
...
 //
 // Hook for generic creation of stuff on new creatures
 //
 /hook/living_new/proc/vore_setup(mob/living/M)
-	M.verbs += /mob/living/proc/insidePanel
 	M.verbs += /mob/living/proc/escapeOOC
 	M.verbs += /mob/living/proc/lick
+	if(no_vore) //If the mob isn's supposed to have a stomach, let's not give it an insidepanel so it can make one for itself, or a stomach.
+		M << "<span class='warning'>The creature that you are can not eat others.</span>"
+		return
```
Arokha: This must return 1 otherwise the hook will consider itself failed. Such is the case with all `/hook`s

Cameron653:
Whoops, let me fix that real quick.

Cameron653:
Done. Should work now.

SilverTalismen:
Very *Mice* work Cammie :)

PR #1586, opened 2017-05-30
Trait Tweaks

Arokha:
`balance intensifies`

Allow 5 traits
Adds blood color as a custom species setting (doesn't apply to synths)
Adds photoresistant trait
Changes haste and slowdown traits to be taj and tesh speeds (higher is now 4 points)


Dhaeleena:
Cattes get 15 okay?

Arokha:
> Cattes get 15 okay?

Cattes are smol. You get 1.5

PR #1601, opened 2017-05-31
Adds Trash Piles

Arokha:
Replaces some lockers and racks in maintenance with trash piles. They are presistent loot piles for people who just can't help but greytide. It allows people who join later in the shift to still get goodies. Each is searchable once per shift by each ckey.

They have three loot lists: alpha, beta, and gamma.

- Alpha has the highest chance and just random maint stuff. All usable items (no random empty beer cans) but nothing you couldn't find elsewhere.
- Beta has a lower chance of spawning and is usually contraband, mechaically useful stuff. Radio jammers and the like.
- Gamma is dangerous or highly illegal contraband, and is unique. Guns (nonlethal!), teleporters, syndicate IDs, etc.

Gamma items have special handling: all of the trash piles share a gamma list, and when an item is given out from the gamma list to someone, it's de-pooled so that another trash pile won't hand it out. The pile-collective maintains a reference to it as an 'allocated' gamma item though. Then, when the gamma list runs out, it reviews allocated gamma items to find any that it might be able to 're'-distribute. Namely ones that have been deleted, or are in cryopod computers because the user left with them. It can then hand that item out again. If it cannot hand out a new gamma item, it gives beta instead.

You can also just manually put gamma-list items back into the piles if you're leaving. Or, if you find a translocator, but already made yourself one in R&D (or later make one), or you're the RD and find a bluespace harpoon, you can stuff the translocator/whatever back into any trash pile and it'll be re-pooled into all of them. You can only return gamma items this way, not beta/alpha. You cannot add 'extra' items either. You may know it hands out translocators, but if it hasn't handed one out this shift, you can't return a translocator to any trash pile.

Additionally, if you find yourself playing a simple animal mob for an event (or you're a mouse), you can hide in trash piles by clicking them. There's a 50% chance you're revealed if the pile is searched. You can climb out by clicking the pile again. I dunno if that'll ever get used but it was easy to code, so eh.


OrbisAnima:
this is crazy insane.

PR #1605, opened 2017-06-01
Adds Test Subject alt-title to assistants.

PontifexMinimus:
Because it would be fun to be used for SCIENCE!!
(Lewd science)


Arokha:
I don't see why not, but you have to add a `//VOREStation Edit` line after that line because it's a stock polaris dm file.

PR #1606, opened 2017-06-01
One more jobslot for librarians.

PontifexMinimus:
Request of Benemuel.


Arokha:
@JoanRisu says this is fine so if this is horrible blame her.

PR #1615, opened 2017-06-04
Smol fixes

Dhaeleena:
Fixes pipes on the recreation area showers. Adds extra mining locker to mining, Fixes telecom access areas and areas naming also fixes one turret that was lethal instead of stun. Syndi-fox is now called Rick, Aronai is butted.


Arokha:
> Aronai is butted

:C

PR #1620, opened 2017-06-04
Fluff Implant for Pear120

Pear120:
http://forum.vore-station.net/viewtopic.php?f=26&t=961

Adds subjugation implant - causes pain in host when host's player triggers implant through Object tab, comes in 2 flavors: moderate and brain damage.


Cameron653:
You may want to make a check that makes sure that triggering it won't _kill_ them from the amount of brain damage they're going to take. Ex: Person has 98 brain damage from various things and hitting it will kill them.

Also, is there a delay for activating it? If not, there should be a 30 second delay or so in between presses so someone doesn't accidentally click it twice at once.

Pear120:
@Cameron653 Well, I'm the only one using it, and the Extreme button would need to be pressed a LOT of times for it to be lethal.

Pear120:
Fixed implant so it won't kill you when your brain is almost pudding.

SpadesNeil:
I don't know why this PR is even here. The application was not approved.

PR #1621, opened 2017-06-04
Soap, Rags, and Buckets

Pear120:
Rags and soap were unable to be used on buckets to wet them, instead they would only clean the bucket.
[Rags can now also clean dirt from floors so maids don't have to use mops](http://www.maids.com/blog/wp-content/uploads/2012/11/cinderella-scrubbing-300x201.png)


Arokha:
https://github.com/VOREStation/VOREStation/blob/master/.github/CONTRIBUTING.md

Pear120:
@Arokha Comments added, will work on slipping it to Polaris.

PR #1622, opened 2017-06-05
Makes rigsuits not get contaminated with plasma.

Cameron653:
This was killing people even if they had it in the AMI/rigsuit and didn't have it equipped, as it was in their contents.

Additionally, rigsuit parts were impossible to clean. 
This makes sure they don't become contaminated with plasma, solving both of those problems


Cameron653:
Changing the method this is done

PR #1547, opened 2017-05-25
Does a real big critical fix or something with hardsuits

Yoshax:
Fixes #1545 maybe idk looks good

PR #1623, opened 2017-06-05
 Makes rigsuits not get contaminated with plasma.

Cameron653:
This was killing people even if they had it in the AMI/rigsuit and didn't have it equipped, as it was in their contents.

Additionally, rigsuit parts were impossible to clean.
This makes sure they don't become contaminated with plasma, solving both of those problems

### DIFF
--- a/code/modules/clothing/spacesuits/rig/rig.dm
+++ b/code/modules/clothing/spacesuits/rig/rig.dm
@@ -199,7 +199,7 @@
 		if(!piece) continue
 		piece.icon_state = "[initial(icon_state)]"
 		if(airtight)
-			piece.item_flags &= ~(STOPPRESSUREDAMAGE|AIRTIGHT|PHORONGUARD) //VOREStation edit
+			piece.item_flags &= ~(STOPPRESSUREDAMAGE|AIRTIGHT)
 	update_icon(1)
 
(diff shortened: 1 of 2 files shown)


code/modules/clothing/spacesuits/rig/rig.dm
```diff
@@ -199,7 +199,7 @@
 		if(!piece) continue
 		piece.icon_state = "[initial(icon_state)]"
 		if(airtight)
-			piece.item_flags &= ~(STOPPRESSUREDAMAGE|AIRTIGHT|PHORONGUARD) //VOREStation edit
```
SpadesNeil: What?

SpadesNeil:
I thought this was fixed in Polaris already.

Cameron653:
As far as I know, this isn't fixed on polaris. Only a select few rigsuits are actually phoronproof. There was this PR: (https://github.com/VOREStation/VOREStation/pull/1547) that tried to fix it but it didn't actually do anything to make the rigsuit pieces, like the boots, gloves, and helmet not get contaminated.

code/modules/clothing/spacesuits/rig/rig.dm
Cameron653: PHORONGUARD does nothing here to actually make it so the rigsuit pieces are unable to be contaminated so I removed it.

PR #1626, opened 2017-06-06
Nanite Implant Framework

Arokha:
Adds the Nanite Implant Framework, the first persistent inter-round item to the game. It's a (very) expensive implant creatable in R&D. It's implanted into the head via the object implanting surgery. After that, there is a 30-minute period of suffering, the first 3 minutes of which the patient will be blind as it attaches to their optic nerve, then after will randomly pass out or have pain. So, it's not a minor surgery. Replacing/re-implanting the NIF after this will only take 1 minute to sync with no side effects.

The fluff is that it is just a nanomachine fabrication hive that builds implants and body mods on the fly from software you buy at NIFSoft Shops, which are just vending machines that only work for people with NIFs. NT doesn't allow self-replicating nanomachine systems though so this is something that requires occasionally being removed and having repairs. 

The crunch is that you spend your hard earned ingame cash on NIFSoft. An assistant won't be able to buy more than one, but the CD/Cap can probably buy 4-5 of them. We're hoping to do persistent money in the future, so here's a good place to spend it. The NIFSofts themselves don't persist for balance reasons, except a civilian AR overlay, and a commlink. NT and their licensing restrictions! Wear is something that persists between rounds, and depends on how often and how many NIFSofts you install, and will require robotics (or someone in R&D anyway) to repair your NIF (sdriver->wires->multitool->sdriver). Medical's job to remove it and hand it to you, though, for you to take to robotics. Make sure you have enough time left in the shift to get it repaired or back in, else it won't be saved with your character. See below.

The current NIFSofts:
* AR HUDs - Version identical to each current AR glasses set. Civ one included for free with every NIF.
* Goggle-alikes - NIFSoft for each 'goggle' ingame (meson, nvg, thermal, etc)
* Nictating Membrane - Basically toggleable welder-vision to protect you from welding. Does not protect you from flashes, but does protect your eyes from the atmosphere outside.
* Fast-Response Filter - Flash protection, like a sechud would have.
* Corrective Lenses - Why have physical corrective lenses when you can have your visual senses altered on-the-fly to be 'correct'?
* Crew Monitor - For the medic on the go.
* Medichines - One version for synth and one for organics, basically increases healing speed (and adds a very minor self heal for very minor damage on synths, 5 or less only). Puts organics in stasis and notifies medical if they reach critical damage (no comm message if being digested).
* Respirocytes - Extra red blood cells stockpile oxygen and allow you to use a small supply of oxygen for emergencies. Only lasts a minute or two, and needs a long time to recharge.
* Commlink - An implanted communicator that displays messages directly into your vision and allows you to make calls and text people without a communicator (also blocked by comm jammer). Included for free after surviving your first shift with your NIF.
* APC Charging - For synths, allows them to recharge from an APC at the cost of the APC cell's charge. Same charge rate as being in a recharger, just more convenient with locations.
* Pressure Seals - For synths, protects synths from pressure damage, but requires enormous power to maintain them, and causes significant wear due to the significant amount of body changes required.
* Heatsinks - Similar to Respirocytes for organics, this allows a synth to sink their heat into heat storage for a short time. Afterwards it must be left off for a significant time to cool down, using energy to cool.
* Mass Alteration - Built in shrinking/growing, but uses a huge amount of energy to do, along with causing significant wear when installed. Plays a visible message so people can tell you're not just using the resize verb.

Illegal/Contraband NIFSoft (available after hacking AND emagging a NIFSoft Shop), these all generate significant wear, and may require a stolen ID or two to pool funds for:
* Bullhide Mod - Adds additional brute armor at the cost of constant energy drain.
* Dragon's Skin - Same, but for burn armor.
* Nova Shock - A toggleable stream of painkillers, but does not track when/if you will overdose, so requires careful management. Will probably lead to addiction.
* Bloodletters - Monofilament claws that significantly increase melee damage.
* Dazzle - Fabricates a single holdout e-gun with only two shots, but causes halloss as you... excrete it.

Obviously many of the above are restricted to jobs that they are sensible for, and some are restricted to heads or captain only. If they CD sees fit, they can pull up the appropriate NIFSoft on the vendor and let another user pay for it if they wish for someone else to buy the soft.

There is also one additional NIFSoft only available via found disk:
* Compliance - The only NIFSoft findable on a disk in trash piles, allows one to upload laws to someone with a NIF. Of course, NT will remove these after the shift is over, but until then...

The NIF does not work in NO_SCAN species (which are most whitelisted species), but a fluff version is available that does so if you want, you can ask an admin about an event or something that gets you one. There's also a cheap knockoff version that doesn't spawn anywhere but maybe I should make it spawn in trash heaps or something.

You need to do one of the following each shift to make sure you keep your NIF implant after you get one:
- Cryo/teleport/tram out with it *in your head*
- Be alive with it at the end of the shift *in your head*
- Get a backup implant and have it for more than 5 minutes while also having a functional NIF

NIFs are given back to you when you're resleeved, they're tied to your mind records. Don't ask how/why the resleeving pod can fabricate a NIF. It's mostly because it's a vore server and having a cool item that means you can't get eaten would suck. So that's just how it works. To prevent duping them, once one has paired with someone, it can't be used by anyone else. On the flipside, already-paired NIFs only take a minute to sync up with their user after being inserted.

The NIF power source is nutrition. So go to the bar and get some good, filling food if you actually want to not get hungry.

~~**Note to admins:** Rejuv'ing people removes ALL implants, NIFs included. This does include backup implants too, 'n always have, but be careful until it's rewritten.~~ Ok added a commit I think fixes this.


code/modules/nifsoft/nif.dm
```diff
@@ -0,0 +1,559 @@
...
+	//First one to spawn in the game, make a big icon
+	if(!big_icon)
+		big_icon = new(icon,icon_state = "nif_full")
+
+	//If given a human on spawn (probably from persistence)
+	if(ishuman(newloc))
+		var/mob/living/carbon/human/H = newloc
+		if(!quick_implant(H))
+			WARNING("NIF spawned in [H] failed to implant")
+			spawn(0)
```
Arokha: I don't know why this comment survived... but this was done.

PR #1627, opened 2017-06-06
chimera bugfixes and tweaks

Screemonster:
Xenochimera now correctly stop breathing while regenerating.

Fixed a bug where chimera that fail to regenerate due to dying before they're finished would end with their cooldown permanently locked on. Now, they can re-attempt it whenever medical injects them with stuff.

Fixed the nutriment check to look in bloodstream/ingested instead of reagents. Also has a minimum threshold for the amount of stuff you need to give them. Protein is more effective than plain nutriment.
No longer take toxloss from the thing needed to revive them.

Feral critters now tick down their shock stage. Their health is unchanged so they still go into softcrit at 0, fullcrit at -50, and die at -100. Softcrit still has the periodic chance of staggering them and will pin their shockstage at 60 every tick as before if it goes under (with the same chance of weakening)
In practical terms, it's a buff to hauling their mangled carcasses off to safety without passing out on the way, they certainly won't be in a condition to do any fighting. Shock stage also affects gun accuracy, but feral mobs can't use guns anyway.

Cold, conversely, now ticks _up_ their shock stage instead of doing halloss. As a result, cold won't render them feral but will give them a fuckton of pain messages and knock them out. At the coldest level, it only takes 10 ticks for them to reach the cap.

### DIFF
--- a/code/modules/mob/living/carbon/human/human_damage.dm
+++ b/code/modules/mob/living/carbon/human/human_damage.dm
@@ -186,15 +186,13 @@
 	..()
 
 /mob/living/carbon/human/proc/Stasis(amount)
-	//if((species.flags & NO_SCAN) || isSynthetic()) vorestation removal
-	if(((species.flags & NO_SCAN) && !istype(species, /datum/species/xenochimera)) || isSynthetic()) //vorestation addition
+	if((species.flags & NO_SCAN) || isSynthetic())
 		in_stasis = 0
 	else
 		in_stasis = amount
 
 /mob/living/carbon/human/proc/getStasis()
-	//if((species.flags & NO_SCAN) || isSynthetic()) vorestation removal
-	if(((species.flags & NO_SCAN) && !istype(species, /datum/species/xenochimera)) || isSynthetic()) //vorestation addition
+	if((species.flags & NO_SCAN) || isSynthetic())
 		return 0
 
 	return in_stasis
--- a/code/modules/mob/living/carbon/human/human_damage.dm
+++ b/code/modules/mob/living/carbon/human/human_damage.dm
@@ -186,13 +186,15 @@
 	..()
 
 /mob/living/carbon/human/proc/Stasis(amount)
-	if((species.flags & NO_SCAN) || isSynthetic())
+	//if((species.flags & NO_SCAN) || isSynthetic()) vorestation removal
+	if(((species.flags & NO_SCAN) && !istype(species, /datum/species/xenochimera)) || isSynthetic()) //vorestation addition
 		in_stasis = 0
 	else
 		in_stasis = amount
 
 /mob/living/carbon/human/proc/getStasis()
-	if((species.flags & NO_SCAN) || isSynthetic())
+	//if((species.flags & NO_SCAN) || isSynthetic()) vorestation removal
+	if(((species.flags & NO_SCAN) && !istype(species, /datum/species/xenochimera)) || isSynthetic()) //vorestation addition
 		return 0
 
 	return in_stasis
--- a/code/modules/mob/living/carbon/human/species/station/station_special_vr.dm
+++ b/code/modules/mob/living/carbon/human/species/station/station_special_vr.dm
@@ -73,7 +73,7 @@
 				H.feral = min(150-H.nutrition, H.feral+1) //Feralness increases while this hungry, capped at 50-150 depending on hunger.
 
 		// If they're hurt, chance of snapping. Not if they're straight-up KO'd though.
-		if (H.stat == CONSCIOUS && H.traumatic_shock >=36 && H.in_stasis == 0) //30 brute/burn, or 18 halloss.
+		if (H.stat == CONSCIOUS && H.traumatic_shock >=36 && !H.inStasisNow()) //30 brute/burn, or 18 halloss.
 			if (2.5*H.halloss >= H.traumatic_shock) //If the majority of their shock is due to halloss, greater chance of snapping.
 				if(prob(min(10,(0.2 * H.traumatic_shock))))
 					if(H.feral == 0)
@@ -110,6 +110,8 @@
 
 	if(H.feral > 0) //do the following if feral, otherwise no effects.
 		var/light_amount = H.getlightlevel() //how much light there is in the place
+
+		H.shock_stage = max(H.shock_stage-(H.feral/20), 0) //if they lose enough health to hit softcrit, handle_shock() will keep resetting this. Otherwise, pissed off critters will lose shock faster than they gain it.
 
 		for(var/mob/living/M in viewers(H))
 			if(M != H) // someone in view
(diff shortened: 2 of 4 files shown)


Screemonster:
Medical players: you need to inject them with 15u of protein, or feed them 30. Alternatively, give them double that amount of non-protein nutriment. For reference, a slab of meat contains 9 units.

code/modules/mob/living/carbon/human/human_damage.dm
```diff
@@ -186,13 +186,15 @@
 	..()
 
 /mob/living/carbon/human/proc/Stasis(amount)
-	if((species.flags & NO_SCAN) || isSynthetic())
+	//if((species.flags & NO_SCAN) || isSynthetic()) vorestation removal
+	if(((species.flags & NO_SCAN) && !istype(species, /datum/species/xenochimera)) || isSynthetic()) //vorestation addition
```
Arokha: Too snowflake for me to include in this generic proc. If you want to have your species treated differently, do it on your species, not on this proc that affects everyone. You should just set in_stasis yourself if you want/need special treatment.

code/modules/mob/living/carbon/human/species/station/station_special_vr.dm
```diff
@@ -160,28 +162,25 @@
...
 		//This is just to prevent them from taking damage if they're in stasis.
 
 	else if(adjusted_pressure2 <= 20) //If they're in an enviroment with no pressure and are NOT in stasis, damage them.
 		H.take_overall_damage(brute=LOW_PRESSURE_DAMAGE, used_weapon = "Low Pressure")
 
-	if(H.bodytemperature <= 260 && H.in_stasis == 1) //If they're in stasis, don't give them them the negative cold effects
+	if(H.bodytemperature <= 260 && H.inStasisNow()) //If they're in stasis, don't give them them the negative cold effects
 		//This is just here to prevent them from getting cold effects
 
 	else if(H.bodytemperature <= 260) //If they're not in stasis and are cold. Don't really have to add in an exception to cryo cells, as the effects aren't anything /too/ horrible.
```
Arokha: Just do this with the normal species vars instead of this extraneous check every life() tick. Set their cold damage levels to 260, 200, and 100 (you currently have them set to -5000 and handle it yourself?).

Also those are basically the defaults for cold damage levels for humans (260,200,120). I thought they were MORE susceptible to cold rather than being MORE RESISTANT to it, only taking HALLOSS instead?

code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Food-Drinks.dm
```diff
@@ -77,6 +77,9 @@
 	if(alien && alien == IS_SKRELL)
 		M.adjustToxLoss(2 * removed)
 		return
+	else if(istype(M.species, /datum/species/xenochimera)) //vorestation addition
+		affect_ingest(M, alien, removed) //like hell are chimera of all things going to be poisoned by injecting them with meat
```
Arokha: Doing a check for your snowflake species literally anytime anyone eats anything on the server ever seems a little much. Why are you doing this, anyway? Only skrell get poisoned by it.

Screemonster: This is the affect_blood, not the affect_ingest. Protein isn't set as an injectable form of nutriment, so the ..() on the line below this one causes them to take tox from it. Can remove it though, it's not a huge amount.

code/modules/mob/living/carbon/human/species/station/station_special_vr.dm
Screemonster: The normal check is "slowly take burn damage at these levels", not "get slowed to a crawl and freeze". I could set the levels to be harsher easily enough too.

code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Food-Drinks.dm
Arokha: I see. To be honest, this is the original code:
```
	if(!injectable)
		M.adjustToxLoss(0.1 * removed)
		return
	affect_ingest(M, alien, removed)
```

I'd say change it to just:
```
	if(injectable)
        	affect_ingest(M, alien, removed)
```
I dunno why it should poison you, but it does make sense that non-injectible ones are not 'useful'.

code/modules/mob/living/carbon/human/human_damage.dm
Screemonster: I can set in_stasis manually, sure. The line below is the real kicker, 'cause all the things that actually check in_stasis are set to return 0 for NO_SCAN species. Almost tempted to just make them not NO_SCAN given the number of things that check it that don't make sense for chimera and oldstyle cloning isn't a thing any more.

code/modules/mob/living/carbon/human/species/station/station_special_vr.dm
Arokha: I'd be more amenable to writing a general handle_cold and handle_hot proc on species that gets called from life and passed the level of problem though.

I guess just leave this for now. Leshana and I want to rewrite how cold/hot works since it's handled in breathing so does nothing for synths or people with a gasmask.

code/modules/mob/living/carbon/human/human_damage.dm
Arokha: NO_SCAN is still used other places though, as a flag for "too complicated to have certain things happen" biology. Like xenochimera will start taking cloneloss and scrambles from radiation among other things.

Screemonster: Strictly speaking, the things that cause cloneloss are things that should fuck them up. Like, a lot. I was originally going to have it (along with brain) be one of the things that regenning didn't clear properly and actually need actual treatment, along with having their cloneloss levels as a thing that increases revival time. Shooting them with a decloner makes sense as a thing that would _ruin_ them.

Arokha: Actually I should point out, in_stasis may not do what you want it to do anymore, you may want to re-evaluate your use of it in general if you just want to "stop life ticks" while they regenerate. I notice there's a 'transforming' var that causes life ticks to just return at the top. You might want to consider using that, rather than wasting time dividing some number by ten thousand every tick.

Screemonster: Oh shit, I didn't know about that one. That probably makes a lot more sense to use.
Looking over it, the only other places it's set are monkeying, gibbing and staff of change, and purely to stop the mob moving while they do their thing.

code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Food-Drinks.dm
Screemonster: Works for me. As far as I can tell the "injectable" var is literally only set for glucose. Why they didn't override glucose's affect_blood rather than checking a var for every nutriment type is beyond me.

Arokha: "Unrealized future plans" is generally the answer to every question like that.

code/modules/mob/living/carbon/human/species/station/station_special_vr.dm
Screemonster: Cool cool. I'll remove the stasis check since nothing can set stasis on them besides regenning (which is being changed to use transforming anyway). This new proc (directly incrementing shock_stage) means that putting them in a cryotube will correctly _fuck them up_.

code/modules/mob/living/carbon/human/human_damage.dm
Screemonster: Okay, looking over it (and testing), transforming really really really fucks up too much stuff. 
HOWEVER turns out the does_not_breathe var does exactly what I wanted out of stasis. It's used by the self-respiration changeling power.
Yes, chimera can bleed out and die while regenerating if they have wounds/IB. I see no reason to make this a thing they can avoid by taking a nap.

Screemonster:
Tweaked as per hidden comments in the review. Less snowflake. Uses the ling self-respiration var instead of stasis since that does exactly what I needed of it.

PR #1639, opened 2017-06-08
Adding Harpywings! (also a kitsune thing!)

LiquidFirefly:
FINALLY

This isn't attached to a character, and isn't personal fluff, so I don't know if this is the proper channel to just go ahead and PR, but whatever.


Screemonster:
I've been wanting to do these for ages and never got around to it, ever since I added nevreans back on oldcode.

LiquidFirefly:
It was hard as heck getting them the way I wanted them, and I'm not so sure I'm fully satisfied with them even now, but I can always fiddle with it, so whatever.

LiquidFirefly:
http://forum.vore-station.net/viewtopic.php?f=45&t=964&sid=6e6f854bb570c711f2e8c4d0f8edaa4e Linking this in here because of Seiga's stuff

PR #1643, opened 2017-06-08
Removes the snowflake human defined verbs from the majority of the reagent implants

Yoshax:
Thanks for @Arokha telling me about a secret DM feature no longer do we need a bunch of coded verbs. We only need one! And types are dynamically created when appropriate using vars on the implant. 

Tempest's, Belle's and Theodora's are converted to use the new system. Roiz's remains the same and gains its own :snowflake: implanted() because it doesn't obey the rules of the rest of it.


Arokha:
thx 4 fix

PR #1648, opened 2017-06-09
Fixes some colourisation issues

Yoshax:
They were just added, white sprites. White does multiply.


Arokha:
t s k
@LiquidFirefly

PR #1654, opened 2017-06-09
Longpup mobs! (plus an approved fluff item committed earlier)

Verkister:
-Plain and just hostile root version. (only spawn if you're looking for a bad time)
-Recruitable follower pet version that starts hostile but has a 50% chance of taming and befriending on petting. (pray for rngesus)
-Same as above but with a cute little collar!
-Guard puppy version that comes tame by default, runs after criminals, and has the 50% chance of befriending and following.
-They may still be hostile towards station pets and simple bots. How rowdy of them and how lazy of me.

-Also includes the code bit for a fluff item (aesthetic science goggles) http://forum.vore-station.net/viewtopic.php?f=26&t=959 (Whoops nice tab management here)


code/modules/mob/living/simple_animal/vore/otie.dm
```diff
@@ -0,0 +1,297 @@
...
+
+/mob/living/simple_animal/otie/attack_hand(mob/living/carbon/human/M as mob)
+	..()
+	switch(M.a_intent)
+
+		if(I_HELP)
+			if (health > 0)
+				LoseTarget()
+				handle_stance(STANCE_IDLE)
+				if(prob(50)) //It's a fiddy-fiddy you may get a buddy pal or you may get mauled and ate. Win-win!
```
Arokha: This will not result in you being mauled and eaten since the parent also checks intent, or maybe that was just a joke. :P Or did you mean WHILE BEING ATTACKED and patting them?

code/modules/vore/fluffstuff/custom_items_vr.dm
```diff
@@ -892,6 +892,14 @@ obj/item/weapon/material/hatchet/tacknife/combatknife/fluff/katarina/handle_shie
 		accessset = 1
 	..()
 
+//verkister: Cameron Eggbert - Science goggles that ACTUALLY do nothing.
+/obj/item/clothing/glasses/science_proper
+	name = "Aesthetic Science Goggles"
+	desc = "The goggles really do nothing this time!"
```
Arokha: I assume this is to avoid the purple overlay?

Regardless, the forum thread you linked, http://forum.vore-station.net/viewtopic.php?f=26&t=957&sid=e4bfb9a1ce015f3e3ab282d998263a68 is not related to this item. Wrong post?

code/modules/mob/living/simple_animal/vore/otie.dm
```diff
@@ -0,0 +1,297 @@
...
+
+		if(A == src)
+			continue
+
+		var/atom/F = Found(A)
+		if(F)
+			T = F
+			break
+		else if(specific_targets)
+			return 0
```
Arokha: You have specific_targets set to true on all of these, so really you could handle your tamed checks in Found() and not override this one here. On the one that does the sechud broadcast, you could override that in set_target instead.

The reason I suggest doing that instead is because this is a big proc with lots of little fiddly bits that may be changed in the future, and you wouldn't get those changes and your mob may stop behaving like one might expect. set_target is less likely to change since it's a very simple proc... there's not really much to fix about it, and Found() will never be changed because it's whole purpose is to be overridden by children.

code/modules/vore/fluffstuff/custom_items_vr.dm
Verkister: Yup had some top tier tab management there.
http://forum.vore-station.net/viewtopic.php?f=26&t=959

Arokha:
This seems fine to me, just checking with staff to make sure it's okay with them.

Arokha:
I meant to squash-merge this. But. Oh welllllllll.

PR #1661, opened 2017-06-09
Yet Another Overlay Avoidance

Arokha:
Now on this turf type. How many times will we have to do this?


Dhaeleena:
owo what is this?

Arokha:
The server randomly has problems (like you saw with the elevator) with turfs becoming corrupted and it keeps moving from turf to turf.

PR #1667, opened 2017-06-10
Added colored eggs and did some minor tweaks.

SpoopyLizz:
I've added new sprites to the eggs I made so they can be recolored with crayons. The new sprites are in a seperate .dmi to prevent merge issues.
I've also did some minor tweaks to the code of the eggs.


code/modules/vore/fluffstuff/custom_items_vr.dm
```diff
@@ -788,7 +788,25 @@ obj/item/weapon/material/hatchet/tacknife/combatknife/fluff/katarina/handle_shie
...
-	bitesize = 4
+	bitesize = 2
+
+/obj/item/weapon/reagent_containers/food/snacks/egg/roiz/attackby(obj/item/weapon/W as obj, mob/user as mob)
+	if(istype( W, /obj/item/weapon/pen/crayon ))
+		var/obj/item/weapon/pen/crayon/C = W
+		var/clr = C.colourName
+
+		if(!(clr in list("blue","green","mime","orange","purple","rainbow","red","yellow")))
+			usr << "\blue The egg refuses to take on this color!"
```
Arokha: No color macros (\blue, \red, etc), and don't use usr. You have 'user' in this proc based on the args it takes, so use that instead. Also don't use the stream operator anymore, use `to_chat`
`to_chat(user,"<span class='warning'>The egg refuses to take on this color!</span>")`

That will make it print with the 'warning' css class, which is actually red italics. You should probably use that for errors. For the blue one, like the success one below, you can use 'notice' instead, which is blue italics.

PR #1669, opened 2017-06-10
Allows rigsuits to be able to carry backpacks

Cameron653:
Because it was removed for some reason, making jobs like mining/xenoarch incredibly annoying.


SpadesNeil:
At some point we should change these to _add_ to the list, rather than override it entirely.

PR #1686, opened 2017-06-11
New dogborgo! (fixed?)

Verkister:
-Oh god this has been my best worst robopup idea so far lmao.
-Added janitor type dogborg.
-Its speciality is a modified sleeperbelly fitted to ingest stuff too!
-This might be very problematic in the wrong hands lmao.
-Increased sleepergurgle item charge from 10 to 100. 10 was so little that the whole code bit for item gurgle charge was practically useless.
-New module hud icons for all dogborgs.
-Removed the unused movement state icons from the static dogborgs.
-Added a new hud icon for brig belly.


code/modules/mob/living/silicon/robot/dogborg/dog_sleeper_vr.dm
```diff
@@ -431,7 +431,36 @@
...
+/obj/item/device/dogborg/sleeper/compactor/afterattack(var/obj/item/target, mob/living/silicon/user, proximity)
+	hound = loc
+	if(!proximity)
+		return
+	if(!isobj(target))
+		return
+	if(target.anchored)
+		return
+	if(target.w_class > ITEMSIZE_LARGE)
+		user << "<span class='warning'>This object is too large to fit into your [src.name]</span>"
```
Leshana: We recommend [src] instead of [src.name]

code/modules/mob/living/silicon/robot/robot_vr.dm
```diff
@@ -30,3 +30,42 @@
...
 			overlays += "[module_sprites[icontype]]-sleeper_r"
+
+/mob/living/silicon/robot/Move(a, b, flag)
+
+	. = ..()
+
+	if(module)
+		if(module.type == /obj/item/weapon/robot_module/scrubpup)
+			var/turf/tile = loc
+			if(isturf(tile))
```
Leshana: Okay.  I think here what we want to do is call  `T.clean(src, src)` which will handle the turf cleaning part for you.  It will also use up some water, which I think is a good idea.    Just need to add a little bit of extra code allowing the dogborg to drink from a sink to refill their reagents with water!

code/modules/mob/living/silicon/robot/robot_vr.dm
```diff
@@ -30,3 +30,42 @@
...
+								cleaned_human.wear_suit.clean_blood()
+								cleaned_human.update_inv_wear_suit(0)
+							else if(cleaned_human.w_uniform)
+								cleaned_human.w_uniform.clean_blood()
+								cleaned_human.update_inv_w_uniform(0)
+							if(cleaned_human.shoes)
+								cleaned_human.shoes.clean_blood()
+								cleaned_human.update_inv_shoes(0)
+							cleaned_human.clean_blood(1)
+							cleaned_human << "\red [src] cleans your face!"
```
Leshana: We can't use \red anymore, use  <span class='warning'> please.

code/modules/mob/living/silicon/robot/dogborg/dog_sleeper_vr.dm
Verkister: [src] was cramming in an extra "the" into the sentence which kinda felt grammatically off.

code/modules/mob/living/silicon/robot/robot_vr.dm
Verkister: Sounds pretty good. I pretty much just copied this whole part from the normal janiborg.

Verkister: Actually idk. Depends if the thing will do the draining for every tile moved over whether they're clean or not. After all this is the part that handles the janiborg cleaning the floors by just running them over.

code/modules/mob/living/silicon/robot/robot_vr.dm
Verkister: Roger that.

code/modules/mob/living/silicon/robot/robot_vr.dm
Verkister: Anyway I'm feeling pretty expended over last night (up til like 5 am).
I'll leave that stretch for someone else to code for now if necessary.

Verkister:
Okay that check failure was none of my doing or even related to this stuff. gg congrats for your random bug travis.

Verkister:
So we good yet?

code/modules/mob/living/silicon/robot/dogborg/dog_sleeper_vr.dm
Verkister: Same as with the recent camera related PR that was about the exact same thing :v

Leshana:
I do not know very much about borgs at all, and don't have time to research them.  However the issues I identified before have been fixed.  I have some suspicions about it cleaning everything it walks over, that should probably be re-visited in the future.

Arokha:
Am somewhat concerned people will just eat literally everything because "lol why would you have given me the ability to if you didn't want me to??", and it's kinda meh that you can't see items that you've eaten, or even know how many you have in you, but codewise it's fine (other than our desire to have it clean the floor with reagents. Also the override to Move() is a little weh.)

Verkister:
I never said it was a very good or well balanced idea in the first place, but folks were way more for than against it. In fact this is the first time anyone (other than me) has brought up any balance concerns. At least the recently added capacity limit should help with having some sort of idea of the clutter amount.

And to the floor cleaning concern, the ability (and code) is still simply copied from the vanilla janiborgs. They have always done it too.

code/modules/mob/living/silicon/robot/dogborg/dog_sleeper_vr.dm
```diff
@@ -449,42 +449,55 @@
...
 	hound = loc
-	if(!proximity)
-		return
-	if(target.anchored)
-		return
-	if(length(contents) > 32)
-		user << "<span class='warning'>Your [src.name] is full. Eject or process contents to continue.</span>"
+	if(isobj(target))
+		if(length(contents) > 11)//grow that tum after a certain junk amount
+			hound.sleeper_r = 1
```
Leshana: Probably should not update fields an icons before the proximity/anchored/etc checks are performed.
Also, setting sleeper_r = 1 will be overwritten by the next call to update_patient() because it doesn't take contents length into account.

Verkister: The icons for the bellies are identical without the lights so it's not a noticeable change. A person inside gives the bulge regardless of the clutter amount and since the ejection is all at once it shouldn't have appearance problems which might appear if having person and just a few items being left inside was even possible.

Anyway the other things could be tidied up a bit yeah.

code/modules/vore/eating/living_vr.dm
```diff
@@ -25,7 +25,7 @@
 		return 1
 	M.verbs += /mob/living/proc/insidePanel
 
-	M.appearance_flags |= PIXEL_SCALE
+	//M.appearance_flags |= PIXEL_SCALE
```
Leshana: I think this is beyond the scope of this PR.  If you want to disable pixel scale on all mobs, that needs to be its own discussion, its bigger than dogborgs.

Verkister: Yeah that wasn't intended to sneak into the pr at all in the first place. Only had that thing in the background for testing purposes as it kept erroring the compile.

PR #1693, opened 2017-06-11
Adds the SleeveMate 3200

Arokha:
**This commit prevents dead people from being backup-implanted!**
But it's fine because you just use this handheld thing to back them up instead and don't waste an implant, and medical starts with 4 so don't freak out. See below.

Adds the SleeveMate 3200 hand-scanner device. This device allows you to scan people to determine several things about them:
- If they have a mind in them
- If they have a client attached
- If their mind matches their body
- If they are alive/conscious

All of that is represented with in-universe messages of course. The real function of it is that it provides several features:
- One-time mind scan: Performs a one-time mind scan for those that don't like getting mind backup implants. This will store that one scan in the database, but you'll forget everything since then if resleeved. This is also useful for resleeving someone who died without an implant, but is not DNR/DNC. Won't notify medical if the person dies, since it has no way of knowing.
- One-time body scan: Adds someone to the body scan database, or, updates their body scan if they obtained a new body. Useful if someone has used the body designer and switched bodies, as you can update their body scan. You should probably check medical records, and LOOC around before using this to resleeve someone that didn't have a body scan originally, because they may have picked that to avoid resleeving.
- Store Whole Mind: Takes the entire mind from someone, storing it into the SleeveMate 3200, rendering the patient mindless. From there, it can be backed up as many times as needed, deleted, or stored into someone with a Soulcatcher NIFSoft and run live in their VR (so now you don't have to digest/kill people for that).

Four SM3200's start in medical (1 in minimed downstairs, 3 upstairs), R&D can fab more, and they occasionally show up in trash piles for those wanting to be sneaky with a Soulcatcher, but who don't like digesting people.

Backup-implanting dead people will no longer work (it was never intended to be a permanent solution, I was always going to design this handheld thingy), just bop them with a SM3200 and do 'One-time mind scan'. Exact same thing.


code/game/objects/items/devices/scanners_vr.dm
```diff
@@ -0,0 +1,209 @@
...
+	else
+		to_chat(user,"<span class='warning'>Not a compatible subject to work with!</span>")
+
+/obj/item/device/sleevemate/attack_self(mob/living/user)
+	if(!stored_mind)
+		to_chat(user,"<span class='warning'>No stored mind in \the [src].</span>")
+		return
+
+	var/choice = alert(user,"What would you like to do?","Stored: [stored_mind.name]","Delete","Backup","Cancel")
+	switch(choice)
```
Leshana: If you add  `if (!stored_mind) return` it will gracefully handle multi-clicking.

code/game/objects/items/devices/scanners_vr.dm
```diff
@@ -0,0 +1,209 @@
...
+		return
+
+	if(href_list["mindsteal"])
+		if(!target.mind)
+			to_chat(usr,"<span class='warning'>Target seems totally braindead.</span>")
+			return
+
+		usr.visible_message("<span class='warning'>[usr] begins downloading [target]'s mind!</span>","<span class='notice'>You begin downloading [target]'s mind!</span>")
+		if(do_after(usr,35 SECONDS,target)) //This is powerful, yo.
+			if(!stored_mind)
```
Leshana: Need to make sure target still has a mind after the 35 second wait.

code/modules/resleeving/infocore.dm
```diff
@@ -67,7 +68,7 @@ var/datum/transhuman/infocore/transcore = new/datum/transhuman/infocore
 			MR.nif_software = null
```
Leshana: `if (!one_time) MR.one_time = FALSE`  perhaps? To handle if someone gets a one time backup, THEN gets an implant later?

code/game/objects/items/devices/scanners_vr.dm
Arokha: Targets who have died actually still have a mind, it's set on BOTH their ghost and their body, that's why post-mortem implants work, but I suppose someone ELSE with a sleevemate could have taken it meanwhile.

Leshana: True true! But still, they might have ghostized in some other way! Someone else with a sleevemate or them ghostfacing or something right?

code/modules/resleeving/infocore.dm
Arokha: Context is not right on this one. Fix is above.

Leshana:
For anyone else viewing this:  This PR has a good description, this is what I like to see for stuff!

PR #1727, opened 2017-06-13
Converts the mob process controller to a StonedMC subsystem

Leshana:


Arokha:
Pretty straightforward!

PR #1728, opened 2017-06-13
Fix preference sanitizing and mousedrop runtimes

Leshana:
Fixes un-sanitized preferences runtimes

* Puts back in code that sanitized preferences after loading, it was overwritten in polaris sync commit 3f1e5c80f25af98e27be802cb06a218026a33769
* Also removes a VOREStation Edit comment on lines that were no longer different from Polaris, introduced at the same time.
* Fixes Runtime in 02_language.dm,21: Cannot read null.language

Fix runtimes in MouseDrop when dragging out of the screen

* MouseDrop()'s over_object may be null if dropping over a stat panel or over other empty space. Fix runtimes from assuming it is not null.
* Fixes Runtime in clothing_accessories.dm,54: Cannot read null.name
* Fixes Runtime in evidencebag.dm,21: Cannot read null.loc

### DIFF
--- a/code/modules/client/preferences.dm
+++ b/code/modules/client/preferences.dm
@@ -243,6 +243,7 @@ datum/preferences
 		load_preferences()
 		load_character()
 		attempt_vr(client.prefs_vr,"load_vore","") //VOREStation Edit
+		sanitize_preferences()
 	else if(href_list["load"])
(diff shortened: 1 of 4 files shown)


Arokha:
> Fixes Runtime in 02_language.dm,21: Cannot read null.language

Yaaaaay.

PR #1729, opened 2017-06-13
Adds 3 new guns

Arokha:
* KHI-98a 'Protector'
* MPA6 'Sickshot'
* PML9 'Pummeler'

image

The Protector is an energy gun that can't be fired on lethal on code green, so it's considered safe for sec to carry regularly, and has a slightly enhanced stun over tasers and normal eguns. It also features an integrated flashlight, toggleable with ctrl-click, and a prototype update_icons proc from /tg/ that makes it all go. An emag can disable the security level lock.

The Sickshot is a stun weapon in handheld form, that induces nausea and vomiting. It has a chance to make predators vomit out prey (regardless of belly) and tends to cause confusion and nausea. It does cause slight burns from the hypersonic frequency sound it uses, but nothing serious or potentially lethal.

The Pummeler is a 'bass cannon' essentially that fires a huge sound wave, tossing the target away from you. It may or may not knock them down, depending on your luck, but it will definitely push them away, and potentially over a railing or wherever you wanted to put them.


SpadesNeil:
I'm not a fan of DNA locks being a regular thing on guns in the game. It gives Security an extra excuse to be reckless with gun carry because they know "hurr hurr even if I lose my gun no one can shoot it" when they shouldn't be losing their gun in the first place.

Also it makes it very one-sided for when an antag or criminal character _does_ steal a gun.

Arokha:
They aren't starting guns or anything, they're made in R&D. None of the 'normal' sec guns have DNA locks. Also since an emag clears the DNA locks (and lethal mode restriction on it), and antags could/would have emags (and you can find them in trash piles), I don't really see the issue. Not to mention, why would an antag want this law enforcement weapon instead of whatever they brought with them?

Considering it's the only gun with a DNA lock, it should be pretty obvious when stealing the gun isn't the best plan, though it was suggested I make DNA locks craftable in R&D too.

Screemonster:
> pummeler
yeeeeeeeeeeeeeessss.

SpadesNeil:
>They aren't starting guns or anything, they're made in R&D. None of the 'normal' sec guns have DNA locks. Also since an emag clears the DNA locks (and lethal mode restriction on it), and antags could/would have emags (and you can find them in trash piles), I don't really see the issue. Not to mention, why would an antag want this law enforcement weapon instead of whatever they brought with them?

I said antags _or_ criminals characters. Criminal characters being people like Jarome who are causing trouble without crossing into blatant antag territory.

>Considering it's the only gun with a DNA lock, it should be pretty obvious when stealing the gun isn't the best plan, though it was suggested I make DNA locks craftable in R&D too.

No one item or gun should be perfect for its purpose. It eliminates variety and challenge. I like the code green restriction. I do not like the DNA lock. It's too many features on one gun. A different gun can come with a DNA lock, though.

Epitaxis:
These sound guns are going to make Teshari hate their lives. Will there be any plans on giving it bonus effectiveness or even ear damage to Teshari as a side effect of Teshari sensitive hearing?

Arokha:
I like that sprite too much and don't feel like arguing so I'm gonna redo that gun to have a different sprite then, and use the nice one for another gun.

code/modules/projectiles/gun.dm
```diff
@@ -333,7 +363,12 @@
 	accuracy = initial(accuracy)	//Reset the gun's accuracy
 
 	if(muzzle_flash)
-		set_light(0)
+		//VOREStation Edit - Flashlights
+		if(gun_light)
```
SpadesNeil: I have concerns with this. The last time I added light to guns, it caused immense problems of lag. Keep in mind I had done it so all _projectiles_ would be lit up. This should be a lot less awful but if a major gunfight breaks out it could be really problematic.

Just don't add light to moving projectiles. Lasers are probably fine because they're instant because they're beams. It's the act of lots of bullets traveling across the screen with light on them that horribly bottlenecks the server, causing lag worse than atmosia could imagine.

code/modules/vore/fluffstuff/custom_guns_vr.dm
```diff
@@ -891,7 +677,7 @@
 	mag_type = MAGAZINE
 
 /obj/item/ammo_casing/stg
-	desc = "A 7.92×33mm Kurz casing."
+	desc = "A 7.92 33mm Kurz casing."
 	icon_state = "rifle-casing"
 	caliber = "kurz"
 	projectile_type = /obj/item/projectile/bullet/rifle/a762
```
SpadesNeil: This looks incorrect. Nothing to do with this PR but this is probably something that should be fixed later.

vorestation.dme
```diff
@@ -2401,6 +2401,11 @@
 #include "code\modules\vore\fluffstuff\custom_items_vr.dm"
 #include "code\modules\vore\fluffstuff\custom_mecha_vr.dm"
 #include "code\modules\vore\fluffstuff\custom_permits_vr.dm"
+#include "code\modules\vore\fluffstuff\guns\bsharpoon.dm"
+#include "code\modules\vore\fluffstuff\guns\dominator.dm"
+#include "code\modules\vore\fluffstuff\guns\protector.dm"
+#include "code\modules\vore\fluffstuff\guns\pummeler.dm"
+#include "code\modules\vore\fluffstuff\guns\sickshot.dm"
```
SpadesNeil: This refactor of how we handle uniquely coded guns is probably how we should do this in future. However I would move these under the original /guns directory, and add a /vore or /vorestation folder to put these in.

Leshana: I agree.  Deese separate files are nice.
This code looks like it will work to me.

code/modules/projectiles/gun.dm
Arokha: Lasers already do all have lights. I doubt it'll be much of an issue on the new lighting system. We have /tg/'s lights, and this gun with a light is from /tg/, so I think it'll work. In fact you can attach a maglight to tons of /tg/ guns, and they have a generic system for adding the light sprite without needing an additional sprite for every gun.

code/modules/vore/fluffstuff/custom_guns_vr.dm
Arokha: I removed the strange non-unicode X character. I have no idea what character that is, but it looks insane and can't be rendered right ingame. Just removed it while I was there. Can readd a normal X or something later.

vorestation.dme
Arokha: I have another gun to add so I'll do that on that one.

code/modules/vore/fluffstuff/custom_guns_vr.dm
SpadesNeil: I meant the projectile_type variable.

SpadesNeil: I don't know why that weird × was there either.

PR #1734, opened 2017-06-14
Fixes objects using walk_to() not being able to be garbage collected.

Leshana:
* Okay evidently walk_to(A, B) doesn't stop when A reaches B, but keeps running in the background forever until it is manually canceled!  Therefore in order to be garbage collected, we must cancel walking on any object that might have initiated it.
* Fixes this on chemsmoke, spiders, and aiholos.
* Also fixed chemsmoke actually getting qdeled at the right time.

### DIFF
--- a/code/game/gamemodes/events/clang.dm
+++ b/code/game/gamemodes/events/clang.dm
@@ -41,6 +41,10 @@ In my current plan for it, 'solid' will be defined as anything with density == 1
 		if(clong && prob(25))
 			src.loc = clong.loc
 
+	Destroy()
+		walk(src, 0) // Because we might have called walk_towards, we must stop the walk loop or BYOND keeps an internal reference to us forever.
+		return ..()
+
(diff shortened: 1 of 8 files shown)


Cyantime:
Is there a reason to not just put walk(src, 0) in atom/movable's Destroy()? That seems like it would also handle any cases that come up in the future.

Leshana:
@Cyantime That would work, and it is safe to do so.  I was thinking about performance.  
Overriding /Destroy adds another proc-call to these specific types (and some of them already had /Destroy anyway)  whereas adding `walk(src, 0)` to  `/atom/movable/Destroy()` would add however much `walk(src, 0)` costs to *every* object in the game.     
I don't know how much `walk(src,0)` actually costs.  Maybe its super cheap and its worth doing? I just played it safe on performance terms after doing a global search for walk_to and walk_towards.

PR #1741, opened 2017-06-15
Fixes players not being removed from the manifest immediately when cryoing.

### LINKED ISSUE #1713: Leaving Round [Outside of Digestion
Leaving the round, such as through Cryo, Does not currently remove you from the Manifest

To vanish from the Manifest

#### What actually happened
I was still on it despite having left

#### Steps to reproduce
- (Step 1)Join and be on the Manifest
- (Step 2)Leave and remain on the Manifest
- (and so on)

#### Code Revision
-Actual revision does not appear, have the code thing instead [a5c8e4780da816dbf5bd7edd17b467ddf55985a3]
 
Leshana:
* The cryopod simply qdel's the datacore records. Old garbage.dm hard-del()'d datums, so it was fine.  Now we need to make sure they are actually removed from the data_core.
* Testing shows this fixes #1713 and in most cases the objects also GC cleanly.  However some older computers still retain hard references to data_core entries. But even in the cases where that prevents a clean GC, the player is still removed from the manifest successfully and promptly, so that can wait.

### DIFF
--- a/code/_helpers/logging.dm
+++ b/code/_helpers/logging.dm
@@ -180,6 +180,8 @@
 	return "[..()] ([ckey])"
 
 /proc/log_info_line(var/datum/d)
+	if(!d)
+		return "*null*"
 	if(!istype(d))
-		return
+		return json_encode(d)
 	return d.log_info_line()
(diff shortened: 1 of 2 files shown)


Leshana:
Note: Even in the event that a computer still has a hard reference, it ends up looking like: 
image
So we can just put fixing that on the back burner, no rush.

PR #1747, opened 2017-06-16
Fixes suffocation when eaten/picked up

Cameron653:
Credit to Shadow Larkens/tigercat2000 for fix


Cameron653:
Tested and can confirm this works. Needs merged into release ASAP

PR #1749, opened 2017-06-16
Adds a unit test to make sure prey can breathe inside pred bellies.

Leshana:
I tested with pre-fix and post-fix code to make sure it really fails when broken and passes when not.


Arokha:
Seeeems useful.

PR #1752, opened 2017-06-17
KHI-102b "NSFW" Handgun

Arokha:
The "Nanotech Selectable-Fire Weapon" is an energy weapon that takes microbatteries to determine the fire modes rather than weapon power cells. There are various weapon cells like stun, shrink, kill, xray that all offer 4 shots of that type, and you can load 3 cells into the magazine at once. So if you load stun-stun-stun you get 12 shots of stun. If you load stun-net-lethal you get 4 of each. Stun-stun-net gives 8/4, etc.

The gun itself isn't craftable ingame. It was given to the ERT commander with four of the ammo types, and to deathsquad with all the ammo types including 'final option' ammotype (brain scan + instagib). Other people could request it as a fluff item with certain ammo types if they want! People expressed that they thought the sprite was really good. It takes longer to recharge than a normal energy gun since you have to take the batteries out and recharge them individually.

image
image

Full disclosure: I also replaced my own fluff netgun with a stun/stun/net NSFW (so, same thing as a netgun, just a cooler sprite). Still falls under the terms of my permit, so should be fine.

I also fixed the slidey walls at centcom while I was there.


SilverTalismen:
I don't have any issue with this, while the final option is questionable, if gives you a brainscan so your not gone forever.

Arokha:
> I don't have any issue with this, while the final option is questionable, if gives you a brainscan so your not gone forever.

That's something only deathsquad has. I just thought it was a neat idea, because Kitsuhana probably wouldn't make a way to permanently kill someone, so... it brainscans them and then gibs them! So you can revive them 'after the fighting' as it were, for interrogations or whatever.

Ketrai:
I'm very anxious at seeing that asset protection will have instagib ammo. It seems way too powerful, and honestly a bit unrealistic if it's just going to gib people regardless of armor, and what the gun looks like. It's like an energy revolver, but bam, instagibbed. if it was like a heavy energy rifle that shot explosive bullets and looked badass, it would be more believable. I'm also like, the colour changes with the ammo types... but how are people supposed to know what each type of ammo does? also, when you quickly switch, you can surprise people, way too easily. make it look like a harmless stun gun, switch, shoot, bam netted. It's like i'll be afraid of anyone using this gun because I don't know what it does. Maybe make it so you have to load in a new weapon cell just to change what it fires? I mean, i'm all up for mystery guns... but if people can get this through normal fluff apps? yikes.

Arokha:
I think we talked about this in chat and you seemed satisfied. Just leaving that here so people who read this notice. Also several people have completely unique fluff guns... you can't always expect to know what they will do just by the name. And you can't tell what ANY eguns are set to by examining the person holding them. They just say egun. They could just say it's set to stun and shoot you with lethal, so that's not really specific to this. If you think that should be changed, sure I might agree with that since someone standing nearby should be able to see the colors on guns that have colors, however that's outside the scope of this PR.

Anyway, the 'super ammo' is literally sitting directly on top of an RPG and ammo in the ERT 'closet' at centcom, and that's the only place it'll go. Considering it's right next to a Pulse Destroyer and other stuff like that, I wouldn't worry too much about it. Deathsquad is pretty much just played for gags here anyway.

Arokha:
You should be more worried about the other Deathsquad starting equipment like this:
image


And I've got four more in the coat! And one on my ear! Because that's how that works yaya.

SpadesNeil:
I like abusey things being tossed in the code sometimes for reasons of **FUN™** but this certainly won't be something that appears regularly, if at all.

I care little if the gib gun is added or not. If I wanted to gib people with a gun I'd use the meteor launcher.

PR #1753, opened 2017-06-17
Reduce soulcatcher cost some

Arokha:
People seem to like it so why not.


SilverTalismen:
Thank you. How much is it now?

Arokha:
Nine fiddy.

PR #1793, opened 2017-07-01
Dakka Dakka

SpadesNeil:
- Removed gun_vr.dmi judge, Polaris has one now.
- Removed vorestation's m44 mags, Polaris has those now.
- Removed detective's .44 magnum because NO YOU CANNOT HAVE A FUCKING MAGNUM
- Corrected inconsistent naming of .44 speedloaders
- Corrected inconsistent naming of .357 speedloaders
- Fixed broken calibers for .357 stuff to be consistent with upcoming Polaris sync.
- Imported fixes from https://github.com/VOREStation/VOREStation/pull/1760 also.


SpadesNeil:
This will not compile until a Polaris sync is done.

Dingosilver:
So what will the detectives get from here on then? The justification for the HoS's .44 being removed and not replaced seems to have been that they have the laser to fall back on with its stun setting, and it thus wouldn't need to be replaced with anything. The detectives just get, what, standard officer gear now minus the armour? If you're going to get rid of everything outside of the armoury that might possibly be lethal, why not make non-lethals more worth using? Maybe replace the magnum with a net gun, or something.

Yoshax:
Take a stun revolver. You can't get more worth using than one of those.

Cameron653:
Also, the SecTech in the security briefing room has a .45 pistol. He could easily get grab one of those, or, as stated above, a stun revolver from the SecTech.

SpadesNeil:
I could give the detective a .32 revolver again probably. Also detectives aren't supposed to be getting into fights, which is why they have the default locker loadout that they do.

There's no rule IC or OOC preventing them from just taking one of the many armored vests from a locker, or any officer gear.

Dingosilver:
The .45 pistols in the SecTech only use flash bullets though, which for all intents and purposes seem like a ranged flash considering how useless they are with sunglasses on. As for tasers/stun revolvers, I still find them incredibly underwhelming for what they are, but it does make them more fair to be on the receiving end of, so I can see the argument for their current state. I always just thought the detective having something like a revolver was to emphasise that their job isn't to be making arrests and that them having a weapon is solely as a means of self-defence from those that may be trying to get in the way of an investigation.

Yoshax:
'Underwhelming'. I'm not sure what you want? They're better than the .44, they're better than the .45, they're better than a .38. By 'underwhelming' do you mean 'don't do enough damage'? Cause, I mean... Maybe it's just my opinion but the detective shouldn't really be trying to do damage to people. Self-defence shouldn't be lethal weapons. Pretty much every single time I've seen a Detective use  the gun it's been excessive force. They have a dumb lethal gun because 'muh ss13 trope', no other reason.

Dingosilver:
By 'underwhelming' I mean the last time I used a taser it took over half the clip to put a single person in any condition to arrest them. Maybe that's been changed since, but if not, I have no clue what you're talking about when you say they're better than the .44. Especially not considering one of the reasons I'm told the .44 is being removed is because they can supposedly be one shot lethals when shooting someone in the head. I find them underwhelming because it took three of the five shots a taser has to make the target no longer in fighting condition, and as far as I can tell one cannot 'reload' them while moving, so if two or more people corner you with a subspace jammer and sunglasses you're left in a position where your weapons are better at being disarmed and used against you than they are when being used on them.

Yoshax:
When was the last time you used any of the things you're actually trying to talk about? Your information is at least six months old, if not so much more.

Dingosilver:
The last time I used a taser was about a month and a half ago, if I'm not mistaken. If .44s have since been nerfed, please feel free to correct me on that one, I'm going entirely off of other's word on that point in particular.

SpadesNeil:
They have SO not been nerfed. They do as much damage as a .357 which does 60 per hit. That's more than a SHOTGUN SLUG.

Dingosilver:
.44s sound more and more ridiculous the more I hear about them, so I can see now why they're being systematically removed, but giving the detectives something for those off cases in which they would need a weapon would be nice. Hell, maybe even just a few rubber magazines for the standardised .45s, if revolvers in general are a little bwoiked right now.

PR #1799, opened 2017-07-02
SilencedMP5A5 Custom Items Changes

Cameron653:
- Slightly changes their robolimb sprites.
- Has their revolver and permit spawn in a box instead of in their backpack.
- Has their custom armor spawn if they are security personnel.
- Gives their robolimbs a tail.
- Slightly changes their custom armor sprite.
- Makes their custom armor cover arms and legs.
- Changes their custom armor's name and description.
- Comments out their now irrelevant conversion kit.

Sprites, names, and descriptions provided by SilencedMP5A5

### DIFF
--- a/code/modules/vore/fluffstuff/custom_boxes_vr.dm
+++ b/code/modules/vore/fluffstuff/custom_boxes_vr.dm
@@ -191,10 +191,8 @@
 
 //SilencedMP5A5:Serdykov Antoz
 /obj/item/weapon/storage/box/fluff/serdykov_antoz
-	name = "Serdy's Gear Box"
+	name = "Serdy's Weapon Box"
 	has_items = list(
-		/obj/item/clothing/suit/armor/vest/wolftaur/serdy,
-		/obj/item/clothing/head/helmet/serdy,
 		/obj/item/fluff/permit/silencedmp5a5,
 		/obj/item/weapon/gun/projectile/revolver/detective/fluff/serdy)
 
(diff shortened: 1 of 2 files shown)


SpadesNeil:
The use of a conversion kit was a deliberate design choice, as they should only have the armor if they have access to the normal armor from their current job.

So like, if they join as a scientist, they won't have the armor.

But I believe the current custom item system can prevent you from spawning with some items based on your access.

Cameron653:
Looking at the code, I think I have an idea of how to make it so they only spawn with the armor if they're a security member. Hasn't been done on any other custom items as far as I can see, but I believe I understand how to get it to work. Going to edit this a bit.

Cameron653:
Alright, there we go. I removed the armor from the box itself and made it so if they have security access then they will spawn with their armor, while keeping their revolver and permit in the box.

List of all accesses here: 
https://github.com/VOREStation/VOREStation/blob/5c7f6f98635fa0814e0e566cc12d419907a8455d/config/example/access%20levels.txt

SpadesNeil:
I need to code in a gun case at some point.

Cameron653:
Anyways, this should be ready for merge if there's no more suggestions.

SpadesNeil:
reeeeeeee conflict

Cameron653:
Ugh, I deleted the fork so I could do pAI things since this PR was completely done and had no merge issues since I can't use branches with github. Let me close this PR and recreate it.

PR #1801, opened 2017-07-03
Replaces carpotoxin in koi with normal toxins

Cyantime:
I didn't realize carpotoxin had real medical uses, now the punishment poison is only that, and also impedrezene.

### DIFF
--- a/code/modules/mob/living/simple_animal/animals/fish_vr.dm
+++ b/code/modules/mob/living/simple_animal/animals/fish_vr.dm
@@ -7,22 +7,25 @@
 /mob/living/simple_animal/fish/koi/poisonous/New()
 	..()
 	create_reagents(60)
-	reagents.add_reagent("carpotoxin", 60)
+	reagents.add_reagent("koipotoxin", 45)
+	reagents.add_reagent("impedrezene", 15)
 
 /mob/living/simple_animal/fish/koi/poisonous/Life()
 	..()
 	var/datum/belly/why = check_belly(src)
-	if(why && prob(20))
+	if(why && prob(10))
 		sting(why.owner)
 
 /mob/living/simple_animal/fish/koi/poisonous/react_to_attack(var/atom/A)
 	if(isliving(A) && Adjacent(A))
 		var/mob/living/M = A
-		visible_message("<span class='warning'>\The [src] flails at [M]!</span>")
+		visible_message("<span class='warning'>\The [src][is_dead()?"'s corpse":""] flails at [M]!</span>")
 		SpinAnimation(7,1)
 		if(prob(75))
 			if(sting(M))
 				to_chat(M, "<span class='warning'>You feel a tiny prick.</span>")
+		if(is_dead())
+			return
 		for(var/i = 1 to 3)
 			var/turf/T = get_step_away(src, M)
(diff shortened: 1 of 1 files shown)


OrbisAnima:
great, now I can add one more poison type for my deadly injections!

code/modules/mob/living/simple_animal/animals/fish_vr.dm
```diff
@@ -34,9 +37,24 @@
...
+	strength = 10
+
 /obj/item/weapon/reagent_containers/food/snacks/carpmeat/koi
-	name = "koi fillet"
-	desc = "A fillet of genetically modified koi meat."
\ No newline at end of file
+	name = "fillet"
+	desc = "A fillet of genetically modified koi meat."
+
+/obj/item/weapon/reagent_containers/food/snacks/carpmeat/koi/New()
```
SpadesNeil: Why not just add a new type of meat instead of koi meat being a thing?

Cyantime: So the koi meats can be used in the same recipes as carp meat, for fish sticks and what not

OrbisAnima:
Couldnt we just use normal toxin instead of a new toxin type? new reagents make interesting chems appear less often in plants and make powergamers panties cream when they notice yet another deadly chem, making their poison mixes one second and a half more lethal.

code/modules/mob/living/simple_animal/animals/fish_vr.dm
SpadesNeil: Because if you're changing this to prevent access to carpotoxin, that negates the purpose, as you get carpotoxin in carp food recipes.

Cyantime: I thought recipes inherited reagents from their ingredients, and I took the carpotoxin out of the meat. let me go test this

SpadesNeil: I'm not sure, as the cooking system has changed a few times. Either way it wouldn't be the authentic versions of the food.

Cyantime: alright, they do inherit reagents but also generate new ones, I'll axe the koi fillets and replace the snowflake poison with something else

Yoshax:
But why?

OrbisAnima:
Players were taking Koi to make hard to get chems, since Koi went each for 60 carpotoxin

Yoshax:
...They were? People already get into trouble for killing pets as is, so y'know.

SpadesNeil:
Yea but this is one of those things that have happened so many times it's looking like a bad addition than problem players.

Yoshax:
I mean... I've been on all day for weeks and I've never seen it happen.

SpadesNeil:
Wait, really? Where the hell am I getting this info then?

Well I know it's something _I'd_ abuse probably, so...

Cyantime:
I mean, the carpotoxin has only been in for like  day at this point. There hasn't been time for people to learn about it yet.

SpadesNeil:
I'm going crazy then. Still I think this is inviting some kind of abuse. Also it makes no sense for koi to have carp toxin.

PR #1807, opened 2017-07-04
Allows PAI's to properly eat people [WIP]

killer653:
- Makes pAIs stomach correctly process people in them.
- Gives pAIs a verb that allows them to eat people while in mobile form. While doing testing, it stayed while they were in card form but had no functionality unless they were in mobile form.
- Makes pAIs spit everything out of their stomach if they exit mobile form/die.

Previously, the only way for a pAI to eat someone was with teleportation/translocators. Even then, the stomach didn't properly digest/absorb them. This makes it so they can properly eat people and, that when the people are in them, they can be digested/absorbed/etc like normal.

DNM. Working on this still.


Cameron653:
Closing it to do this PR differently.

PR #1809, opened 2017-07-04
Fixes Skrell.

RezRaz:
Tajarans were given tentacles instead of Skrell. This fixes that.


Screemonster:
WHOOPS, THAT'S WHAT I GET FOR COPYPASTA

PR #1812, opened 2017-07-05
pAI vore and pAI fat sprites.

Cameron653:
- Makes pAIs stomach correctly process people in them.
- Gives the pAI cat mobile form a fat sprite when standing up and resting if someone is in it.
- Gives pAIs a verb that allows them to eat people while in mobile form. While doing testing, it stayed while they were in card form but had no functionality unless they were in mobile form.
- Makes pAIs spit everything out of their stomach if they exit mobile form/die.

pAI fat cat sprites: https://gyazo.com/99f50f0f6a77a52e975b9a6d806b0c2b
pAI fat cat resting sprites: https://gyazo.com/fc9d29abb0afdc5f8d570b53d289c5ca

Previously, the only way for a pAI to eat someone was with teleportation/translocators. Even then, the stomach didn't properly digest/absorb them. This makes it so they can properly eat people, and, that when the people are in them, they can be digested/absorbed/etc like normal.

Tested it and as so far as I can see there are no issues with it.

The monkey, drone, and mouse pAI mobile forms don't have proper fat sprites yet, but the .dmi is setup so when the fat sprites are made they can simply be put into the .dmi no problem.


SpadesNeil:
Do we prevent folding back up if there's someone in their stomach?

Cameron653:
I've got it setup so that it just spits the person/people out of the pAI if the pAI's clicked.

PR #1818, opened 2017-07-06
tinyfix for outdoors otie

Verkister:
fixed a missing line for "icon_living"
(caused standing mutant oties to revert to the default look.)


SilencedMP5A5:
they also seem to be passive aggressive. they won't attack on sight, like other mobs.

PR #1822, opened 2017-07-06
Adds in body snatcher device

Cameron653:
- Adds in a body snatcher device, based off the sleevemate.
- Adds body snatcher to the protolathe, requiring bio 3, tech 3, and illegal 2 to create.
- Adds body snatcher to trash piles, and makes them as rare as sleevemates.

Recently in the discord there was talk about adding in a body snatcher device, one that would let you swap minds with another person to take over their body, while they have control over your body.

I went ahead and made it so you can swap bodies with: Humans (Including synths), borgs, and the AI.

The body snatcher itself takes 35 seconds of both you and the person it's being used on standing still. After that, it will proceed switch their minds between each other, knock both people out for 10 seconds, leave their vision blurry for 30 seconds, and leave them slurring for 50 seconds as they attempt to get used to their new body.


code/game/objects/items/devices/body_snatcher_vr.dm
```diff
@@ -0,0 +1,66 @@
...
+					var/user_ckey = usr.ckey
+					var/user_ooc_notes = user.ooc_notes
+					admin_attack_log("[usr.ckey] used a Bodysnatcher to swap bodies with [M.ckey]")
+					to_chat(usr,"<span class='notice'>Your minds have been swapped! Have a nice day.</span>")
+					M.ghostize() //If this doesn't happen, it boots them from the server.
+					M.ckey = user_ckey
+					usr.ckey = target_ckey
+					M.computer_id = user_cid
+					usr.computer_id = taget_cid
+					M.lastKnownIP = user_ip
```
Arokha: You should definitely not touch all these manually, and instead look at how the soulcatcher and resleeving transfer minds. Doing this isn't forward compatible if we change things about minds.

Also you should DEFINITELY, under NO CIRCUMSTANCES, edit minds, ever, period. You need to move them.

Cameron653: Ah, alright. Currently looking at resleeving/machines.dm, and I think I see how I can apply the mind transfer to this. Making DNM until I change it.

Cameron653:
Alright. I went ahead and looked at the soul catcher and the resleever. I noticed a proc that would be helpful (transfer_to) but from looking at it, it didn't look like it would work. If I were to have it do M.transfer_to(user) and then user.transfer_to(M) the issue would be that M's mind would transfer into user's body no problem, but it would cause their body's mind to be overwritten, meaning that doing user.transfer_to would result in it copying M's mind back over to the original body.

https://github.com/VOREStation/VOREStation/blob/64eb12fd9c24b929ca27bea181d2b1519c3882be/code/datums/mind.dm#L76
To fix this, I looked at what the proc was doing (.dm file above) and then proceeded to do what it was doing manually, which was to transfer over the mind, then their ckey. Shouldn't be an issue as that portion of the code hasn't been touched in 2+ years.

Additionally, I went and added in "apply_vore_prefs" when the body swapping occurs, so the prey and pred will have the vore organs they're used to and can modify/create more of them to fit the body they're now in.

Cameron653:
Fixed a merge error caused by a previous PR.

Cameron653:
Gah, one second. Trying to fix the trash pile conflict.

Cameron653:
There we go. I think this is ready for merge now if there are no more requested changes, as I believe I've taken care of Arokha's requested changes.

Cameron653:
Removing the DNM since I changed it to swap minds instead of editing the minds, provided Aronai confirms that there is nothing else wrong with this. Everything seems to be working.

SpadesNeil:
@Arokha you think this is good?

Arokha:
>If I were to have it do M.transfer_to(user) and then user.transfer_to(M) the issue would be that M's mind would transfer into user's body no problem, but it would cause their body's mind to be overwritten, meaning that doing user.transfer_to would result in it copying M's mind back over to the original body.

This is easily avoidable with something simple like:
```
var/datum/mind/mob1_mind = mob1.mind
var/datum/mind/mob2_mind = mob2.mind
mob1.mind = null
mob2.mind = null
mob1_mind.transfer_to(mob2)
mob2_mind.transfer_to(mob1)
```

It might seem like I'm being picky, but code forward-compatibility is super important if we want to keep this code working right for the long-term

Cameron653:
Alright. Went ahead and used the code block provided above, compiling it now to test it in game.

Cameron653:
Just tested the code block provided above, modified it a bit to allow OOC_notes to transfer (https://pastebin.com/raw/Kys75nVA) and for some reason it's making the mind of the mob that the user of the device is transferring into null.
https://i.gyazo.com/de6c51ce71359fd75ffb684950658399.png is what it looks like before the mind transfer, and
https://i.gyazo.com/e85b2baa92fedb959c830531bb50df74.png is what it looks like after the mind transfer.

I have no idea why it's doing this, as the code provided _should_ work. @Arokha Any idea as to why it's not working and how to fix it?

SpadesNeil:
I wanna use this in an RP so I wanna see this working soon. c:

Cameron653:
The method currently in this PR _works_ and does the same exact thing that transfer_to does (Transfer the mind & the ckey) just manually. (Although, the method I'm doing in the PR doesn't transfer things like changeling powers like transfer_to does, but that's honestly not used on the server.)

At this point, after fiddling around with the code for hours trying to get the transfer_to proc to work with it, the fact that the transfer_to code is two years old, and the fact that I'm almost sure that using tranfer_to here is literally impossible, I'm honestly not too worried about making certain this is forward compatable.

I'd rather just go back to this code and modify it when(if ever) minds gets changed. This PR has been up for an extremely long time and I'd much rather see it get merged so I can start using it ~~and have my characters bodyswapped then eaten by theirselves.~~ and I can go back and change it in the future if the mind/transfer_to proc ever gets changed.

SpadesNeil:
Leaving it to @Arokha to make the call on this one.

Cameron653:
So... Is this good enough to be merged? The code currently in the PR transfers mind correctly just like transfer_to would but manually. Tested and I can confirm it works.

Fevix:
Could this be added to preferences? I'd much prefer this NOT happen to my characters.

SpadesNeil:
If it doesn't already it should obey disallowing body stealing with normal resleeving.

Cameron653:
@Fevix The bodysnatcher requires both people to be standing next to each other for 35 seconds and gives a warning in chat. There really doesn't need to be a button disallowing it due to the sheer amount of time it takes and the fact that it's designed for scenes. Someone can easily grab & vore you in much less time than the device takes.

@SpadesNeil The device can't be used on bodies that are resleeved without the owner of said body being in it, since resleeved bodies don't have a mind

SpadesNeil:
That isn't what I was referring to.

Cameron653:
@SpadesNeil Oh, you mean that if they have disallow body stealing toggled it shouldn't be able to be used on them? I didn't make it take that into account because the thing with the resleever is that can be used anywhere at any time without the person knowing. With this, it takes 35 seconds the person actually has to be with the standing still with the user of the device for 35 seconds. If someone is going to deliberately pref break they can do other things that are a lot faster than this.

Fevix:
It would make sense. 

IIRC, the error message for attempted body impersonators when they have that option set mentions something about incompatibility. If it's not compatible for resleeving why is it suddenly compatible for bodysnatching? 

Though I do understand that some people may just not want it done _without their permission_, if a body is incompatible it's incompatible.

Cameron653:
That's the problem with adding a check to see if they have body impersonation disallowed. 

If a person is using it to prevent people from using their body without their knowledge, it'd be stupid to completely block them from using the bodysnatcher device in a scene if they're fine with it being used on them with their knowledge. 


The 35 second delay is in there to prevent people from doing it at random to people just to grief.  It gives the person more than enough time to walk away and stop the body switch and let them say no in LOOC.

Arokha:
Can't really review your code if you don't add commits. Hard to review what might be going wrong if I can't actually see your code.

Arokha:
Also I think the pref thing is a little silly because they can ahelp and looc about prefbreak. The other thing is, like Cadence said, something that can happen without you even knowing, and the preference is to prevent that.

Cameron653:
I posted the code in a pastebin above and didn't commit it since it is absolutely broken and won't work, while the current one in the PR works. Here's a pastebin with the fully modified file:
https://pastebin.com/raw/HBZ4K874
As mentioned earlier, the procs are for sure firing, as it swaps their ckey's correctly, but only the first transfer_to in the code block properly moves their mind. The second transfer_to doesn't swap minds properly, resulting in them having NULL for a mind.

Arokha:
Try getting references to the minds before ghostizing. Maybe in some weird cases ghostizing removes the reference to the mind? Unsure. Do you get any runtimes?

Cameron653:
As far as I could see while testing it, there were no runtimes at all.

I messed with the code for hours in every way trying to get transfer_to to work, but seemingly nothing worked except doing it the manual way in the PR. I'm pretty sure the issue is where the transfer_to proc does

```
if(new_character.mind)		
	new_character.mind.current = null
```
which somehow deletes the mind, even though the variable has it beforehand.

I'll try to do the references before ghosting them, but if this doesn't work I'd honestly much rather the PR get merged and I come back and fix the bodysnatcher if the mind code ever gets changed, because it's such a minor thing and the code it's based on hasn't been touched in 2 years, so I'm not afraid of it getting heavily modified.

Arokha:
It's unusual that ghostize doesn't remove current, but you can just manually set 'current' on the mind to null, too. That would fix it if that's the problem.

Cameron653:
The current suggestion worked. Fully tested and everything works as needed.

Ready for merge now.

SpadesNeil:
Awaiting Aro's feedback.

SpadesNeil:
@Arokha MLEM

code/game/objects/items/devices/body_snatcher_vr.dm
```diff
@@ -0,0 +1,70 @@
...
+	item_state = "healthanalyzer"
+	flags = CONDUCT
+	slot_flags = SLOT_BELT
+	w_class = ITEMSIZE_SMALL
+	matter = list(DEFAULT_WALL_MATERIAL = 200)
+	origin_tech = list(TECH_MAGNET = 2, TECH_BIO = 2, TECH_ILLEGAL = 1)
+
+/obj/item/device/bodysnatcher/New()
+	..()
+	flags |= NOBLUDGEON //So borgs don't spark.
```
Arokha: FYI, no big deal, but this is silly. Don't need to change it, just for reference for next time. You already set flags as a variable above, so it's not like it matters what the parent 'device' has. The only reason you'd need to do this is if you were concerned with adding one and not messing with the existing ones, but considering you wipe out all the existing ones by setting flags = CONDUCT, this is kinda irrelevant. I assume these aren't going to spawn every 0.001 seconds and call New() a whole lot though so the overhead of having an additional proc call would be minimal.

PR #1823, opened 2017-07-06
Radio jammer trash frequency nerf

SpadesNeil:
The radio jammer is very obnoxious, and I feel like it should not appear as often as it does. Every shift I have been on since its addition has seen a jammer easily located. It's useful for clever predators, but it shouldn't be this common.


SpadesNeil:
No discussion I guess.

PR #1839, opened 2017-07-08
Tether changes

SpadesNeil:
- Reduced trash piles by probably half. There shouldn't be more than one trash pile in any room.
- Replaced steel beds with comfy beds everywhere.
- Removed syndie ID from trash pile, because there's absolutely no way to detect it by Security, making it unfair for non-antags to find.
- Added sizegun to trash piles. Is technically not contraband, so enjoy.


ArgobargSoup:
Is it too late ta discuss some of the changes?  Because personally I like the current amount of trash piles...  lowering the number of trash piles just seems ta give people less chances of finding something fun ta spice up the shift.

Although some _could_ be moved around... there are some maint rooms that are quite out of the way that don't have a trash pile.

SpadesNeil:
Trust me, there's still plenty. There just aren't 2 - 3 in a single room ever again.

Dankman:
Drop rates were buffed across the board to offset the drop in total piles, yes?

SpadesNeil:
No. The point of this was because there was far too much contraband in circulation per crew member.

PR #1841, opened 2017-07-08
Attempted wilderness performance increase

Cyantime:
Recently, eight more zlevels and several hundred mobs were added to the map, this attempts to lower the performance impact of that.
The mobs on the wilderness maps spawn fully disabled, and are only activated when a player enters the zlevel.
I don't know how to work byond's profiler, so I'm not sure of the actual effect of this, but it can't hurt.


SpadesNeil:
This was badly needed.

SpadesNeil:
Well, have you tested this at least?

Cyantime:
Yes, I at least superficially test all my PRs.

Cyantime:
Comments added

Arokha:
Would be cool if it could re-suspend their life() calls after some timeout and iterating over the players list every X something-ticks to see if it can suspend that z-level again, but maybe you/someone can do that later.

PR #1852, opened 2017-07-11
Fixes teshari smoking sprites

killer653:
Makes the correct sprite show up when teshari smoke.


Draycu:
Oh!  Thank you so much for taking the time to fix this.

Issue #1857, opened 2017-07-13
Mobs don't eat critical condition players


SeigaSeiga:
#### Brief description of the issue
Mobs eat things that are on the ground from resting, stuns, sleeping, etc. Mobs will leave critical condition players alone.

#### What you expected to happen
Vore mobs to eat crit players, especially mobs with the code:
`	vore_pounce_chance = 0 // Beat them into crit before eating.`

#### What actually happened
Mobs with pounce chance zero follow the same rules all mobs do: only eating conscious, knocked down prey.

#### Steps to reproduce
- (Step 1) Be near crit and approach any vore mob, or be at full health near any mob with 	vore_pounce_chance = 0.
- (Step 2) Aggro mob. Behold as you are in crit and not inside belly
- (and so on)

#### Code Revision
Server revision: morelampstuff - 2017-07-13
ddb369d3cd8be0788504328074733398238e2b21
[Testing this on an updated local server with as of yet untouched from base code]
 
#### Anything else you may wish to add:
- I don't know if this is intended behaviour of the general mob vore code and the comments on these mobs with pounce chance zero were made in error, or if mobs should actually be eating players in crit. Either way, people are expecting vore from certain wilderness mobs [dragon, gaslamp], and are coming up disappointed.

SeigaSeiga:
Thinking about it, it's probably because mobs lose interest in you entirely when you go into crit.

SpadesNeil:
I'd give back their pounce chance if https://github.com/VOREStation/VOREStation/issues/541 wasn't a problem AGAIN.

Novacat:
...I think this was fixed a long time ago, even with current vorecode being the mess it is?

PR #1859, opened 2017-07-13
Update gaslamps to be more endo-vore forward in their retaliation.

SeigaSeiga:
Makes gaslamps more inclined to eats. ALWAYS EATS. These changes are partly in reference to #1857, and I have somewhat err'd on the side of making the mob vore aspect people are wandering outside for way more accessible and may have gone overboard with it on making them more vore focused and less attack-focused since now they only attack you with their melee if you have digestion turned off.

Also changed the attacktext to past tense since that's how it should be [otherwise you get 'has thrashes']. Will PR similar fixes for other mob attacktext separately after this one is merged or closed.


SpadesNeil:
I don't like that it's past tense. Past tense makes me cringe.

PR #1860, opened 2017-07-14
Makes some adjustments to prevent temporal distortion while engaging in combat.

SeigaSeiga:


SpadesNeil:
Did you check all the mobs by any chance?

SeigaSeiga:
I did a github search for attacktext and these were the only ones I noticed an issue with. So, yes, afaik.

PR #1867, opened 2017-07-17
Adds a scannable reagent generator

Cameron653:
- Adds in a reagent generator which can scan reagents and produce the scanned reagent over time.

I was recently contacted by ~~A certain fat lizard~~ someone requesting a reagent generator which could scan reagents and then produce them slowly.

Went ahead and whipped this up in ~3 hours, based on the preexisting  reagent generator and using the mech syringe gun as inspiration. 

Notes:
- Generates .2 units per tick.
- Can hold up to 20 units.
- Each tick, 10 nutrition is used to create the .2 units. This means it takes 100 ticks to generate 20 units, resulting in _1000 nutrition required_ for 20 units, making it very costly.
- Can only be used to synthesis reagents that are liquids. This means things like iron, potassium, radium, etc can't be synthesized.

Additionally, I was thinking of putting more safeguards on it (See: Not allowing phoron to be synthesized, as an example.) but decided not to, as someone could play a Vox and generate liquid phoron for some reason.

Currently admin spawn only.
I was personally thinking of having it limited to being a custom item only, or custom item & extremely high R&D level due to simply how powerful it can be if abused.

### DIFF
--- a/code/game/objects/items/weapons/implants/implantreagent_vr.dm
+++ b/code/game/objects/items/weapons/implants/implantreagent_vr.dm
@@ -171,7 +171,7 @@
 			to_chat(user,"<span class='notice'>The container must have one single, pure, liquid reagent inside of it to scan.</span>")
 			return
 		var/datum/reagent/R = container.reagents.get_master_reagent()
-		if(R.reagent_state != 2)
+		if(R.reagent_state != LIQUID)
 			to_chat(user,"<span class='notice'>The reagent inside the beaker must be a liquid.</span>")
 			return
 		var/datum/reagent/R_ID = container.reagents.get_master_reagent_id()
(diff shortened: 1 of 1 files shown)


code/game/objects/items/weapons/implants/implantreagent_vr.dm
```diff
@@ -96,3 +96,90 @@
...
+	for(var/I in src.contents)
+		if(istype(I, /obj/item/weapon/implant/reagent_generator/scannable))
+			rimplant = I
+			break
+	if(rimplant)
+		if(!container.reagents) //This should only apply to things that can't hold reagents.
+			to_chat(user,"<span class='notice'>The container must have one single, pure, liquid reagent inside of it to scan.</span>")
+			return
+		var/datum/reagent/R = container.reagents.get_master_reagent()
+		if(R.reagent_state != 2)
```
Arokha: Use the constants used elsewhere and not just the number please. I have no idea what '2' is.

Cameron653:
Changed the != 2 to != LIQUID.

SpadesNeil:
Not a mechanic I want to see.

PR #1871, opened 2017-07-17
Add new tail: Big snake tail

ShadowLarkens:
`snake tail, colorable`


SpadesNeil:
Ok this is really neat.

PR #1874, opened 2017-07-18
Renames "second-hand NIF" to "bootleg NIF"

Cyantime:
Second-hand implies it was already used by someone, and from what I can tell NIFs bind permanently to whoever they're put into first. The description also already fits it being some sort of bootleg or shoddy replica.

I don't know who the loreperson for NIFs is, if the original name is better please tell me.


Arokha:
I'd be the loreperson and this is 👍

PR #1875, opened 2017-07-18
Adds the icons for the bear hat.

Ketrai:
Approved here : https://forum.vore-station.net/viewtopic.php?f=26&t=956

Don't worry about there being three commits, it's just me being a derp and getting confused over the fact that there is an on mob file, and a custom clothes file. Either people don't put stuff in the on mob file when they're supposed to, or they're already doing it right. (I've seen plenty of on mob files in the file, so I guess it's fine.)

Also arokha makes me think that I'm supposed to rather just attach an icon file to my pull request, but heck, why not just add it to the file if i'm able to!

ALSO: I do realize I made two pull requests now, and i only realized after making this one I made the organized error of directly committing rather than making the code revisions in my forked branch, so sorry for that.


SpadesNeil:
conflict reeeeeeeeeeee

Ketrai:
well, can someone at least explain to me what might be conflicting here? I just added the icons to a forked branch to the dmi file, and then uploaded it, then applied to it here as a pull request.

PR #1876, opened 2017-07-18
Scion ship names.

Benemuel:
Names for Scion ships for the busy space organizations.

### DIFF
--- a/code/modules/busy_space/organizations_vr.dm
+++ b/code/modules/busy_space/organizations_vr.dm
@@ -195,7 +195,7 @@
 					"Star of Tiamat",
 					"Mostly Harmless",
 					"Hero of the Revolution",
-					"Dawnstar",D
+					"Dawnstar",
 					"Freedom",
 					"Fiery Justice",
 					"Bulwark of the Free",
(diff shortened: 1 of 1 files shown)


code/modules/busy_space/organizations_vr.dm
```diff
@@ -224,3 +224,32 @@
 						"a settlement needing our help",
 						"Forward Base Sigma-Alpha in ArCon space")
 
+/datum/lore/organization/gov/Scion
+    name = "Scion Imperium"
+    short_name = "Scion"
+    desc = "A Imperium of Desert Elves, slowly expanding and exploring the galaxy on there own terms instead of those of others./
```
Screemonster: A few typos, mostly there/there mixups in the description. Should read "on their own terms", "Their ships are..." and "Despite their lack of ships, their ground forces..."
History section misspells "calendar".

Benemuel: And this is why autocorrect is useful for me >_>

Benemuel:
Also I can't tell for the life of me why it is failing the check.

SpadesNeil:
If the Scion is just one planet with a couple of military ships I don't see the point in adding this, to be honest. I'd rather not, or everyone will want their own custom races and factions to be included.

Benemuel:
Righto.

PR #1878, opened 2017-07-18
FIXES THE CATWALK INVENTORY BS

Verkister:
Things that are not on the turf floor should no longer drop through catwalks. YOU CAN NOW SAFELY BAG YOUR CRAP WITHOUT LOSING IT TO THE VOID THAT IS THE FLOOR DOWNSTAIRS.


SpadesNeil:
Should probably just make this change upstream.

Verkister:
Their file didn't even have the catwalk stuff in it so dunno if it's just map specific.

Verkister:
Yeah just checked this whole file is like 1/3 our custom code mostly by Leshana.

PR #1879, opened 2017-07-18
Adds object-specific effects for running into them while confused

Cyantime:
Now you can drunkenly flip over the railings like you've always dreamed of.

### DIFF
--- a/code/game/objects/stumble_into_vr.dm
+++ b/code/game/objects/stumble_into_vr.dm
@@ -1,0 +1,126 @@
+/atom/proc/stumble_into(mob/living/M)
+	playsound(get_turf(M), "punch", 25, 1, -1)
+	visible_message("<span class='warning'>[M] [pick("ran", "slammed")] into \the [src]!</span>")
+	to_chat(M, "<span class='warning'>You just [pick("ran", "slammed")] into \the [src]!</span>")
+	M.apply_damage(5, BRUTE)
+	M.Weaken(2)
+
+/obj/structure/table/stumble_into(mob/living/M)
+	var/obj/occupied = turf_is_crowded()
+	if(occupied)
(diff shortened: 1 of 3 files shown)


SilverTalismen:
I guess afew people will *Rail*y hate these changes, but they sound fun!~

Cameron653:
This would be a great thing to push upstream.

code/game/objects/stumble_into_vr.dm
```diff
@@ -0,0 +1,126 @@
...
+/obj/machinery/space_heater/stumble_into(mob/living/M)
+	..()
+	if(on)
+		M.apply_damage(10, BURN)
+		M.emote("scream")
+
+/obj/machinery/suit_storage_unit/stumble_into(mob/living/M)
+	if(!ishuman(M) || !isopen || !ispowered || isbroken || OCCUPANT || HELMET || SUIT)
+		return ..()
+	playsound(get_turf(src), 'sound/effects/clang.ogg', 25, 1, -1)
```
Arokha: Just to make sure, you CAN get out of these, right? Is it that they have to be locked to keep people in? Otherwise I'd be worried about people just getting stuck and needing to ahelp, which is funny, but also annoying. XD

Cyantime: It doesn't lock them in, it does the same thing as climbing in manually

Arokha: I've never shoved anyone in either, so I didn't know if it was like. ... Well whatever. Anyway!

Cyantime: No one ever uses the suit storages for hiding in, and they're incredibly sparse on the new map anyway.

PR #1884, opened 2017-07-20
Deletes Teshari Duplication

Cameron653:
Teshari were duplicated twice in station_vr when they only needed to be there once. 
- Deleted the duplicate.


Cameron653:
This is just minor cleanup that changes literally nothing, so just going to go ahead and merge this.

PR #1885, opened 2017-07-20
Adds autohiss traits.

Cameron653:
Adds traits that allow you to hiss like a unathi or a tajaran.

- Adds Unathi Autohiss Trait
- Adds Tajaran Autohiss Trait

Might want to increase the amount of traits select able by 1 sometime in the future.


SpadesNeil:
Tajaran and Unathi won't have this by default, right?

Cameron653:
As far as I know, no. This is a trait that people can pick if they want to roll their S,X, or R's

Yoshax:
Does this force you to do the thing, or allow you to use the toggle?

Cameron653:
It allows you to use the toggle.

PR #1886, opened 2017-07-20
Allows borgs to use custom stomachs

Cameron653:
Allows borgs to eat people and use custom stomachs, along with fixing how the game handles it when someone ends up in a borg's stomach.

- Makes it so the game properly processes things in a borg's stomach (Previously something could get in via teleportation or whatnot and the game would just hold them there and not actually process them)
- Makes it so borgs have a robot nom verb that allows them to eat people to put them in custom stomachs instead of using the sleeper.

### DIFF
--- a/code/modules/mob/living/silicon/robot/life.dm
+++ b/code/modules/mob/living/silicon/robot/life.dm
@@ -21,7 +21,7 @@
 		process_killswitch()
 		process_locks()
 		process_queued_alarms()
-	handle_internal_content() //VOREStation Edit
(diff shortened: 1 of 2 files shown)


kcin2001:
this will encourage me to play more borg thats for sure

icowom:
As  a synthtetics and silicon heavy player, I wholly support this~ And would definitely make me wanna play them more

Cameron653:
Did some more testing on this and I can't seem to find any bugs at all. This should be ready to merge.

PR #1888, opened 2017-07-20
fixing some messed up medical hardsuit helmet sprites

kcin2001:


SpadesNeil:
What was messed up about them? Also, this fix should go to Polaris.

kcin2001:
i didnt think about pushing to polaris but on 2 of the med rig helmets the east and west sprites were flipped

SpadesNeil:
What are the seromi again? Those aren't our thing, are they? Because if those sprites are fucked up, then Polaris needs this fix.

Cyantime:
Seromi are what teshari are called in the code, it's their original name

SpadesNeil:
Okay yea, don't fix this here. Fix this in Polaris.

PR #1889, opened 2017-07-21
Disable all shell() exec prox for extra security.

Leshana:
We do not use python or nodejs scripts. Our discord bot makes HTTP requests.  Therefore we can entirely disable this code.  Even if we ever want to start using these procs for discord integration, we would pipe it through our HTTP discord proxy, not use shell().
With all instances of shell() disabled, we can't be attacked by it.


OrbisAnima:
The bot then proceeds to execute an eval() to pass the message

MrStonedOne:
https://github.com/tgstation/tgstation/issues/29420

PR #1895, opened 2017-07-23
Design fix

Cameron653:
Accidentally had the bodysnatcher design listed as /sleevemate from when it was copy and pasted, preventing it from being created. This fixes it

- Allows bodysnatcher to be correctly made by R&D


Cameron653:
Extremely simple bugfix. Passed Travis and doesn't actually change anything other than a quick one line bug fix. Merging.

PR #1900, opened 2017-07-25
Update alienwhitelist.txt

Rixunie:


SpadesNeil:
It's Gutterband, not just gutter. :P Fixed it.

Cyantime:
The language is "gutter", if you search for "gutterband" the only file that comes up is the whitelist file.

SpadesNeil:
In that case there's two screwed up.

SpadesNeil:
Fixed it.

PR #1902, opened 2017-07-25
Ivy Baladeva ID Fix

DarkWolfInsanity:
Replaces Ivy Baladeva's ID with a reprinter as to keep access rights and still change the icon.


DarkWolfInsanity:
It says failed, but looking at the log, it appears to have just timed out on the check.

Cameron653:
code/modules/vore/fluffstuff/custom_boxes_vr.dm:141:error: /obj/item/weapon/card/id/fluff/ivyholoid: undefined type path

DarkWolfInsanity:
Should be fixed now.

PR #1906, opened 2017-07-25
Workaround for resist transfers putting people into non-connected bellies

### LINKED ISSUE #763: Transferlocation stomach limbo bug
Deleting a transfer location stomach then transferring prey causes prey to be stuck in limbo
Delete transfer location stomach, have it automatically removed as a transferlocation from all other stomachs
#### What actually happened
Didn't remove it as a transferlocation. When prey was transferred it trapped them in limbo
#### Steps to reproduce
Make a secondary stomach
Select secondary stomach as a transfer location
Delete secondary stomach
Eat prey and have them struggle while special interactions are toggled and transfer chance set to 100
Watch as prey is removed from primary stomach but trapped in limbo as secondary stomach doesn't exist
#### Additional info:
 - **Server Revision**: https://github.com/VOREStation/VOREStation/commit/3c0df7dbc08319ad4b7298a546db4de0fc17d5bb
 - **Anything else you may wish to add**: SEE COMMENTS BELOW FOR HOW TO REPRODUCE.
 Bug is from PR #758.

Cyantime:
The actual solution would be saving transferlocation as the belly name rather then the belly datum itself, but I'm pretty sure that would involve touching save files and I'm terrified of that.
Should fix #763

### DIFF
--- a/code/modules/vore/eating/belly_vr.dm
+++ b/code/modules/vore/eating/belly_vr.dm
@@ -460,6 +460,27 @@
 				return
 
 		else if(prob(transferchance) && istype(transferlocation)) //Next, let's have it see if they end up getting into an even bigger mess then when they started.
+			var/location_found = 0
+			var/name_found = 0
+			for(var/I in owner.vore_organs)
+				var/datum/belly/B = owner.vore_organs[I]
+				if(B == transferlocation)
+					location_found = 1
+					break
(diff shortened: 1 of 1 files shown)


Cameron653:
Your method is wild, but it looks like it works codewise. :+1:

PR #1910, opened 2017-07-26
Slight sergal overhauls

Verkister:
-Apart from ears basically everything is new.
-Full body markings updated as well.
-Added a dual colorable new sergal tail.
-Added a new full body markings set that's basically just some vague general use simple body shading to make up for coloring induced contrast loss.
-Further tweaked the sergal parts to bring up more definition.
-Overhauled sergal ears.
-converted the sergal hardsuits to fit for the new tail.
-Improved the body toning markings, now stronger and more seamless.

c593fc585e589e26296ab6ae7d157688


SpadesNeil:
This should not be merged unless sergal hardsuits are also updated.

Jakksergal:
There's no need to update the hardsuit sprites immediately. Equipping the hardsuits removes the tail sprite all together.

Verkister:
Annd done once again.

SpadesNeil:
Yes but it'll never get done if we don't do it now, and it'll always look dumb.

Luckily it's done now. Gonna review this more thoroughly soon.

PR #1912, opened 2017-07-28
Stops ion storm from blocking PDA messages

Cameron653:
Ion storm made it so the PDA was pretty much unusable due to having common terms and space blocked by the pda messenger.

This makes it so ion storms don't mess with PDA's anymore.


Cyantime:
This isn't a bug, one of the items added to the spam filter is " ", removing that will make it not block anything with a space in it.

Cameron653:
The PDA is pretty much as commonly used at LOOC at this point, so having ion storms make it completely impossible for people to say common things like "hi" or "hello" so it's better to just remove all of it, since it adds no actual gameplay value and just exists to piss people off.

PR #1913, opened 2017-07-28
Size Update

Cameron653:
Size update.
Allows people to chose their size between 25-200% instead of being forced to be 25/50/100/150/200
This gives people more freedom to make their character, say, 90% to show 5'8 and 75% to show 3'8 or something like that.

CHANGES (Everything listed was changed to allow 25-200% to be selectable unless stated otherwise):
- Changes sizegun
- Changes character select size
- Changes bluespace jumpsuit
- Changes size NIF
- Changes size virus
- Changes currently commented out size_change ooc proc
- Changes resleeve design size selection
- Changes resize chemicals to change size by 1% per tick. 1 unit lasts 100 ticks.  This allows for more accurate size changes.
- Allows you to see if you can pick someone up upon examine

Notes:
Left constraints, as they were useful with things like chemicals and a lot of vore mobs apparently use them.
Please let me know of any typos or issues.

Also, special thanks to Aronai for letting me know that IsInRange is a thing.


SpadesNeil:
I have concerns this may screw some things up with the size changing code, as player scale being certain values has always been attached to other snippets of code including but not limited to the way vore mobs determine belly size.

This could break things and needs extensive testing and reviewing.

RezRaz:
This sounds great.

Cameron653:
I don't believe that it will cause issues with vore mobs, at the very least.
The bulk of the mob-vore code (from what I can see) is handled here: 
https://github.com/VOREStation/VOREStation/blob/master/code/modules/mob/living/simple_animal/simple_animal_vr.dm#L47-L85
and most mobs are set to eat anything that is between RESIZE_HUGE (2.0) and RESIZE_TINY (.25).

On the topic of mob elly size, it should be perfectly fine, as they do 
new_fullness = round(new_fullness, 1) //Located on L53 in the .dm above
which makes sure that it rounds to the closest whole number, so it doesn't matter if a mob is .75 or .9, it'll round up to 1, and .49 and below will round to 0.

SpadesNeil:
Oh good.

code/modules/client/preference_setup/vore/02_size.dm
```diff
@@ -30,9 +30,7 @@
 	S["weight_loss"]		<< pref.weight_loss
 
 /datum/category_item/player_setup_item/vore/size/sanitize_character()
-	var/valid_scales = list(RESIZE_HUGE, RESIZE_BIG, RESIZE_NORMAL, RESIZE_SMALL, RESIZE_TINY);
-	pref.size_multiplier	= sanitize_inlist(pref.size_multiplier, valid_scales, initial(pref.size_multiplier))
-
+	pref.size_multiplier= pref.size_multiplier
```
SpadesNeil: tabulation makes me wah but doesn't actually matter

Cameron653: ...Looking at this, it actually looks pretty bad. Let me fix it real quick.

SpadesNeil:
This been tested yet? :U

Cameron653:
Just re-tested everything.

Character select: Works
Bluespace jumpsuit: Works
Size gun: Works
NIF: Works
Size virus: Works as expected. Picks random numbers between 25 and 200%
Resleeve designer: Works, gives an approximation of size (If size is .9, it'll say normal. If it's 1.57, it'll say big. Didn't make it say the % because I have no idea how to deal with .tmpl files. See: designer.dm andbody_designer.tmpl)
Size Examine code: Works

### Mobs eating

Tested w/ a feral otie, which has a stomach capacity of 1 (meaning will eat up to 100%)
Would eat: 
One .75 person (Rounded up to 1.)
Two .25 people (Resulted in .5 which rounded up to 1.)
Two .49 people. (Resulted in .98 which rounded up to 1.)
One .98 person (Rounded up to 1.)

I can't think of anything else that could be bugged by this change. Should be ready for merge.

PR #1861, opened 2017-07-14
Engineering Overhaul

Jakksergal:
An entire redesign of Engineering.

Key features: 
- Engine is now as far away from the hallway as physically possible. Radiation exposure is impossible even with an overcharged SM
- A plethora of air pump ports and scrubber ports were added. Engineering no longer relies on cargo to fill up breaches.
- A new construction area was added where Engineering EVA was. It's multi-z with a balcony and a view!
- Engineering EVA is now recessed inside engineering as it should be. It's directly connected to the workshop for rapid response.
- Introduces a new maintenance area for scenes, RP, and antagery.

Notes: 
- A new map will need to be created for crew monitoring consoles, cameras, and directories. I wouldn't know how to do these so I'll leave the task to someone else.

(Almost) Full image: 
z1 - https://gyazo.com/7d5ac27a5885f191a29310d4dda0b265
z2 - https://gyazo.com/3c06bbe016e42e684c25d99f76dcae85

PR #1929, opened 2017-08-01
Engineering Overhaul 2: Panic in the Repo.

Jakksergal:
Fixes absolutely everything even though there were no problems in the first place. *Cough*

An entire redesign of Engineering.

Key features: 
- Engine is now as far away from the hallway as physically possible. Radiation exposure is impossible even with an overcharged SM
- A plethora of air pump ports and scrubber ports were added. Engineering no longer relies on cargo to fill up breaches.
- A new construction area was added where Engineering EVA was. It's multi-z with a balcony and a view!
- Engineering EVA is now recessed inside engineering as it should be. It's directly connected to the workshop for rapid response.
- Introduces a new maintenance area for scenes, RP, and antagery.

Notes: 
- A new map will need to be created for crew monitoring consoles, cameras, and directories. I wouldn't know how to do these so I'll leave the task to someone else.

Images:
https://gyazo.com/ff17b2dd14489529817d3ffb16fa7fde
https://gyazo.com/f6cc1329d17ef58db48c4778a389be9f
https://gyazo.com/f21327953925c6f503865b98c162f5c5


Jakksergal:
This PR is a pseudo-continuation of https://github.com/VOREStation/VOREStation/pull/1861

Virgo113:
Just at a glance, this has my seal of approval. I'd have to playtest in order to see if any kinks that need ironing crop up, but from what I can see in the images, everything seems logical.
The only thing that slightly confuses me is that waste cooling setup for the engine room...
imagen

 Why put the pump there?

Looking good in every other regard, from my quick five minute review at least.

HawkerTheGreat:
Still dying to have this pushed, I feel like its well done... Rather than re-posting, I shall grab my previous comment. 

htg

ding

(And I don't think Silver will mind if I move this here aswell.)
(These are both from the previous merge discussion.)

Cyantime:
More then enough power can be produced with the current setup without leaking radiaiton anywhere. Blame engineers who need big numbers for your irradiation, not the map.

Jakksergal:
@Virgo113 That was a mistake from a lazy copy-pasta. 
@Cyantime Engineering at its core is about using the tools we're given to make something new and toy with the mechanics of the game. If every engineer setup the engine the exact same way, there'd be no point in even having a setup in the first place. 

Also, I would rather have a large lower level and small upper level than have a medium lower level and a really oblong upper level. The brig just takes up too much of my usable space to do much other than that. Still though, I'll see what I can do.

SpadesNeil:
Are there any final opinions on this before I merge it? Gonna bother engineering again.

Floorface:
@Jakksergal 
You said: "Engineering at its core is about using the tools we're given to make something new and toy with the mechanics of the game. If every engineer setup the engine the exact same way, there'd be no point in even having a setup in the first place."

However, creating a situation where every set up is just as good as another means there is no point in setting up the engine in different ways. Experimentation will not be needed, where almost every set up will be just as safe and just as effective as every other set up. Engineers already have the freedom to set up the engine however they want, but they should have to use their judgement in deciding which solution is best.

HawkerTheGreat:
As I mentioned before hand... the radiation leak is not an issue with over-setting the engine, its an issue with the way engineering is structured. 

Also, just because an engine only needs to be protected from a set amount of radiation doesn't mean it shouldn't be protected from a higher amounts just in case... This is done in real life all the time, for safety - even though there is never a time where it would be logical to do so. Its done anyways, because there isnt a reason _/not/_ to be **safe**.

PR #1934, opened 2017-08-04
Updates the fitness vendor sprite to match perspective

Yoshax:


Cyantime:
3/4 perspective is bad and I hate that more servers are trying to change to it.  It only ever looks good for things that are directly south of a wall, and often not even then.

SpadesNeil:
Looks good to me. 👍

PR #1935, opened 2017-08-04
makes colorable bat wings able to make more colors 

kcin2001:
OLD
image

NEW
image


Screemonster:
I reeeaaaalllllly need to go through and fix the contrast on basically all the "colorable" items but ugh effort

PR #1936, opened 2017-08-06
Reduces effectiveness of tabling as opening combat move.

Cyantime:
Tabling in its current form is an effectively instant stun with no real counters other then being about four tiles from any tables at all times, which can be extended indefinitely with minimal effort. This degenerates much combat into "Grab, mash z, click table, push to ground, table again occasionally to reapply stun".

Tabling now takes a short time to happen, 1/2 second with some additional time added depending on the melee defence of the victim's chest armor. People wearing active magboots cannot be tabled.
Armor also decreases the length of the stun received.

### DIFF
--- a/code/modules/tables/interactions.dm
+++ b/code/modules/tables/interactions.dm
@@ -106,9 +106,11 @@
 				else
 					user << "<span class='danger'>You need a better grip to do that!</span>"
 					return
-			else
-				G.affecting.loc = src.loc
-				G.affecting.Weaken(5)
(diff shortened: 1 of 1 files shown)


SpadesNeil:
Should be sent to Polaris rather than just us, I think.

AeroWing:
I don't think Actual Table is going to appreciate being nerfed.

But humor aside this is something that needs to be done, thanks for putting the effort in.

Cyantime:
PR was put up on polaris a while ago, I should probably have closed this then.

PR #1937, opened 2017-08-06
Modifies trash pile loot pool

Cyantime:


Keekenox:
I assume this is the one that adds gas masks and also voice changers to the loot pool?

PR #1941, opened 2017-08-07
Makes species & taur egg correctly change custom species

Cameron653:


Cameron653:
This is like the most minor bugfix in the history of ever. Merging.

SpadesNeil:
inb4 causes massive organ rejection

Cameron653:
If this causes massive organ rejection I'm deleting BYOND.

PR #1946, opened 2017-08-08
Engineering fixes and improvements.

Jakksergal:
- Fixes a critical issue with the engine's default pipe layout.

- Fixes CE's office not having privacy.

- Adjusts backup atmos' pipe layout to improve functionality.

- Adds bits.

- Adds bobs.

If there's anything else wrong, let me know here.


SpadesNeil:
>Jakksergal wants to merge 2 commits into VOREStation:release from Jakksergal:master 

NO

Jakksergal:
I don't even. It's changed but shouldn't a critical fix be pushed to the front?

SpadesNeil:
You should **never** merge into release directly because:
- You cause conflicts between master and release.
- You break the server's ability to update correctly from release.
- Your changes get deleted next merge to release because that's the only way to fix what you broke.

PR #1947, opened 2017-08-08
Blurry non-default sprites

Cameron653:
Not sure what to call this one.
Disabling this makes sprites appear less sharp and more blurry. Works on smaller sprites but makes Macros look kinda... Off.

Leaving this here for people to discuss and vote.

How it currently is: https://i.imgur.com/vkPrcPM.png?1
How it will look if this is merged: http://i.imgur.com/Q6Be1qE.png
Top part of the above screenshot: 150 160 170 180 190 200

### DIFF
--- a/code/modules/vore/eating/living_vr.dm
+++ b/code/modules/vore/eating/living_vr.dm
@@ -25,7 +25,7 @@
 		return 1
 	M.verbs += /mob/living/proc/insidePanel
 
-	M.appearance_flags |= PIXEL_SCALE
+	//M.appearance_flags |= PIXEL_SCALE //This makes sprites not blurry or fuzzy, and makes them sharp. Macros look good with it but can cause issues.
 
 	//Tries to load prefs if a client is present otherwise gives freebie stomach
 	if(!M.vore_organs || !M.vore_organs.len)


Natje0:
As someone who plays macro/big a lot... I can't support this change.

Dingosilver:
Please no blurry macros...

Dhaeleena:
Have you all seen how every other size looks like?
Yea your macro looks good but now there are 170 new sizes
Any size that isnt 25 100 150 or 200 looks either with NO eyes, With one Less eye, With different sized eyes and in some cases even with limbs of different sizes

Dingosilver:
I'm guessing there's no way to have macros not be blurred and micros be blurred? Blurring does little to solve the problem, other than make it harder to make out.

Cameron653:
Updated the PR desc.
How it currently is: https://i.imgur.com/vkPrcPM.png?1
How it will look if this is merged: http://i.imgur.com/Q6Be1qE.png

Ketrai:
I vote not to merge this.

Renkindle:
I like the fix, since the weird misshapen eyes on some of the larger sizes in the original pic aren't there.

Cameron653:
Honestly, I like how the large macros look with the sharp, but... 

Yeah, I'm going to have to go with making it blurry. The misshapen eyes are a pretty big issue, and the fact eyes are missing from 62-65% is an issue.

Ketrai:
but with the blurr ices are still missing on micros?

Dingosilver:
Yeah, they still seem to be mishappen and missing things, but are just harder to view than before.

Cameron653:
The eyes are still missing in the 60% range and one eye is missing in a few others, but it's _much_ harder to notice on micros, and causes the misshapen eyes to blend together and be less noticeable on the macros.

Ketrai:
hm, Idk, then i'd rather be neutral, or at least see ingame how people react

Dingosilver:
Frankly if it's at the point where the eyes need to blend together to give an illusion of something that is relatively easy to notice upon even relatively close inspection, I'd rather not mess up basically all the macro sprites just to fix something that could be fixed by picking a different size.

Ketrai:
true, I basically had to come at 84% since other sizes derped, you might just wanna experiment.

Vorrarkul:
IMO:
25-64: Too small to affect meaningfully
65-129: Benefits from the change
130-200: Blurry mess, only getting worse the bigger you get.

This all depends on if you want macros or near-normal sizes to look better. Personally, my vote is on sharp, not blurry.

Dhaeleena:
Actually micro is more visible. And it also depends on your screen. If you have a 4k screen then well, there is nothing we can do if that person decides to play on stretch to fit. I play on 32x32 and everything looks perfect, I would say the fur patherns on everything will look a lot better with fuzzy smooth

Dingosilver:
Ehh, I've always had trouble focusing on things that are blurred for too long, so I can't deny I'd be biased because of that. I just feel like, if adjusting resolution to make everything so small you don't notice the blur is the solution, what's the commotion about a few things looking slightly mishappen?

Dhaeleena:
People dont like having the faces of their characters deformed specially when they get to stare at them all day. Atleast when its blurry it still looks more like a face

Dingosilver:
The blurry one looks even less like a face, in my opinion. Looks more like a paint splotch.

Ketrai:
I'm going to shift back to sharp, it really is just a case of shifting the numbers around, and getting something you like.

Vorrarkul:
@Dhaeleena So, we may disagree on micros, but I didn't see anything in your post disagreeing that this turns macros into blurry messes, especially the biggest of them?

HawkerTheGreat:
It looks good on the normal sized sprites.
> 65-129: Benefits from the change

However on the rest of the sprites, it makes them look even worse than the one eye'd amalgamations they originally were as micros or huge eye'd monsters they might turn into as Macros. -

The problem with this that you will never fix is due to the fact of byond limitations, in order to shrink a sprite, you have to get rid of some pixels and when you do... single pixel items like eyes tend to either vanish or become freakishly large.

And I don't think there is a workaround to shrinking them without getting rid of pixels.

As it stands, I **DISAGREE** with merging this.

SpadesNeil:
Honestly I think we should just not merge this. It's how it was before and it was pretty fugly I guess. Gonna close this.

Also micros get even less visible which is awful.

PR #1951, opened 2017-08-08
Comments out an assert check that is unneeded.

Cameron653:
It kept screaming in chat that it failed every time the resize proc occured. This shuts it up.


Cameron653:
Can I go ahead and merge this?

Issue #1959, opened 2017-08-09
Open space item ghosting and duplication.


Jakksergal:
#### Brief description of the issue
Removing items from your person over open space creates a ghostly item on the bottom-most z-level exposed by that open space.

#### What you expected to happen
Pretty sure you know.

#### What actually happened
The issue described.

#### Steps to reproduce
- Wear an item standing over normal turf.
- Move to a catwalk/lattice and remove an item. Keep it in your hand. 
- Move to the bottom-most z-level.
- Get spooked. Picking up that item in your other hand removes the first item you just took off.
I don't know how I did it, but you can completely duplicate items that change states like welding goggles and such.

#### Code Revision
- cf3b3a66adc7c2451ef5eef91b06e88851f6d7b4

SpadesNeil:
Could have sworn we fixed this. Did we?

Jakksergal:
I'll test again sometime in the future.

PR #1966, opened 2017-08-09
Yet more engineering fixes and adjustments.

Jakksergal:
- Adjusts the construction area's area to make doors powered from the outside.
- Attempts to fix CE's access to secure storage. If the fix doesn't work it wont break anything for the other engineers.
- Adds missing air alarms to maintenance.
- Adds a drink vendor and borg charger to the lobby.
- Minor aesthetic changes.
- Adds an electrical shaft with a ladder and an upwards wire.
- Removes the catwalks in the locker area and replaces it with normal turf until https://github.com/VOREStation/VOREStation/issues/1959 is fixed.

Let this one sit for a day just in case something else comes up.


SpadesNeil:
`access = list()` is "you need all of these to get through this door"
`access_one = list()` is "you need any _one_ of these to get through this door.

Mix and match as necessary.

Jakksergal:
I know what they're for but some reason `req_access = list(11)` and `req_one_access = list(10)` on the same door makes the CE not able to enter it. It doesn't make any sense to me, this is just attempting to fix the issue.

SpadesNeil:
That doesn't make any sense. Although I was mistaken earlier. req_access overrides req_one_access I think.

https://github.com/VOREStation/VOREStation/blob/master/code/game/jobs/job/engineering.dm
https://github.com/VOREStation/VOREStation/blob/master/code/game/jobs/access_datum.dm

Jakksergal:
That doesn't seem to be the case. Players reported that removing secure tech storage's airlock control board and reconstructing the door with the same board allows them to access it with CE clearance. It turns out that the door starts with the vars I stated above but reconstructing it sets `req_one_access` to null. Although I'm frankly not certain what I should set that var in DM to to achieve the same result. 
Is it `null`, `list(null)`, `list()`, `{}` or what?

SpadesNeil:
None of the above.

`req_one_access = newlist()`

PR #1969, opened 2017-08-10
Micro Crush Adjustments

Cameron653:
- Increases max end damage by being stepped on
- Decreases low end damage by being stepped on
- Spreads damage across all body parts
- Gurantees damage will heal over time, provided they were only stepped on once, even if on walk intent. They may be able to be stepped on more than one time and still heal up naturally, provided RNG is nice.
- fixes a minor typo. 

Running over prey will now do 5 damage min, 26.25 damage max, depending on size & RNG. Max means the pred is 2.0 and the prey is .25, and they hit the max RNG number.

Walking over prey will now do 17.5 damage min, 91.875 damage max. This sounds like a lot, but it's spread out over 10 limbs (both arms, both legs, both hands, both feet, head, and chest.) and the max damage requires, again, a 2.0 size person and a .25 person, along with bad RNG. This means that one walk will heal up naturally over time, guaranteed, and break no bones, even if they hit the max damage.

I fiddled around with the numbers somewhat and found this was the best


SpadesNeil:
I'm in approval of this change, frankly.

PR #1971, opened 2017-08-10
Makes wings their own seperate entity on char setup

Cameron653:
- Makes wings their own seperate entity on character setup
- Leaves custom wings&tails people have under the tails, since they're going to be using both at once anyways.
- Only touches 2 normal .dm files (dna2 and dna2_helpers). Made sure to leave a //VOREStation Edit comment on the lines that were edited, or just stuck it in the areas where there was a //VOREStation Add Begin and //VOREStation Add End. I did as much as possible to touch as little of the not _vr .dm files, but these two were quite essential to do. If there's ever a conflict, it can be easily edited up a layer by changing a few numbers around.
- Adds a vflap feature, enabling flapping wings in the future. (Enabled, but no wings has this feature.)
- Adds a flapping variable to humanoids, allowing them to flap their wings around with the vflap.
- ~~Sucked the life out of my soul because it took 5 hours~~
- Adds Roiz's fat lizard tail http://i.imgur.com/MUseeMf.png http://i.imgur.com/BHkhIE8.png
- I pretty much copied the tail code and edited it for wings with a few touches here and there.
- Fixes people spawning invisible due to a null scale http://i.imgur.com/Qnyx2KJ.png

 Tested, and I can confirm it works for tails. The wings are setup so they'll be above the tail layer, so tails won't sit above the wings.


SpadesNeil:
You know what'd be neat but totally unfair? The ability to fly up Z-levels. :)

HawkerTheGreat:
Eyy Cameron, You get a cookie. - I've been waiting so long for this. Thx. - I think we should be able to fly wherever we choose, **#spacewings**

Cameron653:
Ugh, I accidentally left a Roiz's WIP tailin taurs_vr.

Since it isn't actually used, and is still a WIP, isn't a big issue.

Cameron653:
I know including bugfixes in feature PR's is bad, but I saw the solution to solve the invis bug and was too lazy to open another branch.

PR #1973, opened 2017-08-11
Removes redspace tech from sizegun

Cameron653:
All it did was cause runtimes.


Cameron653:
Just a small fix that stops a runtime due to a non-existing tech tree.

PR #1978, opened 2017-08-11
Attempt to fix dual/ear color DNA issues.

Verkister:
-Mostly slipped lines between already existing vorestation edits in these files.
-Should probably fix the DNA related issues with dual color tails and ear colors.
-Such as resleeving breaking the forementioned on characters.
-Basically just added ear and secondary tail color definitions to the DNA codes.
-Untested though as I have no idea how to even test this in solo.


SpadesNeil:
fuckit, merged

PR #1993, opened 2017-08-12
Size gun fix

Cameron653:
- Fixes size gun description to show the correct %age
- Sets the design name to "Size gun" instead of "Shrink ray"


Cameron653:
Testing this before I merge.

Cameron653:
Tested and can confirm it works. Merging.

PR #2002, opened 2017-08-13
Fixes Promethean Humans looking really, really bad

Yoshax:


SpadesNeil:
Polaris might want this too, but I'll merge for now.

PR #2003, opened 2017-08-13
Trash piles can now spawn mice

Cyantime:
Trash piles will now slowly restock maint with wild mice.
Each pile can only spawn one mouse each half-hour, and no mice will be spawned if there's already a mouse visible from the pile, to prevent horrible mouse overpopulation.

### DIFF
--- a/code/game/objects/structures/trash_pile.dm
+++ b/code/game/objects/structures/trash_pile.dm
@@ -8,6 +8,7 @@
 
 	var/list/searchedby	= list()// Characters that have searched this trashpile, with values of searched time.
 	var/mob/living/hider		// A simple animal that might be hiding in the pile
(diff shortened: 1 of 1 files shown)


SpadesNeil:
Can you do me a huge favor with this?

Can you also create a generic object for respawning mobs over time? I have intentions of using it for gateway maps.

Cyantime:
Sure, why not.

Cyantime:
Do you want them to be destructible, like gauntlet enemy spawners?

SpadesNeil:
I want an option for them to be destructible.

I also want to be able to check if any players exist on the same Z-level, because it's supposed to be like a respawn mechanic for if players leave the map, to repopulate it for later

PR #2015, opened 2017-08-14
Enables vore organ borgo fueling.

Verkister:
-Borgs now gain recharge juice from the recently enabled vore organ gurgles.


SpadesNeil:
I think Travis croaked.

PR #2016, opened 2017-08-14
More Engineering fixes and adjustments.

Jakksergal:
- Fixes the waste loop not having a pump
- Squishes the radiator to 8 tiles rather than 10.
- Gives a smidge more detail to maintenance.

I really wish people would have commented and critiqued this stuff when it was open for a month.


Ketrai:
look, when you make big PR's, ask two people or so to playtest, or take a look at the map. it's going to iron out more issues.

PR #2018, opened 2017-08-15
Revert "Trash piles can now spawn mice"

SpadesNeil:
Reverts VOREStation/VOREStation#2003 until there's a change that limits the number of mice so long as the last mouse is still alive.

It's causing horrible lag.


Ketrai:
would always be nice to like, have mice spawn rather fast, but limit at 20-30, and they have to be dead, since mouse corpses don't cause any lag.

PR #2025, opened 2017-08-16
Configurable mob spawners, re-implementation of trash mouse spawning.

Cyantime:
By request of @SpadesNeil, robust mob spawners for whatever purpose might need them.
Re-adds trash mouse spawning using the spawners, each pile can have no more then two living mice at a time, with a respawn time of ten minutes.
Includes pre-made subtype that only spawns when no living players are on the same zlevel, as proof of concept for flexibility.


SpadesNeil:
This continues to cause crashing.

Cyantime:
I have no idea how an extra 50 or so would be crashing the server, it handled hundreds of mobs being added with the wilderness with only some lag.

PR #2026, opened 2017-08-16
Shock collars now come with exciting shocking action!

Gearshy:
Salvaged some electropack and radio code and made shock collars plug into the signaller system. Because some of us are probably into that. Collars come with an interface to select frequency and code like any other radio device, an on/off switch, and start switched off to prevent immediate zappings.


Cameron653:
Looks good to me. I'll go ahead and merge this later tomorrow if there's no additional comments about it since it's been sitting here for a while.

SpadesNeil:
As amusing as this is, I think it's very unwise to have shock collars spawn as accessories.

Gearshy:
I'm not sure I understand. They've been freely accessible through the loadout since before I first poked in here, if in a purely cosmetic form.

Cameron653:
Seeing as they start set to off, I can't imagine it being an issue or abused in any way. Collars are a pain to put on someone without handing them it and having them put it on, and the person would have to manually turn it on if they wish to use it for RP.

SpadesNeil:
Oh, nevermind then.

PR #2033, opened 2017-08-19
Add wood wall sprites

CashMoneyMoney:
image

Adds in new wall sprites for walls made out of wood.


CashMoneyMoney:
For any people wondering if it might merge with existing walls:

image

SpadesNeil:
Not really a fan of the riveted look. The lack of texture makes it look more like brown plastic.

CashMoneyMoney:
image

Cameron653:
The sprites look good, but this touches a normal .dm file.
You'd probably be better off sending this to Polaris.

Cameron653:
Upon second look, apparently Arokha modified these .dmi files already. So we're going to have conflicts if these files ever get changed on Polaris anyways, so it's not an issue to add these in. Not sure what to do with this PR, though.

Cameron653:
Looked at the code again and saw why. It'd be a fucking nightmare to modify the code instead of having the .dmi conflict.

See: https://github.com/VOREStation/VOREStation/blob/d63303f136b124d2c91dbd98b3a600a3be95cc5b/code/game/turfs/simulated/wall_icon.dm

In this case, modifying a regular .dmi file is fine. This should be ready for merge.

Belsima:
I might steal this for Polaris in my next wave of PRs, just as a note, since I heard that might explode things.

SpadesNeil:
Marking this in case we need to look up the conflict later.

PR #2055, opened 2017-08-24
Yes I figured out the shuttle issue.

Verkister:
-No, the line 84 did not replace or relate to the timer part you commented out.
-Due to being a less relevant call, the code skipped the line 84 stuff, moving onto the next thing below, slapping the shuttle through instantly because of the timer having been commented out from inbetween.

To clarify, as far as I could figure out, the line 84 exists to handle the custom shuttle crashing event.
The proc stuff of the line pointing towards the shuttle crash event file (code/modules/shuttles/crashes.dm) which had nothing to do with the shuttle timer part.


SpadesNeil:
Tested?

Verkister:
Tested and working

PR #2061, opened 2017-08-25
Makes vore organ transfer use the target organ's sound.

Verkister:
-Ye it's no longer hardlimited to squish2.ogg


SpadesNeil:
Is squish2.ogg an option still?

Verkister:
yeah I think it's still in the options for each belly

PR #2062, opened 2017-08-26
Makes the ID's on-mob sprite invisible.

SilencedMP5A5:
Since the change to Northern star from the station before that, the (rather ugly) under the arm ID sprite has been visible again. Honestly, it bugs the hell out of me. I don't know how many other people are bothered by the same thing.

Here's a picture with it on. https://gyazo.com/80b21439d59caff575fa7b9768e113ee
Here's a picture with it off. https://gyazo.com/29155101c035530335087b8b4876dea3

The obvious solution most people would probably bring up is 'just don't wear your ID.' 
But that means you're an unidentified person now until you put it on again.
The change isn't for me to say 'it's good' obviously. I just don't like the way the sprite looks. If someone could make it look nicer, sweet.
But honestly, I prefer it invisible.
disclaimer: all i did was go into the mob.dmi file and remove the visible pixels for 'id' and 'id2'
edit: Yes, i'm aware how minuscule the change is. Aaaaaaaaaaa.


Vorrarkul:
I can see a counterargument regarding the 'realism' of having the ID be visible. Perhaps there's a compromise here- shoes have a verb called "switch shoe layer". This verb doesn't seem to do much, until your character is wearing a floor-length dress or other foot-covering article of clothing, at which point its purpose becomes much clearer.

Perhaps the ID pixels can be similarly hideable as a verb option on the ID to move it to a layer below the jumpsuit?

SilencedMP5A5:
The thought crossed my mind a couple of times while thinking about it, however I have very little knowledge of coding, and didn't know how I would go about making a verb like that.

Also. If I just moved it to the bottom-most clothing layer, it would still be visible on people with nudity permits.

..right?

Yoshax:
This change won't be able to be merged anyway. It modifies a stock .dmi, which isn't allowed.

SpadesNeil:
tbh I prefer it invisible too

Like what is the point of it?

But we're not editing a stock .dmi file to do it.

PR #2064, opened 2017-08-26
Ketrai custom item

Cameron653:
Adds Ketrai's custom injector to let them become a farwa. They notified me it was accepted, so I tested the h2m injector and found out it didn't work. Went ahead and tossed together a ghetto h2m injector that only lets Ketrai (Kitchi) use it.

App here: https://forum.vore-station.net/viewtopic.php?f=45&t=1070


SpadesNeil:
As long as this doesn't end up like Catbug.

PR #2066, opened 2017-08-27
Prevents multiple extend votes.

Cameron653:
Have you ever looked at the manifest, saw "6:26" and thought "Oh! Well, it's been extended once, so I'll wait for next round..." and then coming back a few minutes later and seeing it at "6:40" and knowing that the round was voted to be extended _again_?

It is absolutely awful, and annoying. Shifts shouldn't have the possibility to go on _forever._ 
Having a hard-cap of one extend vote max (Or however many the server sets it to. One in this case.) makes it easier for people to plan when the next round is going to be so they don't constantly go "Alright, I'll join after this round since it has 40 minutes left." Now an extra hour. And another. And so on. 

I tested a simplified version of it (Setting the timer to 10 minutes) on a test server and it autocalled the shuttle when the timer was reached, so I'm confident this will work as planned.

### DIFF
--- a/code/controllers/autotransfer.dm
+++ b/code/controllers/autotransfer.dm
@@ -3,6 +3,7 @@ var/datum/controller/transfer_controller/transfer_controller
 datum/controller/transfer_controller
 	var/timerbuffer = 0 //buffer for time check
 	var/currenttick = 0
+	var/shift_hard_end = vote_autotransfer_initial + vote_autotransfer_interval //VOREStation Edit
 datum/controller/transfer_controller/New()
 	timerbuffer = config.vote_autotransfer_initial
 	processing_objects += src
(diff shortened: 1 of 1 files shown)


Ketrai:
This should definitely be a thing.

Dhaeleena:
Yeeeesssss, Finally

SpadesNeil:
There should be some kind of global message stating that:

"Warning: The round will not be extended again. Wrap up your scenes in the next 60 minutes."

Cameron653:
I can look into the global message. Shouldn't be too hard. 
Admins can override it as normal by just simply clicking the re-call shuttle command.

Cameron653:
Added a global announcement when the first extend vote is launched.

code/controllers/autotransfer.dm
```diff
@@ -3,15 +3,22 @@ var/datum/controller/transfer_controller/transfer_controller
...
 	currenttick = currenttick + 1
-	if (round_duration_in_ticks >= timerbuffer - 1 MINUTE)
+	if (round_duration_in_ticks >= shift_hard_end - 1 MINUTE)//VOREStation Edit START
+		init_shift_change(null, 1)
+		shift_hard_end = timerbuffer + config.vote_autotransfer_interval //If shuttle somehow gets recalled, let's do this.
+		timerbuffer = timerbuffer + config.vote_autotransfer_interval //Just to make sure a vote doesn't occur immediately afterwords.
+	else if (round_duration_in_ticks >= timerbuffer - 1 MINUTE) //VOREStation Edit END
 		vote.autotransfer()
 		timerbuffer = timerbuffer + config.vote_autotransfer_interval
+		world << "Warning: If the extend vote passes, the round will not be extended again. Wrap up your scenes in the next 60 minutes if the vote succeeds."
```
SpadesNeil: // Vorestation Edit

SpadesNeil:
Now is this configured to allow only one extension? Or can it be changed?

Cameron653:
Only one extension, but you can easily make it more by changing
shift_hard_end = config.vote_autotransfer_initial + config.vote_autotransfer_interval
to
shift_hard_end = config.vote_autotransfer_initial + (config.vote_autotransfer_interval * (number of extensions))

SpadesNeil:
Toss in the code for it and just make a comment explaining what it does in case we ever need it.

Cameron653:
Alright.

Cameron653:
There we go. Added the code and left comments on how to use it. Went ahead and revamped how the global announcement works, too. Announced 1 minute before the vote occurs that it will be their last vote.

PR #2067, opened 2017-08-27
More z-hole backup backup checks for non-turf junk

Verkister:
To maybe finally make sure only on-turf objects may fall down.
Onsite edit because fuck this. I had it fixed and working before.
NO BAD Z-HOLE. NO EATING SHIT OFF HANDS AND FOLKS YOU BAD. ONLY FLOOR FOOD FOR YOU


SpadesNeil:
Will this also stop flying cargo trams?

Verkister:
No idea actually. Probably has some odd check elsewhere. This thing is just supposed to make clear that the objects that aren't on the turf cannot fall through and since the cargo trams and such already go on turfs and so on

PR #2073, opened 2017-08-29
Late August Polaris Sync

Cameron653:
It compiles.
No idea what was in it. Was only a few conflicts that were so minor I fixed them in like 2 seconds.
automatic/carbine was replaced with automatic/bullpup. Commenting it out in trash.dm caused compile errors (for some odd reason.) so I just put its replacement in.

- Teshari stuff added to Polaris, so the teshari stuff on our side was removed to prevent duplicates.


Cameron653:
I lied. It doesn't' compile. one second.

Cameron653:
Compiles now.

SpadesNeil:
Travis disagrees.

Funce:
I'm all over it

Cameron653:
Now it should work. I though that was added to the same but guess not.

Funce:
Woo teamwork?

PR #2075, opened 2017-08-29
Medical Borgs Assorted Fixes

### LINKED ISSUE #1443: Allow Mediborgs/Medihounds remote access to the crew monitor
With the new AR-Ms, medical folks have access to the crew monitor anywhere, but, far as I can tell playing a medihound right now, there's no way for me to look at that without running over to the console in medical.

I'd imagine this might be very simply solved by giving the borgs AR-Ms, instead of the Medical hud item in their loadout?

Funce:
Fixes #1443 
And also amends the god-awful OVERWRITING that #1994 did.
Gah.

So Crisis Borgs, and Surgeon Borgs, are now selectable again
and Medihounds now have the Crew Monitor like every other medical borg.

If a changelog is needed lemme know.
image

### DIFF
--- a/code/modules/mob/living/silicon/robot/robot_modules.dm
+++ b/code/modules/mob/living/silicon/robot/robot_modules.dm
@@ -155,7 +155,7 @@ var/global/list/robot_modules = list(
 	src.modules += new /obj/item/device/flash(src)
 	src.modules += new /obj/item/weapon/crowbar/cyborg(src)
 	src.modules += new /obj/item/weapon/extinguisher(src)
-	vr_new() // For modules in robot_modules_vr.dm
+	vr_new() // Vorestation Edit: For modules in robot_modules_vr.dm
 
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/silicon/robot/robot_modules.dm
```diff
@@ -155,6 +155,7 @@ var/global/list/robot_modules = list(
 	src.modules += new /obj/item/device/flash(src)
 	src.modules += new /obj/item/weapon/crowbar/cyborg(src)
 	src.modules += new /obj/item/weapon/extinguisher(src)
+	vr_new() // For modules in robot_modules_vr.dm
```
SpadesNeil: Add "Vorestation Edit" here so we can easily find it if we ever need to update this.

Funce:
Cheers @SpadesNeil

PR #2076, opened 2017-08-29
Adds Handshakes and Head Patting

AaronSherwood117:
Thanks to Cameron for helping get this working. Been way too long since I started trying to learn DM.

Target a hand or the head on 'Help' intent (with an empty hand) and you can now perform classic human greetings and/or pet your friends!

Finally using Git, hope this works.

Feature in action:
pon58mf0txme7ubjnh_edw


SpadesNeil:
I think Polaris might like these too. We'll add them here first but offer this to Polaris later.

PR #2079, opened 2017-08-29
Fixes Ketrai's injector & gives lang. injector

Cameron653:
- Fixes Ketrai's injector to actually work. attack_self apparently does jack.
- Gives Ketrai a language injector to understand galcom, since you really can't scene without understanding what people are saying.

### DIFF
--- a/code/modules/vore/fluffstuff/custom_items_vr.dm
+++ b/code/modules/vore/fluffstuff/custom_items_vr.dm
@@ -1362,7 +1362,7 @@ obj/item/weapon/material/hatchet/tacknife/combatknife/fluff/katarina/handle_shie
 /obj/item/weapon/fluff/kitchi_injector/attack(mob/living/M, mob/living/user)
 
 	if(M.ckey != "Ketrai")
-		user << "<span class='warning'>Something compels you to <i> not </i> use this injector.</span>"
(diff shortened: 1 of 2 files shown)


code/modules/vore/fluffstuff/custom_items_vr.dm
```diff
@@ -1359,12 +1359,16 @@ obj/item/weapon/material/hatchet/tacknife/combatknife/fluff/katarina/handle_shie
 	icon = 'icons/obj/items.dmi'
 	icon_state = "dnainjector"
 
-/obj/item/weapon/fluff/kitchi_injector/attack_self(mob/user as mob)
+/obj/item/weapon/fluff/kitchi_injector/attack(mob/living/M, mob/living/user)
 
-	if(user.ckey != "Ketrai")
-		user << "<span class='warning'>You inject yourself.... Nothing happens.</span>"
+	if(M.ckey != "Ketrai")
+		user << "<span class='warning'>Something compels you to <i> not </i> use this injector.</span>"
```
SpadesNeil: `<i> not </i>` should be `<i>not</i>`

Cameron653: Changed it.

PR #2087, opened 2017-08-31
Removes whitelist requirements from Daemon and Enochian

Yoshax:
The whitelist existing currently does nothing more than hinder RP.

- The process of both getting approved for, and then added to the whitelist is tedious, to say the least. Especially for something that is actually rather small.
- So few people are whitelisted for the languages that there is nobody to talk to if you have them.
- As such, few people desire to gain the whitelist.
- Both Table and I believe the whitelist should not exist.

### DIFF
--- a/code/modules/mob/language/station_vr.dm
+++ b/code/modules/mob/language/station_vr.dm
@@ -55,7 +55,6 @@
 	exclaim_verb = "incants"
 	colour = "daemon" //So fancy
 	key = "n"
-	flags = WHITELISTED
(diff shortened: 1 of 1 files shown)


SpadesNeil:
Since it's Table's lore, I'll allow this. Also damn it @Cameron653 that is not what the misc tag is for. :U

Cameron653:
I mean, removing a whitelist on a preexisting feature isn't really a feature. It's a misc change if anything.

SpadesNeil:
Feature tag = Changing or adding features.
Bug = Fixing features that don't work as intended.
Refactor = A feature isn't changed. Not even the code. It's just moved or cleaned up.
Misc = Other stuff, such as changing the Github readme. Has nothing to do with the server.

Cameron653:
Oh, so _that's_ where the tag definitions are. Well, will keep that in mind next time.

PR #2089, opened 2017-09-01
Adds a ghost verb to alert blank pAI cards.

Verkister:
-The verb causes available empty pAI devices to briefly flash an alert on their screens.
-Didn't add the sound part due to spam protections failing for the sake of annoyance prevention. Just gotta hope someone walks right past when ya do it.
-You can now indicate the things might actually be of use for once without having to resort to metabegging.


SpadesNeil:
This is a neat thing Polaris should have.

PR #2093, opened 2017-09-02
Makes it so cargo can't buy the buggy buggy or the humvee

Cameron653:
They were making people invisible and defying gravity.


SpadesNeil:
Remove the MULE while you're at it. It's busted on this map.

Cameron653:
Done.

PR #2095, opened 2017-09-02
Adds Taur-Specific Rad/Bombsuits

Neopatra:
Also makes the file utility_vr since it did not previously exist in order to add said taur-specific suits. Will require adding the taursuit sprites as well.

### DIFF
--- a/code/modules/clothing/suits/utility_vr
+++ b/code/modules/clothing/suits/utility_vr
@@ -1,0 +1,125 @@
+/*
+ * Contains:
+ *		Bomb protection
+ *		Radiation protection
+ */
+
+/*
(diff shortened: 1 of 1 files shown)


Neopatra:
These should also be added (The sprite file and the security voidsuit codes respectively. Also I have no idea how github works)
https://www.dropbox.com/s/mjqz0h347doxwv2/taursuits_vr.dmi?dl=0
https://www.dropbox.com/s/sac3j93c1ic03ec/void_vr.dm?dl=0

Cameron653:
The file needs to be changed to utility_vr.dm and included in the .dme

PR #2101, opened 2017-09-04
Increase jobslots. Puts job changes in 1 file.

Cameron653:
Janitor 2         :arrow_right: 3
Cargo Tech 2 :arrow_right: 3
Borg 2            :arrow_right: 3
Mining 3        :arrow_right: 4


Cameron653:
Feel free to comment on the numbers and request changes.

Cameron653:
Travis you're a menace.

Dhaeleena:
I think bartender would be good at only 3
Janitors will need more lockers of gear.
Medbay might need more medical doctor closets too
everything else looks okay

SpadesNeil:
JOB CHANGES:
>MED: 5                :arrow_right: 7
why

>PSYCH 1              :arrow_right: 2
reasonable

>BORG 2               :arrow_right:4
reasonable

>JANITOR 2         :arrow_right: 4
It's a big station

>SECURITY 4       :arrow_right: 6
reasonable

>CARGO 2            :arrow_right: 4
holy shit did we only have two cargo techs?

>MINING 3           :arrow_right: 4
I guess

>CHEF 2                :arrow_right: 3
okay but why though

>ENGI 5                 :arrow_right: 6
We do _not_ need six engies.

>ATMO 3              :arrow_right: 4
Sure I guess. Phoron leaks and all.

>BARTENDER 2    :arrow_right: 4
_No_

code/game/jobs/job/z_all_jobs_vr.dm
```diff
@@ -0,0 +1,49 @@
+//Contains all modified jobs for easy access and editing.
+
+/datum/job/hop
+	alt_titles = list("Deputy Director", "Crew Resources Officer")
+
+/datum/job/doctor
+	total_positions = 7
+	spawn_positions = 7
```
SpadesNeil: Make sure this many spawn positions even exist.

SpadesNeil:
Marked "Needs Testing" until we confirm all the job slots have enough spawns on the map.

Cameron653:
Have any suggestions as to what the numbers should be changed to? I was going of what people in the discord were saying. Also, the amount of spawns on the map doesn't matter. Just means that two people might spawn ontop of each other IIRC.

Renkindle:
Mining's gonna need another locker on the map. Not sure about other jobs.

Yoshax:
These are too many slots, in my opinion. The only time this is remotely relevant is rounds like last night, which are once in a blue moon. There's simply not space in departments, nor the need, for most of this.

The departments would be exceedingly crowded anytime people filled it.

See Medical. There would be 7 Doctors, 2 Paramedics, 2 Chemists, a CMO and 2 Psychiatrists. That's a total of 16 people. The same goes for the other departments with their own numbers.

Why do we need this?

Cameron653:
I really have no idea why we need it. I'm just putting the PR up since people were requesting more job slots and bumping the numbers up to what people said in the discord.

 (The chef increase was my own doing, since too many cooks reference, and having 3 chefs isn't that bad of a thing seeing as people aren't making food 24/7 as a chef.)

Yoshax:
If we don't have a reason, then I don't think we should. As Ace always says we need a reason to do things, not a reason not to.

SpadesNeil:
You have learned well, young grasshoppah. :U 

But yes we shouldn't do this without a reason.

Like, extra janitors? I'm fine with that. The station is huge.

An extra Atmos Tech? Sure. That makes sense.

An extra chef? There's usually up to 50 people they have to feed so that's okay by me.

The rest of this is crazy.

Yoshax:
If we are going to be changing things like this, I think we should start small and incremental. Add one, not two, or more.  Some of them make sense, and are okay. Stuff like: Janitor, Cargo Tech, Miner. I feel like the others do not.

@SpadesNeil the concept behind that Atmos Tech is good, but in practice, I feel like the slots are rarely full, I would name an Atmos regular, but I can't think of one.

I feel like the intentions behind some of these are good, but in practice they're not needed and the equipment and workplace for them doesn't exist.

I am for:
Janitor - > 3
Cargo Tech - > 3
Borg -> 3
Mining -> 4

For the others, it would overflow the department when all the rules were filled (Engineering, Medical) or the equipment and workplace for them just doesn't exist to allow everyone to have a fun time at once (Psyche, Bartender, Chef.)

Keep in mind that for a lot of these, it would be taking work away from people who want to do it, and have joined to do it. That's somewhat unfair on them.

SpadesNeil:
That seems reasonable.

Cameron653:
Changed it to Yosh's suggestions

chaoko99:
>Questions on why here.

Basically because when all the [x]'s fuck off into maint, get eaten, and then don't quit the round, WHOOPS, WE HAVE NO [X] SLOTS!

SpadesNeil:
Has this been checked with the map?

Cameron653:
As far as I know, they'll just have multiple people spawn on the same tile if there's not enough spawn slots. Haven't tested since that'd require me to have multiple accounts, but I've seen it before where multiple people spawned on the same tile.

SpadesNeil:
I'd rather make sure it's fixed _now_ though.

PR #2103, opened 2017-09-04
Adds 4 more monkeys for Prometheans to transform into

Renkindle:
Sparra, Wolpin, Sobaka, and Saru

This is my first time making a PR, so if there's any mistakes I can try to change them. This did work on my test server.


Cameron653:
I see no issues with this seeing as the other 3 lesser forms were already selections for Prometheans.

PR #2115, opened 2017-09-06
Makes corner decals less anal about placement order.

chaoko99:
Does this even NEED a changelog?


SpadesNeil:
Fucking with layers is generally not a good idea actually now that I think of it.

chaoko99:
Works fine locally. What you guys really need is a proper set of defines and whatnot for planes and layers (Why don't you actually?)

chaoko99:
BLeh. Polaris actually could use a proper plane master port.
I'll do it later.

SpadesNeil:
Kay.

PR #2119, opened 2017-09-07
Adds Giant Tail/Fixes Zorren Colorable Ear

Natje0:
Adds the Giant Tail I made for Kyuu, approved here https://forum.vore-station.net/viewtopic.php?f=26&t=1082
Made it colorable and avaliable for public use.

Fixed the Zorren ears north-facing directional sprite. You'd see the inside of the eat from behind, and that was sad. No more, I say!

...My first ever PR. Hopefully I didn't do any major mistakes. Pls review.


SpadesNeil:
I _really_ do not like how that tail looks. It just doesn't look like a tail.

Natje0:
Okay, guess I'll either rework or make it locked to myself or something then. The original idea was that it wasn't *just* a tail, but also a large part of Kyuus body anyway, only reason I made it non-locked to myself was because someone asked me why I was locking it to myself and I couldn't come up with a good reasoning at the time.

PR #2121, opened 2017-09-08
Numbing bite and sharp melee fix

Cameron653:
**- Adds sharp numbing bite. Injects some numbenzyme into the person bit depending on the attack_damage RNG.**

**- Adds a fluff-syringe for special characters that would have numbing bites.**

**- Fixes a bug where the sharp melee attack trait didn't work.** 
I found this out because I couldn't get the syringe listed above to work, so I searched into the code and noticed that the way attacks are initialized is really fucking weird and awful (It's done on species/New) so I made a custom proc that occurs on spawn (update_attack_types) that makes people spawn in with the correct unarmed attack types.

**- Also made a custom proc for the numbing syringe fluff item.**

**- Changed kitchi_injector to fluff/injector/monkey, and made a base fluff/injector item for other items to be based off of.**

**- Gives numbing enzyme an OD variable once you have an extremely large amount in you to make it somewhat dangerous if you get a large amount pumped into you.**

I know bugfixes in feature PRs isn't proper, but I literally had to fix the bug in order to get the feature to work.


code/modules/mob/living/carbon/human/human.dm
```diff
@@ -1169,6 +1169,7 @@
 
 	species.handle_post_spawn(src)
 
+
```
Cameron653: What. Let me fix this.

SpadesNeil:
Numbing bite really should be a separate PR. I don't want to merge this until numbing bites are discussed.

Cameron653:
Agh, that's what I was afraid of. I'll split the bugfix and the numb bite after the bugfix is merged.

PR #2122, opened 2017-09-08
Adds a stairway to access the third level surface plateau

Cyantime:
The only way previously was to break through a wall or window somewhere, and adding a fourth airlock just for this seems unnecessary.
The stairs are just north of the main surface airlock, recessed into the cliff face a little.
staris


Dhaeleena:
I was actually going to do a thing like this but without making actual stairs. Just adding unfinished 'construction projects' and laving or two conveniently placed lattices to be used, still gives you the feel or risk and adventure of climbing up

Cyantime:
I mostly wanted stairs so that you can drag objects up and down, you can't make stairs ingame and can't drag things up ladders or lattices.

PR #2123, opened 2017-09-08
Updates SilencedMP5A5's fluff gun.

Cameron653:
https://forum.vore-station.net/viewtopic.php?f=26&t=1084
- Replaces revolver with a .45
- Preloads gun with rubber rounds
- Adds a soundfile for the gun firing


SpadesNeil:
Their permit needs to be re-added I think.

Cameron653:
As far as I can see, it's still in
https://github.com/VOREStation/VOREStation/blob/c1b2700b0238f9e960f5b97453520c9ed9016ac8/code/modules/vore/fluffstuff/custom_permits_vr.dm#L39-L42

PR #2126, opened 2017-09-09
Allows ghosts to become a pAI

Cameron653:
## Allows ghosts to become a pAI if it's empty and sitting on the floor.
- Makes it so pAI cards get deleted on death of the pAI, so they can't infinitely take control back over it.

- This makes it so that when a ghost clicks an empty pAI, they can take control over it.
- Asks them:
     - Are they sure they want to become a pAI?
     - What do they wish to be named?

If they don't answer either one of these, the pAI sits there like they never clicked it.
This makes it so players who want to play pAI don't have to go around to people in LOOC asking them to search with a pAI.
By doing it this way instead of choosing from all pAIs in the world, it makes it so that the chance of accidentally interrupting a scene is minimal.

Tested all the possible things people could do with it and I found no blatant or abusable bugs.


Cameron653:
Also made it so pAI cards get deleted on death of the pAI, so they can't infinitely take control back over it.

PR #2133, opened 2017-09-09
Atmos revamp, take 1

Gearshy:
I told you I'd do it! Or did I? Either way, this is my first attempt at making atmos better. Probably not ready to be actually merged, but it's there for reference/perusal/complaints.

Screenshots:
- https://puu.sh/xvMXF/901960acc3.png
- https://puu.sh/xvN18/1dcb675b3c.png

Features:
- Supply line will pressurize to a gentle 101.325kPa from round start.
- Two vents for dumping the scrubber loop outside.
- Optional radiator for cooling the pre-filtering waste gasses.
- One more vent for dumping the waste section.
- Hopefully more robust gas filtering.
- Less overheating air reserves?

Known issues:
- The air alarm that steers the scrubbers starts set to expect normal indoor conditions. I don't know how to make it more open-minded during mapping, unfortunately.


Cyantime:
Don't forget pipe caps on the open ends at the air, CO2, and plasma storage tanks.

Gearshy:
Considering pipe caps for some goddamn reason can't be removed when they're overpressure, I'd say putting them on defeats the point of leaving most of those ends available. >.>

Virgo113:
Consider capping the ends off with a dvalve and a cap, then. Wherever space allows for it

Dhaeleena:
Looks neat. Did you test it ingame yet to see how it all behaves and stuff?

Screemonster:
+1 just for having the "divert to outside" on the waste line before it even _reaches_ the filter loop, solving 99% of all "atmos is clogged so we can't so anything until it clears" problems before they start.

Cyantime:
The only thing that I don't like about this is the dvalve leading out of main air storage, I feel like that should be another normal pump that defaults to on and 1atm.
The scrubbers in the places of the old intake vents could also be moved to fit in line with all the new atmos devices out there, too.

Gearshy:
>Consider capping the ends off with a dvalve and a cap, then. Wherever space allows for it

I'll do that. And since https://github.com/PolarisSS13/Polaris/pull/3809 is a thing now, we may move to just pipe caps soon anyway.

>Did you test it ingame yet to see how it all behaves and stuff?

I gave it a small test. That is to say, I loaded it up, checked that all the pipes are connected together and such, then spawned a few cans of phoron in the atrium and introduced them to a welder. I have the sneaking suspicion I don't have access to quite the same ZAS/phoron presets the server uses, but fire is fire. The scrubber loop got really pressurized and really hot, but the outside vents seemed to be working decently well. When allowed to push the gasses through filtering, a single cooler on the post-N2O piping was enough to keep the N2/O2 mix at room temperatures. I've already figured out how to nudge things to let the radiator-cooled piping connect to either waste or scrubbers, so hopefully the next iteration will work better still.

SpadesNeil:
# MAP MERGE TOOL REEEEEE

Gearshy:
Wat.

Gearshy:
If only it'd work reeeeeeee.

PR #2139, opened 2017-09-10
Fixes #2138

### LINKED ISSUE #2138: Railing layer is being set to 5, and this is not correct
image
This is annoying.

Railing sprite layer should be 3.2 according to the code.

#### What actually happened
Fuck you, it's layer 5.

#### Steps to reproduce
N/A

#### Code Revision
- https://github.com/VOREStation/VOREStation/commit/d0e5ab12120c0951493273f1da9b895b9b64aaa2
 
- #2125 tried to fix this and failed.

SpadesNeil:


Cyantime:
That line makes it so railings facing south show up above mobs, like this:
railing
When massive people are stretching into the tile above them and being under railings on that tile, it's more a problem with resizing.
The file also doesn't need an edit comment, railings aren't from polaris.

SpadesNeil:
I'm aware that is the intended effect but that's not the result we end up with majority of the time.

PR #2143, opened 2017-09-10
Adds typing indicator to borgs

Cameron653:
See: title


Cameron653:
Someone is working on a better fix.

PR #2148, opened 2017-09-10
Numbing bite v2

Cameron653:
- Adds sharp numbing bite. Injects some numbenzyme into the person bit depending on the attack_damage RNG.

- Gives numbing enzyme an OD variable once you have an extremely large amount in you to make it somewhat dangerous if you get a large amount pumped into you. It now has special OD effects!

- Also made a custom proc for the numbing syringe fluff item.

- Adds a fluff-syringe for special characters that would have numbing bites.

- Adds numbing bites to the traits system, as numbing bites & sharp melee. 

- Increased the metabolism of numbing enzyme from 0.02 -> 0.2, so the effects won't last too long.

- Makes numbing enzyme slow down the person bit by a noticeable but not major amount.


Cameron653:
Also, I should mention, this sets out the framework and how other reagent injecting bites could work as an example in the code. This could be branched off to inject things like toxin, or have other various effects (Like causing the person to take burn damage over some time.)

SilencedMP5A5:
maybe make it cost 4 points like haste+

Cameron653:
Adding numbing enzyme OD. Will remove DNM afterwords and probably make it a perk.

code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Medicine_vr.dm
```diff
@@ -24,8 +24,37 @@
...
+		to_chat(H,"<span class='warning'>Your body feels numb, a light, tingly sensation spreading throughout it.</span>")
+
+/datum/reagent/numbing_enzyme/overdose(var/mob/living/carbon/M, var/alien)
+	//..() //Add this if you want it to do toxin damage. Personally, let's allow them to have the horrid effects below without toxin damage.
+	if(ishuman(M))
+		var/mob/living/carbon/human/H = M
+		if(prob(1))
+			to_chat(H,"<span class='warning'>Your entire body feels numb and the sensation of pins and needles continually assaults you. You blink and the next thing you know, your legs give out momentarily!</span>") 
+			H.AdjustWeakened(5) //Fall onto the floor for a few moments.
+			H.Confuse(15) //Be unable to walk for a bit longer.
```
Cameron653: walk properly*

Cameron653:
Guh. I wish it was possible to squash commits for when I do a minor edit. Anyways, any more feedback, especially when it comes to being a perk.

Cameron653:
holy conflict errors.

Let me fix this.

Cameron653:
Fixed the conflicts.

Ketrai:
I don't really see the point of a numbing bite? I mean I get that it's venom, but I'd rather have it be a slow attack someone can apply. for ykno, scenes. I also don't want it to be too strong. oh naga having lewd sex with the doctors, then they get called for an emergency and continue to waddle sideways. Ok, I have to admit that sounds pretty funny, but still, skeptical.

Cameron653:
Hmm, let me see about making the slow part of the numbing enzyme. It makes sense.

Cameron653:
This should be ready for merge provided there's no more requests.

Tasald:
I still hope this is going to be a white listed ability. I'm unsure how I feel about giveing a pain med a slow down, if anything instead of slowdown it should make some one "slightly" dizzy like narcotics do. If that's what the slow down is sapose to signify then never mind. If their saying that the special verb should take more time then just click spam some one with it, that I could agree with as well. Make it like that would make it harder to just spam it in combat and lol kill some one with toxin damage, not that people should be doing that anyway, but yea.

Cameron653:
I made it specifically so it doesn't do toxin damage whatsoever by overrriding the /overdose proc with a /overdose proc specific for the numbenzyme. The "worst" effect it has is making someone lose their breath for a few moments. And yeah, the slowdown is supposed to signify that they're numb, they're having a hard time doing things since making them dizzy (causing their screen to infinitely shake) would be extremely deliberating for some players.

Cameron653:
This is ready for merge if there's no more suggestions (Seeing as nobody's commented for 7 days, I'm imagining not.)

SpadesNeil:
eeehhhhhhhhhhhhhhhhhhhhhhhh

SpadesNeil:
fuckit

PR #2153, opened 2017-09-12
Reenables size changing.

Cameron653:
- Reenables size changing verb.
- Makes the size changing verb Promethean only.
- Makes the size changing verb available to pAIs due to lack of ways to change size.
- Allows non-humanoids to use the size changing verb.

People want it back for Prometheans and other things.


Dhaeleena:
Yyeeess.
I want this because we are already changing and adding so many things to facilitate role-play and making features more role-play friendly...I cant think of a more role-play friendly tool than re-enabling this.

kcin2001:
this can be very usefull and cool since for prommies they could technically intake more mass and grow from it by turning it to goop and borgs dont keep their size between rounds if my memory serves and some people may want to have a larger or smaller size of robot

SilencedMP5A5:
i support this because:
big (and small) borgs are CUTE
and also prometheans should be able to mass change.
(and so people can reverse accidental/intentional sizeray shots they don't approve of)

SpadesNeil:
Considering how flagrantly abused size changing was previously, I have really no desire to see it added back in.

Prometheans can have the ability if they want but they're _prometheans_.

kcin2001:
I at least want to be able to spawn as bigger or smaller than 100% with a Borg

Ketrai:
tbh, I'd be in favour, but all the roleplay aiding elements we have are still mostly IC, and something you have to achieve. There's also plenty of ways to change size. I do agrees borgos etc should be able to change easily without someone else to use a sizegun on em, however everyone else should just use the nifsoft or the size gun, it's not that hard.

Kligor:
I personally think its a good idea for if you want to roleplay things you normally could. Sure you COULD do some things but if its abused just.... tell the player to knock it off? I feel like if someone's abusing size changing its for reasons that'd probably get them in more trouble than just abusing size. If Im wrong, tbh, please let me know what they can abuse with it? I really dont see much use to abusing it.

Zeronis:
The only reason I can fathom for it being removed in the first place was because it was being abused, and then that's not the problem with the tool, that's a problem with the people abusing the tool.
Don't remove features, remove the fuckups fucking it up for everyone else. If the temptation is really so strong to break the rules and abuse features with blatant disregard then they either don't respect or don't care about the rules they're breaking and need to be dealt with.
I am in favor of bringing back this tool despite the massive amount of size changing equivalents that exist, due in part to the people/characters who could benefit from it that aren't covered by the existing alternatives, but I also won't shed any tears if it falls through.

Renkindle:
I feel like it should only be available to prometheans/borgs. Prometheans because it makes sense species-wise and borgs because they can't eat the mushrooms or use a size gun on themselves or anything.

DemgelZalvine:
Personally I don't want to see this back for everyone at all. I was glad when it was gone, as the people claiming magic and abusing the OOC use size-change button was really getting annoying. There are many means in station to change your size now, via NIF, Size gun, etc, that can set you however you'd like and not terribly hard to get. 

Arguably, enabling something like this for borgs wouldn't be so bad, maybe a size change module for them? Maybe promethians as it makes sense...but they already have a Ton of perks as is.

Arokha:
I'm with Zalvine on this. Adding options for those that can't size-change because none of the existing methods work well with them or are mostly unavailable is a good idea, but I'm of the opinion that magical size-changing requires an SCA, like any other magical ability. I could see maybe an exception for prometheans since goo.

kcin2001:
Yeah and its not a new concept that slimes can increase mass by eating since thats how xenobio smiles work essentially

AliciaJewel:
I used the size change verb when I played my Promethean for a mass increase when she ate, so I might be a bit biased toward saying that they should at least be given to them.

Cameron653:
Alright. General consensus from what I can see if Prometheans should have it. Going to change it.

Cameron653:
Done. It's been made Promethean only. Tested it and the only species that has access to the proc is Prometheans.

Cameron653:
_NOW_ it's done.
Only pAI's (Due to lack of ways to change size) and Prometheans (Due to them being slime) can use the size changing verb.

kcin2001:
@Cameron6531 what about borgs?

SpadesNeil:
We changed our mind with borgs. We don't need tiny borgs.

kcin2001:
and pais are fine being even tinier?

Cameron653:
Seeing as pAIs can't:
1. Open ID locked doors
2. Grab people
3. Hit people
4. Use machines
etc
I don't think it's really that much of an issue, while borgs could do all those things.

kcin2001:
ok then, if there was a way to restrict the size change to certain limits easily i would suggest that

SpadesNeil:
>Allows non-humanoids to use the size changing verb.

What?

Cameron653:
There was a !ishuman check that prevented borgs, pai's, etc from using the verb. I removed said check.

SpadesNeil:
I kinda don't want borgs able to change size freely.

Cameron653:
They can't

SpadesNeil:
oh ok

PR #2154, opened 2017-09-12
Makes the thermal vision NIF software illegal.

Cyantime:


Natje0:
It's already locked to Director Access(Or something like that). I think it's fine that way, tbh...

SpadesNeil:
I don't think it should even exist tbh.

Cyantime:
It's currently available to anyone with captain access, combat cyborgs that are only available during code red and bound by laws had theirs taken away, does the captain get a pair of (undetectable) thermals?

SpadesNeil:
That is why it shouldn't exist.

Cyantime:
Making them illegal is pretty similar to not existing, the machine needs to be both hacked and emagged for someone to get them. They also fit better with the other combat software, like the one that generates an egun.

kcin2001:
illegal would work probably

SpadesNeil:
I don't think that one should exist either!

But at least if it's illegal medical crew will generally be like "NO I WON'T INSTALL THAT! IN FACT I'M CALLING SECURITY!"

Cyantime:
You don't install nifsofts, you just go to a vendor and then you have them. You only need to have the nif itself installed by medical.

kcin2001:
Ace medical doesnt install the software only the hardware

SpadesNeil:
Oh. <_>

SpadesNeil:
tbh thermal and weapon nifs should be _removed_

Ketrai:
Why? they're really hard to get. e.g you need to emag and hack the nifsoft machine. (altho emags are rather rare) then they're SO expensive you need to be a high paying job before you can even get them. and if anything people rarely use em, ever.

Renkindle:
I've never seen the illegal weapon NIF software even used before because of those hurdles to get to it.

SpadesNeil:
Expect me to abuse this.

PR #2161, opened 2017-09-14
Boosts air alarm thermostats

Gearshy:
So, air alarms are our only thermal control that doesn't rely on replacing air or handheld power cells. Problem is, they suck. Buuuuut apparently this variable actually lets me change that. It can even be overriden by mappers/admins to make specific air alarms *even more effective*. Worth a shot, I think.

### DIFF
--- a/code/game/machinery/alarm.dm
+++ b/code/game/machinery/alarm.dm
@@ -39,7 +39,8 @@
 	anchored = 1
 	use_power = 1
 	idle_power_usage = 80
-	active_power_usage = 1000 //For heating/cooling rooms. 1000 joules equates to about 1 degree every 2 seconds for a single tile of air.
+	//active_power_usage = 1000 //For heating/cooling rooms. 1000 joules equates to about 1 degree every 2 seconds for a single tile of air.
+	active_power_usage = 15000 //Vorestation edit
 	power_channel = ENVIRON
 	req_one_access = list(access_atmospherics, access_engine_equip)
 	var/alarm_id = null


code/game/machinery/alarm.dm
```diff
@@ -39,7 +39,8 @@
 	anchored = 1
 	use_power = 1
 	idle_power_usage = 80
-	active_power_usage = 1000 //For heating/cooling rooms. 1000 joules equates to about 1 degree every 2 seconds for a single tile of air.
+	//active_power_usage = 1000 //For heating/cooling rooms. 1000 joules equates to about 1 degree every 2 seconds for a single tile of air.
+	active_power_usage = 15000 //Vorestation edit
```
SpadesNeil: this seems unwise to supercharge _all_ the air alarms, and furthermore, this is not how you would make a vorestation edit in this scenario

you'd make a new DM file called alarm_vr.dm and change only that one var.

```
/obj/machinery/alarm
	active_power_usage = 15000
```

Gearshy:
Okay, fair enough.

code/game/machinery/alarm.dm
Gearshy: Supercharging all the alarms would mean there wouldn't be a part two (and three, and four, and five...) that consists of selectively map-editing a hypothetical more powerful variant in, but okay. Fair enough on _vr.
Would it be better if I instead defined a few alarm subtypes with varying thermostat power ratings? Something like...
```
/obj/machinery/alarm/thermo2
    active_power_usage = 7500
/obj/machinery/alarm/thermo3
    active_power_usage = 15000
/obj/machinery/alarm/atrium
    active_power_usage = 100000

SpadesNeil: I have to edit the map anyway. Why is this any better?

Gearshy: You said that you'd rather not supercharge all air alarms everywhere. I'm trying to figure out what, if anything, I can do to make the idea acceptable.

PR #2162, opened 2017-09-14
A miniscule fluff item fix too small to deserve a PR but gets one anyway

Verkister:
-Makes the recent purely cosmetic fluff goggles able to be worn additionally on the extra ear slot and thus not be completely hidden by the character's hairstyle in the "up" state.
-Also fixes my fluff stuffs missing some directory fluff. I ain't fixing this for other people's stuffs this time tho.


SpadesNeil:
This item directory should have `/fluff/` in it also.

PR #2166, opened 2017-09-16
Neopatra's taursuits

Neopatra:
Adds in several more taur variants of voidsuits, biohazard suits, and radiation suits for use by our bigger friends. Also includes code and sprites for taur-specific bomb suits but there is currently no way to summon them naturally as there is no bomb suit crate or equivalent.


Ketrai:
I helped test this and can confirm it's quite good. also yes, this does add extra suits to existing crates, making them more valuable, but the amount of normal players is far greater to taurs, and taur suits are not applicable to anthros.

SilencedMP5A5:
aw ye.

Neopatra:
Oh, whoops. I just committed the same file twice. No wonder it failed the second time. Derp.

PR #2168, opened 2017-09-17
Adds size digestion and adjustable stomach examine.

Cameron653:
- Adds three new digestion modes
1. Shrink. Slowly shrinks prey inside of you and gives you nutrition
2. Grow. Slowly grows prey inside of you and reduces your nutrition
3. Size steal. Slowly steals preys nutrients and size and gives it to you.
- Adds two types of new toggles.
1. Bulge size. This allows you to set how large prey has to be to cause your stomach to show a message on examine.
2. Shrink/grow size. Determines at what point you will stop growing/shrinking your prey.
- Absorbed people don't show up on examine anymore, since they shouldn't make a bulge/whatnot when they're fluff


SpadesNeil:
Nifty.

SilencedMP5A5:
tbh i kinda liked having absorbed prey be visible on examine. my solution was just to move them to another belly labelled 'pudge' or something.

PR #2171, opened 2017-09-18
Game Kit - Reborn

Zap-zapper:
It's time for this wonderful little device to come back again into life. I originally was porting it for other server, but I noticed, that this device can be very useful for this server too. 

This device allows you to play checkers, chess, or whichever game involving those pieces.

I refactored it's kinda ugly code somewhat, removed redundant src's, images are now loading from the game server, not from the cloud, and also I added a feature which allows players to invert board.

Screenz:

gmkt1

gmkt2

gmkt3

gmktkrch


SpadesNeil:
This edits a lot of core files for no good reason, but really, we should give something this neat to Polaris anyway.

JoanRisu:
It's more like 3-4 files. The other files are just the image assets. I'm going through it and it seems to just be cleaning up existing code to render assets off the server instead of an outside source and moving the gaming kit to its own file instead of being in code/defines/obj/weapon.dm.

If anything, the edits to some of our core files were unavoidable but I agree this would probably be more useful and efficient as an upstream pull request at Polaris.

Anewbe:
Oh hell yes, if this works, want.

SpadesNeil:
I have no confirmation if it works because I haven't _personally_ tested it but it _should_ work just fine in your code since it seems to work in ours.

JoanRisu:
If anything, a framework is there that Travis likes.

DickFreedomJohnson:
so when will this be added to polaris and thus added to us?

SpadesNeil:
When someone wants to add it.

PR #2177, opened 2017-09-21
Ports Autotransfer for bellies

Poojawa:
Ohai.

So recently I enabled a megafauna the ability to eat people, then pushed them automatically from maw -> gullet -> gut for gurgles and eventual fate.

Sadly, this requires the porting of addtimers and the subsquent CALLBACK in order to run properly, as these systems need to be able to check if someone's been moved already or is there to be moved. 

Since that's more upstream, I figure I'd be a good neighbor and port it for y'all. 

c70e6195d833a0c608655108a5b9261c
example of being hands off, yet mushed through the connections.


code/modules/vore/eating/belly_vr.dm
```diff
@@ -515,16 +537,8 @@
 	if(!(content in internal_contents))
 		return
 	internal_contents -= content
-	target.internal_contents += content
-	if(content in items_preserved)
-		items_preserved -= content
-		target.items_preserved += content
```
Verkister: Psst I un-nukified the loot gurgle code so you may wanna slip this part back in or things may get kinda borky heh.

SpadesNeil:
This ever gonna be unbroken?

PR #2192, opened 2017-09-23
Fixes the dead running into walls and disposal bins.

Cyantime:


code/modules/mob/living/living.dm
```diff
@@ -138,7 +138,7 @@ default behaviour is:
 			..()
 			if (!istype(AM, /atom/movable) || AM.anchored)
 				//VOREStation Edit - object-specific proc for running into things
-				if((confused || is_blind()) && prob(50) && m_intent=="run")
+				if(((confused || is_blind()) && stat == CONSCIOUS)) && prob(50) && m_intent=="run")
```
Cameron653: if((confused || is_blind()) && stat == CONSCIOUS && prob(50) && m_intent=="run")
Should work. No need to put ()'s around confused and blind, since it already has one doing that.

Cyantime: This is what I get for trying to code while falling asleep

PR #2193, opened 2017-09-23
Adds nom-on-bump behaviour

Screemonster:
Adds a new behaviour to certain simplemobs, giving them a chance of deciding to gobble people up that bump into them regardless of hostility.

This bumpnom behaviour will respect the usual checks like ignoring indigestible people and not eating when full. It _does_ bypass the check for whether it can eat standing people, but assuming it doesn't also pounce you to the floor (with the same chances as it has in combat) you can just move away to avoid it. If you stand still while a wah's casually stuffing you in its mouth you deserve to get ate.

The proc will only trigger in situations where you bump it on non-help intents, or where you would normally swap places. If you decide to go AFK in a room with a catgirl or a gaslamp, buckle in. Likewise, people who are resting or otherwise  have canmove==false won't trigger this, so SSD people _should_ be safe.

Simplemobs don't seem to trigger this against each other (seriously, I filled a room with them bumping into each other all over the place, debug messages everywhere, nothing) so there's no worry about a petting zoo turning into that scene in futurama where they introduce Nibbler. There's an anti-cannibalism check to make sure they won't eat mobs that share a type anyway.

If a mob _does_ try to eat you when you bump it, you'll end up on the same tile rather than swapping. This is because swapping looked really fucky, especially when they pounced and your prone body went sliding across the floor as they moved into your old tile, before getting grabbed back to where you were. (only damper)

### DIFF
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -92,12 +92,6 @@ default behaviour is:
 				var/turf/oldloc = loc
 				forceMove(tmob.loc)
 
-				// VOREStation Edit - Begin
-				// In case of micros, we don't swap positions; instead occupying the same square!
-				if (handle_micro_bump_helping(tmob)) return
-				// TODO - Check if we need to do something about the slime.UpdateFeed() we are skipping below.
-				// VOREStation Edit - End
-
 				//VOREstation Edit - Begin
 				if (istype(tmob, /mob/living/simple_animal)) //check bumpnom chance, if it's a simplemob that's bumped
 					tmob.Bumped(src)
@@ -107,10 +101,12 @@ default behaviour is:
 					now_pushing = 0
 					return
 
-				//VOREstation Edit - End
+				// In case of micros, we don't swap positions; instead occupying the same square!
+				if (handle_micro_bump_helping(tmob)) return
+				// TODO - Check if we need to do something about the slime.UpdateFeed() we are skipping below.
+				// VOREStation Edit - End
 
 				tmob.forceMove(oldloc)
-
 				now_pushing = 0
 				return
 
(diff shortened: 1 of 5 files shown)


Screemonster:
Mobs that have been given this behaviour (so far):

- Catgirls. 5% chance, so it's not _hugely_ likely but it'll happen occasionally.

- Wahs. 10% for normal wahs, 75% for greedy purplewahs. Wah nom behaviour has been adjusted after a chat with Zal.

- Gaslamps. Relatively huge 90% chance, but their pounce chance is so low that as long as you move away from them you'll be fine. They're big passive drifty things, so casually eating whatever bumps into their tentacles seems appropriately jellyfishy.

Ketrai:
Add the bump proc for solargrubs, they do eat people, just only ones who surrender, I feel like it could be neat for em

HzDonut:
Maybe it ignores the standing check for micros? After all, tackling someone so small to the floor to eat them would be silly.

Screemonster:
That might be a worthwhile thing for regular pounces, this just skips it. The can_eat_standing var is used in combat to decide whether hostilemobs can eat people, this behaviour only comes into play outside of combat. As it happens, the bumpnom check is done immediately after the "are you small enough for me to step on" check that micros go through.

KotetsuRedwood:
A 50% chance for solargrubs sounds neat and maybe some chances for oties, too. Probably like 5% for ones on station and 30% for wilderness ones, maybe?

Screemonster:
DNM 'cause I need to add another check to disable the behaviour for player-controlled simplemobs.

Screemonster:
They now won't do the thing if they have a ckey. Also moved the check to avoid possible conflict with micro step behaviour. (ie. it won't skip the nom check for people small enough to walk under them now)

PR #2196, opened 2017-09-23
Smol fix for grub event.

Ketrai:
This makes it so grubs are less tedious to deal with. They always spawn with around 4, and always in the same area. This way security shouldn't have to hunt them down.

- Also for people who don't know yet, their chem injection has been fixed, so now they don't do insane amounts of damage like they're not supposed to anyways.

### DIFF
--- a/code/modules/events/grubinfestation_vr.dm
+++ b/code/modules/events/grubinfestation_vr.dm
@@ -1,35 +1,35 @@
 /datum/event/grub_infestation
-    announceWhen = 10
-    endWhen = 11
-    var/location
-    var/locstring
- 
+	announceWhen = 10
+	endWhen = 11
+	var/location
+	var/locstring
+
 /datum/event/grub_infestation/start()
- 
-    location = rand(0,8)
-    var/list/turf/simulated/floor/turfs = list()
-    var/spawn_area_type
-    spawn_area_type = /area/tether/outpost/solars_outside
-    locstring = "the solar farm"
- 
-    for(var/areapath in typesof(spawn_area_type))
-        var/area/A = locate(areapath)
-        for(var/turf/simulated/floor/F in A.contents)
-            if(turf_clear(F))
-                turfs += F
- 
-    var/list/spawn_types = list()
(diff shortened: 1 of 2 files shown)


Ketrai:
How is there a dme error when I literally downloaded a fresh branch like an hour ago?

code/modules/events/grubinfestation_vr.dm
```diff
@@ -1,26 +1,35 @@
...
+			var/spawn_type = pick(spawn_types)
+			new spawn_type(T)
 
-	while((spawncount >= 1) && vents.len)
-		var/obj/vent = pick(vents)
-		new /obj/effect/spider/spiderling/grub(vent.loc)
-		vents -= vent
-		spawncount--
+/datum/event/grub_infestation/announce()
+	command_announcement.Announce("Bioscans indicate that solar grubs have somehow entered through the gateway. Clear them out, before this starts to affect the powergrid.", "Vermin infestation")
```
SpadesNeil: The gateway? Do they only ever spawn in the gateway?

Ketrai: Yup, any other areas you want them to spawn in? people complained about them ending up all over the station, even in like telecoms

SpadesNeil: I say let them show up anywhere except some blacklisted areas, but _not_ use vents.

Or conversely, whitelisted areas, as defined in the original infestation code for mice and lizards.

Ketrai: Actually am going to gather opinions on this via a vote.

Ketrai: https://strawpoll.com/5c8ry1s7

SpadesNeil: "All over station, excluding the dorms" has huge majority compared to other stuff.

Ketrai: Alrite, i'll close this pull for now and go work on a fix so they can still spawn everywhere, just not the dorms.

PR #2200, opened 2017-09-24
Bumps borgs up by one.

Cameron653:
Increases borgs, so one more borg can latejoin.

https://i.imgur.com/X7MIKPZ.png

### DIFF
--- a/code/game/jobs/job/z_all_jobs_vr.dm
+++ b/code/game/jobs/job/z_all_jobs_vr.dm
@@ -29,8 +29,8 @@
 	spawn_positions = 4
 
 /datum/job/cyborg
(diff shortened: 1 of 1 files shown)


icowom:
Mhmm, this would be perfect for setting up a squad of doggo borgs! Also some spread where every dept has at least one borg to call into service

SpadesNeil:
I'm not sure we need so many borgs tbh.

Cameron653:
People are requesting it and ahelps for borg slots to be opened isn't too uncommon, along with the fact borgs have such a different playstyle than humanoids that people seem to enjoy.

SpadesNeil:
fuckit, but no more

PR #2202, opened 2017-09-24
Fixes loincloth description

Cameron653:
See: title.


Cameron653:
This was requested by the creator of the loincloth, and is just a simple typo fix+desc addition.

PR #2229, opened 2017-09-30
Gets rid of a duplicate file

Cameron653:


SpadesNeil:
I was wondering wtf this was

PR #2234, opened 2017-09-30
Makes neutral traits not take trait slots

Cameron653:
- Makes neutral traits not take up trait slots.

There was virtually 0 reason in game to take something like coldblooded, autohiss, increased metabolism, bloodsucker trait, etc due to its small impact on gameplay when you could take positive+negative traits and actually see a difference.

This change encourages users to actually choose neutral traits, as they're neutral and shouldn't take up slots when they do barely anything to gameplay.


Yewmi:
I would love to see this PR be merged. There are plenty of neutral traits I would love to use on most of my characters that I cannot use as I want to keep my positive and negative traits as well.

PR #2236, opened 2017-09-30
Makes all external airlocks glass.

Cyantime:


Ketrai:
will this horribly break things again?

Cyantime:
It shouldn't?

SpadesNeil:
This looks like it was done correctly.

PR #2241, opened 2017-10-01
Limits bluespace harpoon teleportation range.

Cyantime:
Prevents infinite teleportation distance via camera.


Ketrai:
great! MERGE DIS

SpadesNeil:
Oh come on where's the **FUN**™ in that?

I'm kidding, this is a serious exploit and should be fixed.

PR #2253, opened 2017-10-03
Adds open space turfs above all stairs.

Cyantime:
Fixes
```
Runtime in turf.dm,87: list index out of bounds
   proc name: update icon (/turf/simulated/open/update_icon)
   src: (38,52,7) (/turf/simulated/open)
   call stack:
   (38,52,7) (/turf/simulated/open): update icon()
   openspace (/datum/controller/process/open_space): update turf( (38,52,7) (/turf/simulated/open))
   openspace (/datum/controller/process/open_space): doWork(1)
   openspace (/datum/controller/process/open_space): setup()
```

In the future, please don't put floor tiles above stair objects.
This probably isn't related to the recent crashes on the actual server, though.


SpadesNeil:
screenshots please

because I have a much larger map update inbound and I am 90% sure it will cause conflicts

Cyantime:
There's no changes in-game, because stairs change the tiles above them to open space on initialization. But they don't do it in time or correctly or something and it causes that runtime occasionally on server startup. Tiles changed are the ones above the primary surface and asteroid stairwells, and the ones above the atmos stairs.

SpadesNeil:
aaaand I caused conflicts, knew I would :C

Cyantime:
Conflicts cleared

PR #2267, opened 2017-10-05
Readds reaction for slime jelly to purple slime

aztectornado:
I KNEW something felt like it was missing! 

This readds the ability to get slime jelly from actual slimes, instead of only being able to make it from phoron and craptons of reagents. 

TG's version of the reaction pays out 10 slime jelly, but considering what ours can do, I nerfed it to 5u per extract reaction.

The reason for picking 5u is twofold: Xenobio2 used to give you the same amount per slime core you blended, and making 5u in the lab would require 20u phoron, which is a full sheet's worth. 

Hopefully overall we'll start to see slime jelly and its various reactions start to become viable again, but if this is still too little for things, we may want to increase the payout up to 10, since purples have other, arguably better uses and of course tend to mutate into other colors instead of staying purple, thus making it hard to get a lot of extracts in the first place.

### DIFF
--- a/code/modules/xenobio/items/extracts.dm
+++ b/code/modules/xenobio/items/extracts.dm
@@ -289,7 +289,7 @@
 	name = "purple slime extract"
 	icon_state = "purple slime extract"
 	description_info = "This extract can create a slime steroid agent when injected with phoron, which increases the amount of slime extracts the processor \
-	can extract from a slime specimen."
+	can extract from a slime specimen. The extract can also create a strange, morphic toxic jelly if injected with sugar."
 
 
 /datum/chemical_reaction/slime/purple_steroid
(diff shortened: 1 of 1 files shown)


Yoshax:
https://github.com/PolarisSS13/Polaris/pull/3952

SpadesNeil:
Already handled in Polaris. We'll get it in our next sync.

aztectornado:
Wow. And theirs is significantly more OP than mine.

aztectornado:
I'd like this to be re-opened. Polaris' version of this is SIGNIFICANTLY more OP than this

5u slime jelly currently eats 20u phoron (a full sheet)

Polaris' version requires less than 0.0334u phoron per 15u and a common as dirt extract
My proposed version uses an uncommon extract and sugar to make only 5u jelly

Considering how end-game level jelly reactions are, I think this would be a lot fairer

SpadesNeil:
Okay well even still I'm not reopening this PR only because it was done incorrectly to begin with.

See here: https://github.com/VOREStation/VOREStation/blob/36ddd8f93a72d51e647f869992fb6774cea8b75b/.github/CONTRIBUTING.md

PR #2268, opened 2017-10-05
Fixes #2249

### LINKED ISSUE #2249: Mixing Gin+Sake bug
Mixing equal measures of Gin and Sake (Tested using the bottles from the vendor) makes the glass instantly empty itself of both.

I expected a delicious japanese sake-based drink.

#### What actually happened
I got despair and nothing to soothe my desire for weeb-drinks.
(But really, what happened was that the glass became completely empty)

#### Steps to reproduce
- (Step 1) Pour X units of Griffeater Gin in some container (min size 2X units)
- (Step 2) Pour X units of Mono-No-Aware luxury sake into the same container.
Done, container is now empty of both.

#### Code Revision
Server revision: release - bf51b997afd5f1c066bbb24692bcb8c3a65babe3 

- Tested this on the bottled versions of the alcohol, but I suspect it works on normal alcohol also.

aztectornado:
Someone misspelled something that was literally one line under the correct spelling. 

Whatever. This fixes the bug where saketini makes your booze vanish instead of making the proper drink

### DIFF
--- a/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Food-Drinks.dm
+++ b/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Food-Drinks.dm
@@ -2322,7 +2322,7 @@
 
 /datum/reagent/ethanol/saketini
 	name = "Saketini"
-	id = "sakitini"
+	id = "saketini"
 	description = "For when you're too weeb for a real martini."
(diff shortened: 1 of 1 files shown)


SpadesNeil:
This should be fixed for Polaris, or it should have `// Vorestation Edit` added.

PR #2270, opened 2017-10-06
Disables spider infestations.

Ketrai:
Seeing as these spiderlings, which have S.amount_grown = -1 are still able to grow into mature spiders, whilst their code suggests they would spawn as duds, hence never be able to form into dangerous spiders that attack people on sight. Seems to only be a 50% chance a spiderling grows up, doesn't remove the danger.

### DIFF
--- a/code/modules/events/infestation.dm
+++ b/code/modules/events/infestation.dm
@@ -72,6 +72,7 @@
 			spawn_types = list(/mob/living/simple_animal/lizard)
 			max_number = 6
 			vermstring = "lizards"
+/*
 		if(VERM_SPIDERS)
 			spawn_types = list(/obj/effect/spider/spiderling)
 			max_number = 3
@@ -91,6 +92,7 @@
 				var/spawn_type = pick(spawn_types)
 				new spawn_type(T)
 
+*/
 
 /datum/event/infestation/announce()
 	command_announcement.Announce("Bioscans indicate that [vermstring] have been breeding in [locstring]. Clear them out, before this starts to affect productivity.", "Vermin infestation")


Ketrai:
actually seeing as this edits a dm file... hang on

Ketrai:
yah, probably should make a new _vr file overwriting the code for this.

Cyantime:
Why?

Screemonster:
Is this a PR to remove something that inconvenienced you last night?

code/modules/events/infestation.dm
```diff
@@ -72,6 +72,7 @@
 			spawn_types = list(/mob/living/simple_animal/lizard)
 			max_number = 6
 			vermstring = "lizards"
+/*
```
SpadesNeil: // VORESTATION EDIT

Cameron653:
Seeing as the spider infestation is disabled, I'm assuming the spiders spawning from the vermin event was unintended.

Ketrai:
It's pretty much something that has always bothered me. if someone finds a way to make the spiders guaranteed not to grow up, then I wouldn't see an issue with the code.

Cameron653:
If you wanted to, you could make 
``` /obj/effect/spider/spiderling/no_grow```
And then do
```
/obj/effect/spider/spiderling/no_grow/New(var/location, var/atom/parent)
	..()
	amount_grown = 1
```

Along with changing 
https://github.com/VOREStation/VOREStation/blob/5c7f6f98635fa0814e0e566cc12d419907a8455d/code/modules/events/infestation.dm#L76
and line
https://github.com/VOREStation/VOREStation/blob/5c7f6f98635fa0814e0e566cc12d419907a8455d/code/modules/events/infestation.dm#L88
to /obj/effect/spider/spiderling/no_grow and /obj/effect/spider/spiderling/no_grow/S respectively

SpadesNeil:
Indeed, spiderlings in the infestation event was something I have previously removed for this very reason.

Ketrai:
Could do, will look into that once I got time.

Screemonster:
As long as they're non-breeding spiders, they're not all that much worse than solargrubs. Which are fine in their current state. Track the sucker down, two whacks with a fireaxe, ezpz. It's when they're the breeding sort that spiders go crazy.

SpadesNeil:
This is a problem but this is not how I want it fixed.

PR #2277, opened 2017-10-09
Removes access requirements from a few missed doors.

Cyantime:
Replaces doors leading out of the tram and to the surface three shuttlepad with public access ones.
I'm not sure how this isn't an issue on the main server already, but people are running into it when compiling from the current code for testing.


Neopatra:
While we're at it can we please change the security bathroom doors and all the other bathroom doors to public access too?

JoanRisu:
Which tram doors exactly?

Cyantime:
The two doors next to the spawnpoint, that lead into the rest of the station.

PR #2282, opened 2017-10-11
Improves solargrub larva

Cyantime:
Grub larva are now new mobs instead of spiderlings, they wiggle around, ventcrawl, and get into machines to drain power. After they drain enough power they grow into the bigger grubs.
Whacking a machine with a larva in it with a multitool pops the grub out.
Grubs and larva both glow slightly yellow and make sparks and noises occasionally, to make them a bit easier to find.

This also stops spiderlings from ventcrawling into the dorms, and stops grubs from spawning there.


SilverTalismen:
Now if only they would craw into FBPs...

Ketrai:
They already asked me if I were okay with this, and I am. So no need to ask me again ^^

SpadesNeil:
Where does the code check whacking with a machine?

Cyantime:
Bottom of the larva file, /obj/item/device/multitool/afterattack()

SpadesNeil:
Oh so they latch _externally_ onto the machine.

Cyantime:
The ~lore~ idea I had was that they crawl inside, and the multitool gives the entire machine an electropulse or something and the larva doesn't like that and pops out.

Cameron653:
Build rrestarted

SpadesNeil:
So how do you know they are in a machine?

Cyantime:
The machine has a flashing effect on it, and sparks somewhat often.
The area APC also has a much higher load on it than usual, which can be seen from a power monitering console.

Screemonster:
Do solargrubs affect the security status of a powernet, as shown on the power monitoring console?

SpadesNeil:
Whenever we can, I want a wiki page on solar grubs ASAP.

Ketrai:
Smol solargrubs probably won't trigger an alert, merely cause some anxiety on alert engineers. But the big grubs have and will always cause alerts focused around an area, especially with sub grids enabled. (which people really should enable, I've seen shifts where the station had a bit of a brown out coz of the grubs)

And I'll make a wiki page.

PR #2283, opened 2017-10-11
Adds toggleable hunger growls to vore panel.

Verkister:
-Disabled by default.
-One tile range as with the other bellynoises.
-10% chance to manifest when "extremely hungry", same but a little louder when "starving".


SpadesNeil:
I'm not a fan of the audio quality... Let me see if I can do something about that.

SpadesNeil:
Okay I can't fix these. Audio files are too short to get a noise profile.

Verkister:
Yeah might have to acquire some fresh clips. Gonna ask some folks if they got any good ones and maybe get some oc done heh.

Verkister:
How's this stuff?

SpadesNeil:
Much better, but still has some background noise 'hissing'.

https://www.podfeet.com/blog/recording/how-to-remove-noise-with-audacity/ Here is how to fix it.

Verkister:
I even went as far as to manually draw the waveform to remove a couple snap/click artifacts and i feel like some big time audio engineer now lmao

SpadesNeil:
Now we're talking!

These sound excellent.

PR #2288, opened 2017-10-13
Some ear stuff and hair tweaks.

Verkister:
-Added separate dual colorable vulpkanin ears.
-Converted colorable cat, fox, fennec, and mouse ears to dual color multiply format.
-Converted fox and solid mouse ears to multiply format.
-Made antlers multiply colorable.
-Applied the style-independent improved contrast tweaks to human_face_vr.


code/modules/vore/appearance/sprite_accessories_vr.dm
```diff
@@ -112,7 +112,7 @@
 	desc = ""
 	icon_state = "fenears"
 
-/datum/sprite_accessory/ears/sergal
+/datum/sprite_accessory/ears/sergal //Redundant
```
SpadesNeil: Redundant how?

Verkister: Essentially a white uncolorable dupe for the colorable version, which in the multiply format will look just same in white tint.

PR #2290, opened 2017-10-13
Port in chemistry bag from Paradise. Add two to chemistry pill closet.

Anasari:
As per title. This port in the chemistry bag which can hold bottle & pill from Paradise, and add two to the chemistry closet. 

No more clicking on 1000 bottles. Should also make stocking pills practical and chemistry less tedious.


Screemonster:
If it's a vorestation-specific edit, put the icon changes in a separate file with a _vr suffix, in this case storage_vr.dmi - changing the base polaris icon files breaks them if there's a change upstream.

Anasari:
Closing PR and pushing it to Polaris

PR #2291, opened 2017-10-13
Makes the metabolism traits affect movement drain and weight gain/loss.

Verkister:
-Overrides the carbon move proc for humans to use species/traits hunger factor instead of forced global default.
-Bigger hunger factor now burns more flab on running and smaller should burn less.
-Hopefully. Not sure if it actually manages to override the default proc values or just add to them. Anyway we're talking about heckin' miniscule numbers here, the default drain being 0.005 (default / 10) so the difference is next to nonexistent and therefore below the error margins I could detect on test runs hah.
-Yeah the default value means you'd have to run 100 tiles in order to lose 1 point of nutrition. For contrast, the normal nutrition level between hungry and glutted is 250 to 450.
-Moves our weight gain/loss specific metabolism var from mob to species.
-Makes the metabolism traits affect the forementioned var.
-Makes our weight gain/loss mechanics use the forementioned vars instead of one hard set to mob vars.

### DIFF
--- a/code/modules/mob/living/carbon/human/life_vr.dm
+++ b/code/modules/mob/living/carbon/human/life_vr.dm
@@ -50,3 +50,24 @@
 
 	//Process regular life stuff
 	nif.life()
+
+//Overriding carbon move proc that forces default hunger factor
+/mob/living/carbon/Move(NewLoc, direct)
+	. = ..()
+	if(.)
+		if(src.nutrition && src.stat != 2)
+			if(ishuman(src))
+				var/mob/living/carbon/human/M = src
+				if(M.stat != 2 && M.nutrition > 0)
+					M.nutrition -= M.species.hunger_factor/10
+					if(M.m_intent == "run")
+						M.nutrition -= M.species.hunger_factor/10
+			else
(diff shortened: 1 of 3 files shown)


Verkister:
Gonna do some math. According to my calculations;
-Currently the forced default movement drain value means **walking will get you 200 tiles far for 1 nutrition point. Running will take you 100 tiles far for 1 nutrition point.**
What comes to the stuff in this PR, in the intended case of overriding the values, the numbers would be;
**-250 tiles walking, 125 tiles running on slow metabolism.
-50 tiles walking, 25 tiles running on fast metabolism.
-20 tiles walking, 10 tiles running on apex.**
In the worse case where the values might stack instead of overriding, it'd still be;
**-100 tiles walking, 50 tiles running on default.
-111 tiles walking, 55 tiles running on slow metabolism.
-40 tiles walking, 20 tiles running on fast metabolism.
-18 tiles walking, 9 tiles running on apex.**
And all of this for 1 nutrition point, the middle ground between hungry and glutted being at 350 points :v

SilverTalismen:
Well let's hope this gets in before the update, these would be nice changes :D

PR #2293, opened 2017-10-13
Logs stepping on people.

Cameron653:
Actually logs when you step on or pin people down via foot.


Cameron653:
That should fix all the compile errors.

PR #2296, opened 2017-10-14
Crashing Fix Attempt.

Cameron653:
Instead of giving the game .1 second to apply decals, let's give it 5 seconds. This should be _more_ than enough time for it to properly apply the decals before the tiles start to get initialized. Or something. I have no idea what the fuck I'm doing, but a small delay doesn't hurt.

Tested and it doesn't break everything: https://i.imgur.com/efn0m4j.png
No guarantee if it actually fixes, but after a bunch of startups and restrarts it seems to not be crashing.

All I had was http://www.byond.com/forum/?post=2245438&page=2 to go off of along with it crashing on initializing tiles, so hopefully this works.

### DIFF
--- a/code/controllers/master_controller.dm
+++ b/code/controllers/master_controller.dm
@@ -71,7 +71,7 @@ datum/controller/game_controller/proc/setup_objects()
 		if(T.decals) T.apply_decals()
 		CHECK_SLEEP_MASTER
 	floor_decals_initialized = TRUE
-	sleep(1)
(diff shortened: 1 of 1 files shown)


Arokha:
This will not fix that problem, and will probably make the server take much longer to start.

SpadesNeil:
Yeah I wasn't sure this would do anything either, or it'd be merged by now.

Cameron653:
Well, there goes that idea. Any idea what is actually causing it to crash? Because as far as I can see in the forum post, it tries to overlay the image on a tile and crashes for some reason, but that's just a guess from what I'm seeing.

PR #2299, opened 2017-10-15
Adds in a "succubus" drain.

Cameron653:
- Adds in a new drain type.
- Slight adjustment to naming of bloodsucker drain name. Clarifies that it sucks a portion of their blood instead of draining them completely.
- Adds a toggle for the succubus drain.
Activated in the ability tab.
Two types:
Non-lethal (Only drains nutrition. Deals 100 halloss once all nutrition has been absorbed.)
Lethal (Drains nutrition and then lifeforce (deals brain damage.) until the hold is broken or the prey dies and is fully drained.)

Lethal Drain (for prey): https://i.imgur.com/K1fR41q.png
Lethal Drain (for pred): https://i.imgur.com/90COLwk.png


Screemonster:
Would be good for the lethal mode to respect the prey's digestion prefs, 'cause I can see people getting blindsided by this - especially if it's dealing halloss beforehand and they won't notice their health doll changing state.

Also halloss implies this is rather painful for the prey, and it'll push shock_stage up over time and lead to pain messages, rather than any kind of peacefully draining someone into a stupor until they pass out.

Dhaeleena:
How much does it take for someone to die from this?

Cameron653:
Hmm, I didn't think about the halloss causing pain messages. I'll go ahead and remove the halloss. Was mainly there to show "Hey, they haven't stopped yet"

I'll also add the digestable check.

And it depends on the amount of nutrition the prey has. The more  nutrition, the longer it'll take for it for them to be sucked dry

If someone has 10000 nutrition, it'll take 220 seconds to get them below 100 nutrition and then 60 (+5) seconds for them to be drained to death.

Screemonster:
What I'd suggest is something like calling paralyse() directly rather than relying on halloss to knock them out at the end. You could also apply effects like eye_blurry and the like.

This is what happens when halloss hits 100, _assuming_ the mob you're draining actually has a health of 100 (or less).
https://github.com/VOREStation/VOREStation/blob/c2a3f9a1f3631366f4453ec23aa50e28ef5f9331/code/modules/mob/living/carbon/human/life.dm#L943

>		if(halloss >= species.total_health)
>			src << "<span class='notice'>You're in too much pain to keep going...</span>"
>			src.visible_message("<B>[src]</B> slumps to the ground, too weak to continue fighting.")
>			Paralyse(10)
>			setHalLoss(species.total_health - 1)

Cameron653:
Went ahead and added in an eye_blurry that slowly stacks up over time. The longer the prey is drained, the longer they'll have blurry vision once the draining stops. Went ahead and changed the halloss to do their max health in halloss, so when they wake up they can move around (Makes halloss heal slower standing up) or lay there until it goes away, which is much faster.

SpadesNeil:
Conflict.

Cameron653:
@SpadesNeil Conflict fixed.

Cameron653:
DNM, adding one last thing.

Cameron653:
Done. Added a toggle so the death/KO doesn't come out of nowhere in the middle of a scene.

blazerules:
Would be neat if it said that the user felt an immense amount of pleasure or something seeing how it is a "succubus" drain. We all know what they drain ;)

Cameron653:
I knooow. I just wanted to make sure it was ambiguous enough for it to be able to be used for characters other than succubi.

PR #2300, opened 2017-10-15
Xenochimera tweaks

Screemonster:
A few tweaks for the bugs.

1. Internal bleeds are sealed at the start of their revive so they won't die while healing the thing they're trying to heal.

2. Adds a "purge impurities" button. The main thing it does is purge the shit out of infections because a being literally made from a colony of flesh-eating microbes shouldn't give a shit about an infected papercut. Takes a little while and has a 5-minute cooldown after use which also locks out their main regen. Depending on how much grossness they're purging, it might make them throw up.

3. Adjusts the threshold for going feral. It now scales with nutrition (capped at 600) so a well-fed monster can take up to 50 brute (up from 36) before the game starts rolling for feralness, while those on the brink of snapping from hunger anyway might flip out at a single punch.

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/station_special_abilities_vr.dm
+++ b/code/modules/mob/living/carbon/human/species/station/station_special_abilities_vr.dm
@@ -25,6 +25,13 @@
 				C << "You begin to reconstruct your form. You will not be able to move during this time. It should take aproximately [round(time)] seconds."
 				C.does_not_breathe = 1 //effectively makes them spaceworthy while regenning
 
+				for(var/obj/item/organ/E in C.bad_external_organs)
+					var/obj/item/organ/external/affected = E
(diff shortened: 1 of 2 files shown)


SilverTalismen:
Sounds awesome, now if only they could mend bones~ 

Haha but seriously this'll be nice!

Screemonster:
Well, if the other PR goes through, you'll be able to use the full hibernate/regen to fix them without ruining your icons again.

SilverTalismen:
LEt's hope so :D It's great to have a way to fix infections tho, for those times when Medical takes a hour and a half to get to me.

Screemonster:
If your organ is already dead from going necrotic, it won't save said organ. It doesn't repair damage to organs, so medical treatment (or a full revive) will be needed - just like spaceacillin will cure an infection, but won't repair the damage.

PR #2304, opened 2017-10-15
Adds 4 new Jukebox Tunes

Dustbiter:
Adds 4 new jukebox tunes, including a temporary spooky halloween track


SpadesNeil:
@Leshana will need to make these work.

We use a special code to stream Jukebox tunes that loads the files from a serverside host. This lets us bypass copyright laws because it ends up falling under fair use like a radio service or something. I forget all the legal shenanigans but Leshana did it that way for a reason.

Anyway talk to them about getting these songs loaded to the radio proper. We can't port them here but I'll leave this PR up so it gets attention.

Dustbiter:
Aw crud, whats the best way of getting ahold of Leshana? Are they on the discord?

Leshana:
For the new jukebox.json we'll want to know the artist as well as the title.

PR #2317, opened 2017-10-18
Updates Travis to 512.1392

Cameron653:
Updates Travis to 512.1392
Will compile code in 512.1392. Merge this whenever we update to 512.1392. If we update to something like 512.1397 later when it comes, comment here and I'll change it in this PR.


Leshana:
Well Well Well

PR #2324, opened 2017-10-18
Fixes omni filters

Cameron653:
Until we do another sync.


SpadesNeil:
// VORESTATION EDIT

Cameron653:
It's already fixed exactly the same way as Polaris. This'll make it so it (hopefully) doesn't cause a conflict.

SpadesNeil:
Pretty sure it won't work that way but whatevs. :U

Cameron653:
Well I hope it works! Else I'll have to go and fix a conflict and that's annoying >->

PR #2326, opened 2017-10-18
Makes soulcatcher actually purchasable for most jobs

Cameron653:
https://i.imgur.com/5Aq26T7.png
This really shouldn't happen. This makes it so it's cheap & all jobs can get it. It's an RP only item, so being next to free shouldn't be an issue.


SilverTalismen:
As Camilla I agree, its really expensive as it takes up 90% or all of your funds.

PR #2332, opened 2017-10-19
Adds a mind backup nifsoft and fixes sleevemate nif and quit issues.

Verkister:
-Adds a nifsoft that can be used for manual mind backups on the go sleevemate style.
-Fixes sleevemate being able to scan people who have quit the round. https://github.com/VOREStation/VOREStation/issues/2336
-Fixes one-time (first time) backups not saving nifs (sleevemate and this thing)
-Apparently the core backup proc only saved nifs on already backed up minds after the second sync.
-The nifsoft's currently priced at 250. Increase/decrease?
-Extremely useful for those skeletal and overly busy shifts where all the implanters stay tauntingly locked away in plain sight.
-Imo it'd actually make more sense for the nif to backup itself than for the implant to copy the nif anyway.
-Adds something actually useful and frequetly installable to wear out people's nifs faster.

To all concerned medical players afraid of losing their jobs because of this, don't worry, I'll be making up for it with some additional workplace hazard updates to give you some actual work to do instead of just being walking implant dispensers :v

### DIFF
--- a/code/__defines/nifsoft.dm
+++ b/code/__defines/nifsoft.dm
@@ -16,30 +16,31 @@
 #define NIF_UVFILTER			12
 #define NIF_FLASHPROT			13
 //Health-related
-#define NIF_CREWMONITOR			14
-#define NIF_ORGANIC_HEAL		15
-#define NIF_SYNTH_HEAL			16
-#define NIF_AUTOSTASIS			17 //These two are just part of
-#define NIF_MED_ALARM			18 //medichines right now
-#define NIF_TOXHEAL				19 //And this, for organics
-#define NIF_SPAREBREATH			20
+#define NIF_BACKUP				14
+#define NIF_CREWMONITOR			15
+#define NIF_ORGANIC_HEAL		16
+#define NIF_SYNTH_HEAL			17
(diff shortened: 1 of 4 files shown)


Gearshy:
Hmm? Hmm. To be clear: would this allow NIF'd people to just get the mind backup NIFsoft, forgo going to medical for a backup implant, and then keep their NIF on resleeve after a gurgly end?

Verkister:
Yeah pretty much. Should I up the price some? Currently at 250.
Just kinda tricky to get the proper implants done some times with skeletal or overly busy shifts with the implanters all locked away and such. What I could tell from looking at the sleevemate code was that it was supposed to backup the nif too, but the backup proc itself only had the nif part for automated follow-up syncs with none on the initial one.

Gearshy:
I think the price is okay as is. But then, I'm also for having an extra way to get backed up without setting foot in medbay. :P

SpadesNeil:
I think I already mentioned this to you but in case I didn't

When referencing a bug, say something like, "Fixes #0000" or whatever. This automatically closes the issue when the item is merged.

PR #2339, opened 2017-10-20
Removes the minimum limits to gurgle damage and scales the nutrition gain with them.

Verkister:
-Set the minimum gurgle damages to 0, the lower limit obsolete with the nutrition scaling. No damage, no food.
-A tick by tick digestion keeps the delay updated autoheal away just fine.
-Even a 0+1 digestion damage confirmed lethal for every race on a varedited test run so off the loo with the counterarguments.
-Also made the digestion nutrition gain scale with the gurgledamages so that's an even bigger blow against the "meta".
-Formerly 10/sizediff for default damage, now 2*(brute+burn)/sizediff.

### DIFF
--- a/code/modules/vore/eating/bellymodes_vr.dm
+++ b/code/modules/vore/eating/bellymodes_vr.dm
@@ -81,9 +81,9 @@
 					s_owner = owner
 					s_owner.cell.charge += 100
 				if(offset) // If any different than default weight, multiply the % of offset.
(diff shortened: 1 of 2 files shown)


Cameron653:
Seeing as autoheal doesn't happen if you're taking damage, I give this :+1:

Verkister:
I totally just realized that the gain scaling would make the thing entirely unabusable even with 0 damage gurgles. no damage, no nutrition

SpadesNeil:
>-Also made the digestion nutrition gain scale with the gurgledamages so that's an even bigger blow against the "meta".

This just fixed any problem I might have had with this and I wonder why I didn't think of that sooner.

PR #2340, opened 2017-10-20
Replaces live otie grabs with a different feature.

Verkister:
-Grabbing a live otie no longer does the classic "break sprite offset and piss off the mob" function.
-Instead it now just leads to a disapproving growl and a loss of bonding if the grabber had befriended the thing before.
-Now there is also a way to make them stop following you when you get tired of it.
-Grabbing a dead one does the old stuff tho so ya can still 'dispose' of the bodies or whatever.
-Serious there's always someone who goes grabbing the nonhostile ones fgshjd :v

### DIFF
--- a/code/modules/mob/living/simple_animal/vore/otie.dm
+++ b/code/modules/mob/living/simple_animal/vore/otie.dm
@@ -247,7 +247,7 @@
 
 	switch(M.a_intent)
 		if(I_HELP)
-			if (health > 0)
+			if(health > 0)
 				M.visible_message("<span class='notice'>[M] [response_help] \the [src].</span>")
 				LoseTarget()
 				handle_stance(STANCE_IDLE)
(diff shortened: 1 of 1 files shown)


SpadesNeil:
Not sure if loss of bonding is wise, since grabbing is a pretty standard thing to do.

SilverTalismen:
A suggestion I have is similar to what slimes do and be able to ask them to follow you.

Verkister:
The loss only makes it stop following the person and such. The grabbing is a pretty dumb thing to do with these to begin with though considering how the base code makes the mob consider it an attack to retaliate. It's like every time someone has a docile mob like this around, someone's gonna more or less accidentally grab it into attacking. Happened last round like twice.

SpadesNeil:
Yes but let's not make the bonding part lost when you grab them. Just have it growl and ignore the grab.

Verkister:
Alright even though the bonding thing only tells it who to follow around with no proper way to stop it other than passing the bond to the next person who pets it enough.

PR #2345, opened 2017-10-21
Reverts Ace's changes to make parapens functional

Yoshax:
Neurotoxin is an alcoholic drink that doesn't do as Ace likely expected when injected, because it is what it is. Drink affects like that of Neurotoxin or Beepsky Smash are applied upon ingest. Currently the parapen is non-functional.


SpadesNeil:
Used to instantly knock you down. :\

PR #2346, opened 2017-10-22
Brutal Predation & minor adjustments.

Cameron653:
- Adds "Brutal Predation" 
- Gives xenochimera & prometheans all the special verbs, since they can't use traits to get them. Allows them to use the verbs as appropriate.

Brutal Predation works as so:
0. Choose the Brutal Predation trait or be a xenochimera/promethean.
1. Get a neckgrab on someone.
2. Click the "Damage/Remove Prey's Organ" (HV) verb.
3. Select the external limb you wish to damage/remove.
4. It will then warn you if it's a vital limb or not. (Head is vital, for example.)
5. Click "yes" and it'll prompt you asking if you wish to damage/tear out an internal organ that's contained within that external organ (Groin contains the kidneys, for example.)
6. Select an internal organ or not.
7. It will then ask you if you wish to swallow the organ into your stomach, and give you a list of all your current organs you could swallow it into.
8. Select a stomach (or not).
9. The action will then start being performed. Takes 10 seconds.

The action itself:
1. If only an external organ was selected, it will damage it for 25 damage, or, if it already has 25 or more damage, it will tear the limb off (Provided it doesn't have the cannot_amputate var. If so, it just does 25 more damage.) and, if a stomach was selected, will eat it. Otherwise, the limbdrops onto the ground.
2. If an internal organ was selected, it will damage it for 25 damage or, if it already has 25 damage, it will tear the limb off and put it in a stomach if selected. Otherwise, the organ falls down onto the same tile the prey is on.

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/prometheans_vr.dm
+++ b/code/modules/mob/living/carbon/human/species/station/prometheans_vr.dm
@@ -18,8 +18,10 @@
 		/mob/living/proc/set_size,
 		/mob/living/carbon/human/proc/succubus_drain,
 		/mob/living/carbon/human/proc/succubus_drain_finialize,
-		/mob/living/carbon/human/proc/succubus_drain_lethal
-		)
+		/mob/living/carbon/human/proc/succubus_drain_lethal,
+		/mob/living/carbon/human/proc/bloodsuck,
+		/mob/living/carbon/human/proc/shred_limb
+		) //Prometheans get all the special verbs since they can't do traits.
 
 
 /datum/species/shapeshifter/promethean/handle_death(var/mob/living/carbon/human/H)
--- a/code/modules/mob/living/carbon/human/species/station/prometheans_vr.dm
+++ b/code/modules/mob/living/carbon/human/species/station/prometheans_vr.dm
@@ -18,10 +18,8 @@
 		/mob/living/proc/set_size,
 		/mob/living/carbon/human/proc/succubus_drain,
 		/mob/living/carbon/human/proc/succubus_drain_finialize,
-		/mob/living/carbon/human/proc/succubus_drain_lethal,
-		/mob/living/carbon/human/proc/bloodsuck,
-		/mob/living/carbon/human/proc/shred_limb
-		) //Prometheans get all the special verbs since they can't do traits.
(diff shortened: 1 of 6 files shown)


SpadesNeil:
Uhh, maybe we shouldn't give prometheans the "Brutal Predation" thing.

Sergals on the other hand...

code/modules/mob/living/carbon/human/species/station/prometheans_vr.dm
```diff
@@ -18,8 +18,10 @@
 		/mob/living/proc/set_size,
 		/mob/living/carbon/human/proc/succubus_drain,
 		/mob/living/carbon/human/proc/succubus_drain_finialize,
-		/mob/living/carbon/human/proc/succubus_drain_lethal
-		)
+		/mob/living/carbon/human/proc/succubus_drain_lethal,
+		/mob/living/carbon/human/proc/bloodsuck,
+		/mob/living/carbon/human/proc/shred_limb
```
SpadesNeil: Yeeeeaaaahhh Prometheans aren't gonna get this one.

Cameron653:
Well, in terms of prometheans they'd be melting the limb. They were given it in case they wanted to, say, "melt" off someone's arm into their gooey body. Sergals weren't given it since they have nothing special when it comes to the species system, and their body style can be selected as a custom species.

code/modules/mob/living/carbon/human/species/station/prometheans_vr.dm
Cameron653: I can remove the verbs from them. They were there mostly for "In case they want to use these fluff verbs" but I can do that in another PR later down the road, as to not conflict with what this current PR is really for.

Cameron653:
@SpadesNeil Done. The verbs have been removed from the prometheans.

SpadesNeil:
Give shred to sergals tho.

Virgo113:
I support this motion on totally unbiased grounds.

Cameron653:
Bah. Eventually! For now sergals can use the custom species w/ sergal base.

SpadesNeil:
_do it now_ C:

PR #2348, opened 2017-10-22
Resleever grower pod processes meat automatically

Anasari:
This let resleever's growing pod processes meat automatically next to it for 50 unit of biomass each. Also include meatpie and such because of code weirdness. Shouldn't cause an issue.

(Doesn't change cloning pod as that'll cause issue when synced with upstream)

### DIFF
--- a/code/modules/resleeving/machines.dm
+++ b/code/modules/resleeving/machines.dm
@@ -106,9 +106,13 @@
 
 /obj/machinery/clonepod/transhuman/process()
 
+	var/visible_message = 0
 	for(var/obj/item/weapon/reagent_containers/food/snacks/meat in range(1, src))
 		qdel(meat)
 		biomass += 50
+		visible_message = 1 // Prevent chatspam if multiple meat are near
+
+	if(visible_message)
 		visible_message("[src] sucks in and processes the nearby biomass.")
 
 	if(stat & NOPOWER)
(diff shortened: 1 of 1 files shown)


Cameron653:
Does this do all of them at once? If so, would the message result in a huge wall of chatspam?

Anasari:
Seems like it does. Lemme fix that.

Anasari:
Fixed!

Cameron653:
:+1:

Screemonster:
Does it just make 50 biomass per stack, as opposed to counting the amount of reagents in the stack? 'cause I'm just picturing ~~scree~~ some fatty doctor taking a bite out of each slab-o-meat without it affecting the biomass

Cameron653:
As far as I can see, just 50 biomass per slab. Nobody really ever implemented it to be based on reagents, as even the old cloning machine just adds 50 bio
https://github.com/VOREStation/VOREStation/blob/163c686d183ca817998f34b6d9362cc29b93f7a8/code/game/machinery/cloning.dm#L244-L248

Now, it would be neat to have it based on reagents, though. Inject meat with extra nutrition for more biomass. Although that might be outside the scope of the PR

Anasari:
I think that case will be quite rare enough to not matter that much.

Besides it's hilarious.

PR #2353, opened 2017-10-22
Fixes and Tweaks to Micro Stomping

Razer69:
Takes care of a few issues I noticed with how micro bumping is handled.

 - Fixes some spelling issues.
 - Fixes a wrong return value.
 - Wraps some visible messages in css classes.
 - Micros now get pushed to the ground immediately rather than waiting for the next life tick.
 - ~~Bumping micros now also gets triggered when they are prone.~~
 - Micros that aren't on help intent do no longer block bigger people that are.
 - Now uses relative size difference (instead of absolute) for checks and damage dealing. Assailant needs to be at least 30% bigger for any actions.

Did some very minor code cleanup since I had the chance. This entire proc consists of a lot of copypaste and some hacked in code, I might look into rewriting this entire thing in the future so it can be easily added upon and it's less easily broken when touched.

### DIFF
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -113,14 +113,15 @@ default behaviour is:
 			if(!can_move_mob(tmob, 0, 0))
 				now_pushing = 0
 				return
-			if(a_intent == I_HELP || src.restrained())
-				now_pushing = 0
-				return
 
 			// VOREStation Edit - Begin
 			// Handle grabbing, stomping, and such of micros!
 			if(handle_micro_bump_other(tmob)) return
 			// VOREStation Edit - End
+
+			if(a_intent == I_HELP || src.restrained())
+				now_pushing = 0
+				return
 
 			if(istype(tmob, /mob/living/carbon/human) && (FAT in tmob.mutations))
 				if(prob(40) && !(FAT in src.mutations))
--- a/code/modules/vore/resizing/resize_vr.dm
+++ b/code/modules/vore/resizing/resize_vr.dm
@@ -109,7 +109,7 @@ var/const/RESIZE_A_SMALLTINY = (RESIZE_SMALL + RESIZE_TINY) / 2
  * @return false if normal code should continue, 1 to prevent normal code.
  */
 /mob/living/proc/attempt_to_scoop(var/mob/living/carbon/human/H)
-	var/size_diff = H.get_effective_size() - src.get_effective_size()
+	var/size_diff = src.get_effective_size() / H.get_effective_size()
 	if(!src.holder_default && src.holder_type)
 		src.holder_default = src.holder_type
 	if(!istype(H))
@@ -117,7 +117,7 @@ var/const/RESIZE_A_SMALLTINY = (RESIZE_SMALL + RESIZE_TINY) / 2
 	if(H.buckled)
 		usr << "<span class='notice'>You have to unbuckle \the [H] before you pick them up.</span>"
 		return 0
-	if(size_diff >= 0.50)
+	if(size_diff >= 2)
 		src.holder_type = /obj/item/weapon/holder/micro
 		var/obj/item/weapon/holder/m_holder = get_scooped(H)
 		src.holder_type = src.holder_default
@@ -137,7 +137,7 @@ var/const/RESIZE_A_SMALLTINY = (RESIZE_SMALL + RESIZE_TINY) / 2
 		// Both small! Go ahead and
 		now_pushing = 0
 		return 1
-	if(abs(src.get_effective_size() - tmob.get_effective_size()) >= 0.50)
+	if(abs(src.get_effective_size() / tmob.get_effective_size()) >= 1.3)
 		now_pushing = 0
 		if(src.get_effective_size() > tmob.get_effective_size())
 			var/mob/living/carbon/human/H = src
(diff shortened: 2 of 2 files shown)


Ketrai:
Stomping on micros that are lying down was never intentional I think, ask jasmine. or find the orginal pr for reference to double check.

Razer69:
Not claiming it was intentional, but now it's definitely possible. Didn't make much sense before that once a micro was prone, it'd be impossible for you to do this.

Ketrai:
It think it was to avoid killing a micro too easily.

Razer69:
Right now, you can already kill a micro easily be stepping on them with 'walk' and 'hurt' intent as many times as you want before they drop to pain- or hardcrit.

Cameron653:
The reason I made it so micros couldn't be stepped on while down is so you can't go and run over them with disarm intent to shove them to the ground, swap to walk & harm intent and kill them.

Razer69:
Wouldn't that be powergaming and worse, griefing, on an otherwise 24/7 extended server? You can already kill someone by stepping on them two or three times with walk intent when they aren't expecting it. On the other hand, this now makes you able to use those functions in RP when people are prone, which has been a missing feature for me before.

SpadesNeil:
Mechanics should reflect what we want to see in RP. Enforcing rules when mechanics encourage otherwise is bad practice.

Razer69:
A small cooldown would safely prevent any abuse of mechanics by 'spamwalking' I guess, making it safer than it was before.

SpadesNeil:
Separate fixes verses changes to intended features.

Razer69:
The contentious change has been removed from this PR.

Screemonster:
As per the discussion in issue #2398 - harm damage is pretty brutal right now. It's not a lot of damage to each limb, but it adds up to something like 10-15 brute, putting it on a par with things like circular saws - and that's for running, not walking. This PR, according to the comments, actually raises the maximum damage per limb.

Razer69:
Due to taking relative size difference in account now rather than absolute (which only made sense before where the size options were discrete rather than continuous between 25% and 200%), the numbers changed a little.

Average damage by running changed from 15.625 to 17.5 and average damage by walking changed from 54.6875 to 52.2. These are calculated for a 200% against a 25% character. For half the size, it'd be half of that, ect. These changes in numbers are miniscule and an artifact of using relative measurements rather than simply subtracting one size from the other.

Looking over it, I guess I can change the damage multiplier for running to 2.5 though, making the average damage 15 and even closer to what it was before.

Razer69:
Bump.

SpadesNeil:
Crap, uh, I forgot about this. Anything else we need to go over?

Razer69:
From my side it's all ready to go, testing was sufficiently done.

code/modules/mob/living/living.dm
```diff
@@ -130,6 +127,10 @@ default behaviour is:
 					return
 			// VOREStation Edit - End
 
+			if(a_intent == I_HELP || src.restrained())
```
Cameron653: These lines of code changed should have //VOREStation edit on it if they're moved.

code/modules/vore/resizing/resize_vr.dm
```diff
@@ -112,15 +112,15 @@ var/const/RESIZE_A_SMALLTINY = (RESIZE_SMALL + RESIZE_TINY) / 2
  * @return false if normal code should continue, 1 to prevent normal code.
  */
 /mob/living/proc/attempt_to_scoop(var/mob/living/carbon/human/H)
-	var/size_diff = H.get_effective_size() - src.get_effective_size()
+	var/size_diff = src.get_effective_size() / H.get_effective_size()
```
Cameron653: Looking at the math now.

src.get_effective_size() / H.get_effective_size()

SRC is the mob trying to pick up someone. H is the person being picked up.

The minimum size to pick someone up is 50% size difference.
NEW version: (If size diff is >= 2, pick them up)
2.0/1.5 = 1.333
1.5/1 = 1.5
1.0/0.5 = 2.0
0.75/0.25 = 3.0

OLD version: (If size diff is >= .5, pick them up)
2.0-1.5 = 0.5
1.5-1 = 0.5
1.0-0.5 = 0.5
0.75-0.25 = 0.5


If I'm reading this correctly, this would mean that large characters can't pick up people that are ~50% of them, making the micro pickup/crush code not work.

code/modules/vore/resizing/resize_vr.dm
```diff
@@ -180,82 +180,94 @@ var/const/RESIZE_A_SMALLTINY = (RESIZE_SMALL + RESIZE_TINY) / 2
...
+					src << "You push [tmob] to the ground with your foot!"
+					tmob << "<span class='warning'>[src] pushes you down to the ground with their foot!</span>"
+				log_and_message_admins("has pushed [tmob] to the ground by stepping on them.") //Both humans and mobs, since stepping on mobs can be abused.
 				admin_attack_log(src, tmob, "Pinned [tmob.name] under foot.", "Was pinned under foot by [src.name].", "Pinned [tmob.name] under foot.")
 				return 1
-			if(src.m_intent == "walk") //Most likely intentionally stepping on them.
-				var/size_damage_multiplier = (src.size_multiplier - tmob.size_multiplier)
-				var/damage = (rand(15,30)* size_damage_multiplier) //Since stunned is broken, let's do this. Rand 15-30 multiplied by 1 min or 1.75 max. 15 holo to 52.5 holo, depending on RNG and size differnece.
+			else // Most likely intentionally stepping on them.
+				var/size_damage_multiplier = 0.25 * (src.size_multiplier / tmob.size_multiplier)
```
Cameron653: Before:
var/size_damage_multiplier = (src.size_multiplier - tmob.size_multiplier)
Max of 1.75, min of 0.5

New:
var/size_damage_multiplier = 0.25 * (src.size_multiplier / tmob.size_multiplier)
2.0/1.5 = 1.333 * 0.25 = .325
1.5/1 = 1.5 * 0.25 = 0.375
1.0/0.5 = 2.0 * 0.25 = 0.5
0.75/0.25 = 3.0 * 0.25 = .75
Min of .325 and max of .75. This means damage from halloss is greatly nerfed. Not only that, but 2.0 can't step on 2.0 can't step on 1.5, and 1.5 also can't step on 1.0 as shown above.

code/modules/vore/resizing/resize_vr.dm
```diff
@@ -180,82 +180,94 @@ var/const/RESIZE_A_SMALLTINY = (RESIZE_SMALL + RESIZE_TINY) / 2
...
-			now_pushing = 0
+	else if(src.a_intent == I_HURT)
+		if((src.get_effective_size() / tmob.get_effective_size()) >= 2)
 			src.forceMove(tmob.loc)
-			var/size_damage_multiplier = (src.size_multiplier - tmob.size_multiplier)
-			var/damage = (rand(1,3)* size_damage_multiplier) //Rand 1-3 multiplied by 1 min or 1.75 max. 1 min 5.25 max damage to each limb.
-			var/calculated_damage = damage/2 //This will sting, but not kill. Does .5 to 2.625 damage, randomly, to each limb.
+			now_pushing = 0
+			tmob.Weaken(4)
+			var/size_damage_multiplier = 0.25 * (src.size_multiplier / tmob.size_multiplier)
```
Cameron653: Same as above.

code/modules/vore/resizing/resize_vr.dm
```diff
@@ -180,82 +180,94 @@ var/const/RESIZE_A_SMALLTINY = (RESIZE_SMALL + RESIZE_TINY) / 2
...
-							tmob.take_overall_damage(calculated_damage, 0) // 5 damage min, 26.25 damage max, depending on size & RNG. If they're only stepped on once, the damage will heal over time.
+							tmob.take_overall_damage(calculated_damage, 0) // 5 damage min, 30 damage max, depending on size & RNG. If they're only stepped on once, the damage will heal over time.
 						M.drip(0.1)
 						admin_attack_log(src, M, "Crushed [tmob.name] under foot for [damage * 10] damage.", "Was crushed under foot by [H.name] for [damage * 10] damage.", "Crushed [M.name] for [damage * 10] damage.")
 				return 1
 
-			if(src.m_intent == "walk") //Oh my.
-				damage = calculated_damage * 3.5 //Multiplies the above damage by 3.5. This means a min of 1.75 damage, or a max of 9.1875. damage to each limb, depending on size and RNG.
+			else //Oh my.
+				damage = calculated_damage * 3 //Multiplies the above damage by 3. This means a min of 1.5 damage, or a max of 9. damage to each limb, depending on size and RNG.
```
Cameron653: Minor damage decrease.

code/modules/vore/resizing/resize_vr.dm
```diff
@@ -264,27 +276,30 @@ var/const/RESIZE_A_SMALLTINY = (RESIZE_SMALL + RESIZE_TINY) / 2
 						admin_attack_log(src, M, "Crushed [M.name] under foot for [damage * 10] damage.", "Was crushed under foot by [H.name] for [damage * 10] damage.", "Crushed [M.name] for [damage * 10] damage.")
 				return 1
 
-	if(src.a_intent == I_GRAB && src.canmove && !src.buckled)
-		if((src.get_effective_size() - tmob.get_effective_size()) >= 0.50)
-			now_pushing = 0
+	else if(src.a_intent == I_GRAB)
+		if((src.get_effective_size() / tmob.get_effective_size()) >= 1.5)
```
Cameron653: Same as above math. 2.0 can't pick up a 1.5.
2.0/1.5 = 1.333

code/modules/vore/resizing/resize_vr.dm
Razer69: Anyone can pick up mobs that are at most half the size of themself with this change, simply put.

code/modules/vore/resizing/resize_vr.dm
Razer69: Your calculations are a bit odd here.
```
Maximum: 2   = 0.25 * max_size/min_size
Minimum: 0.5 = 0.25 * 2

The fact that a 2.0 can't step on a 1.25 is lent to the fact that the size now requires to be twice the value. Back when using discrete values however, the size of 1.25 did not exist, so applying the same logic to it does not make sense here.

File: code/modules/vore/resizing/resize_vr.dm (hunk 6 above)
Comment (Razer69) [reply code/modules/vore/resizing/resize_vr.dm]: What coefficient are you suggesting?

File: code/modules/vore/resizing/resize_vr.dm (hunk 5 above)
Comment (Razer69) [reply code/modules/vore/resizing/resize_vr.dm]: Use the calculated average for comparison, the difference is negligible.

File: code/modules/vore/resizing/resize_vr.dm (hunk 4 above)
Comment (Razer69) [reply code/modules/vore/resizing/resize_vr.dm]: As above, your calculated values for maximum and minimum damage are wrong.

File: code/modules/mob/living/living.dm (hunk 1 above)
Comment (Razer69) [reply code/modules/mob/living/living.dm]: They shouldn't, all that was done here was that he edit was moved.

Comment (Cameron653) [reply code/modules/mob/living/living.dm]: Ah, very well. Was just wondering why it showed the line being modified.

File: code/modules/vore/resizing/resize_vr.dm (hunk 2 above)
Comment (Cameron653) [reply code/modules/vore/resizing/resize_vr.dm]: >Anyone can pick up mobs that are at most half the size of themself

I can't support this. 200% should be able to pick up 150% and 150% should be able to pick up 100%. If they can't do that, this PR has absolutely no support from me.

File: code/modules/vore/resizing/resize_vr.dm (hunk 3 above)
Comment (Cameron653) [reply code/modules/vore/resizing/resize_vr.dm]: >now requires to be twice the value

I really don't like this. This isn't a tweak, but a complete change.

The calculations I did to calculate the damage are fine as long as I can see. I changed it to be (src.size_multiplier / tmob.size_multiplier) * 0.25 to make it easier to read.

File: code/modules/mob/living/living.dm (hunk 1 above)
Comment (Cameron653) [reply code/modules/mob/living/living.dm]: Mm, very well. Was wondering why github referenced this.

File: code/modules/vore/resizing/resize_vr.dm (hunk 2 above)
Comment (Cameron653) [reply code/modules/vore/resizing/resize_vr.dm]: >Anyone can pick up mobs that are at most half the size of themself with this change, simply put.

# This is a hidden and massive nerf that wasn't mentioned in the PR description. 

Not once did it mention that in the PR. The closest thing you said was "Now uses relative size difference (instead of absolute) for checks and damage dealing."

This isn't "fixes and tweaks" this is just a flat out nerf. This means that 200% can't pick up 150, and 150% can't pick up 100%

# 200% ALSO can't step on 101%

This huge nerf was hidden and not mentioned at all, which **is not fucking OK**

File: code/modules/vore/resizing/resize_vr.dm (hunk 3 above)
Comment (Cameron653) [reply code/modules/vore/resizing/resize_vr.dm]: Redid the math, I see the mistakes I made here.

.25 * (2.0/1.5) = .33333333333 (Unable to do so with this nerf.)
.25 * (2.0/1.0) = .5
.25 * (2.0/.75) = .6666666666
.25 * (2.0/.5) = 1
.25 * (2.0/.25) = 2
.25 * (1.5/1.0) = .375 (Unable to do with this nerf)
.25 * (1.5/.75) = .5
.25 * (1.5/.5) = .75
.25 * (1.5/.25) = 1.5
.25 * (1.0/.5) = .5
.25 * (1.0/.25) = 1
.25 * (.75/.25) = .75
.25 * (.5/.25) = .5

So the lower damage is still the same, and the upper damage is slightly buffed. This I'm fine with, but _the fact 200% can't pick up 150% or _anyone above 100%, including 101% for that matter_ is fucking sneaky and not ok whatsoever.

File: code/modules/vore/resizing/resize_vr.dm (hunk 4 above)
Comment (Cameron653) [reply code/modules/vore/resizing/resize_vr.dm]: Fixed them.

File: code/modules/vore/resizing/resize_vr.dm (hunk 6 above)
Comment (Cameron653) [reply code/modules/vore/resizing/resize_vr.dm]: Yes. 1.3 would work, as it'd allow 200 to pick up 150, and 150 to pick up 100.

File: code/modules/vore/resizing/resize_vr.dm (hunk 2 above)
Comment (Razer69) [reply code/modules/vore/resizing/resize_vr.dm]: I didn't hide the fact that the system was swapped to using a relative difference calculations. It implies that there will be changes.

What coefficient is used is completely open to debate and you can make suggestions to it.

No need for using title fonts, I can read.

File: code/modules/vore/resizing/resize_vr.dm (hunk 6 above)
Comment (Razer69) [reply code/modules/vore/resizing/resize_vr.dm]: I initially used a quotient of 2 to keep it intuitive, but for the sake of making 200% to 150%-interaction possible, I suppose 1.3 is okay.

File: code/modules/vore/resizing/resize_vr.dm (hunk 2 above)
Comment (Cameron653) [reply code/modules/vore/resizing/resize_vr.dm]: The issue with making it check for pred_heigh/prey_height is that it causes extremely odd interactions across the board, while having it check for 50% difference worked well  and provided enough of a size difference.

200/150 = 1.33. 200/150 = 1.5 Making it 1.3 works perfectly in these two cases. 
200/153 = 1.307. This means 200% can pick up _just_ a bit higher than before. Not that big of a change.

But as you start to get lower sizes, things start to get crazy.

0.100/0.76 = 1.31. 100% would be able to pick up 76%, and there's only a 24% size diff here.
0.75/0.57 = 1.31 75% would be able to pick up 57%, which is only a 18% size difference.
0.50/0.38 = 1.31 50% would be able to pick up 38%, which is again a 12% size difference
0.33/0.25 = 1.32 33% would be able to pick up 25%, which is only a 8% size difference.

Trying to get it to work with relative size is good, but it just... Really doesn't work out. If you increase it, 200% can't pick up 150%. Having it be a 50% size difference makes it so it's even across the board and you don't get crazy things like, say, a micro managing to step on another micro when they're almost the exact same size.

Also, I do apologize for the boldfont. I was under the impression that the size changes to make it so, say, 200% can't step on anyone above 100% were being snuck in intentionally without including info about it.

Comment (Razer69) [reply code/modules/vore/resizing/resize_vr.dm]: I understand the sentiment. Though a set difference of 0.75 makes it so you can only step on 0.25 when normal-sized, and not on anything when being the slightest bit smaller.

If sizes past 200% were allowed the interactions among large characters would become quite extreme as well, but since those sizes will most likely remain inaccessable I guess the model is good enough for that.

So, since 200% being able to interact with 150% is apparently wanted, how about a coefficient of 1.3 for a victim of size 100% or greater, and 2.0 for victims below that size?

Comment (Cameron653) [reply code/modules/vore/resizing/resize_vr.dm]: Doing 1.3 for 100% or greater and 2.0 for victims below 100% could work, but then the issue of 99% requiring someone to be 198 and such arises, which would result in a whole slew of issues. Maybe make it so... 
>= 75% = 1.3 (Looking back at this, it could possibly be increased to 1.4. Make it so 200 can pick up <= 140, resulting in a small decrease in pickup size while not lowering it too much.) 
< 75 && >50 (This is where it's somewhat annoying to get the calculations.) =1.6
<= 50% = 2.0

Comment (Razer69) [reply code/modules/vore/resizing/resize_vr.dm]: On second thought, at this point the mathematical model used is becoming unnecessary convoluted. It should be kept at a simple x % size advantage or x size difference.

**Comment by Arokha** [comment]
Considering how often you're using it, you might want to consider just making a
`#define SIZE_DIFF_RATIO 1.3`
`#define SIZE_DIFF_CHECK(x,y) ((x.get_effective_size * y.get_effective_size) >= SIZE_DIFF_RATIO)`

**Comment by Arokha** [comment]
Has conflicts, can't be merged as-is, has undocumented changes, needs additional changes.

Given the history of this PR, I'm closing it. If you decide to reconcile the above issues, reopen it as a new PR.

File: code/modules/vore/resizing/resize_vr.dm [hunk 7]
```diff
@@ -257,27 +267,30 @@ var/const/RESIZE_A_SMALLTINY = (RESIZE_SMALL + RESIZE_TINY) / 2
...
 				else
 					src << "You pin [tmob] down onto the floor with your foot and curl your toes up around their body, trapping them inbetween them!"
-					tmob << "[src] pins you down to the floor with their foot and curls their toes up around your body, trapping you inbetween them!"
+					tmob << "<span class='warning'>[src] pins you down to the floor with their foot and curls their toes up around your body, trapping you inbetween them!</span>"
 			else if(istype(H) && istype(H.tail_style, /datum/sprite_accessory/tail/taur/naga))
 				src << "You squish [tmob] under your large, thick tail, forcing them onto the ground!"
-				tmob << "[src] pins you under their large, thick tail, forcing you onto the ground!!"
+				tmob << "<span class='warning'>[src] pins you under their large, thick tail, forcing you onto the ground!!</span>"
 				tmob.resting = 1
+				tmob.Weaken(1)
```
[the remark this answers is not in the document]
Razer69: It forces a direct update of the mob's state, whereas only modifying a variable will only take effect on the next life tick.

Probably going to open a clean PR with only a subset of changes, as this PR turned into a huge mess.

Issue #2355, opened 2017-10-22
Custom species and lost limbs


SilverTalismen:
#### Brief description of the issue
Custom species don't tell you what limb they are missing

#### What you expected to happen
For it to tell me what arm or leg was missing

#### What actually happened
https://gyazo.com/f2e791e984fbaa8070b1bfef96932c5d

#### Steps to reproduce
- (Step 1) Be a custom species
- (Step 2)lose a arm somehow
- (and so on)

#### Code Revision
- dcddd0d64a82f76e991e166c8d9dc3893690fbe9
 
#### Anything else you may wish to add:
- (Location if it's a mapping issue, screenshots, sprites, etc.)

PR #2359, opened 2017-10-23
Fixes missing limbs showing no names and shit on custom species.

Verkister:
- Yeah found out the reason, pretty clear by logic after figuring it out, but still somehow functional for normal species apparently(itjustworks.jpg)???
- The stuff was relying on some sort of "descriptor" in the has_limbs list that does not seem to even exist.
- Adding this "descriptor" to custom species limbs list reliably fixed the issue on test runs.
- No fucking clue how the other species somehow manage without that thing however.

kuva

Fixes https://github.com/VOREStation/VOREStation/issues/2355

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/blank_vr.dm
+++ b/code/modules/mob/living/carbon/human/species/station/blank_vr.dm
@@ -28,6 +28,20 @@
 
 	var/list/traits = list()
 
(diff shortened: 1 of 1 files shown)


SpadesNeil:
In future just say:

`- Fixes #IssueNumber`

PR #2374, opened 2017-10-27
Makes bloodbags have 10 uses.

Cameron653:
Makes bloodbags have 10 uses for vampires and not provide them with enough nutrition to fuel everything ever.


Cameron653:
I'm just going to merge this so I can start on the sync. It's a minor bugfix.

PR #2378, opened 2017-10-27
Revert "Just some contrast adjustments on hair and teshari bases."

SpadesNeil:
Reverts VOREStation/VOREStation#2333

These are all done wrong.


SpadesNeil:
Nevermind, this is so much of a fucking mess now that I can't change it through reverts anymore.

@Verkister **You've been doing all of these sprites wrong for everything you've done so far regarding color changing.** So stop.

The correct way of doing it is the _base_ sprite is supposed to be **as black as you ever want it to be**, and we do not use color multiply anymore because it's a horrible system. Apparently though you've changed **every species** to use this system. This is completely incorrect and should be changed back immediately once we have the correct sprites for it.

Hair sprites for humans are already correctly configured for this method.

PR #2380, opened 2017-10-27
Polaris Sync

Cameron653:
Fixes job access
Gives whistle, shuffling cards, and clapping a sound.
Captain's spare is a new path.
Other stuff I missed


Anewbe:
Yeah, sorry 'bout the bug, didn't think anyone actually used that config.

PR #2386, opened 2017-10-28
Organizes alien whitelist in alphabetical order.

Cameron653:
- Makes alien whitelist alphabetical order.
- Removes capital letters in the ckeys. (This isn't required as the config files say, just looks nicer.)


Cameron653:
This changes literally nothing other than organization. Merging.

PR #2387, opened 2017-10-28
Allows flight with wings.

Cameron653:
- Allows flight with wings and a 2 point trait. (Xenochimera, Nevrean, and Harpies have it by default.)
- Makes you bob while flying
- Allows you to pick people up into the air while flying and they won't fall
- Allows you to drop people from the air while flying.
- Allows you to grab someone as they jump over a railing/climb over a railing to prevent them from dropping. 
- Base Polaris code touched as little as possible. (Which, had to be touched quite a lot due to the code requiring a lot of procs to be adjusted.) Went ahead and left //VOREStation Edits on all parts of code with edited lines.
- Example: https://i.imgur.com/7KCq56H.gifv
( The above gif has the skrell being dragged along due to resting. If it wasn't resting, it'd be held up against the flyer.)
- Stops people from floating on open Z tiles. If someone is on an open space and they shouldn't be, they'll fall (Fall proc added to life to check this.) (This was added to make people fall if they get tased while flying midair, but has the added benefit of fixing the floating bug.)
- Makes it so if someone flying is tased or incaped in the air, they fall down.
- Fixes a minor typo. (There was a comma in an incorrect place.)
- Has a max nutrition and min nutrition needed to fly. Otherwise, you can't fly or will stop flying midair.
- You "glide" while wings are enabled. https://i.imgur.com/yp8j3IL.gifv
- You lose 2 nutrition per tile you fly, along with 2 nutrition per tick.
- You can use a hover proc in exchange for 25 nutrition. This stops you from gliding momentarily, allowing you to fly upwards.
- Var moved to /mob instead of /human to allow porting
- Makes it so if you're flying, you will bump into stuff in your way,


Yoshax:
I'm not the biggest fan of this, to be quite honest. I don't really feel like it adds anything, and could hinder some things going forwards. Especially in the current where we have an on-going mapping competition for away missions. I know I brainstormed with some people and some people had the ideas to use verticalness to some degree to create puzzles, or interesting stuff. The possibility of which would be dramatically decreased and any thought required removed because people can fly.

Not only that, but it would allow people to access places where others cannot, and cannot do much about it. It could become a massive pain for Security, and for people to just be asses with in general. I don't really what we need it, or what it would add. 

Further, it not only adds very little, but touches a lot of base Polaris code, which seems to be becoming more and more common for things of little impact.

code/modules/multiz/movement.dm
```diff
@@ -49,6 +49,23 @@
 			else
 				to_chat(src, "<span class='warning'>You gave up on pulling yourself up.</span>")
 				return 0
+		else if(ishuman(src)) //VOREStation Edit Start. Are they human (H above.), have a wing style, and do they have the trait?
+			var/mob/living/carbon/human/H = src //VOREStation Edit. Used for below
+			if(H.wings_flying == 1)
+				if(H.stat || H.paralysis || H.stunned || H.weakened || H.lying || H.restrained() || H.buckled)
```
Yoshax: if(H.incapacitated())

Should work and would be neater.

code/modules/multiz/movement.dm
```diff
@@ -140,14 +157,23 @@
 
 	if(throwing)
 		return
+	if(ishuman(src)) //VOREStation Edit Start. Code to see if they have wings.
+		var/mob/living/carbon/human/H = src
+		if(H.wings_flying == 1) //Some other checks are done in the wings_toggle proc
+			if(H.stat || H.paralysis || H.stunned || H.weakened || H.lying || H.restrained() || H.buckled)
```
Yoshax: if(H.incapacitated())

Should work and would be neater.

Cameron653:
The security problem is solved by making it a long (7 seconds) time before moving upwards. This can be adjusted if need be.

The mapping contest is completely out of my control. If a map  really needs no flight zones, they could add an area var named "no_flight" to areas and make the move up proc check that to prevent movement.

Most areas are easily accessable. This would mainly allow movement up and down the atrium (Which can be accessed by removing a window and placing down lattices) or moving up to different levels of the outpost Z level (Which can be done via the stairwell.) I can't think of any other places that are inaccessible on the map.

People requested a feature to allow them to fly, so I added it. If someone wants to have a kinky in the air RP, so be it. Additionally, people like engineers could more easily maneuver to areas in order to fix breaches or do repairs. (This is one of the reasons for its point cost.)

The touching Polaris code was unavoidable. I did as best I could to stay out of the procs, but quickly found it unavoidable. It isn't much of an issue seeing as I do the syncs. If a conflict occurs, I'll fix it or (If worst comes to worst, which is highly unlikely.) I'll disable the flying code and go about it another angle.

Yoshax:
Saying it's fine because you do the syncs is a poor comment. There's no guarantee you'll be doing it next month, or the one after that, or the three after that. Leshana did them before, Aro did before, neither do now. That just shows what I've said. 

I'm just not sure I see a reason for this.

Cameron653:
I'm not saying it's fine just because I do the sync. I'm saying it's fine because there is physically no other way (As far as I can see) to do this without injecting the flying code into multi-z code as this does, along with the fact it can easily be disabled or adjusted.

I made sure to have comments in every place where flying is so it's easy to look and see what the code does and where. If it's a problem, it can easily be disabled until fixed. Conflicts are (generally) easy to fix. Even more so if properly commented on.

SilverTalismen:
Personally, this seems fine, I'm certain ways can be made to prevent flight, and I'm certain that Security can manage to catch whoever is abusing flight to escape them and even Admins can intervene if needed

Benemuel:
I am a fan of the idea, if just because it would allow for some interesting RP with people who fly. Its as abusable as Haste is when it comes to sec, especially if combined, but well, its part of the trait system. You ARE going to get min maxers.

Cameron653:
Like I said. If the time to fly up needs to be edited, I can do so. It is currently at 7 seconds, plus whatever other slowdown effects you have on you. This means if security shoots you with a taser once, you take even longer to move back up if you go down.

Natje0:
I love this. It makes me want to revisit my harpy character idea... I see a lot of potential in this, and it's likely to make winged characters a bit more popular.

Cameron653:
Let me fix something real quick. Incap check doesn't work.

code/modules/multiz/movement.dm
Cameron653: Tried this and it didn't work for whatever reason. Went ahead and did incap check & a few of the checks above. Restrained and buckled were left out since, well, they really aren't needed. Can still flap wings while restrained, and you really shouldn't be able to be buckled while in a space where all objects will fall.

Cameron653:
Alright. After vigorous testing, I'm marking this as done. Everything is working as intended and there are no bugs I'm finding with it.

Only adjustments done to this code now will be suggestions.

code/modules/mob/living/carbon/human/species/station/station_special_abilities_vr.dm
```diff
@@ -720,3 +720,21 @@
...
+/mob/living/carbon/human/proc/wings_toggle()
+	set name = "Toggle Flight"
+	set desc = "Toggle to allow for draining to be prolonged. Turn this on to make it so prey will be knocked out/die while being drained. Can be toggled at any time."
+	set category = "Abilities"
+
+	var/mob/living/carbon/human/C = src
+	if(!C.wing_style) //The species var isn't taken into account here, as it's only purpose is to give this proc to a person.
+		to_chat(src, "You cannot fly without wings!!")
+		return
+	if(C.incapacitated() || C.stat || C.lying || C.stunned || C.weakened || C.paralysis)
```
Razer69: Should just be `C.incapacitated(INCAPACITATION_ALL)`.

code/modules/multiz/movement.dm
```diff
@@ -49,6 +49,25 @@
 			else
 				to_chat(src, "<span class='warning'>You gave up on pulling yourself up.</span>")
 				return 0
+		else if(ishuman(src)) //VOREStation Edit Start. Are they human (H above.), have a wing style, and do they have the trait?
+			var/mob/living/carbon/human/H = src //VOREStation Edit. Used for below
+			if(H.wings_flying == 1)
+				if(H.incapacitated() || H.stat || H.lying || H.stunned || H.weakened || H.paralysis)
```
Razer69: `C.incapacitated(INCAPACITATION_ALL)`

code/modules/multiz/movement.dm
```diff
@@ -140,14 +159,24 @@
 
 	if(throwing)
 		return
+	if(ishuman(src)) //VOREStation Edit Start. Code to see if they have wings.
+		var/mob/living/carbon/human/H = src
+		if(H.wings_flying == 1) //Some other checks are done in the wings_toggle proc
+			if(H.incapacitated() || H.stat || H.lying || H.stunned || H.weakened || H.paralysis) //Incap by itself doesn't work, for some reason.
```
Razer69: `C.incapacitated(INCAPACITATION_ALL)`

Cameron653:
Hmm, good idea. I'll make it so if someone is flying they can't wrap up a micro and all that. Going to test out the INCAPACITATION_ALL, too.

code/modules/multiz/movement.dm
Razer69: In current code you *are* checking for restrained and buckled in `incapacitated()`. Also if you allow flight while buckling, you'll allow flight in wheelchairs, office chairs and similar, which doesn't make too much sense to me. In its current state, it would also get players stuck when the chair falls a level below and they are unable to unbuckle.

Cameron653:
Alright. Changed to incap all and flying checks done.
Also, crossed doesn't need to be touched as far as I can see. Only deals with muelbots
https://github.com/VOREStation/VOREStation/blob/master/code/modules/mob/living/carbon/human/human.dm#L267-L276

Cameron653:
Tested. Everything works it'd appear. Also, did a test with micro in hand. They don't fall down and break everything while in a pocket/in hand, so that's good.

Razer69:
Crossed() came to my mind for mouse traps and mulebots, yes. It likely isn't used for anything else that would matter.

code/modules/vore/resizing/resize_vr.dm
```diff
@@ -200,7 +200,7 @@ var/const/RESIZE_A_SMALLTINY = (RESIZE_SMALL + RESIZE_TINY) / 2
 				tmob.apply_damage(damage, HALLOSS)
 				tmob.resting = 1
 				var/mob/living/carbon/human/H = src
-				log_and_message_admins("has stomped on, [tmob] pinning them to the ground and dealing [damage] HALLOSS.") //Both humans and mobs, since stepping on mobs can be abused.
+				log_and_message_admins("has stomped on [tmob] pinning them to the ground and dealing [damage] HALLOSS.") //Both humans and mobs, since stepping on mobs can be abused.
```
Razer69: Corrected that in my PR as well, though the comma should be after `[tmob]`. It was merely misplaced.

Screemonster:
Away missions are single-Z as far as I'm aware anyway, I don't think you can use multi-z with them as the multi-z zone is used for the main station.

SpadesNeil:
I don't like the idea of a flying mechanic even on our own station.

Is there at least a limit to how much/how often you can fly?

Character nutrition should play a role. Flying should not be possible if you're hungry, and should also not be possible if you're too full.

SilverTalismen:
I, personally don't have much of a issue with flight, while it could easily be abused it is also easily countered, there's only a handful of areas you can fly about in, one is in the toxic atmosphere with very few ways in and outwithout breaking windows, which becomes atmos grief at that point, and the other is well within range of Security's tasers so they can easily capture you or wait for you to exhaust yourself.

Plus if it is abused then admins can easily counter it and punish the abuser or warn them

SpadesNeil:
>Plus if it is abused then admins can easily counter it and punish the abuser or warn them

Or we could just not put in features that are likely to be abused. Otherwise we add something and get angry when people use it. This is bad practice.

Yoshax:
As far as I'm aware, Multi-Z can be used for away missions. I remember Aro and I discussing it before. Thought I'm not certain.

I'm with Ace on this though, I just don't like the idea of it. I don't have the best description as to why, but it just leaves me with a bad feeling.

SpadesNeil:
Also, the interior of the station is large, but not so large that powered flight should be feasible.

Adding a trait to allow reduced fall damage, or no fall damage, due to having wings, all that is fine by me. It's such a niche kind of accident that 'gliding' would be okay in my book.

Cameron653:
I can add a nutrition check and make it drain your nutrition over time, yeah. That'd actually be quite easy to do. Can make it piggyback off the fall() proc, along with putting checks to see if you're too full or not in the activation proc.

Cameron653:
Done. Added a limit to flying. Each tick you're up in the air, you lose 10 nutrition. Upon getting hungry, the game warns you that you're getting hungry and you need to get down to the ground. If you run out of nutrition, you fall.
Additionally, if you're too full, you can't fly or, if you get fat in the air, somehow, you fall.

Min nutrition (will fall if you hit this point): 20 or less
Max nutrition (Will fall if you hit this point): 1000 or more.

Cameron653:
Also, @Yoshax as I said before, multi Z gateway missions aren't an issue. Just add a var to areas called no_fly. Have the zMove proc check to see if the area has this var (Much like it does with gravity) and if it returns 1, tell it not to let them move up.

Example:
```c
				if(area.no_fly)
					to_chat(src, "<span class='notice'>You can't fly upwards in this location!</span>")
					return 0
```
Example of it included in the code: https://pastebin.com/giXFZ5aR

Very-Soft:
I think the point cost should be higher than 1 for something that provides such a significant movement advantage. Flying would make some people completely inaccessible to others. Just the ability to fly up to the third level of the atrium alone gives you a normally inaccessable hiding spot, where no one can see you unless they just to happen to build out there. Not to mention literally the entirety of the exterior of the station.

1 point for such an advantage seems a little bogus.

Cameron653:
Hmm. I can bump it up a point to make it more balanced.

Screemonster:
Personally I'd like it if it was gliding rather than full flight/hovering, as in "if flying, you constantly move forwards in the direction you're facing whether you want to or not". I dunno how feasable that is. It means someone wouldn't be able to fly up in the middle of the atrium and AFK until their hunger ticks down to evade security as they'd have to be constantly flying in circles to avoid flying into the windows.

Mainly because the idea of birds splattering against the windows then tumbling to the floor with a crash is fucking hilarious to me.

Cameron653:
I'll have to look into how jetpack stabilization works and see if it's possible to make it glide people above open spaces.

Cameron653:
Done. You now 'glide' while wings are enabled, much like if you were to fly in space without jetpack stabilization on. This means that in order to move upwards, you'd have to place down a lattice, stop flying, steady yourself next to the  lattice, start flying and move upwards again, making abuse of this to evade security practically impossible.

Very-Soft:
It would be good if you had to keep moving to keep flying, and if you stop, you'll fall down (for no fall damage) but if you run into something while flying, you'll fall. 

Tl;dr I'm 100% with scree, birds smacking into windows and falling down would be hilarious.

Runa-Dacino:
Why no hovering? The whole shtick of flying for me is for a micro to hover at head level and pester people that way. That also solves issue of size - for a micro, the atrium is -massive-.

Anewbe:
Not that my opinion really matters here, but I think it might be a better design choice to make a more general "Flying" var for mobs, and just have the wings allow you to change that var on yourself.

Because, while Polaris doesn't have wings, this does very much make me want rocketpacks, and I'd port this in a heartbeat if it wasn't quite this specific.

Cameron653:
Hmm. Good idea. Let me see about wings_flying to flying and put it on /living instead of /human

Cameron653:
@Anewbe Done. I made the flying procs /mob, included a flying check on mob/Check_Shoegrip and made the var "flying" instead of "wings_flying"
This should now be a lot easier to port. Additionally, the 	
```if(H.grabbed_by.len) ```
on line 183 in movement.dm in this PR could allow for some neat "holding someone over the edge while they do repair work" and the calling of fall() being called on line 70 in life.dm will prevent floating in the middle of open spaces when people aren't supposed to be there.

**Comment by Cameron653** [comment]
Fixing some runtimes. DNM for the moment.

**Comment by Cameron653** [comment]
Runtimes resolved. Well, other than the one below, which isn't related to this PR
```C
20:00:00] Runtime in seed_packets.dm,79: Cannot execute null.create random seed().
  proc name: New (/obj/item/seeds/random/New)
  src: the packet of seeds (/obj/item/seeds/random)
  src.loc: the abandoned crate (/obj/structure/closet/crate/secure/loot)
  call stack:
  the packet of seeds (/obj/item/seeds/random): New(the abandoned crate (/obj/structure/closet/crate/secure/loot))
  the abandoned crate (/obj/structure/closet/crate/secure/loot): generate loot()
  the abandoned crate (/obj/structure/closet/crate/secure/loot): New(the floor (102,73,11) (/turf/simulated/shuttle/floor/black))```

File: code/modules/mob/animations.dm [hunk 7]
```diff
@@ -86,7 +86,11 @@ note dizziness decrements automatically in the mob's Life() proc.
 	if(anchored||buckled)
 		make_floating(0)
 		return
-
+	if(ishuman(src)) //VOREStation Edit Start. Floating code.
+		var/mob/living/carbon/human/H = src
+		if(H.flying == 1)
```
ShadowLarkens: `== 1` in this case is redundant.

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -1207,7 +1207,11 @@
 			verbs |= /mob/living/carbon/human/proc/succubus_drain_finialize
 			verbs |= /mob/living/carbon/human/proc/succubus_drain_lethal
 		if(species.hard_vore_enabled == 1) //Hardvore verb.
-			verbs |= /mob/living/carbon/human/proc/shred_limb //VOREStation Edit End
+			verbs |= /mob/living/carbon/human/proc/shred_limb
+		if(species.can_fly == 1)
```
ShadowLarkens: Same here, and infact for all of the above conditionals- with binary flipping variables, `== 1` is unnecessary, redundant code.

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -1447,6 +1451,8 @@
 /mob/living/carbon/human/Check_Shoegrip()
 	if(shoes && (shoes.item_flags & NOSLIP) && istype(shoes, /obj/item/clothing/shoes/magboots))  //magboots + dense_object = no floating
 		return 1
+	if(flying == 1) //VOREStation Edit. Checks to see if they have wings and are flying.
```
ShadowLarkens: More redundant `== 1`

code/modules/mob/living/carbon/human/human_movement.dm
```diff
@@ -141,6 +141,8 @@
 		if(((!check_drift) || (check_drift && thrust.stabilization_on)) && (!lying) && (thrust.allow_thrust(0.01, src)))
 			inertia_dir = 0
 			return 1
+	if(flying) //VOREStation Edit. If you're flying, you glide around!
```
ShadowLarkens: This should probably either stay like it is and have all of the other code changed to match it, or changed to match `== 1`.

code/game/turfs/turf.dm
```diff
@@ -145,6 +145,9 @@ var/const/enterloopsanity = 100
 			M.lastarea = get_area(M.loc)
 		if(M.lastarea.has_gravity == 0)
 			inertial_drift(M)
+		if(M.flying == 1) //VORESTATION Edit Start. This overwrites the above is_space without touching it all that much.
```
ShadowLarkens: `== 1` in this case is redundant.

code/modules/mob/living/carbon/human/species/station/station_special_abilities_vr.dm
```diff
@@ -720,3 +720,56 @@
...
+	if(!C.wing_style) //The species var isn't taken into account here, as it's only purpose is to give this proc to a person.
+		to_chat(src, "You don't have wings!")
+		return
+	if(C.incapacitated(INCAPACITATION_ALL))
+		to_chat(src, "You cannot hover in your current state!")
+		return
+	if(C.nutrition < 50 && !C.flying) //Don't have any food in you?" You can't hover, since it takes up 25 nutrition. And it's not 25 since we don't want them to immediately fall.
+		to_chat(C, "<span class='notice'>You lack the nutrition to fly.</span>")
+		return
+	if(C.anchored == 1)
```
ShadowLarkens: Is this `== 1` necessary? Isn't anchored another true/false variable?

code/modules/mob/living/carbon/human/species/station/station_special_abilities_vr.dm
```diff
@@ -720,3 +720,56 @@
...
+		to_chat(C, "<span class='notice'>You are already hovering and/or anchored in place!</span>")
+		return
+
+	if(C.anchored != 1 && !(C.pulledby)) //Not currently anchored, and not pulled by anyone.
```
ShadowLarkens: Would this not be more clear as `if(!C.anchored && !C.pulledby)`?

code/modules/mob/living/living.dm
```diff
@@ -146,7 +146,7 @@ default behaviour is:
 			..()
 			if (!istype(AM, /atom/movable) || AM.anchored)
 				//VOREStation Edit - object-specific proc for running into things
-				if((confused || is_blind()) && stat == CONSCIOUS && prob(50) && m_intent=="run")
+				if(((confused || is_blind()) && stat == CONSCIOUS && prob(50) && m_intent=="run") || flying)
```
ShadowLarkens: If I'm reading this correctly, walking into any anchored object while flying will 100% result in stumbling? Is that correct?

code/modules/mob/mob_movement.dm
```diff
@@ -517,6 +517,8 @@
 	return dense_object
 
 /mob/proc/Check_Shoegrip()
+	if(flying == 1) //VOREStation Edit. Checks to see if they  and are flying.
```
ShadowLarkens: `== 1` in this case is redundant.

code/modules/multiz/movement.dm
```diff
@@ -49,6 +49,25 @@
 			else
 				to_chat(src, "<span class='warning'>You gave up on pulling yourself up.</span>")
 				return 0
+		else if(ismob(src)) //VOREStation Edit Start. Are they a mob, and are they currently flying??
+			var/mob/H = src
+			if(H.flying == 1)
```
ShadowLarkens: `== 1` in this case is redundant.

code/modules/multiz/movement.dm
```diff
@@ -140,14 +162,39 @@
 
 	if(throwing)
 		return
+	if(ismob(src))
+		var/mob/H = src //VOREStation Edit Start. Flight on mobs.
+		if(H.flying == 1) //Some other checks are done in the wings_toggle proc
```
ShadowLarkens: `== 1` in this case is redundant.

code/modules/multiz/movement.dm
```diff
@@ -49,6 +49,25 @@
...
+			var/mob/H = src
+			if(H.flying == 1)
+				if(H.incapacitated(INCAPACITATION_ALL))
+					to_chat(src, "<span class='notice'>You can't fly in your current state.</span>")
+					H.stop_flying() //Should already be done, but just in case.
+					return 0
+				var/fly_time = max(7 SECONDS + (H.movement_delay() * 10), 1) //So it's not too useful for combat. Could make this variable somehow, but that's down the road.
+				to_chat(src, "<span class='notice'>You begin to fly upwards...</span>")
+				destination.audible_message("<span class='notice'>You hear the flapping of wings.</span>")
+				H.audible_message("<span class='notice'>[H] begins to flap \his wings, preparing to move upwards!.</span>")
```
ShadowLarkens: There's a stray period after the `!` here.

code/modules/multiz/movement.dm
```diff
@@ -140,14 +162,39 @@
...
+				return
+			else if(H.nutrition < 100 && H.nutrition > 89)
+				to_chat(H, "<span class='danger'>You're seriously fatigued! You need to get to the ground immediately and eat before you fall!</span>")
+				return
+			else if(H.nutrition < 2) //Should have listened to the warnings!
+				to_chat(H, "<span class='danger'>You lack the strength to keep yourself up in the air...</span>")
+				H.stop_flying()
+			else
+				return
+		if(H.grabbed_by.len) //If you're grabbed (presumably by someone flying) let's not have you fall. This also allows people to grab onto you while you jump over a railing to prevent you from falling!
```
ShadowLarkens: Can people grab each other at the same time? If so, this might allow people to walk across the void by holding onto each other.

code/modules/vore/resizing/resize_vr.dm
```diff
@@ -166,6 +168,14 @@ var/const/RESIZE_A_SMALLTINY = (RESIZE_SMALL + RESIZE_TINY) / 2
  */
 /mob/living/proc/handle_micro_bump_other(var/mob/living/tmob)
 	ASSERT(istype(tmob)) // Baby don't hurt me
+	if(ishuman(src))
+		var/mob/living/carbon/human/P = src
+		if(P.flying == 1) //If they're flying, don't do any special interactions.
```
ShadowLarkens: `== 1` in this case is redundant.

code/modules/vore/resizing/resize_vr.dm
```diff
@@ -141,6 +141,8 @@ var/const/RESIZE_A_SMALLTINY = (RESIZE_SMALL + RESIZE_TINY) / 2
 		now_pushing = 0
 		if(src.get_effective_size() > tmob.get_effective_size())
 			var/mob/living/carbon/human/H = src
+			if(H.flying == 1)
```
ShadowLarkens: `== 1` in this case is redundant.

code/modules/vore/resizing/resize_vr.dm
```diff
@@ -166,6 +168,14 @@ var/const/RESIZE_A_SMALLTINY = (RESIZE_SMALL + RESIZE_TINY) / 2
  */
 /mob/living/proc/handle_micro_bump_other(var/mob/living/tmob)
 	ASSERT(istype(tmob)) // Baby don't hurt me
+	if(ishuman(src))
+		var/mob/living/carbon/human/P = src
+		if(P.flying == 1) //If they're flying, don't do any special interactions.
+			return
+	if(ishuman(tmob))
+		var/mob/living/carbon/human/D = tmob
+		if(D.flying == 1) //if the prey is flying, don't smush them.
```
ShadowLarkens: `== 1` in this case is redundant.

code/modules/mob/living/living.dm
Cameron653: Yes. If they bump into anything at all, they immediately "stumble" (see: crash) into it.

code/modules/mob/mob_movement.dm
Cameron653: Going to fix all of these == 1's

code/modules/multiz/movement.dm
Cameron653: Good catch.

code/modules/multiz/movement.dm
Cameron653: Hmm, that is problematic. Didn't think of that. Going to test some potential fixes for this. Possibly doing a check "if var/obj/item/weapon/grab/G in contents? Will fiddle with it.

Cameron653:
Done. All the == 1's have been purged, and the walking in the air w/ someone else has been fixed by checking the person's hands contents if they're grabbed. Tested and the fix works.

Cameron653:
Alright. I'm going to call this a done deal.

It is relatively balanced, easy to learn hard to master (especially at faster movement speeds,) quite a high cost trait (two points is enough to get 25% resistance!) and can be quite dangerous to the user if they're not careful.

Additionally, if/when flying framework is ported to Polaris, this'll touch a lot less files than it does in its current state.

SilverTalismen:
It looks great on my end, I don't know much about code but from what I've seen as this PR developed and changed it's become pretty great and is decently balanced in my own opinion..

SpadesNeil:
Xenochimera should not have flight by default. They have wings, that doesn't mean they work properly.

SilverTalismen:
`Xenochimera should not have flight by default. They have wings, that doesn't mean they work properly.`
But, it makes sense that Xenochimera could fly if they eat or possess wings?

SpadesNeil:
Xenochimera already have too many cool things. Let's not make it worse. :U

SilverTalismen:
It's not like we have many repetative Xenochimera though, There's me, and there's Scree and occasionally Natje I don't often see any others..

SpadesNeil:
Fiiiiiiiiiiiine.

PR #2389, opened 2017-10-29
Makes mice unable to push things around.

Verkister:
-No more absolute disarray and blocked routes caused by maintenance mice pushing crates and welder tanks all over the place.


SpadesNeil:
Maybe give this to Polaris later.

PR #2390, opened 2017-10-29
Removes a redundant double check from digestion_death and adminlogs digested players

Verkister:
-If we get into this proc the victim is already dead and there's no need for a re-kill.
-Essentially the cause behind mouse carcasses screaming in the gut.
-Thing was calling the death proc for the already dead mob and the death proc is what contains the scream on mice.


Screemonster:
While you're messing around in there, it'd be nice to log gurgledeaths.

Verkister:
Done

PR #2394, opened 2017-10-30
Fixes some issues with subtles

Yoshax:
Should/Might actually make things less expensive in general for emoting. Blame Aronai if it doesn't, he wrote it :))))

Also this actually makes emotes with simple mobs without keys actually work properly.


Yoshax:
Gonna leave this till I have the whisper fix too, actually.

PR #2407, opened 2017-11-03
Adds Serana Te'saln's drink

Neopatra:
I've been making these on-station for almost a year, so I figured I should make it an actual thing in the code at this point. It has a few alterations to make it act like it used to without being too outrageous in how much you needed to drink to proc any effects.


SpadesNeil:
Polaris _clearly_ needs this.

Neopatra:
Why does Polaris need this? I made this on-station here for almost a year now, it's kinda my thing.

SpadesNeil:
-shrug- Figured they'd enjoy it, but okay. It can be Vorestation unique.

PR #2409, opened 2017-11-03
Replaces the ADD type icons for DSI lizard and taj with multiply ones.

Verkister:
-They're just the organic multiply icons tho for now.
-Kept both, and will do proper sets after next polaris sync (species/limbs combo blend mode)


SpadesNeil:
Expect polaris sync soonish.

PR #2412, opened 2017-11-04
Adds scaling to character previews!

Verkister:
-How cool is that?!
-Also centers the icons so they don't push each other out of frame.
-(edited after the pics below) Evened out the spacing between the sprites and lined the bottom to the background tiling.
kuva
kuva
kuva
kuva
kuva
kuva
kuva
Shit gets a bit crowded on macro taurs but whatever :v


Dhaeleena:
It would have been good if a Toggle for this was added

PR #2416, opened 2017-11-04
Fixes vending machines giving out bad protein shakes

Cameron653:


Cameron653:
Wait, drinks/drinkingglass.dm is commented out. Whoops.

PR #2418, opened 2017-11-05
POLARIS SYNC THAT FIXES THE THINGS

Cameron653:


Cameron653:
Fixes species colors from being WHITE ON ALL COLORS.
Really. You could set the color to almost pure black and you'd be white.

PR #2429, opened 2017-11-05
Some gurgle gain tweaks.

Verkister:
-Dead mobs now give nutrition/charge based on their mob_size.
-Broadened the scope of vulnerable food items for item friendly mode.
-Also made severed organs vulnerable to item friendly mode.
-Made severed organs give 50 nutrition per piece to make up for HV health loss.
-Just one single unit of ingested nutriment reagent gives 30 points so come on!

### DIFF
--- a/code/modules/vore/eating/bellymodes_vr.dm
+++ b/code/modules/vore/eating/bellymodes_vr.dm
@@ -54,10 +54,10 @@
 				owner << "<span class='notice'>" + digest_alert_owner + "</span>"
 				M << "<span class='notice'>" + digest_alert_prey + "</span>"
 
-				owner.nutrition += 10*M.mob_size // dead mice 10, dead people 200 etc.
+				owner.nutrition += 5*M.mob_size // Dead mice 5, dead people 100 etc. Miniscule amount, but come on, just one ingested unit of nutriment reagent gives 30 points (Chemistry-Reagents-Food-Drinks.dm)
 				if(isrobot(owner))
 					s_owner = owner
-					s_owner.cell.charge += 50*M.mob_size
+					s_owner.cell.charge += 25*M.mob_size
 				var/deathsound = pick(death_sounds)
 				for(var/mob/hearer in range(1,owner))
 					hearer << deathsound
--- a/code/modules/vore/eating/bellymodes_vr.dm
+++ b/code/modules/vore/eating/bellymodes_vr.dm
@@ -54,7 +54,9 @@
 				owner << "<span class='notice'>" + digest_alert_owner + "</span>"
 				M << "<span class='notice'>" + digest_alert_prey + "</span>"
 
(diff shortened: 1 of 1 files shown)


Cyantime:
Consider using the mob size defines here https://github.com/VOREStation/VOREStation/blob/master/code/__defines/mobs.dm#L114 to determine nutrition gain from dead mobs. Mice are size 1, humans are size 20.

Verkister:
Roger that.

SpadesNeil:
Ehhh no. This needlessly screws up the balance I already had on this. Unless we make digestion _damage_ result in _no nutrition_ gain, this is an awful idea.

Verkister:
It's not much really. One liquidfood gives like 500 or something at least. Would make at least some sense to have a dead body give a little more than _a half of a raw donk pocket_ especially with the new HV mechanics often resulting to the victim dying before getting ate. Currently HV doesn't feed for shit with the organs only giving the trace amount akin to non-nutritional other crap and we hardly have any "edible" mobs around that weren't mice or crew already so what even is there to lose?

Verkister:
If you think anything over 20 on a scale of **0-to-450-to-infinity** is game-breakingly excessive, I recommend snorting up some hard math.

Arokha:
> One liquidfood gives like 400 or something at least.

Then liquifood rations are OP.

SpadesNeil:
_Liquid food rations **are** OP._

SpadesNeil:
I am 90% sure liquid food rations **do not** give 400 nutrition.

SpadesNeil:
Here's what needs to happen instead of this:

Digestion needs to be reworked completely. Instead of a person dumping X-amount of nutrients into a stomach _on top of_ the nutrients you **already** get from digestion, one of these has to be eliminated. If you want dead bodies to give a full body's worth of nutrients, then you need to delete what gives nutrients _during_ digestion. I'd be okay with that.

In this current state, this will not be merged.

Verkister:
Apparently liquidfood rations give 600. 20x30.
kuva
kuva

Verkister:
So yeah. Just one ingested unit of nutriment reagent gives 30 points. 20 points for dead bodies is quite literally less than a single bite off most foods so keep telling me it's 'too much' all you want.

Verkister:
They already do give the 20 for dead bodies regardless of mob size. _This only scales it some so that dead mice and dead crew members don't count as the same thing_. I did give the severed organs a buff to make up for the hard vore stuffs though, and wanted the main body count for at least for little more than _**0.66u nutriment.**_ (that really is what 20 points equals for) People more or less die when you rip their organs out, and that totally nullifies the gains you'd be getting had you ate them on full health.

SpadesNeil:
This still isn't being done in the manner which I want it to be done.

I could have changed the vore system to give substantial nutriment on death only but I am not doing that. I could make it do both but I am not doing that either. The changes this PR makes are undesirable so I am going to close it.

PR #2434, opened 2017-11-06
Fixes the severed legs crashing? (early access edition)

Verkister:
-Hopefully finally yes.
-Got no proper environment to test if it actually fixes the thing, but extensive troubleshooting and pinpointing has eliminated the issue to the rendering of leg icons, which apparently had nothing in it to skip missing ones.
-Thing affects both, living mobs and ghosts alike, and being a cached icon generation also explains why it only happens once and not after the victims have rejoined.
-Pushed to Polaris too, but it's been like 5 minutes since the last sync and so on.


SpadesNeil:
Just push it to Polaris.

Cameron653:
Tell me when Polaris merges this.

Verkister:
https://github.com/PolarisSS13/Polaris/pull/4232 It's been merged.

PR #2440, opened 2017-11-08
Makes HV at least somewhat feasible with organ nutrition.

Verkister:
-Yeah you don't even get a single food bite worth of nutrition if your victim dies.
-This balances that some by giving severed organs an actual nutritional value instead of garbage treatment.
-Also broadens the scope of vulnerable food items for item friendly mode.
-This includes organs. They will no longer be classified as generic gurgle resistant "items"
(yeah yeah same branch with the old commits showing up but this time it doesn't have the dead things stuff, just the organ and food path things)

### DIFF
--- a/code/modules/vore/eating/bellymodes_vr.dm
+++ b/code/modules/vore/eating/bellymodes_vr.dm
@@ -54,10 +54,10 @@
 				owner << "<span class='notice'>" + digest_alert_owner + "</span>"
 				M << "<span class='notice'>" + digest_alert_prey + "</span>"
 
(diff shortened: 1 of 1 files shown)


code/modules/vore/eating/bellymodes_vr.dm
```diff
@@ -197,6 +200,9 @@
 							internal_contents += M
 						internal_contents -= H
 						qdel(H)
+					if(istype(T,/obj/item/organ))
```
SpadesNeil: This is good.

PR #2455, opened 2017-11-10
Loadout edits

Cameron653:
- **Makes tracking implant that spawns in you cost 0.** 
People can teleport to you and get instanommed & it tells security where you are if they look at the monitor. Not really _too_ useful outside of a telenom, and the downside of bigbrother always knowing where you are makes it costing points unreasonable.
- **Makes wallets (excluding orange wallet) free.** 
Wallets are so common that you can easily find them. This saves people a few minutes of searching maint for a wallet so they can have a fancy armband or something. Orange wallet still costs 1, since it  spawns with cash in it. Requested change.

### DIFF
--- a/code/modules/client/preference_setup/loadout/loadout_accessories.dm
+++ b/code/modules/client/preference_setup/loadout/loadout_accessories.dm
@@ -29,14 +29,18 @@
 /datum/gear/accessory/wallet
 	display_name = "wallet, orange"
 	path = /obj/item/weapon/storage/wallet/random
+	cost = 1 //VOREStation Edit
 
 /datum/gear/accessory/wallet_poly
 	display_name = "wallet, polychromic"
 	path = /obj/item/weapon/storage/wallet/poly
+	cost = 0 //VOREStation Edit
+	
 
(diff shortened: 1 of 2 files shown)


LittleBigKid2000:
Why not remove the money from the orange wallet and make it cost 0 points, since the extra money is largely useless anyway? Aren't they the only kind the spawns in maintenance?

Cameron653:
I have no idea. The orange wallet can stay & 1 point in exchange for cash, since cash can be used to buy NIF upgrades and whatnot.

SpadesNeil:
I am not sure a tracking beacon in people is a good idea, mostly because the point of tracking beacons is that there is supposed to be a limited supply. I do not want that added.

Cameron653:
Well, at the moment, everyone can spawn with a tracking beacon.  It just costs 2/3 of your cosmetic points. Have any suggestions as to what the price should be for it?

SpadesNeil:
eh fuck it

Cameron653:
As it turns out you can't teleport to the implants. The locator is in teleportation.dm, which tells you _where_ the tracking implant is.

PR #2465, opened 2017-11-12
Un-dupes head marking cache generation.

Verkister:
-Polaris bug yeah, pushed there too. Ours got some our edits within the snippets anyway.
-As the title suggests, this is likely also related to the organ icon cache related crashes.


SpadesNeil:
This might actually fix some crashing. It would explain the fuckery with why Polaris doesn't suffer these bugs.

However I'd like @Cameron653 or @Arokha to doublecheck the code.

SpadesNeil:
Conflicts. Probably because this was fixed elsewhere.

PR #2467, opened 2017-11-12
Vulp ear compatibility update.

Verkister:
-Filled up the holes in our vulp hairstyles.
-Separated vulp ears from the base head icon. Yes you can now be a vulp with custom ears replacing the default ones instead of just pasting on top of them. We now have proper vulp ears in the ear selection.
-Basically made the "ear holes" obsolete by recently introducing separate vulp ear set to put over the hair.
-Now the "ear holes" won't be a problem for any non default vulp ear folks.
-Also adjusted the highlight brigtness for certain vulp hairstyles that had the shine brightness way over necessary.


Christykitsune:
Supported for this yes, Vulp is great

PR #2470, opened 2017-11-13
Dogborg wreck and rest icons now work properly!

Verkister:
-Yeah I got em working real nice.
-Dogborg wreck icons are working now.
-Dogborg resting icons are working now too.
-Made borg rest verb not cause a complete blackout.
-Tweaked some rest icons and added missing ones.
-Also supports the taller borgo file.
-Also fixed the tongues not cleaning floors properly. (As in single post-wipe step bringing all the dirt back onto the tile at once.)
-Fixed a couple less relevant runtimes I discovered while testing this stuff.
kuva
kuva
kuva


SpadesNeil:
Do the bellies work correctly?

Verkister:
Yeah, although it does get hidden for the resting icons. The sleepers blast empty on death anyway so it's not an issue for the wrecks.

PR #2471, opened 2017-11-13
Prevents vatborns from spawning

Cameron653:
Our traits system & full customization makes them useless.


Cyantime:
They also have no appendix, which you can't do with the traits system.

SilverTalismen:
Yeah I don't really see any reason to remove them..? I dunno.

PR #2475, opened 2017-11-14
Rebalanced the Energy Luger

SpadesNeil:
@Yoshax noticed that the Energy Luger can do a retarded amount of damage over its Energy Gun counterpart. The Energy Luger has 140% more shots on all settings than the standard Energy Gun. The Energy Luger has the potential to do 2.4x more damage. This is absurd, and not its intended design. This was an oversight that occured during a recent rework of Polaris energy guns and how batteries work.

I have three possible solutions. This is one of them.

- A) Rebalance it to match the energy gun, including the ability to reload it (requires a new sprite which I do not have).
- B) Lock the battery, making it impossible to reload without a charge station, but allow it to keep twice the capacity. (I'll fix the numbers so it's exactly 2x the Egun capacity, not 2.4x)
- C) Lock the battery, but remove the firing delay. **This is what I did in this PR.**

Discuss the three solutions and decide what is best. Option A is not an option without a sprite, though, so if anyone wants option A, make a sprite.

### DIFF
--- a/code/modules/vore/fluffstuff/custom_guns_vr.dm
+++ b/code/modules/vore/fluffstuff/custom_guns_vr.dm
@@ -419,7 +419,7 @@
 // ------------ Energy Luger ------------
 /obj/item/weapon/gun/energy/gun/eluger
 	name = "energy Luger"
-	desc = "The finest sidearm produced by RauMauser; this pistol can punch a hole through inch thick steel plating. This ain't your great-grand-daddy's Luger! Can switch between stun and kill."
+	desc = "The finest sidearm produced by RauMauser. Although its battery cannot be removed, its ergonomic design makes it easy to shoot, allowing for rapid follow-up shots. It also has the ability to toggle between stun and kill."
 	icon = 'icons/obj/gun_vr.dmi'
 	icon_state = "elugerstun100"
 	item_state = "gun"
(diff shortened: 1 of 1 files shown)


code/modules/vore/fluffstuff/custom_guns_vr.dm
```diff
@@ -423,13 +423,14 @@
 	icon = 'icons/obj/gun_vr.dmi'
 	icon_state = "elugerstun100"
 	item_state = "gun"
-	charge_cost = 100 //How much energy is needed to fire.
-	projectile_type = /obj/item/projectile/beam/stun
```
SpadesNeil: `projectile_type = /obj/item/projectile/beam/stun` was redundant code.
`charge_cost = 100` was also redundant code.

SilverTalismen:
Version A: Would be fine with me, as someone who is also applying for it to be a item for one of my character's PDW.

Yoshax:
Should just be a pure reskin, IMO, so A.

The battery-locking, as far as my knowledge extends, is only done the weapons which auto-recharge.

SpadesNeil:
>The battery-locking, as far as my knowledge extends, is only done the weapons which auto-recharge.

Not true. All self-recharging batteries are locked, but not all locked batteries are self-recharging. I checked in the code for this. See energy.dm for details.

We'll do option A if I can get a new sprite though.

SpadesNeil:
I'm just gonna do this for now since version A isn't an option with my current sprites. Else the sprite disappears.

PR #2477, opened 2017-11-14
Adds an inconveniently evil new negative trait.

Verkister:
-"Lightweight". Cost -4. Makes every non-helpful bump received have a 50% chance to knock you down.

### DIFF
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -124,7 +124,7 @@ default behaviour is:
 			if(ishuman(tmob))
 				var/mob/living/carbon/human/H = tmob
 				if(H.species.lightweight == 1 && prob(50))
-					H.visible_message("<span class='warning'>[src] has bumped [H] off their balance, knocking them into the turf!</span>")
+					H.visible_message("<span class='warning'>[src] bumps into [H], knocking them off balance!</span>")
 					H.Weaken(20)
 					now_pushing = 0
 					return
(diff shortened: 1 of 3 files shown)


SpadesNeil:
Table is going to love this lol

SilverTalismen:
All the better to abuse Briana with

code/modules/mob/living/living.dm
```diff
@@ -120,6 +120,13 @@ default behaviour is:
 			// VOREStation Edit - Begin
 			// Handle grabbing, stomping, and such of micros!
 			if(handle_micro_bump_other(tmob)) return
+			// Plow that nerd.
+			if(ishuman(tmob))
+				var/mob/living/carbon/human/H = tmob
+				if(H.species.lightweight == 1 && prob(50))
+					H.Weaken(20)
```
Cameron653: T-that's a long knockdown time! But seeing as the cost is -4, I can support this

Also, if it doesn't already, may want to make it include a custom knockdown message. Otherwise, it'll just look like someone just rested for no reason.

code/modules/mob/living/living.dm
```diff
@@ -124,6 +124,7 @@ default behaviour is:
 			if(ishuman(tmob))
 				var/mob/living/carbon/human/H = tmob
 				if(H.species.lightweight == 1 && prob(50))
+					H.visible_message("<span class='warning'>[src] has bumped [H] off their balance, knocking them into the turf!</span>")
```
SpadesNeil: This is a little awkwardly worded.

`[src] knocks over [H]!` is much better.

SpadesNeil:
Close enough.

PR #2478, opened 2017-11-14
Polaris Sync

LNSmith458:
Fixes a spelling mistake on the GUI for the Central Atmospherics Controller.

Yay first PR. Tell me if I made mistakes.


Cyantime:
This is a Polaris typo, and should be fixed there.
https://github.com/PolarisSS13/Polaris/

SpadesNeil:
Good catch. But yes, we should send this to Polaris.

LNSmith458:
Will do. Thank you.

Cameron653:
This isn't a polaris sync. This'll just cause conflicts if we commit this here. We'll get this with a real polaris sync if they have it upstream.

PR #2479, opened 2017-11-15
Tajaran base contrast tweaks.

Verkister:
-Now with 75% more face.
-Forgot the polaris pr with this edit was reverted due to complications and never made it down here.


SpadesNeil:
Before/After please?

Verkister:
kuva

PR #2481, opened 2017-11-15
Adds sound/verb edit for ejection and the framework for different vore audio sets.

Verkister:
-With this stuff you can choose what sort of noises your guts make, were you a fleshbag, slimebag, scrapheap, or just otherwise acoustically deviant entity doing something beyond imagination to the poor sod inside you.
-This also allows you to set those to make no sound at all.
-In fact that is the only non-default option at this point. The other sets need their sounds made before they can be of use!
-Also allows picking sound and verb for ejection like it already is for insertion.
-Adds test preview button for gurgles and gurgledeaths that plays a random sound from the selected selection.
-Adds a couple extra insertion/ejection sounds. (existing ingame assets for now, some machine sounds and coughs)

### DIFF
--- a/code/_helpers/global_lists_vr.dm
+++ b/code/_helpers/global_lists_vr.dm
@@ -136,6 +136,7 @@ var/global/list/vore_sounds = list(
 		"None" = null)
 
 var/global/list/release_sounds = list(
+		"Splat" = 'sound/effects/splat.ogg',
 		"F cough" = 'sound/effects/mob_effects/f_cougha.ogg',
 		"M cough" = 'sound/effects/mob_effects/m_cougha.ogg',
 		"Synthcough" = 'sound/effects/mob_effects/m_machine_coughc.ogg',
--- a/code/_helpers/global_lists_vr.dm
+++ b/code/_helpers/global_lists_vr.dm
@@ -43,6 +43,13 @@ var/global/list/important_items = list(
 		/obj/item/clothing/suit/storage/hooded/wintercoat/roiz)
 
 var/global/list/digestion_sounds = list(
+		"organic" = digest_organic,
+		"none" = digest_none/*, // NEED NOISES FOR THE REST,
+		"mech" = digest_mech,
+		"goop" = digest_goop,
+		"cyber" = digest_cyber*/)
+
+var/global/list/digest_organic = list(
 		'sound/vore/digest1.ogg',
 		'sound/vore/digest2.ogg',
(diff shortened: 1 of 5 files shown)


Cyantime:
If you try to change the release sound, you can't re-pick the original sound because splat.ogg isn't in the list right now.

Verkister:
shit thanks

Arokha:
Seems okay if it tests okay.

Cameron653:
Is this good to merge?

Verkister:
On framework level yeah I suppose. No clear eta on the actual sound files yet.

code/_helpers/global_lists_vr.dm
```diff
@@ -43,6 +43,13 @@ var/global/list/important_items = list(
 		/obj/item/clothing/suit/storage/hooded/wintercoat/roiz)
 
 var/global/list/digestion_sounds = list(
```
Leshana: Could use a comment explaining this is a list of lists (of sounds)   since it used to just be a list of sounds

code/modules/vore/eating/belly_vr.dm
```diff
@@ -12,7 +12,11 @@
 	var/name								// Name of this location
 	var/inside_flavor						// Flavor text description of inside sight/sound/smells/feels.
 	var/vore_sound = 'sound/vore/gulp.ogg'	// Sound when ingesting someone
+	var/digestion_sounds_pref				// Digestion sound set.
```
Leshana: Datatype of `digestion_sounds_pref` is a list().  since it was named _pref I at first thought it was going to hold a key like "organic" but its a list()  `digestion_sounds_list` maybe would be clearer.  But not a big deal, could also just clarify in the comment.

code/_helpers/global_lists_vr.dm
```diff
@@ -56,7 +63,30 @@ var/global/list/digestion_sounds = list(
 		'sound/vore/digest11.ogg',
 		'sound/vore/digest12.ogg')
 
+var/global/list/digest_none = list(null)
```
Leshana: Won't this result in  `sound(null)` being sent to the client?  Have you tested what happens when you do this?

Arokha: (Hint: Sending 'null' as a sound stops all sounds playing on the client)

Verkister:
gonna do some fixing when i have time

Arokha:
Just open a new one then with these changes when that happens.

PR #2485, opened 2017-11-16
Increases main engine SMES default output

Cyantime:
This prevents the tram and cafe area from blacking out within minutes of roundstart, and gives engineers more time to arrive and start setting up power before people start yelling about being trapped in the tram spawnpoint.


Arokha:
Fair enough

PR #2488, opened 2017-11-16
Fixes grub larva meat type

Cyantime:
Fixes #2487


Cyantime:
Whoops, that's what I get for not looking before doing things

PR #2489, opened 2017-11-16
Attempt to clean up player_list if an issue is found

Arokha:
Dangerous? Not any more than just leaving it, probably. I can't think of a case where player_list -= something wouldn't work, even in the case of nulls probably. And the no-depth [mob] lookup at the end is probably fine, even if it's a null, since it'll use byond's internal "what is this" handler to print the name or an icon or whatever it is.

### DIFF
--- a/code/_helpers/game.dm
+++ b/code/_helpers/game.dm
@@ -281,9 +281,12 @@
 
 	//A list of every mob with a client
 	for(var/mob in player_list)
+		//VOREStation Edit - Trying to fix some vorestation bug.
 		if(!istype(mob, /mob))
-			crash_with("There is a null or non-mob reference inside player_list.")
+			player_list -= mob
+			crash_with("There is a null or non-mob reference inside player_list ([mob]).")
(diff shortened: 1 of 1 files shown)


Cyantime:
Shouldn't this go to Polaris?

SpadesNeil:
I know not what this does.

SpadesNeil:
@Arokha Shouldn't this have `//Vorestation Edit` or go to Polaris?

SpadesNeil:
I still do not know what this does.

Arokha:
Yes, to both of you. I think I was tired when I made this.

Arokha:
Added vorestation edit comments.
@SpadesNeil This is because, randomly in the list of players, there's something that's not a player. I'm not sure if it's just not a mob, or if it's a null or what, but this should fix that, and log it each shift it happens, and we can maybe figure out what's mistreating this list (either leaving nulls, or putting nonmobs into it).

PR #2495, opened 2017-11-18
Adds Slime Feeding

Cameron653:
- Allows slimes to feed people
- Think succubus drain but in reverse
- If the slime runs out of nutrition & drain finalization is on, they disappear into whatever gut the person they're feeding has selected.

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/station_special_abilities_vr.dm
+++ b/code/modules/mob/living/carbon/human/species/station/station_special_abilities_vr.dm
@@ -606,15 +606,80 @@
 			C.absorbing_prey = 0
 			return
 
+/mob/living/carbon/human/proc/slime_feed()
+	set name = "Feed prey with self"
+	set desc = "Slowly feed prey with your body, draining you in the process. You may only do this to one person at a time."
+	set category = "Abilities"
+	if(!ishuman(src))
+		return //If you're not a human you don't have permission to do this.
+	var/mob/living/carbon/human/C = src
+	var/obj/item/weapon/grab/G = src.get_active_hand()
+	if(!istype(G))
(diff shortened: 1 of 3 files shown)


code/modules/mob/living/carbon/human/species/station/station_special_abilities_vr.dm
```diff
@@ -606,15 +606,80 @@
...
+	for(var/stage = 1, stage<=100, stage++) //100 stages.
+		switch(stage)
+			if(1)
+				to_chat(C, "<span class='notice'>You begin to feed [T]...</span>")
+				to_chat(T, "<span class='notice'>An odd sensation flows through your body as [C] begins to feed you!</span>")
+				T.nutrition = (T.nutrition + (C.nutrition*0.05)) //Drain a small bit at first. 5% of the prey's nutrition.
+				C.nutrition = C.nutrition*0.95
+			if(2)
+				to_chat(C, "<span class='notice'>You feel weaker with every passing moment of feeding [T].</span>")
+				src.visible_message("<span class='notice'>[C] seems to be doing something to [T], [T]'s body looking stronger with every passing moment!</span>")
```
Cyantime: `[T], [T]` could probably be rephrased in a way that isn't repeating the target's name twice like that

Screemonster:
... don't think I'm going to let that branch name slip by without a comment.

Cameron653:
Changed [T], [T]'s to [T], resulting in [T]'s

SpadesNeil:
>Changed [T], [T]'s to [T], resulting in [T]'s

https://www.youtube.com/watch?v=KufiqCh3Qd4

Cameron653:
I edited the way it looks.
Instead of: "Person seems to be doing something to Zee, Zee's body looking...
It does: "Person seems to be doing something to Zee, resulting in Zee's body looking...

PR #2450, opened 2017-11-10
Adds toggleable character pref for sharp/fuzzy scaling. (Tested.)

Verkister:
**Tested working.**
-Yeah you should now be able to choose how your character looks without forcing a global mode over everyone.
-Especially benefitical for any character scaled to those atrociously fucked up deformation ranges.
-Kinda based on #1947 Go there for preview examples.
-Because some characters just look better soft and symmetrical instead of something that got run over with one of those landfill variety steamrollers.

PR #2496, opened 2017-11-18
Fixes #2450 that the sync conflicts broke.

Verkister:
-Oh neato beans the conflict update actually made fixing my toggle update easier than it was to get working with the old placement of bixel scale.

-Fixes #2450 that the sync conflicts loss.jpg'd on the last second reee.

kuva


Cameron653:
Looks good to me.

PR #2499, opened 2017-11-18
Fixes Chain of Command guide book

SpadesNeil:


SpadesNeil:
Actually, gonna keep it canon that Ace wrote this.

PR #2500, opened 2017-11-18
Gives pAIs Brutal Predation

Cameron653:
- Allows PAIs to chew off limbs
- The proc takes 45 seconds to occur, and both have to be standing still. This is to make up for the fact they don't need a neckgrab for it.
- pAIs are vastly underused, so this will be a rare occurrence & allow those which like to do HV to be able to perform HV as a pAI


Mewchild:
I'm not salty that a siri PDA gets this and my prommie doesn't, not at all :c

Cameron653:
I need to give Prometheans their own custom hard-digestion-esk slime melt sometime.

PR #2501, opened 2017-11-18
Firing Range Updates

SpadesNeil:
Main:
- Gives green glasses to Security firing range. For safety.
- Gives RnD a firing range.
image

Unrelated:
- Removes meat hook in the hangar, because it's not an engine hoist.


ArgobargSoup:
I would really like to request that, should this go through, that some extra maintenance tunnels are added so that the large chunk of maintenance to the left, located in a very large area ripe for digging into, isn't suddenly cut off and accessible only to Science.

SpadesNeil:
>I would really like to request that, should this go through, that some extra maintenance tunnels are added so that the large chunk of maintenance to the left, located in a very large area ripe for digging into, isn't suddenly cut off and accessible only to Science.

I am attempting to do just that.

SpadesNeil:
And done.

PR #2505, opened 2017-11-20
Finished borgopup port job and more!

Verkister:
-Old original features brought back:
--Defib paws for medihound.
--Pounce for k9. (some halloss plus disarm tier knockdown rng, 3 tile range, hefty cell charge cost(10% per hop on default cell), requires navigating popup menus for each attempt aka absolutely no point-and-click spam either)
--Laser/taser wielding overlays.
--Pounce as emag addition for all dogborgs.
-Some code cleanups.
-Removed the old botched feeding attempt that was also overriding my 0 force polaris fix.
-Finally fixed my original blooper of accidentally making the sprite files 62p wide instead of 64.
-Also fixed the taller fluff file.
-Enabled the sparky wreck overlay we apparently had.
-Separated janipup eye lights to their own overlay.
-Removed the obsolete hud glasses from module inventory.
-Made pounce have guaranteed knockdown against lightweight trait just because.


Dhaeleena:
--Pounce as emag addition for all dogborgs.

What does this mean?

Yoshax:
That leap doesn't sound too balance, to be honest. There has been a long move towards not having instant knockdowns or stuns, or anything of the sort, and introducing them is a poor idea.

code/modules/mob/living/silicon/robot/dogborg/dog_modules_vr.dm
```diff
@@ -332,11 +288,80 @@
...
+	..()
+	flags |= NOBLUDGEON
+
+/obj/item/weapon/dogborg/pounce/attack_self(mob/user)
+	var/mob/living/silicon/robot.R = user
+	R.leap()
+
+/mob/living/silicon/robot/proc/leap()
+	if(last_special > world.time)
+		src << "Your leap actuators are still recharging."
```
Yoshax: This should be to_chat() as should the others

code/modules/mob/living/silicon/robot/dogborg/dog_modules_vr.dm
```diff
@@ -332,11 +288,80 @@
...
+	var/mob/living/T = input(src,"Who do you wish to leap at?") as null|anything in choices
+
+	if(!T || !src || src.stat) return
+
+	if(get_dist(get_turf(T), get_turf(src)) > 4) return
+
+	if(last_special > world.time)
+		return
+
+	if(stat || paralysis || stunned || weakened || lying || restrained() || buckled)
```
Yoshax: This should be if(usr.incapacitated(INCAPACITATION_DISABLED)) or so.

Verkister:
I pretty much just pasted those things from the existing xeno code. What comes to the balance issues, it has jack shit of any robusting use as the leap targeting requires picking the target from a popup window and is impossible to robust on the go.

Yoshax:
Clearly you didn't see Vox with it. A few seconds is more than enough time to put handcuffs on someone.

Verkister:
Man wouldn't it be horrible if a criminal got cuffed for once

SpadesNeil:
We're not adding instapounce.

>Man wouldn't it be horrible if a criminal got cuffed for once.

implying that security is always incompetent all the time

>just to briefly inconvenience the fleeing powergamer

implying the pounce ability won't be abused by powergamers, only used against them

Verkister:
1, It is a massive metal dog slamming onto you, which somehow should be the same as getting bapped to the knee with a cardboard tube?
2, It's a 3 second inconvenience for the victim with a 8 second cooldown and fat charge cost for the borg and only has the range of 3 tiles behind the hassle of navigating your target from the popup menus.
3, Let's not let powergaming become the expected standard of gameplay, please.

SpadesNeil:
It _is_ the expected gameplay, though. If a feature exists like this, and it provides a distinct advantage while working as intended, it will result in honest players over-using it.

And I'm not going to result to banning people when an overpowered feature is added to begin with. There is a very good reason that instant stuns have been removed.

This will not be merged with instant stuns.

Verkister:
It's purely a fluff feature now.

code/modules/mob/living/silicon/robot/dogborg/dog_modules_vr.dm
```diff
@@ -364,4 +364,4 @@
 		src << "<span class='warning'>You miss!</span>"
 		return
 
-	T.Weaken(3)
+	T << "<span class='warning'>Please lay down for the arrest to commence immersively.</span>"
```
SpadesNeil: This is just silly now.

I didn't say we had to delete the pounce feature, just that it cannot be accepted when it's as overpowered as it is. Consider what mobs are _normally_ able to do, such as using disarm to knock someone down.

Verkister: Alright. What I can tell is that disarm has a 25% chance to push, which will apply the same 3 second weaken if the target has <60 armor. I'm just having difficulties trying to figure out how to make the chances feasible for it to not require too many costly attempts while still not being an insta stun. Could possibly take down the cooldown and just rely on the charge costs so that an rng-unlucky borg will hop its batteries dry.

Verkister:
Made it deal some halloss and gave it similar push chance as disarm has. Also took the cooldown down to 2 seconds and charge up to 750.

Cameron653:
:+1:

Verkister:
Oi did ya even check the changes before skipping this shit from the rush release?

SpadesNeil:
Considering that this change is more than just a few lines, meaning I have to review it more thoroughly, and that the release was a meager quick hotfix for bugs and mapping issues, the answer is no. No I did not.

You're beginning to get on my nerves with your attitude as of late. Knock it off.

SpadesNeil:
That said, this is _much_ more fairly balanced with a pounce _chance_ instead of just immediate knock-down. I'm not sure I'm too keen on it being ranged, which is the only lingering doubt of mine.

But, I don't think it's a huge issue. I want to run this by people and see how they think of it.

Yoshax:
> Oi did ya even check the changes before wordlessly skipping this shit from the rushed-ass release?

Maybe implement the changes people request to bring your code in from the 00s and you'd have more chance.

Verkister:
What even is the main difference between the to_chat and the apparently incorrect polaris preferred method?

SpadesNeil:
>What even is the main difference between the to_chat and the apparently incorrect polaris preferred method? Almost everything in the base code uses << and it works fine?

Legacy support. The method that works isn't the method that's correct.

SpadesNeil:
Needs discussion tag removed. No protest to ranged pounce. The range perk is counteracted by the cooldown drawback.

Cameron653:
Looking at it, it looks good to me. If it's been tested and everything works, It's got my support. 

The holo damage w/ the 25% chance to weaken looks pretty balanced. I'd say a 50% chance would be better balanced, since you need to be _right_ up on the person to actually use this, along with the fact it does

var/mob/living/T = input(src,"Who do you wish to leap at?") as null|anything in choices

Which means that if anyone else is within 3 tiles from you with the person, it'll be hard to hit on the fly, so I feel like a 50% knockdown chance is more apt.

It's much more balanced than it's initial variant in the PR, but at this point it feels like it'd be a bit too weak. 

Other than that, everything else seems to be general fixes & cleanup which is always good.

Screemonster:
The main issue with knockdowns that nobody's explicitly mentioned when it comes to balance things is that even a 1-second stun with a 10-second cooldown will still make you drop all your shit, and when the borg slams into you and fucks off dragging your LMG behind them, leaving you unarmed facing the rest of the sec team, that's a pretty sharp reversal

Verkister:
Yeah true. The balance stuffs were kind of messed up all over the place on citadel/tg in general back when the original pounce stuffs were added with the disabler gun being the only non-instant stun iirc. On both versions the pounce was pretty much just copied and refitted from the xenomorph abilities which I suppose were never intended to be balanced to begin with on either base heh. Polaris xeno pounce even seems to come with automatic grab in addition to the instant stun.

code/modules/mob/living/silicon/robot/dogborg/dog_modules_vr.dm
```diff
@@ -372,5 +372,5 @@
 	var/armor_block = run_armor_check(T, "melee")
 	var/armor_soak = get_armor_soak(T, "melee")
 	T.apply_damage(20, HALLOSS,, armor_block, armor_soak)
-	if(prob(25))
+	if(prob(40)) //Slightly higher chance to push than unarmed disarm shoves. Didn't feel like going all the way up to 50% tho.
```
SpadesNeil: no

Verkister: okay

SpadesNeil:
@Cameron653 
>Which means that if anyone else is within 3 tiles from you with the person, it'll be hard to hit on the fly, so I feel like a 50% knockdown chance is more apt.

I'm not comfortable with making it that high just yet. We'll determine how useful it currently is with a normal knockdown.

PR #2506, opened 2017-11-20
Some various map tweaks, and also carp nerf

SpadesNeil:
- Space carp now only spawn on the bottom level of the asteroid. Changed the message to reflect this.
- A few minor map tweaks and fixes. Mostly asthetic. Just threw some dirt around.
- Added Gaslamps to the maps, which will now wander around uncaringly until antagonized.
- Fixed access to R&D firing range.
- Added request console to Security and Engineering head offices.


SpadesNeil:
>!! FAILURE !! [ZAS: Mining Area (Vacuum)]: Pressure out of bounds: 101.325 | Turf: the floor |  Location: 24 // 88 // 7

That's weird. I can't find any issues at those coordinates.

PR #2507, opened 2017-11-20
Gives Tas coat same allowed items as HoS oat

Cameron653:


code/modules/vore/fluffstuff/custom_clothes_vr.dm
```diff
@@ -503,6 +503,7 @@
 /obj/item/clothing/suit/storage/det_suit/fluff/tas_coat
```
SpadesNeil: Why not just make it a sub-class of the HoS coat?

Cameron653: No idea. I didn't want to fiddle w/ it too much, just fix the single error on it.

SpadesNeil:
-shrug-

PR #2510, opened 2017-11-22
Adds a traitless nutriment sink speed ability.

Verkister:
-Activating it gives a slight boost to movement speed (40% of minor haste) with the drawback of full apex mode hunger drain (both passive and motion based when active).
-Added the ability for couple species (sergal, akula, xenochimera)
-Mostly just did it because of such predatory species being locked to the pathetic default drain that gives very little incentive to even search for a snack vendor more than once in a blue moon.

Opinions?

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/station_special_abilities_vr.dm
+++ b/code/modules/mob/living/carbon/human/species/station/station_special_abilities_vr.dm
@@ -669,9 +669,6 @@
 			C.absorbing_prey = 0
 			return
 
-
-
-
 /mob/living/carbon/human/proc/succubus_drain_finialize()
 	set name = "Drain/Feed Finalization"
 	set desc = "Toggle to allow for draining to be prolonged. Turn this on to make it so prey will be knocked out/die while being drained, or you will feed yourself to the prey's selected stomach if you're feeding them. Can be toggled at any time."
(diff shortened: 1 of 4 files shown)


SpadesNeil:
I do not think this is wise.

code/modules/mob/living/carbon/human/species/station/station_special_abilities_vr.dm
```diff
@@ -841,3 +838,22 @@
...
+/mob/living/carbon/human/proc/hastemode()
+	set name = "Haste Mode"
+	set desc = "Run slightly faster with high nutriment drain."
+	set category = "Abilities"
+
+	var/mob/living/carbon/human/C = src
+	if(C.haste_run != 1)
+		C.haste_run = 1
+		to_chat(C, "<span class='notice'>Haste Mode enabled.</span>")
+		C.species.slowdown -= 0.2
```
Screemonster: wait, if I were to use this as a xenochimera, would this not affect _every chimera in the game?_ You're changing the _species_ values.

Verkister: Hm yeah that might be quite an oversight I forgot to think about.

Screemonster: You could achieve the effect by putting an "if (haste_run) tally -= 0.2" in movement_delay() somewhere after this line:

https://github.com/VOREStation/VOREStation/blob/1aa1843fff22ebba745a24cd3a8480746e79bde5/code/modules/mob/living/carbon/human/human_movement.dm#L7

That wouldn't affect metabolism or any of the other species things though.

PR #2514, opened 2017-11-22
Gives Reasonable Species Brutal Predation

Cameron653:
- Gives Xenohybrids Brutal Predation, because they have sharp teeth and claws.
- Gives Teshari Brutal Predation, since they also have sharp teeth and claws.
- Gives Unathi Brutal Predation, ditto
-  Gives Tajaran Brutal Predation , ditto
- Gives Vox Brutal Predation, ditto.
- Gives Flatland Zorren Brutal Predation
- Gives Highland Zorren Brutal Predation
- Gives Vulpakins Brutal Predation

Forgot to give these species the verb when I was initially adding in Brutal Predation

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/station_vr.dm
+++ b/code/modules/mob/living/carbon/human/species/station/station_vr.dm
@@ -121,6 +121,7 @@
 	name_language = LANGUAGE_BIRDSONG
 	color_mult = 1
 	can_fly = 1
+	inherent_verbs = list(/mob/living/carbon/human/proc/shred_limb)
 
 	min_age = 18
 	max_age = 80
@@ -183,6 +184,7 @@
 		"You feel uncomfortably warm.",
 		"Your overheated skin itches."
 		)
+	inherent_verbs = list(/mob/living/carbon/human/proc/shred_limb)
 
 /datum/species/fl_zorren
 	name = "Flatland Zorren"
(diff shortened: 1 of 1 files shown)


Dhaeleena:
how many species have this by this point?

Cameron653:
Sergal, Akula, and custom species.

Hmm... Thinking about it, I should also give Tajaran it.

Cameron653:
Alright. Checked all the species edited to make sure their verbs weren't accidentally overwritten.

Cyantime:
Why not just give it to everyone at this point?

SpadesNeil:
That's just it. I don't want everything to have it. D:

SpadesNeil:
But, that is just me, hence the tag.

LittleBigKid2000:
Why not give it to Vulpkanin? They have sharp teeth and claws, too.

Cameron653:
Whoops, forgot about vulp.

Also, @SpadesNeil The verb is intended to be for scenes, not an actual mechanical feature to be used & whatnot. If someone does a HV scene as a species, like xenohybrid and then wants to actually mechanically take a limb/organ out without having to use a hatchet or whatever, this makes it easy.

There's so many restrictions on it that it makes it completely non-viable for abuse.

SpadesNeil:
Eh good point.

Cameron653:
Hmm. I'll need to give Promethean their own custom shred limb verb sometime where they actually turn the limb to goop instead of removing it, but that's for another time.

This PR is ready for merge.

SpadesNeil:
Let's not give Prometheans any such thing please. Else it's literally just humans, Diona, and Skrell who don't have brutal predation, and that is silly.

Cameron653:
Mm, very well.
~~PR is ready for merge aaahhh~~

PR #2515, opened 2017-11-23
Allows prometheans to change their hair color seperately from body color, and also headpat

Cyantime:


Cameron653:
You should give the hair changing ability to Polaris upstream.

Cyantime:
Prometheans can't have hair color that isn't their body color on Polaris, and they're the only shapeshifter species.

Cameron653:
I'm assuming that has to do with the monochromatic variable

Cyantime:
Right, that's also what made slimes that change their body color also change their hair color to the same.

Belsima:
```
November 23, 2017
[8:37 AM] Belsima: https://github.com/VOREStation/VOREStation/pull/2515

Just saw this downstream with a 'send to polaris' tag. You want, or no?

[1:22 PM] Mechoid: ..That would be incredibly nice
```

From Promethean loremaster. May want to return that Polaris tag.

Cyantime:
Closing this for now, will re-open with changes depending on what gets merged into Polaris

Cyantime:
Re-opening and updating to match the polaris changes

Cyantime:
Travis fell over and died, can someone give it a kick.

Cameron653:
Don't edit prometheans.dm. That'll automatically be changed when we sync with polaris.

The other two are fine, I'll just have to fix the conflict with the proc being in our codebase and theirs.

Cyantime:
I had just changed it to match the Polaris file, in case this is merged and the server is updated before a Polaris sync. Should I unchange that?

Cameron653:
Correct. It'll just cause conflicts otherwise.

Cyantime:
Done

Cameron653:
Alright, that's better.

There will still be conflicts, as species_shapeshift.dm is being edited, but we won't need to adjust prometheans_vr when the sync occurs.

Cyantime:
There'll be conflicts, but it shouldn't matter which side is used to resolve them, because they'll both be identical. Unless I'm completely mistaken about how conflicts are resolved like this.

Cameron653:
You're correct. They'll both be identical, but Github will still scream at me that the files are completely different even though they're not. It only takes me about a minute or two per file for something as basic as this, however. So it'll be fine.

PR #2522, opened 2017-11-27
Mobs Lazy Spawner

OrbisAnima:
This is an upgrade of the current mob spawner, what it does is use a
process identical to that of the proximity scanner to check for people
nearby, and if found, it will spawn a mob using the usual rules of the
spawner to get a new mob.
- It wont scan for new users if its on the mob cooldown.
- It comes with a range var to set which tiles should be scanned.

The idea behind this was to prepare a map with the proof of concept, and
launch this lazy loader for the wilderness, since it would fix most of
the lag issues we had given that mobs would only show up when they are
needed.

### DIFF
--- a/code/game/objects/mob_spawner_vr.dm
+++ b/code/game/objects/mob_spawner_vr.dm
@@ -104,3 +104,43 @@
 		if(T.z == L_T.z)
 			return 0
 	return 1
+
+
+/*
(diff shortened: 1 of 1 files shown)


OrbisAnima:
@Arokha review requested

SpadesNeil:
If you get the chance, set up the currently unused wilderness maps so I can re-enable them.

OrbisAnima:
I been more busy than I liked too it seems... I dont want to promise I will, but I will try

SpadesNeil:
Tis fine.

PR #2523, opened 2017-11-27
Item re add: Meta Glasses

OrbisAnima:
Re added the old style metamorphic glasses that changed shape with the
drink it had in them. I left clear examples of how to add new drinks if
needed, as the current itteration only have the original sprites.

Since these glasses are supposed to be rarer, I added the following ways
to get them:

-With one coin, a booze mat machine will be able to vend you ONE.
-Mining abandoned crates have a good chance to get them, and half of
that chance to get a box with four of them.
-Contraband crates will have a decent chance to have a box of 4 of them
-The trashpile has a rare chance (beta) to get a single glass.

This was supposed to be the end quest reward from the Resort Gateway
map, sadly, the project is no longer viable, and thus was dropped, we
still felt the effort shouldnt be wasted, so both this code and the lazy
spawner were released.


OrbisAnima:
@Arokha review requested

Anewbe:
I'm pretty sure the reagents themselves have a var to check what image the glass should be taking, so you don't need that full list.

OrbisAnima:
Are you sure? because when I followed the original code I saw that one var that was consistent with it, and it seemed to make sense to bring it. You could remove the list and see what happens, if it shows the drink or not.

What I will admit I had no idea what it was for was the whole x,y for the image position/centering, I only assumed it was for that, copy pasted it and called it a day, since I had no idea how ugly or not it would be if I did not brought it with me for the port.

OrbisAnima:
@Anewbe , do you want this in polaris, rather?
@Cameron653 does this need any changes for merge?

Cameron653:
Not that I can see, assuming this is just a glass that makes the old drink sprites visible when a drink is put in it.

I'd personally like them to be abundant, as the unique drink sprites were really neat and made bartending more fun as you could mix things and have the bar counter look fancy with a bunch of crazy drink mixes. Now we just have really bland and dull looking drink glasses, but that's just a personal gripe.

Anewbe:
I'd appreciate it on the Polaris end, assuming it works.

OrbisAnima:
It does, I always test stuff before making the PR.

so, lets talk specifics then, @Anewbe , Polaris is your server, I can port the items with just that file, but you need to tell me how you want them distributed.... I rather not touch the map at all if it can be avoided, so, how should we do with that?

https://cdn.discordapp.com/attachments/282670646159343617/366970693348491264/unknown.png

https://cdn.discordapp.com/attachments/295519200779436043/366970774487040000/unknown.png

should the Bar Vendor give them for free, should it take the coin, should it be a prize for searching stuff, or be a thing of the abandoned crates?

tell me the specifics and I will close this one, then make the pr to polaris, since Virgo gets the downstream, the item would still be received

Anewbe:
Just toss like 10 of them into the bar machine, don't worry about the map.

OrbisAnima:
Understood, closing the pr then

Issue #2449, opened 2017-11-10
Colorable bear ears do not preserve color when being sleeved


Dinnel:
#### Brief description of the issue
One of my characters has colorable bear ears as part of her appearance. When she is resleeved, her new body has ears that seem brown or yellowish instead of being black.

#### What you expected to happen
For the new body to have black bear ears

#### What actually happened
They were not

#### Steps to reproduce
- Make character with black bear ears
- Make sleeve of said character
- Behold!

#### Code Revision
- release - a5844f786ef2a09948e6718a7d418b05d95f9324 
 
#### Anything else you may wish to add:
- Color of ear is 82 on all values , used custom species, Tajara as base.

SpadesNeil:
@Arokha This

PR #2524, opened 2017-11-27
Fixes cloning fucking up ear colors and wings.

Verkister:
-Yeh this fixes those.
kuva

### DIFF
--- a/code/game/dna/dna2.dm
+++ b/code/game/dna/dna2.dm
@@ -33,13 +33,13 @@
 #define DNA_UI_GENDER      14
 #define DNA_UI_BEARD_STYLE 15
 #define DNA_UI_HAIR_STYLE  16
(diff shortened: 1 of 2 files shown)


Dinnel:
If that's right, then this fixes this issue I had before. https://github.com/VOREStation/VOREStation/issues/2449

Verkister:
Yeah it does.

Cameron653:
Does this also fix the one machine that lets you make a sleeve?

Verkister:
Probably. Seemed the dna code had it snipping out the uis past 27, and if the issues on that machine are dna related and the symptoms including the same borkage from ear color blue level, secondary ear color, wings, and wing color, then I'm pretty sure it could be the very same issue with this.

Cameron653:
Alright. Happy to see someone found out the cause of this issue.

PR #2525, opened 2017-11-27
Revert "Makes mice unable to push things around."

Cyantime:
Reverts VOREStation/VOREStation#2389

This ~~is being~~ has been done in a cleaner way on Polaris.
https://github.com/PolarisSS13/Polaris/pull/4360


Cameron653:
Next update we'll hopefully do a sync, so we'll get the Polaris version next update.

PR #2533, opened 2017-11-29
De-equips combat hardsuit from stolen package

Cyantime:
It's still a walking tank, just not a walking tank with a built in laser cannon, grenade launcher, thermal scanner, combat chem injector, etc.


SpadesNeil:
I am okay with this.

Ketrai:
Well do add a different way to get it ingame? It's still interesting for security in case of emergencies. Then again in events mins can always spawn one if they see fit so.

PR #2537, opened 2017-12-01
Adds ability to change gender identity mid-round

Runa-Dacino:
Adds the ability for any carbon to change their gender identity mid round. This does NOT change the sex of the individual. Tested and works.

**Reasons for this feature:**

1. Makes it possible to "hide" your gender identity on chargen until you get resleeved or TF'd. At the moment, getting resleeved wouldn't change the identity, forcing you to either start as target gender and reveal your cards, or be stuck with non-prefered pronouns until round end or despawn/respawn. Basically, support for dubious con TG/TF.
2. Enables people to "disguise" their gender. Useful for drag, trap and sissy folks.
3. Enables gender fluid people to be fluid without being squishy 

Includes changelog for player notification

### DIFF
--- a/html/changelogs/RunaDacin-GenderIdentityToggle.yml
+++ b/html/changelogs/RunaDacin-GenderIdentityToggle.yml
@@ -1,0 +1,5 @@
+author: Runa-Dacino
+delete-after: True
+changes: 
+  - rscadd: "Adds the ability to change gender identity on the fly. It can be found under "IC" tab."
+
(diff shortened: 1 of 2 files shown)


SpadesNeil:
I see some hilarious shenanigans occurring as a result of this.

html/changelogs/RunaDacin-GenderIdentityToggle.yml
```diff
@@ -0,0 +1,5 @@
+author: Runa-Dacino
```
SpadesNeil: We don't use the changelogs. Those are for Polaris. You can delete this entirely.

PR #2542, opened 2017-12-02
Add pAI bears

Dinnel:
Because robo teddy bears!

image


Cameron653:
Don't modify pai.dmi, only pai_vr.dmi

Dinnel:
I'll undo that. Thought I had to be extra careful.

PR #2543, opened 2017-12-02
pAI Bear Take-2

Dinnel:
I had no idea how do undo what I did before, so I made a new pull request. This time it only changes the pai_vr.dmi file.
I learned that I shouldn't modify the other file because apparently it causes conflicts and ours uses the pai_vr file only anyway.


Cameron653:
Looks good to me.

Belsima:
Have you looked into having this for Polaris? I can think of a use for it.

SpadesNeil:
I don't think Polaris would want this but they are welcome to grab it.

Dinnel:
You are free to use the sprite for whatever, so go ahead if you want!

Belsima:
Merci

PR #2548, opened 2017-12-03
Slimes that fail to *flip lose limbs

Cyantime:
*squish


SilverTalismen:
More like splat!

OrbisAnima:
Careful though, while I support the hillariousness of prommies going splat, the possible problems I see with this one is things like aliens losing their limbs and having no way to get it back, granted, they deserve it for morons, but its something to have in mind, since anything with non destructible limbs might get the effect, and not prommies who have the regrow limbs feature.

Cyantime:
The only species that this effects are prometheans and diona, both of which have the regenerate limbs verb.

PR #2550, opened 2017-12-04
Adds a non-restricted shaved beard that works for our custom species.

Verkister:
-Why in the shit would the only beardless option be restricted to polaris species only?


SpadesNeil:
Conflict.

Verkister:
Was apparently done in te xenochimera icon base pr as well. Closing

PR #2555, opened 2017-12-08
Tweak dogborgs

Arokha:
Tweak dogborg interfaces, change drug injection routine. No longer autoinject inaprov because it will make tricord whenever anyone tries to inject them with dylo. It's a manual button. Remove inaprov from janihounds, too. (I fixed that weird space above the inaprov button between the word 'Injector' and the first button, but only after the screenshots)

image

image

image


Arokha:
Changing something.

PR #2579, opened 2017-12-16
Fix compile warning

Arokha:
Cyantime removed the uses of this var but didn't remove the definition, so this removes that too, to prevent that compiletime warning.


Cyantime:
That's what I get for using the github editor, sorry about that.

PR #2581, opened 2017-12-18
Wilderness Optimization

OrbisAnima:
-Re enables the wilderness, with the temple and a rather poor crash
site.
-Updates the Lazy Mob spawner with some more optimizations, and a few
uses.
-Leaves a version of the crashsite with the original items and the new
system, I wasnt able to make it work without butchering it, and since I
am not that good of a mapper, I opted to leave it on the hands of people
that know what they are doing to improve the original, and making it
work, while leaving one that works.
-Changed the path of the teleports just to fuck up with Ketrai :)

Note to admins: Since this introduces 8 new z levels, all of which have
a check to generate or not a tree (think roughly a bit less than 100k
tiles checking for those), game init for objects went from 32 seconds to
71 seconds, so expect a rougher start on update.
Note to Ace: Does this counts as my Gateway Contest map? :P

additional notes: If performance takes a severe hit, init could be
severely reduced by reducing the trees chance to 2, or even placing
trees by hand on the maps.


OrbisAnima:
@SpadesNeil , please Review

SpadesNeil:
Glad to finally have this back.

OrbisAnima:
I tested it twice and had no crashes or problems with this commit. Real test begins when it faces all the players the server has to offer though.

Yoshax:
A more than double increase in game object initialisation time doesn't like a good tradeoff. Nor does adding countless more mobs to the map when we already have the mob controller inexplicably climb to 500+% usage.

SpadesNeil:
Even when it was done **far less efficiently**, trust me, it wasn't nearly as bad as you make it sound.

OrbisAnima:
Strictly speaking, this is more of a framework. Ideally real mappers would come and actually fill that wilderness with more interesting things than just lazy spawners and random trees in a maze like area. And the initialization could be reduced to 32 only by setting all the tree chance to 0. 
As for the mobs, they are only there when players activate them, if no player ever goes into the wilderness, there wont be mobs.

OrbisAnima:
(Also, it would be nice to have mappers of the away missions follow the example with the lazy spawners, and use them for the away missions, that would not only help with the lag of away missions, but avoid ghosts from seeing where those appear. Also, as we have trees, we could also implement the same tree chance for spawners, which would make it so nobody knows where the spawners are set.)

Yoshax:
It being better than absolutely awful isn't a reason to add it @SpadesNeil.

I don't need to say anything other than our mob controller is almost constantly at 500%. Not one thing relating at all to mobs should be merged until that is resolved.

SpadesNeil:
The mob controller thing, iirc, started when we added respawning mice. Is it possible they aren't getting deleted properly?

SpadesNeil:
Also can someone **please** make an issue post about the mob controller issue? Because that's why most bugs don't get fixed.

When a developer looks at "Issues" and says, "I'm gonna fix some of these today!" they end up being reminded only of the bugs that people _bothered to write down_.

OrbisAnima:
you could make the mice spawner to use the scanner instead of the base controller.

OrbisAnima:
@Yoshax , what is the problem with the mobs?

Yoshax:
I dunno. :man_shrugging:

OrbisAnima:
You say the mob controler climbs up to 500%, thats what I am asking, I had seen this, but it seems to only happen at char init, on my test with this PR that happened and then it went back to 2, with spikes that went to 10, even when running like a lunatic on the wilderness or getting eaten inside the temple due to running out of charge, I havent got any significant ping on the mob controler.

Yoshax:
It happens all shift, every shift. Live on server right now it's a 642%.

OrbisAnima:
Were you able to reproduce it locally?

Yoshax:
Haven't tried. But it happens consistently every round on the server. 668% now.

OrbisAnima:
My tests show the MC going bananas when I  join with a character. stabilizes less than two minutes after that. This is with my PR or not.

After the stabilization period, the mobs and the spawners do not generate any insignificant blip, that is to say, anything bigger than 25, and rarely go over 10 on my rig, which again, its a laptop, I doubt the server is running on a laptop.

I think its the human mobs you guys want to block, and rethink if there is anything that might be creating issues, rather than the simple mobs.

SpadesNeil:
I remain suspicious that the act of killing/spawning mice is what is murdering the mob controller. In my experience, the issue only appears when mice are being slaughtered enmasse. I bet I know how we can test this.

OrbisAnima:
I admit I havent been murdering mouse en masse during my tests... but, given how the base spawner works, if you kill all the mice of the server at the same time, you would trigger all the spawners that generated them at the same time, this wouldnt be different than 20 players running around in different areas of the wilderness at the same time killing everything

Cyantime:
Hello I made the mouse spawners. When they get a report that the mouse linked to them has died the respawn timer is set for some point between then and one hour from then, they do not all respawn their mice at the same time.

OrbisAnima:
I got an idea then, @Cyantime , think it would be possible to make a PR to block all spawners from working?

If we deploy that and see on the server, and then check the MC, we could see the differences and find the culprit.

OrbisAnima:
I spoke with Cyan, going to make a separate PR to disable all the spawners, if anything it will show wherever that is the true origin of the MC, and if it makes sense or not to block this PR.

I will ask you guys to merge that one asap on the main server, since 668% usage of the MC is significant. 

Personally, I think there wont even be any difference with the spawners on and off. but we are all humans here, errors happen.

Arokha:
The slowdown is from human Life() calls and was improved somewhat by @Leshana on a Polaris PR we have now. Has about a 50% reduction in time by just redoing how it generates the little health puppet on the right side.

Arokha:
So how do people get to the wilderness? And how can we cut down the initialize time?

SpadesNeil:
People get to the wilderness via an opening at the bottom of the map that has/had teleporters mapped in. Currently, those teleporters are mapped into a wall and the wall is impassible, but the spot is still visible by a little patch of dirt along the south wall. Players enter the wilderness through this opening.

The rest is a Lost Woods style maze.

OrbisAnima:
The init Time is big not because of the spawners, but because of the trees.

If you want to cut down the init times, make the trees by hand and remove the tree chance from the turf, that will reduce it to normal levels again.

In this PR the tether map was updated with the removal of that wall.

OrbisAnima:
Interestingly, I feel like the tree chance thing would bebetter to be used for the spawner themselves, making it so players dont know where the heck mob spawners are, instead of the trees, but since it has so much of an impact to init times...

Not to mention the other bit, in that it doesnt matter for the players to know where the spawners are in the wilderness... is just effin full of them, but, something to think in the future for certain maps, might be a good idea to consider one such ocassion.

Arokha:
I mean as an alternative to "every turf in the wilderness needs to decide if it gets a tree" why don't you just figure out how many you want on average (based on what the spawn chance is) and just have ONE object in each z that rolls a rand(min,max) for how many trees you're getting, then rolls rand(x,y) for where it's going to put them, checks to make sure the spot is the type of terrain you wanted, and puts them down.

This reduces the processing time from "We're going to run code for every tree we get AND don't get" to "We're going to run code for every tree we get" which is less, I hope. The same thing could be applied to mob spawners.

OrbisAnima:
Honestly? I though about it but decided against since it was out of the scope of what I Was doing, you know, just the whole monster spawning. The trees were something that someone already did, and were just using...

I like your approach, and I think it would be the better solution for this problem.

maps/tether/tether.dm
```diff
@@ -21,7 +21,9 @@
 	#include "tether-08-ships.dmm"
 	#include "tether-09-empty-surface.dmm"
 	#include "tether-10-empty-space.dmm"
-//	#include "tether-11-wild-surface.dmm" // Wilderness stuff removed until mobs can be optimized better.
+	#include "tether-11-wild-surface.dmm"
+	#include "tether-12-wild-crash-alt.dmm"
+	#include "tether-13-wild-temple.dmm"
```
Arokha: For now, why not remove these #includes and the tether-01 change and any other changes that enable the maps, and we can merge the rest of the other changes/fixes in. Then Leshana and I or anyone can work on optimizing the large startup time penalty and map gen stuff. In fact, it can probably be optimized using the new /tg/ maploader.

OrbisAnima: tomorrow I make the changes then

OrbisAnima:
@Arokha , changes were made as you requested

SpadesNeil:
So now what? We merge? :U

OrbisAnima:
Thanks for all the fish!

PR #2591, opened 2017-12-21
Cleans up two custom item files

Cyantime:
Removes all relative pathing.
Adds to_chat() wherever needed.
Removes some `src.`

### DIFF
--- a/code/modules/vore/fluffstuff/custom_clothes_vr.dm
+++ b/code/modules/vore/fluffstuff/custom_clothes_vr.dm
@@ -89,12 +89,12 @@
 /obj/item/clothing/suit/storage/hooded/wintercoat/roiz/proc/ToggleHood_roiz()
 	if(!suittoggled)
 		if(ishuman(loc))
-			var/mob/living/carbon/human/H = src.loc
+			var/mob/living/carbon/human/H = loc
 			if(H.wear_suit != src)
-				H << "<span class='warning'>You must be wearing [src] to put up the hood!</span>"
+				to_chat(H, "<span class='warning'>You must be wearing [src] to put up the hood!</span>")
 				return
 			if(H.head)
-				H << "<span class='warning'>You're already wearing something on your head!</span>"
+				to_chat(H, "<span class='warning'>You're already wearing something on your head!</span>")
 				return
 			else
 				H.equip_to_slot_if_possible(hood,slot_head,0,0,1)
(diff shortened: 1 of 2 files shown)


SpadesNeil:
I feel like there was a reason @Arokha did them this way but I am not sure.

Arokha:
Why remove the relative pathing? I'd agree it's sort of bad most places, but makes this file which is edited VERY OFTEN much easier to visually navigate. Relative pathing has shenanigans when used willynilly but I think this use is probably fine.

The removal of some unnecessary src, fine. The to_chat() changes, cool. But I dunno about removing the relative pathing. That's a coder QOL thing for these files.

Cyantime:
Personally I think it's the opposite of QOL and makes the file more obnoxious to read and to search in, you can't search for `/obj/item/clothing/under/rank/khi/cmd` and find it when it's split up into relative paths.
Keeping all paths absolute makes it entirely clear what is what, and is a standard across all SS13 codebases for a reason.

Arokha:
While I agree that's the standard for pretty much every code file, I'd make the argument that this file isn't really so much actually 'programming' and more just 'design'. Like people aren't going to search for a path most of the time, they're going to search for a character name or ckey, and it helps visually to see where each item and the versions of said item end, demarcated by the next full-path-entry in that file. It makes those character/ckey comments so much harder to notice when scrolling when every path abuts the left margin. Since this is the only file where ostensibly the comments are more important than the paths in terms of finding what you want, that's why it was like that.

Originally I proposed that each fluff person just have their own .dm file by ckey. It's not like we'll run out of filenames. And if you don't want a huge thing in the left, don't expand the fluff_items folder or something, but since that was denieeeed... I'm just trying to keep this file sort of 'readable'.

Cyantime:
If the comments are what is being searched for, why keep it relative at all?

Things like this
relative1

or this
relative2

or this
relative3

have no reason to be kept relative. The next person's fluff items can be clearly identified, none of the paths are long enough to make that a meaningful issue, and even if they were, the comments that you just mentioned are their own method of splitting up different ckey related items.
Death to relative pathing.

Screemonster:
Considering that there have been times where I've had to fix fluff items because the base item they're based on has changed path (the HoS coats being one particularly memorable thing - all custom hoscoats broke because they changed from jacket to armor upstream or something) and being able to change all the coat/hos/fluff in one go by changing a single line and not having to ctrl-f every single now-broken absolute path (which is what I _did_ have to do, and it was a fucking _pain in the asshole_) would have been a godsend

PR #2598, opened 2017-12-21
Debug Action: Disable Mob Spawner

OrbisAnima:
This disables all spawners to help Virgo track down why the Mob
Controller is at 668%

Relevant to #2581 discussion.

### DIFF
--- a/code/game/objects/mob_spawner_vr.dm
+++ b/code/game/objects/mob_spawner_vr.dm
@@ -51,7 +51,7 @@
 	return 1
 
 /obj/structure/mob_spawner/proc/choose_spawn()
-	return pickweight(spawn_types)
+	return 0//pickweight(spawn_types)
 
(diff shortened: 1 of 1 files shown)


OrbisAnima:
@Cyantime , followed your instructions, thanks.
@SpadesNeil , if we can get this merged and then tested by one of the devs, it would help us clear the Issue reported by @Yoshax .

I suggest that samples are taken before and after this change. With round time and characters playing please. Hell, its not even needed to merge this change, just perform it on the server and compile would suffice.

Arokha:
The high tick consumption is due to a player mob, not another spawned mob. I've deleted every mob/living on the server other than humans and robots, and the tick usage remains high, indicating it's code related to a player mob. Consequently, I don't think this will help.

OrbisAnima:
Roger, closing the PR then, Thanks Aro!

PR #2600, opened 2017-12-21
Add ability for 'extensive' mob controller logging

Arokha:
Requires mid-round var edit to enable, can be used to determine how much of each tick each mob's Life call uses. Will not tell you WHY, of course, but will aim you at a specific mob hopefully. Worst case, we find out that it's actually just every mob.


OrbisAnima:
now this is interesting.

PR #2607, opened 2017-12-22
Teshari no longer get passtable

Yoshax:
It's quite silly and strange that Teshari get this while no other species does. It causes more problems than it is worth. Many of the Teshari players that spoke out in Discord do not want it either. 

Future solutions that are out of scope:
- Make only mobs under a certain scale get pass table

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/station_vr.dm
+++ b/code/modules/mob/living/carbon/human/species/station/station_vr.dm
@@ -337,6 +337,7 @@
 	min_age = 18
 	push_flags = ~HEAVY //Allows them to use micro step code.
 	swap_flags = ~HEAVY
+	pass_flags = 0
 
 	inherent_verbs = list(
 		/mob/living/carbon/human/proc/sonar_ping,


Arokha:
There's a better place to do this, at the bottom of station_vr.dm there are overrides for the existing species. Can you stuff this down there instead? Setting it to 0 is the equivalent (species default is 0).

Draycu:
I must respectfully disagree with this sentiment that it is an actual issue.  The fact that Teshari are able to also still fall from railings, yet would be unable to jump over tables as they would with their small agile bodies would be unacceptable.  

I must recommend that this change only be adopted if Teshari are also not able to fall from ledges with rails blocking them as well.  They are weak enough as it is, and the loss of being as mobile as they are would greatly harm the enjoyment of playing as Teshari.

I also feel this might prevent Teshari from jumping over Hydroponics trays, which would also be a major detriment to playing as the small, vulnerable, and quick species.

Draycu:
As something further to ask, would this not conflict with the Teshari's ability to hide _under_ tables to simple pass through them anyway?

Cyantime:
PASSTABLE is the flag that allows tesh to pass through railings, this disables that as well.

Draycu:
Good, glad to hear this.  However, it would still be an issue when working in hydroponics.  One of the primary reasons I enjoy playing as a Teshari is the ability to pass over the trays.  

My main question, then, is why this might be considered an issue needing to be fixed.  It was mentioned this current feature causes problems, what problems?

Cameron653:
If you want to make it so teshari have all functionality but don't go through railings it can be done in about two seconds.

https://github.com/VOREStation/VOREStation/blob/db8a51cf98bfcc837552600b1d71250c7d8f6afe/code/game/objects/structures/railing.dm#L40-L41

Comment out these two lines and the issue which most teshari have (the railings) will be solved and they will be able to have their original functionality. 

By commenting out these two lines instead of removing the flag you'll make it so they don't fly over railings while also not gutting the race entirely.

Draycu:
_Thank you!_

SpadesNeil:
I'd rather just not allow Teshari to dart over tables. It's incredibly annoying.

Draycu:
It's annoying?  For who?  You watching them and not getting to do so as another species?  It is what makes Teshari unique.

Cameron653:
If their table walking is removed, they need a rework/readjustment.
The table walking is the only positive trait they have w/ the exception of Haste plus.

Taking a look at custom species:
You start with 1 point
Haste+ Is -5 points (Costs 5)
50% HP is 3 points (Gives you 3)
Teshari have 35% extra damage to both brute and burn (a bit more than the normal weaknesses. 10% to be exact. So 4 points at the min.)
That's 8 points for a 5 point trait, leaving three points that need to be allotted somewhere.

Yoshax:
Standard species are not balanced around our points system. There's no obligation to attempt to do so. If you wanted to play that game, then we should look at all the species you've given constant free traits to, and smack them with some heavy downsides.

SpadesNeil:
>It's annoying? For who? For watching them and not getting to do so as another species? It is what makes Teshari unique.

For trampling all over the table like zippy little spaz monkeys with feathers while I'm trying to enjoy a nice discussion and food in the bar, for example.

No species _should_ have this trait. It's obnoxious. You imply I'm jealous. The ability should not exist to begin with.

Arokha:
I'd agree with removing `PASSTABLE` but I do think you should probably balance the species to be in line with the points at least. When we were adding the points, the discussion included that the stock species should always have 'slightly better' traits than attainable with points, so as to encourage people to play them rather than snowflake species. I'd at least reduce the health downside a bit.

Cameron653:
"Standard species are not balanced around our points system. There's no obligation to attempt to do so."

A species shouldn't be _that_ far in the negative. They have one upside and three downsides if this PR is passed.

Draycu:
This is a bit much, I must say, and feels hastily put through without full consent of all that play as Teshari, does it not?  

I did not mean to imply being jealous, I am speaking in terms that this removes the unique quality of this species.  At worst, make Teshari a whitelist then like Polaris does if people are abusing table running to annoy you.

Draycu:
Can you not use Cameron's changes?  This would satisfy those that play as Teshari.

Arokha:
>without full consent of all that play as Teshari

While I think they do need to be balanced a little, I just want to remind that it's not a democracy. We don't *ever* need to get consent from players before making changes. We want people to be happy, obviously, so we ask for opinions from time to time, but it's definitely not a requirement.

Draycu:
I see.  That's unfortunate.

Cameron653:
Why not code the future solution in along with this? This PR really makes no sense whatsoever.

Verkister:
Could make them do the climbing thing a lot faster to make up for the loss of passtable, even instant maybe, just kinda needing a manual input to go over instead of flat out ignoring the structures.

SpadesNeil:
... Y'know, we _could_ whitelist Teshari. :)

oh boy the outrage would be great but it would solve so many problems

AaronSherwood117:
I'd like to voice support for Cameron's version of fixing the issue.  Nobody has really been complaining that they shouldn't have the ability in chat.  "It causes more problems than it is worth. Many of the Teshari players that spoke out in Discord do not want it either." Both of these are false.  I was the only one who said it, and I meant in the context of the railing crap.  This is a really lazy fix made on impulse.  @SpadesNeil And I can't help but feel you're being awfully rude, Ace.  Actively mocking people who are upset at a half-assed change, cmon man.

SpadesNeil:
There's a big controversy over a one line change that removes a very small feature that really shouldn't have been added to begin with. If people want to get buttmad about it or leave, be my guest. It's not like we're whitelisting Teshari, although I'd rather be doing _that_ to be completely honest. Someone else suggested that possibility.

Pick something more substantial to be upset about, or expect people to laugh at you for being upset over things that aren't that important. What I am doing is not mocking, but I **am** ridiculing.

Screemonster:
smolcritters being able to hide under tables is neat (and cute) but it's something that should be gated to size rather than species. _Maybe_ give teshari a bonus that lets them tablehide at larger relative sizes than most species, but at the moment it's frankly silly that a 200% scale teshari can hide under a table as an eight foot tall feather-raptor, while a 25% scale barely-visible micro-tajara cannot.

Draycu:
@SpadesNeil, I am only passionate about preventing the species from being fixed with what seems to be a quick change without much care.  

The name of this code change, for instance, "lolteshari," stands out.  You say supporting keeping passtable is lazy coding?  A simple line change that this current suggestion is making, seems to be as well, in my eyes.

I'm honestly not trying to make a mountain out of a mole hill, but it is apparent that you are pushing rather hard for this blunt force code change.  

I know this isn't a democracy, but I know this is a community that simply wishes to enjoy role playing over mechanics.  Balance changes to mechanics like this isn't worth pushing so much for, when those that play as the species do not want them.

SpadesNeil:
I reiterate that sticking the species on a whitelist is the alternative solution. Would you prefer that? Because we chiefly hear complaints about Teshari above all, so there's clearly some aspects about Teshari that is encouraging bad behavior.

Screemonster:
People can name their branches what they want unless they're outright racist. I've lost count of the number of meme or obscene branch names and commit titles I've seen or created over the years.

> I know this isn't a democracy, but I know this is a community that simply wishes to enjoy role playing over mechanics. Balance changes to mechanics like this isn't worth pushing so much for, when those that play as the species do not want them.

Tell that to the people who used the tablehiding mechanic to run around and annoy people and robust them in the bar by hiding under tables to become unshootable and unhittable and thus making a lot of people negatively disposted towards teshari in general and tabledashing in particular.

Tell that to the people who got sick of tesh dashing into their departments before anyone could stop them, then dashing off faster than they can be chased because the person they just stole shit from has to climb a table (or go the long way round to the door) to follow them.

don't get me wrong, I _love_ teshari when they're played well, I think they're adorable. The mental image of a little birdything waitress at a diner nyooming along the counter and giving people their drinks at mach 3 is cute as hell. This isn't the behaviour people are complaining about though.

Compare and contrast diona - nobody gave a shit about their ability to split into nymphs, until one guy in particular became infamous for taking command slots and splitting at roundstart, usually right in the middle of the bar, to meme around and annoy people as a ventcrawling, thieving little jackass. It eventually got bad enough that people playing diona as the slow, thoughtful pacifists they're meant to be got sick of it because the default reaction of crew to a diona was "ugh, it's one of these tanky assholes again, how long before he turns into nymphs?"

Ketrai:
I wouldn't remove this. at this rate tesh will loose too much of their fluff, and people may just play custom species tesh instead of tesh as they're meant to be. which would be weird.

SpadesNeil:
The more people suggest not removing this, the more I consider setting Teshari as a whitelisted species.

_No_ species should have the passtable trait imo.

AaronSherwood117:
Threatening to hide the species behind the whitelist just because we don't want a certain change happening so we'll stop discussing seems a bit disingenuous, Ace.  If you're trying to say "I think some people in the past have abused this and such a feature should be whitelisted." instead of what it's coming off as, it would be nice to know that.

I do, however, agree that it could be removed in favor of a size-based system.  Scree also makes a good point with the history of Diona, and if people have actually been shitty with Tesh mechanics instead of these totally unsubstantiated claims we keep getting in discord chat, a whitelist might not be out of the question.

I just don't think an impulse feature removal made out of some sort of weird spite is a good way to handle changes.

Belatedly, mocking and ridiculing are the same thing.

Verkister:
Any opinions on the idea of replacing the passtable with non-delayed or faster climbing that would require a manual action instead of the movement just flatout ignoring the structures?

Screemonster:
I'd be happy with teshari to have buffs to tableclimbing, and _if_ passtable is linked to size, then a buff to how big they can be without losing it, just to avoid the immersion-shattering giant teshari under the table.

SpadesNeil:
@AaronSherwood117 I apologize if it came off as a threat.

Mocking and ridiculing are not the same thing but I suppose that isn't the point. It's disrespectful either way and I apologize for that as well.

Either way we have a problem with Teshari and I am not sure how to properly fix it.

Screemonster:
one solution would be to ban the ones that run around doing squawktide but that _has_ eased off a bit lately, I think a chunk of the anger people have towards 'em is just residual from when that was a thing.

PR #2611, opened 2017-12-23
Adds a proc to toggle passtable

### LINKED ISSUE #1522: Railings should stop Teshari movement
Teshari can walk straight off of railings

Railings stop Teshari movement unless you actually click and/or drag yourself over them

#### What actually happened
Lagged for a few seconds, got to watch my character simply walk off the railing

#### Steps to reproduce
- Play Teshari
- Walk towards railing

#### Code Revision
release - f2c9a3a54e980f6e578ad278651b12887b610f3c

Cameron653:
- Adds a proc to toggle passtable
- Any /living mob can be given this via VV, or have it coded in to spawn with
- Gives Teshari this proc so they can turn it on to prevent accidentally walking onto tables/falling off railings,

fixes #1522


Arokha:
Please make the text for `will now` and `will not` more different. The same length and 1-character-different enable/disable messages are easy to read wrong. Maybe `can` and `cannot` or `will now` and `cannot` or `will` and `will not`.

SpadesNeil:
This might be out of scope for this PR but can we turn passtable off for Teshari by default, but leave the option to turn it _on_ later?

Arokha:
Blarrrr. Actually can you move this to IC category? Abilities rarely has anything in it. Having it be gone entirely would be nice. (The 'Exit VR' thing is already removed in master)

Cameron653:
Changed. Made it say "will" or "will NOT". It's late and I can't think of any more of a clearer way to do it.

Moved to IC tab.

Also, they have the hardvore & Drain procs available to them, which are in the abilities tab, so they'll have the abilties tab anyway unless those are also moved to IC.

Screemonster:
#1522 is the relevant issue for this.

Cameron653:
Done. Set it up to auto close 1522 when this is merged.

Arokha:
You raise a good point but I don't think I'd be convinced of the usefulness of a tab until it gets like 4+ buttons constantly rather than like 1-2 buttons for some people with some traits. Maybe we should bring back the vore tab, idk.

Cameron653:
I'd personally like them to stay under abilities as it currently is, because IC has a lot of stuff in it and finding a vore-specific verb in it can sometimes be confusing. Having all the vore stuff organized neatly under one tab makes it a lot easier to find.

Arokha:
We can have a duel about it later, or maybe shift other stuff around to justify a new tab having more than like 3 things.

PR #2612, opened 2017-12-23
Fixed pAI OOC info

AbbieFland:
This is my first VS13 code change, but I think I didn't do anything stupid.

Previously, pAI OOC info didn't appear to be getting set correctly,
whether installed through the normal recruit dialog or through an
observer directly inhabiting a loose pAI card.

There was a VOREStation-specific check during pAI creation to copy over
OOC info from preferences, but I tested multiple ways and it didn't seem
to ever succeed.

This change moves that check from creation to chassis unfold. This is
tested to work correctly with recruit, observer-inhabit, and admin
force-pAI modes of becoming a pAI. OOC notes were already intentionally
only displayed for unfolded pAI, so the OOC notes not copying until they
unfold doesn't get in the way of anything.

This delayed check also has the unintentional bonus that if you play as
a pAI, you can change your OOC notes between the time you inhabit the
card and the time you unfold, in case you have different prefs for the
same character in normal mode and pAI mode.

### DIFF
--- a/code/modules/mob/living/silicon/pai/pai.dm
+++ b/code/modules/mob/living/silicon/pai/pai.dm
@@ -104,12 +104,6 @@
 	verbs += /mob/living/silicon/pai/proc/choose_chassis
 	verbs += /mob/living/silicon/pai/proc/choose_verbs
 
-	// Vorestation Edit: Meta Info for pAI's
-	if (client)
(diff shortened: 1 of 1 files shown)


AbbieFland:
After discussion, changed it to read the notes in login rather than unfold, as that's triggered by BYOND whenever the controlling player changes.

PR #2627, opened 2017-12-26
Updates expeditionary paths due to Polaris change

Cyantime:


Burrito-Justice:
Bless your soul

Arokha:
Wait, how does changing all these paths fix the icons

Cyantime:
`/obj/item/clothing/under/utility/expeditionary/whatever` only exists in solgov_vr.dm, which was used to override some Polaris names and descriptions. The actual clothes were changed to `utility/sifguard/whatever`, so the expeditionary versions just went back to using the basic `/utility` sprites without the colored markings.

Arokha:
OH. They're overrides. Gotcha.

PR #2629, opened 2017-12-26
Fix NIF surgery post-update

Arokha:
This line was put into the wrong block (a good example of why relative pathing is bad, since it made it unclear where it was before in normal code).

### DIFF
--- a/code/modules/surgery/implant.dm
+++ b/code/modules/surgery/implant.dm
@@ -122,7 +122,6 @@
 			return
 		if(affected && affected.cavity)
 			var/total_volume = tool.w_class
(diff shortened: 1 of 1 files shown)


Cyantime:
What the heck is that syntax
`if(istype(tool,/obj/item/device/nif)){var/obj/item/device/nif/N = tool;N.implant(target)}`
Curly braces and a semicolon?

Arokha:
The "I'm super-concerned about upstream mergability and want to make this a single-line change rather than a three line change with two more lines for begin/end comments for a vorestation edit" syntax.

Byond has always supported curlybraces and semicolons tho if that's what you meant. Just nobody uses them.

PR #2651, opened 2017-12-29
Makes the soulcatcher able to have 2048 characters

Cameron653:


Screemonster:
But where are you gonna find 2048 characters on a station that tops out at 75 on a busy night?

PR #2652, opened 2017-12-30
Swaps two words around.

Cameron653:
- Unless it was made to sound southern, that is.


Arokha:
Probably intentional. XD

PR #2658, opened 2018-01-01
no fun allowed

Cameron653:


Arokha:
Due to the way Travis failed (Travis was down at the time) I can't manually restart this build. Closing/reopening the PR is the only way to convince it that it's a good idea.

PR #2659, opened 2018-01-01
Prevents low health bodies from dying on creation

Cyantime:
Humans with the -50% max health trait will no longer die instantly when the growing process starts.

### DIFF
--- a/code/modules/resleeving/machines.dm
+++ b/code/modules/resleeving/machines.dm
@@ -70,7 +70,7 @@
 	H.original_player = current_project.ckey
 
 	//Apply damage
-	H.adjustCloneLoss(150)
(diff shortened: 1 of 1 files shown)


Cameron653:
Oddly enough, I tested this multiple times on my test server. Someone with 50% max HP would only take 100 cloneloss and not die when being cloned. No idea how/why, though.

Cyantime:
It might have been something to do with the order in which things were processed, it does 150 cloneloss which would bring a 50 maxhealth human to exactly the point that they'd die, but if the cloner processed before the mob it would bring them up to above the death point. Things might have just gotten out of order during that one round or something. Either way, this should keep it from happening in the future.

Arokha:
Due to the way Travis failed (Travis was down at the time) I can't manually restart this build. Closing/reopening the PR is the only way to convince it that it's a good idea.

PR #2662, opened 2018-01-02
Destroys botany forever

Arokha:
Why do we have a robot that literally does the entire job. Basically people leave it and it becomes a lag generator forever.

Removes farmbot harvesting and removal of dead plants, plus adds an idle timer for 150 process cycles without a target causing it to turn itself off. So basically it will tend the plants and keep them alive, but you need to come harvest them sometimes.

⬇️ Screaming goes below

### DIFF
--- a/code/modules/mob/living/bot/farmbot.dm
+++ b/code/modules/mob/living/bot/farmbot.dm
@@ -19,7 +19,7 @@
 	var/replaces_nutriment = 0
 	var/collects_produce = 0
 	var/removes_dead = 0
-
+	var/times_idle = 0 //VOREStation Add
 	var/obj/structure/reagent_dispensers/watertank/tank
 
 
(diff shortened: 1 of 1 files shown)


lbnesquik:
Farmbot is important for botany because it is realy tedious otherwise. I don't think there is many other jobs that realy need so much attention on the gameplay while said gameplay isn't social interaction. (Like bartender who has a job where they just interact.)

Chargae:
I don't see an issue with farmbots being a thing. Just with them doing _literally every part of the job_ except collecting the stuff it harvests.

Which also happens to make walking past botany a laggy mess once it's been going for an hour or so.

Verkister:
With a farmbot that does everything it becomes a zero attention job.

MrSign:
No issue with this. Prevents botany from becoming a laggy mess AND promotes to botanist that they still gotta do their dang job in providing fruits and veges to the kitchen, instead of leaving it for the chef to wander over and pick it up.

SpadesNeil:
Marking this as 'feature' since we're technically just removing a badly coded and badly thought out feature rather than fixing a bug.

OrbisAnima:
Alternatively, you could remove the robot entirely and then just add that feature that TG had in which you connected trays that were togeder and they all shared the water they received, making it less of a pain to do the job.

lbnesquik:
We got two extremes. Either no attention or full attention. Can we bring both closer toward the middle?

Very-Soft:
This seems like it is the middle. The bot will maintain your plants for a while, but you have to do the actual harvesting. You literally plant what you want planted, and come back to harvest it when it's done. 

I don't see how anyone could actually seriously have an issue with this. Having huge laggy stacks of food doesn't do anything positive for anyone.

Draycu:
I feel like this is a sensible thing to change.  As someone who tends to stick with botany and xenobotany, I’ve often used the bot to maintain the plants while I do other things.  

No matter how many times I’d try to make sure the bot was set to not collect produce, somehow it seemed it would be set back and would lag the area with massive stacks of harvested plants.

My only other suggestion would be to perhaps add a multitool that could spawn in Hydroponics for the cryo option on the tray.  However, that’s more a mapping request than anything to do with the code.

Definitely feel this is a helpful feature that could hopefully prevent the headache of cleaning up massive amounts of plants, and hopefully stop some of the lag such stacks caused after several hours!

LittleBigKid2000:
Make the bot plant seeds and deliver the harvests to the kitchen

Very-Soft:
why not just remove the botany slots while we're at it?

SpadesNeil:
stop posting shit

PR #2664, opened 2018-01-02
Fixes non-dogborgs butting into dogborg icon update territory.

### LINKED ISSUE #2625: Non-dog cyborgs turn invisible when resting.
Non-dog borgs turn invisible when resting.

Their eyes to turn off, and to not become invisible.

#### What actually happened
They became invisible.

#### Steps to reproduce
Be an actual, useful cyborg, press 'rest' or get stepped on by some giant person.

#### Code Revision
efed34b8f7f765e9bfd8cd6708e3d6f80a28764a
 
#2470 @Verkister

Verkister:
-No more invisible resting standard borgs.
-Also adds a dogborg var for special functions such as the icon bs.

Fixes #2625

### DIFF
--- a/code/modules/mob/living/silicon/robot/robot_modules/station_vr.dm
+++ b/code/modules/mob/living/silicon/robot/robot_modules/station_vr.dm
@@ -81,6 +81,7 @@
 	//R.icon_state = "k9"
 	R.pixel_x 	 = -16
 	R.old_x 	 = -16
+	R.dogborg = 1
 	..()
 
 /obj/item/weapon/robot_module/robot/knine/respawn_consumable(var/mob/living/silicon/robot/R, var/amount)
@@ -131,6 +132,7 @@
 	//R.icon_state = "medihound"
 	R.pixel_x 	 = -16
 	R.old_x  	 = -16
+	R.dogborg = 1
 	..()
 
 /obj/item/weapon/robot_module/robot/ert
@@ -157,6 +159,7 @@
 	R.hands.icon = 'icons/mob/screen1_robot_vr.dmi'
 	R.pixel_x 	 = -16
 	R.old_x 	 = -16
+	R.dogborg = 1
 	..()
 
 /obj/item/weapon/robot_module/robot/scrubpup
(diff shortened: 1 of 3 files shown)


code/modules/mob/living/silicon/robot/robot_modules/station_vr.dm
```diff
@@ -179,6 +182,7 @@
 	//R.icon_state = "scrubpup"
 	R.pixel_x 	 = -16
 	R.old_x 	 = -16
+	R.dogborg = 1
```
Arokha: Current standard is `TRUE` and `FALSE` for boolean stuff. Just make these `TRUE` and the other one `FALSE`

code/modules/mob/living/silicon/robot/robot_vr.dm
```diff
@@ -17,7 +18,7 @@
 
 /mob/living/silicon/robot/updateicon()
 	..()
-	if(icon == 'icons/mob/widerobot_vr.dmi' || 'icons/mob/64x64robot_vr.dmi')
+	if(dogborg == 1)
 		if(stat == CONSCIOUS)
 			if(sleeper_g == 1)
```
Arokha: No need to nest these.
`if(dogborg == TRUE && stat == conscious && sleeper_g == TRUE)`

code/modules/mob/living/silicon/robot/robot_vr.dm
```diff
@@ -17,24 +18,22 @@
...
-				overlays += "taser"
-			if(resting)
-				overlays.Cut() // Hide that gut for it has no ground sprite yo.
-				icon_state = "[module_sprites[icontype]]-rest"
-			else
-				icon_state = module_sprites[icontype]
-		else if(stat == DEAD)
-			icon_state = "[module_sprites[icontype]]-wreck"
-			overlays += "wreck-overlay"
+	if(dogborg == TRUE && stat == CONSCIOUS && sleeper_g == TRUE)
```
Arokha: Why this change? It was actually much more efficient, CPU-time-wise, before.
Previously it'd only check if they were conscious once and if they were a dogborg (based on the poor icon test, glad you fixed that) once.

Now it checks both 10 times for some reason?

PR #2666, opened 2018-01-02
Makes belly bulge examine add up prey sizes.

### LINKED ISSUE #2560: Belly examine size only counts for first prey and doesn't add up.
Belly examine size thing only counts for the first prey.

To have the belly show up in examine after eating a second prey bigger than the set minimum size after the smaller first prey.

#### What actually happened
Belly stayed invisible until the examine limit was set below the first prey's size.

#### Steps to reproduce
- (Step 1) Set belly examine size bigger than your first tiny prey.
- (Step 2) Eat the tiny prey.
- (Step 3) Have no visible bulge as supposed.
- (Step 4) Eat a bigger prey into the same belly.
- (Step 5) Still have no visible bulge.

#### Code Revision
- https://github.com/VOREStation/VOREStation/commit/d2194a759af92a64e9392232841f351bf15d409f

Verkister:
-If prey 1 is too small to show up, ya get no visible bulge.
-But if you eat another prey also too small to show up, and they take enough space together with the first one, ya get bulge.
-Also if you eat a small person, and then a big person, ya now get bulge too. (unlike before)

OwO

Fixes #2560

### DIFF
--- a/code/modules/vore/eating/belly_vr.dm
+++ b/code/modules/vore/eating/belly_vr.dm
@@ -192,15 +192,18 @@
 	if(internal_contents.len && examine_messages.len)
 		var/formatted_message
 		var/raw_message = pick(examine_messages)
+		var/total_bulge = 0
 
 		formatted_message = replacetext(raw_message,"%belly",lowertext(name))
 		formatted_message = replacetext(formatted_message,"%pred",owner)
 		formatted_message = replacetext(formatted_message,"%prey",english_list(internal_contents))
 		for(var/mob/living/P in internal_contents)
-			if(!P.absorbed && P.size_multiplier >= bulge_size && bulge_size != 0) //This is required first, in case there's a person absorbed and not absorbed in a stomach.
-				return("<span class='warning'>[formatted_message]</span><BR>")
-			else if(P.absorbed || P.size_multiplier < bulge_size || bulge_size == 0) //Are they absorbed, too small to show up, or examining this stomach is disabled?
-				return ""
+			if(!P.absorbed) //This is required first, in case there's a person absorbed and not absorbed in a stomach.
+				total_bulge += P.size_multiplier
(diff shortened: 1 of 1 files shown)


code/modules/vore/eating/belly_vr.dm
```diff
@@ -192,15 +192,18 @@
 	if(internal_contents.len && examine_messages.len)
 		var/formatted_message
 		var/raw_message = pick(examine_messages)
+		var/total_bulge = 0
```
Arokha: OwO

code/modules/vore/eating/belly_vr.dm
```diff
@@ -192,15 +192,18 @@
...
 		for(var/mob/living/P in internal_contents)
-			if(!P.absorbed && P.size_multiplier >= bulge_size && bulge_size != 0) //This is required first, in case there's a person absorbed and not absorbed in a stomach.
-				return("<span class='warning'>[formatted_message]</span><BR>")
-			else if(P.absorbed || P.size_multiplier < bulge_size || bulge_size == 0) //Are they absorbed, too small to show up, or examining this stomach is disabled?
-				return ""
+			if(!P.absorbed) //This is required first, in case there's a person absorbed and not absorbed in a stomach.
+				total_bulge += P.size_multiplier
+		if(total_bulge >= bulge_size && bulge_size != 0)
+			return("<span class='warning'>[formatted_message]</span><BR>")
+		else if(total_bulge < bulge_size || bulge_size == 0) //Are they absorbed, too small to show up, or examining this stomach is disabled?
```
Arokha: You don't actually need the `if` in this. Just do `else` and return the empty string, since the only way to get to the `else` is with the conditions you want right there. There's no point in testing for them since we already know either total_bulge is less than bulge_size or bulge_size is 0

PR #2670, opened 2018-01-03
Adds advanced medical items for centcom/ERT

Arokha:
They should probably have these because they're fancy.


Chargae:
SOOOO FANCY.

PR #2671, opened 2018-01-03
Electrostaff Custom Item

blazerules:
This is the code for the stunstaff as was approved here

https://forum.vore-station.net/viewtopic.php?f=27&t=1204


code/modules/vore/fluffstuff/custom_items_vr.dm
```diff
@@ -1668,3 +1668,136 @@
 		return
 	else
 		..()
+
+//jacknoir413:Areax Third
+/obj/item/weapon/melee/baton/stunstaff
```
Arokha: Please give this a `/fluff/` in the path. So probably `/obj/items/weapons/melee/baton/fluff/stunstaff`. Find/replace should do the trick.

code/modules/vore/fluffstuff/custom_items_vr.dm
```diff
@@ -1668,3 +1668,136 @@
...
+/obj/item/weapon/melee/baton/stunstaff/dropped()
+	..()
+	if(wielded)
+		wielded = 0
+		spawn(0)
+			update_held_icon()
+
+/obj/item/weapon/melee/baton/stunstaff/apply_hit_effect(mob/living/target, mob/living/user, var/hit_zone)
+	if(isrobot(target))
+		return ..()
```
Arokha: Why duplicate `apply_hit_effect`? What's different from the normal baton?

code/modules/vore/fluffstuff/custom_items_vr.dm
blazerules: I honestly have no idea if the item will even work if I do that. I guess I could just copy paste all the baton code but that kinda begs the question of why bother with having baton in the filepath. I mean if it works cool.

code/modules/vore/fluffstuff/custom_items_vr.dm
blazerules: Probably wanted to change something there but then didn't need to anymore and forgot to remove it.

code/modules/vore/fluffstuff/custom_items_vr.dm
Arokha: It will work exactly the same.

If I have an item that's `/obj/item/thingy` and define another item named `/obj/item/thingy/this/is/a/really/long/path/but/it/exists` and change like... `name = "Crazy Item"`, it will work, look, and otherwise be exactly like `/obj/item/thingy` except for the name.

Arokha: Just to try and further explain the path inheritance for your reference, you could do...
```
/obj/item/thingy
name = "A thingy"
desc = "This is a thingy"
icon = 'the_original.dmi'

/obj/item/thingy/special
name = "A special thingy"

/obj/item/thingy/special/very_special
icon = 'glowy_thingy.dmi'
```
`/obj/item/thingy` will have what you'd expect. `/obj/item/thingy/special` will have the same desc and icon as thingy, but a new name. And `/obj/item/thingy/special/very_special` will have a special icon, AND also have the name from `/obj/item/thingy/special` and the desc from `/obj/item/thingy`.

Basically just think of it as saying "Oh, you didn't define anything special for this variable/proc/whatever on your sub-type of that path? You must want the original from the parent(s)."

blazerules:
I have no idea why the check failed. All I did was rename some stuff.

config/custom_items.txt
```diff
@@ -890,3 +890,9 @@ ckey: zodiacshadow
 character_name: Nehi Maximus
 item_path: /obj/item/device/radio/headset/fluff/zodiacshadow
 }
+
+{
+ckey: jacknoir413
+character_name: Areax Third
+item_path: /obj/item/weapon/storage/box/fluff/stunstaff
```
Screemonster: The reason given for accepting the application is that it's functionally identical to a baton, so it should probably be restricted to jobs that are normally allowed a baton.

blazerules: Done, slipped my mind

PR #2676, opened 2018-01-05
Fixes some bugs relating to absorption and struggling, makes struggle feedback messages clearer

### LINKED ISSUE #1998: Resisting mob stomachs doesn't work
Resisting from a mob gut should get you spat up.

See above.

#### What actually happened
"You struggle inside the (X)'s stomach..." And no other text.

#### Steps to reproduce
- Get eaten by a mob.
- Resist until you get something other than making its stomach rile up or failing to escape.
- Fail to escape anyway.

#### Code Revision
- https://github.com/VOREStation/VOREStation/commit/78f2ee4939d7293ace6d2b3698bae30c8c3a2a53

Screemonster:
- Fixes a bug where absorbed prey in simplemobs wouldn't get released when the predator was killed, which in some cases resulted in prey getting GC'd.
- Fixes an indentation error in the messages given to prey when an otherwise successful escape attempt is blocked by the predator setting their belly to inescapable before the escape timer is up.
- Fixes an edge case where prey could roll a successful escape attempt, get absorbed in the meantime, and then somehow manage to escape anyway.
- Makes struggle messages a little clearer as to the outcome, hopefully fixes #1998 as it's obvious when you're actually successfully resisting now. (previously, it displayed the "you attempt to climb out" message in all cases, then gave _no_ further message if your attempt was successful)
- Random mode-switches will no longer tell you they're changing mode if they're already in the mode that is being switched to.
- Changed "switching to absorb mode" messages so you can tell the difference between a belly that just started absorbing and one that just started digesting.

### DIFF
--- a/code/modules/mob/living/simple_animal/simple_animal_vr.dm
+++ b/code/modules/mob/living/simple_animal/simple_animal_vr.dm
@@ -36,6 +36,8 @@
 /mob/living/simple_animal/Destroy()
 	for(var/I in vore_organs)
 		var/datum/belly/B = vore_organs[I]
+		for(var/mob/living/M in B.internal_contents) // because release_all_contents doesn't release absorbed things
+			M.absorbed = 0
 		B.release_all_contents() // When your stomach is empty
 	prey_excludes.Cut()
 	. = ..()
@@ -127,6 +129,8 @@
 /mob/living/simple_animal/death()
 	for(var/I in vore_organs)
 		var/datum/belly/B = vore_organs[I]
+		for(var/mob/living/M in B.internal_contents) // because release_all_contents doesn't release absorbed things
+			M.absorbed = 0
 		B.release_all_contents() // When your stomach is empty
 	..() // then you have my permission to die.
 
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/simple_animal/simple_animal_vr.dm
```diff
@@ -127,6 +129,8 @@
 /mob/living/simple_animal/death()
 	for(var/I in vore_organs)
 		var/datum/belly/B = vore_organs[I]
+		for(var/mob/living/M in B.internal_contents) // because release_all_contents doesn't release absorbed things
+			M.absorbed = 0
```
Arokha: I like all the changes except this one. I know you told me about this one and asked if that was the way to do it in Discord but sometimes it's hard to get context like that.

I think a better way to do it would be to add a parameter to release_all_contents, like
```
/datum/belly/proc/release_all_contents(var/include_absorbed = FALSE)
	if (internal_contents.len == 0)
		return 0
	for (var/atom/movable/M in internal_contents)
		if(!include_absorbed && istype(M,/mob/living))
			var/mob/living/ML = M
			if(ML.absorbed)
				continue
```
Then just call `B.release_all_contents(include_absorbed = TRUE)`. It will be faster.

Screemonster: Fixed as per discord convo to avoid problems of people being released with absorbed still true and stuck with weldervision forever

PR #2681, opened 2018-01-05
Gives the HoS a KHI-102b NSFW

Arokha:
It's just my fluff gun. But sharing is caring.

From the ingame fluff:
"Variety is the spice of life! The 'Nanotech Selectable-Fire Weapon' is an unholy hybrid of an ammo-driven energy weapon that allows the user to mix and match their own fire modes. Up to three combinations of energy beams can be configured at once. Ammo not included."
"This gun is an energy weapon that uses interchangable microbatteries in a magazine. Each battery is a different beam type, and up to three can be loaded in the magazine. Each battery usually provides four discharges of that beam type, and multiple from the same type may be loaded to increase the number of shots for that type."
"The Kitsuhana 'Nanotech Selectable Fire Weapon' allows one to customize their loadout in the field, or before deploying, to achieve various results in a weapon they are already familiar with wielding."

The HoS's includes 2 lethal, 2 stun, 1 net, and 1 ion battery. There are other batteries (like instagib!) but there's no way to get them normally. There's also one that strips people. 👀 

image

image

image

### DIFF
--- a/code/game/objects/structures/crates_lockers/closets/secure/security.dm
+++ b/code/game/objects/structures/crates_lockers/closets/secure/security.dm
@@ -147,8 +147,9 @@
 		new /obj/item/ammo_magazine/m44/rubber(src)
 		new /obj/item/ammo_magazine/m44(src)
 		new /obj/item/ammo_magazine/m44(src)*/ // NO YOU DO NOT GET A .44 MAGNUM! -Ace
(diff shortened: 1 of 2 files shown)


Arokha:
Wanna give better info when examining about what ammo is loaded, so lemme add that.

PR #2685, opened 2018-01-05
Tweak lethal succubus drain

Arokha:
Mostly remove the OOC magic-implying wording about 'life energy' (since setting is w/o magic) so it can be whatever in your private RPs (someone was asking about using it to represent stealing bodyheat which is cool).
`absorbing_prey` was never unset if they were found to be indigestible, so fixed that
Two stages also sent two messages at the same time
Removed the needless casting of src to C since it's defined on human already

Maybe in the future consider using cloneloss instead of brainloss? Alkysine is one of the easiest drugs to make, and if I jack myself up with it I'd maybe be immune to having my nutrition/heat/whatever stolen which seems \~unusual\~

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/station_special_abilities_vr.dm
+++ b/code/modules/mob/living/carbon/human/species/station/station_special_abilities_vr.dm
@@ -511,13 +511,12 @@
 			return
 
 /mob/living/carbon/human/proc/succubus_drain_lethal()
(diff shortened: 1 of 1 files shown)


Screemonster:
cloneloss would be neat (since it's what slimes do) but a few species on-station are outright immune to cloneloss. Anyone with the NO_SCAN flag essentially.

Arokha:
> few species on-station are outright immune to cloneloss

That's true. Maybe it'd be better off as just a timer until death or something.

PR #2689, opened 2018-01-05
Sizegun QoL/Nerf

Arokha:
* I want clicking it in-hand to let me change the size (so I can do it with Z). And I want the box to show the currently set size when I'm changing it instead of always just showing 200. So that's what it does now.
* Also jeez. If we have the technology to make a military-grade battery recharge in just a few seconds, why don't we use it other places? Makes the thing not self-recharge. Just stuff it in a recharger. There are plenty.
* Added various rechargers everywhere in public places on the map, since there weren't enough, just to be sure.

(The odd diff is from rearranging. Previously it was like...
gun
bullet
gun
gun
gun)


Arokha:
Somehow I 'lost' this other change that actually did what the desc said.

PR #2690, opened 2018-01-05
Commlink QoL

Arokha:
Makes it properly auto-register your name instead of coming as Unset every time.
Makes messages to Commlinks properly stay in the commlink user's head.

### DIFF
--- a/code/modules/nifsoft/software/14_commlink.dm
+++ b/code/modules/nifsoft/software/14_commlink.dm
@@ -42,7 +42,7 @@
 		..()
 		nif = newloc
 		nifsoft = soft
-		register_device(nif.human)
+		spawn(10) register_device(nif.human) //Need to outwait the other spawn in communicators, sigh.
 		qdel_null(camera) //Not supported on internal one.
 
 	Destroy()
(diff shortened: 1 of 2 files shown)


code/modules/nifsoft/software/14_commlink.dm
```diff
@@ -42,7 +42,7 @@
 		..()
 		nif = newloc
 		nifsoft = soft
-		register_device(nif.human)
+		spawn(10) register_device(nif.human) //Need to outwait the other spawn in communicators, sigh.
```
Leshana: Acceptable for now (until `/obj/item/device/communicator/New()` is fixed)  but both that spawn and this spawn should be in initialize() not New()

Arokha: I assume fixing the other New() is contingent upon like. Fixing the order things happen during character spawn which might be "a while", but no I don't like the DOUBLE RACE either.

Very-Soft:
I've noticed that I still seem to have to set my name in the Commlink. Playing as a custom species, if that's important at all.

PR #2691, opened 2018-01-06
Stop doors from sleeping during machine controller process() execution

Leshana:
* Sleeping during process() is BAD BAD BAD!
* Since close() sleeps, we make sure to spawn() before calling it.

### DIFF
--- a/code/game/machinery/doors/airlock_control.dm
+++ b/code/game/machinery/doors/airlock_control.dm
@@ -21,8 +21,7 @@ obj/machinery/door/airlock/receive_signal(datum/signal/signal)
 	if(id_tag != signal.data["tag"] || !signal.data["command"]) return
 
 	cur_command = signal.data["command"]
-	spawn()
-		execute_current_command()
+	execute_current_command()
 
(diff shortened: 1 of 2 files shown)


Leshana:
If it works it should go to polarisland also.

Arokha:
I see spawns but I don't see any sleeps removed/changed?

Leshana:
Right, it still needs to sleep, but now the sleeps won't block the machine controller because in the process() proc it spawn()'s before calling close()

Neerti:
I think you can also do ``set waitfor = 0``

Leshana:
You can indeed.   In fact I actually tried that first before adding spawns, and it seemed to work.
I was concerned about this here: https://github.com/VOREStation/VOREStation/blob/080a9234d610fab4bd50448815c4e5f8c946987f/code/game/machinery/doors/airlock_control.dm#L62   where close() is called followed by lock(), and so I was concerned about what would happen if close() returned then lock() was called before close() was completed.
This is a priority fix, so after more research may change it to `waitfor = 0` if it turns out to be properly safe.

PR #2694, opened 2018-01-06
Redoes the armory

Arokha:
Space Obama here, to take all the guns away. This removes the entire secure side of the armory and replaces it with a shuttle with guns and heavy armor that NT can send if the station is at code red. It's in the same place, but has fewer guns oh noooo. The console to call it is in the not-as-gunny side of the armory, but still has access to both sides (from the breakroom and from the other side of the armory).

image


SpadesNeil:
Kind of a weirdly complicated thing to do when you just have to call the shuttle to get the weapons for the armory.

Doesn't make sense in-character and also completely removes first responder capability.

I do not approve of this.

Dhaeleena:
I kinda like the idea but i can see Ace's point

Mewchild:
The secure side of the armory isn't only used in code red. As an example, it wouldn't be unreasonable for security to be given energy weapons in the event of a slime outbreak or something similar. Considering as well that that side of the armory contains the riot gear (shields, armor, etc), I feel this is a wholly restrictive change that has negative gameplay implications. The only potential upside I can see of this is that it would encourage more crosstalk with research, but only in the context of "Please help us rebuild our neutered armory."

HzDonut:
I don't really see the point of removing the big guns at all, we're not the kind of server that has to worry about people breaking in and using all that against us or anything, and if that gear got handed out willy nilly that person would be demoted in a heartbeat. If anything, this just causes problems since when we need that gear, we NEED that gear.

maps/tether/tether_shuttles.dm
```diff
@@ -20,6 +20,20 @@
...
+	shuttle_tag = "Danger Armory"
+	req_one_access = list(access_heads)
+	//needs code to stop use below code red
+	//!get_security_level() == "red"
+
+/obj/machinery/computer/shuttle_control/danger_armory_shuttle/ui_interact(mob/user, ui_key = "main", var/datum/nanoui/ui = null, var/force_open = 1)
+	if(get_security_level() == "red")
+		return ..()
+	else
+		to_chat(usr,"<span class='warning'>The screen remains static. This console only works on Alert Level Red or higher.</span>")
```
Cyantime: Doesn't actually work on alerts higher than red, the check is only if the alert is specifically red.

Benemuel:
To me it seems like trying to fix a broken leg with a amputation. Yes, it TECHNICALLY works, but it seems unneeded compared to just fixing the issue, which tends to be people in sec jumping to lethals first thing, as opposed to trying to negotiate. All this would do is give them a extra step to do that.

LiquidFirefly:
Ace, when do you need guns as security?   Like, really, really need guns.   Because the only time I can think of that you’re dealing with something that requires lethals is an event.  In the case of an event, if the event requires lethals, lethals can be spawned by the admin overseeing the event via fax.   There’s no reason they require those guns.

SpadesNeil:
So you would prefer players wait 10 to 20 to 30 minutes to get their lethals for an event.

How about no. Security already has to go through the warden at minimum to get lethal weapons. If the warden is handing them out like candy then they should not be Warden. There are numerous other ways to fix this incredibly rare problem then something so bombastic and impractical as this.

SpadesNeil:
And yes I'm acutely aware that an admin may still be doing the event actively and although response time on fax can be quick it does not take away from the lengthy time it requires to complete the process. We're also not just going to magically teleport the guns there because that's stupid.

Screemonster:
The other issues with the current armory is that it's frankly overstocked, leaving virtually no reason for sec to ever order anything from cargo or science since everything they could possibly need has been powercrept into starting gear. Most shifts, the biggest thing they're liable to have to deal with is grubs and the occasional wandering surface mob, so as long as the standard armory has _some_ means of dealing with those things (and for grubs, an egun is _more than adequate), some sort of barrier to just busting out the carbines and snipers makes sense.

That said, if it's gonna take any longer than 2 minutes then this armory might as well not be on the map at all since cargo is right down the hall and that's how long the shuttle takes.

Rather than having it done as a sort-of-shuttle, how about keeping the "code red" armory like the vault in that the doors are bolted at roundstart, with the access controls to unbolt them tied to the alert level in some way - even if it's a warning along the lines of "the tactical armory is for emergencies only. Use of this override outside of CODE RED will result in a notification to Central Command. Improper use may result in charges of exceeding official authority."

SpadesNeil:
I would be fine with a tactical Armory. We had that on the old map and honestly I don't know why we got rid of it. I do agree that the current Armory is certainly overstocked.

Chargae:
You don't even need an egun for grubs, as a note. Tazers work.

maps/tether/tether_shuttles.dm
Arokha: Delta isn't actually achievable on this server anyway.

SpadesNeil: That isn't true. Delta can be achieved through the use of admin tools or if we were to spawn and activate a nuke. There has been at least one incident where we have nuked the station.

Arokha:
You added energy snipers to the armory. It has 15 suits in it, too. The energy snipers have the same damage as a pulse rifle.

If you give security that stuff with no restrictions, how can you ever hope to make any sort of antags that have any chance of not just getting shot from off the screen? Every event with armed antags we've had has gone horribly. It's time to make changes. We already discussed this as admins a long time ago, but nobody implemented it, then we discussed it again last night.

Also the shuttle takes 60 seconds, I'm unsure where this "30 minute wait" is coming from.

Arokha:
I have an idea: Why don't we TRY this and see if the results are as apocalyptic as the naysayers predict?

Mewchild:
The most prevalent example I can think of where weaponry is needed is the case of slime outbreaks, which do happen a decent amount. Tasers don't affect them, and I feel it's unreasonable to ask officers to go into melee combat on an outbreak using extinguishers. In addition to this, there are the somewhat rare (but still present) occasions where a crew-member is eaten by a gaslamp, requiring security response, or the much more common examples where security is called due to a report of someone with a weapon, necessitating responding officers be armed for the potential of engagement.

Arokha:
@Mewchild This is why normal eguns moved into the armory on the left, so you can have lasers for times when you need lasers to pew at things like that.

Mewchild:
There are examples of lethal weaponry being authorized and required outside of code red. Even in code blue, lethals are able to, [by SoP](https://wiki.vore-station.net/Standard_Operating_Procedure#Code_Blue_-_Suspected_Threat), be issued by the Warden, HoS, or Colony Director. Code Blue can be called for a number of situations which wouldn't mandate a full station-wide code red, with all it entails. Examples of this, that come to mind, are when a poorly hidden Rape/HV scene is discovered, or other evidence has come to light that there may be someone armed/dangerous aboard the station. Code blue calls for the permitting of random searches of crew- something that may be met with violence but is ultimately within the job requirements of officers to perform.

Arokha:
@Mewchild This is why normal eguns moved into the armory on the left, so you can have lasers for times when you need lasers to pew at things like that.

Mewchild:
My concern boils down, mainly, to the biggest threat the station regularly faces, that again being research. Xenobiological outbreaks happen roughly once a week or so, and oft warrant heavily armed response. From experience, and keeping in mind this is mere anecdote, the average slime can take 2-3 shots to down from an egun, much more if there's a pink (healing slime? Can't remember color). Couple that with miss chance, and the officer is only able to down 1-2 before needing to retreat, to spend 1-2 minutes recharging. I'd push for a downsizing of the armory, but removing it completely from the station seems an extreme measure to the problem security has been facing.

JoanRisu:
Those are examples of smaller emergencies meant for a lighter armory, one that would hold things like riot gear and ion rifles.

As it stands, the current armory makes a base meant for research look more like a military outpost. It's overkill for things that are not preludes to terrorism, invasions, or full blown outbreaks.

Arokha:
> spend 1-2 minutes recharging

Sounds like someone made the poor choice to carry literally nothing else with a battery. Armory isn't going to fix that no matter how it's designed.

SpadesNeil:
The singular energy sniper was added in direct response to more than one hostage situation that was not produced by admin events. However the SMGs should really just be removed as they have no practical purpose that the laser carbine cannot already fulfill.

SpadesNeil:
Also the ion rifles are there because of people who like to run around in heavily-armed mechs.

SpadesNeil:
I wish I could edit posts on my phone. Just so you know these changes were implemented a very long time ago and they were indeed discussed before hand.

SpadesNeil:
Also why did you remove my tag?

Arokha:
> singular energy sniper

There were two.

> the ion rifles are there because of people who like to run around in heavily-armed mechs

Rogue combat mech would be code red I'd think? Immediate threat to station.

> Also why did you remove my tag?

Because like I said, we DID discuss this already. Both months ago when nobody decided to actually do the map work involved, and last night. The armory has gotten MORE guns since our decision last year to remove some, so obviously things are not going well with nobody actually implementing that decision.

Screemonster:
The has been an issue lately when it comes to events of any kind for security to wildly gear up without any code changes, I don't know if this is people coming over from action servers and treating this place like the places they used to play or what.

A mob arrives and you don't know whether it's hostile? Better grab the tactical armory just in case. No need to declare code blue, that'll just panic the crew. (and get them yelling at you for declaring code blue/red when nobody's getting shot yet)

At the very _least_ I'd like some sort of confirmation box to get at them, like "I acknowledge that busting this gear out WILL escalate the situation if it isn't already violent" so we can see if there's any pattern to it, 'cause it's getting to the point where it's barely worth running events at all if SOP is going to be for sec to charge down and shoot whatever it is immediately, and I don't want to have to deal with the shitstorm of whining that will result if I start HoSbanning people who grab the riot guns in code green and yell BUT WE HAVE ACCESS TO THEM when told to chill the fuck out on the murderhoboing.

As another consideration - unlike box and northern star, security on tether is RIGHT NEXT to cargo, who are more than capable of ordering a few ablative crates if for some reason security are wanting to gear up more people than they can ever have roundstart officers.

HzDonut:
All this will change is that instead of security instantly gearing up for something, they'll just call the shuttle and do the same thing. Putting the armory behind a timer won't solve the problem, the players are the problem. I just can't see the positive results to this, but I guess a testing phase won't hurt anyone.

Screemonster:
It's not so much putting it behind a timer that I'm in favour of as it is saying "central (read: admins) won't be happy if you routinely bust out the tactical armory without a halfway decent reason for doing so" and notifying us when it happens.

That and bringing the overstocking under control a bit. Bug cargo if you need to gear up more than half a dozen officers for whatever reason.

Benemuel:
Honestly, if it makes you NEED cargo, thats fine. If it makes cargo the way better option then your own armoury, that is a issue.

Arokha:
The large part of this is thinning down the guns in the armory in general, to me, anyway. The fact that it's on a shuttle was just a fun play on Paradise's bluespace armory. To be honest, we were also going to lock down what guns can be ordered from cargo, as well, behind code locks and stuff. It's just not part of this PR.

The idea is more 'deescalation' from where we are now which is "they have every gun from the start and access to insane weapons". We'd rather give both antags and security "less firepower" so shooting events aren't "you got shot from offscreen in the head and because it's the same damage as a pulse destroyer you're killed instantly".

Screemonster:
Bear in mind that when I mentioned the code locks (on the military crates and the like - remember the "EMERGENCIES ONLY" crate with the LMG that was routinely ordered for science every round?) I also mentioned bypassing that restriction by hacking the console since you're ordering from shady black marketeers at that point instead of sending your order to central.

Though I would like a downside to roundstart hacking the ordering console to go with this, such as increased prices compared to normal, or a risk of getting sent dodgy goods, but that's outside the scope of this PR.

SpadesNeil:
>There were two (snipers).

Okay then remove one. There's really only a need for one.

>Rogue combat mech would be code red I'd think? Immediate threat to station.

So what do they do for the 10-20 minutes that the mech is roaming around doing god knows what while they wait for the heads of staff to set code red, fax CentCom, CentCom to reply, CentCom to send the shuttle...?

>Because like I said, we DID discuss this already. Both months ago when nobody decided to actually do the map work involved, and last night.

I am aware of no such discussion and we're having the discussion now, so...

>The armory has gotten MORE guns since our decision last year to remove some, so obviously things are not going well with nobody actually implementing that decision.

This decision doesn't exist as far as I am aware. If it exists, I have no idea what the community input on it is, because so far, the community response in this thread is negative outside of admins. However it's just three people.

Do not remove that tag in future. We talked about not making big changes without discussing them and if we haven't discussed this in a year and never implemented those changes then clearly the last discussion is out of date and possibly for a good reason.

SpadesNeil:
Also why are we immediately doing the most dramatic change of **removing the entire armory** instead of the logical path of doing something far more reasonable and realistic like just _taking away some of the gear that isn't needed?_

No one has bothered to audit what's in the armory. Even when we first changed to the new map, we have had **substantially more stuff** in the armory, so don't blame me for these changes.

Arokha:
I'm just going to redo the entire armory area, and fine no shuttle. Maybe even be tiered for each alert level, who knows.

SpadesNeil:
Tiered for alert level would be an excellent alternative solution. Plus we can tie admin alerts to it if we make the unlock button do special things.

LiquidFirefly:
I mean, you say that like it has to take 10-15 minutes to say "Hey, this and this is going down, we need guns," and then maybe a short time for an armory shuttle to arrive.    The cargo shuttles take less than a few minutes to go back and forth.  And again, I'm saying the warden is a part of security.   They might not be handing them out like candy, but there are certainly situations where security does not need weapons for an event, or jumps the gun because the HoS and the Warden can just immediately, no questions asked get in there.   We really don't have combat events here, like, ever.   And if we do it's not an immediate overtaking.  I don't see how the span of less than a few minutes from request to drop is a massive deal.

Issue #1130, opened 2017-03-08
Using a newspaper on nonhostile mobs should make them spit out prey


Screemonster:
#### Brief description of the issue
No.. no! Runtime! BAD! GET BACK HERE! That's a patient!

*WHAP*

Right, now, OUT.

#### What you want to happen
Whapping pets with a newspaper should "encourage" them to spit out anyone they've ate.

#### What actually happened
It doesn't.

SpadesNeil:
That's fucking hilarious.

ArbyTheDragon:
I feel as though there are several items that should do this. A space cleaner bottle should work too.

Very-Soft:
Yes please! XD

Arokha:
Made a little harder by the fact that Runtime and Renault /are/ hostile mobs.

PR #2703, opened 2018-01-07
Returns Noodle to the CE office

Cyantime:


Screemonster:
Commenting here as a reminder to include noodle when I work on #1130
except snek so I'm sorely tempted to say a rolling pin should be used to force prey out

Cyantime:
You were going to do that too? It was on my to_do.txt, but in the form of `Pick up cat & squeeze to remove contents`

PR #2706, opened 2018-01-08
Improves medibot speaker switch

Cyantime:
The speaker control on medibots now controls all audio messages, instead of only idle ones.


Cyantime:
wait this was supposed to go to polaris, ignore me

Arokha:
I USUALLY DO

PR #2711, opened 2018-01-08
Advanced NIF Repair Tool

Arokha:
Creatable at the protolathe. Requires the same tech levels as the NIF, ~but no exotic materials~.

It's a device that accepts nanopaste, and programs the nanite swarms into NIF repair swarms, which are a liquid reagent ingame. It has a 60u storage capacity, and needs to be emptied into any container.

From there, it can be injected via syringe, made into pills, or drank (though the latter two reduce effectiveness by 50%). Note that synths don't process reagents: this is only an option for organics.

Each unit nanopaste makes 15u repair nanites, so a full repair from 0% to 100% of a normal NIF requires 7 nanopaste converted to reagents (100u).


HzDonut:
The last thing we need is science (specifically roboticists, since we're the 'NIF repair people') having even less to do because people will just request these repair nanites instead, plus we already have a problem with non-scientists repairing NIFs, it shouldn't be made _easier_ for them to do so. All I ever hear about is people wanting more 'inter-departmental interaction' yet so many updates come through that limit this interaction, it's bizarre.

Arokha:
I'm unsure where you got the idea that robotics was "the NIF repair department" other than perhaps that it's sort of like a machine shop and that's where some spare cable coils are. I've always thought of Science in general as being responsible.

Regardless, I don't understand how this alters that. R&D can make the repair tool and give it to robotics to manage. People can come to robotics and get drinks/pills/shots. Administering nanotech stuff to repair physical hardware is not really medical's domain.

Screemonster:
The people coming in to get their NIF repaired are not going to be doing so on the same shift where they got the NIF, it's not like we're spawning one of these in the robotics lab. It still has a fairly hefty research gate.

HzDonut:
The machine existing isn't the point I'm protesting here at all, I think it's great personally. I never said I don't like the _machine_. I don't like how it enables people with no knowledge on how to repair NIFs to now be able to repair NIFs, which we just agreed shouldn't be possible since it's more complicated IC than just stabbing a syringe into someone or yourself, so problem solved.

Natje0:
I'm honestly not a big fan of this.
NIF's are a neat thing that for once made surgery actually wanted at times. Unlike organs, and even bones if medical get their hands on carpotoxin, it was one thing that couldn't be fixed with an injection. Fixing a NIF gave a surgeon, and usually a member of Research actual WORK to do.

This removes a large part of the already scarce amount of work Surgeons have.

Arokha:
Added rare materials requirements so it's more on par with a NIF itself, so it doesn't replace surgery as readily.

KonBear:
Oh I see Jesse's point now! I totally agree with that, especially after a netflix sci-fi show called "Travellers" and how their concept works with nanomachines in the human body. Basically the idea is that the nano machines are very hard to come by in the first place making actual doctors and surgery more viable than what the nano-machines would do for their cost, which I think you replicated well by making sure RnD has the levels and increased the cost of the little buggers. Another issue with the nano-bots was that they required to be programmed by a programmer in the show, needed to be told what to do and how to repair the specific body. Maybe coding in some sort of mini-game type thing so that the robotocists have to program the bots in a sense? Have it ID locked like some of the borgs when trying to do upgrades on them?

KonBear:
Of course we have our own lore, I just like their ideas and concepts of nano-bots. Making it not a thing that anyone can just inject into someone and "hurray, you did it, person who only knew to stab someone and press the button on the back of the needle"

Arokha:
Blerrrr. I dunno. I can't code a minigame, but I see the sides of the argument you have here yes.

KonBear:
Yeah a mini-game would be a lot for this, unless someone has a sudden inspiration to make one. So I'll just shoot some ideas, keep in mind I have very bare bone basic knowledge when it comes to coding. Perhaps a tool in robotics that can be used after an injection to activate the nanites, or like I said before with their PDA, if say aimed at their head while laying down with help intent on and the correct cartridge in their PDA for it?

Arokha:
PDA cartridges are a horrible horrible mess so not a PDA cartridge. The problem is that medical people think this is going to remove their surgeries, which they enjoy, and robotics people think this is going to remove their NIF repairs, which they enjoy. :v I don't know which is right, or which to care about more.

Draycu:
I would say the added high material cost of this proposed NIF repair tool would be more balanced.  It could be an _option_ rather than hoping that people with robotics knowledge and surgery skills can work together to repair the NIF.  

At the moment, it is a two person job, as the repair process of the NIF itself requires a multitool and wires, not something the Surgeon should necessarily be able to do on their own.

Overall, with the high cost, I would see it would still be better to rely on surgery for multiple people, but if there is only one person requesting repairs on a shift without a surgeon, this could be a valid option over faxing central for a single job.

PR #2712, opened 2018-01-09
Added supply shuttle export reports

AbbieFland:
Added a new option to the cargo console to view supply points gained
from the most recent shuttle trip.

image

It breaks down the value per crate, but to preserve the fun of trial and
error for folks that enjoy that doesn't itemize exactly what inside the
crate provided that value.

Part of the supply shuttle changes could probably be instead done as a
sell_crate hook, but because of where that's called it would have to
duplicate all the manifest-detection code.


AbbieFland:
Amended to fix a typo in the HTML

SpadesNeil:
This is nifty.

Very-Soft:
Yessss~

PR #2719, opened 2018-01-09
Make laying people be under other people

Arokha:
Unsure what consequences this would have. PLEASE REVIEW. (It works fine ingame, but I can't test every object with every mob)

### DIFF
--- a/code/modules/mob/living/carbon/human/update_icons.dm
+++ b/code/modules/mob/living/carbon/human/update_icons.dm
@@ -169,11 +169,13 @@ Please contact me on #coderbus IRC. ~Carn x
 		M.Scale(size_multiplier) //VOREStation Edit. Look at Polaris pull #4267 to see things edited.
 		M.Translate(1,-6)
 		src.transform = M
+		src.layer = MOB_LAYER -0.1 //VOREStation Edit. Laying people under other people. LEWD.
 	else
 		var/matrix/M = matrix()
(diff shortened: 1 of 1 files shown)


Screemonster:
the only downside I can see for this would be for LAPSITTING.

Arokha:
Generally lapsitting people are both standing. This only affects when sprites are 90deg rotated flat.

Screemonster:
So basically only an issue when someone does the cat thing of flopping in someone's lap, lying belly-up and demanding pets :v
(which even I don't do all that often)

Arokha:
Tru. But I think that is less common than every other time. Like just walking over someone who's laying, or using disarm-pin to pin someone, or just being on a bed over someone, etc.

Very-Soft:
I'm honestly fine with how it presently is? All it takes is the person on bottom moving off the tile then back on to get on top. I don't really understand why you would want to make it so that people wouldn't have at least that minor control over it.

Arokha:
That is not how it works.

Very-Soft:
Ah as I said in discord, I'd not seen it actually behave this way on server. if that's how it do, then yeh, that should not be a thing. I'd prefer if it behaved like all the other things, where the person who last entered the tile is on top.

Arokha:
If it breaks the universe you can yell at me.

SpadesNeil:
I have **never** seen it behave like that. The fuck got changed to break this?

PR #2722, opened 2018-01-10
Fixes dogborg sprites.

Jakksergal:
Fixes an issue with a previous PR overwriting a recent update.


Jakksergal:
This was actually made redundant by #2727 They included the fixed sprites in that PR

PR #2723, opened 2018-01-10
Gives sci dogborg a sheet loader like the sciborg

Cameron653:
Because they can't produce items for the research gripper without loading the protolathe.


Draycu:
Having worked on R&D myself, I can say the regular science borg has the material sheet loader already usable which allows the metal and glass sheets to be placed into the protolathe to create the first few items to be deconstructed.  This is a rather key component for basic research levels at the early start of a round, and cannot be progressed very far without mining's help.  

With the science dog borg being unable to load the protolathe, there is little the recently added module is able to do in the department, limited as they are at the moment.

PR #2725, opened 2018-01-10
Refactor Digestion

Arokha:
Refactors digestion to be about 220 lines less. Now items have a `digest_act(var/belly)` instead of having snowflake code repeated several times.

This also makes it possible for each item to easily have special things that happens when digested. Refactored a little of contamination as well.

Also in case it's important, removed the 'digestable' person check on digesting people's items. Not having your PDA digested is not an ERP pref. And in fact lots of people have the pref for the opposite (having their clothes/other stuff digested off).

NB: Also removed the NIF flavortext, it doesn't belong in the *vore* panel. Left the vars on humans so it can be readded to be settings on the NIF in another PR.

### DIFF
--- a/code/modules/vore/eating/belly_vr.dm
+++ b/code/modules/vore/eating/belly_vr.dm
@@ -37,7 +37,7 @@
 	var/tmp/mob/living/owner					// The mob whose belly this is.
 	var/tmp/list/internal_contents = list()		// People/Things you've eaten into this belly!
 	var/tmp/is_full								// Flag for if digested remeans are present. (for disposal messages)
-	var/tmp/emotePend = 0						// If there's already a spawned thing counting for the next emote
+	var/tmp/emotePend = FALSE						// If there's already a spawned thing counting for the next emote
 	var/tmp/list/items_preserved = list()		// Stuff that wont digest.
 	var/tmp/list/checked_slots = list()			// Checked gear slots for strip digest.
 	var/list/slots = list(slot_back,slot_handcuffed,slot_l_store,slot_r_store,slot_wear_mask,slot_l_hand,slot_r_hand,slot_wear_id,slot_glasses,slot_gloves,slot_head,slot_shoes,slot_belt,slot_wear_suit,slot_w_uniform,slot_s_store,slot_l_ear,slot_r_ear)
@@ -95,7 +95,7 @@
 	//Mostly for being overridden on precreated bellies on mobs. Could be VV'd into
 	//a carbon's belly if someone really wanted. No UI for carbons to adjust this.
 	//List has indexes that are the digestion mode strings, and keys that are lists of strings.
-	var/list/emote_lists = list()
+	var/tmp/list/emote_lists = list()
 
 // Constructor that sets the owning mob
 /datum/belly/New(var/mob/living/owning_mob)
@@ -294,7 +294,6 @@
 		if(!H)
 			H = owner
 		for(var/obj/item/W in H)
-			//_handle_digested_item(W,M) //The gut handles them now.
 			if(istype(W,/obj/item/organ/internal/mmi_holder/posibrain))
 				var/obj/item/organ/internal/mmi_holder/MMI = W
 				var/atom/movable/brain = MMI.removed()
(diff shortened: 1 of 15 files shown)


code/modules/vore/fluffstuff/custom_clothes_vr.dm
```diff
@@ -676,6 +679,9 @@
 		else
 			return 1
 
+/obj/item/clothing/under/fluff/screesuit/digest_act(var/datum/belly/belly = null)
+	return FALSE
```
Screemonster: oh lol the screesuit's still in? I thought I removed it entirely when markings became a thing. Musta just removed it from the fluff items .txt

code/modules/vore/eating/digest_act_vr.dm
```diff
@@ -0,0 +1,88 @@
...
+	if(belly)
+		for(var/obj/item/AM in contents)
+			AM.forceMove(belly.owner)
+			belly.internal_contents |= AM
+
+		belly.owner.nutrition += (1 * w_class)
+		if(isrobot(belly.owner))
+			var/mob/living/silicon/robot/R = belly.owner
+			R.cell.charge += (50 * w_class)
+	qdel(src)
```
Verkister: Needs borg sleeper support (borg sleeper is not a vorepanel belly) or everything will just nuke along with the possible contents for zero gains.

Arokha:
> The digest_act will not work properly on the borg sleeper bellies as they are not the vore organ datum type stuff, right?

Now it will.

code/modules/vore/eating/belly_vr.dm
```diff
@@ -303,22 +302,12 @@
...
-			else
-				if(!(istype(W,/obj/item/organ) || istype(W,/obj/item/weapon/storage/internal) || istype(W,/obj/screen)))//Don't drop organs or pocket spaces
-					H.remove_from_mob(W,owner)
-					internal_contents += W
+			for(var/slot in slots)
+				var/obj/item/thingy = M.get_equipped_item(slot = slot)
+				if(thingy)
+					M.unEquip(thingy,force = TRUE)
+					thingy.forceMove(owner)
+					internal_contents |= thingy
```
Cyantime: This will make any items that are removed as a result of removing the intended item fall onto the floor below the person. Pockets/belt/ID from uniform, suit storage from suits, etc.

Arokha: Uniform is one of the last things in that list, and as such it doesn't get removed. Uniform is the only object I'm aware of that causes you to throw all your stuff everywhere, and said stuff has already been removed at that point.

I did test it, and it works fine. I digested several people with full gear on and nothing falls on the floor (except for one time I digested someone with a ton of brute damage (100) and their feet came off and so their shoes fell off 👀 BUT THAT ASIDE...)

code/modules/vore/eating/contaminate_vr.dm
```diff
@@ -1,45 +1,39 @@
 var/image/gurgled_overlay = image('icons/effects/sludgeoverlay_vr.dmi')
 
 /obj/item
-	var/gurgled = 0
+	var/gurgled = FALSE
 	var/cleanname
 	var/cleandesc
 
-/obj/item/proc/gurglecontaminate()
+/obj/item/proc/gurgle_contaminate(var/list/internal_contents = null, var/atom/movable/item_storage = null)
```
Leshana: Would be good to document the meaning and purpose of the arguments.

code/modules/vore/eating/digest_act_vr.dm
```diff
@@ -0,0 +1,88 @@
+//Please make sure to:
+//return FALSE: You are not going away, stop asking me to digest.
+//return non-negative integer: Amount of nutrition/charge gained (scaled to nutrition, other end can multiply for charge scale).
+
+// Ye default implementation.
+/obj/item/proc/digest_act(var/list/internal_contents = null, var/atom/movable/item_storage = null)
```
Leshana: Would be good to document the meaning and purpose of the arguments.

PR #2726, opened 2018-01-10
Fixes science borg analyzer belly and couple other things. (Ready)

Verkister:
-Digestive Analyzer belly now works as intended.
-Can only load one item/victim at a time.
-It now tells the item's research value on nom and updates internal files on gurgles. Sync button is in the sleeper UI for the belly. The sync button now only shows up if there are unsynced updates.
-Fixed UI refreshing on all borg sleeper bellies. (updates correctly whenever contents change and such)
-Updated the entire sleeper code to use to_chat.
-Starting drain for sleeper gurgles is now defined by the type, the DA one having lower than others.
-Cannot load large items into the analyzer belly. (such as loaded backpacks etc.)
-Yes this does also exclude mech equipment and some other things, but that's when the human crew comes in handy.
-If you're playing a human scientist and just equipping a monkey with a loaded backpack to feed for the borg, you will only make yourself look like a fool as it will take lot less time and effort to just use the lab DA in correct order instead of enduring the slow wait for the borg to gurgle the monkey and its stuff one by one in a completely random order.
-Fixed new secborg belly overlay having inverted side directions.
-Removed borg baton charging (uses borg cell)
-Updated code for digestion refactor.
-Enabled mob_energy var to replace hardcoded charge gain.
-Cleaned up some things.


Arokha:
No belly that DA's every item in it, pls. One at a time, with cost for each.

code/modules/mob/living/silicon/robot/dogborg/dog_sleeper_vr.dm
```diff
@@ -126,6 +128,9 @@
 		dat += "<font color='red'><B>Current load:</B> [length(contents)] / [garbo.max_item_count] objects.</font><BR>"
 		dat += "<font color='gray'>([list2text(contents,", ")])</font><BR><BR>"
 
+	if(istype(src, /obj/item/device/dogborg/sleeper/compactor/analyzer))
+		dat += "<A href='?src=\ref[src];sync=1'>Sync Files</A><BR>"
```
Arokha: Use a var instead of this istype check, it's less expensive.

If you want to know what you should REALLY do, put some #defines at the top that represent capabilities, and add a var just called `capabilities` and make it a bitfield of capabilities, which will be very fast.

code/modules/mob/living/silicon/robot/dogborg/dog_sleeper_vr.dm
```diff
@@ -223,13 +229,30 @@
 		sleeperUI(usr)
 		return
 
+	if(href_list["sync"])
+		var/success = 0
+		for(var/obj/machinery/r_n_d/server/S in machines)
```
Arokha: This list is very large, and as far as I'm aware, there's no cooldown on clicking this button in this UI. Might be an easy DDoS target. Maybe you should just have them smush their face against the server since they have access, and are borgs (or add a cooldown).

code/modules/mob/living/silicon/robot/dogborg/dog_sleeper_vr.dm
```diff
@@ -409,6 +432,10 @@
 					return
 				if(T in items_preserved)
 					return
+				if(istype(src,/obj/item/device/dogborg/sleeper/compactor/analyzer) && istype(T,/obj/item/weapon))
+					var/obj/item/weapon/tech_item = T
```
Arokha: Why `/obj/item/weapon`? Are there not some `/obj/item` that have techlevels?

code/modules/mob/living/silicon/robot/dogborg/dog_sleeper_vr.dm
```diff
@@ -205,11 +211,11 @@
 					return
 				else
 					cleaning = 1
-					drain(500)
+					drain(100)
```
Arokha: Changing drain on every dogborg belly is shenanigans. Isn't this PR just about DA bellies?

code/modules/mob/living/silicon/robot/dogborg/dog_sleeper_vr.dm
```diff
@@ -570,7 +594,66 @@
...
 			hound.updateicon()
-			if(UI_open == 1)
+			if(UI_open == TRUE)
+				sleeperUI(usr)
+		return
+	return
+
+/obj/item/device/dogborg/sleeper/compactor/analyzer/afterattack(var/atom/movable/target, mob/living/silicon/user, proximity)//GARBO NOMS
+	hound = loc
+
```
Arokha: NTY to this giant copypaste.

code/modules/mob/living/silicon/robot/dogborg/dog_sleeper_vr.dm
Verkister: obj/item popped up undefined var error so not as far as I know.

Cyantime: That's strange, because origin_tech is defined on the /item level
https://github.com/VOREStation/VOREStation/blob/master/code/game/objects/items.dm#L21

And /weapon adds almost literally nothing, and is mostly a remnant of oldcode
https://github.com/VOREStation/VOREStation/blob/master/code/game/objects/weapons.dm#L1

Verkister: Huh yeah found the real reason for the error. The thing wasn't specified as obj/item before the part.

code/modules/mob/living/silicon/robot/dogborg/dog_sleeper_vr.dm
```diff
@@ -517,13 +547,58 @@
...
+			if(trashman.buckled)
+				to_chat(user, "<span class='warning'>[trashman] is buckled and can not be put into your [src.name].</span>")
+				return
+			user.visible_message("<span class='warning'>[hound.name] is ingesting [trashman] into their [src.name].</span>", "<span class='notice'>You start ingesting [trashman] into your [src.name]...</span>")
+			if(do_after(user, 30, trashman) && !patient && !trashman.buckled && length(contents) < max_item_count)
+				trashman.forceMove(src)
+				trashman.reset_view(src)
+				update_patient()
+				processing_objects.Add(src)
+				user.visible_message("<span class='warning'>[hound.name]'s internal analyzer groans lightly as [trashman] slips inside.</span>", "<span class='notice'>Your garbage compactor groans lightly as [trashman] slips inside.</span>")
```
CockatriceXL: Psst. Still calling it a garbage compactor on the message that the borg sees.

Verkister: whoops thanks

Verkister:
Updated for the refactor and added a couple other fixes and cleanups as well.

code/modules/mob/living/silicon/robot/dogborg/dog_sleeper_vr.dm
```diff
@@ -223,13 +231,32 @@
 		sleeperUI(usr)
 		return
 
+	if(href_list["sync"])
+		synced = TRUE
+		var/success = 0
+		for(var/obj/machinery/r_n_d/server/S in machines)
```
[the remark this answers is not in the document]
Verkister: It's the same thing in the regular portable DA, and there's no need for cooldown because it only shows up once if there are any new unsynced updates and immediately refreshes away on the click. It requires analyzing a new item in order to bring the button back

Arokha:
At least squash your commits please.

PR #2730, opened 2018-01-11
Renames leg cuffs to leg wraps, renames the gold ones

LittleBigKid2000:
Mainly to fix the inconsistent capitalization, partially to make it consistent with the names in the loadout, partially because I think they're too long to be considered a cuff.

I'd also change the pathing so that the cuffs aren't a subtype of black shoes for no apparent reason, but I'd rather not accidentally break anything.

I'm also going to take this opportunity to say that I want an all-black/grey variant of these.


LittleBigKid2000:
Accidentally left in a part of the title that was for a change I decided against doing, due to tiredness.

PR #2737, opened 2018-01-13
Stop explosions or singularity or whatnot from ripping holes in space.

Leshana:
* When we say ChangeTurf() space, we mean it.  Don't override us like that.


Screemonster:
wait, so changeturf to space was supposed to not allow space to be created when it was above another turf?

Then why the fuck were RCDs doing it in the atrium?

PR #2742, opened 2018-01-14
Plaaaaaaanessssssss

Arokha:
Converts several things to use Byond planes:
Lighting, Ghosts, AI Eye, The UI, In-game 'HUD' glasses huds

Adds the potential to perform shenanigans (these are not ingame yet):
* Adds colorblindness support for 13 types of colorblindness
* Allows for some things to be seen by some players and not others, without using invisibility
* Allows for making things look different for some players and not others, which can fix darksight

Adds an admin verb as a sort of PoC for plane things:
* Ghost View allows an admin to see ghosts while still in a mob

Fixes a few things:
* Mesons/material/NVG/various-other-goggles now properly remove darkness when worn and on
* Toggle darkness for ghosts now works better/faster

Improves a few things:
* HUDs like medhud/sechud/etc are now updated immediately and have no delays
* Cleans up all the garbage HUD code from Life() since it's unused
* Glasses now just have a list of planes they make you able to see, no proc-work necessary

Yes, there are things this PR touches areas that are not optimal. I am definitely not making anything slower with this, though. If anything, things should be much faster, especially in Life(), and for ghosts. 

I intend to come back in separate PRs and redo things like:
* The fullscreen overlays (mesons, blindness, welder vision, etc) and put them on planes properly
* Human update_icons can be vastly improved by using MA's instead of what it does now
* Reworking various other things into planes (pipes, cables, plating)
* Fixing darksight to actually remove the darkness layer around you (a bigger area for better darksight)


Screemonster:
Is this a thing polaris would like or nah? Or you sending it to them too?

Arokha:
> Is this a thing polaris would like or nah? Or you sending it to them too?

I'm testing it here because I can quickly fix it here and have server access to make copies and revert here if it breaks everything. But it's in two commits so I can cherrypick that one to them.

SpadesNeil:
Marking this "Send to Polaris" for later.

PR #2750, opened 2018-01-15
Make travis not compile Tether twice

Arokha:
What it says. Prevents tether from being compiled on the 'premaptest' time.


SpadesNeil:
Travis be like, "no"

Arokha:
Ye, these marks are all the places where Tether is referenced in hardcoded spots, which is bad. So we should fix those. This can sit here until then

Arokha:
Accidentally deleted this branch, oops.

SpadesNeil:
you can restore it yknow

PR #2754, opened 2018-01-15
Colorblindness Traits

Arokha:
Adds Monochromia, and the Vulpkanin and Tajaran colorblindness from Paradise. Can add other varieties if people want, I have all the RL varieties available in the backend.

For comparison


Jakksergal:
Would it be possible to add reduced vision range or fuzzy vision to this commit as well?

Arokha:
I can probably do that, but I'd do it as a separate PR since this is just colorblindness.
Next though I really just want to fix darksight.

Belsima:
It'd be awesome to integrate this into Polaris' trait system?

Arokha:
> It'd be awesome to integrate this into Polaris' trait system?
@Belsima We have that commented out on ours, but it'd be super easy for Neerti to do it probably.
Leshana and I are working on the PR that enables Planes and whatnot for Polaris.

PR #2755, opened 2018-01-15
Polaris Sync 2018 01 10

Leshana:
## Merged Polaris Upstream
This sync brings us only to 2018-01-10  (just before PolarisSS13/Polaris#4519) instead of today  because of timing vs. the hud/planes project.
### Features
Adds a species flag for preventing infections. Added to Prometheans and Xenos. PolarisSS13/Polaris#4515
### Fixes
Updates misc atmospherics machinery init and destroy PolarisSS13/Polaris#4514
Fixes misspelling 'variant' -> 'varient' PolarisSS13/Polaris#4510

### DIFF
--- a/code/game/machinery/cryo.dm
+++ b/code/game/machinery/cryo.dm
@@ -30,16 +30,9 @@
 	var/turf/T = src.loc
 	T.contents += contents
 	if(beaker)
-		beaker.loc = get_step(src.loc, SOUTH) //Beaker is carefully ejected from the wreckage of the cryotube
-	..()
-
-/obj/machinery/atmospherics/unary/cryo_cell/initialize()
-	if(node) return
-	var/node_connect = dir
-	for(var/obj/machinery/atmospherics/target in get_step(src,node_connect))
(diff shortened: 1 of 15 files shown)


code/game/machinery/cryo.dm
```diff
@@ -30,16 +30,9 @@
...
-/obj/machinery/atmospherics/unary/cryo_cell/initialize()
-	if(node) return
-	var/node_connect = dir
-	for(var/obj/machinery/atmospherics/target in get_step(src,node_connect))
-		if(target.initialize_directions & get_dir(target,src))
-			node = target
-			break
+		beaker.forceMove(get_step(loc, SOUTH)) //Beaker is carefully ejected from the wreckage of the cryotube
+		beaker = null
+	. = ..()
```
Arokha: Isn't this now indented incorrectly? Looks like 2 tabs then 1 tab.

Leshana: No it is correct.   Look in Split view, it just looks strange since I modified Destroy and *removed* initialize()

PR #2758, opened 2018-01-16
3000% human/update_icons() speed improvement

Arokha:
Using overlays in lists and a mutable_appearance.

Lots of little changes in files since I needed to read each time update_icons() was called to decide if it could call update_icons_layers/body/huds instead by itself, and determine when places were inappropriately calling it with the default TRUE parameter or not.


code/modules/mob/living/carbon/human/update_icons.dm
```diff
@@ -247,9 +286,10 @@ var/global/list/damage_icon_parts = list()
 	//0 = destroyed, 1 = normal, 2 = robotic, 3 = necrotic.
 
 	//Create a new, blank icon for our mob to use.
-	if(stand_icon)
-		qdel(stand_icon)
-	stand_icon = new(species.icon_template ? species.icon_template : 'icons/mob/human.dmi',"blank")
+	var/icon/stand_icon = new(species.icon_template ? species.icon_template : 'icons/mob/human.dmi',"blank")
+
+	//Clean old ma_body
```
Leshana: old `list_body`

code/modules/mob/living/carbon/human/appearance.dm
```diff
@@ -211,4 +209,4 @@
 /mob/living/carbon/human/proc/force_update_limbs()
 	for(var/obj/item/organ/external/O in organs)
 		O.sync_colour_to_human(src)
-	update_body(0)
+	update_icons_body()
```
Leshana: I think this can be update_icons_body(FALSE)  since the old one didn't call update_icons either.

PR #2762, opened 2018-01-16
Fixes science borg analyzer belly and couple other things. (*squish edition)

Verkister:
-Digestive Analyzer belly now works as intended.
-Can only load one item/victim at a time.
-It now tells the item's research value on nom and updates internal files on gurgles. Sync button is in the sleeper UI for the belly. The sync button now only shows up if there are unsynced updates.
-Fixed UI refreshing on all borg sleeper bellies. (updates correctly whenever contents change and such)
-Updated the entire sleeper code to use to_chat.
-Starting drain for sleeper gurgles is now defined by the type, the DA one having lower than others.
-Cannot load large items into the analyzer belly. (such as loaded backpacks etc.)
-Yes this does also exclude mech equipment and some other things, but that's when the human crew comes in handy.
-If you're playing a human scientist and just equipping a monkey with a loaded backpack to feed for the borg, you will only make yourself look like a fool as it will take lot less time and effort to just use the lab DA in correct order instead of enduring the slow wait for the borg to gurgle the monkey and its stuff one by one in a completely random order.
-Fixed new secborg belly overlay having inverted side directions.
-Removed borg baton charging (uses borg cell)
-Updated code for digestion refactor.
-Enabled mob_energy var to replace hardcoded charge gain.
-Cleaned up some things.

### DIFF
--- a/code/modules/mob/living/silicon/robot/dogborg/dog_sleeper_vr.dm
+++ b/code/modules/mob/living/silicon/robot/dogborg/dog_sleeper_vr.dm
@@ -11,15 +11,19 @@
 	var/min_health = -100
 	var/cleaning = 0
 	var/patient_laststat = null
-	var/mob_energy = 30000 //Energy gained from digesting mobs (including PCs)
+	var/mob_energy = -20000 //Energy gained from digesting mobs (including PCs)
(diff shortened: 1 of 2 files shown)


kcin2001:
Thank you very much verk

PR #2767, opened 2018-01-16
Fix dependencies on tether map in main codebase

Leshana:
1) Move unit test except areas into tether_defines where they should be.
2) Make the persistence of spawn-points at end of be configurable per map.
Added three vars to the map datum to confgure which spawn point should be persisted for players who are dead, evac'd or stayed on station.
3) Make the Metor Strike random event be configurable per map.
Added a variable to configure which areas are vulnerable to meteor strikes instead of hard coding it.


Leshana:
**Fixed the rest of the dependencies** They were all "wilderness" code.
* Reverted random.dm back to Polaris as there is no need for the VOREStation customs.
* Fixed /obj/random/outside_mob so it will actually work (You can't set vars on a type!)
* Removed all wilderness code outside of the tether map.  It was only setting life_disabled on mobs spawning there. We can do that with a subsystem or by dynamic maploading instead.

PR #2769, opened 2018-01-17
Changes AR-E glasses to show engineering alarms

AbbieFland:
They were linked to the base alarm_monitor, which meant they got a blank
alarm panel.

This changes them to use alarm_monitor/engineering which has the
atmos/fire/power alarms common to engineering.


Arokha:
Travis has been down for like 18 hours and it seems to be doing bull.

PR #2774, opened 2018-01-17
Prevent stalls in dogborg self-clean cycle

AbbieFland:
When something unexpected (such as mice, which are neither humans nor
items) is in a scrubpup's garbage compactor, it could stall forever when
performing a self-clean. This changes it to treat anything it doesn't
know how to digest as an indigestible item, treating it like other
indigestible items such as ID cards.

This also changes the self-clean damage ticks to harm anything living,
rather than just human.

### DIFF
--- a/code/modules/mob/living/silicon/robot/dogborg/dog_sleeper_vr.dm
+++ b/code/modules/mob/living/silicon/robot/dogborg/dog_sleeper_vr.dm
@@ -372,7 +372,7 @@
 	if(air_master.current_cycle%3==1 && length(touchable_items))
 
 		//Burn all the mobs or add them to the exclusion list
-		for(var/mob/living/carbon/human/T in (touchable_items))
+		for(var/mob/living/T in (touchable_items))
 			if((T.status_flags & GODMODE) || !T.digestable)
 				src.items_preserved += T
 			else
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/silicon/robot/dogborg/dog_sleeper_vr.dm
```diff
@@ -372,7 +372,7 @@
 	if(air_master.current_cycle%3==1 && length(touchable_items))
 
 		//Burn all the mobs or add them to the exclusion list
-		for(var/mob/living/carbon/human/T in (touchable_items))
+		for(var/mob/living/T in (touchable_items))
 			if((T.status_flags & GODMODE) || !T.digestable)
```
Arokha: It's good that this changed, since it should be `/mob/living`, but this does mean that those will still stay forever.

The only code that actually deletes mobs from bellies still checks to see if it's a human, on 387 (the check) and 430 (the qdel).

Also while you're there change 391's "T.stat & DEAD" because stat is not a bitfield, it should be ==

AbbieFland: Changing that check you mention on 387 works and gives everything the desired behavior: living things get digested once they die and indigestible things get set aside.

PR amended.

code/modules/mob/living/silicon/robot/dogborg/dog_sleeper_vr.dm
```diff
@@ -384,11 +384,11 @@
...
 		//Handle the target being a mob
-		if(ishuman(target))
-			var/mob/living/carbon/human/T = target
+		if(isliving(target))
+			var/mob/living/T = target
 
 			//Mob is now dead
-			if(T.stat & DEAD)
+			if(T.stat == DEAD)
 				message_admins("[key_name(hound)] has digested [key_name(T)] as a dogborg. ([hound ? "<a href='?_src_=holder;adminplayerobservecoodjump=1;X=[hound.x];Y=[hound.y];Z=[hound.z]'>JMP</a>" : "null"])")
```
Arokha: This isn't a performance issue, but just as a favor to me (and other admins), can you ammend the message admins to be in an `if(ishuman))` block? Just the message_admins thing, because I don't want to get spammed as dogborgs eat every mouse like I know CERTAIN PEOPLE do.

AbbieFland: It didn't even occur to me that the change also meant logging was no longer restricted to ishuman stuff. Amended.

PR #2780, opened 2018-01-18
Fix a runtime in deleting human

Arokha:
What it says. Don't do these if we're deleting ourselves and dropping all our stuff.


SpadesNeil:
I assume send this to Polaris?

Arokha:
They already did this there in a separate PR. I'm just doing it here because I don't want to wait. :p

SpadesNeil:
oh ok :U

PR #2793, opened 2018-01-18
Remove reaction reagent powergaming

Arokha:
Don't just delete all the reagents in our mob simply because there was a reaction. That doesn't make any sense. Pls.

image


Cyantime:
Time to fill someone with water and hook them up to a potassium IV for constant explosions.

SpadesNeil:
Medical outrage in 3... 2... 1...

Burrito-Justice:
I've neither used this nor seen it used (outside of accidentally tossing potassium in a beaker when I was making cryoxadone) so I'm not outraged.

_yet._

Arokha:
I've seen it used a couple of times. :p

PR #2795, opened 2018-01-18
More guns for the gun box

SpadesNeil:
- NT Mk98 (re)added.
- P92X added.

The NT Mk98 is just a M1911 clone as far as game mechanics and balance goes. It's a gun you should all be familiar with already.

The P92X is a 9mm pistol with a 10 round capacity. It's a straight upgrade from the 1911 if using rubber ammo because 3 extra rounds, but a downgrade per shot if using live ammo.


Arokha:
No love for the Glock/"Enforcer"? ;(

SpadesNeil:
Do we even have such a thing?

PR #2796, opened 2018-01-19
Removes redundant hypospray

Cyantime:
Removes hypospray/vr, which was previously ported to Polaris as hypospray/vial.
Removes the vial/vr subtype.
Includes some QOL stuff that's already been sent to Polaris.


Arokha:
how dare

Leshana:
oh ok!

PR #2803, opened 2018-01-20
Does what #2795 intended to do

Cameron653:
Makes the gunbox spawn.
Makes the taser spawn correctly
Some efficiency . = ..()


Arokha:
Why remove the taser tho, what if they prefer it and don't want to steal one out of an officer's locker

Cameron653:
I assumed the taser was placed there previously as a placeholder. I can add it to the detective closet itself, since having it mapped in is kinda eh.

Arokha:
Taser was there for 1+ years, along with the gun box

Cameron653:
Fixed. Goes to show how often I play detective

code/game/objects/structures/crates_lockers/closets/secure/security.dm
```diff
@@ -297,7 +297,7 @@
 		..()
 		new /obj/item/clothing/accessory/badge/holo/detective(src)
 		new /obj/item/clothing/gloves/black(src)
-		//new /obj/item/gunbox(src) // Vorestation edit. If you can't be trusted to not be stupid with lethals, you can't be trusted with less-than-lethals either.
+		new /obj/item/gunbox(src)
```
SpadesNeil: NYET.

Gun box contains _guns_. We just got through discussing why we ultimately decided not to give the detective guns as default equipment.

Gun box also does not contain tasers stop.

Arokha:
Ok then

PR #2806, opened 2018-01-21
Entopic Framework + AR Features

Arokha:
Lays down the framework for Entopics, which are AR-only objects, visible to anyone with AR glasses, or an AR overlay enabled in a NIF. Notable changes in this PR include:
* Allows people in Soulcatchers to project into AR, visible/hearable to anyone nearby with AR glasses or an AR overlay enabled in their NIF. They are dragged around by their pred, share the same vision range, and can't leave their sight. Preds can disable this feature on their soulcatcher.
* Replacing of NIFSoft Shops with essentially an AR object visible to people with AR on
* Reduction of the hunger/charge cost of AR overlays in the NIF by 5x
* A new NIFSoft that makes everyone look like farm animals ("World Bender"). Future features to come like "see the floor as grass" and such.

Other less-visible features are:
* Increase in the speed of visible_message and audible_message, and ensure they can be heard by everyone nearby, even if they are inside someone/an object/a soulcatcher
* Fix a lazy list deletion on ghosts

NB: For the Soulcatcher projection, you hear/see from the pred, so just like a holopad if you can see someone projecting in AR and start talking to them, they can only hear you if you can see their pred.


Cyantime:
The nifsoft shop can be infected by the malware random event, and needs to be visible to normal people so they can fix it.

Arokha:
>needs to be visible to normal people

It is, it has a small floor panel with wire and broken states and whatnot.

PR #2809, opened 2018-01-21
Dynamic Engine Loading

Leshana:
Includes the dynamic engine loading system and four engine maps:
- Supermatter
- Tesla
- R-UST
- Singularity

Aro Edit: This just adds the ability to load random engines. We'll just start with Supermatter and Tesla, because Singulo and R-UST need tweaks.


SpadesNeil:
I should summon engineering to discuss this new feature.

Jakksergal:
As much as I like having the option of different engines, I don't really like the idea of having a different engine each time. It's really hard to believe that any one engineering character could remember how to setup four different types of engines. 

Methinks we should incorporate a secondary engine on the surface and have another in space, maybe swap them out every time there's a disaster. That way, there's reason behind why each engine is where it is, because someone botched the last one.

Knightfall5:
I know how to operate all these engines, and will gladly teach others how to as well if this is put in.

Arokha:
Just FYI it would probably only be Supermatter and Tesla until tweaks to singulo and RUST are done.

Cameron653:
This is quite nice. Can see the map templates being used for additional things of this nature in the future, too. Like making randomly generated maintenance areas or the such.

Leshana:
It has been changed so that the Tesla energy ball rapidly dissipates energy when it is zapping machines and such, so that it will not last very long at all.  Also reduced the explosions slightly so it will still destroy machines etc but not destroy the entire floor too.

PR #2812, opened 2018-01-21
Couple janihound processor fixes.

Verkister:
-Fixes ingested decal type remains becoming uncleanable.
-Fixes nonhuman mobs giving human charge. (mouse overload)
-Fixes janigut not swelling up when reaching the limit.


code/modules/mob/living/silicon/robot/dogborg/dog_sleeper_vr.dm
```diff
@@ -448,6 +447,9 @@
 					items_preserved |= T
 				else
 					hound.cell.charge += (50 * digested)
+			if(istype(target,/obj/effect/decal/remains))
```
AbbieFland: Should this be else if? I think the else below might try and preserve already-digested items as-is.

PR #2813, opened 2018-01-22
Move dorms/cafe to Z1

Arokha:
Moves dorms/cafe down here, give cafe an outdoor view of glowing trees and gaslamps. This should protect it from any accidental engine shenanigans as people learn the new engines and destroy absolutely everything. Throws in a free maint door to the tram area for solargrub hunting or hanging out at that table.

Also adds a construction site right next to it for yer sauna shenanigans or whatever.

Also adds a stairwell up from right there to just outside the bar.

It looks about like this:
image

Also fixes a bug in the AR update. Fairly harmless one but might as well fix.


SpadesNeil:
I actually kind of like this.

Very-Soft:
I really like it!

PR #2826, opened 2018-01-22
Fixes RUST bugs and invisible gas blocker

AbbieFland:
The cross of invisible particle accelerator/gyrotron targets where the
field goes was set as impassible to everything but mobs. Including
air. Now it's impassible to beams and particles only.

One place was capping the 0-1 percent_instability float to 100 instead
of 1, easy fix.


Cyantime:
Should be sent to Polaris.

AbbieFland:
Ah you're right, I'll redo it there

PR #2834, opened 2018-01-23
Some borg sleeper tweaks

Verkister:
-Cleans up some redundant srcs.
-Tweaks and fixes some patient update and overlay behavior.
-Makes the sleeper gurgles use damage based gains instead of big crunch after prey death. Also drastically reduces the dead mob gain.
-Moves the reduced dead mob gain back to its original non-human exclusive spot.
-Utilizes more drain proc instead of clunky cell charge hardcode.
-Removes an unbenefitical DA nerf. (can take same size stuff as janigut again. using this to gamblefeed loaded backpacks still makes you look like an idiot shooting your own foot though.)
-Makes compactor and analyzer functions a variant in base sleeper afterattack proc instead of separate subtype proc.

### DIFF
--- a/code/modules/mob/living/silicon/robot/dogborg/dog_sleeper_vr.dm
+++ b/code/modules/mob/living/silicon/robot/dogborg/dog_sleeper_vr.dm
@@ -44,9 +44,6 @@
 	if(length(contents) > (max_item_count - 1))
 		to_chat(user, "<span class='warning'>Your [src.name] is full. Eject or process contents to continue.</span>")
 		return
-	if(target.buckled)
-		to_chat(user, "<span class='warning'>The user is buckled and can not be put into your [src.name].</span>")
-		return
 	
 	if(analyzer == TRUE)
 		if(istype(target, /obj/item))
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/silicon/robot/dogborg/dog_sleeper_vr.dm
```diff
@@ -279,10 +280,11 @@
 //For if the dogborg's existing patient uh, doesn't make it.
 /obj/item/device/dogborg/sleeper/proc/update_patient()
 	hound = src.loc
+
 	//Well, we HAD one, what happened to them?
 	if(patient in contents)
 		if(patient_laststat != patient.stat)
-			if(patient.stat & DEAD)
+			if((patient.stat & DEAD) || cleaning)
```
Arokha: stat is not a bitfield, don't use bitwise and. This only works coincidentally because DEAD is 2 and is a power of 2. Like if we add a constant in the future that is 6, this will return true, which would be unexpected.

code/modules/mob/living/silicon/robot/dogborg/dog_sleeper_vr.dm
```diff
@@ -293,13 +295,15 @@
...
+		if(UI_open == TRUE)
+			sleeperUI(usr)
 		return(patient)
 
 	//Check for a new patient
 	else
 		for(var/mob/living/carbon/human/C in contents)
 			patient = C
-			if(patient.stat & DEAD)
+			if((patient.stat & DEAD) || cleaning)
```
Arokha: not-bitfield-stat

code/modules/mob/living/silicon/robot/dogborg/dog_sleeper_vr.dm
```diff
@@ -522,7 +525,7 @@
 	if(istype(src,/obj/item/device/dogborg/sleeper/compactor/analyzer))
```
Arokha: The sort of general lack of polymorphism in this file is becoming a problem. `istype(src` should never be necessary.

Arokha:
Using Travis as your only testing environment is not acceptable. It appears like you are just making changes, and waiting to see if Travis accepts them, rather than testing your code on your machine. All your edits are obviously done in the web editor, which would seem to indicate you're not testing them locally at all.

The fact that you've made so many commits to this one file over the span of several PRs with many, many commits in each: No thank you.

We're done accepting these for now. Please fork vorestation, clone it on your machine, make your changes, test, make changes, test, in a normal iterative programming technique. Squash/rebase your changes into a single or a small number of commits for each feature, and then make a PR from your fork.

PR #2842, opened 2018-01-25
Allow pulling objects through step teleports

Arokha:
Such as the map edge loads for solars and cargo.


Screemonster:
oh man that one time i had to drag a miner-filled gaslamp indoors without bumping it when I was unable to pull it across the level transition

PR #2845, opened 2018-01-26
Various borg sleeper tweaks and fixes (tested working)

Verkister:
-Cleans up some redundant srcs.
-Tweaks and fixes some patient update and overlay behavior.
-Makes the sleeper gurgles use damage based gains instead of big crunch after prey death. Also drastically reduces the dead mob gain.
-Moves the reduced dead mob gain back to its original non-human exclusive spot.
-Utilizes more drain proc instead of clunky cell charge hardcode.
-Removes an unbenefitical DA nerf. (can take same size stuff as janigut again. using this to gamblefeed loaded backpacks still makes you look like an idiot shooting your own foot though.)
-Makes compactor and analyzer functions a variant in base sleeper afterattack proc instead of separate subtype proc.


Verkister:
travis did a fuck

PR #2847, opened 2018-01-26
Convert ZAS into a StonedMC Subsystem

Leshana:
* Creates the SSair subsystem which replaces the Setup, Start(), and Tick() procs of air_master.
* Remove the obsolete Setup and Tick procs.
* Adjust admin and debug verbs that dealt with some ZAS internals to work with the new implementation.
* Move all other procs from air_master into SSair type

Note: This leaves the global var `air_master` intact, now just pointing to SSair.  This was purely to simplify code review and debugging for the first stage.  A subsequent PR will rename air_master to SSair globally.
Note 2: Baystation also converted airflow into a subsystem.  This is forthcoming.
Note 3: Europa and Baystation also incorporate the c_airblock macroization optimization.  This is forthcoming.


Leshana:
I've removed the non-essential uses of `src`   the remaining ones are copying datum references to local references for sanic speed purposes.

PR #2855, opened 2018-01-27
Add explorer and SAR lockers to map + recharger

Arokha:
Useful! So they don't have to make ALL their gear.

Also adds borg charger to shuttle since there's no other way to recharge in the field (we should code some sort of field recharger thing...)


Cyantime:
Borgs can have their power cell swapped in the field, FPBs can drain from APCs or other sources.

Arokha:
FBPs can only do that if they have a NIF, bought that NIFsoft, and there are APCs on the mission. Realistically this shuttle HAS a generator somewhere, right? It's obviously powering a sleeper and airlocks and stuff. Might as well also power a borg charger. Human crew can carry food. FBPs can't, really, and have to walk all the way back to the shuttle. That's already enough of a disadvantage without being like "No you need to FLY BACK TO THE STATION"

PR #2856, opened 2018-01-28
QoL map changes

KotetsuRedwood:
Changes direction of science front desk windoor so that greyshirts can set their contraband for research on the desk and leave it there.
Adds a filing cabinet to the forensics lab for outdated/irrelevant evidence details.


LittleBigKid2000:
>for outdated/irrelevant evidence details

So, everything I collect as detective?

PR #2862, opened 2018-01-28
Allow admins to see say/emote near their own body

Arokha:
image

yw


Screemonster:
oh man, I don't have to LOOC people to let 'em know I'm dealing with some ahelp bullshit any more

SpadesNeil:
Polaris would probably love this.

PR #2863, opened 2018-01-28
Nasty contamination vocabulary update.

Verkister:
-Now in even more creative flavors for digested crap!
-Also fixes a subtle blinking pixel in sludge overlay effect.

Yeah I had some fun :v


code/modules/vore/eating/contaminate_vr.dm
```diff
@@ -12,7 +12,7 @@ var/image/gurgled_overlay = image('icons/effects/sludgeoverlay_vr.dmi')
 	if(!gurgled)
 		gurgled = TRUE
 		overlays += gurgled_overlay
-		var/gurgleflavor = pick("soggy","soaked","dirty","nasty","slimy","drenched","sloppy")
+		var/gurgleflavor = pick("soggy","soaked","dirty","nasty","slimy","drenched","sloppy","grimy","sludgy","stinky","mucky","stained","soiled","filthy","saucy","foul","icky","tarnished","unsanitary","messy","begrimed","cruddy","funky","disgusting","repulsive","noxious","gruesome","gross","putrid","yucky","tainted","feculent","putrescent","unsavory","smelly","fetid","smutty","acrid","pungent","unclean","contaminated","gunky","gooey","sticky","drippy","oozing","sloshed","digested","sopping","damp","gloppy","begraggled","churned")
```
SpadesNeil: Only remove fetid and feculent because those words are derived from 'feces'. We don't want those.

PR #2864, opened 2018-01-28
Remove sleep from fuel injection

Cyantime:
Particles already start moving when created, there's no need to start them here.


Cyantime:
whoops wrong github, again

PR #2872, opened 2018-01-29
Removes Assistant, replaces with PTO system

Arokha:
* Removes the Assistant job.
* Adds an Intern job and associated department alt-titles, expected to work and assist in that department. Useful for players trying to learn the game.
* Adds a Visitor job, which is basically what old assistant was, without the 'resident' alt-title. You can live at the colony, but you don't live on the station, so you're a visitor when you come to the station.
* Adds a new job that is essentially super-assistant for every department, which is "Off-Duty [Department]"
* Tracks how long every player plays every department for, records it in a database, essentially allowing them to accrue paid time-off.
* Allows players to play "Off-Duty" versions of their department, which consumes their paid time-off hours.

Off-Duty slots are not limited in number.
Visitor and Intern are not limited for now, but will be in the near future once people have some PTO hours built up. Admins can free more slots for both if necessary using the normal free job slot mechanics.

The current rate of hours accrual is 1 hour in a normal job for 3 hours in an off-duty job. This may be adjusted in the future, and can be adjusted per job. You can have a negative PTO balance (it's not like it forces you to log out when you reach 0 on your off-duty job) but you need to get back to > 0 before using the off-duty version again.

A timeclock is forthcoming to check your hours, and if there are available slots in your department, to 'clock in' and consume one of those slots. You can be granted additional PTO hours for `reasons`. We may or may not make this a feature of the HoP job for rewarding good works, but there are concerns about metagaming it, so it warrants further discussion.

Off-Duty people should probably be asked before faxing CC to ask for additional staff in their department, but they aren't required to do anything, and they are allowed to grumble loudly about being made to work on their time off if they do accept the offer of a job and pay and such.

EDIT: I should mention that time-off is per player, not per character. The goal is to have players occasionally play real jobs, not to force every character into a real job. If you want to have one hard-working medical character and then one slacker who always seems to be off-duty medical, go for it. I realize this isn't realistic because somehow one character earns PTO for another but the goal is "have players play real jobs" which is accomplished by just having it per player, no need to go as far as per-character.


Burrito-Justice:
Yes. Yes. Yes. **Yes.**
Destroy the assistants that don't assist. Make people do things. Salt everything.

Cyantime:
```/datum/reagent/sodiumchloride
	name = "Table Salt"
	id = "sodiumchloride"
	description = "A salt made of sodium chloride. Commonly used to season food."
	taste_description = "salt"
	reagent_state = SOLID
	color = "#FFFFFF"
	overdose = REAGENTS_OVERDOSE```

**Comment by SingoDingo** [comment]
**Possible Intern name changes:**
Engineering Intern: Engineering Apprentice
Medical Intern: Student Doctor
Research Intern: Lab Assistant
Security Intern: Junior Officer
Cargo Intern: Retail Worker

And perhaps give them access to their respective departments, like the junior officer role had. I tend to find that incentivizes people to actually teach them, cause they can't just be put out of sight and out of mind.

**Comment by SilencedMP5A5** [comment]
junior officer is already a sec job title.
bring back security cadet.

File: code/game/jobs/job/offduty_vr.dm [hunk 1]
```diff
@@ -0,0 +1,81 @@
+//
+// "Off-duty" jobs are for people who want to do nothing and have earned it.
+//
+
+/datum/job/offduty_civilian
```
SpadesNeil: I actually very much like the idea of adding "off-duty" roles.

SpadesNeil:
The only thing I don't like about this is the addition of "Paid Time Off" mechanics.

I like there being "Off-Duty" roles though. If roles are needed, it makes it easy to know "This person is capable of doing this job, no need to fax CC."

SingoDingo:
The idea is that you remove junior officer from the alt titles and put it as an intern position as well.

After all, why would literally anyone play the “I’m in security but be patient I’m learning” role without access, if there was a “I’m in security but be patient I’m learning” role **with** access.

Arokha:
> The only thing I don't like about this is the addition of "Paid Time Off" mechanics.

S'what we decided in the admin forum thread was the best way to coax people into playing jobs and not staying assistant forever.

Remember back to the olden days when we'd know everyone and PM people who played assistant too much and started trying to get them into job positions.

With as many players as we have now, it's slightly less possible to do that and with 25+ assistants and nobody in some departments on some shifts it's too heavy-handed to just start like... writing down assistant names and trying to figure out when one of them plays assistant too much.

Very-Soft:
The paid time off mechanics as described are also very generous. If you play 1 full shift, you'll be able to take 3 shifts as the PTO role. Encourages people to play sometimes, but still lets them be off duty more often than they have to be on.

Which seems good if you're just trying to gently encourage people into jobs, and not necessarily trying to get people into jobs ALL the time.

SilencedMP5A5:
to be honest, having glanced over the space lobby a couple of times i thought you were going to be completely gutting the assistant role. 

imo, this is a pretty neat change.

code/game/jobs/jobs.dm
```diff
@@ -44,6 +44,7 @@ var/const/ASSISTANT			=(1<<11)
 var/const/BRIDGE			=(1<<12)
 var/const/CLOWN				=(1<<13) //VOREStation Add
 var/const/MIME				=(1<<14) //VOREStation Add
+var/const/INTERN			=(1<<15) //VOREStation Add
```
Leshana: If INTERN is a rename of assistant (/datum/job/assistant) why do we need a new constant instead of the ASSISTANT constant?

Haery70:
This does raise questions:

Are the Interns gonna have basic access to their respective departments?

Will this encourage players to work rather than be ignored by the higher ranks that sometimes abandon their work for scenes and don't acknowledge their existence?

Will they have different names other than interns for some departments? (Because, seriously, Security Intern doesn't ring in my ears well.)

There might be more questions from me about this in the future.

code/game/jobs/jobs.dm
Arokha: Fixed

Arokha:
> Are the Interns gonna have basic access to their respective departments?

No for now, though we thought about doing that. Might change in the future, but right now 'basic access' is A LOT of access for some departments. Like the 'basic' science access would allow them to do literally all of R&D, make circuits, etc.

> Will this encourage players to work rather than be ignored by the higher ranks that sometimes abandon their work for scenes and don't acknowledge their existence?

Maybe? I'm not aware of that issue, I don't think this will affect it though, this is more about getting players to stop playing assistant only.

> Will they have different names other than interns for some departments? (Because, seriously, Security Intern doesn't ring in my ears well.)

I just updated them to `list("Apprentice Engineer","Medical Intern","Lab Assistant","Security Cadet","Jr. Cargo Tech")` based on feedback.

Arokha:
(Ok joke's over, real title time)

Arokha:
People and staff seem to think it's fine so in it goes.

blazerules:
So no more resident?

Screemonster:
Resident hasn't made sense since we moved to a map that lacks a residential elevator.

PR #2877, opened 2018-01-29
Makes all speech louder to micros

Arokha:
Improves realism by appending a message describing how loud all speech is to micros, who have tiny ears and are very small.


KonBear:
Can we make micro speech very small too~. I don't think my ears could pick up the pipsqueak all the way across the bar without *really* trying to listen. :^)

SpadesNeil:
I suspect this was a joke PR but I'm also **really tempted** to accept it.

Arokha:
Nooooo

PR #2878, opened 2018-01-29
Fix macro movement

Arokha:
Whenever a macro walks if there is not a solid wall under them, the floor collapses.


SilencedMP5A5:
this actually sounds hilarious

SpadesNeil:
I want to keep this for an April Fools day event.

PR #2880, opened 2018-01-29
All living mobs maintain an average size across every living mob

Arokha:
All living mobs have their size averaged and whenever any mob's size changes, the average goes up/down and all living mobs resize. The fewer mobs, the more a resize affects things. Repeatedly shooting one mob over and over with a shrink ray slowly shrinks every mob on the server, and vice-versa.


Arokha:
To explain, what I mean is if you have people A B C and their sizes are 100% across the board, when one of them tries to resize to 200%, it does:

100% + 100% + 200% = 400%, divided by 3 people = 133%, and then sets everyone on the server's size to 133%.

Repeated attempts to grow/shrink will result in furthering this number.

SpadesNeil:
https://www.youtube.com/watch?v=KufiqCh3Qd4

SilencedMP5A5:
i couldnt have phrased that any better myself, ace. :v

LittleBigKid2000:
I understand what is being said, but why?

Issue #2884, opened 2018-01-29
Quality of Life for miners is terrible


SpadesNeil:
#### Brief description of the issue
Mining is slow and tedious and uninteresting. It's not hard, but it's also too slow to be easy. But giving them fully loaded rigs is imbalanced. There needs to be a proper solution to this, probably changing multiple features in mining to make mining actually fun.

So I want people to discuss ideas.

I have some hotfixes for now but I want a good long term solution for this.

Ideas accepted so far:
- [x] Ripley needs a speed boost.
- [x] Ripley needs a battery drain nerf.
- [x] Ripley needs an ore scanner.
- [x] Mining needs a recharge pad for ripley. (I think this was resolved.)
- [x] Mobs in the mining area. Maybe tameable. Maybe just a threat to safety. Spices things up either way.
- [x] ~~Return of asteroid mining?~~ Underdark!
- [ ] Return of secret rooms.
- [ ] Ripley bumping into airlocks that cycle should make the airlock cycle to open that door. This is a feature on /tg/, I don't know why it doesn't work for us.
- [x] Make the big drills report on the supply channel when they fail, possibly with ID number.

#### Code Revision
- https://github.com/VOREStation/VOREStation/commit/7f48af2b4bb7ed4bd7a6dc5c363830bfd1e8c280

Screemonster:
I'm going to give ripleys a buff to movement. Step_in delay of 6 is ludicrous and a step_energy_drain of 10 is _fucking_ ludicrous. Probably give them a built-in ore scanner too.

KotetsuRedwood:
Something that was mentioned in the discord was treasure rooms making a comeback. Gives them random loot, and let's be honest, who doesn't like loot.

Tasald:
some kinda loot machine like on other servers? where they turn in ore to get better gear/ toys?

Dankman:
ore satchels that hold more than 50 ores at a time

SpadesNeil:
I'm adding the industrial hardsuits back as a **temporary** fix. Expect it to be removed again in the near future when we have some QoL fixes.

Another issue is the lack of recharge pads for ripleys in the field.

>ore satchels that hold more than 50 ores at a time

We have those. Bring an ore box with you, click on it with a full satchel. Holds infinite ores.

Natje0:
[...]

PR #2887, opened 2018-01-29
Improves the ripley slightly

Screemonster:
- Gives it a slight buff to movespeed. Can revise later if it's not enough.
- Slashes the movement power usage because holy fuck who thought 10 energy per step was a good idea?
- Gives it a mounted ore scanner in the exosuit interface tab so you don't have to keep getting out if you're using the big drills (which you should, being able to carry all the drill parts at once is like one of the biggest advantages of using a ripley to mine)

Hopefully helps with #2884 a bit.

### DIFF
--- a/code/game/mecha/working/ripley.dm
+++ b/code/game/mecha/working/ripley.dm
@@ -3,7 +3,8 @@
 	name = "APLU \"Ripley\""
 	icon_state = "ripley"
 	initial_icon = "ripley"
-	step_in = 6
+	step_in = 5 // vorestation edit, was 6 but that's PAINFULLY slow
+	step_energy_drain = 5 // vorestation edit because 10 drained a significant chunk of its cell before you even got out the airlock
 	max_temperature = 20000
 	health = 200
 	wreckage = /obj/effect/decal/mecha_wreckage/ripley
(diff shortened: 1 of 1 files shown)


code/game/mecha/working/ripley.dm
```diff
@@ -66,4 +67,51 @@
...
 
+// VORESTATION EDIT BEGIN
 
+/obj/mecha/working/ripley/verb/detect_ore()
+	set category = "Exosuit Interface"
+	set name = "Detect Ores"
+	set src = usr.loc
+	set popup_menu = 0
+
+	usr << "You start scanning for metal deposits."
```
Arokha: to_chat() and other stream operators below

I'm not a huge fan of the copypaste from ore scanner. If the scanner changes (like if ore categories change) you'd have to remember to paste it here. I assume this is just because you can't make do_after interrupt if they move the mech?

code/game/mecha/working/ripley.dm
```diff
@@ -3,11 +3,13 @@
...
 	icon_state = "ripley"
 	initial_icon = "ripley"
-	step_in = 6
+	step_in = 5 // vorestation edit, was 6 but that's PAINFULLY slow
+	step_energy_drain = 5 // vorestation edit because 10 drained a significant chunk of its cell before you even got out the airlock
 	max_temperature = 20000
 	health = 200
 	wreckage = /obj/effect/decal/mecha_wreckage/ripley
 	cargo_capacity = 10
+	var/obj/item/weapon/mining_scanner/orescanner
```
Arokha: Not marked as vorestation edit

Screemonster: FUCK

PR #2888, opened 2018-01-30
nerfs micromech punches

Screemonster:
Micromech punches no longer stun and can no longer break through walls. Copypasta strikes again.


Cyantime:
It's something to note that paralyse() isn't even the stun that'd make sense, it's 'punch into momentary unconsciousness'. Mechs are awful.

PR #2889, opened 2018-01-30
Changes to Arrivals Locker room

Very-Soft:
Turned the locker room into 2 rooms, 1 with more lockers, 1 security checkpoint

arrivals 1

arrivals 2

arrivals 3

I mostly did this as a test, as I'm learning how to map and things, but also thought that, what with the dorms being so close by, what with all their toilets, that maybe the locker room should have more lockers. And then there's a small security checkpoint, so you can roleplay papers please with Visitors.


SilencedMP5A5:
neat

SingoDingo:
I like it, personally!

Leshana:
I like how there is a bookshelf, and how there are donuts.

LittleBigKid2000:
Needs a speaker playing the Papers, Please theme on loop.

PR #2892, opened 2018-01-30
Latex Maid Dresses!

KotetsuRedwood:
Doesn't add it to the map because mapmerge is being a little binch. If someone could add one in the maintenance dungeon that'd be awesome


Very-Soft:
I'm planning to try to move a couple of trash piles around eventually, I'll try to remember to add one to the maintenance dungeon when I do that, though shouldn't really be necessary if they are obtainable from loadout, crates, and trash piles.

KotetsuRedwood:
It's more for the dungeon's sake than the dress's, to be fair. More kinky stuff in there, y'know?

PR #2893, opened 2018-01-30
Makes mining drills faster

Ketrai:
- All drills but borgdrill, and diamond drill are now 10% faster, which is a significant boost.
- Diamond drill is 20% faster
- Borg drill is roughly ~ 15% faster

Felt like with the removal of the rigsuit, mining feels extremely slow and clunky again. Ripley has gotten buffed, which makes it an excellent contendor now, with it being slightly slower whilst restricting movement, but better at tunneling, and gathering up veins. As well as allowing the hauling of large amouns of crates, and stationary drills.

This change should make the normal drill feel smooth enough for the early stages, especially nice if there is no research, such as on skeleton shifts, yet you wanna do some mining for abandoned crates and such. It also makes upgrading the drills feel even nicer. 

Before update:
Basic drill is 8x as slow as a diamond drill.
Advanced drill is 6x as slow as a diamond drill.
Even the best drill, aka jackhammer and plasma cutter, is 4x as slow as a diamond drill.
Borgs were around 3x as slow as a diamond drill; Meaning they start with far superior gear compared to a miner.

After update:
Basic drills are still 8x times as slow as a diamond drill, so that upgrade still feels just as powerful
Advanced drills are now 6.75x as slow as a diamond drill.
Jackhammer/plasma cutters are now 4.5x as slow as a diamond drill.
Borg drills are now 3.25x as slow as a diamond drill.

### DIFF
--- a/code/modules/mining/mine_items.dm
+++ b/code/modules/mining/mine_items.dm
@@ -53,7 +53,7 @@
 	item_state = "jackhammer"
 	w_class = ITEMSIZE_LARGE
 	matter = list(DEFAULT_WALL_MATERIAL = 3750)
-	var/digspeed = 36 //moving the delay to an item var so R&D can make improved picks. --NEO
+	var/digspeed = 40 //moving the delay to an item var so R&D can make improved picks. --NEO
 	origin_tech = list(TECH_MATERIAL = 1, TECH_ENGINEERING = 1)
 	attack_verb = list("hit", "pierced", "sliced", "attacked")
 	var/drill_sound = 'sound/weapons/Genhit.ogg'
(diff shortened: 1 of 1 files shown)


Arokha:
Alternatively you can override all of those values by making a mine_items_vr.dm file in the same folder and just doing:
```
/obj/item/weapon/pickaxe/drill
    digspeed = 27

 /obj/item/weapon/pickaxe/plasmacutter
    digspeed = 18
```
etc

Ketrai:
Ah yeah, I forgot. Been a while since I've edited vore station files.

Ketrai:
Moved to another pull, closing.

PR #2907, opened 2018-01-31
Rewrites beach mission to be on Virgo 4

Cyantime:
As per the Lore Book on Virgo Erigone, https://github.com/VOREStation/VOREStation/blob/release/code/modules/lore_codex/lore_data_vr/important_locations.dm.

Virgo 2 is described with things like ```The very high temperatures, dangerous (sometimes magma-filled) caves, and the only presence of civilization being mining operations has made tourism for Virgo 2 mostly non-existent, with the exception of explorers who specifically seek out hellish landscapes, which are plentiful with all the ruins, volcanoes, twisting caves, and general lawlessness. The occasional remains of previous explorers near certain hotspots somehow does not deter them."```
There shouldn't be an ocean for beaches, the atmosphere should be unbreathable. This is almost a perfect description for a Lavaland style environment, which could be explored in future away missions.

In comparison, ```Virgo Prime is the fourth planet of Virgo-Eirgone. Although primarily desert, temperatures linger around only 287 kelvin (~14°C). It is the only planet in the system with an environment that supports oxygen-breathing lifeforms. The planet has a very shallow ocean, which is only an average of 1000 meters deep.```
This is almost perfect for the beach area, and adds more possibilities for future missions on the planet.

### DIFF
--- a/maps/tether/submaps/alienship/alienship.dm
+++ b/maps/tether/submaps/alienship/alienship.dm
@@ -1,7 +1,7 @@
 // -- Datums -- //
 
 /datum/shuttle_destination/excursion/alienship1
-	name = "Virgo 2 Orbit"
+	name = "Virgo 4 Orbit"
 	my_area = /area/shuttle/excursion/space
 	preferred_interim_area = /area/shuttle/excursion/space_moving
 	skip_me = TRUE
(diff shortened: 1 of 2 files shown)


Screemonster:
The fact that it's the major inhabited planet of the system gives away teams even more of a reason not to go guncrazy when they see other people around who might have more of a right to be there than they do.

PR #2913, opened 2018-01-31
Changes voremob default gurglemode to itemweak

Verkister:
It's just kind of a bummer to lose all your important gear after losing a fight if backup doesn't arrive in time. After all the item friendly mode was introduced way after vore mobs.

-Changes voremob default gurglemode to itemweak.
-Makes digest chance struggle failure trigger item-friendly gurgles first.
-Second struggle failure during itemweak gurgles kicks in the full gurgles.

### DIFF
--- a/code/modules/vore/eating/belly_vr.dm
+++ b/code/modules/vore/eating/belly_vr.dm
@@ -466,7 +466,7 @@
 			digest_mode = DM_ABSORB
 			return
 		
-		else if(prob(digestchance) && digest_mode != (DM_ITEMWEAK | DM_DIGEST)) //Finally, let's see if it should run the digest chance.
+		else if(prob(digestchance) && digest_mode != DM_ITEMWEAK && digest_mode != DM_DIGEST) //Finally, let's see if it should run the digest chance.
 			R << "<span class='warning'>In response to your struggling, \the [name] begins to get more active...</span>"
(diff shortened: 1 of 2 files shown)


Screemonster:
Might be worth doing the same for their belly-interaction strugglechances too.

code/modules/vore/eating/belly_vr.dm
```diff
@@ -465,10 +465,16 @@
 			owner << "<span class='warning'>You feel your [name] start to cling onto its contents...</span>"
 			digest_mode = DM_ABSORB
 			return
-
-		else if(prob(digestchance) && digest_mode != DM_DIGEST) //Finally, let's see if it should run the digest chance.
+		
+		else if(prob(digestchance) && digest_mode != (DM_ITEMWEAK | DM_DIGEST)) //Finally, let's see if it should run the digest chance.
```
Screemonster: Remember this affects player bellies too that have interactions turned on, but tbh I can't see it being a problem. (I need to ask around or do some sneaky VVing and see if anyone actually uses them)

Leshana: digest_mode is not a bitfield.  You cannot use bitwise OR to test it.    I think you mean   `else if(prob(digestchance) && !(digest_mode == DM_ITEMWEAK || digest_mode == DM_DIGEST))`

PR #2914, opened 2018-01-31
Exercise equipment changes/refactor

LittleBigKid2000:
- Reduces copypasta
- Makes the kilograms variable for the scale not unnecessarily a part of the object anymore
- Changes the visible message from the scale so that it's not apparently speaking or something
- Makes the weight loss from exercise equipment respect the user's set weight_loss and their species' metabolism
- Nutrition loss isn't affected by metabolism or anything because I don't know exactly what variable to use

_Do tell me if I did anything horribly wrong, I don't actually have much of an idea what I'm doing_


code/modules/vore/weight/fitness_machines_vr.dm
```diff
@@ -129,13 +90,11 @@
...
 /obj/machinery/scale/attack_hand(var/mob/living/user)
 	if(user.loc != src.loc)
 		user << "<span class='notice'>You need to be standing on top of the scale for it to work!</span>"
 		return
 	if(user.weight) //Just in case.
-		kilograms = round(text2num(user.weight),4) / 2.20463
-		user << "<span class='notice'>Your relative weight is [user.weight]lb / [kilograms]kg.</span>"
-		user.visible_message("<span class='warning'>[user]'s relative weight is [user.weight]lb / [kilograms]kg.</span>")
+		var/kilograms = round(text2num(user.weight),4) / 2.20463
+		src.visible_message("<span class='notice'>[src] displays a reading of [user.weight]lb / [kilograms]kg when [user] stands on it.</span>")
```
Arokha: It's not important, but the current standard is to not use `src` unless you need to. Like the use inside the string in `[src]` is unavoidable, but the `src.visible_message` should really just be `visible_message` (If you decide you want to change this, change the src.loc on 95 too)

Arokha:
Tested?

LittleBigKid2000:
I did test it, then made changes I was 99% sure would work just fine. Tested it again just now, only problem I noticed is that I forgot to set the click cooldown for the lifter.

PR #2922, opened 2018-02-01
Prevent communicators from connecting to the exonet when off station.

Leshana:
Presumably they are out of range of our station's exonet node when far away.


Mewchild:
Will communicators work properly if a telecomms relay is on the layer?

Arokha:
Maybe in the future, but not for now, because things like headset check this whenever anyone talks which is somewhat expensive (they have to loop over a bunch of stuff), but communicators check this every 2 seconds to hang up on calls or video or change the little icon at the top so it'd be a lot more expensive to make that change on communicators.

Maybe if we mess with telecomms to make it less bad. Just imagine it's a high-bandwidth thing for now and requires a good connection at the station.

SpadesNeil:
Should update the wiki, then.

Arokha:
@SpadesNeil Updated. It's actually just one generic SoP right now that covers both: 
https://wiki.vore-station.net/Standard_Operating_Procedure#Scientific_Expeditions

PR #2945, opened 2018-02-03
Assistants/Interns don't earn or spend PTO

Arokha:


SpadesNeil:
why not put in assistant_vr.dm?

Arokha:
IT'S TOO LATE also I mostly want to be aware if they change anything about assistant so it conflicts so we can fix it. That's the excuse I'm going with.

PR #2967, opened 2018-02-04
Makes the energy carbine crate consistent.

Ketrai:
It made no damn sense to get 3 of them for the same price as the energy gun crate, which gives 2 energy guns, despite laser carbines being considered superior in some regards.

### DIFF
--- a/code/datums/supplypacks/munitions.dm
+++ b/code/datums/supplypacks/munitions.dm
@@ -46,9 +46,9 @@
 	containername = "Experimental weapons crate"
 	access = access_armory
 
-/datum/supply_packs/munitions/energyweapons
+/datum/supply_packs/munitions/energyweapons 
 	name = "Laser carbine crate"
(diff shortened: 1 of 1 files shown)


code/datums/supplypacks/munitions.dm
```diff
@@ -46,9 +46,9 @@
 	containername = "Experimental weapons crate"
 	access = access_armory
 
-/datum/supply_packs/munitions/energyweapons
+/datum/supply_packs/munitions/energyweapons 
 	name = "Laser carbine crate"
-	contains = list(/obj/item/weapon/gun/energy/laser = 3)
+	contains = list(/obj/item/weapon/gun/energy/laser = 2) //VOREStation Edit - Made to be consistent with the energy guns crate.
```
SpadesNeil: I'd normally say "just do this in munitions_vr.dm" but I think we should actually send this to Polaris as well.

PR #2968, opened 2018-02-04
Adds a grub knight.

Ketrai:
-Does not spawn in the grub infestation event.
-Currently unobtainable, expect through cargo. (seperate PR) #2971 
-Mainly just implementing them for grub-related events. Feel free to spawn them on away missions to fuck over synths. (Two EMP procs is generally enough to down you, if you get bad luck. then they'll eat you similar to other hostile mobs. Do not fight these as synth without backup. Do not rush these unprepared as organic. Organics generally have the benefit of being able to stock up on burn meds to counter them easily.)
-More robust than normal grubs, EMP more frequently, cause slightly more poison damage, slightly more often.
-Encourages the use of the chemical Vermicitol, can be made with grubmeat. (metabolizes slower than dermaline and ketolane. It's a lot less potent)
-Hostile, unlike the normal solargrub. They are generally seen as protectors of the hive, so they don't fuck around.


Ketrai:
The added file edits the 64x32 icons file, generally where all the otie sprites are located. Because these grubs are visually larger.

Ketrai:
Don't merge this just yet, testing a few more things

Ketrai:
Kay, good to merge, but will have to write out a little fix for grubs not injecting poison.

Ketrai:
Goin to use these good bois in a special away mission at some point. Generally since synths are being used as meatshields rn, and I've seen synths take out like, 20+ spiders without going down. So this is like, an indirect nerf towards synths and their combat effectiveness for explorer missions.

SpadesNeil:
I like this.

code/modules/mob/living/simple_animal/vore/solargrub.dm
```diff
@@ -125,4 +124,73 @@ List of things solar grubs should be able to do:
...
+	var/poison_chance = 65
+	var/emp_chance = 30
+
+/mob/living/simple_animal/hostile/solargrubknight/PunchTarget()
+	if(target_mob&& prob(emp_chance))
+		target_mob.emp_act(4) //The weakest strength of EMP
+		visible_message("<span class='danger'>The grub releases a powerful shock!</span>")
+	..()
+
+/mob/living/simple_animal/hostile/solargrubknight/PunchTarget()
```
Cameron653: Shouldn't these be  merged into just one /mob/living/simple_animal/hostile/solargrubknight/PunchTarget ?

code/modules/mob/living/simple_animal/vore/solargrub.dm
```diff
@@ -125,4 +124,73 @@ List of things solar grubs should be able to do:
...
+				L << "<span class='warning'>You feel a shock rushing through your veins.</span>"
+				L.reagents.add_reagent(poison_type, poison_per_bite)
+
+/mob/living/simple_animal/hostile/solargrubknight/handle_light()
+	. = ..()
+	if(. == 0 && !is_dead())
+		set_light(2.5, 1, COLOR_YELLOW)
+		return 1
+
+/mob/living/simple_animal/hostile/solargrubknight
```
Cameron653: This should be included with the rest of the vars above instead of in its own section down here.

Ketrai:
There we go!

code/modules/mob/living/simple_animal/vore/solargrub.dm
```diff
@@ -125,4 +122,70 @@ List of things solar grubs should be able to do:
...
+	var/poison_type = "shockchem"
+	var/poison_chance = 65
+	var/emp_chance = 30
+	
+	vore_active = 1
+	vore_capacity = 1
+	vore_pounce_chance = 0
+	vore_default_mode = DM_ITEMWEAK
+
+/mob/living/simple_animal/hostile/solargrubknight/PunchTarget()
```
Arokha: You should use `do_punch` for this rather than `PunchTarget`

code/modules/mob/living/simple_animal/vore/solargrub.dm
```diff
@@ -125,4 +122,70 @@ List of things solar grubs should be able to do:
...
+	. = ..()
+	if(target_mob&& prob(emp_chance))
```
Arokha: This will let them EMP non-adjacent things, because ..() will return null if they aren't adjacent, but you're not checking it here, and the mob will still have target_mob even if it's moved away. Which is why you should override `do_punch` for your special attacks.

code/modules/mob/living/simple_animal/vore/solargrub.dm
```diff
@@ -125,4 +122,70 @@ List of things solar grubs should be able to do:
...
+	if(ishuman(.))
+		var/mob/living/carbon/human/L = .
+		if(L.reagents)
+			if(prob(poison_chance))
+				L << "<span class='warning'>You feel a shock rushing through your veins.</span>"
+				L.reagents.add_reagent(poison_type, poison_per_bite)
+
+/mob/living/simple_animal/hostile/solargrubknight/handle_light()
+	. = ..()
+	if(. == 0 && !is_dead())
```
Arokha: Should check for `FALSE` not 0, it's a boolean. Should use TRUE/FALSE instead.
You could also compress this and the above line to `if(!..() && !is_dead())` but the way you're doing it is safer if something expects a return value from handle_light (I just know that nothing does). Don't need to change that.

Ketrai: mhm, gimme some time.

Ketrai:
Sigh, I'm not exactly a good coder you know. And now I'm gonna have to update the image file again.

Arokha:
Reopen (preferably squashed) when you've got it all in one piece.

PR #2969, opened 2018-02-04
makes vermicetol more potent.

Ketrai:
Seeing as I disregarded the effect of metabolism, vermicetol was unintentionally weak. So incredibly slow that the effect was barely noticable.

The potency now should cause vermicetol to heal at a rate of 2.2. which is roughly 63% lower than ketolade, but lasts a whole lot longer.

### DIFF
--- a/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Medicine_vr.dm
+++ b/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Medicine_vr.dm
@@ -73,4 +73,4 @@
 
 /datum/reagent/vermicetol/affect_blood(var/mob/living/carbon/M, var/alien, var/removed)
 	if(alien != IS_DIONA)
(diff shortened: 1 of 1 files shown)


code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Medicine_vr.dm
```diff
@@ -73,4 +73,4 @@
 
 /datum/reagent/vermicetol/affect_blood(var/mob/living/carbon/M, var/alien, var/removed)
 	if(alien != IS_DIONA)
-		M.heal_organ_damage(0, 21 * removed) //more potent than keloderm
+		M.heal_organ_damage(0, 110 * removed) //Not as potent as ketolade, but lasts LONG.
```
Arokha: What the hell is ketolade

Ketrai: Kelotane

PR #2972, opened 2018-02-04
Removes several ways to know every language

Arokha:
* Removes EAL implant from loadout
* Removes ear translator from loadout
* Alters handheld translator to not fit in pockets
* Make some languages untranslate-able

### DIFF
--- a/code/game/objects/items/devices/translator.dm
+++ b/code/game/objects/items/devices/translator.dm
@@ -4,7 +4,7 @@
 	desc = "This handy device appears to translate the languages it hears into onscreen text for a user."
 	icon = 'icons/obj/device.dmi'
 	icon_state = "translator"
-	w_class = ITEMSIZE_SMALL
+	w_class = ITEMSIZE_NORMAL //VOREStation Edit
 	origin_tech = list(TECH_DATA = 3, TECH_ENGINEERING = 3)
 	var/mult_icons = 1	//Changes sprite when it translates
 	var/visual = 1		//If you need to see to get the message
--- a/code/modules/client/preference_setup/loadout/loadout_ears.dm
+++ b/code/modules/client/preference_setup/loadout/loadout_ears.dm
@@ -8,8 +8,8 @@
 /datum/gear/ears/headphones
 	display_name = "headphones"
 	path = /obj/item/clothing/ears/earmuffs/headphones
-
+/* VOREStation Edit - Make languages great again
 /datum/gear/ears/translator
 	display_name = "universal translator, ear"
 	path = /obj/item/device/universal_translator/ear
-	cost = 8
+	cost = 8*/
--- a/code/modules/client/preference_setup/loadout/loadout_utility.dm
+++ b/code/modules/client/preference_setup/loadout/loadout_utility.dm
@@ -115,11 +115,11 @@
 	cost = 0 //VOREStation Edit. Changed cost to 0
 	slot = "implant"
 	exploitable = 1
-/*VOREStation Edit - Make language great again
+
 /datum/gear/utility/translator
 	display_name = "universal translator"
 	path = /obj/item/device/universal_translator
-	cost = 8*/
+	cost = 8
 
 /datum/gear/utility/pen
 	display_name = "Fountain Pen"
--- a/code/modules/client/preference_setup/loadout/loadout_utility.dm
+++ b/code/modules/client/preference_setup/loadout/loadout_utility.dm
@@ -101,13 +101,13 @@
 
 /datum/gear/utility/implant
 	exploitable = 1
-
+/* VOREStation Edit - Make languages great again
 /datum/gear/utility/implant/eal //This does nothing if you don't actually know EAL.
 	display_name = "implant, language, EAL"
 	path = /obj/item/weapon/implant/language/eal
 	cost = 2
 	slot = "implant"
-	exploitable = 1
+	exploitable = 1*/
 
 /datum/gear/utility/implant/tracking
 	display_name = "implant, tracking"
@@ -115,11 +115,11 @@
 	cost = 0 //VOREStation Edit. Changed cost to 0
 	slot = "implant"
 	exploitable = 1
-
+/*VOREStation Edit - Make language great again
 /datum/gear/utility/translator
 	display_name = "universal translator"
 	path = /obj/item/device/universal_translator
-	cost = 8
+	cost = 8*/
 
 /datum/gear/utility/pen
 	display_name = "Fountain Pen"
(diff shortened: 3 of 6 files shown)


kcin2001:
can you get the eal implant outside of loadout?

Ketrai:
Does this mean the CD and HoP, who  typically have to negociate a lot, still get access to translators?

kcin2001:
this feels alot like overkill

aztectornado:
I think that translation devices aren't that unheard of or insane. I disagree with this entire PR.

kcin2001:
the loadout translators were pretty balanced

Draycu:
Not exactly sure I like the sound of this, particularly for work as a HoP, CD, and psychologist as Ketrai pointed out.   

Also, if the implants are removed, can RnD still produce the ear translators and hand held translators?  I would certainly hope so, as only having the option to use the soul catcher is a bit much, I personally feel.

aztectornado:
What even is the lore for using a soulcatcher that way, anyway? That seems very _very_ overkill

Mewchild:
I personally don't see the reason for this nerf, honestly. The loadout items cost over half the points you have available. It's a big payin for them

Arokha:
The points, like R&D levels, are a totally pointless way to balance anything. People don't run out of points just like it's easy to get max R&D levels within 20 minutes of a shift starting. They're more a suggestion of difficulty/expense/balance than an actual barrier.

To the above examples of job usage, those departments spawn with translator items in their department already (unless someone deleted them off the map like an asshole).

R&D can produce translators still, yes.

The problem is that about half the server is able to understand every language all the time. It makes the languages pointless when potentially any random person could understand them. What's the point of having languages? It's not just 'fluff', it's also 'so I can talk privately to people who also know my language' which the omnipresence of translators defeats.

The fact is that y'all are generally a bunch of powergamers. Like the expeditions of all synth characters because no enemies do ion damage at the moment.

Mewchild:
But on the other side of this, how is that a problem? People being able to speak in their native tongue, and still have it understood can be every much a fluff thing as if only certain people can understand it? Given that races don't automatically have their lore-appropriate languages, it removes the incentive to ever even attempt to use anything outside of common, because you've no idea if anybody will be able to understand it, at all. I've seen several instances of people sitting in the bar for instance, three different languages being spoken in the same conversation, and everybody enjoying themself.

Ketrai:
I generally use farwa language if I want to speak in private, because it's non-translatable. Otherwise, go ahead, remove it from the loadouts, as long as stuff is still printable ^^.

Also outright calling people powergamers isn't very nice. 

Also, if you want enemies to fuck up synths on away missions, go add the grub knights I just made. they're aimed at messing up synths.

aztectornado:
Why are you even removing the EAL implant? What was the issue with scientists who could understand what is essentially space versions of IBM beep codes or morse?

Mewchild:
I'm okay with this now, based on the handheld able to be kept.

aztectornado:
Keep the EAL implant though?

Arokha:
No, what's the point of EAL if any organic can speak it with a super cheap implant? It was enjoyable when it was only able to be used by synths, which is how it should be.

kcin2001:
synths shouldnt need the implant

Cameron653:
Mm. I can agree with the PR seeing as the handheld translator now actually has to be held in your hand. Allows for eavesdropping but makes it so you are blatantly doing so.

Arokha:
> synths shouldnt need the implant

They don't need it to speak EAL. Anyone, everyone, has to take the language to speak EAL. The implant just lets you speak it, if you know it. It doesn't teach you EAL.
```
[4:34 PM] Aronai: No, you don't spawn with any languages other than galcom
[4:34 PM] Aronai: you can take the langauge though
[4:34 PM] Aronai: it's like any other
[4:34 PM] Aronai: but only synths can speak it (or anyone with the implant, which is dumb)
[4:36 PM] Aronai: @Jasmine/Yiris/Katelynn/Diana Previously EAL could be taken by anyone, but only synths could speak it, which was :thumbsup: because it was like people talking to R2D2 in star wars, where nearby people hear half the conversation and it's great for RP (because that's such a fun narrative trope)
[4:36 PM] Aronai: But now for like trivial points, anyone can speak/understand EAL completely, like synths
[4:36 PM] Aronai: Which essentially removes that specialness entirely
```

Screemonster:
Finally criminals can speak gutter without every sec officer having an ear translator to understand them, and every sec officer understanding gutter so they could bust criminals was the reason it got whitelisted in the first place.

Handheld translators are fine, you can't _passively_ listen in on every damn language with a handheld translator, and people can see when someone's got one out.

The fact is, we've had people OOCly snarking at players for _not_ taking the earpiece translator in their loadout, in a "why would you ever not take this" kinda way, so I'm absolutely fine with that toy being taken away if people can't use it properly. I wasn't a fan of them being added in the first place.

code/game/objects/items/devices/translator.dm
```diff
@@ -4,7 +4,7 @@
 	desc = "This handy device appears to translate the languages it hears into onscreen text for a user."
 	icon = 'icons/obj/device.dmi'
 	icon_state = "translator"
-	w_class = ITEMSIZE_SMALL
+	w_class = ITEMSIZE_NORMAL //VOREStation Edit
```
SpadesNeil: This makes it not fit in pockets. That's the only thing I think is bad. Fitting in pockets is totally reasonable. It's _handheld_ after all.

Screemonster: That was mainly to stop people turning it on and putting it in their pocket and still being able to use it, since it's supposed to read the translation out on a screen (which doesn't make sense in a pocket).
People used to use voice recorders as translators, and that required using them in-hand to get the translation.

Very-Soft:
I'm mostly okay with this though, one question;

If the EAL implant isn't in the loadout anymore, can it have persistence like NIFs do? 

Ultimately this might just lead to everyone just getting the implant, but that could come with some dangers I think. Maybe if you get emped and you have the implant, you can't speak at all anymore.

Maybe after installation you can't speak for a while or something.

I'm fine with it being removed from loadout, but I sort of feel like if it is not either persistent, or VERY EASY to install yourself once you get it from research, then no one will ever use it.

code/game/objects/items/devices/translator.dm
SpadesNeil: Fair enough.

ArgobargSoup:
I'm a _little_ bummed ta be losin' the ear translator from loadout.  It was useful as, say, a bartender/barista/chef, to be able ta understand customers who were more comfortable with their native language.  And, well, bein' able ta go "I heard you were talkin' shit like I wouldn't hear" at times.

I mean, half the conversations in different languages I've heard were over comms, which _don't_ get translated.

If the issue is bein' able ta understand certain really snowflake languages, why not remove the ability to translate those languages from the translator?  I mean, like @Screemonster mentioned, there's gutterband.  Which has already been established as an ever-evolving language that's constantly being updated to keep ahead of the authorities.  So _why would the translator have the files to translate it?_

I mean, at this point, I don't think I'll get anything but further nerfs, but yah.  :P

Very-Soft:
You can still get the headset translator from RnD it seems, and those aren't expensive to get at all. If anyone's running science, you could probably just ask for one. Having more things that people WANT to get from science is good in my opinion. Gives more of an actual job to science.

Dhaeleena:
No common no service!

PR #2974, opened 2018-02-04
Removes binoculars from xenoarch lockers

Cyantime:
What possible use does a xenoarch need binoculars for, spying on the rocks?

### DIFF
--- a/code/modules/xenoarcheaology/misc_vr.dm
+++ b/code/modules/xenoarcheaology/misc_vr.dm
@@ -1,9 +1,0 @@
-/obj/structure/closet/excavation
-	name = "Excavation tools"
-	icon_state = "toolcloset"
-	icon_closed = "toolcloset"
-	icon_opened = "toolclosetopen"
-
-	New()
-		..()
-		new /obj/item/device/binoculars(src)
--- a/vorestation.dme
+++ b/vorestation.dme
@@ -2674,7 +2674,6 @@
 #include "code\modules\xenoarcheaology\manuals.dm"
 #include "code\modules\xenoarcheaology\master_controller.dm"
 #include "code\modules\xenoarcheaology\misc.dm"
-#include "code\modules\xenoarcheaology\misc_vr.dm"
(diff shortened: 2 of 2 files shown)


KonBear:
Yeah! Then they should totes transfer it to the exploration team lockers. :^)

Cameron653:
Binoculars help xenoarch when you need to locate where an artifact clump is. You can use them to see the little brown artifact deposits that are out of range/eyesight.

Cyantime:
That's what digging exploratory tunnels is for, should miners also get binocs to spy on ore deposits?

Draycu:
I have to disagree, digging exploratory tunnels like that still risks damaging surrounding sites.  

I’m not sure why there is a need to remove the item, to tell the truth.  What harm is there in leaving it?

Cyantime:
Binoculars are an incredibly powerful item that when combined with any ranged weapon, render hostile mobs harmless because you can safely shoot them from far beyond the ranges that they can retaliate in. They're a 5 telecrystal traitor uplink item for a reason. Seeing the digsite markings from the edge of the normal 7-tile view range should be plenty of forewarning to check for the surrounding unmarked sites.

Cameron653:
The fact you can shoot mobs further away isn't an issue, since there's built in RNG when shooting from far away. If you're trying to snipe a mob with an egun you're going to run out of charge and _maybe_ get one or two hits, if that. It's practically useless unless you want to waste a shitload of time for a small payoff.

Additionally, the fact they're a traitor uplink item doesn't apply here. We don't have antag rounds, so it has little to no use in PVP (which is, again, subject to the same RNG penalties.)

I tested this the other day. Not only do you have to _see_ the mob to hit them (Meaning it doesn't work in places like on the asteroid area of expeditions) but you also have to deal with the RNG penalty. 


The binoculars are an extremely useful tool for xenoarch and saves a lot of time when it comes to locating digsites.

Arokha:
In my longstanding tradition of ruining everything, we'll see how this goes, since I'm told explorers pretty much just always take them.

PR #2976, opened 2018-02-05
"Planetary" Airlock Controller

Leshana:
### Ports cycle_to_external_air mode for airlock controllers
- Adds a flag to controllers that instructs the airlock program to always purge between IN/OUT and also to pump air to "outside" instead of tanks when purging to go OUT.
- Tweak the way the airlock fudges target pressures so it more reliably operates when trying to pump down to vacuum.

### Add "shuttle" sensors and vent pumps
- Adds an exterior airlock sensor that reads the air pressure in *front* of it, so it can be mounted on a shuttle but measure air outside the shuttle.
- Adds an exterior vent pump that pumps air to/from the turf *front* of it, so it can be mounted on a shuttle but exchange air outside the shuttle.

### DIFF
--- a/code/game/machinery/embedded_controller/airlock_program.dm
+++ b/code/game/machinery/embedded_controller/airlock_program.dm
@@ -109,13 +109,18 @@
 	var/shutdown_pump = 0
 	switch(command)
 		if("cycle_ext")
+			//If airlock is already cycled in this direction, just toggle the doors.
+			if(!memory["purge"] && IsInRange(memory["external_sensor_pressure"], memory["chamber_sensor_pressure"] * 0.95, memory["chamber_sensor_pressure"] * 1.05))
+				toggleDoor(memory["exterior_status"], tag_exterior_door, memory["secure"], "toggle")
 			//only respond to these commands if the airlock isn't already doing something
 			//prevents the controller from getting confused and doing strange things
-			if(state == target_state)
+			else if(state == target_state)
 				begin_cycle_out()
 
 		if("cycle_int")
-			if(state == target_state)
+			if(!memory["purge"] && IsInRange(memory["internal_sensor_pressure"], memory["chamber_sensor_pressure"] * 0.95, memory["chamber_sensor_pressure"] * 1.05))
+				toggleDoor(memory["interior_status"], tag_interior_door, memory["secure"], "toggle")
+			else if(state == target_state)
 				begin_cycle_in()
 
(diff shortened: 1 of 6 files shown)


code/game/machinery/embedded_controller/airlock_program.dm
```diff
@@ -38,6 +44,10 @@
 
 	if (istype(M, /obj/machinery/embedded_controller/radio/airlock))	//if our controller is an airlock controller than we can auto-init our tags
 		var/obj/machinery/embedded_controller/radio/airlock/controller = M
+		cycle_to_external_air = controller.cycle_to_external_air
+		if(cycle_to_external_air)
+			tag_pump_out_external = "[id_tag]_pump_out_external"
+			tag_pump_out_internal = "[id_tag]_pump_out_internal"
```
Arokha: So the X_pump_out_external are pumps outside the airlock on the potentially-planet? And X_pump_out_internal are inside the potentially-spaceship?

Leshana: Almost.  There is a second set of pumps to do the "pump out" operation.  An internal one is *inside the airlock* and an external one on the planet.

Arokha: So pump_out pumps are used when cycling to the outside, then? pump_in is cycling to internal? Or, can you just list the tags and where they go?

Leshana: ### Setup for a "cycle_to_external_air" shuttle airlock:
- `X_pump_out_external` - A vent pump on the "outside" of the shuttle, connected by pipes to the pump_out_internal vent.
- `X_pump_out_internal` - A vent pump inside the airlock chamber, connected by pipes to the pump_out_external vent.
- `X_pump` - A vent pump inside the airlock chamber, connected by pipes to air canisters.
- `X_exterior_sensor` - Sensor on the "outside" of the shuttle.  Must set this tag manually on the controller and sensor, its not auto-populated onto controller.
#### Cycling to "outside"
1. Purge the chamber by draining thru X_pump into the canisters.
2. Fill the chamber by siphoning planetary air via X_pump_out_external thru the pipe and out X_pump_out_internal.
#### Cycling to "inside"
1. Purge the chamber by draining thru X_pump_out_internal thru the pipe and out X_pump_out_external
2. Fill the chamber from the canisters thru X_pump

PR #2978, opened 2018-02-05
Removes Licking from context menu

SpadesNeil:
So you no longer right click on people and lick them instead of, say, pressing "give".

Also some grammar fixes for OOC escape.

### DIFF
--- a/code/modules/vore/eating/living_vr.dm
+++ b/code/modules/vore/eating/living_vr.dm
@@ -272,9 +272,10 @@
 // Clearly super important. Obviously.
 //
 /mob/living/proc/lick(var/mob/living/tasted in living_mobs(1))
-	set name = "Lick Someone"
+	set name = "Lick"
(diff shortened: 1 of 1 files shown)


Draycu:
Gotta say, I really appreciate this change.  Would be nice to not accidentally lick people when in a hurry, that’s for sure!

For clarification, it’ll be possible to still lick through other means, right?

KotetsuRedwood:
I'm not sure I like it. Makes it easier to slobber people up and let's be honest, that's _fun_

SpadesNeil:
>For clarification, it’ll be possible to still lick through other means, right?

Yes. Use the IC menu.

Leshana:
Booooooooo

ArgobargSoup:
Oh darn.  There goes a perfect way of adding hilarity and awkwardness to a situation.  :P

PR #2980, opened 2018-02-06
Makes the typing indicator work as it previously did.

Cameron653:
It suddenly started functioning properly (see: Disappearing in 30 seconds) which made it so it was practically useless to show that you're typing up a /me, which is what it was used for prior.

This makes it so it doesn't just suddenly disappear after an obscenely short amount of time, simply lasting until you send a message/close the window.

### DIFF
--- a/code/modules/mob/typing_indicator.dm
+++ b/code/modules/mob/typing_indicator.dm
@@ -1,4 +1,4 @@
-#define TYPING_INDICATOR_LIFETIME 30 * 10	//grace period after which typing indicator disappears regardless of text in chatbar
+#define TYPING_INDICATOR_LIFETIME 30 * 1000	//grace period after which typing indicator disappears regardless of text in chatbar //VOREStation Edit. Changed so people can use it to show if they're typing a /me.
 
 mob/var/hud_typing = 0 //set when typing in an input window instead of chatline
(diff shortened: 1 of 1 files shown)


LittleBigKid2000:
I'm glad I'm not the only one who noticed this and was bothered by this.

Draycu:
Yup, it'll help with long posts, for sure!  Thanks for working on this, seriously!

KonBear:
So the typing indicator will show when using the verb me too right?

Cameron653:
It should.

Arokha:
\*peers at web edit\* Did you test this?

Cameron653:
Just actually checked it and this _isn't_ the issue.

The issue is when update_icons is called on the mob. It gets rid of the typing indicator.

Arokha:
This is because update_icons (and calling any proc that calls update icons, like virtually everything in update_icons.dm) strips all overlays as it builds a new mutable_appearance to apply to the mob with their body/equipment/whatever overlays.

You can either add the typing indicator check to somewhere in update_icons (HUD? Seems like the most related) and then when you want someone to start typing, just call update_icons_huds. Then replace the existing set_typing_indicator to just toggle the variable on/off and call update_icons_huds

PR #2981, opened 2018-02-06
Fix Telecomms Turret Controls

Leshana:
I was thinking it would be good if the outer turret controls actually controlled the outer turrets.

### DIFF
--- a/code/game/machinery/turret_control.dm
+++ b/code/game/machinery/turret_control.dm
@@ -48,6 +48,8 @@
 /obj/machinery/turretid/initialize()
 	if(!control_area)
 		control_area = get_area(src)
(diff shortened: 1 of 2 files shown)


Knightfall5:
On a related but slightly less important note, the lethal turret controller for telecomms is powered by the APC inside the cooled freezer room... right next to the lethal turrets. If power goes out, you have to go inside without turning them off first, reenable power, and have a second person then shut off the turrets before they murder you to death.

Leshana:
@Knightfall5 That is no longer the case.  I didn't specifically mention it, but that is one of the things I changed on the map.  Both turret controls are now powered by the room they are actually in.

PR #2987, opened 2018-02-07
Adds frontier lasers to expedition prep room

Cyantime:
Adds some new guns, and gives the explorers some of them. They're weaker laser carbines that can't be fired on the station unless emagged or swiped with an ID that has armory access. Clicking them with an off-hand, instead of removing a power cell, activates the `crank charger` and slowly recharges the gun, for extended use in foreign environments.


SpadesNeil:
I have wanted crank charge guns for a while. I approve.

Draycu:
YES!  This will be a perfect weapon for explorers, I feel, particularly with the crank style charging while out on an expedition.  Thanks for working on this!

TheDavestDave:
Can we change it to the weapons being in a locker that can be opened off station.  It seems a little silly to have safties on a gun more complex than the gun its self.

Cameron653:
Could you perhaps make it so in R&D you can make them with metal, glass, & silver or perhaps gold? (Silver or gold to make it so you can't immediately create them/create a mass amount of them.) Since there'd be no other way to create them, only spawning with as many that spawn in the locker.

Cyantime:
Oh, I knew I forgot something yeah.
Also I imagined the locks were something NT installed afterwards and are just sort of stuck onto the gun, not an intended part of them.

Cyantime:
I decided to add them to cargo rather than RnD, because they aren't supposed to be any sort of high-tech things.

SilencedMP5A5:
its about goddamn time

KonBear:
How perfect~. Now sec doesn't have to bother us about it, we don't have to worry about battery power out there and we are not over powered. Will there be just one firing mode for it, or is there plans for more?

Cyantime:
It has normal shots which are lasergun beams without armor pen, a slightly longer firing delay, and 8 shots from a full cell. And 'low-power' shots which are the suppressive mode on the carbine also with a slightly higher fire delay.

KonBear:
Perfect, thanks~! Good job on them!

PR #2994, opened 2018-02-08
Adds a fancy new voremob.

Verkister:
-"corrupt hound". Basically a messed up evil feral borgo.
-Currently only available without bus shenanigans through the dangerous pred crate. Could be added to some expeditions and such I guess.
-May loot drop illegal borg module, vtec module, spooky cult knife, or compliance disk.

badboi


code/modules/mob/living/simple_animal/vore/badboi.dm
```diff
@@ -0,0 +1,66 @@
...
+	max_n2 = 0
+
+	vore_active = 1
+	vore_capacity = 1
+	vore_pounce_chance = 30
+	vore_icons = SA_ICON_LIVING
+	vore_stomach_name = "fuel processor"
+	vore_stomach_flavor = "You have ended up in the fuel processor of this corrupted machine. This place was definitely not designed with safety and comfort in mind. The heated and cramped surroundings oozing potent fluids all over your form, eager to do nothing less than breaking you apart to fuel its rampage for the next few days... hours... minutes? Oh dear..."
+
+	loot_list = list(/obj/item/borg/upgrade/syndicate)
```
Arokha: Make this like a 5% chance (this list actually allows for `loot_list = list(/type = prob)` because if I add these to a spawner people are going to get like 30 per away mission.

Draycu:
OH, heck yes!  Love this new mob, fantastic work!

I really hope this does get added for the new map, think it might fit well, personally!

code/modules/mob/living/simple_animal/vore/badboi.dm
```diff
@@ -0,0 +1,66 @@
...
+	icon_living = "badboi"
+	icon_dead = "badboi-dead"
+	icon_rest = "badboi_rest"
+	faction = "corrupt"
+	maxHealth = 200
+	health = 200
+	minbodytemp = 150
+	hostile = 1
+	investigates = 1
+	retaliate = 1
```
Arokha: I know where you copied these from uses 1 but be a good pupper and do TRUE for all the bool vars in this file

code/modules/mob/living/simple_animal/vore/badboi.dm
```diff
@@ -0,0 +1,66 @@
+/mob/living/simple_animal/badboi
```
Arokha: Should be a /hostile subtype, then remove the hostile = 1, retaliate = 1 etc. Look at the hostile subtype for vars you can remove (it's at the bottom of simple_animal.dm)

code/modules/mob/living/simple_animal/vore/badboi.dm
```diff
@@ -0,0 +1,66 @@
...
+/mob/living/simple_animal/badboi/PunchTarget()
+	if(istype(target_mob,/mob/living/simple_animal/mouse))
+		return EatTarget()
+	else ..()
+
+/mob/living/simple_animal/badboi/death(gibbed, deathmessage = "shudders and collapses!")
+	resting = 0
+	icon_state = icon_dead
+	update_icon()
+	..()
```
Arokha: Why? ..() calls update_icon(), which sets icon_state = icon_dead after 0.3s. Might still need to do the resting thing though if that's special to your pup. Should do . = ..() too for good measure in case anything cares about the death() return value.

code/modules/mob/living/simple_animal/vore/badboi.dm
```diff
@@ -0,0 +1,67 @@
...
+/mob/living/simple_animal/hostile/badboi/PunchTarget()
+	if(istype(target_mob,/mob/living/simple_animal/mouse))
+		return EatTarget()
+	else ..()
+
+/mob/living/simple_animal/hostile/badboi/death(gibbed, deathmessage = "shudders and collapses!")
+	.=..()
+	resting = 0
+	icon_state = icon_dead
+	update_icon()
```
Leshana: This call to update_icon shouldn't be necessary because parent is already going to call it in 3 ds.  Otherwise the rest of this seems proper.

PR #2996, opened 2018-02-08
Fixes colorblindness disabling HUDs

Verkister:
->Thing that updates hud visuals requires vis_enabled list in the mob to work.
->The thing that creates the vis_enabled list in the first place is in mob login proc.
->The forementioned list is only created alongside the mob's plane_holder if the mob getting logged in lacks plane_holder.
->The trait created plane_holder for the mob on spawn, skipping the forementioned login proc.
->Fixed the thing by making the trait also create the vis_enabled list while at it.

kuva


Arokha:
Actually imo in mob/initialize just do LAZYINITLIST(vis_enabled)

Verkister:
Having trouble finding that part.

Verkister:
https://github.com/PolarisSS13/Polaris/pull/4762

PR #2998, opened 2018-02-08
AwooCosplayCrate

Jwguy:
After talking with Natje and getting permission, I am continuing their efforts to create a full set of Awoo/Wolfgirl clothing and props. Most of what I have here is based off of their efforts, with some changes to the shield sprites, and extending the sword sprites into two sets for every directional facing. The original sword code was lost, so I tried my best to rebuild it to Natje's description (i.e. a switchable mode sword that could be activated to do more damage, but does a measly 1 HALLOSS when not activated). I think it came out well.

Furthermore, as per their request, I've added these items as well as their wolfgirl fluff clothing (see Awen Henry fluff app) to a 'Wolfgirl Cosplay Crate' that can be ordered from Cargo for 50 supply points. In summary, this does the following:

Adds Wolfgirl Cosplay Crate to Supplypacks under Recreation.
Adds Wolfgirl Replica Sword (Ignore the screenshot's name for it) and Autumn Shield.
Adds the above and three existing items to aforementioned crate.

Previous discussions of these items seemed to be mostly positive with the exception that the shield may have been too bulky for a fluff item (depending on how large it actually is, it may or may not be. If you need I can supply reference pictures, as I am wont to do).

As far as the shield blocking things.... personally, I've tested it thoroughly and it only seems to block melee attacks (i.e. made from within a single tile's range) and only every now and then, and even then, only if they come from directly in front of you. Ranged attacks of any kind bypass it completely and even melee weapons can circumvent it most of the time. Personally, that doesn't seem that bad, to me, but if we have to do something about it, I'd be happy to talk it out.

As of yet, there's not much lore on Wolfgirls, but here's hoping I can figure out how to make an exploration map, and then I'll try to make some through that. =)

vhdqsm2
eq6k6jn
capture


SpadesNeil:
More cosplaying crates and costumes in general could be fun.

code/modules/vore/fluffstuff/custom_items_vr.dm
```diff
@@ -1911,4 +1911,110 @@
...
+	attack_verb = list("attacked", "slashed", "stabbed", "sliced", "torn", "ripped", "diced", "cut")
+	sharp = 1
+	edge = 1
+	icon_state = "[active_state]_sharp"
+	damtype = BRUTE
+
+
+/obj/item/weapon/melee/fluffstuff/awoosword/deactivate(mob/living/user)
+	if(active)
+		user << "<span class='notice'>The [src] grows dull!</span>"
```
Leshana: Preferred to be  `to_chat(user, "<span class='notice'>The [src] grows dull!</span>")`

Jwguy:
Added the preferred change to the deactivation text.

code/modules/vore/fluffstuff/custom_items_vr.dm
```diff
@@ -1911,4 +1911,110 @@
...
+	slot_flags = SLOT_BACK | SLOT_OCLOTHING
+	active_force = 15
+	active_throwforce = 7
+	active_w_class = ITEMSIZE_LARGE
+	force = 1
+	throwforce = 1
+	throw_speed = 1
+	throw_range = 5
+	w_class = ITEMSIZE_SMALL
+	origin_tech = "materials=7"
```
Arokha: This is supposed to be a list and not a string. Like `origin_tech = list(TECH_MATERIALS = 7)`

Jwguy: Gotcha'. I honestly have no idea what that value is, I just copied it from a handful of the other fluff items like Joan's Rapier, which is materials=7.

I'll get it changed now.

Jwguy:
Changes made; Origin_tech value has been corrected to `origin_tech = list(TECH_MATERIAL = 2, TECH_COMBAT = 1)`

PR #3001, opened 2018-02-08
Adds explorer alt titles back

Cameron653:
Got pinged that people that wanted the alt titles back since they got removed for whatever reason.

They each have their own specialized items they spawn with. Nothing too gamechanging, however.
These items can be seen here:
https://github.com/VOREStation/VOREStation/blob/300bcd123f71d3698ced93452699bedc3eb9ea3c/maps/southern_cross/job/outfits.dm#L20-L30

Tech spawns with a full toolbelt. Medic spawns with a first aid kit.

Really nothing major/anything that can't be obtained in ~1 minute.

### DIFF
--- a/code/game/jobs/job/z_all_jobs_vr.dm
+++ b/code/game/jobs/job/z_all_jobs_vr.dm
@@ -47,3 +47,8 @@
 /datum/job/atmos
 	total_positions = 3
 	spawn_positions = 3
+
+/datum/job/explorer
+	alt_titles = list(
+ 		"Explorer Technician" = /decl/hierarchy/outfit/job/explorer2/technician,
+		"Explorer Medic" = /decl/hierarchy/outfit/job/explorer2/medic)


Cyantime:
Explorer medic is SaR, what does that alt-title actually bring to the job? These both seem to just support the "I'm an explorer but ALSO I can do all of the doctoring/engineering so we don't need to bring an actual one of those" mindset.

Cameron653:
I have no idea. Search and rescue is more specialized while medic is more general?

People mentioned it was removed and (as far as I can see w/ blame) there really wasn't a reason given and asked if it could be added back in.

TheDavestDave:
well engineers are kind of not ment to leave the station. And if they did they would be expected to do non-combat things. An exploer tech should be expected to open doors they find or knock down a wall that wherer they might get exposed to fire, but not set things like the relay or build a base

TheDavestDave:
maybe drill paths in the cave for the team, but not go mineing there, or digging for artifacts

Arokha:
The SoP says specifically that you can/should take an engineer with you. They we're removed on Pol because people took it to mean they basically replaced SAR and engineers, and I didn't bother avoiding the change here because I agree.

TheDavestDave:
SoP say may take an engineer. And it may invole takeing the only station engineer to get shot at on a misson to open a door. If you made it so the com thingy had to work on sight, a real engineer would have to come to set it out, work out power. Help start a basic treatment center while the tech drill holes, shoot shit, and take down doors. It's be the diffrence between SAR and docter comeing along

Cameron653:
Should I remove the gear they get and simply leave the titles, then?

TheDavestDave:
maybe add the tilte to the Engineering department. I think a pay diffrence would be in an engineering job were you build a wall, and one you build a wall getting shot at

Cameron653:
Oh, just read the comment before that. If people are taking the alt title instead of an engineer just _because_ of the alt title, I can understand it's removal.

Arokha:
We tell people a lot of things and express distaste for them, and people often ignore said things. It usually ends up having mechanical changes to enforce them (whether that's kicks/bans or code changes). I /know/ there are explorers who think exptech and expmed 'should be allowed' to do surgery and everything else, and though I can tell them not to, I trust them about as far as I can throw them to actually not do it. And we can't always watch. I can't add an admin log to report whenever any nonmedical person is doing surgery (well, I can, but I'd rather not)

Very-Soft:
I'm not really sure why the alt titles are necessary, and I don't really like what they imply. You've already got SaRs which go on the ship. And as stated above, the ship's supposed to get one of the engineers to come with them.

These alt titles are an easy avenue for confusion in one's role on the shuttle, -especially- if they spawn with stuff to reinforce that confusion. If one comes on as an explorer medic, and spawns with a medkit, they're gonna think they're medical.

I'd certainly like alt titles for the shuttle jobs, but I don't think these ones are good.

Cameron653:
Mm, very well. Going to close this seeing as people were apparently using it to replace actual required crew.

PR #3004, opened 2018-02-09
Ion gun / EMP bugfix and refactor

Screemonster:
- Fixes a bug where emp_act() was being called on all of a human mob's organs twice. EDIT: fix has now been merged on polaris, next sync will get it.

- Makes synth internal organs have special behaviour for EMPs, rather than just taking damage. Cells lose power (aka nutriment), eyes go blurry, MMIs gain system instability. Should be less of a pain to repair, but watch out for that instability 'cause you can't just fix that with wire.

- Changes the way ion guns work. Rather than just having a fixed pulse size, we can now specify a radius for each severity of pulse. Ion rifles now only deliver their max severity on a direct hit, but they gain an extra radius of weak severity. Pistols now have no AoE, but still deliver full severity. Hivebots and UFO mission turrets deliver a weaker bolt that does the rifle's splash damage on a direct hit, and a weak pulse on the AoE. (these may be buffed back to rifle shots if the above bugfix turns out to make the missions too easy)


SpadesNeil:
Polaris might like these changes.

TheDavestDave:
does this mean I can stop wondering how a player getting hit 3 tiles away kills me but not the droid fireing it? Is this going to make synths too OP for getting shot at?

code/modules/organs/internal/organ_internal.dm
```diff
@@ -62,3 +62,21 @@
 		if (prob(50))
 			take_damage(1,silent=prob(15))
 
+// Vorestation addition begin
```
Cameron653: All the code in this file would be better suited in a _vr file to prevent conflicts.

code/modules/projectiles/guns/energy/special.dm
```diff
@@ -23,7 +23,10 @@
 	force = 5
 	slot_flags = SLOT_BELT|SLOT_HOLSTER
 	charge_cost = 480
-	projectile_type = /obj/item/projectile/ion
+	projectile_type = /obj/item/projectile/ion/pistol // vorestation edit, still packs a punch but no AoE
+
+/obj/item/weapon/gun/energy/ionrifle/weak // Vorestation addition
```
Cameron653: Would also be good to move the edits to a _vr file.
Could also overwrite the ion rifle weak above by doing
```c
/obj/item/weapon/gun/energy/ionrifle/weak
	projectile_type = /obj/item/projectile/ion/pistol
```

and also overwrite the ion pistol by doing

```c
 /obj/item/weapon/gun/energy/ionrifle/pistol
	projectile_type = /obj/item/projectile/ion/pistol
```
instead of modifying this file.

Screemonster: Will do that when I get home from work.

Screemonster:
The emp_act bugfix has been sent to polaris here:
https://github.com/PolarisSS13/Polaris/pull/4767

code/modules/organs/internal/organ_internal_vr.dm
```diff
@@ -0,0 +1,15 @@
+/obj/item/organ/internal/eyes/emp_act(severity)
+	if(!(robotic >= ORGAN_ASSISTED))
+		return
+	owner.eye_blurry += (4/severity)
+
+/obj/item/organ/internal/cell/emp_act(severity)
+	if(!(robotic >= ORGAN_ASSISTED))
```
Arokha: This will never be NOT robotic. It's not like you get this if you pick mechanical heart or anything like eyes might be.

code/modules/organs/internal/organ_internal_vr.dm
```diff
@@ -0,0 +1,15 @@
...
+		return
+	owner.nutrition = max(0, owner.nutrition - rand(10/severity, 50/severity))
+
+/obj/item/organ/internal/mmi_holder/emp_act(severity)
+	if(!(robotic >= ORGAN_ASSISTED))
```
Arokha: Same

Issue #2995, opened 2018-02-08
Engineering Mr. Handy has no eyes overlay


Heroman3003:
#### Brief description of the issue
Mr. Handy skin of Engineering cyborg module doesn't have eyes overlay

#### What you expected to happen
Mr. Handy skin of Engineering cyborg module should have eyes overlay to showcase active/not active state

#### What actually happened
it doesn't

#### Steps to reproduce
- (Step 1) Choose Engineering cyborg module and then Handy skin
- (Step 2) Look at him

#### Code Revision
- f9feb190c403c43650399fb5a4558187d5d155be 
 
#### Anything else you may wish to add:
- Issue appears to stem from robots.dmi sprite file having two different eyes-handy-science sprites and no eyes-handy-engineer. Recommended solution is to rename orange version of eyes-handy-science into eyes-handy-engineer

Arokha:
https://github.com/PolarisSS13/Polaris/pull/4831

Arokha:
You probably fixed this in your borg sprites PR

PR #3007, opened 2018-02-09
Fixes 2995 for real I swear

Burrito-Justice:
Fixes https://github.com/VOREStation/VOREStation/issues/2995 for real this time.

Renames a sprite to what it should probably be named to so that the engineering model of Mr. Handy can use light overlays.


Verkister:
This should be a Polaris fix

Burrito-Justice:
Seems https://github.com/VOREStation/VOREStation/pull/3008 is already fixing this and more so I'll just close this

PR #3019, opened 2018-02-11
Updating Whitelist

Mewchild:
Adds WTFismyname to Xenomorph Hybrid - https://forum.vore-station.net/viewtopic.php?f=46&t=1256


config/alienwhitelist.txt
```diff
@@ -24,6 +24,7 @@ singo - Gutter
 some~user - Species
 tastypred - Xenochimera
 verkister - Xenochimera
+WTFismyname - Xenomorph Hybrid
```
Arokha: These are ckeys not keys, therefore need to be all lowercase with no formatting.

Arokha: These are ckeys not keys, therefore need to be all lowercase with no formatting.

Arokha:
Leshana and I are a bit "MMMMMMMMMMM" about the fact that there's a PR merge commit in this that should probably be removed. And 3 commits for 1 line is too much.

Mewchild:
You right

PR #3020, opened 2018-02-12
Added "Circuit Designer" scientist alt-title

Pear120:


code/game/jobs/job/z_all_jobs_vr.dm
```diff
@@ -47,3 +47,8 @@
 /datum/job/atmos
 	total_positions = 3
 	spawn_positions = 3
+	
+/datum/job/scientist
+	total_positions = 5
+	spawn_positions = 3
```
Arokha: Why positions defines?

Pear120: The file lacked a scientist entry. And seeing that every other entry in the file had position counts, I copied them over from scientist.dm.

MrSign:
Idunno about 'circuit designer'. It's a bit of a mouth full. You want to be able to splurt it out in a short sentence and circuit  designer isn't suited for it. Are you able to brainstorm for other titles? I've been thinking Electronics. Programmer. Coder. Something along the lines of programming circuits rather than designing it

Dankman:
Machinist maybe?

JoanRisu:
Electronics Engineer? Still a mouthful though.

Pear120:
Phoron Researcher, Xenobiologist, Mechatronic Engineer. Are these not mouthfuls? And having been piecing pre-made circuit modules together, I can say, that this is circuit design. Programming involves code. I'm sticking with Circuit Designer, because it's literally designing circuits.

Jwguy:
I will say that while I recognize that we ought be `moving forward with solutions` rather than `using past mistakes as examples`, that "Mechatronic Engineer"... or "Biomechanical Engineer", or half of the titles we're currently using, do no good service to the idea of holding back a title because it might be a mouthful. Those two examples are 7 and 9 syllable words... this one is a mere 5.

And even if it is a mouthful, is that really a reason to hold it back? It's a science title; mouthfuls are kind of their thing. Not to mention it is a simple and accurate description of one of the many rooms/jobs one can work with in Science.

Arokha:
I can definitely see both sides. Circuit Designer is a long name. But then again I can't think of anything more summary than that. I'm fine with the alt-title being added in general, because it does have a dedicated area in science for it, which is pretty much the minimum justification for an alt-title IMO.

Pear120:
I fail to see the problem of having a "long" job title on a HRP server. This isn't LRP where you need to shout "CHEF KILLING ME MAINT" in 5 seconds before he removes your headset. Maintenance Technician, Engine Specialist, alt-titles specify what you're good at instead of just being a generic Engineer or Scientist. Why would we NEED them to be short?

SpadesNeil:
Because long titles sound as stupid as they look.

Pear120:
Why not put that to a community vote then if there's no legitimate reason to deny the addition of professional-styled job titles aside from opinions? Especially since this incredibly simple pull request has sat around for days now with little to no progress.

Screemonster:
"why SHOULDN'T we add this" is not an adequate argument for adding anything. The burden is on the person making the request to justify it, not on others to justify refusal.

Pear120:
We have a job for each lab (except RnD, which seems to be just the generic Scientist lab) except for the Workshop. Would a Xenobiologist be expected to explain why they're not doing RnD?

Jwguy:
Well, we're not exactly discussing objective universal truths, either, here. There's nothing to prove, aside from comparing it to other station titles, which again, it meets or beats par in terms of length and syllables.

Personally, I don't really think it is that long or over the top, and as Aronai mentioned, it does seem as though it'd be appropriate, since the title covers a job actually performed in science, which has a whole lab devoted to it. Furthermore, at this point, the job is done; We're just more accounting for taste, it seems like... or perhaps distaste would be more appropriate.

SpadesNeil:
Pretty sure Polaris also removed a bunch of alt titles that for some reason we still have.

Pear120:
Polaris Github seems to still have the same alt-titles we do for our jobs:
Engineer: ("Maintenance Technician","Engine Technician","Electrician")
Scientist: ("Xenoarchaeologist", "Anomalist", "Phoron Researcher")
Roboticist: ("Biomechanical Engineer","Mechatronic Engineer")

Arokha:
Tired of seeing this on the tracker and it seems at least "okay" to me.

PR #3021, opened 2018-02-12
Fixes feeding people to mobs.

Verkister:
-Simpleanimals had a dumb override in polaris code to skip the attempt_vr part.
-Also fixed guard oties not switching to hold mode when eating 'human' prey as supposed.


Arokha:
There's a better way to do this (no offense) and I already typed it in some shadekin changes I'm making. I'll make a PR for it tomorrow (and include the otie fix).

PR #3026, opened 2018-02-13
Updates Fluff Ring

Mewchild:
Updates Vietsi fluff ring so gloves can be worn over it, and so it functions as a stamp (similar to normal signet ring).

### DIFF
--- a/code/modules/vore/fluffstuff/custom_clothes_vr.dm
+++ b/code/modules/vore/fluffstuff/custom_clothes_vr.dm
@@ -1697,19 +1697,6 @@ Departamental Swimsuits, for general use
 	icon_state = "vietsi_ring"
 	var/nameset = 0
 
-/obj/item/clothing/gloves/ring/seal/fluff/vietsi/attack_self(mob/user)
-	if(nameset)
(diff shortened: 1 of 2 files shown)


code/modules/vore/fluffstuff/custom_clothes_vr.dm
```diff
@@ -1689,15 +1689,15 @@ Departamental Swimsuits, for general use
...
+/obj/item/clothing/gloves/ring/seal/fluff/vietsi
 	name = "signet ring"
 	desc = "A signet ring carved from the bones of something long extinct, as a ward against bad luck."
 
 	icon = 'icons/vore/custom_clothes_vr.dmi'
 	icon_state = "vietsi_ring"
 	var/nameset = 0
 
-/obj/item/clothing/gloves/fluff/vietsi/attack_self(mob/user)
+/obj/item/clothing/gloves/ring/seal/fluff/vietsi/attack_self(mob/user)
```
Leshana: If you make it `/obj/item/clothing/gloves/ring/seal/signet/fluff/vietsi`  then you can eliminate this proc entirely as its identical to `/obj/item/clothing/gloves/ring/seal/signet/attack_self`

Issue #2999, opened 2018-02-08
Search and Rescue cannot spawn with medical loadout items


aztectornado:
#### Brief description of the issue

When building a loadout for an SaR character, medical loadout items do not spawn, indicating that the game thinks they aren't part of medical or something

#### What you expected to happen

I could spawn with white drop pouches, a KHI med uniform, or even a stethescope or med hud when making a loadout. The list of affected items is large, though the department jacket seems to be an exception, oddly enough?

#### What actually happened

I didn't get anything. :c 

#### Steps to reproduce
- Get some medical stuff in your loadout
- Spawn in as Search and Rescue
- Be sad that you can't get drop pouches

#### Code Revision
- f9feb190c403c43650399fb5a4558187d5d155be 
 
#### Anything else you may wish to add:
- This job still deserves hazard pay, imo

KonBear:
Same thing for explorers and science gear

Arokha:
Done on Pol

PR #3027, opened 2018-02-13
Adds Search and Rescue and Scientist to loadouts

aztectornado:
This will allow anything a Paramedic could spawn with to also be spawned with by Search and Rescue, fixing #2999. 

It also does the same with Scientist and Explorer, but leaves Pilot untouched because they're only civilian, and not medical or science. 

Now you can finally spawn with drop pouches as an SaR instead of waiting to raid the locker and praying the vest is still there and not given to some explorer! You can also totally wear the Star Trek jumpsuits down on the planets and be a walking reference!

### DIFF
--- a/code/modules/client/preference_setup/loadout/loadout_accessories.dm
+++ b/code/modules/client/preference_setup/loadout/loadout_accessories.dm
@@ -135,32 +135,38 @@
 /datum/gear/accessory/brown_vest
 	display_name = "webbing, brown"
 	path = /obj/item/clothing/accessory/storage/brown_vest
-	allowed_roles = list("Station Engineer","Atmospheric Technician","Chief Engineer","Security Officer","Detective","Head of Security","Warden","Paramedic","Chief Medical Officer","Medical Doctor")
+	//VOREStation Edit - Adds Search and Rescue to allowed roles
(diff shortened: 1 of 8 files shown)


Cyantime:
Should probably be done on Polaris

aztectornado:
Polaris doesn't have SaR and Explorers as part of their regular game, do they?

Cyantime:
They do. The jobs were added for their new map, Southern Cross.

aztectornado:
Well heck. I guess it does need to go upstream, but this could totally hold us over until the next merge, and does the vr_ specific things anyway. 

I'll make one for upstream that's specific to them.

Arokha:
> They do. The jobs were added for their new map, Southern Cross.

This change upstream will not pass their Travis checks, as one of the Travis checks makes sure all checked files can compile with no map checked, which will fail there. The southern cross jobs and such are directly included in the dme for us so we can use them, and it passes here.

aztectornado:
Wait a sec, it looks like they already have most of these changes upstream. 

Sans the vr specific ones, anyway.

Arokha:
Probably just wait for a Polaris sync then, and reduce your changes to just the VR ones.

aztectornado:
Lemme do some time travel then

aztectornado:
Actually, do you want these both in a single commit?

aztectornado:
Closing this, will reopen with new changes

PR #3030, opened 2018-02-15
Make light switches constructable

Leshana:
Adds constructable light switches using Brown's construction stage icons.


Very-Soft:
\o/ yay!!

Arokha:
Interesting new system, we'll see how it goes.

PR #3042, opened 2018-02-15
Fix for crushing and new descriptions for Taur stompies

Heroman3003:
Fixed crushing not dealing any damage
Insect, Hoofed, Tentacled and Slug taurs now have more appropriate descriptions for stomping (horse/cowtaur grab one is temporary, open to suggestions on that one)


Arokha:
I don't like how this file is looking, so I'm going to merge this, and then just rewrite this using your quotes.

PR #3043, opened 2018-02-16
Refactor step code, and taur messages

Arokha:
Move all those messages onto the taur bodies themselves. Now they're easy to define and that file isn't quite as much as a horrorshow.

Also fixed a bug preventing harm steps from doing full damage on walk intent, and the damages logged being inaccurate.

### DIFF
--- a/code/modules/vore/appearance/sprite_accessories_vr.dm
+++ b/code/modules/vore/appearance/sprite_accessories_vr.dm
@@ -1056,6 +1056,33 @@
 	do_colouration = 1 // Yes color, using tail color
 	color_blend_mode = ICON_MULTIPLY  // The sprites for taurs are designed for ICON_MULTIPLY
 
+	//Could do nested lists but it started becoming a nightmare. It'd be more fun for lookups of a_intent and m_intent, but then subtypes need to
+	//duplicate all the messages, and it starts getting awkward. These are singletons, anyway!
+
+	//Messages to owner when stepping on/over
+	var/msg_owner_help_walk		= "You carefully step over %prey."
+	var/msg_owner_help_run		= "You carefully step over %prey."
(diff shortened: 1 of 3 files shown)


code/modules/vore/appearance/sprite_accessories_vr.dm
```diff
@@ -1056,125 +1056,302 @@
...
+	msg_owner_help_run = "You carefully slither around %prey."
+	msg_prey_help_run = "%owner's huge tail slithers past beside you!"
+
+	msg_owner_disarm_run = "Your tail slides over %prey, pushing them down to the ground!"
+	msg_prey_disarm_run = "%owner's tail slides over you, forcing you down to the ground!"
+
+	msg_owner_disarm_walk = "You push down on %prey with your tail, pinning them down under you!"
+	msg_prey_disarm_walk = "%owner pushes down on you with their tail, pinning you down below them!"
+
+	msg_owner_harm_run = "Your heavy tail carelessly slides past %prey, crushing them!"
```
Sebbe9123: Doesn't going "past" someone imply slithering around them (missing) instead of onto them or over them?(hitting/crushing)

Heroman3003: I think 'past' when applied to sliding can be used that way.

Sebbe9123: Hmm... Still feel like "carelessly sliding over %prey" works better, might be because of English being second language. Up to @Arokha of course.

Screemonster: "Sliding" is not the correct verb tense to use in this sentence. "Slides" is.
Substituting "over" for "past" is fine though.

Arokha: I mean, I didn't write any of these messages. If someone wants to "grammarize" them, they can go for it. These are the *existing* taur messages, unaltered, just moved to another file.

PR #3051, opened 2018-02-17
Fix pipes in engineering over openspace

Arokha:
These, by putting plating under them.

image


Arokha:
Leshana says they'll fix this the real way later.

PR #3059, opened 2018-02-18
Ringfix v2

Mewchild:
Replaces a Removed Proc- which broke a fluff item


Mewchild:
For clarification, removing the proc as suggested in the last update to this made it so I wasn't able to change the ring's name at all. It's the same proc as used by the base ring/signet, of which this is a child, but BYOND is a weird place

Leshana:
That is because you didn't change the type path.  You should remove the proc AND change the type path as suggested to `/obj/item/clothing/gloves/ring/seal/signet/fluff/vietsi`

PR #3061, opened 2018-02-18
Smol Sync

Arokha:
* Bay survival boxes
* Borg intent fix
* Fix suit coolers not working in hot environments
* Power monitoring consoles now find all things


Leshana:
ok

PR #3071, opened 2018-02-19
Make gas turbine constructable and operational

Leshana:
* Make gas turbine constructable, as well as modernizing its code. Borrowed some from /tg
* Converted the gas turbine control computer to nano ui.

### DIFF
--- a/code/_helpers/unsorted.dm
+++ b/code/_helpers/unsorted.dm
@@ -509,6 +509,16 @@ Turf and target are seperate in case you want to teleport some distance from a t
 //		mob_list.Add(M)
 	return moblist
 
+// Format a power value in W, kW, MW, or GW.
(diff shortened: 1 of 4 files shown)


Cyantime:
More engines for the engine pile!

Jakksergal:
Would it be possible to do this with the pipe-based variant as well?

Screemonster:
The codebases that have this don't have meltable walls and airlocks - what sort of temperatures are needed for it to be useful?

Leshana:
@Jakksergal Probably.  But I was able to leverage mostly existing code and /tg's working turbines to do this quickly.   Plus the sprites aren't terrible.   The pipe based sprites are super bleh.
@Screemonster That sounds like a great thing for SCIENCE to find out, through RESEARCH!

PR #3078, opened 2018-02-19
Adds "monster tamer" drink, plus some chimera tweaks.

Screemonster:
- Adds "monster tamer", a drink made by mixing animal protein and whiskey. Combination of good ol' meat with a neural depressant calms them down from feralness faster (provided they're not still in the state that MADE them feral)

- Renamed "begin reconstitute form" into just "reconstitute form".

- Moved their revive into its own proc so it can be called without the verb.

- Added a chemical reaction (5u monster tamer, 5u clonexadone) that will do the above. Will give feedback as to whether it was successful or not, ie. whether they have nutriment in them, aren't reviving already. Has no effect on live ones. Mainly exists so the chimera in deadchat don't have to keep returning to their body and mashing the button to see if they've been injected with enough nutriment yet.

- Made them NO_INFECT, removed Purge Impurities. (no slyly removing toxloss with a verb that's meant to reflect their strong immune systems. They have actually strong immune systems now)

- Moved handle_feral calls under a spawn(0) because the first thing that happens in that proc after the "am I handling hallucinations already" check is a sleep() and that doesn't seem like a good thing to have happening in a critter's life() ticks. Removed redundant if(!handling_hal) checks because the first thing in the proc they call is that exact check.

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/station_special_abilities_vr.dm
+++ b/code/modules/mob/living/carbon/human/species/station/station_special_abilities_vr.dm
@@ -1,5 +1,5 @@
 /mob/living/carbon/human/proc/begin_reconstitute_form() //Scree's race ability.in exchange for: No cloning.
-	set name = "Begin Reconstitute Form"
+	set name = "Reconstitute Form"
 	set category = "Abilities"
 
 	if(world.time < last_special)
@@ -9,41 +9,18 @@
 
 	var/confirm = alert(usr, "Are you sure you want to completely reconstruct your form? This process can take up to twenty minutes, depending on how hungry you are, and you will be unable to move.", "Confirm Regeneration", "Yes", "No")
 	if(confirm == "Yes")
-		var/mob/living/carbon/human/C = src
-		var/nutrition_used = C.nutrition/2
+		chimera_regenerate()
 
-		if(C.reviving == 1) //If they're already unable to
-			C << "You are already reconstructing, or your body is currently recovering from the intense process of your previous reconstitution."
-			return
+/mob/living/carbon/human/proc/chimera_regenerate()
+	var/mob/living/carbon/human/C = src
+	var/nutrition_used = C.nutrition/2
 
-		if(C.stat == DEAD) //Uh oh, you died!
-			if(C.hasnutriment()) //Let's hope you have nutriment in you.... If not
-				var/time = (240+960/(1 + nutrition_used/75))
-				C.weakened = 10000 //Since it takes 1 tick to lose one weaken. Due to prior rounding errors, you'd sometimes unweaken before regenning. This fixes that.
-				C.reviving = 1
-				C.canmove = 0 //Make them unable to move. In case they somehow get up before the delay.
-				C << "You begin to reconstruct your form. You will not be able to move during this time. It should take aproximately [round(time)] seconds."
-				C.does_not_breathe = 1 //effectively makes them spaceworthy while regenning
+	if(C.reviving == 1) //If they're already unable to
+		C << "You are already reconstructing, or your body is currently recovering from the intense process of your previous reconstitution."
+		return
 
-				for(var/obj/item/organ/E in C.bad_external_organs)
-					var/obj/item/organ/external/affected = E
-					for(var/datum/wound/W in affected.wounds) // Fix internal bleeds at the start of the rejuv process.
-						if(istype(W, /datum/wound/internal_bleeding))
-							affected.wounds -= W
-							affected.update_damages()
-
-				spawn(time SECONDS)
-					if(C) //Runtime prevention.
-						C << "<span class='notice'>Consciousness begins to stir as your new body awakens, ready to hatch..</span>"
-						C.verbs += /mob/living/carbon/human/proc/hatch
-						return
-					else
-						return //Something went wrong.
-			else //Dead until nutrition injected.
-				C << "Your body is too damaged to regenerate without additional nutrients to feed what few living cells remain."
-				return
-
-		else if(C.stat != DEAD) //If they're alive at the time of reviving.
+	if(C.stat == DEAD) //Uh oh, you died!
+		if(C.hasnutriment()) //Let's hope you have nutriment in you.... If not
 			var/time = (240+960/(1 + nutrition_used/75))
 			C.weakened = 10000 //Since it takes 1 tick to lose one weaken. Due to prior rounding errors, you'd sometimes unweaken before regenning. This fixes that.
 			C.reviving = 1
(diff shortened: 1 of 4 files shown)


Screemonster:
Also just spotted and fixed a bug that popped up in testing and for the life of me I couldn't reproduce deliberately - namely that really minuscule (but still positive) values for feral were halting life, as handle_feral() had a sleep(rand(200,500)/(feral/10)) statement in it.

So a feral value of 4E-6 (which is what it had when it happened) was making it sleep _for goddamn ever_.

Have made handle_feral outright reject feral values under 10 so the maximum delay is 500.

code/modules/mob/living/carbon/human/species/station/station_special_abilities_vr.dm
```diff
@@ -209,7 +158,7 @@
 	if(handling_hal) return //avoid conflict with actual hallucinations
 	handling_hal = 1
 
-	if(client && feral) // largely a copy of handle_hallucinations() without the fake attackers. Unlike hallucinations, only fires once - if they're still feral they'll get hit again anyway.
+	if(client && feral >= 10) // largely a copy of handle_hallucinations() without the fake attackers. Unlike hallucinations, only fires once - if they're still feral they'll get hit again anyway.
 		sleep(rand(200,500)/(feral/10))
```
Arokha: Do not sleep in Life() ever. It sleeps the entire mob controller. When you sleep, the mob controller will bail, and stonedmc master will sigh at poor code and pause the entire controller until your sleep is done.

I cannot merge something that sleeps in Life()

It's best if you come up with a way that doesn't involve sleep, but the alternatives are using spawn, or `set waitfor = 0` which means the proc does above does NOT wait for this proc. Be wary of doing this, because the second you sleep, your proc will return `.` at that exact moment, and there's no potential to return other data (without shenanigans like mutual object references).

aztectornado:
Damn but this looks cool. Almost makes me wanna go Xenochimera, but I'd have to think of a character to play for that. 

Does the Lazarus thing work on simple mobs too, or is it just a xeno thing?

code/modules/mob/living/carbon/human/species/station/station_special_abilities_vr.dm
Screemonster: The handle_feral proc is called under a spawn. Will that still wreck the mob controller?

If so, handle_hallucinations still needs to be fixed as well as it does the exact same thing and is called directly by life() - again, under a spawn(0).

Screemonster:
Got rid of the sleep() and replaced it with a spawn(). Since this code was copypasted from hallucination code, hallucination code could probably do with an overhaul.

While I was in there, I also got rid of the client-view-spinning thing since it broke every sprite on the screen, and replaced it with SCARY VOICES. Like this:

image

Screemonster:
The "discount lazarus" reaction only works in human mobs that have the "reconstitute form" verb, and has the same effect as pressing the verb and hitting yes on the confirmation box. It doesn't work on simple mobs.

code/modules/mob/living/carbon/human/species/station/station_special_abilities_vr.dm
```diff
@@ -9,119 +9,68 @@
...
 	if(confirm == "Yes")
-		var/mob/living/carbon/human/C = src
-		var/nutrition_used = C.nutrition/2
+		chimera_regenerate()
 
-		if(C.reviving == 1) //If they're already unable to
-			C << "You are already reconstructing, or your body is currently recovering from the intense process of your previous reconstitution."
-			return
+/mob/living/carbon/human/proc/chimera_regenerate()
+	var/mob/living/carbon/human/C = src
```
Arokha: Why cast this? There's no need. If this is defined on human, you can just refer to everything with implicit things. Like one line below, just doing nutrition.

code/modules/mob/living/carbon/human/species/station/station_special_abilities_vr.dm
```diff
@@ -9,119 +9,68 @@
...
-						C << "<span class='notice'>Consciousness begins to stir as your new body awakens, ready to hatch..</span>"
-						C.verbs += /mob/living/carbon/human/proc/hatch
-						return
-					else
-						return //Something went wrong.
-			else //Dead until nutrition injected.
-				C << "Your body is too damaged to regenerate without additional nutrients to feed what few living cells remain."
-				return
+	if(C.reviving == 1) //If they're already unable to
+		C << "You are already reconstructing, or your body is currently recovering from the intense process of your previous reconstitution."
```
Arokha: to_chat(src,"Words"), here and everywhere else

code/modules/mob/living/carbon/human/species/station/station_special_abilities_vr.dm
```diff
@@ -9,119 +9,68 @@
...
 			C.canmove = 0 //Make them unable to move. In case they somehow get up before the delay.
 			C << "You begin to reconstruct your form. You will not be able to move during this time. It should take aproximately [round(time)] seconds."
 			C.does_not_breathe = 1 //effectively makes them spaceworthy while regenning
 
+			for(var/obj/item/organ/E in C.bad_external_organs)
+				var/obj/item/organ/external/affected = E
+				for(var/datum/wound/W in affected.wounds) // Fix internal bleeds at the start of the rejuv process.
+					if(istype(W, /datum/wound/internal_bleeding))
+						affected.wounds -= W
+						affected.update_damages()
```
Arokha: I dunno about this. It seems... shady. Those datums may or may not get GC'd, I don't know if they're referenced anywhere else. Maybe consider just healing them?

code/modules/mob/living/carbon/human/species/station/station_special_abilities_vr.dm
```diff
@@ -9,119 +9,68 @@
...
+
 			spawn(time SECONDS)
-				if(C.stat != DEAD) //If they're still alive after regenning.
+				if(C) //Runtime prevention.
 					C << "<span class='notice'>Consciousness begins to stir as your new body awakens, ready to hatch..</span>"
 					C.verbs += /mob/living/carbon/human/proc/hatch
```
Arokha: What if they are revived via other means in the meantime?

code/modules/mob/living/carbon/human/species/station/station_special_abilities_vr.dm
```diff
@@ -9,119 +9,68 @@
...
-	C << "<span class='notice'>You have finished purging your body of impurities.</span>"
-
-	spawn(300 SECONDS) //5 minute wait until you can purge or regenerate again.
-		C.reviving = 0
+					return //Something went wrong.
+		else //Dead until nutrition injected.
+			C << "Your body is too damaged to regenerate without additional nutrients to feed what few living cells remain."
+			return
 
+	else if(C.stat != DEAD) //If they're alive at the time of reviving.
```
Arokha: Misleading comment, as this is not 'reviving' anyone.

code/modules/mob/living/carbon/human/species/station/station_special_abilities_vr.dm
```diff
@@ -209,188 +158,189 @@
...
-					if(6) src << 'sound/effects/Glassbr3.ogg'
-					if(7) src << 'sound/machines/twobeep.ogg'
-					if(8) src << 'sound/machines/windowdoor.ogg'
-					if(9)
-						//To make it more realistic, I added two gunshots (enough to kill)
-						src << 'sound/weapons/Gunshot.ogg'
-						spawn(rand(10,30))
+	if(client && feral >= 10) // largely a copy of handle_hallucinations() without the fake attackers. Unlike hallucinations, only fires once - if they're still feral they'll get hit again anyway.
+		spawn(rand(200,500)/(feral/10))
+			var/halpick = rand(1,100)
```
Arokha: What if they aren't feral anymore when this procs?

code/modules/mob/living/carbon/human/species/station/station_special_abilities_vr.dm
```diff
@@ -209,188 +158,189 @@
...
+						hal_screwyhud = 0
+				if(16 to 25) //10% chance
+					//Strange items
+					//src << "Traitor Items"
+					if(!halitem)
+						halitem = new
+						var/list/slots_free = list(ui_lhand,ui_rhand)
+						if(l_hand) slots_free -= ui_lhand
+						if(r_hand) slots_free -= ui_rhand
+						if(istype(src,/mob/living/carbon/human))
```
Arokha: Gross istype of src.

code/modules/mob/living/carbon/human/species/station/station_special_abilities_vr.dm
```diff
@@ -209,188 +158,189 @@
...
+							var/mob/living/carbon/human/H = src
+							if(!H.belt) slots_free += ui_belt
+							if(!H.l_store) slots_free += ui_storage1
+							if(!H.r_store) slots_free += ui_storage2
+						if(slots_free.len)
+							halitem.screen_loc = pick(slots_free)
+							halitem.layer = 50
```
Arokha: Unlikely this will show up on the HUD at all, considering the HUD has a very high plane and you're not setting any on this. You should use hud_layerise()

code/modules/mob/living/carbon/human/species/station/station_special_abilities_vr.dm
```diff
@@ -209,188 +158,189 @@
...
+						if(possible_points.len)
+							var/turf/simulated/floor/target = pick(possible_points)
+
+							switch(rand(1,3))
+								if(1)
+									//src << "Space"
+									halimage = image('icons/turf/space.dmi',target,"[rand(1,25)]",TURF_LAYER)
+								if(2)
+									//src << "Fire"
+									halimage = image('icons/effects/fire.dmi',target,"1",TURF_LAYER)
```
Arokha: The flick_overlay proc can replace most of this if you have suitable animations to go with. Fire would probably work. Not necessary.

code/modules/mob/living/carbon/human/species/station/station_special_abilities_vr.dm
```diff
@@ -209,188 +158,189 @@
...
+								'sound/hallucinations/look_up1.ogg', 'sound/hallucinations/look_up2.ogg', 'sound/hallucinations/over_here1.ogg', 'sound/hallucinations/over_here2.ogg', 'sound/hallucinations/over_here3.ogg',\
+								'sound/hallucinations/turn_around1.ogg', 'sound/hallucinations/turn_around2.ogg', 'sound/hallucinations/veryfar_noise.ogg', 'sound/hallucinations/wail.ogg')
+							src << pick(creepyasssounds)
+				if(56 to 60) //5% chance
+					//Flashes of danger
+					//src << "Danger Flash"
+					if(!halbody)
+						var/list/possible_points = list()
+						for(var/turf/simulated/floor/F in view(src,world.view))
+							possible_points += F
```
Arokha: This is scary. It'd be a lot cheaper to just pick an x/y randomly within world.view until you found what you want than processing a whole full-screen view most of the time (making sure to bail if you fail too many times (or world.view*world.view times, since that'd be the whole area)).

code/modules/mob/living/carbon/human/species/station/station_special_abilities_vr.dm
```diff
@@ -209,188 +158,189 @@
...
+							if(client) client.images += halbody
+							spawn(rand(50,80)) //Only seen for a brief moment.
+								if(client) client.images -= halbody
+								halbody = null
+				if(61 to 85) //25% chance
+					//food
+					if(!halbody)
+						var/list/possible_points = list()
+						for(var/turf/simulated/floor/F in view(src,world.view))
+							possible_points += F
```
Arokha: Same

code/modules/reagents/Chemistry-Recipes_vr.dm
```diff
@@ -47,6 +47,22 @@
 	s.start()
 	holder.clear_reagents()
 
+/datum/chemical_reaction/xenolazarus
+	name = "Discount Lazarus Injector"
```
Arokha: I'm not sure I'd name a REACTION an INJECTOR.

code/modules/mob/living/carbon/human/species/station/station_special_abilities_vr.dm
Screemonster: In practical terms? It'll take longer than spawn(500) for them to tick 10 feralness down normally, but can add an "if(!feral) return" easily enough though to be sure.

code/modules/mob/living/carbon/human/species/station/station_special_abilities_vr.dm
Screemonster: Any suggestions how? This is how it's done in the equivalent changeling power.

code/modules/mob/living/carbon/human/species/station/station_special_abilities_vr.dm
Screemonster: Now _that's_ an edge-case I don't think either me or cameron accounted for as I can't remember if defibs were around at the time.

code/modules/mob/living/carbon/human/species/station/station_special_abilities_vr.dm
Screemonster: Hallucination code.

code/modules/mob/living/carbon/human/species/station/station_special_abilities_vr.dm
Screemonster: Hallucination code.

code/modules/mob/living/carbon/human/species/station/station_special_abilities_vr.dm
Screemonster: Hallucination code.

code/modules/mob/living/carbon/human/species/station/station_special_abilities_vr.dm
Screemonster: Hallucination code.

code/modules/mob/living/carbon/human/species/station/station_special_abilities_vr.dm
Screemonster: Hallucination code.

Screemonster:
Moved the actual hatching part into its own proc rather than having redundant parts everywhere. Also fixed an edge case where a dead 'mera that was regenerating gets revived by whatever means - if this happens, the proc just aborts and resets the cooldown when the timer is up. Also put a sanity check in handle_feral just in _case_ they unferal between the proc being called and the sleep taking effect.

Updated a whole bunch of "C <<" into "to_chat" and got rid of redundant var/C = src shenanigans.

PR #3081, opened 2018-02-19
Allow shadekin phase-shift noms

Arokha:
Allows shadekin to phase in on people, and if they are set to be a dropnom pred in the vore panel, they'll belly the person they land on.


Very-Soft:
This pleases me.

Screemonster:
and now to make the actual mobs occasionally do this as a behaviour :3

PR #3090, opened 2018-02-21
Make spy bugs spawn spy cameras

Arokha:
They are just using New() from their parent which spawned a normal camera. Now they store a type.


Arokha:
pol merged, will just get from there, keep history shorter.

PR #3091, opened 2018-02-21
Adds a toggle to the bluespace jumpsuit

Cameron653:
Makes the bluespace jumpsuit showing bulges toggle able or not.

### DIFF
--- a/code/modules/clothing/under/miscellaneous_vr.dm
+++ b/code/modules/clothing/under/miscellaneous_vr.dm
@@ -47,7 +47,7 @@
 /obj/item/clothing/under/bluespace/verb/toggle_fibers()
 		set category = "Object"
 		set name = "Adjust fibers"
-		set desc = "Adjust your suit fibers. This makes it so your stomach(s) will show or not.
+		set desc = "Adjust your suit fibers. This makes it so your stomach(s) will show or not."
 		set src in usr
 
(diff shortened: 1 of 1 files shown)


Cameron653:
Accidentally deleted a " while doing a quick addition of a description at the last second.

code/modules/clothing/under/miscellaneous_vr.dm
```diff
@@ -42,13 +42,36 @@
...
+		set src in usr
+
+		adjust_fibers(usr)
+		..()
+
+/obj/item/clothing/under/bluespace/proc/adjust_fibers(mob/user)
+	if(hides_bulges == FALSE)
+		hides_bulges = TRUE
+		to_chat(user, "You tense the suit fibers, hiding your stomach(s).")
+	else
```
Arokha: It doesn't really matter, and you don't need to change it, but for future reference you can do this entire proc in two lines:
```
hides_bulges = !hides_bulges //This inverts the setting (setting it to 'not' what it is right now)
to_chat(user, "You [hides_bulges ? "tense" : "relax"] the suit fibers, [hides_bulges ? "hiding" : "showing"] your stomach(s).")
```
It technically has one more if() statement than yours, since both ternary operators are just as expensive as if statements. You could reduce it to the same cost by just saying 'adjust' the suit fibers.

PR #3101, opened 2018-02-22
[Pls Review] Xenochimera Cleanup

Arokha:
Posted for review.

### DIFF
--- a/code/_helpers/unsorted.dm
+++ b/code/_helpers/unsorted.dm
@@ -1467,11 +1467,21 @@ var/mob/dview/dview_mob = new
 		if(337.5)
 			return "North-Northwest"
 
+//This is used to force compiletime errors if you incorrectly supply variable names. Crafty!
+#define NAMEOF(datum, X) (#X || ##datum.##X)
 
+//Creates a callback with the specific purpose of setting a variable
+#define VARSET_CALLBACK(datum, var, var_value) CALLBACK(GLOBAL_PROC, /proc/___callbackvarset, weakref(##datum), NAMEOF(##datum, ##var), ##var_value)
 
-
-
-
-
-
-
+//Helper for the above
+/proc/___callbackvarset(list_or_datum, var_name, var_value)
+	if(isweakref(list_or_datum))
+		var/weakref/wr = list_or_datum
+		list_or_datum = wr.resolve()
+	if(!list_or_datum)
+		return
+	if(length(list_or_datum))
+		list_or_datum[var_name] = var_value
+		return
+	var/datum/D = list_or_datum
+	D.vars[var_name] = var_value
--- a/code/modules/mob/living/carbon/human/species/station/station_special_abilities_vr.dm
+++ b/code/modules/mob/living/carbon/human/species/station/station_special_abilities_vr.dm
@@ -1,137 +1,147 @@
-/mob/living/carbon/human/proc/begin_reconstitute_form() //Scree's race ability.in exchange for: No cloning.
+/mob/living/carbon/human/proc/reconstitute_form() //Scree's race ability.in exchange for: No cloning.
 	set name = "Reconstitute Form"
 	set category = "Abilities"
 
-	if(world.time < last_special)
-		return
-
-	last_special = world.time + 50 //To prevent button spam.
-
+	// Sanity is mostly handled in chimera_regenerate()
+	
 	var/confirm = alert(usr, "Are you sure you want to completely reconstruct your form? This process can take up to twenty minutes, depending on how hungry you are, and you will be unable to move.", "Confirm Regeneration", "Yes", "No")
 	if(confirm == "Yes")
 		chimera_regenerate()
 
 /mob/living/carbon/human/proc/chimera_regenerate()
-	var/nutrition_used = nutrition/2
-
-	if(reviving == TRUE) //If they're already unable to
-		to_chat(src, "You are already reconstructing, or your body is currently recovering from the intense process of your previous reconstitution.")
-		return
-
-	if(stat == DEAD) //Uh oh, you died!
-		if(hasnutriment()) //Let's hope you have nutriment in you.... If not
-			var/time = (240+960/(1 + nutrition_used/75))
-			reviving = TRUE
-			to_chat(src, "You begin to reconstruct your form. You will not be able to move during this time. It should take aproximately [round(time)] seconds.")
-			//don't need all the weakened, does_not_breathe, canmove, heal IB crap here like you do for live ones'cause they're DEAD.
-
-			spawn(time SECONDS)
-				if(src) //Runtime prevention.
-					if (stat == DEAD) // let's make sure they've not been defibbed or whatever
-						to_chat(src, "<span class='notice'>Consciousness begins to stir as your new body awakens, ready to hatch.</span>")
-						verbs += /mob/living/carbon/human/proc/hatch
-						return
-					else // their revive got aborted by being made not-dead. Remove their cooldown.
-						to_chat(src, "<span class='notice'>Your body has recovered from its ordeal, ready to regenerate itself again.</span>")
-						reviving = FALSE
-						return
-				else
-					return //Something went wrong.
-		else //Dead until nutrition injected.
-			to_chat(src, "Your body is too damaged to regenerate without additional nutrients to feed what few living cells remain.")
+	//If they're already regenerating
+	switch(reviving)
+		if(REVIVING_NOW)
+			to_chat(src, "You are already reconstructing, just wait for the reconstruction to finish!")
+			return
+		if(REVIVING_DONE)
+			to_chat(src, "Your reconstruction is done, but you need to hatch now.")
+			return
+		if(REVIVING_COOLDOWN)
+			to_chat(src, "You can't use that ability again so soon!")
 			return
 
-	else if(stat != DEAD) //If they're alive at the time of regenerating.
-		var/time = (240+960/(1 + nutrition_used/75))
-		weakened = 10000 //Since it takes 1 tick to lose one weaken. Due to prior rounding errors, you'd sometimes unweaken before regenning. This fixes that.
-		reviving = TRUE
-		canmove = 0 //Make them unable to move. In case they somehow get up before the delay.
-		to_chat(src, "You begin to reconstruct your form. You will not be able to move during this time. It should take aproximately [round(time)] seconds.")
-		does_not_breathe = 1 //effectively makes them spaceworthy while regenning
+	var/nutrition_used = nutrition * 0.5
+	var/time = (240+960/(1 + nutrition_used/75))
 
-		spawn(time SECONDS)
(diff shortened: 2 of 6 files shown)


code/modules/mob/living/carbon/human/species/station/station_special_abilities_vr.dm
```diff
@@ -80,55 +77,55 @@
...
 			chimera_hatch()
 			visible_message("<span class='danger'><p><font size=4>The dormant husk of [src] bursts open, revealing a new, intact copy in the pool of viscera.</font></p></span>") //Bloody hell...
-			return
-		else
-			return //Runtime prevention.
 
 /mob/living/carbon/human/proc/chimera_hatch()
-	nutrition -= nutrition/2 //Cut their nutrition in half.
-	var/old_nutrition = nutrition //Since the game is being annoying.
+	last_special = world.time + 1 HOUR
```
Screemonster: The "reviving" var was what was used for the cooldown, so if you're using the last_special timer instead, remember to set it to false again on hatching.

code/modules/mob/living/carbon/human/species/station/station_special_vr.dm
```diff
@@ -56,142 +56,189 @@
...
+		//If they lose enough health to hit softcrit, handle_shock() will keep resetting this. Otherwise, pissed off critters will lose shock faster than they gain it.
+		H.shock_stage = max(H.shock_stage-(H.feral/20), 0)
+
+		//Don't bother doing heavy lifting if we weren't going to give emotes anyway.
+		if(!prob(1))
+
+			//This is basically the 'lite' version of the below block.
+			var/list/nearby = H.living_mobs(world.view)
+
+			//Always handle feral if nobody's around.
```
Screemonster: Shouldn't do any feral stuff if they're in the dark, unless you wanna make handle_feral have a light level check.

code/modules/mob/living/carbon/human/species/station/station_special_vr.dm
```diff
@@ -56,142 +56,189 @@
...
+		if(H.jitteriness)
+			H.feral = max(H.feral, H.jitteriness-100)
+
+		//Still hungry? More feral.
+		if(H.feral + H.nutrition < 150)
+			H.feral++
+		else
+			H.feral = H.feral
+
+		//Did we just finish being feral?
```
Screemonster: Can't see feral being decremented. "If (hungry || shock || jittery) do the above things, else feral--" maybe?

Arokha: Typo, it was intended to be on L167

code/modules/mob/living/carbon/human/species/station/station_special_abilities_vr.dm
Arokha: Meant to, somehow removed it.

code/modules/mob/living/carbon/human/species/station/station_special_abilities_vr.dm
```diff
@@ -1,72 +1,69 @@
...
+			to_chat(src, "<span class='warning'>Your body is too damaged to regenerate without additional nutrients to feed what few living cells remain.</span>")
 
-	else if(stat != DEAD) //If they're alive at the time of regenerating.
-		var/time = (240+960/(1 + nutrition_used/75))
-		weakened = 10000 //Since it takes 1 tick to lose one weaken. Due to prior rounding errors, you'd sometimes unweaken before regenning. This fixes that.
-		reviving = TRUE
-		canmove = 0 //Make them unable to move. In case they somehow get up before the delay.
+	//Clicked regen while NOT dead
+	else
 		to_chat(src, "You begin to reconstruct your form. You will not be able to move during this time. It should take aproximately [round(time)] seconds.")
```
Screemonster: Need to set reviving = TRUE here.

code/modules/mob/living/carbon/human/species/station/station_special_abilities_vr.dm
```diff
@@ -80,55 +77,56 @@
...
-					chimera_hatch()
-					visible_message("<span class='danger'><p><font size=4>The lifeless husk of [src] bursts open, revealing a new, intact copy in the pool of viscera.</font></p></span>") //Bloody hell...
-					brainloss += 10 //Reviving from dead means you take a lil' brainloss on top of whatever was healed in the revive.
-					return
-				else
-					return //Runtime prevention
-			else //don't have nutriment to hatch! Or you somehow died in between completing your revive and hitting hatch.
-				to_chat(src, "Your body was unable to regenerate, what few living cells remain require additional nutrients to complete the process.")
-				reviving = FALSE // so they can try again when they're given a kickstart
+	if(confirm == "Yes" && reviving)
```
Screemonster: If they're NOT reviving, the verb should probably be removed. That said, there shouldn't be any circumstances where they GET the verb if they're not reviving.

code/modules/mob/living/carbon/human/species/station/station_special_abilities_vr.dm
```diff
@@ -80,55 +77,56 @@
...
+				chimera_hatch()
+				visible_message("<span class='danger'><p><font size=4>The lifeless husk of [src] bursts open, revealing a new, intact copy in the pool of viscera.</font></p></span>") //Bloody hell...
+				brainloss += 10 //Reviving from dead means you take a lil' brainloss on top of whatever was healed in the revive.
 				return
 
-		else if(stat != DEAD) //If they're alive at the time of regenerating.
+			//Don't have nutriment to hatch! Or you somehow died in between completing your revive and hitting hatch.
+			else
+				to_chat(src, "Your body was unable to regenerate, what few living cells remain require additional nutrients to complete the process.")
+				reviving = FALSE //So they can try again when they're given a kickstart
```
Screemonster: Remove verb here.

code/modules/mob/living/carbon/human/species/station/station_special_abilities_vr.dm
```diff
@@ -80,55 +77,56 @@
...
-				reviving = FALSE // so they can try again when they're given a kickstart
+	if(confirm == "Yes" && reviving)
+
+		//Dead when hatching
+		if(stat == DEAD)
+			//Check again for nutriment (necessary?)
+			if(hasnutriment())
+				chimera_hatch()
+				visible_message("<span class='danger'><p><font size=4>The lifeless husk of [src] bursts open, revealing a new, intact copy in the pool of viscera.</font></p></span>") //Bloody hell...
+				brainloss += 10 //Reviving from dead means you take a lil' brainloss on top of whatever was healed in the revive.
```
Screemonster: Apparently just setting the brainloss var doesn't work, just VV'd after reviving from dead and it was 0. Might need to call adjustBrainLoss() instead.

code/modules/mob/living/carbon/human/species/station/station_special_abilities_vr.dm
```diff
@@ -80,55 +77,56 @@
...
+	var/old_nutrition = nutrition * 0.5
+	var/braindamage = brainloss * 0.5
+
+	//I did have special snowflake code, but this is easier.
+	revive()
 	mutations.Remove(HUSK)
 	nutrition = old_nutrition
-	brainloss = braindamage //Gives them half their prior brain damage.
-	update_canmove()
+	brainloss = braindamage
```
Screemonster: Changelog shows brainloss was moved to the brain organ. Use setBrainLoss(braindamage).

Arokha:
Will squash after review.

code/modules/mob/living/carbon/human/species/station/station_special_abilities_vr.dm
Screemonster: Uh. Just had a check of what last_special is used for elsewhere. This will render the chimera unable to use the resist button for an hour.

code/_helpers/unsorted.dm
```diff
@@ -1466,11 +1466,17 @@ var/mob/dview/dview_mob = new
...
 
-
-
-
-
-
-
-
-
+#define NAMEOF(datum, X) (#X || ##datum.##X)
```
Leshana: Please add comment:
This macro is used for compile checked var-name-as-string references.   Normally if a var name is used as a string (for use in `vars` or such), A typo in the string (or the actual var being renamed in a later commit) will not generate a compile error, you'll just get runtimes.  Use this macro to make the compiler check that your string is a valid var name.

code/modules/mob/living/carbon/human/species/station/station_special_abilities_vr.dm
```diff
@@ -80,55 +78,61 @@
...
+	//Unfreeze some things
+	does_not_breathe = FALSE
+	update_canmove()
+	weakened = 2
+
+	//Visual effects
+	var/T = get_turf(src)
+	new /obj/effect/gibspawner/human/scree(T)
+
+	schedule_callback_in(1 HOUR, VARSET_CALLBACK(src, reviving, FALSE))
```
Screemonster: This'll leave them stunlocked for an hour after reviving 'cause of the code on line 64 in station_special_vr. I dunno if there's a better var to use though.

Arokha: Are they incapacitated? Because we can just make them sleep instead. Or stunned. I mean are they supposed to be aware of what's going on around them?

Screemonster: Making them sleep until they hatch would also work, as long as they can't be woken up early. I think cadence only used weaken instead of sleep so the chimera player could sort-of-RP being passive and useless (or possibly tasty in the event of someone deciding to omf them while they're there helpless.

HOWEVER - is there any reason why the reviving var can't just have multiple states? Like "reviving = 1" for on cooldown, "reviving = 2" for actively reviving, "reviving = 0" for neither.

code/modules/mob/living/carbon/human/species/station/station_special_vr.dm
```diff
@@ -56,142 +56,198 @@
...
+	var/jittery = H.jitteriness >= 100
+
+	//To reduce distant object references
+	var/feral = H.feral
+
+//Handle feral triggers and pre-feral messages
+	if(!feral && (hungry || shock || jittery))
+
+		// If they're hungry, give nag messages (when not bellied)
+		if(H.nutrition >= 100 && prob(0.5) && !ismob(H.loc))
```
Screemonster: the !ismob(H.loc) check won't work for new bellies, 'cause the location of the prey is now the belly object rather than the mob.

Arokha:
@Leshana @Screemonster 

Made suggested changes, rebased and combined into two commits.

Screemonster:
Apart from REVIVING_DONE needing set in a few spots (basically anywhere the verb is added) and the ismob() check on line 105 needing replaced with an istype(H.loc, belly) kinda thing, the flow on this looks pretty good.

PR #3119, opened 2018-02-24
Preliminary rewrite of pipe construction.

Leshana:
- Moved pipe construction defines into __defines/construction.dm
- Unified pipe *unwrenching* by creating a standard proc along with the `construction_type` var.
- Eliminated the pipe fitting name & icon_state lookup tables by adding `pipe_state` var on atmos machinery and referencing that.
    - Each pipe which can be made from a fitting object should override `pipe_state` with the icon state to be used on the pipe fitting object.
- Eliminated the giant switch statement of doom in pipe construction by delegating that work to `on_construction` proc.
    - To make this work, every pipe must implement `get_neighbor_nodes_for_init` which returns a list of nodes which should be re-initialized on that pipe's construction.
- Combined the SCRUBBERS, SUPPLY and REGULAR pipe fitting classes together by storing the `piping_layer` variable and using the `setPipingLayer` procs
- Standardized the code for searching for node neighbors into the `can_be_node` proc.
    - This proc is also improved in that is a mutual check, `check_connectable` is called on BOTH objects, so they have to mutually agree to connect as nodes. Eliminates lots of special edge case logic.
    - Updated all the `amos_init` procs to use `can_be_node`.  In the most common cases, even that boilerplate code is consolidated into the `STANDARD_ATMOS_CHOOSE_NODE` macro.
- Implemented `pipe_flags` which lets pipes declare (or override) certain requirements.
- Adds a "pipe_recipe" datum to help out things that construct pipes.  By taking it out of the dispenser, we open the road for multiple dispenser types.  No, no RPD yet.  Soon.
    - Enhances the pipe dispenser to operate on pipe recipe datums instead of hard coded lists of pipes it can construct.   These datums are also (partially) initialized from the pipe machine types themselves, reducing having to define stuff in multiple places.
    - Switched pipe dispenser UI to use browse().   Not a NanoUI, but makes it a bit prettier with low effort.
    - Changed pipe dispenser to use a button selector to switch between Regular/Scrubbers/Supply instead of having separate list items.
- Added icon states to HE pipes to support the "connected on neither side" state.


Leshana:
@AbbieFland look now we have improved pipe construction

AbbieFland:
I like the new dispenser UI; it no longer constantly recentering every click is fantastic!

A couple issues I ran in to:
 * Chillers don't seem to get any air inside them despite seeming connected. I noticed this when the R&D server room air alarm went off: upon checking the loop was pressurized just fine but the chiller was detecting zero pressure.
 * The one-per-turf settings seem to be a little too aggressive in some cases. It doesn't seem to allow straight pipes to cross over any more, or even to replace the default scrubber-supply same-tile setup after disassembly:
image

Leshana:
@AbbieFland Good catch! I had concentrated on mapped in pipes working and newly constructed pipes connecting.   Un-wrenching and re-wrenching mapped in pipes was a good test.   
I believe I have corrected all of that, along with the issue of gas chillers and heaters connecting to pipenets.

PR #3128, opened 2018-02-25
Adds hollow bones trait

aztectornado:
~~DO NOT MERGE - Requesting help~~

This adds a new negative trait which will cause all limbs to be much, much easier to break. Most are decreased to around 20-15 damage before break, as opposed to 35-30, with the exception of hands/feet, which were already fragile as heck and only see a minor decrease since any more would result in way too many breaks.

~~However, as I tested, I had a thought: "What about robot limbs? Will this apply to them too?"~~

~~And irritatingly, after testing, I found that it did not in fact pass the new limbs to FBPs when initializing them. Everything else works as expected, but I need help locating why the roboticize() isn't simply running on the /weak versions of limbs.~~

~~Until this is fixed this should not be merged; it'd basically give FBP powergamers a free 4 trait points. Anyone able to tell where I went wrong?~~

Edit: Fixed the issue with help from Arokha. It should be solid now.

### DIFF
--- a/code/modules/mob/living/carbon/human/human_powers.dm
+++ b/code/modules/mob/living/carbon/human/human_powers.dm
@@ -173,7 +173,7 @@
 
 	for(var/obj/item/organ/external/EO in organs)
 		if(EO.brute_dam || EO.burn_dam)
-			output += "[EO.name] - <span class='warning'>[EO.burn_dam + EO.brute_dam > ROBOLIMB_REPAIR_CAP ? "Heavy Damage" : "Light Damage"]</span>\n"
+			output += "[EO.name] - <span class='warning'>[EO.burn_dam + EO.brute_dam > EO.min_broken_damage ? "Heavy Damage" : "Light Damage"]</span>\n" //VOREStation Edit - Makes robotic limb damage scalable
 		else
 			output += "[EO.name] - <span style='color:green;'>OK</span>\n"
 
--- a/code/modules/mob/living/carbon/human/species/station/traits_vr/negative.dm
+++ b/code/modules/mob/living/carbon/human/species/station/traits_vr/negative.dm
@@ -96,6 +96,17 @@
 	cost = -1
 	var_changes = list("flash_mod" = 2.0)
 
+/datum/trait/hollow
+	name = "Hollow Bones/Aluminum Alloy"
+	desc = "Your bones and robot limbs are much easier to break."
+	cost = -2 //I feel like this should be higher, but let's see where it goes
+
+/datum/trait/hollow/apply(var/datum/species/S,var/mob/living/carbon/human/H)
+	..(S,H)
+	for(var/obj/item/organ/external/O in H.organs)
+		O.min_broken_damage *= 0.5
+		O.min_bruised_damage *= 0.5
+
 /datum/trait/lightweight
 	name = "Lightweight"
 	desc = "Your light weight and poor balance make you very susceptible to unhelpful bumping. Think of it like a bowling ball versus a pin."
--- a/code/modules/organs/organ_external.dm
+++ b/code/modules/organs/organ_external.dm
@@ -400,7 +400,7 @@
 		user << "<span class='notice'>Nothing to fix!</span>"
 		return 0
 
-	if(damage_amount >= ROBOLIMB_REPAIR_CAP)
+	if(damage_amount >= min_broken_damage) //VOREStation Edit - Makes robotic limb damage scalable
 		user << "<span class='danger'>The damage is far too severe to patch over externally.</span>"
 		return 0
 
@@ -1085,6 +1085,7 @@ Note that amputating the affected organ does in fact remove the infection from t
 
 	dislocated = -1
 	cannot_break = 1
+	min_broken_damage = ROBOLIMB_REPAIR_CAP //VOREStation Addition - Makes robotic limb damage scalable
 	remove_splint()
 	get_icon()
 	unmutate()
@@ -1133,7 +1134,7 @@ Note that amputating the affected organ does in fact remove the infection from t
 	return !(status & (ORGAN_MUTATED|ORGAN_DEAD))
 
 /obj/item/organ/external/proc/is_malfunctioning()
-	return ((robotic >= ORGAN_ROBOT) && (brute_dam + burn_dam) >= 25 && prob(brute_dam + burn_dam))
+	return ((robotic >= ORGAN_ROBOT) && (brute_dam + burn_dam) >= min_broken_damage*0.83 && prob(brute_dam + burn_dam)) //VOREStation Edit - Makes robotic limb damage scalable
 
 /obj/item/organ/external/proc/embed(var/obj/item/weapon/W, var/silent = 0)
 	if(!owner || loc != owner)


Arokha:
Because FBPs recreate all their limbs when they call robotize(). Definitely don't do it this way (no offense).

I'll find a suggestion for you.

aztectornado:
Yeah, figures the simple way wouldn't work for this, especially once you consider FBPs and the weirdness that is their code. 

The other way I thought of was to add a trait-specific apply(), but even now I can't think of a way that it could work without wasting processor time, or inserting a vore edit into roboticize that checks a species var; though I'm not sure if it'd be able to access that var safely.

Arokha:
Honestly, do it in apply, yes. Iterate over external organs, and alter that var.
`min_broken_damage *= 0.75` or something similar. Then you do it once (or again if they respawn, but whatever.

aztectornado:
I suppose that'd work, yeah. I thought apply was done prior to spawn

Arokha:
No, apply is passed the human, so they can be manipulated however.

aztectornado:
Overwrote the old commit with the new process. It works on FBP too now.

aztectornado:
...wait fuck vore internals aren't external organs, are they?

aztectornado:
They aren't, so what gives with Travis? I don't know what would cause those runtimes.

KasparoVy:
Travis took a second lap and made it this time.

code/modules/mob/living/carbon/human/species/station/traits_vr/negative.dm
```diff
@@ -96,6 +96,30 @@
...
+	..(S,H)
+	for(var/obj/item/organ/external/O in H.organs)
+		if(istype(O, /obj/item/organ/external/head))
+			O.min_broken_damage = 25 //Skulls are a bit denser
+
+		else if(istype(O, /obj/item/organ/external/chest) || istype(O, /obj/item/organ/external/groin))
+			O.min_broken_damage = 20
+
+		else if(istype(O, /obj/item/organ/external/arm) || istype(O, /obj/item/organ/external/leg))
+			O.min_broken_damage = 15
```
Arokha: This stuff is why I suggested doing `min_broken_damage *= 0.75` (or something similar) since that would reduce every organ to 75% of the original value, effectively being a 25% debuff regardless of what the original values were, specifically.

code/modules/mob/living/carbon/human/species/station/traits_vr/negative.dm
```diff
@@ -96,6 +96,30 @@
 	cost = -1
 	var_changes = list("flash_mod" = 2.0)
 
+/datum/trait/hollow
+	name = "Hollow Bones/Aluminum Alloy"
+	desc = "Your bones and robot limbs are much easier to break."
+	cost = -4
```
Arokha: I'll have to start a discussion with admins about the cost, since that's a very high point return, considering most people on the server don't get into combat, or get injured.

code/modules/mob/living/carbon/human/species/station/traits_vr/negative.dm
aztectornado: The problem is that'd result in weird rounding errors that my OCD dislikes.. Plus, since hollow bones are not uniformly weaker in birds, I feel that it should be reflected in code.

code/modules/mob/living/carbon/human/species/station/traits_vr/negative.dm
aztectornado: This was the suggested point value for it, since without osteodaxon/pink slime, one or two hits from anything will result in cracked bones. This makes it even more brutal than a 50% damage increase, since at least with those you can bandage it. 

Also, this makes a broken bone from falling any distance exceedingly likely. A single slime swat is nearly guaranteed to break something. Heck, most mobs can break any of these in one or two hits as well. 

Basically, so far as combat drawbacks go, this almost guarantees that if you get into combat, you go to medbay right after, if you're not getting floored from agony, or starting to bleed out from jostling the bones around. I consider that a sizable curse to carry; it basically makes melee combat impossible.

SilencedMP5A5:
meem

code/modules/mob/living/carbon/human/species/station/traits_vr/negative.dm
Arokha: No offense but code cleanliness comes before your OCD. Don't use hardcoded organ types. If you really hate non-integer results, put a round(,1) around it.

code/modules/mob/living/carbon/human/species/station/traits_vr/negative.dm
```diff
@@ -96,6 +96,30 @@
...
+
+		else if(istype(O, /obj/item/organ/external/arm) || istype(O, /obj/item/organ/external/leg))
+			O.min_broken_damage = 15
+
+		else if(istype(O, /obj/item/organ/external/hand) || istype(O, /obj/item/organ/external/foot))
+			O.min_broken_damage = 10 //About as low as it can go without misclicks shattering things
+			O.min_bruised_damage = 7
+
+		else //Whatever it is, it needs a change too
+			O.min_broken_damage = 15
```
Arokha: Bad form to assume 15 is higher than any future organs ever added. Again, use a multiplier.

aztectornado: Y'mean lower, Arokha? 15 is the lowest any external organ (hands/feet specifically have this) has currently, it seems doubtful that there'll be anything worse... And even if there is, we can change that when it occurs.

code/modules/mob/living/carbon/human/species/station/traits_vr/negative.dm
aztectornado: It's more that I don't want a uniform change in bones since the skull is a little more dense than other bones, as well as the fact that the foot/hand starts so low that it requires a bruised damage change too. This would be easier to fine tune in the event that we find that something is too strong or more likely, too weak

aztectornado:
Altered the commit, now it's just a flat 50% easier break rate. (25% resulted in numbers too lenient, in my opinion)

code/modules/mob/living/carbon/human/species/station/traits_vr/negative.dm
Screemonster: I'd be interested to see this maybe have a lower point return, BUT reduce the nutrition costs for flying. Because you're lighter.

Arokha: Synth limbs don't break, so I'd be interested in making sure synths can't take this, because it would have no effect. Otherwise there's no reason every synth won't just take this for automatic haste+ affordability. Also interested in maybe just reducing the cost to -3, but still talking about it with others.

aztectornado: Wait, I thought that synth limbs did go into that sparking mess where they drop items/fall over from the damages? Does it not check min_broken for that?

aztectornado:
Made changes to the robotic limb stats; now it starts malfunctioning when it hits min_broken_damage, and stops being externally repairable at MBD+10. This should allow for scaling robolimb quality, if ever anyone ends up wanting to make the different robotic limb brands have differing stats.

aztectornado:
Any news on this one?

code/modules/mob/living/carbon/human/human_powers.dm
```diff
@@ -173,7 +173,7 @@
 
 	for(var/obj/item/organ/external/EO in organs)
 		if(EO.brute_dam || EO.burn_dam)
-			output += "[EO.name] - <span class='warning'>[EO.burn_dam + EO.brute_dam > ROBOLIMB_REPAIR_CAP ? "Heavy Damage" : "Light Damage"]</span>\n"
+			output += "[EO.name] - <span class='warning'>[EO.burn_dam + EO.brute_dam > EO.min_broken_damage + 10 ? "Heavy Damage" : "Light Damage"]</span>\n" //VOREstation edit - Makes robotic limb damage scalable
```
Arokha: I meannnn. I'm not okay with you changing how robolimb repair works for people who don't have this trait, in a PR titled "Adds hollow bones trait" and doesn't mention this change in the description, just from a general perspective.

Definitely not this one, since this change makes synths without this trait far more robust, raising things like their torso repair cap from 30 to 45.

code/modules/mob/living/carbon/human/species/station/traits_vr/negative.dm
Arokha: Reduce to -2.

code/modules/organs/organ_external.dm
```diff
@@ -400,7 +400,7 @@
 		user << "<span class='notice'>Nothing to fix!</span>"
 		return 0
 
-	if(damage_amount >= ROBOLIMB_REPAIR_CAP)
+	if(damage_amount >= min_broken_damage + 10) //VOREDtation Edit - Makes robotic limb damage scalable
```
Arokha: See previous comment about this.

code/modules/organs/organ_external.dm
```diff
@@ -1133,7 +1133,7 @@ Note that amputating the affected organ does in fact remove the infection from t
 	return !(status & (ORGAN_MUTATED|ORGAN_DEAD))
 
 /obj/item/organ/external/proc/is_malfunctioning()
-	return ((robotic >= ORGAN_ROBOT) && (brute_dam + burn_dam) >= 25 && prob(brute_dam + burn_dam))
+	return ((robotic >= ORGAN_ROBOT) && (brute_dam + burn_dam) >= min_broken_damage && prob(brute_dam + burn_dam)) //VOREStation Edit - Makes robotic limb damage scalable
```
Arokha: See previous comment, though this math here is different. You'd be raising some from 25 to 35, and lowering some from 25 to 15, etc.

aztectornado:
H'okay. Redid both those commits, this time actually checking my goddamn math. (What the fuck Past-Aztec??)

Hollow Bones is now a -3 trait; I'll ask to bump it to -4 if it seems appropriate after play testing. It also now adjusts min_bruised_damage, since I realized that on hands/feet, the bruised level would actually be higher than the break level, meaning you'd snap bones before being bruised. Bruised damage doesn't seem to affect almost anything on external organs at the moment, but I figure it's prudent to change it in case it ever is used.

Robotic limbs still all stop being repairable at 30, and start malfunctioning at 83% of that, but can now be scaled according to min_broken_damage. (Almost equivalent to the way it was before; they malfunction at 24.9 instead of 25 now. That was as close as I could get. :/ )

If they have the aluminum alloy, they stop repairs at 15, and malfunction at 12.45. A tiny window for sure, but it's exactly in line with the ratios in vanilla.

PR #3133, opened 2018-02-25
Convert turfs to use add_overlays()

Leshana:
- Convert fire alarms to use add_overlay() A good first test.
- Fix runtime on universal pipe adapter update_icons
- Unfortunately BYOND's initialization order is strange, and the appearance_bro var is only half initialized when map starts to load, causing errors.  We temporarily fix by moving it to be a global-scoped global.
- Converted as much as I could find about turf overlays to use add_overlay().
- This should be enough to stop BYOND from crashing, so we can eliminate the turf_overlay_holder hack.
- This also lets us remove the anti-corruption hacks from walls and open space.
- ZAS gas overlays can use priority overlays, so this also fixes the gas-goes-away-when-crowbarring-plating issue.


Leshana:
Testing by @AbbieFland and @Leshana show it working so far.

PR #3138, opened 2018-02-25
Gives prometheans the trash eater trait.

Cameron653:


Screemonster:
macro promethean janitor cleans up the hallways (2562, colorised)
image

Cyantime:
Verk mentioned that this also need to give them the `/mob/living/proc/eat_trash` verb.

Arokha:
Depends on what they wanna do. If they wanna eat trash items, this would do it. If they want to eat 'a ton of things' like pAIs and stuff then they'd need the verb.

Cameron653:
Doesn't trashcan var also give them the eat_trash verb?

Cameron653:
OH, I see. The code was changed for it.

PR #3139, opened 2018-02-25
Promethean Water Edits

### LINKED ISSUE #3136: Prometheans are weak to water
Recently a change was made to Prometheans such that they'd take toxins damage from water, for no discernible reason other than "Lol slimes are weak to it for some reason"

I could still propose sexy shower times or pool fun to slime gals. 

#### What actually happened
They take a stupidly minor amount of toxins damage from drinking or touching water; so minor it's barely actually a weakness so much as it is an annoyance, and a deterrent to RPing in those situations despite the fact I know a lot of people are probably into it. In addition, this made protein shakes poisonous to them. This all makes little sense in the lore overall (don't get me started on slimes), but this is especially true coupled with the fact that not all our promethean players are actually xeno-bio prometheans; we have slime girls, tentacled monsters, mimics, and so many other things on station at the regular this does not feel like a sane thing to add to _all_ of them.

#### Steps to reproduce
- Play with or as prommie
- drink water or go to a shower
- sigh as your scene is constantly interrupted by "Your body stings slightly" messages

#### Code Revision
- 22568ffdf997493f46ae28b3f572e793e53a3899 
 
- Remember that watery amoeba boss from Legend of Zelda? Yeah just-... Any of the games, really. Always a water boss. Slimes are water aligned from a cultural sense. I still don't get why we've kept this as their weakness for so long.

Cameron653:
Resolves #3136

- Makes prometheans not die to water. Seriously, it's flat out annoying and a dumb 'feature.'
- Keeps the promethean weakness to water, however. They won't heal if they have water _on_ them. 
Due to the way the code works, they still heal if water is injected into them or drank by them, but that's simply how it's coded.

### DIFF
--- a/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Core.dm
+++ b/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Core.dm
@@ -170,10 +170,10 @@
 		remove_self(needed)
 
 /datum/reagent/water/affect_blood(var/mob/living/carbon/M, var/alien, var/removed)
-	if(alien == IS_SLIME)
+	/*if(alien == IS_SLIME) //VOREStation Edit Start. Stops slimes from dying from water.
 		M.adjustToxLoss(6 * removed)
-	else
-		..()
+	else*/
+	..() //VOREStation Edit End.
 
 /datum/reagent/fuel/affect_ingest(var/mob/living/carbon/M, var/alien, var/removed)
 	if(alien == IS_SLIME)
--- a/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Core.dm
+++ b/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Core.dm
@@ -170,7 +170,7 @@
 		remove_self(needed)
 /*  //VOREStation Edit Start. Stops slimes from dying from water. Fixes fuel affect_ingest, too.
 /datum/reagent/water/affect_blood(var/mob/living/carbon/M, var/alien, var/removed)
-	if(alien == IS_SLIME) //VOREStation Edit Start. Stops slimes from dying from water.
+	if(alien == IS_SLIME)
 		M.adjustToxLoss(6 * removed)
 	else
 		..()
@@ -180,7 +180,7 @@
 		M.adjustToxLoss(6 * removed)
 	else
 		..()
-/*  //VOREStation Edit End.
+*/  //VOREStation Edit End.
 /datum/reagent/fuel
 	name = "Welding fuel"
 	id = "fuel"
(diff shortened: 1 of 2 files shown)


Screemonster:
This weakness I'm fine with 'cause it won't interfere with 1: eating, or 2: lewding in the showers. Both of which are frankly essential features around these parts.

~next step is making them weak to space cleaner~

code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Core.dm
```diff
@@ -170,10 +170,10 @@
 		remove_self(needed)
 
 /datum/reagent/water/affect_blood(var/mob/living/carbon/M, var/alien, var/removed)
-	if(alien == IS_SLIME)
+	/*if(alien == IS_SLIME) //VOREStation Edit Start. Stops slimes from dying from water.
 		M.adjustToxLoss(6 * removed)
-	else
-		..()
+	else*/
+	..() //VOREStation Edit End.
```
Arokha: You have commented out this entire proc's contents... by editing all but 1 line inside of it (and that one line is commented out)
Just put the entire thing inside a block comment using the whitespace above and below.

code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Core.dm
```diff
@@ -170,10 +170,10 @@
...
 
 /datum/reagent/fuel/affect_ingest(var/mob/living/carbon/M, var/alien, var/removed)
```
Screemonster: There's a PR upstream to change this 'cause they copypasted and put "fuel" in instead of water. Just comment out both procs entirely, 'cause this probably breaks fuel's _actual_ affect_ingest() too.

code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Core.dm
Cameron653: Resolved. Did a /* */ above water and below fuel, since they're both unneeded/borked

code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Core.dm
Cameron653: Resolved. Did a /* */ above water and below fuel, since they're both unneeded/borked

Arokha: Fuel doesn't have an affect_ingest(), so there would be nothing to mess up. It would make prometheans take more damage from drinking fuel (which would be fine, really). But like mentioned, it's already fixed upstream to water.

PR #3142, opened 2018-02-26
Rewrites vore

Arokha:
Things you may care about:
* There's an 'eating sounds' and 'digestion sounds' toggle in global settings for those who don't like to hear all that
* Vore sounds adjusted in general to not be quite as deafening, and have sharp falloff ranges, but not be be the inefficient 'X in range(1)' thing it was before. Some are also blocked by walls. (May require volume tweaking, bear with us on that)
* Sickshot now works on living mobs, not just humans
* Less lag caused by bellies (probably? maybe?)
* Somewhat unlikely to delete all your bellies

Things you probably don't care about but I think are cool:
* All bellies are now objects rather than the abstract datums they have been for years
* The Bellies subsystem handles all belly processing rather then hooking into Life()
* Bellies are less processor-intensive when not doing anything
* /tg/'s playsound system is probably more efficient
* Framework for making sure things don't "fall out" of preds as much


AbbieFland:
Absorbing seems kind of funny.

From the prey's perspective, things seem basically as before.

From the pred's perspective:
 * Once it completes, the prey disappears from the vore panel rather than turning purple.
 * Unabsorb mode appears to do nothing.
 * Unsorbitol appears to do nothing.

code/modules/mob/living/carbon/brain/brain.dm
```diff
@@ -8,6 +8,7 @@
 	use_me = 0 //Can't use the me verb, it's a freaking immobile brain
 	icon = 'icons/obj/surgery.dmi'
 	icon_state = "brain1"
+	no_vore = TRUE //VOREStation Edit - PLEASE. lol.
```
Cyantime: brain_vr.dm?

code/modules/mob/living/carbon/human/human_species.dm
```diff
@@ -2,6 +2,7 @@
 	real_name = "Test Dummy"
 	status_flags = GODMODE|CANPUSH
 	has_huds = FALSE
+	no_vore = TRUE //VOREStation Add - Dummies don't need bellies.
```
Cyantime: human_species_vr.dm?

code/game/sound.dm
```diff
@@ -136,60 +82,54 @@ var/const/FALLOFF_SOUNDS = 0.5
...
-			if (M.hallucination)
-				S.environment = PSYCHOTIC
-			else if (M.druggy)
-				S.environment = DRUGGED
-			else if (M.drowsyness)
-				S.environment = DIZZY
-			else if (M.confused)
-				S.environment = DIZZY
-			else if (M.sleeping)
-				S.environment = UNDERWATER
```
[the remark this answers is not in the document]
Arokha: Thanks for pointing that out. I'll put it back in, but probably in a get_sound_env() proc on mobs.

code/modules/mob/living/silicon/pai/pai.dm
```diff
@@ -389,9 +389,7 @@
 	if(src.loc == card)
 		return
 
-	for(var/I in vore_organs) //VOREStation edit. Release all their stomach contents. Don't want them to be in the PAI when they fold or weird things might happen.
-		var/datum/belly/B = vore_organs[I] //VOREStation edit
-		B.release_all_contents() //VOREStation edit
+	release_vore_contents()
```
Cyantime: Should have `//VOREStation Edit`

code/modules/mob/living/carbon/brain/brain.dm
Arokha: Don't really like making a file for a single line change. It's below my "do I want to complicate things for future mergers, or future code-searchers" threshold.

code/modules/mob/living/carbon/human/human_species.dm
Arokha: This one, yes, since that file already exists.

code/modules/mob/living/silicon/pai/pai.dm
Arokha: Will add

code/modules/vore/eating/belly_obj_vr.dm
```diff
@@ -354,49 +351,58 @@
...
+		to_chat(R,"<span class='warning'>You attempt to climb out of \the [lowertext(name)]. (This will take around [escapetime/10] seconds.)</span>")
+		to_chat(owner,"<span class='warning'>Someone is attempting to climb out of your [lowertext(name)]!</span>")
 
 		if(do_after(R, escapetime, owner, incapacitation_flags = INCAPACITATION_DEFAULT & ~INCAPACITATION_RESTRAINED))
-			if((owner.stat || escapable) && (R in internal_contents)) //Can still escape?
+			if((owner.stat || escapable) && (R.loc == src)) //Can still escape?
 				release_specific_contents(R)
 				return
-			else if(!(R in internal_contents)) //Aren't even in the belly. Quietly fail.
+			else if(!(R.loc == src)) //Aren't even in the belly. Quietly fail.
```
Cyantime: `!=`?

Arokha: Would look nicer, yes

code/game/sound.dm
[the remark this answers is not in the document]
Arokha: Readded as get_sound_env() on mobs.

AbbieFland:
Absorption issues from earlier seem fixed by the fixup, but in testing once that was resolved I came across a (seemingly-unintended) change in functionality:

Unchanged from before, a pAI with normal prey in their belly drops them on the floor when they collapse chassis.

However, previously a pAI was able to collapse chassis and still keep all absorbed prey. Now, collapsing chassis also drops all absorbed prey.

Arokha:
> However, previously a pAI was able to collapse chassis and still keep all absorbed prey. Now, collapsing chassis also drops all absorbed prey.

Not sure how I feel about that tbh. It's an easy fix by passing FALSE on pai.dm:392. But I'm not sure that pAIs would correctly handle absorbed prey in all cases where they abandon being a pAI or get wiped or anything. Hm.

AbbieFland:
I'll experiment with it a bit and at least see if anything immediately obvious breaks in those cases.

Since playing pAI often means being folded up by the owner depending what they're doing (and any random person walking by can with one click do a boop and force fold-up), the change would make any scenes involving absorption even more of a hassle to deal with than playing a pAI already is.

I can understand making the tradeoff if it causes Bad Things, but it is a real cost.

Arokha:
> but it is a real cost

So's being deleted or moved to nullspace by poorly thought-out code. :p

AbbieFland:
Guess they got really really absorbed!

AbbieFland:
Tested it out with the flag changed and while folded with an absorbed prey:
 * Leaving due to wipe software, pAI suicide, or EMP death ghosts the absorbed prey, much like if a regular pred cryos. This is probably ok.
 * If in a belly (with trash-eat trait) and the outer pred cryos/etc., the pAI ghosts and the absorbed prey ghosts. This is probably ok.
 * If in a pocket and the carrier cryos/etc., the pAI goes into the asset retention console. The pAI can still control themself and can unfold chassis to pop out, and the absorbed prey is still in them. This is not ideal but is maybe ok - if pAI was afk the absorbed prey would have to ooc escape to get out of the console, but then if the pAI went afk without being carried off then ooc escape would be the only way out anyway.

code/modules/vore/eating/simple_animal_vr.dm
```diff
@@ -34,20 +34,19 @@
 	var/mob/living/carbon/human/user = usr
 	if(!istype(user) || user.stat) return
 
-	var/datum/belly/B = vore_organs[vore_selected]
 	if(retaliate || (hostile && faction != user.faction))
 		user << "<span class='warning'>This predator isn't friendly, and doesn't give a shit about your opinions of it digesting you.</span>"
 		return
-	if(B.digest_mode == "Hold")
+	if(vore_selected.digest_mode == "Hold")
```
Leshana: I recommend switching to DM_* constants

Arokha:
> If in a belly (with trash-eat trait) and the outer pred cryos/etc., the pAI ghosts and the absorbed prey ghosts. This is probably ok.

What about digested in a belly with trash eat trait though?

code/modules/vore/eating/simple_animal_vr.dm
Arokha: Didn't touch that, but I suppose this is a refactor, so fixed.

code/modules/vore/eating/belly_obj_vr.dm
```diff
@@ -97,56 +99,69 @@
...
+
+// Called whenever an atom enters this belly
+/obj/belly/Entered(var/atom/movable/thing,var/atom/OldLoc)
+	if(isliving(thing))
+		var/mob/living/M = thing
+		if(desc)
+			to_chat(M, "<span class='notice'><B>[desc]</B></span>")
+		var/taste
+		if(can_taste && (taste = M.get_taste_message(FALSE)))
+			to_chat(owner, "<span class='notice'>[M] tastes of [taste].</span>")
```
Leshana: It looks like this may print the taste message *before* the "x slides into your y" message.

AbbieFland:
> > If in a belly (with trash-eat trait) and the outer pred cryos/etc., the pAI ghosts and the absorbed prey
> > ghosts. This is probably ok.

> What about digested in a belly with trash eat trait though?

I'll have to test that next time I'm experimenting. I don't think any of the tests today involved digesting a card in a trash-eat belly.

code/modules/vore/eating/belly_obj_vr.dm
```diff
@@ -348,55 +345,67 @@
...
 
+//Determine where items should fall out of us into.
+//Typically just to the owner's location.
+/obj/belly/drop_location()
+	//Should be the case 99.99% of the time
+	if(owner)
+		return owner.loc
+	//Sketchy fallback for safety, put them somewhere safe.
+	else
+		var/fallback = pick(latejoin)
```
Leshana: If this ends up happening, might be worth logging

code/modules/vore/eating/vorepanel_vr.dm
```diff
@@ -99,8 +97,8 @@
 	dat += "<HR>"
 
 	dat += "<ol style='list-style: none; padding: 0; overflow: auto;'>"
-	for(var/K in user.vore_organs) //Fuggin can't iterate over values
-		var/datum/belly/B = user.vore_organs[K]
+	for(var/belly in user.vore_organs) //Fuggin can't iterate over values
```
Leshana: Now there is no need to iterate over associated values!

Very-Soft:
I did test that, pAI cards don't seem to be digestible.

code/modules/vore/eating/belly_obj_vr.dm
Arokha: Weird that it doesn't consider this comment outdated, but, changed.

Arokha:
Rebased.

PR #3145, opened 2018-02-27
Readds NIF examine text

Arokha:
Recoded it a little different. The verb is in OOC, if you have a NIF. It's saved as part of the NIF, so if your head explodes and you lose your NIF, you'll need to describe your new one.

Also fixes an unrelated bug where ghosts examining people print all sorts of crazy messages about being able to trample people.

### DIFF
--- a/code/modules/nifsoft/nif.dm
+++ b/code/modules/nifsoft/nif.dm
@@ -70,7 +70,12 @@ You can also set the stat of a NIF to NIF_TEMPFAIL without any issues to disable
 
 	//Put loaded data here if we loaded any
 	save_data = islist(load_data) ? load_data.Copy() : list()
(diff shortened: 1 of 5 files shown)


code/modules/nifsoft/nif.dm
```diff
@@ -69,6 +70,7 @@ You can also set the stat of a NIF to NIF_TEMPFAIL without any issues to disable
 
 	//Put loaded data here if we loaded any
 	save_data = islist(load_data) ? load_data.Copy() : list()
+	examine_msg = save_data["examine_msg"] || "There's a certain spark to their eyes."
```
Leshana: Does this mean that clearing the message will get reset to spark after next load?

Arokha: Yes, fixed.

PR #3152, opened 2018-02-28
Adds explorer meeting room

Mewchild:
Adds a meeting room to the expedition hangar.
Adds a shipping destination to cargo, to the expedition hangar.
Adds pilot spawn locations (In the cockput of the ship).
Updates NanoMap to reflect current state of the map


Mewchild:
Figured out the issues. Will make a new PR in a day or two, once I remake it.

PR #3154, opened 2018-02-28
Fixed Language upgrade not being removed after being applied

Heroman3003:
Language upgrade simply said an error and did not disappear, while borg received all languages. Not intended behavior at all, fixed.


code/game/objects/items/robot/robot_upgrades_vr.dm
```diff
@@ -5,4 +5,7 @@
 		R.add_language(LANGUAGE_CANILUNZT,		1)
 		R.add_language(LANGUAGE_ECUREUILIAN,	1)
 		R.add_language(LANGUAGE_DAEMON,			1)
-		R.add_language(LANGUAGE_ENOCHIAN,		1)
\ No newline at end of file
+		R.add_language(LANGUAGE_ENOCHIAN,		1)
+		return 1
+	else
+		return 0
```
Arokha: You can actually do this more simplyon line 2 by doing `if((. = ..()))` which both sets our own return value equal to, and checks the parent's return value. Then you don't need to return anything in particular. You don't need to change it though, just so you know for the future.

PR #3156, opened 2018-02-28
Small wide/dog borg tweaks/fixes

Verkister:
-Adds a new var to specify nonstandard borg sprite sizes. No longer dogborg specific.
-Makes the dogborg var usable for standard sizes as well (for the special features like rest poses and gun/gut overlays)
-Makes the borg reset module reset the forementioned vars.
~~-Excludes wideborg from the standard size polaris/vore vr_sprite_check to avoid unintended default file reverts.~~(quickfixed in master)
-Makes the wideborgs var skip the vr_sprite_check anyway for less unnecessary file swapping.


Leshana:
So does this help with the ones that are big?

Verkister:
Basically yeah. With this stuff the reset module clears both, wide and dogborg settings. Also now the dogborg sprite features can also be applied to normal sized borg types if applicable.

PR #3160, opened 2018-03-01
Make syringe reuse a bad idea

Arokha:
Syringes come capped, and can infect and transfer viruses to people if you start reusing them. In exchange, syringe boxes have 20 syringes each (and medivends still have 12 each to dispense individually).

Also ports Goon's syringe sprites. They're bigger, easier to click, and have a little more detail (also color-coded inject/draw headers).


Jakksergal:
I'm not in favor of this unless there is an infinite source of capped syringes somewhere convenient in medical. An infinite source that only requires a single click with a storage box in hand to replenish.

Arokha:
There are hundreds, and they can be recycled down the hall at cargo in the autolathe.

Draycu:
I do love the addition of something which would make more sense, with a more realistic syringe bringing a new level of depth to the system.

However, I do agree with Jakksergal, there needs to be some sort of reliable way to make sure medical will not run out of syringes after using several on a patient.  

I have to ask if it would be possible to make a syringe dispenser that would require sheets of glass and metal and have it located somewhere in medical.  I do recall that cargo typically has left over glass and metal when filling up their autolathe at round start, and this could be used for the syringe dispenser.  As long as medical players don't mind interacting with cargo players and needing to work together.

Draycu:
Or, yes, simply asking cargo for boxes to be filled and provided would work just as well without edits to the map.

Arokha:
Medical literally starts with 100+ though, already. I doubt you will run out. VERY MUCH doubt.

Arokha:
Going to test merge this, basically. If you all find that it makes medical way too unplayable or something, we will revisit it, and modify or change it however it's needed.

Jakksergal:
How does this effect cyborg syringes?

Arokha:
> How does this effect cyborg syringes?

https://github.com/VOREStation/VOREStation/pull/3163

SpadesNeil:
It was decided in https://github.com/VOREStation/VOREStation/issues/3972 that we will be reverting this.

However, I want to keep the whole goal of stopping doctors from abusing the same syringe over and over and over again.

Syringes will be changed to no longer cause infection. However, they will be made to break after too many uses.

PR #3169, opened 2018-03-03
Bellies save via serialization to JSON

Arokha:
Change bellies to save via serialization out to JSON using an abstract serializer written by Tigercat2000 for Paradise. (Thanks TC for pointing it out!)

Prevents the 'server doom' issues from before.

Also reverts the revert of a commit I thought was originally responsible due to it being similar to server crashes related to floor decals.


KasparoVy:
:tiger: :ok_hand:

PR #3170, opened 2018-03-03
Return of the Fluff Items Workshop 

JoanRisu:
~Moves Crescent Rose to its own gun file just to make things clean and modular.
~Changes Crescent Rose to have a more up to date fire mode selecter.
+Adds the Gun Buster, a gun whose battery doubles as a sword. 
+Adds security alert level compliance to the Dominator.


code/modules/vore/fluffstuff/guns/gunsword.dm
```diff
@@ -0,0 +1,134 @@
...
+
+/obj/item/weapon/cell/device/weapon/gunsword/update_icon()
+	overlays.Cut()
+
+	if(charge < 0.01)
+		return
+	else if(charge/maxcharge >=0.995)
+		overlays = null
+	else
+		overlays = null
```
Arokha: Am confused as to what's going on here.

You remove all the overlays at the start of this by doing overlays.Cut(), which deletes them all.

Then you ... may or may not set overlays to null (which is not a thing byond lets you do, anyway since overlays must always be a list).

You never add any other overlays or anything. Why even have these conditions if all they do are: return, or, do something byond doesn't even let you do.

JoanRisu: That's just something I copied over from the original power cells code and edited. Would it keep the overlays.Cut() if I remove the if statements and leave it 

`
/obj/item/weapon/cell/device/weapon/gunsword/update_icon()
	overlays.Cut()
`

Arokha: Only worth doing if there's somewhere that adds overlays. Is there somewhere that does?

Leshana: Just remove that proc.

Arokha: I mean, shouldn't the power cell's update_icon() handle that itself? If you remove this entire proc, it'll just be calling /obj/item/weapon/cell/device/wepaon's update_icons, which should remove that, should it not?

code/modules/vore/fluffstuff/guns/gunsword.dm
```diff
@@ -0,0 +1,134 @@
...
+			"<span class='danger'>\The [user] is falling on \the [src]! It looks like [TU.he] [TU.is] trying to commit suicide.</span>"))
+		return (BRUTELOSS|FIRELOSS)
+
+/obj/item/weapon/cell/device/weapon/gunsword/update_icon()
+	overlays.Cut()
+
+	if(charge < 0.01)
+		return
+	else if(charge/maxcharge >=0.995)
+		overlays = null
```
Arokha: Please determine if overriding update_icon and not calling parent is necessary, and if not, remove this child. Even if it is required, overlays = null is not a valid thing to do in Byond.

code/modules/vore/fluffstuff/guns/gunsword.dm
JoanRisu: I don't know if I'm not being clear enough or if I just don't understand what is being asked. 

I didn't have that block of code in before so it inherited the overlay. Because of that, I added what I now understand included an invalid statement into that file to prevent the overlay from appearing at all. I would not have added that block of code if I had not encountered a problem in the first place.

PR #3174, opened 2018-03-04
Make syringes not require 3 clicks to inject

Arokha:
Previously they took (usually) 30ds to inject and you'd click 3 times to stack them and inject 5u each time. A little silly if you ask me, kind of a weird mechanic.

Now they take their injection time (so 30ds usually) and 66% of that is for starting the first injection, then it repeats the remaining 33% for it's injection size until it's empty.

So for a normal syringe this would be one click to get: 2s to do the first 5u, then 1s for the next 5u, then 1s more for the last 5u. Assuming there's that much in the syringe, anyway.

### DIFF
--- a/code/modules/reagents/reagent_containers/syringes.dm
+++ b/code/modules/reagents/reagent_containers/syringes.dm
@@ -214,13 +214,13 @@
 				mode = SYRINGE_DRAW
 				update_icon()
 
-				dirty(target,affected) //VOREStation Add
 			if(trans)
 				to_chat(user, "<span class='notice'>You inject [trans] units of the solution. The syringe now contains [src.reagents.total_volume] units.</span>")
 				admin_inject_log(user, target, src, contained, trans)
 			else
 				to_chat(user, "<span class='notice'>The syringe is empty.</span>")
 
+			dirty(target,affected) //VOREStation Add
 
 	return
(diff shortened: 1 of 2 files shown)


code/modules/reagents/reagent_containers/syringes.dm
```diff
@@ -188,32 +190,39 @@
...
 
 				if(injtime == time)
-					user.visible_message("<span class='warning'>[user] is trying to inject [target] with [visible_name]!</span>")
+					user.visible_message("<span class='warning'>[user] is trying to inject [target] with [visible_name]!</span>","<span class='notice'>You begin injecting [target] with [visible_name].</span>")
 				else
-					user.visible_message("<span class='warning'>[user] begins hunting for an injection port on [target]'s suit!</span>")
+					user.visible_message("<span class='warning'>[user] begins hunting for an injection port on [target]'s suit!</span>","<span class='notice'>You begin hunting for an injection port on [target]'s suit!</span>")
 
-				user.setClickCooldown(DEFAULT_QUICK_COOLDOWN)
+				user.visible_message("<span class='warning'>[user] injects [target] with the syringe!</span>")
```
Leshana: Why this line?  Doesn't this just result in double message printing?

PR #3180, opened 2018-03-04
Fixes Rischi's egg implant to actually work

Konabird:


Arokha:
I'm told this doesn't actually fix the issue, but you know, it's important to fix this anyway.

PR #3181, opened 2018-03-04
Fixes a fluff item

Cameron653:
Makes the egg implant able to lay eggs properly for Konabird


code/modules/vore/fluffstuff/custom_items_vr.dm
```diff
@@ -1103,7 +1103,7 @@
 			rimplant = I
 			break
 	if (rimplant)
-		if(rimplant.reagents.total_volume <= rimplant.transfer_amount)
```
Arokha: *Oh*

PR #3185, opened 2018-03-05
A whole bunch of vore related fixes.

### LINKED ISSUE #3173: Prey cannot hear the pred's gurgles from inside.
While otherwise audible in the same way as before (although hella quiet) the gurgles cannot be heard at all from inside the pred. You can hear them as the pred, next to the pred, on top of the pred, and after your spiritual remains ascend out of them, but not at all while you are inside.

For the prey to hear those noises especially well from their perspective.

#### What actually happened
Mime convention.

#### Steps to reproduce
- (Step 1) Get ate.
- (Step 2) Get gurgled.
- (Step 3) Hear absolutely nothing while getting gurgled.
- (Step 4) Become a burp ghost.
- (Step 5) Hear your crap still churning about ~~loud~~ and clear now.

#### Code Revision
- 7d899aeb7f6cfc60cb51726c36688e91abc14042
 
- Also hunger noises are MIA right now.

Verkister:
-Fixes prey not hearing gurgles from inside the gut.
-Tweaked the miniscule vore sound falloff that made things barely audible at all. People not in the immediately adjacent tiles may hear some faint noises if they listen intently enough, but that feels like a less of a problem compared to the stuff being barely hearable on ground zero tile. We already have a toggle pref now for the gurgle intolerant crowd.
-Fixes hunger noises.
-Hunger noises now scale gradually from 0 to 100% vol in the span of 250 to 0 nutrition.
-Fixes attack anims messing up pixel offsets on our big voremobs and dogborgs.
-Removes prey size multiplier from mob vore "fullness" will_eat blocker that would prevent 1-seater mobs from touching any even slightly taller than average chars. They already have a separate max size blocker in the will_eat thing.
-Makes bellies hide prey deathmessages again.
-Fixes security oties not changing to hold mode for human arrests.(I swear I had fixed this thing before???)
-Also disables the loud and spammy ejection on empty guts. Mobs were literally doing the thing every time they'd get killed, digested, or deleted and I found that extra bothersome while testing on the other stuffs. Gurgling a belly enabled mob caused at least 2 **SPLATS** every time.

Fixes #3173

### DIFF
--- a/code/modules/vore/eating/bellymodes_vr.dm
+++ b/code/modules/vore/eating/bellymodes_vr.dm
@@ -472,7 +472,5 @@
 			put_in_egg(P,1)
 
 	if(play_sound)
-		for (var/mob/living/M in contents)
-			M << sound(play_sound)
-		playsound(src, play_sound, vol = 100, vary = 1, falloff = VORE_SOUND_FALLOFF, ignore_walls = FALSE, preference = /datum/client_preference/digestion_noises)
+		playsound(src, play_sound, vol = 100, vary = 1, falloff = VORE_SOUND_FALLOFF, ignore_walls = TRUE, preference = /datum/client_preference/digestion_noises)
 	return SSBELLIES_PROCESSED
--- a/code/modules/vore/eating/bellymodes_vr.dm
+++ b/code/modules/vore/eating/bellymodes_vr.dm
@@ -3,7 +3,7 @@
 	if((times_fired < next_process) || !contents.len)
 		recent_sound = FALSE
 		return SSBELLIES_IGNORED
-	
+
 	if(loc != owner)
 		if(istype(owner))
 			loc = owner
@@ -41,7 +41,7 @@
 
 		if(prob(50)) //Was SO OFTEN. AAAA.
 			play_sound = pick(digestion_sounds)
-			
+
 		for (var/mob/living/M in contents)
 			//Pref protection!
 			if (!M.digestable || M.absorbed)
@@ -472,5 +472,7 @@
 			put_in_egg(P,1)
 
 	if(play_sound)
+		for (var/mob/living/M in contents)
+			M << sound(play_sound)
 		playsound(src, play_sound, vol = 100, vary = 1, falloff = VORE_SOUND_FALLOFF, ignore_walls = FALSE, preference = /datum/client_preference/digestion_noises)
 	return SSBELLIES_PROCESSED
--- a/code/modules/mob/death.dm
+++ b/code/modules/mob/death.dm
@@ -70,7 +70,7 @@
 
 	if(stat == DEAD)
 		return 0
-	if(src.loc && istype(loc,/mob/living)) deathmessage = "no message" //VOREStation Add - Prevents death messages from inside mobs
+	if(src.loc && istype(loc,/obj/belly)) deathmessage = "no message" //VOREStation Add - Prevents death messages from inside mobs
 	facing_dir = null
 
 	if(!gibbed && deathmessage != "no message") // This is gross, but reliable. Only brains use it.
--- a/code/modules/mob/living/carbon/human/life.dm
+++ b/code/modules/mob/living/carbon/human/life.dm
@@ -901,6 +901,11 @@
 		if(overeatduration > 1)
 			overeatduration -= 2 //doubled the unfat rate
 
+	if(noisy == TRUE && nutrition < 250 && prob(10)) //VOREStation edit for hunger noises.
+		var/growlsound = pick(hunger_sounds)
+		var/growlmultiplier = 100 - (nutrition / 250 * 100)
+		playsound(src, growlsound, vol = growlmultiplier, vary = 1, falloff = 0.1, ignore_walls = FALSE, preference = /datum/client_preference/digestion_noises)
+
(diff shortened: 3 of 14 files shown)


code/modules/vore/eating/bellymodes_vr.dm
```diff
@@ -472,5 +472,7 @@
 			put_in_egg(P,1)
 
 	if(play_sound)
+		for (var/mob/living/M in contents)
+			M << sound(play_sound)
```
Poojawa: You can use M.playsound_local(loc, play_sound) here

Verkister: Dunno I feel like something's broke the get_turf which now is get_step instead which from inside the belly obj pointed to nullspace instead or someshit and that's what I think was the reason for the prey not hearing them at all after the update changed it from direct to local. Found out the old get_turf that actually did account for the locs var was scrapped nearly 2 years ago or something.

Verkister: Yeah didn't work. Testing with M.loc now

Poojawa: weeeird. 

`M.playsound_local(get_turf(M), play_sound, 65)` should work

Verkister: bah kinda redundant anyway as all it would add would be distance/volume/falloff calculations for a prey that would literally be inside the thing in any case before doing the << thing anyway at the end

Poojawa: playsound_local can be set to no falloff. also if the prey is literally on top of the location, then falloff can't occur because they're on top of the source. 

But yee, the << will work just as well.

SpadesNeil: << will work, but IIRC, it's not the desired means of doing it.

@Arokha have any better ideas?

Verkister: That's pretty much what everything leads to.
https://github.com/VOREStation/VOREStation/blob/master/code/game/sound.dm#L90

code/modules/vore/eating/bellymodes_vr.dm
```diff
@@ -472,5 +472,8 @@
 			put_in_egg(P,1)
 
 	if(play_sound)
+		for (var/mob/living/M in contents)
+			if(M.client.is_preference_enabled(/datum/client_preference/digestion_noises))
+				M << sound(play_sound)
 		playsound(src, play_sound, vol = 100, vary = 1, falloff = VORE_SOUND_FALLOFF, ignore_walls = FALSE, preference = /datum/client_preference/digestion_noises)
```
Leshana: Okay I figured it out.  It is because of ye old issue where hearers() will not include people inside bellies.   It can be fixed by changing this line to `ignore_walls = TRUE`  then you won't need the loop.

KasparoVy: Good find

Verkister: Doesn't explain the playsound_local not working for them either tho.

Verkister: https://github.com/VOREStation/VOREStation/commit/2fc033ee0eeb76a2665465403c8c807871ee439a
Yeah I'm pretty sure the get_step thing here draws from the thing's xyz coordinates while the non-turf loc getting substituted with the working locs list feature was removed

Leshana: `get_turf()` works perfectly fine and is faster than that old method.   get_step() is the method used by all codebases for years.   If you don't believe me, try it.

Poojawa: also you may want to do `if(M.client && M.client.is_preference_enabled` just to ensure that byond actually knows there's a client there. it's an annoying thing.

Verkister: Well I'd provide this exact thing here with the deaf prey thing. The prey's coordinates inside the modern obj based belly is 0,0,0, and the playsound_local shrinks the volume to none from the distance from there to the sound source turf or at least that's what I think is happening. The ignore_walls part is only present at the playsound, and not in the playsound_local, whic has been tested completely dead for the prey as well.

Poojawa: so just call `M.playsound_local(get_turf(owner), play_sound, 100)` 

because the belly is inside of the predator, who owns it. and the prey are inside of the obj/belly, so if you get the owner of said belly's turf, you're fine. It's how I did it and y'all have the same sound code as citadel does.

Leshana: Calling `get_turf(prey)` where prey is inside a belly will return the /turf that the pred is standing on.

Verkister: Alright gonna test that when I have time. There's still the problem that the playsound calculations both are based also on the prey mob's get_turf hardcoded into the playsound_local as calling the get_turf(src) for the receiving mob.

Verkister: Alright gonna try the stuff when I get the chance. I still do kinda think that the ignore_walls thing was put there because of people getting annoyed by things carrying over from neighboring dorms to others and such.

Verkister: You were right. Thanks

PR #3189, opened 2018-03-07
Readies our borg stuff for next sync fixes.

Verkister:
-Fixes the default module icon missing from our _vr.dmi
-Makes our custom modules enable ui_style_vr to pull the hud icon from our dmi instead.
-Makes relogging etc not revert to default ui file if ui_style_vr is enabled for the borg.
-Moves the clumsy VS edit borg reset stuff to a more appropriate place.
-Adds the recently fixed attack anim offset fix and the ui_style_vr to the reset module.

-The rest of the fixes are waiting here for now https://github.com/PolarisSS13/Polaris/pull/5027


Verkister:
Not necessarily. The polaris borg ui dmis will stay broken until then though.

code/_onclick/hud/robot.dm
```diff
@@ -276,6 +276,8 @@ var/obj/screen/robot_inventory
 		r.client.screen -= r.robot_modules_background
 
 /mob/living/silicon/robot/update_hud()
-	..()
+	if(ui_style_vr) //VOREStation Edit
+		hands.icon = 'icons/mob/screen1_robot_vr.dmi'
 	if(modtype)
 		hands.icon_state = lowertext(modtype)
+	..()
```
Arokha: Should probably put //VOREStation Edit comments on this line too, or better yet, comment out the return at the top and leave it with an edit comment.

Or better-better-yet, put this in a crazy sideways override in the robot_vr.dm file and make sure to call ..() after

PR #3191, opened 2018-03-07
Armory tweak

Mewchild:
Re-adds the implantpad to the armory (along with a box of tracking and a box of chem implants_
Moves the hand labeller from the briefing room up into the green armory, near officer equipment.


Mewchild:
..It appears I did a fucking dumb, somehow, in my branching

PR #3194, opened 2018-03-08
Fix un-eqipping gear with a single click.

Leshana:
Restored this functionality to the way it was, with the exception of uniforms.  That drops everything when you remove it, so is still protected.


Arokha:
What does this do for suits with storage, given that they aren't uniforms? I assume it opens the storage, given that those are a child type of this and get called earlier (though hopefully don't call ..() before opening storage...), but just checking

Leshana:
@Arokha It works fine, the change is just putting it back the way it was, so it works noice.

PR #3195, opened 2018-03-08
Fixes an embarrassing belly update oversight

Verkister:
-Prey within gurgled prey no longer drops on the floor, but instead ends up in the pred's belly as supposed.
(not my blooper btw)


Verkister:
found even more of these. brb making a proper pr.

PR #3197, opened 2018-03-08
Some more vore fixes.

Verkister:
-Fixes gurgled prey's own prey dropping onto the floor instead of pred's gut.
-Fixes itemweak gurgles not getting past first layer of containment.
-Fixes entered belly contents from gurgled things already in the same gut causing noises and alerts. (such as prey loot and contents from digested storage items)
-Fixes strip digestion. No longer dumps everything at once and empties uniform pockets/id/belt as it used to.
-Strip digestion now also works on multiple prey at once and respects equip order. (as in oversuit goes before uniform etc)


code/modules/vore/eating/bellymodes_vr.dm
```diff
@@ -121,12 +121,22 @@
 			for(var/slot in slots)
 				var/obj/item/thingy = M.get_equipped_item(slot = slot)
 				if(thingy)
-					M.unEquip(thingy,force = TRUE)
-					thingy.forceMove(owner)
+					if(!M.canUnEquip(thingy))
+						continue
+					if(thingy == M.get_equipped_item(slot_w_uniform))
```
Arokha: Why do this? Seems more expensive than just looking at the slot you selected earlier, since if this is the case, slot would also equal slot_w_uniform (though this would be handled for you if you just fix unEquip)

code/modules/vore/eating/bellymodes_vr.dm
```diff
@@ -121,12 +121,22 @@
...
+						var/list/stash = list(slot_r_store,slot_l_store,slot_wear_id,slot_belt)
+						for(var/stashslot in stash)
+							var/obj/item/SL = M.get_equipped_item(stashslot)
+							if(SL)
+								SL.forceMove(src)
+								contents |= SL
```
Arokha: May call Entered twice for this item. Contents is a Byond-managed list, there's no need to ever update it manually (nor should you try)

code/modules/vore/eating/bellymodes_vr.dm
```diff
@@ -121,12 +121,22 @@
...
-					M.unEquip(thingy,force = TRUE)
-					thingy.forceMove(owner)
+					if(!M.canUnEquip(thingy))
+						continue
+					if(thingy == M.get_equipped_item(slot_w_uniform))
+						var/list/stash = list(slot_r_store,slot_l_store,slot_wear_id,slot_belt)
+						for(var/stashslot in stash)
+							var/obj/item/SL = M.get_equipped_item(stashslot)
+							if(SL)
+								SL.forceMove(src)
```
Arokha: You aren't calling u_equip on these like you are for the main item, which will result in them remaining on the screen of the prey.

code/modules/vore/eating/belly_obj_vr.dm
```diff
@@ -350,7 +349,7 @@
 			for(var/slot in slots)
 				var/obj/item/thingy = M.get_equipped_item(slot = slot)
 				if(thingy)
-					M.unEquip(thingy,force = TRUE)
+					M.u_equip(thingy)
```
Arokha: The real fix for this is making unEquip use drop_location(), this is sort of a hack I'm not sure if I like.

code/modules/vore/eating/contaminate_vr.dm
```diff
@@ -18,7 +18,7 @@ var/image/gurgled_overlay = image('icons/effects/sludgeoverlay_vr.dmi')
 		name = "[gurgleflavor] [cleanname]"
 		desc = "[cleandesc] It seems to be covered in ominously foul residue and needs a wash."
 		for(var/obj/item/O in contents)
-			gurgle_contaminate(item_storage)
+			O.gurgle_contaminate(item_storage)
```
Arokha: Sorry, did everything but this fix in this PR my way instead, can you fix the merge conflict and I can merge this part?

PR #3198, opened 2018-03-08
Implement drop_location(), some vore fixes

Arokha:
Duplicates a few things in @Verkister 's https://github.com/VOREStation/VOREStation/pull/3197 though done in a way I think is more efficient in terms of being futureproof.

Doesn't duplicate the gurglecontaminate fix from that PR though.

### DIFF
--- a/code/game/area/areas.dm
+++ b/code/game/area/areas.dm
@@ -339,3 +339,9 @@ var/list/mob/living/forced_ambiance_list = new
 
 /area/proc/shuttle_departed()
 	return TRUE
+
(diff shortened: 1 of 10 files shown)


Arokha:
To be honest, no, but that's something that I'd like to change in the future. Right now when you're in objects, you tend to be unable to drop items, or change your equipment. I assume it's related to canmove or something like that. I'd be interested in changing that.

Arokha:
Returning true on that now, even though you can't drop items. Something to fix for the future.

PR #3200, opened 2018-03-10
Reduce qdel log file by 150000%

Leshana:
Fix indenting error that caused it to log repeatedly.  Also put it in its own file to match the pattern used by overlays and initialize.


Leshana:
Sent to polaris

PR #3201, opened 2018-03-10
Robot module change animations (fxd Edition)

Heroman3003:
Okay. Found the reason for animation breaking when resetting module from nonstandard sprite file. Bug fixed, all animations should be working all right now.


Arokha:
Conflicts, plus lots of crazy commits. Please clean it up.

PR #3206, opened 2018-03-10
Gives RD explorer comms

Mewchild:
Title. Adds explorer comms to RD headset


aztectornado:
This hopefully will cut down on the number of times explorers are ignored/not noticed by the RD.

...I still don't understand why this wasn't already a thing.

Mewchild:
To clarify on why I made this PR: https://i.imgur.com/i7exOBL.png

Arokha:
Was just an oversight. Polaris doesn't have Explorers as part of research, they are their own 'planetary' department based on their Sif planet. We don't have that (since we don't have a single 'planet' to be exploring like that) so they're part of Science and I/nobody had the idea to add Explorer comms to RD, though I did add Science comms to Explorer.

PR #3208, opened 2018-03-10
Prevent ghosts from ejecting borgs from rechargers

Arokha:


Cameron653:
Should probably send this to Polaris.

Arokha:
Basically any of my commits that have the `POLARIS:` at the start are destined for that.

PR #3210, opened 2018-03-11
Module change and reset animations for borgs

Heroman3003:
Should be mergeable now.


code/game/objects/effects/temporary_visuials/temproary_visual.dm
```diff
@@ -0,0 +1,39 @@
...
+/obj/effect/temp_visual/dir_setting
+	randomdir = FALSE
+
+/obj/effect/temp_visual/dir_setting/Initialize(mapload, set_dir)
+	if(set_dir)
+		setDir(set_dir)
+	. = ..()
+*/		//More tg stuff that might be useful later
+
+/obj/effect/temp_visual/proc/setDir(var/newdir)
```
Leshana: We already have `/atom/proc/set_dir(new_dir)`  please remove this and use that instead.

code/modules/mob/living/silicon/robot/robot.dm
```diff
@@ -970,6 +971,12 @@
 			icontype = module_sprites[1]
 	else
 		icontype = input("Select an icon! [triesleft ? "You have [triesleft] more chance\s." : "This is your last try."]", "Robot Icon", icontype, null) in module_sprites
+		if(notransform)				//VOREStation edit start: sprite animation
+			src << "Your current transformation has not finished yet!"
```
Leshana: Please use to_chat()

code/modules/mob/living/silicon/robot/robot_modules/station.dm
```diff
@@ -54,6 +54,7 @@ var/global/list/robot_modules = list(
 	vr_add_sprites() //Vorestation Edit: For vorestation only sprites
 
 	R.set_module_sprites(sprites)
+	sleep(50) //VOREStation edit: sprite animation
```
Leshana: I'm cautious about any sort of sleep in a `New()` proc.   Are 100% sure this object will never be present somewhere mapped in? Is it only created at runtime?

Heroman3003: Yes, robot_module existing on the map is a mistake of a mapper, as it is only supposed to exist as part of borg and be tightly tied to it. Otherwise it is only created upon module choice or, for borgs with pre-installed modules, upon intialization.

code/game/objects/effects/temporary_visuials/temproary_visual.dm
```diff
@@ -0,0 +1,36 @@
...
+	icon = 'icons/effects/effects.dmi'
+	icon_state = "nothing"
+	anchored = TRUE
+	layer = ABOVE_MOB_LAYER
+	mouse_opacity = 0
+	var/duration = 10 //in deciseconds
+	var/randomdir = TRUE
+
+/obj/effect/temp_visual/initialize()
+	//. = ..()
```
Leshana: Sorry, didn't notice this before, but initialize needs to call its parent and return an init hint.  (easiest way to do this is to do . = ..()

And that TG code looks nice to me, so I recommend adding it back in.
```
	. = ..()
	if(randomdir)
		set_dir(pick(global.cardinal))
```

code/modules/mob/living/silicon/robot/robot_modules/station.dm
Heroman3003: Okay, apparently I fixed the thing that required that sleep(), but simply didnt remove it. Fixed now.

PR #3213, opened 2018-03-11
Added integrated radio circuit

Heroman3003:
Added integrated radio circuit.
Inputs: Channel(TEXT), Message(TEXT)
Outputs: Channel(TEXT), Speaker(TEXT), Message(TEXT)
Activation pins: Send Message(IN), On Message Received(OUT), On Translation(OUT)

To set available channels, swipe ID through the circuit. Channels are same as ones you can use on intercom with the access of that card. Common and Entertainment channels are always available.

Upon being pinged in through Send Message, will send input message into input channel.
Pings out On Message Received whenever it hears speech on one of available channels that does not come from another radio circuit.
Pings out On Translation whenever it hears speech in non-common language that it can translate on one of available channels.

Basically microphone, but radio-only.
Complexity is 25.

Right now review from both balance and implementation standpoint is requested


Arokha:
At the very least, needs some logging, especially tying who created it to the circuit, probably by doing a log_game whenever one is created, logging the ref of it and then whenever it speaks, logging the ref so the two can be correlated.

Heroman3003:
Closing for history cleanup.

PR #3214, opened 2018-03-11
Fixes #3209, fixes the cancel button in ears/wings/tail selection clearing the respective selection

### LINKED ISSUE #3209: Sign Language key is broken
Sign language, unlike all other languages, cannot presently be used without swapping it to be your default language.

say ",s Heckie

To make me say "Heckie" in sign language

#### What actually happened
The character said ",s Heckie"

#### Steps to reproduce
- (Step 1)Have sign
- (Step 2)type ,s before a message as with other languages
- (Step 3)confuse your friends

#### Code Revision
- release - 
377dea3bd4dfcffbe7bae0d14c7cc36c3cc2ee14
 
- Shadekin empathy also uses ,s for its key. I'd suspect there being some kind of conflict there. If I might make a suggestion for a new key for empathy, I'd say ,m

Little-119:
The shadekin empathy key is now "m" for mar. I've tested shadekin empathy, sign language, and medical radio to make sure they work. Fixes #3209

Also fixes the ears/wings/tail selection being cleared when the cancel button in the menu for selecting the respective thing is clicked. Also changes the name of the variable that stores the new selection to match the ones for hair, why not. (Sorry if those two sentences don't make any sense, I can't find better ways to word them.)

### DIFF
--- a/code/modules/client/preference_setup/vore/01_ears.dm
+++ b/code/modules/client/preference_setup/vore/01_ears.dm
@@ -186,8 +186,9 @@
 				pretty_ear_styles[instance.name] = path
 
 		// Present choice to user
-		var/selection = input(user, "Pick ears", "Character Preference") as null|anything in pretty_ear_styles
-		pref.ear_style = pretty_ear_styles[selection]
+		var/new_ear_style = input(user, "Pick ears", "Character Preference", pref.ear_style) as null|anything in pretty_ear_styles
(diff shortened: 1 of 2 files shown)


Little-119:
I should probably note that the last argument `input` was supposed to be removed, as it doesn't actually do anything, but I never pushed the changes. Not important enough for a PR though since it doesn't have any effect.

PR #3216, opened 2018-03-11
Makes the monstertamer/clonexadone reaction force hatching too

Screemonster:
It's a bit of an edge case, but there was a potential problem where a dead chimera could be sleevecarded by a doctor that didn't know any better, and because they're dead would be unable to sleeve their mind back into the body, thus leaving them unable to hit the revive button and taking them out of the round permanently without admin intervention.

Rather than yelling "no don't sleevecard m- oh for fuck's sake" in LOOC, figured it'd be better to let the existing reaction mash the buttons on the chimera's behalf. Now if you encounter a deadmera, you can happily sleevecard them, give them a whole bunch of injections, wait for them to bake at gas mark 5 for 20 minutes while you do teasy things to them in your nif or whatever, give them another injection to wake them up, then sleeve their mind back into their newly-hatched body.

I swear I'd catch these edge cases more often if I wasn't so risk-averse in game. Like seriously when was the last time Scree actually had to use his revive? I caught a bug in the code that time too.

### DIFF
--- a/code/modules/reagents/Chemistry-Recipes_vr.dm
+++ b/code/modules/reagents/Chemistry-Recipes_vr.dm
@@ -57,11 +57,19 @@
 	if(ishuman(holder.my_atom))
 		var/mob/living/carbon/human/H = holder.my_atom
 		if(H.stat == DEAD && (/mob/living/carbon/human/proc/begin_reconstitute_form in H.verbs)) //no magical regen for non-regenners, and can't force the reaction on live ones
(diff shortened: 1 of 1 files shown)


Mewchild:
Scree for expedition team engineer. Put him in the risk zone, give him more chance to use revive. Clear solution

KasparoVy:
Buff buff buff.

Screemonster:
It still takes up to 20 minutes for them to do their revive thing with a one-hour cooldown afterwards, so there's a _reason_ most chimera tend to be risk-averse compared to other, clonable, races :p

Issue #3199, opened 2018-03-09
Unused syringes cannot be digested.


Verkister:
#### Brief description of the issue
~~At least by borgs, but they now use the same system anyway.~~
**Confirmed broken for organic guts as well.**
Tested with maint drugs and roids, empty unused syringes, used syringes, and handled ones.
Seems only the used ones could be processed.

#### What you expected to happen
To be able to dispose of trash needles as before.

#### What actually happened
Needles kept accumulating and taking up more and more missed gurglecycles as they weren't getting flagged undigestable either. Ended up with a few drug shots, a roid one, and one unused tossed out by medical or something. Used ones behaved fine tho. At the roundend tested with a fresh box of syringes, which all became untouchable as well to confirm the case.

#### Steps to reproduce
- (Step 1) Be janipup.
- (Step 2) Find drug needles lying around in trash and do your job.
- (Step 3) Possibly beg medical for control samples for the other sort of needles, used and unused probably.
- (Step 4) Rev up that gut.
- (Step 5) Watch all other crap becoming fuel in a gradually slowing pace as the percentage of the contents made up by the "cleanable" needles grows until there is nothing but uninjected needles left.
- (Step 6) Ponder about how your gut is now essentially stuck processing with none of the remaining contents showing up as uncleanable either.

#### Code Revision
- 7d899aeb7f6cfc60cb51726c36688e91abc14042

#### Anything else you may wish to add:
- Tested as both, Goodboi the borg and later as Gabriella the RD

PR #3218, opened 2018-03-11
Runtime Fixes - March

Leshana:
- Fix DEBUG: Runtime in syringes_vr.dm,17: Cannot execute null.Cut()
- Fixes Runtime in item_spawning.dm,38: Cannot create objects of type null by removing obsolete custom items. There is no such type as `/obj/item/device/pda_mod`, and there has not ever been such a type on this codebase.  Last time it existed was in pre-github baystation code.


Verkister:
I feel like this is gonna fix #3199

PR #3220, opened 2018-03-12
Reduce lightweight trait point return

Arokha:
No single "Free Haste+" traits pls.

Given that it only has a 50% chance of happening, and causes no damage, this isn't worth -4.

Also this was basically 4 free points for any micro, because it completely ignored it for size-difference steps, thus resulting in a trait that did literally nothing, and gave you 4 points.

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/traits_vr/negative.dm
+++ b/code/modules/mob/living/carbon/human/species/station/traits_vr/negative.dm
@@ -110,7 +110,7 @@
 /datum/trait/lightweight
 	name = "Lightweight"
 	desc = "Your light weight and poor balance make you very susceptible to unhelpful bumping. Think of it like a bowling ball versus a pin."
-	cost = -4
+	cost = -2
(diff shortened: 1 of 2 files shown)


Little-119:
Realistic micros that can't reach, hold, or wear anything that's not specifically tailored to them when?

KasparoVy:
Two-pixel Tim rides again.

Arokha:
> Realistic micros that can't reach, hold, or wear anything that's not specifically tailored to them when?

I brought that up in the admin chat when I noticed this. :p

KasparoVy:
Can't hold that chainsaw, too damn big.

SilencedMP5A5:
100% support this because while lightweight does stun you for like 20 seconds (maybe a bit more or less???) if you're a hyperspeed micro, you don't need to worry about it in the slightest :V

aztectornado:
We need a micros nerf/rebalance

KasparoVy:
This'll be fun.

PR #3228, opened 2018-03-13
Block prometheans from using synthparts

Arokha:
Aside from being the most powergamey combination I can think of, it appears to make the game go insane trying to do all the special species things for you, PLUS all the special robolimbs things. It may make you explode randomly, which is probably not good, not to mention the verbs probably do completely insane things when you try to use them.


TheDavestDave:
That sounds like a neat mob type. 50:50 it can ignore or damage or blow up if something bumps into it. The rare glass tank.

PR #3229, opened 2018-03-13
Vore fixes and additions. (gurgledeath item drop, vorepanel load button, corrupt prettyboi)

Verkister:
-Fixes prey dropping their crap on pred's location instead of the gut on gurgledeath.
-Adds vorepanel feature to reload and apply belly prefs from current character slot. Handy for copying finished setups from other char slots, and to apply those on spawned mobs as well. Comes with a confirmation warning and ejects all existing bellies before rebuilding.
-Adds a new corrupt hound type based on yesterday's shenanigans.

https://i.gyazo.com/960761bfe4bb389c64fc698f74a3179b.gif


code/modules/vore/eating/belly_obj_vr.dm
```diff
@@ -441,6 +441,9 @@
 /obj/belly/AllowDrop()
 	return TRUE
 
+/obj/belly/onDropInto(var/atom/movable/AM)
+	return src
```
Arokha: Looks like onDropInto is a sort of half-kinda-sorta partial implementation of drop_location from Bay or something. It appears to only be used two places. Maybe just remove it and replace with drop_location?

Verkister: This would need changes to the dropInto proc as it happens to be one of those things critically using the onDropInto

Verkister: Actually I kinda have no idea how to do it. The things would have to be dropping into the prey mob's drop_location, and beyond remove_from_mob's dropInto there isn't much room for the prey mob. Possibly change the dropInto's "destination.onDropInto(src)" part into like "loc.drop_location" of the item?

Arokha: I had a really long post typed up, but then started looking at the /tg/ implementation that I was going to suggest using, and it's a lot of typing. I'll just do it after this PR I guess.

Verkister: Alright should work.

PR #3235, opened 2018-03-14
Adds an alt vulp tail style.

Cameron653:
It's the first frame of the, constantly wagging one the species gets by default.
https://i.imgur.com/vwyWqZR.png - Alt style
https://i.imgur.com/Y4yWsuM.png - The normal vulp tail that's selectable otherwise.


Little-119:
How's this different from `vulpkanin, colorable`? If the difference is that one wags, shouldn't the name indicate that that one wags and this one doesn't instead of the misleading 'alt style'?

Cameron653:
The vulpkanin colorable looks like https://i.imgur.com/q30pIlB.png. This is alt style doesn't have the tail jutting out.

PR #3239, opened 2018-03-15
Little more realistic gurgle gain tweak and oxyloss compensation.

Verkister:
Increases the preyhealth to nutrition ratio to 1/4.5.

Currently a full health same size average weight prey would only give maximum 400 nutrition to the pred (more or less or less when counting possible size/weight difference and ruthless oxyloss/hunger drain during slow gurgles), which in immersion is about worth a single plate of spaghetti or something. The range from absolutely starved (as the case in damn near every non grab&gulp scene there is) to the first time the grey burger appears is 450 nutrition. I'm getting so bummed by most of my non-micro/macro gurgle scenes still ending with the pred still at red burger levels lmao.

Now the total gain before inevitable taxes is around 900, which is closer to the originally intended number of 1050. After deducting the nutrition tax during the gurgle time, ~~the prey health oxyloss tax during the time between critical treshold and death~~(see below), the prey likely being smaller than the pred size difference tax, non-tanky prey health/resistance tax, and the skinny prey tax, a pred should now be able to finish their scene without getting left starving.

Also added prey oxyloss compensation. Researched up how crit oxyloss affects gurgles on full health default human. Default damage (2+2) oxylossed 5.2% of total gain, max damage (6+6) lossed 2.2%, tickle damage (0+1) lossed 12.5%.

Time to get **fat**


code/modules/vore/eating/bellymodes_vr.dm
```diff
@@ -85,9 +85,9 @@
 				var/mob/living/silicon/robot/R = owner
 				R.cell.charge += 20*(digest_brute+digest_burn)
 			if(offset) // If any different than default weight, multiply the % of offset.
-				owner.nutrition += offset*(2*(digest_brute+digest_burn)/difference) // 9.5 nutrition per digestion tick if they're 130 pounds and it's same size. 10.2 per digestion tick if they're 140 and it's same size. Etc etc.
+				owner.nutrition += offset*(3*(digest_brute+digest_burn)/difference) // 9.5 nutrition per digestion tick if they're 130 pounds and it's same size. 10.2 per digestion tick if they're 140 and it's same size. Etc etc.
```
KasparoVy: I suppose the comment on the end about nutrition per tick is no longer accurate (but has it ever been :thinking:)?

Verkister: Yeah hasn't really been since the customizable damage values.

SpadesNeil:
You might want to edit this even more. If you look at the examine.dm file it should contain descriptions for different levels of fullness. If I recall correctly, the intended nutrition gain from a full size, same size, default weight prey was 1050.

Leshana:
Okay, the problem with this is that now whenever you eat someone you're going to go into overfed fat mode and slow down, even if you were completely starving beforehand.  Isn't this *way* too much nutrition to gain?

Verkister:
You're eating a whole person.

code/modules/vore/eating/bellymodes_vr.dm
```diff
@@ -51,6 +51,7 @@
 			if(M.stat == DEAD)
 				var/digest_alert_owner = pick(digest_messages_owner)
 				var/digest_alert_prey = pick(digest_messages_prey)
+				var/compensation = M.getOxyLoss()
```
Leshana: I had no idea what this compensation thing was about until someone else explained it.  Its weird enough that it requires documenting in comments I'd say.

Verkister: Roger that

Arokha:
"Slight" != 210%

Verkister:
Alright. The borg number does look kinda big and I'm gonna half it down to being under 1 default cell capacity (7500).
What comes to humans though, I'm gonna take the number 10% lower, to around 1000. If you're going to idle the whole round, you should just cryo out as a human. Thing is a single liquidfood ration is already worth 600 nutrition, and I can tell I've managed to expend more than one of those each round even with my slowest metabolism characters easily.

Verkister:
And what comes to fatness, I beg to differ.
kuva
This character of mine was sort of an experiment, going from default weight to this all by ingame gains (not leaving the round bloated) withing the occasional shifts in the span of like 3 days. Granted he has apex metabolism, which also scales the weight gain/loss with the metabolism speed, but the rate is the same even if the slower modes failed to pudgify the full load during the round.

Verkister:
This becoming one of those PRs all maintainers are stuck playing the hot potato of "not my job" with?

lm40:
pls merge tyty, tired of starving after digesting an entire person

SpadesNeil:
@Arokha :U Are the numbers still bad?

All I know is the numbers should look like this.

1. Start at neutral nutrition value as a normal size character.
2. Eat one whole person of base weight and normal size, fully digest them.
3. Nutrition overflow should result in a weight gain of at least 26 pounds, assuming the person jumps into cryo or the server ends right that second.

The way the system was supposed to work is that excess nutrition is calculated and dumped into your body weight at round end. The rate of weight gain during the shift is **supposed** to be obscenely low.

SpadesNeil:
This has been sitting in limbo for a while. Can we get a followup, Arokha?

Verkister:
7th week added to limbo.

code/modules/vore/eating/bellymodes_vr.dm
```diff
@@ -113,7 +114,14 @@
 				play_sound = pick(death_sounds)
 				digestion_death(M)
```
Arokha: Honestly digestion_death() should probably handle all this but this is fine for now. Refactor later.

PR #3242, opened 2018-03-16
Communicator + NIF Refactor

Arokha:
Refactors NIFs slightly:
* They now save your commlink name, so you can be named "FoxMaster 9000" each shift or whatever.
* Now work properly on species that don't have their brains in their head (and expect to be put into whatever organ that has the brain)

Refactors communicators slightly:
* Allows you to set the name without providing a mob (no mob-related features/vars were ever accessed so it was basically pointless)

### DIFF
--- a/code/game/objects/items/devices/communicator/UI.dm
+++ b/code/game/objects/items/devices/communicator/UI.dm
@@ -133,11 +133,7 @@
 	if(href_list["rename"])
 		var/new_name = sanitizeSafe(input(usr,"Please enter your name.","Communicator",usr.name) )
 		if(new_name)
-			owner = new_name
-			name = "[owner]'s [initial(name)]"
-			if(camera)
-				camera.name = name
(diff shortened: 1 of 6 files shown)


Dhaeleena:
Can you make this happen with PDA ringtone?

Arokha:
Nope

Little-119:
> Now work properly on species that don't have their brains in their head (and expect to be put into whatever organ that has the brain)

Does this mean prometheans don't need the bioadaptive NIFs?

Arokha:
> Does this mean prometheans don't need the bioadaptive NIFs?

No, that's a gameplay thing, not a bug. It does mean you need to put them into their torso now though.

Little-119:
Seems odd to me that you need to fax CC to get an admin to bluespace one in, in that case.

Arokha:
Not super-related to this PR, tbh.

PR #3251, opened 2018-03-16
Heal robolimbs when you heal everything

Arokha:
Adds include_robo in adjustBruteLoss/adjustFireLoss so that you can specify if you want to heal robolimbs too, which by default, are ignored.

### DIFF
--- a/code/modules/mob/living/carbon/human/human_damage.dm
+++ b/code/modules/mob/living/carbon/human/human_damage.dm
@@ -108,7 +108,8 @@
 		amount += O.burn_dam
 	return amount
 
-/mob/living/carbon/human/adjustBruteLoss(var/amount)
+//'include_robo' only applies to healing, for legacy purposes, as all damage typically hurts both types of organs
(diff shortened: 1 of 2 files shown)


Cyantime:
"Only magic and stuff" what
heal_overall_damage is used whenever adjustBruteLoss or adjustFireLoss is negative, in changeling revives, diona light healing, slimepeople healing from slime jelly

Arokha:
> slimepeople healing from slime jelly

Irrelevant since they can't use synthlimbs anymore

> diona light healing

I should add them to that list too

> adjustBruteLoss or adjustFireLoss is negative

Hmm, didn't notice that. Suppose I'll just give them an additional var unless you have another suggestion.

Arokha:
In looking, this is soooo close to an infinite loop. Human `adjustBruteLoss` calls `heal_overall_damage`, which on humans, does this stuff.

On /mob/living, `heal_overall_damage` calls `adjustBruteLoss`. Somewhere along the way, things went backwards.

PR #3252, opened 2018-03-16
Protean species

Arokha:
Adds a **whitelisted** species, Proteans. Unlike all the other species in which the word means 'Society' for some reason, this one is both closer to and further from the actual meaning of 'species' at once.

## Lore
Proteans are people who have opted to sleeve into a fully synethetic body made up of nanotech, essentially making them big piles of animated dust that can take various shapes. In the case of Proteans, they can take the shape of any of the robolimb manufacturer parts.

They are not from any particular culture or society, but rather just a generic term for any sort of nanoswarm body that might exist. They could be from Kitsuhana, or SPESS ELVES, or wherever.

NanoTrasen is very paranoid about nanotech operating at this level, and requires all Proteans to obtain a P.A.N. card, or 'Permit for Advanced Nanotechnology', which must be renewed monthly at CentCom, to allow them on the premises. A protean must keep this card on them at all times while on NanoTrasen property, and can be removed from the station if they don't appear to be licensed to operate what is effectively their body.

Typically societies that produce things like proteans can also handle resleeving, or things like resleeving, which is good since whitelisted species cannot be body-printed on station. Might I suggest a life crystal?~

## Body
They are exceptionally weak to burn and electric and EMP attacks. If you have a lit torch, you can slice off a limb with each click. A single lethal laser shot will typically cripple them. By contrast, they are basically impossible to punch to death. Punching piles of dust doesn't really work, you know?

They can drop their humanoid form to turn into an amorphous blob of nanites which can do sneaky scary things like ventcrawl and hide under tables, and spread itself over floor tiles to blend in. Ventcrawling is obviously *a thing* which is one of the reasons this is whitelisted, and why if you were to be whitelisted and then abuse that in ways other players find distasteful, you might find yourself losing that whitelist.

They have three internal organs:
* Protean posibrain - A more advanced posibrain capable of operating the other two organs in the body. Is a vital organ.
* Orchestrator module - A highly-parallel compute module capable of directing the nanites in coordination to perform tasks and take shapes. Is a vital organ.
* Refactory module - The module that converts metals into nanites for various purposes. Can consume steel to produce more 'normal' nanites but other metals can produce other effects. Not a vital organ.

## Metal
They 'eat' metal, and various metals have various effects, with steel being the basic thing they need. Steel lets them regenerate, mhydrogen makes them move faster, diamond increases their burn resistance, plasteel increases their brute resistance. I can add more in the future, but not right now.

Not sure what else I could do, feel free to suggest stuff, but I won't implement it as part of this PR since it's already big enough.

## Bugs
I should mention that this needs testing, and you shouldn't apply for a whitelist thing for a couple of weeks until I've actually found and fixed all the bugs. I've fixed the ones I've found, but there are further plans I have for the species that I haven't had time to implement, and I'd like to go ahead and put this work here until whenever I feel like doing those things.


Arokha:
This depends on some of the other PRs before it and will fail to compile until those are merged

code/modules/mob/living/carbon/human/species/station/protean_vr/protean_species.dm
```diff
@@ -0,0 +1,299 @@
...
+	return ..() //Wut
+
+/datum/species/protean/get_blood_colour(var/mob/living/carbon/human/H)
+	return rgb(80,80,80,230)
+
+/datum/species/protean/get_flesh_colour(var/mob/living/carbon/human/H)
+	return rgb(80,80,80,230)
+
+/datum/species/protean/handle_death(var/mob/living/carbon/human/H)
+	to_chat(H,"<span class='warning'>You died as a Protean. Please sit out of the round for at least 30 minutes before respawning, to represent the time it would take to ship a new-you to the station.</span>")
```
Screemonster: Whoa. I didn't know this was a thing and should probably add something similar to chimeras.

Little-119:
So, sort of like prometheans but nanomachines instead of goo?

Arokha:
Basically, yup. A little fancier, comes with abilities besides shapeshifting and stuff.

SilverTalismen:
I would love to apply for these guys, seems very interesting, What would one of those feel like anyway?

PR #3265, opened 2018-03-17
"0x32" is not a valid single-bit flag

Arokha:
It's just not, okay

### DIFF
--- a/code/__defines/mobs.dm
+++ b/code/__defines/mobs.dm
@@ -10,7 +10,7 @@
 #define CANPUSH     0x8
 #define LEAPING     0x10
 #define HIDING		0x20
-#define PASSEMOTES  0x32    // Mob has a cortical borer or holders inside of it that need to see emotes.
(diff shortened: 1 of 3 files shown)


KasparoVy:
Lmfao, I'm glad 0x20 was.. Whew, bullet dodged.

Atermonera:
`0x` indicates that it's a string of hexadecimal, which is 4 binary bits to a bit of hexadec. 
`0x20 = 0010 0000`
`0x32 = 0011 0010`
As you can see, `0x32` occupies 3 bits, and so is only useful if you have a special case that's true when `0x20, 0x10, 0x02` are all true.

Arokha:
Yes, I know, however the other bits it included had nothing to do with what it was used for. It was an obvious accident. 32 is a power of 2 and can be used as a flag, but someone had mistakenly just put 0x or not converted it at some point when converting them from 1 2 4 8 etc.

It's only used by cortical borers and none of the other bits it set made any sense with them.

PR #3278, opened 2018-03-18
Adds a pref verb to switch your scaling mode ingame.

Verkister:
-You no longer have to endure the respawn hassle in order to find out which mode fits your character better.


Arokha:
Does this take effect immediately if you are scaled, or do you need to apply a transform for it to update?

Verkister:
Worked immediately on my test runs.

PR #3281, opened 2018-03-18
QoL fixes/additions

Mewchild:
Added door buttons to the RD and CE offices
Added a filing cabinet to the expedition hangar
Added paperwork machines to RD office


Arokha:
I assume the buttons aren't ON the windows/walls and are actually inside the offices, pixel-moved to appear on the walls, so people outside can't push them? Also I see there's no specialfunction (or whatever that var is) on the buttons, is the default to just open the doors?

Mewchild:
The id(on the button) and ID_Tag(on the door) vars matching makes the door open. I used the HoS Door control button as a reference for this.

The buttons themselves are on the head of staff desk, so they can remain seated and allow people in/out of their office

Arokha:
Mmk

PR #3288, opened 2018-03-18
Shadekin convert nutrition into energy

Arokha:


SilverTalismen:
What is a Shadekin?

Arokha:
`desc = "Some sort of fluffer. Big ears, long tail. "`

SilverTalismen:
Ah okay then.

Arokha:
(They're animalthings we use for events that can phase shift and stuff)

code/modules/mob/living/simple_animal/vore/shadekin/shadekin.dm
```diff
@@ -191,6 +191,11 @@
 	. = ..()
 	if(ability_flags & AB_PHASE_SHIFTED)
 		density = FALSE
+
```
Leshana: I suggest they only do dis if dey be non-dead.

PR #3296, opened 2018-03-19
Ody reagent whitelist

Arokha:
It can analyze and produce anything in the chem dispenser, plus some medical drugs, but mostly nothing with phoron, except Dexalin.

### DIFF
--- a/code/game/mecha/equipment/tools/medical_tools.dm
+++ b/code/game/mecha/equipment/tools/medical_tools.dm
@@ -404,245 +404,293 @@
 	origin_tech = list(TECH_MATERIAL = 3, TECH_BIO = 4, TECH_MAGNET = 4, TECH_DATA = 3)
 	required_type = /obj/mecha/medical
 
-	New()
-		..()
-		flags |= NOREACT
-		syringes = new
-		known_reagents = list("inaprovaline"="Inaprovaline","anti_toxin"="Dylovene")
-		processed_reagents = new
-		create_reagents(max_volume)
-		synth = new (list(src),0)
+	//This is a list of datums so as to allow id changes, and force compile errors if removed.
+	var/static/list/allowed_reagents = list(
+		//Med Reagents,
+		/datum/reagent/inaprovaline,
+		/datum/reagent/bicaridine,
+		/datum/reagent/kelotane,
+		/datum/reagent/dermaline,
+		/datum/reagent/dylovene,
+		/datum/reagent/dexalin,
+		/datum/reagent/tricordrazine,
+		/datum/reagent/paracetamol,
+		/datum/reagent/tramadol,
+		/datum/reagent/spaceacillin,
+		//Dispenser Reagents,
+		/datum/reagent/water,
+		/datum/reagent/aluminum,
+		/datum/reagent/carbon,
+		/datum/reagent/chlorine,
+		/datum/reagent/copper,
+		/datum/reagent/ethanol,
+		/datum/reagent/fluorine,
+		/datum/reagent/hydrogen,
+		/datum/reagent/iron,
+		/datum/reagent/lithium,
+		/datum/reagent/mercury,
+		/datum/reagent/nitrogen,
+		/datum/reagent/oxygen,
+		/datum/reagent/phosphorus,
+		/datum/reagent/potassium,
+		/datum/reagent/radium,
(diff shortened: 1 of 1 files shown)


Wickedtemp:
Add soporific and ickypak for Virgo's specifically. Otherwise, there's pretty much no reason at all to ever use the syringe-gun.

Arokha:
Why would you want an ickypack-equipped Ody, what sort of 'mass predation' events are you encountering

Wickedtemp:
Because the only time the Ody's syringe gun has ever been useful outside of producing reagents has been when it's used to either sedate somebody or force somebody to 'release' another somebody, from my own experience. So I think having those to reagents available on the whitelist would go a little ways toward keeping the syringe gun actually in-use.

Arokha:
Why not use the normal syringe gun? I guess the question is, why do you need to use the Ody specifically for those rather than just the normal syringe gun. What benefit does it have, aside from "I get infinite reagents of those things"

Wickedtemp:
I'm not saying it's necessary, I'm saying it's the only thing I've ever had to actually use them for, and that, well, without these two reagents at it's disposal, I'd probably not even bother ordering an Ody with the syringe gun on it. The bonus over the syringe gun being rapid-fire in case one misses, and the fact you don't have to prepare each syringe into a dart.

Arokha:
I've never gotten any particular use out of the syringe gun in the 3+ years I've played medical here, except for times where I've convinced myself I should use it to justify it's rather limited existence. I've used the sleepers, by comparison, plenty of times.

I'd replace it with some sort of Bic/Kel/Dex/Dyl autoinjector if I didn't think people would ree.

nerdass:
Science uses the oddy fairly regularly for creating glycerol / corn oil when making charges for mining to go blast mining. Could Glycerol be added to the list to keep that alive?

Arokha:
No, that's not what the Ody is for. It's not for "dupe reagents that are hard to get", I think you're missing the point of the PR, which is to mechanically stop people from doing that.

If glycerol is hard to get, then just make an issue explaining how you normally make it, how long it takes, why you think it should be changed, etc, and someone can improve that.

code/game/mecha/equipment/tools/medical_tools.dm
```diff
@@ -404,245 +404,293 @@
...
+	if(!A.reagents || istype(A,/mob))
+		occupant_message("<span class=\"alert\">No reagent info gained from [A].</span>")
 		return 0
+	occupant_message("Analyzing reagents...")
+	for(var/datum/reagent/R in A.reagents.reagent_list)
+		if(R.reagent_state == 2 && add_known_reagent(R.id,R.name))
+			occupant_message("Reagent analyzed, identified as [R.name] and added to database.")
+			send_byjax(chassis.occupant,"msyringegun.browser","reagents_form",get_reagents_form())
+		else
+			occupant_message("Reagent \"[R.name]\" too complex to analyze, skipping.")
```
Leshana: I believe this message will print if you try to analyze a container of something you already analyzed.   So click twice on something and it will say
_Reagent analyzed, identified as milk and added to database._
_Reagent "milk" too complex to analyze, skipping_

Natje0:
You say that this isn't what the Ody is for.
Then, WHAT is it for? It's damn near useless for paramedical use, due to the clunkyness, elevator bottlenecking fast response, airlocks if they're outside, and generally being un-userfriendly. Its syringegun internals literally state it's SYNTHESIZING medicine. It is a synthesizer.

How is synthesizing chemicals not intended for a synthesizer?
It is the only thing that makes an Odyssey useful, so why take that away? Because a tiny minority of players have made use of it which has caused no harm (But rather the opposite) to anyone?

I'm opposed to this change, as it invalidates an already rarely utilized tool.

Arokha:
Based on your comments in the adminchat after this I assume you've altered your opinion.

Natje0:
Correct. If a Reagent generator that doesn't have massive restrictions is added, the Odyssey's replaced with a better, more fitting option that makes a whole lot more sense. I'm curious to see how that'll end up being like...

code/game/mecha/equipment/tools/medical_tools.dm
Arokha: Ok, altered the branching there. How's that?

PR #3300, opened 2018-03-19
Adds a verb to switch tail layer.

Verkister:
IC tab. No issues encountered on test runs.

1885e8eb5562d9018d4773fb1c1b7752

### DIFF
--- a/code/modules/mob/living/carbon/human/update_icons.dm
+++ b/code/modules/mob/living/carbon/human/update_icons.dm
@@ -10,10 +10,10 @@ var/global/list/damage_icon_parts = list() //see UpdateDamageIcon()
 ////////////////////////////////////////////////////////////////////////////////////////////////
 // # Human Icon Updating System
 //
-// This system takes care of the "icon" for human mobs.  Of course humans don't just have a single 
-// icon+icon_state, but a combination of dozens of little sprites including including the body, 
+// This system takes care of the "icon" for human mobs.  Of course humans don't just have a single
+// icon+icon_state, but a combination of dozens of little sprites including including the body,
 // clothing, equipment, in-universe HUD images, etc.
-// 
+//
 // # Basic Operation
 // Whenever you do something that should update the on-mob appearance of a worn or held item, You
 // will need to call the relevant update_inv_* proc. All of these are named after the variable they
@@ -25,14 +25,14 @@ var/global/list/damage_icon_parts = list() //see UpdateDamageIcon()
 // only get assigned to the mob once per tick.)
 // As a corrolary, this means users of this system do NOT need to tell the system when you're done
 // making changes.
-// 
+//
 // There are also these special cases:
 //  update_icons_body()	//Handles updating your mob's icon to reflect their gender/race/complexion etc
 //  UpdateDamageIcon()	//Handles damage overlays for brute/burn damage //(will rename this when I geta round to it) ~Carn
 //  update_skin()		//Handles updating skin for species that have a skin overlay.
 //  update_bloodied()	//Handles adding/clearing the blood overlays for hands & feet.  Call when bloodied or cleaned.
 //  update_underwear()	//Handles updating the sprite for underwear.
-//  update_hair()		//Handles updating your hair and eyes overlay 
+//  update_hair()		//Handles updating your hair and eyes overlay
 //  update_mutations()	//Handles updating your appearance for certain mutations.  e.g TK head-glows
 //  update_fire()		//Handles overlay from being on fire.
 //  update_water()		//Handles overlay from being submerged.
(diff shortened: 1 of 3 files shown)


code/modules/mob/living/carbon/human/update_icons.dm
```diff
@@ -10,10 +10,10 @@ var/global/list/damage_icon_parts = list() //see UpdateDamageIcon()
 ////////////////////////////////////////////////////////////////////////////////////////////////
 // # Human Icon Updating System
 //
-// This system takes care of the "icon" for human mobs.  Of course humans don't just have a single 
-// icon+icon_state, but a combination of dozens of little sprites including including the body, 
+// This system takes care of the "icon" for human mobs.  Of course humans don't just have a single
+// icon+icon_state, but a combination of dozens of little sprites including including the body,
 // clothing, equipment, in-universe HUD images, etc.
-// 
+//
```
Arokha: All these

code/modules/mob/living/carbon/human/update_icons.dm
```diff
@@ -25,14 +25,14 @@ var/global/list/damage_icon_parts = list() //see UpdateDamageIcon()
 // only get assigned to the mob once per tick.)
 // As a corrolary, this means users of this system do NOT need to tell the system when you're done
 // making changes.
-// 
+//
```
Arokha: Random whitespace

code/modules/mob/living/carbon/human/update_icons.dm
```diff
@@ -25,14 +25,14 @@ var/global/list/damage_icon_parts = list() //see UpdateDamageIcon()
...
-// 
+//
 // There are also these special cases:
 //  update_icons_body()	//Handles updating your mob's icon to reflect their gender/race/complexion etc
 //  UpdateDamageIcon()	//Handles damage overlays for brute/burn damage //(will rename this when I geta round to it) ~Carn
 //  update_skin()		//Handles updating skin for species that have a skin overlay.
 //  update_bloodied()	//Handles adding/clearing the blood overlays for hands & feet.  Call when bloodied or cleaned.
 //  update_underwear()	//Handles updating the sprite for underwear.
-//  update_hair()		//Handles updating your hair and eyes overlay 
+//  update_hair()		//Handles updating your hair and eyes overlay
```
Arokha: Changes are

code/modules/mob/living/carbon/human/update_icons.dm
```diff
@@ -337,7 +338,7 @@ var/global/list/damage_icon_parts = list() //see UpdateDamageIcon()
 		return
 
 	remove_layer(SKIN_LAYER)
-	
+
```
Arokha: Not okay for

code/modules/mob/living/carbon/human/update_icons.dm
```diff
@@ -558,7 +559,7 @@ var/global/list/damage_icon_parts = list() //see UpdateDamageIcon()
 
 	//Shoes can be affected by uniform being drawn onto them
 	update_inv_shoes()
-	
+
```
Arokha: A Polaris .dm file

Verkister: I don't really even know what's doing those. Some sorta setting in the editor I've somehow managed to hit?

KasparoVy:
Satan's own whitespace

code/modules/mob/living/carbon/human/update_icons.dm
```diff
@@ -879,17 +883,18 @@ var/global/list/damage_icon_parts = list() //see UpdateDamageIcon()
 	return tail_icon
 
 /mob/living/carbon/human/proc/set_tail_state(var/t_state)
-	var/image/tail_overlay = overlays_standing[TAIL_LAYER]
+	var/used_tail_layer = tail_alt ? TAIL_LAYER_ALT : TAIL_LAYER // VOREStation Edit - Alt Tail Layer
+	var/image/tail_overlay = overlays_standing[used_tail_layer] // VOREStation Edit - Alt Tail Layer
 	
-	remove_layer(TAIL_LAYER)
+	remove_layer(used_tail_layer) // VOREStation Edit - Alt Tail Layer
```
Arokha: Not really consistent with the other uses of alt layering. Might not actually work. Consider the following:
1) tail_alt is false, your tail is in TAIL_LAYER
2) You use the verb, tail_alt is inverted, and set_tail_state is called
3) set_tail_state then clears TAIL_LAYER_ALT out of overlays_standing and cuts it off the mob, leaving TAIL_LAYER, which remains populated with an image, and in your overlays, when really it should clear both, because: We're not using the old layer now, and we're about to update the new one so it needs to be gone

See belts and shoes, they both clear both potential layers at the top to avoid this case.

Verkister: @Leshana

Arokha:
Leaving this alone for a sec while Leshana ports stuff to/from Polaris, so it doesn't cause merge conflicts, but it seems fine to me for now.

PR #3310, opened 2018-03-20
Reduces lightweight knockdown time

Verkister:
It was never meant to be debilitating, just inconvenient but something messed up with the calculations. Even less reason to keep it this awful after the point return decrease.


Arokha:
Weakened is reduced by 1 every Life() tick, so 20 would be 20*2 seconds. 5 is 10 seconds.

PR #3315, opened 2018-03-21
Links up a stranded atmos section

Mewchild:
Links the dorms area into master atmos


Mewchild:
@Arokha The thought occurs: Are the vents in the dorms blacklisted from things such as spiders, grubs, and the like? Can a mouse crawl into the dorms if it's player controlled? This may be the reason that section was isolated

Arokha:
It was isolated due to accident. Mice interrupting scenes is ahelpable. Solargrubs and such avoid the dorms, yes.

PR #3322, opened 2018-03-22
Change to Teshari to test for bug

Arokha:
Removing these, to see if they are related to the Teshari bug. These are the only two unique sprite features that I can see.


KasparoVy:
Couldn't crash anyone with my crashy character in testing today after this merge.

PR #3323, opened 2018-03-22
Amends body_hair Caching and Fixes Teshari Tail Hair Rendering

KasparoVy:
Was inspired to write this PR after witnessing a fiesta of crashes (bad enough to cue "Let the Bodies hit the Floor" and have it actually work for the situation) 😁.

The amendment to `body_hair` caching is purely speculative, but I saw no reason why limbs with `body_hair` shouldn't use the standard caching methods in limb icon generation. Mob health-doll still works just fine. _Can port to Polaris_

This PR also fixes Teshari tail hair not rendering at all. Polaris has this same issue from what I can tell by looking at their code. _Can port to Polaris_

_Teshari Tail Hair Busted_
tesharitailhaircolourbusted

_Teshari Tail Hair Fixed_
tesharitailhaircolourfixed

### DIFF
--- a/code/modules/organs/organ_icon.dm
+++ b/code/modules/organs/organ_icon.dm
@@ -159,12 +159,10 @@ var/global/list/limb_icon_cache = list()
 					icon_cache_key += "[M][markings[M]["color"]]"
 
 			if(body_hair && islist(h_col) && h_col.len >= 3)
-				var/cache_key = "[body_hair]-[icon_name]-[h_col[1]][h_col[2]][h_col[3]]"
-				if(!limb_icon_cache[cache_key])
-					var/icon/I = icon(species.get_icobase(owner), "[icon_name]_[body_hair]")
-					I.Blend(rgb(h_col[1],h_col[2],h_col[3]), ICON_MULTIPLY) //VOREStation edit
-					limb_icon_cache[cache_key] = I
-				mob_icon.Blend(limb_icon_cache[cache_key], ICON_OVERLAY)
+				var/icon/I = icon(species.get_icobase(owner), "[icon_name]_[body_hair]")
+				I.Blend(rgb(h_col[1],h_col[2],h_col[3]), ICON_MULTIPLY) //VOREStation edit
+				mob_icon.Blend(I, ICON_OVERLAY)
+				icon_cache_key += "[body_hair][rgb(h_col[1],h_col[2],h_col[3])]"
 
 	if(model)
 		icon_cache_key += "_model_[model]"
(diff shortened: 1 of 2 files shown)


Arokha:
This PR will be pointless for a bit given the one I opened right before it

KasparoVy:
My bad, didn't see-- sorry. Want me to close for now?

KasparoVy:
Sounds good to me.

code/modules/organs/organ_icon.dm
```diff
@@ -159,12 +159,10 @@ var/global/list/limb_icon_cache = list()
...
-				var/cache_key = "[body_hair]-[icon_name]-[h_col[1]][h_col[2]][h_col[3]]"
-				if(!limb_icon_cache[cache_key])
-					var/icon/I = icon(species.get_icobase(owner), "[icon_name]_[body_hair]")
-					I.Blend(rgb(h_col[1],h_col[2],h_col[3]), ICON_MULTIPLY) //VOREStation edit
-					limb_icon_cache[cache_key] = I
-				mob_icon.Blend(limb_icon_cache[cache_key], ICON_OVERLAY)
+				var/icon/I = icon(species.get_icobase(owner), "[icon_name]_[body_hair]")
+				I.Blend(rgb(h_col[1],h_col[2],h_col[3]), ICON_MULTIPLY) //VOREStation edit
+				mob_icon.Blend(I, ICON_OVERLAY)
+				icon_cache_key += "[body_hair][rgb(h_col[1],h_col[2],h_col[3])]"
```
Arokha: Um, so you're just entirely removing caching for these?

KasparoVy: Yeah, it no longer pulls directly from a snowflake cache in `limb_icon_cache` and just uses the same methods that limbs normally use (appendiglng to `icon_cache_key` and getting to `limb_icon_cache` from there.
This section is now treated the exact same as markings

KasparoVy: -snipped off a bunch of BS- Yeah, I'm taking it off because I'm wondering if this behaviour is related to the crashing issues, nothing more. The caching looks perfectly fine to me.

I could leave it off this PR to see if just fixing the business in `update_icons.dm` does the trick.

Arokha: Is this the only thing that used limb cache...?

KasparoVy: The only other place `limb_icon_cache` is used is for the base limb icons in health-doll icon generation over at https://github.com/VOREStation/VOREStation/blob/15d55d5d06afc518e1167f639acfcc662468a2bf/code/modules/organs/organ_icon.dm#L240

KasparoVy:
Haven't crashed or been able to crash people with a bugged non-Teshari character after 512.1417/18.
Will cut the caching changes tonight as they no longer seem necessary and probably wouldn've fixed anything, anyway.

PR #3324, opened 2018-03-22
Promethean tweaks (hunger, regen time)

Arokha:
* Half their hunger rate (from 4x human to 2x human)
* Decrease regenerate time by 10 seconds (from 30 to 20)

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/prometheans_vr.dm
+++ b/code/modules/mob/living/carbon/human/species/station/prometheans_vr.dm
@@ -1,15 +1,26 @@
 /datum/species/shapeshifter/promethean
-	min_age = 18
+	min_age = 18 //Required for server rules
 	max_age = 80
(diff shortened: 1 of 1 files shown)


Benemuel:
They REALLY need some increased limb HP too, because even with their brute resist a single punch is often enough to remove a limb. 10HP might be decent, at least to see if its enough to deal with the fact that at the current moment they are incapable of going on expeditions if their is going to be danger. Especially if the amount of nutrition that is needed stays the same.

Arokha:
Not really as a part of this PR, tho.

HzDonut:
The reason they lose limbs and such so easily is because of the fact that it's nigh impossible to actually kill them via brute damage and the fact they can just regenerate themselves. If you don't believe that they're massive tanks, just ask around about that VR test event where everyone was a promethean and had rocket launchers. Death was practically impossible and most of the time you were just sat there in critical after taking multiple rockets directly to the body. T'was quite the experience.

PR #3325, opened 2018-03-22
Removes excavation drill since it's been ported upstream

Cameron653:
Excavation drill was ported upsteam & in a much better fashion codewise.


Arokha:
I'll do this before the next sync, but not yet, so that it doesn't disappear if we need to update between now and then.

Cameron653:
Alrighty.

Cameron653:
Seeing as the sync has gone through, should this be merged?

PR #3326, opened 2018-03-22
Release to test Teshari crash fix

Arokha:


KasparoVy:
Couldn't crash anyone during the test period today after this merge.

I didn't notice any crashing yesterday on a bone-stock Teshari (still crashed people with that one character) without a custom tail, but I don't know if that's legit.

I figure it's down to the body_hair, but d'you think we could uncomment the tail hair to see if that's the ticket?

KasparoVy:
Still getting/seeing others experience a bunch of crashes, just not from Teshari

PR #3329, opened 2018-03-23
Fixes default synth tails necessitating potentially harmful mystery delays in mannequin generation.

Verkister:
-Now get_tail_image() can check the owner's syntheticness without having to wait for dress_preview_mob() to crapshoot fingers crossed for some arbitrary irrelevant life tick function to unreliably call isSynthetic() to retrieve and apply the vars required by get_tail_image before the mannequin deletes itself.
-This also fixes default synth tails not showing up at all on previews with naked human bases.

### DIFF
--- a/code/modules/vore/appearance/update_icons_vr.dm
+++ b/code/modules/vore/appearance/update_icons_vr.dm
@@ -17,11 +17,11 @@ var/global/list/wing_icon_cache = list()
 
 /mob/living/carbon/human/proc/get_tail_image()
 	//If you are FBP with tail style and didn't set a custom one
-	if(isSynthetic())
-		if(synthetic.includes_tail && !tail_style)
-			var/icon/tail_s = new/icon("icon" = synthetic.icon, "icon_state" = "tail")
-			tail_s.Blend(rgb(src.r_skin, src.g_skin, src.b_skin), species.color_mult ? ICON_MULTIPLY : ICON_ADD)
-			return image(tail_s)
+	var/datum/robolimb/model = isSynthetic()
+	if(istype(model) && model.includes_tail && !tail_style)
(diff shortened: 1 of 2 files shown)


code/modules/vore/appearance/update_icons_vr.dm
```diff
@@ -17,10 +17,11 @@ var/global/list/wing_icon_cache = list()
 
 /mob/living/carbon/human/proc/get_tail_image()
 	//If you are FBP with tail style and didn't set a custom one
-	if(synthetic && synthetic.includes_tail && !tail_style)
-		var/icon/tail_s = new/icon("icon" = synthetic.icon, "icon_state" = "tail")
-		tail_s.Blend(rgb(src.r_skin, src.g_skin, src.b_skin), species.color_mult ? ICON_MULTIPLY : ICON_ADD)
-		return image(tail_s)
+	if(isSynthetic())
+		if(synthetic.includes_tail && !tail_style)
```
Arokha: Your way will work fine, but there's a very slightly better way to do this, to make it a little more future proof. My concern is: "Well, if we change/don't set/alter how isSynthetic() works so that it doesn't set the synthetic var anymore, this will compile fine, but runtime and synths won't get tails" which isn't the case right now, but who knows.

So, what you could do, is... because isSynthetic() actually returns the manufacturer, you could just do:
```
var/datum/robolimb/model = isSynthetic()
if(istype(model) && model.includes_tail && !tail_style)
```
Which (a) cannot runtime, because at worst, isSynthetic will return nothing, or something crazy, and the istype will gracefully return false, and (b) allows us to not worry about what synthetic is set to on the mob at all.

Verkister: Alrighty

Arokha:
So, the tails do show up in the preview with this?

Verkister:
yeah

PR #3331, opened 2018-03-23
Possible fix for drop-nomming

Heroman3003:


code/modules/multiz/movement_vr.dm
```diff
@@ -52,7 +53,40 @@
...
+				var/tdamage
+				for(var/i = 1 to 10)
+					tdamage = rand(0, 10)/2
+					pred.adjustBruteLoss(tdamage)
+					prey.adjustBruteLoss(tdamage)
+				pred.updatehealth()
+				prey.updatehealth()
+*/
+/mob/proc/dropped_onto(var/atom/hit_atom)
+	if(isliving(hit_atom) && isliving(src)) //THIS WEAKENS THE PERSON FALLING & NOMS THE PERSON FALLEN ONTO. SRC is person fallen onto.  hit_atom is the person falling. Confusing.
```
Arokha: Ok, if this is defined on the thing being fallen ONTO, and ONLY does anything when they are living, this should be moved to /mob/living/dropped_onto and the isliving(src) check removed, along with the isliving(hit_atom) changing to be the opposite check with an early return of ..() instead of indenting the entire proc inside an if statement

PR #3332, opened 2018-03-23
Makes putting down plating not place airless plating

### LINKED ISSUE #2460: putting platting on latice causes pressure damage
putting plateing down in the aitruim make take pressue damage

to have air in the room when I put floor down.

#### What actually happened
that

#### Steps to reproduce
- (Step 1) put lattice down
- (Step 2)put plateing down
#### Code Revision
 a5844f786ef2a09948e6718a7d418b05d95f9324 
putting lattice down may make you be in the middle of floor two, with unsported lattice just floating there

Cameron653:
Fixes #2460
https://github.com/VOREStation/VOREStation/issues/2460

### DIFF
--- a/code/modules/multiz/turf.dm
+++ b/code/modules/multiz/turf.dm
@@ -129,7 +129,7 @@
 			qdel(L)
 			playsound(src, 'sound/weapons/Genhit.ogg', 50, 1)
 			S.use(1)
-			ChangeTurf(/turf/simulated/floor/airless)
+			ChangeTurf(/turf/simulated/floor) //VOREStation Edit
 			return
 		else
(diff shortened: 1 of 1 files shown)


SilencedMP5A5:
finally, now samba will never again baffle medical by coming in with a burst lung without having left the station.

SpadesNeil:
Uh, keep in mind, this will make oxygen appear when you try placing tiles. I don't think this is a good change.

I think this should be coded in a more elaborate manner wherein the air contents of the original tile are copied.

Arokha:
Agree with what @SpadesNeil said, take a look at shuttle code in unsorted.dm, namely like lines 815 and down.

Cameron653:
Hmm, will look into this and see about bringing it over to Polaris.

PR #3338, opened 2018-03-24
Removes an inconsistency in custom species code

Verkister:
This line had no real use at all. It may even have contributed to various issues.
All basic humanoid species have had their appearances constructed onto an _existing_ blank 32x32 icon_state named and called for as "nothing" inside the effects.dmi.
There are no untitled blanks either inside the effects.dmi file, so the funky deviant custom species icon_states would never have had any proper fallback.

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/blank_vr.dm
+++ b/code/modules/mob/living/carbon/human/species/station/blank_vr.dm
@@ -87,7 +87,6 @@
 
 	//Set up a mob
 	H.species = new_copy
-	H.icon_state = lowertext(new_copy.get_bodytype())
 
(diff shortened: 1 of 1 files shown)


Verkister:
disregard for now, it's a bit deeper than that and still got some digging to do

Verkister:
Deeper instance goes to polaris side. Testing effects to decide on further actions.

Verkister:
https://github.com/PolarisSS13/Polaris/pull/5100

Arokha:
Probably not related to any issues, but given icon_state is an item on appearances and one less appearance-change while spawning is good, seems fine to me.

PR #3340, opened 2018-03-24
Balance pass on custom species traits

Screemonster:
Custom species have been facing some severe powercreep. The original design intent for these was to allow some extra flexibility for people who wanted custom things, but that stock species would always have a slight edge in terms of stats, but trait costs haven't really reflected this, as well as some being outright superior to anything stock species get, _especially_ when you stack them.

Generally speaking, some of the crazier positive traits have had their costs increased or been outright removed, while negative traits have had their point rebates reduced to discourage stacking.

The changes are as follows:

POSITIVE TRAITS:
- Haste costs 4 points, up from 3. Haste plus removed (wanna go sonic, be a tesh).
- High Endurance increased from 2 to 4. Very High Endurance removed.
- 15% Brute/Burn resistance trait costs increased from 1 to 2.
- 25% resistances increased from 2 to 3 points, and cannot be stacked with the other resistance.
- 50% resistance traits removed.

NEGATIVE TRAITS:
- Slow reduced from 3 to 2. Major slow reduced from 5 to 3.
- Conductive reduced from 2 to 1, Major conductive reduced from 3 to 2.
- Colorblindness traits reduced to 1.

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/traits_vr/negative.dm
+++ b/code/modules/mob/living/carbon/human/species/station/traits_vr/negative.dm
@@ -1,13 +1,13 @@
 /datum/trait/speed_slow
 	name = "Slowdown"
 	desc = "Allows you to move slower on average than baseline."
-	cost = -3
+	cost = -2
 	var_changes = list("slowdown" = 0.5)
 
 /datum/trait/speed_slow_plus
 	name = "Major Slowdown"
 	desc = "Allows you to move MUCH slower on average than baseline."
-	cost = -5
+	cost = -3
 	var_changes = list("slowdown" = 1.0)
 
 /datum/trait/weakling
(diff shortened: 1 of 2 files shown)


Screemonster:
For comparison's sake: it was possible, for 9 points, to make a character that was 50% resistant to both brute and burn, and had 150HP - meaning they'd be able to tank 300 damage before falling over. And that's before gear.
That's twice as much as the best, tankiest, stock species.
9 points wasn't all that hard to get. Especially since colourblind gave you 3 and had no downside in terms of away mission effectiveness whatsoever, and since few mobs have electric attacks, Major Conductive was another 3 freebies for explorers.
As for why that's important, it makes it impossible to balance away missions, events and PoIs when some of the people that show up to them are playing stock species or non-crazy custom ones, and others are there like NANOMACHINES, SON.
Basically if we balance for the former, the latter are there facetanking deathclaws and (in some cases) actually calling other players bad at the game for dying, when they only died due to a strict numerical disadvantage. If we balance for the latter group, the missions become unbelievably punishing for people who aren't minmaxing.

Very-Soft:
YES!! I am in 100% support of all of this.

Ketrai:
It was even worse when it came to synths. Specifically people building a synth character to be near unkillable. tbh, with the regeneration of fleshbags having gotten nerfed so hard, there has barely been a reason not to pick a synth over a fleshbag when looking to min-max your way through events and missions. Admittedly I have used haste + myself, because It thought it would fit a small, 4'7, hyperactive fennec-fox like. And that it made getting away from danger instead of fighting it easy, as well as making getting around the station a cakewalk. (Seriously tho, we have too many empty halls :p) I hope these changes will put a stop to that anyways. Maybe take a look at those mechanites too? if they still exist? You know the stuff you can get from nif machines that essentially gives you regen.

PR #3351, opened 2018-03-25
alraune species

Screemonster:
Adds a new plant person race. They have plant species tags so you can shoot them with a somatoray and make them fat.

Organs are limited to eyes, brain, kidneys and liver. All planty equivalents. Brains are in their torso.

Species features:

- Slow metabolism. Slow runspeed, but not slowed down much further by gear. Very slow bleed rate.
- Broad comfort zone for temperature tolerance, but weak to extremes in either direction.
- No lungs! They breathe through their skin. That means that if you put internals on and you're not wearing a sealed suit, unless you're standing in a _very_ low pressure environment (and probably dying as a result) your environment is gonna be the one you breathe from. Yes, this _does_ mean you die from poisoning if you go outside wearing the winter coats instead of a sealed suit, though the temperature is tolerable. 
- Gas masks offer no protection against chemical smoke. Covering your face doesn't help when it's your entire body surface that does the breathing.
- If they're in a CO2-enriched environment, things get better for them. While they might gasp a bit if there's no oxygen, they can actually survive breathing pure CO2. If they have CO2 _and_ light, they heal brute and burn damage, but they need more light than they can get while wearing a spacesuit to benefit from this.
- If they're really hungry, they gain nutrition from light, albeit at a slower rate than diona. The nutrition cap they can be taken to is raised if they're breathing CO2. Again, take your suit off.
- Yes I know Virgo 2 outdoors is mostly CO2. Don't try it. You'll burst into flames and die.
- Species is whitelisted _for now_ but honestly I'm planning for it not to be once it's done. Don't bother applying 'cause once they're finished they'll be open for everyone.
- Gave them bloodsuck and succubus drain to you can have your fun times of wrapping someone in your thorny vines and sapping out their vital fluids I guess.


ParadoxSpace:
wtb succubus drain

Little-119:
alarune me are familiar faces

Shouldn't the temperature tolerances be the opposite? Weak to cold, strong to heat? Plants tend to thrive in warm and humid climates while only hardy plants can survive in colder climates.

Screemonster:
That's a good point. That said, if you stuck me outside naked this last winter, I'm sure I'd last less time than the things in my garden (especially that fucking pyracantha the previous owners planted).
An interesting way to do it would be to give them a rather tolerant cold/heat level 1, but very weak level 2/3 points so while they'have a broader range of temperatures they're comfortable in, they'd suffer *fast* if taken outside that zone.

Vorrarkul:
Would they be resleeved like other species? I had the notion of "what if when they died, inside their body is a seed which needs to be planted in hydroponics to regrow them instead?"
Along that line, make the botanist their version of a doctor. And obviously weak to plant-b-gone if doable.

Screemonster:
It'd be _interesting_ to give them a brain that could be regrown into a new body with third-party intervention. iirc that's how prometheans were originally supposed to work, dropping a core that could be reconstituted into a body.

Sadly, the plant-b-gone thing requires the IS_DIONA reagent tag which also fucks up a lot of other chemical interactions. For a start, it makes toxin not poison them. With that tag they can literally walk outside and not take toxloss from breathing phoron, because the way breath code works, breathing phoron adds toxin to your bloodstream instead of directly causing toxloss. Don't ask me why.

Vorrarkul:
Would they have Gazer's (I think it was gazer at least) alraune's flower bottom that can close over them?

Screemonster:
Temp tolerances have been adjusted. They've got broader tolerance of heat and cold than humans, but the extreme-damage-zone once they get out of that comfy area is much narrower.

pontifexminimus2:
https://hastebin.com/raw/fesejabixa

Current lore for them.

Screemonster:
Now they've got their own breath code instead of human handle_breath(), we don't _have_ to make breathing phoron give them the toxin reagent and can deal toxloss directly if we want, which is the reason I didn't give them the IS_DIONA reagent tag.

Having a quick peek through the code, here's what that tag would do:

- everything in chemistry-reagents-medicine.dm that mentions the IS_DIONA tag will have no effect. Too many to list. Includes basically all the basic medicines.
- basically all toxins are made to have no effect either. Again, too many to list.
- Food, drink, booze? Again, no effect for a huge chunk of them.
- plantbgone is horribly poisonous (which would be appropriate, but 50 tox per unit appropriate?)
- They can't overdose. On anything.
- Honestly diona might as well just be synths for all the effect most reagents have on them besides hyperzine.

tl;dr the only desirable and appropriate reagent trait that they get from all of this is plantbgone being poisonous. Plantbgone is already poisonous to humans, being a child of toxin, so on balance it's best not to give them the tag and just have them process reagents like regular crew.

Natje0:
The plantbulb is a fluff part for my character Pumila, not Gazers :P and it would need a colorable version if there'd be a publicly accessible variant.

Little-119:
What do the Alarune (Alarunes? Alari?) look like, anyway?

Vorrarkul:
@Natje0 Sorry, I forgot who had the character.

Screemonster:
Okay, these should be... viable? At least they won't break everything and they're playable. Organs exist for them now.

code/modules/mob/living/carbon/human/species/station/alraune.dm
```diff
@@ -0,0 +1,336 @@
...
+
+
+/datum/species/alraune/handle_environment_special(var/mob/living/carbon/human/H)
+	if(H.inStasisNow()) // if they're in stasis, they won't need this stuff.
+		return
+
+	//setting these here 'cause ugh the defines for life are in the wrong place to compile properly
+	//set them back to HUMAN_MAX_OXYLOSS if we move the life defines to the defines folder at any point
+	var/ALRAUNE_MAX_OXYLOSS = 1 //Defines how much oxyloss humans can get per tick. A tile with no air at all (such as space) applies this value, otherwise it's a percentage of it.
+	var/ALRAUNE_CRIT_MAX_OXYLOSS = ( 2.0 / 6) //The amount of damage you'll get when in critical condition. We want this to be a 5 minute deal = 300s. There are 50HP to get through, so (1/6)*last_tick_duration per second. Breaths however only happen every 4 ticks. last_tick_duration = ~2.0 on average
```
Arokha: Why are these in allcaps and not defines?

code/modules/mob/living/carbon/human/species/station/alraune.dm
```diff
@@ -0,0 +1,336 @@
...
+		O_EYES =     /obj/item/organ/internal/eyes/alraune,
+		)
+
+/datum/species/alraune/can_breathe_water()
+	return TRUE //eh, why not? Aquatic plants are a thing.
+
+
+/datum/species/alraune/handle_environment_special(var/mob/living/carbon/human/H)
+	if(H.inStasisNow()) // if they're in stasis, they won't need this stuff.
+		return
```
Arokha: This is like 200 lines of copypaste? There's no better way?

code/modules/mob/living/carbon/human/species/station/alraune.dm
Screemonster: For some reason it refused to compile when I pointed them at the HUMAN_MAX_OXYLOSS define. No idea _why_ it can't find them when it's further down the tree from where they're defined.

code/modules/mob/living/carbon/human/species/station/alraune.dm
Screemonster: I'd love to just use handle_breath() from line 132 onwards, but there's a whole lot of stuff in handle_breath() that's really bad for them, like making them gasp when there's oxygen present because it's their exhale type, damage from hot air specifically targeting their head and giving messages about icicles in the lungs, and so on. Given that they're plants and don't have lungs I'd honestly be fine with massively simplifying their breath code here though.

SpadesNeil:
Do these things need a special hardsuit or are they just gonna wear human hardsuits?

Screemonster:
I'm gonna assume just human hardsuits. They use the human spritesheet. If we put the "pick a spritesheet" thing like customspecies, we can use the same system they use where you need to use the suits that match your spritesheet.

code/modules/mob/living/carbon/human/species/station/alraune.dm
Arokha: They may be #undef'd at the bottom of their file or somewhere earlier.

Screemonster: Aha, I might have a look at that when I get out of work.

SpadesNeil:
I think we are otherwise all set. Remove "requires discussion" if you think we're set. Last requirement is to add a wiki page and an entry to https://wiki.vore-station.net/Backstory if you have not already. Also maybe edit https://wiki.vore-station.net/Timeline to introduce the species.

Screemonster:
The main thing that might be worth doing is seeing if there's a shorter/simpler way to handle their breath code than just a yard of copypasta. I'm sure there's more stuff that _could_ be done with them but it's either beyond me or will need, yanno, actual testing beyond the simple functional stuff I've done.

Still whitelisted for now so it's in a similar position to proteans. Dragor, Natje and Tabl came up with lore stuff for them.

PR #3352, opened 2018-03-26
Adds Keypad Auth to CE Office

Mewchild:
The CE office was somehow missing this, because they're less of a command staffer than anyone else 👎 
But ye, fixes it


Jakksergal:
It's always the wall-mounted objects I forget.

PR #3358, opened 2018-03-27
Fixes feathered tesh tail into proper dual-color format.

Verkister:
-Fixes the feathered tesh tail in tail selection showing up as the feather overlay only without the base tail showing up at all.
-Feathered tesh tail now uses the proper dual color system now that it no longer comes with the default tail.

-Had a little whoops earlier forgetting to check other open PRs, #3323 this one already having the other stuff I tried to fix so I only left the tail part here.


Verkister:
Tail markings was never a thing I think. Just had the arm/leg feathers as markings

Issue #3353, opened 2018-03-26
Some messed up camera console tags.


Verkister:
#### Brief description of the issue
There's a bunch of multiselecting dupe names that also tend to have no relation to the duped name. Missing tags?
-Medbay recovery room, shows up as dupe of MED - Surgery Hallway
-Medbay staff room west, shows up as dupe of MED - Surgery Hallway
-Medbay staff room southwest, shows up as dupe of MED - Surgery Hallway
-Chemistry lab, shows up as dupe of MED - Virology Quarantine Airlock
-Medbay staff room east, shows up as dupe of MED - Virology Quarantine Airlock
-Virology west/east cells, show up as dupes of Virology #420
-Cargo office south, shows up as dupe of CRG - Mining Airlock
-Cargo bay east, shows up as dupe of CRG - Mining Airlock
-Cargo bay warehouse, shows up as dupe of CRG - Mining Airlock
-Command meeting room east, shows up as dupe of Gateway Access
-Bridge hallway north, shows up as dupe of Gateway Access
-Bridge northeast, shows up as dupe of Gateway Access
-HoP office, shows up as dupe of Gateway Access
-Green armory, shows up as dupe of SEC- Head of Security's Office
-Warden side equipment storage, shows up as dupe of SEC- Head of Security's Office
-Red armory east, shows up as dupe of SEC- Security EVA
-Red armory west, shows up as dupe of SEC- Vault
-SEC - Vault Exterior cameras all show up in the same tag, no north/west/south distinction
-Brig visitor area, shows up as dupe of SEC- Warden's Office
-Brig center, shows up as dupe of SEC- Warden's Office
-Bridge teleport access, shows up as COM - Bridge Starboard (no dupe, just tag mismatch)
-Morgue north, shows up as MED - Examination Room (no dupe, just tag mismatch)
-Research upper hallway north and west show up as Research and Development cameras (no dupes, just tag mismatch)
-Anomaly lab containment cameras show up on Research Outpost network instead of station side Research like the rest.

#### What you expected to happen
Properly tagged cameras.

#### What actually happened
These things, each dupe button selecting all of the ones under the name at once, requiring to jump via another different cam to switch between the collectively toggled dupes.

#### Steps to reproduce
- (Step 1) Access surveillance camera monitor.
- (Step 2) Press buttons.

#### Code Revision
- 0b4a7e3a7ea4baed068ae7fcd1ce68a4632a2fc8
 
#### Anything else you may wish to add:
[...]

PR #3363, opened 2018-03-28
Fixes incorrect camera naming/networks.

Mewchild:
Fixes issue #3353

### DIFF
--- a/maps/tether/tether-01-surface1.dmm
+++ b/maps/tether/tether-01-surface1.dmm
@@ -13949,7 +13949,8 @@
 /area/rnd/external)
 "aHg" = (
 /obj/machinery/camera/network/research_outpost{
-	dir = 4
+	dir = 4;
+	network = list("Research")
 	},
 /obj/structure/cable/green{
 	d1 = 1;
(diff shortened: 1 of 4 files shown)


maps/tether/tether-01-surface1.dmm
```diff
@@ -13949,7 +13949,8 @@
 /area/rnd/external)
 "aHg" = (
 /obj/machinery/camera/network/research_outpost{
-	dir = 4
+	dir = 4;
+	network = list("Research")
```
Leshana: Correct fix here is to change the *type* to `/obj/machinery/camera/network/research` instead of overriding the network.

Mewchild: I was actually unaware of that. Thank you.

Leshana: No problem, except you forgot to actually remove the network var on this one.  It was removed on the one below.

Leshana: No problem, except you forgot to actually remove the network var on this one. It was removed on the one below.

PR #3369, opened 2018-03-30
Fixes infinite loop in FBP naming w/o client

Arokha:
Kelshark is a lazy bum


Repede:
It was either I spend 20 minutes making the PR or you did it in like 10 seconds. :^)

PR #3380, opened 2018-04-02
Nerfs solargrub drain speed.

Ketrai:
Due to the lack of saturation in the moderate event container, solargrubs show up way too frequently, and changing the frequency doesn't help. The container has to pick. This change will just bump the drain speed down so severely, that if you kill one or two grubs, the station should barely feel the impact anymore. As opposed to having to kill 4/5 grubs on average.

Gradual nerfs seem logical, as grubs get more and mroe of a day-to-day experience, and most people know about them now. Time to stop the witch hunting for those annoying grubs :p

### DIFF
--- a/code/modules/mob/living/simple_animal/vore/solargrub.dm
+++ b/code/modules/mob/living/simple_animal/vore/solargrub.dm
@@ -77,8 +77,8 @@ List of things solar grubs should be able to do:
 				sparks.start()
 			anchored = 1
 			PN = attached.powernet
-			PN.draw_power(150000)
-			var/apc_drain_rate = 750 //5000 seems a bit high, knocking it down to 4000
+			PN.draw_power(100000) // previous value 150000
+			var/apc_drain_rate = 750 //Going to see if grubs are better as a minimal bother. previous value : 4000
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/simple_animal/vore/solargrub.dm
```diff
@@ -78,7 +78,7 @@ List of things solar grubs should be able to do:
 			anchored = 1
 			PN = attached.powernet
 			PN.draw_power(150000)
-			var/apc_drain_rate = 4000 //5000 seems a bit high, knocking it down to 4000
+			var/apc_drain_rate = 750 //5000 seems a bit high, knocking it down to 4000
```
Zeldazackman: -"5000 seems a bit high, knocking it down to 4000"
Should update this line.

Arokha: True. Also not lowering the 150000? Want that rate to stay the same, or?

Ketrai: That draws power directly from the powernet, doesn't affect it that much. Apc drain rate is much worse? I'll knock it down a bit anyways.

PR #3385, opened 2018-04-04
Overrides the base Energy Gun sprite with a good looking one

Michiyamenotehifunana:
:cl: Toriate
imageadd: Added new Energy Gun sprite.
code: Created nuclear_vr.dm and eguns_vr.dmi for overriding eguns
/:cl:

dreamseeker_2018-04-05_00-12-39

This PR will replace the regular egun's sprite.

The original is horrendous.
dreammaker_2018-04-04_23-29-25

The new one has received positive response from the discord.
dreammaker_2018-04-06_17-32-05

Due to how all eguns are child objects of the base egun, the AEG and burst fire egun has to have their icon vars pointing back to the default one. The only other solution is to mirror all the default egun sprites in eguns_vr.dmi.


Arokha:
Pathing breaks convention. Should be same names as files they're overriding content from, so a _vr.dm and _vr.dmi file in each folder. Please fix. See https://github.com/VOREStation/VOREStation/blob/master/.github/CONTRIBUTING.md

code/modules/projectiles/guns/energy/nuclear_vr.dm
```diff
@@ -0,0 +1,8 @@
+/*
+//	Virgo's very own file for overriding Polaris' eguns! Should include: Regular egun, Advanced egun, burst egun, mounted egun
+//
+//	NOTICE: All the egun sprites have to be mirrored in /icons/obj/eguns_vr.dmi because someone made them all child objects of the base thing
```
Arokha: Paths are good now, though not super-happy about this part. Trying to think of a way to circumvent this.

Michiyamenotehifunana: I could just add an entry to each individual child object to point their icon back to the default one.

Little-119:
In my personal opinion, I think they might look better with black/greyscale bodies instead of... brown? Whatever you would call that color.

Michiyamenotehifunana:
It's only the highlight that's in a contrasting, hue-shifted color. The rest of the gun is a somewhat bright gunmetal grey. I might try tweaking the colors to make it less conspicuous.

Michiyamenotehifunana:
Improved contrast:
dreammaker_2018-04-06_17-32-05

Original:
dreammaker_2018-04-04_23-29-56

PR #3387, opened 2018-04-05
Fix Teshari Tail-hair Rendering

KasparoVy:
This PR fixes Teshari tail hair not rendering at all.
Polaris didn't have this issue because they fixed the name of the icon state in `icons/species/effects.dmi` some time ago, though due to restrictions on altering stock `.dmi` files a Virgo edit to the icon state string is necessary.

_Teshari Tail Hair Busted_
tesharitailhaircolourbusted

_Teshari Tail Hair Fixed_
tesharitailhaircolourfixed

### DIFF
--- a/code/modules/mob/living/carbon/human/update_icons.dm
+++ b/code/modules/mob/living/carbon/human/update_icons.dm
@@ -871,8 +871,8 @@ var/global/list/damage_icon_parts = list() //see UpdateDamageIcon()
 		// The following will not work with animated tails.
 		var/use_species_tail = species.get_tail_hair(src)
 		if(use_species_tail)
-			var/icon/hair_icon = icon('icons/effects/species.dmi', "[species.get_tail(src)]_[use_species_tail]")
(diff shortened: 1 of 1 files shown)


Arokha:
What would the port be, just the _s?

KasparoVy:
Yeah, that's all.

KasparoVy:
Polaris is missing the _s. I can recall checking the icon file on their end a few days ago and seeing that the icon states were the same as ours. Wouldn't that make this need a port job rather than a Virgo edit comment?

KasparoVy:
The .dmi is now the exact same as the one currently live on the Polaris codebase's master branch.

Arokha:
Regardless, this commit will modify it, and git cannot compare binaries. It will conflict. You cannot edit stock Polaris DMI files.

KasparoVy:
Aw, damn. Well that's a fun situation.

Virgo edit it'll be then I guess. Thanks for clearing it up

KasparoVy:
Reverted back to previous state where only the icon state string was altered.

PR #3389, opened 2018-04-05
Fixes resting and vore icon issues with mobs and silicons

### LINKED ISSUE #3388: Mob vore icons not updating anymore
Seems like nowadays whenever mobs eat folk, their icons to show their bellies don't work. Sometimes they work for a BIT, then reset back to the default mob sprite. Real shame considering it makes it harder to tell when things have people inside of them with the quieter belly noises, AND who doesn't love a good mob belly, let's be honest.

Mob eats person, gets belly sprite.

#### What actually happened
Mob eats person, nothing happens OR mob eats person, sprite updates, then resets back to default.

#### Steps to reproduce
- Just have any vore mob eat someone.

#### Code Revision
-  release - 
0b4a7e3a7ea4baed068ae7fcd1ce68a4632a2fc8

### LINKED ISSUE #3394: Hound Size Change bug
When a V2 medihound has a pose change. Either by its panel being opened or by resting, it will reset the sprite to 100%, but the game will still think your character is at 200%. As such in order to fix the sprite One must first shoot the medihound back to 100% before shooting it again with 200%

To rest at 200% or have my panel opened without shrinking.

#### What actually happened
I shrunk to the sprite size of 100% without the code recognizing me as 100% for no reason.

#### Steps to reproduce
- Set a medihound's size to 200%
- Have the hound rest or open its panel

#### Code Revision
- 0b4a7e3a7ea4baed068ae7fcd1ce68a4632a2fc8
 
- I am unsure if it true with other hound sprites, but it probably is.

### LINKED ISSUE #3397: Voremobs gib as failsneks
Mobs using vore.dmi or vore64x64.dmi have no gibbing icons, and those both files have the failsnek as blank fallback, causing the mobs to explode as failsneks.

No graffitied sneks flying out with other guts.

#### What actually happened
Gibbed mobs briefly explode as graffitied failsneks.

#### Steps to reproduce
- (Step 1) Gib a voremob using those files, in this case just butcher a dead grub or something.
- (Step 2) Witness the snek come and go.

#### Code Revision
- 0b4a7e3a7ea4baed068ae7fcd1ce68a4632a2fc8
 
Verkister:
-Fixed simpleanimal vore icon updates being out of date with parent update_icon()
-Made the forementioned vore icon thing a proc to stick into the new mutable_appearance based system.
-Added SA_ICON_REST to mobs with belly enabled resting icons.
-Made mob icon fullness rounding count for total prey in case of multiple guts.
-Fixed resting verb breaking the scaling with mobs/borgs/etc.
-Added vorestation resize compatibility to nonhuman update_transform() (the reason behind the resting issues)
-Added missing generic gib icons to voremob sprite files to prevent them gibbing as failsneks.

Fixes #3388 
Fixes #3394 
Fixes  #3397


code/modules/mob/living/simple_animal/simple_animal.dm
```diff
@@ -314,6 +314,9 @@
 	else
 		ma.icon_state = initial(icon_state)
 
+	if(update_vore_icon()) // VOREStation edit
+		ma.icon_state = update_vore_icon()
```
Arokha: Bad, calling the same proc twice in a row without need, waste of CPU. Store the value somewhere and check it like:
```
    var/vore_icon_state = update_vore_icon()
    if(vore_icon_state)
        ma.icon_state = vore_icon_state
```
Put //VOREStation Edit Start and end

code/modules/mob/living/simple_animal/simple_animal_vr.dm
```diff
@@ -44,21 +44,25 @@
...
-		new_fullness = round(new_fullness, 1) // Because intervals of 0.25 are going to make sprite artists cry.
+	new_fullness = round(new_fullness, 1) // Because intervals of 0.25 are going to make sprite artists cry.
 	vore_fullness = min(vore_capacity, new_fullness)
 
-/mob/living/simple_animal/update_icon()
-	. = ..() // Call sideways "parent" to decide state
+/mob/living/simple_animal/proc/update_vore_icon()
 	if(vore_active)
 		update_fullness()
 		if(!vore_fullness)
```
Arokha: I'd remove the if(vore_active) test here, make update_fullness() always just return early if(!vore_active), and then your if(!vore_fullness) here will always return and prevent vore icons on those without vore_active (becauase it will forever be 0). Then you can un-indent this entire proc.

code/modules/mob/living/simple_animal/simple_animal_vr.dm
```diff
@@ -44,21 +44,25 @@
...
+			return "[icon_living]-[vore_fullness]"
+		else if(stat >= DEAD && (vore_icons & SA_ICON_DEAD))
+			return "[icon_dead]-[vore_fullness]"
+		else if(((stat == UNCONSCIOUS) || resting || incapacitated(INCAPACITATION_DISABLED) ) && icon_rest && (vore_icons & SA_ICON_REST))
+			return "[icon_rest]-[vore_fullness]"
+	return 0
+
+/mob/living/simple_animal/lay_down()
+	..()
+	update_icons()
```
Arokha: . = ..() for safety when doing sideways overrides, in case someone alters the procs to use return values at some point.

What's the purpose of this, though? Is it for zshadows? That's the only thing done by /mob/living/update_icons. The simple animal proc is update_icon(). If it's for zshadows, you should give it to Polaris since they also use multiz and would benefit, and should really put it in update_icon() with a comment explaining why.

Verkister: update_icon() doesn't preserve scaling on simpleanimals

Arokha: Why?

Verkister: The resize_vr.dm only has the thing for update_icons(). Seems the resting verb uses update_icon() and apparently does the same for borgs too. #3394

code/modules/mob/living/simple_animal/simple_animal.dm
```diff
@@ -314,6 +314,10 @@
 	else
 		ma.icon_state = initial(icon_state)
 
+	var/vore_icon_state = update_vore_icon() //VOREStation edit
```
Arokha: Like my comment said, do vorestation edit start/end comments around this block.

code/modules/vore/resizing/resize_vr.dm
```diff
@@ -45,6 +45,14 @@ var/const/RESIZE_A_SMALLTINY = (RESIZE_SMALL + RESIZE_TINY) / 2
 	M.Translate(0, 16*(size_multiplier-1))
 	src.transform = M
 
+/mob/living/update_transform()
+	. = ..()
+	ASSERT(!ishuman(src))
```
Arokha: This proc already exists in `living.dm` on 1092, but it's commented out. You should probably just uncomment it and add a single line for size_multiplier scaling there rather than duplicating it all here.

Verkister:
Tested working on my end.

PR #3390, opened 2018-04-05
Small Shock/Holo collar tweaks.

Verkister:
-Updates their old << dialogue.
-Makes shock collars taggable.

Was also gonna fix them showing up in forced lowercase on people but that turned out to be a separate polaris issue that I just opened a fix for up there.


Verkister:
You can set a tag text for names/instructions/winkwonks/exposed zapcode bravery/etc. without requiring a hand labeler sorta same way as with the holocollar.

SilverTalismen:
Aah okay.

PR #3405, opened 2018-04-07
Constructable Operating Table and sleeper/bodyscanner boards in RnD.

Heroman3003:
Operating table is now constructable. Requires regular machine frame, a board (RnD level 2 in material and biology) and a scanning module.

Also added sleeper, bodyscanner and their consoles' boards to RnD (level 2 in magnets and bio).

### DIFF
--- a/code/game/machinery/OpTable_vr.dm
+++ b/code/game/machinery/OpTable_vr.dm
@@ -1,0 +1,15 @@
+/obj/machinery/optable
+	circuit = /obj/item/weapon/circuitboard/optable
+
+/obj/machinery/optable/attackby(obj/item/weapon/W as obj, mob/living/carbon/user as mob)
+	..()
+	if(default_deconstruction_screwdriver(user, W))
+		return
+	else if(default_deconstruction_crowbar(user, W))
+		return
+
+/obj/machinery/optable/New()
+	..()
+	component_parts = list()
+	component_parts += new /obj/item/weapon/stock_parts/scanning_module(src)
+	RefreshParts()
--- a/code/game/machinery/OpTable.dm
+++ b/code/game/machinery/OpTable.dm
@@ -90,7 +90,7 @@
 	C.resting = 1
 	C.loc = src.loc
 	for(var/obj/O in src)
-		O.loc = src.loc
+		if(!(O in component_parts) && !(O == circuit)) O.loc = src.loc	//VOREStation Edit: constructable optables
 	add_fingerprint(user)
 	if(ishuman(C))
 		var/mob/living/carbon/human/H = C
--- a/code/game/objects/items/weapons/circuitboards/circuitboards_vr.dm
+++ b/code/game/objects/items/weapons/circuitboards/circuitboards_vr.dm
@@ -53,3 +53,12 @@
 	build_path = /obj/machinery/computer/timeclock
 	board_type = new /datum/frame/frame_types/timeclock_terminal
 	matter = list(DEFAULT_WALL_MATERIAL = 50, "glass" = 50)
+
+//Constructable optable voard
+/obj/item/weapon/circuitboard/optable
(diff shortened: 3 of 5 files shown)


Arokha:
I'm unsure if Pol would want this, but we should discuss anyway.

Heroman3003:
Yeah, the main reason I didnt send it to Polaris is because I don't want to mess with balance that I do not understand.

Cameron653:
I can't see the bodyscanner being too big of an issue. The patient can generally tell you what's broken (leg, arm, eyes, etc) with a few exceptions, like internal organs. Even then, you can use a health analyzer & see their damages and be able to tell in most cases if they have heavy damage to an organ.

Sleeper is debatable, but I don't see it being abused. The biggest two things it can do is take reagents out of people & produce infinite tricord.

Op table is more... Questionable, as it's kind of accepted that if you are doing medical anywhere but the medical bay, you're taking bigger risks and an op table would be extremely powerful.

With that being said, I'd recommend bumping the bio origin_tech of the op-table to 4, or perhaps even 5, the bodyscanner bio origin_tech to 3, and the sleeper origin_tech to 4.

Heroman3003:
Bodyscanner and sleeper boards were not changed, they already existed and even were acquirable (although there's just two of each on the map) and I left them as they are, while making them makeable in RnD with their current origin_tech as req_tech. Perhaps its req_tech you want bumped, and not origin?

Cameron653:
Ah, yes. My bad on that. I was reading origin_tech annd thought that was the req_tech I was reading.

Heroman3003:
Yeah, I would agree that tech cost should be higher. origin_tech is rather low, but the results of construction can be rather valuable.

KasparoVy:
@Cameron653 In that case I guess you could have it such that you can only construct an optable if the area it's in has a sleeper/scanner/other medical equipment endemic to a fully equipped medbay.

Arokha:
Sort of a wonky test, though, scanning every item in range to determine if it should be allowed. Also people will just powergame and build one machine, build optable, then disassemble the first machine. Or leave it. I'm sure people will build an entire medical base if able.

KasparoVy:
🤷‍♂️

code/game/machinery/OpTable_vr.dm
```diff
@@ -0,0 +1,15 @@
+/obj/machinery/optable
+	circuit = /obj/item/weapon/circuitboard/optable
```
Arokha: No construct-able optable unless we can think of a way to limit it from being used for exploration. Please remove it if you want this to be merged, otherwise it can stay as a WIP indefinitely until such a time as we have an idea about how to do that. Up to you.

Screemonster:
my initial gut reaction was "hooray, yet more stuff that explorers will refuse to leave until they have".

Very-Soft:
While I would love it if basically everything in the game tech-wise could be researched and built, I can't think of a good reason why there should be a op tables anywhere outside of medical and robotics. 

If there's an event big enough to put both of those places out of commission, then there are probably admins on to graciously supply new op tables where appropriate.

Arokha:
Plan?

Ketrai:
I wouldn't approve this. seems like only one instance where this would be used. and explorers get a sleeper to counter the exact weakness that no table is available. SAR already have too much power. Let's not give them any more.

code/game/machinery/OpTable_vr.dm
Ketrai: You know, you could just give the explorers a single circuit board. they still have to get all the rest from engineering. it'd be the only one of its kind.

Screemonster:
Yeah, let's not let SAR stretch their job description more than they already do.

KasparoVy:
Highground _again_? VETO THIS!

PR #3406, opened 2018-04-08
Adds SilencedMP5A5's fluff items

Cameron653:
https://forum.vore-station.net/viewtopic.php?f=27&t=1302


Cameron653:
Correction: Need to update the name to be NT-HASD instead of serdykov antoz.

Cameron653:
Fixed.

SilencedMP5A5:
err, the way the suit is currently means anyone can put it on
like, even nontaurs
is there a way to fix this?

KasparoVy:
What do you mean? It checks for ckey.
https://github.com/VOREStation/VOREStation/pull/3406/files#diff-746910cb10d59a2f75a8b280085ab23eR1777

Cameron653:
Hmm, I can add a check to make sure the user is a taur like the normal taur hardsuits.

Cameron653:
Done. Added a check to see if they're a synthlizard & that their ckey is yours.

code/modules/clothing/spacesuits/void/void_vr.dm
```diff
@@ -368,3 +368,21 @@
 			else
 				H << "<span class='warning'>You need to have a horse, wolf, or naga half to wear this.</span>"
 				return 0
+
+/obj/item/clothing/suit/space/void/security/hasd
```
Arokha: Why is this not in custom_clothes_vr.dm? It's ckey locked, so it definitely should be.

Cameron653: Ack, let me move that. Forgot about that.

code/modules/clothing/spacesuits/void/void_vr.dm
```diff
@@ -134,7 +134,7 @@
 			else
 				H << "<span class='warning'>You need to have a horse, wolf, or naga half to wear this.</span>"
 				return 0
-				
+
```
Cameron653: I don't know why these lines were touched. They don't do anything, though.

Arokha:
Fixy da file

Cameron653:
File fixed

PR #3407, opened 2018-04-08
Changes wedding rings to be a subtype of ring

Little-119:
Also moves them to their own .dm, since they're now a subtype of the thing in rings.dm. Also changed bethrothed to spouse, as the former is usually before marriage and not after.

Benefits of this include: Being able to wear gloves over wedding rings. Making them tiny instead of small. Making their gender neuter instead of plural, so it wont say "some wedding ring" when you examine someone wearing them. Probably more good things.

Issue reported by HzDonut.

I'm really tired, so forgive me for any mistakes, please.


SpadesNeil:
Looks fine to me!

PR #3408, opened 2018-04-08
Adds a machine for recoloring apparel ingame.

Verkister:
-Not mapped in currently.
-Still unsure whether to have it whitelisted for white/grey items only or to let people have their fun experimenting on nasty aesthetics.
-Also supports bags and belts. Yes, I've heard the pleas of the people not liking the default grey being the only public backpack option there is.

86e56f28ae70849b1b0180b8a08c3509


code/game/machinery/painter_vr.dm
```diff
@@ -0,0 +1,98 @@
...
+	else if(inoperable())
+		icon_state = "colormate_off"
+	else if(processing.len)
+		icon_state = "colormate_active"
+	else
+		icon_state = "colormate"
+
+/obj/machinery/gear_painter/Destroy()
+	for(var/atom/movable/O in processing)
+		O.loc = src.loc
```
Arokha: O.forceMove(drop_location())

code/game/machinery/painter_vr.dm
```diff
@@ -0,0 +1,98 @@
...
+	if(default_deconstruction_screwdriver(user, W))
+		return
+	if(default_deconstruction_crowbar(user, W))
+		return
+	if(default_unfasten_wrench(user, W, 40))
+		return
+
+	if(is_type_in_list(W, allowed_types) && !inoperable())
+		user.drop_item()
+		W.loc = src
```
Arokha: W.forceMove(src)

code/game/machinery/painter_vr.dm
```diff
@@ -0,0 +1,98 @@
...
+		return
+	if(default_deconstruction_screwdriver(user, W))
+		return
+	if(default_deconstruction_crowbar(user, W))
+		return
+	if(default_unfasten_wrench(user, W, 40))
+		return
+
+	if(is_type_in_list(W, allowed_types) && !inoperable())
+		user.drop_item()
```
Arokha: `user.drop_from_inventory(W)` or `user.unEquip(W)` might be better, don't think you should call this since this is just pretending the user is dropping the item. Will work but is sorta sketchy to emulate the user doing things like that.

code/game/machinery/painter_vr.dm
```diff
@@ -0,0 +1,98 @@
...
+		..()
+	update_icon()
+
+/obj/machinery/gear_painter/attack_hand(mob/user as mob)
+	if(!inoperable())
+		interact(user)
+	update_icon()
+
+/obj/machinery/gear_painter/interact(mob/user as mob)
+	if(stat & BROKEN)
```
Arokha: Why bother, given the attack_hand() check for inoperable()? Or, in the case that other things can interact with it (ai clicks like from borgs), then it should also check for lack of power probably.

code/game/machinery/painter_vr.dm
```diff
@@ -0,0 +1,98 @@
...
+	else
+		for(var/atom/movable/O in processing)
+			dat += "Item inserted: [O]<HR>"
+		dat += "<A href='?src=\ref[src];select=1'>Select new color.</A><BR>"
+		dat += "Color: <font color='[activecolor]'>&#9899;</font>"
+		dat += "<A href='?src=\ref[src];paint=1'>Apply new color.</A><BR><BR>"
+		dat += "<A href='?src=\ref[src];eject=1'>Eject item.</A><BR><BR>"
+
+	user << browse(dat, "window=colormate")
+	onclose(user, "colormate")
```
Arokha: Steal the fake nanoUI CSS from another fake nanoUI machine. Let's not add any more boring plain white UIs.

code/game/machinery/painter_vr.dm
```diff
@@ -0,0 +1,98 @@
...
+			activecolor = newcolor
+
+	if(href_list["paint"])
+		for(var/atom/movable/O in processing)
+			O.color = activecolor
+		playsound(src.loc, 'sound/effects/spray3.ogg', 50, 1)
+
+	if(href_list["eject"])
+		for(var/atom/movable/O in processing)
+			O.loc = src.loc
```
Arokha: O.forceMove(drop_location())

code/game/machinery/painter_vr.dm
```diff
@@ -0,0 +1,98 @@
...
+	var/dat = "<TITLE>Color Mate Control Panel</TITLE><BR>"
+	if(!processing.len)
+		dat += "No item inserted."
+	else
+		for(var/atom/movable/O in processing)
+			dat += "Item inserted: [O]<HR>"
+		dat += "<A href='?src=\ref[src];select=1'>Select new color.</A><BR>"
+		dat += "Color: <font color='[activecolor]'>&#9899;</font>"
+		dat += "<A href='?src=\ref[src];paint=1'>Apply new color.</A><BR><BR>"
+		dat += "<A href='?src=\ref[src];eject=1'>Eject item.</A><BR><BR>"
```
Arokha: Add a button for 'remove paint' (or some synonym) that sets color to initial(item.color)

code/game/machinery/painter_vr.dm
Verkister: Gonna see if I can figure out the template stuff

code/game/machinery/painter_vr.dm
```diff
@@ -0,0 +1,105 @@
...
+	else
+		for(var/atom/movable/O in processing)
+			dat += "Item inserted: [O]<HR>"
+		dat += "<A href='?src=\ref[src];select=1'>Select new color.</A><BR>"
+		dat += "Color: <font color='[activecolor]'>&#9899;</font>"
+		dat += "<A href='?src=\ref[src];paint=1'>Apply new color.</A><BR><BR>"
+		dat += "<A href='?src=\ref[src];clear=1'>Remove paintjob.</A><BR><BR>"
+		dat += "<A href='?src=\ref[src];eject=1'>Eject item.</A><BR><BR>"
+
+	user << browse(dat, "window=colormate")
```
Arokha: Still no fake nanoui?
I believe it's something like...
```
var/datum/browser/menu = new(user, "colormate","Color Mate Control Panel", 400, 600, src)
menu.set_content(dat)
menu.open()
```
... to replace the user << browse line and onclose line.

You'd wanna remove the <TITLE> thing at the top of the dat since title would be set by that 'color mate menu' thing there instead. Just have dat start as an empty string.

Test that to see if that works? Should end up with the nice black/blue nanoUI theme rather than white. Might need to putz with the 400/600 there to get a nice width/height.

Arokha: That's just the real one. The 'browser' datum is a fake nanoUI thing if they have pretty menus enabled in character setup. It's not real nanoui, doesn't use templates, etc.

Arokha:
Test please and lemme know if it works coded as such.

PR #3413, opened 2018-04-09
Fixes #3368 - Teleporting now often respects dropnom pref

### LINKED ISSUE #3368: Telebeacon noms should respect drop-prefs
The telenom code was added when we had the old map, before we had falling, and thus before we had dropnoms, and thus before we had a preference toggle that referred to the possibility of randomly ending up in a complete stranger's gut in one swift motion when certain conditions are met.

#### What currently happens
Telebeacons have the big "do you reeeaaally want to pick this up?" message/warning and so on, and there are discussions that pop up on whether it's valid to start beating the shit out of someone from the inside when you randomly teleport into them and your prefs are for not being prey and so on, but there's no way for the pred to know whether the person that's about to come through the teleporter is okay with teleporting into their stomach.

#### What it should be made to do instead
If the prey doesn't have their "drop nom prey" pref set to true, just dump them on the tile instead.

Arokha:
If you are teleporting someone to a beacon inside yourself, it's assumed that you know what you're doing (since it's ahelpable to just knowingly prefbreak people), but otherwise for standing teleporters and translocators, 'accidental' televore only happens if you have can_be_drop_prey on.

### DIFF
--- a/code/datums/helper_datums/teleport_vr.dm
+++ b/code/datums/helper_datums/teleport_vr.dm
@@ -1,13 +1,17 @@
 /datum/teleport/proc/try_televore()
 	//Destination is in a belly
-	if(isbelly(destination.loc))
+	if(isbelly(destination.loc) && isliving(teleatom))
+		var/mob/living/L = teleatom
 		var/obj/belly/B = destination.loc
-	
+		
+		if(!L.can_be_drop_prey) //Overloading this as a pref for 'want to be unexpectedly eaten'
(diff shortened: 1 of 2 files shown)


Screemonster:
I wonder if checking someone's OOC notes should give a little report of things like someone's dropnom status, digestibility, and so on. Out of scope for this though.

Arokha:
Out of scope yes, but not a bad idea.

PR #3418, opened 2018-04-09
Fix shuttle corner turfs disappearing when shuttle moves.

Leshana:
When shuttles land on a turf, it takes the appearance of the turf it lands on as an underlay.  Must make sure this underlay is on FLOAT_PLANE otherwise it will likely end up OVER the turf itself.

So we don't get shuttles looking like this:


Leshana:
PR Quick-merged to include it in tonight's release.

PR #3423, opened 2018-04-10
Some vore tweaks. (new gurglemodes will hopefully be back in another PR)

Verkister:
(Some parts have been removed to make room for their updated versions in new system)
-Makes absorption mute the drained absorbee's hunger noises if enabled.
-Makes borg sleeperbelly sounds use the new pref-respecting noise system.
-Adds a finisher gurgle and ding sound to finished borg belly self-clean cycle.
-Makes digested ID description state clearer that it is not indeed destroyed, but salvageable and recommended to return to hop instead of trashing okay. Bet most of you never had any idea.


Verkister:
Well I've kinda been seeing the ID digestion thing as a feature of the full digestion mode. Sounds kind of ridiculous how a card that can survive gurgles that destroy near everything are suddenly weaker than sheets of paper what comes to the mode literally designed to keep the prey's belongings so they can be returned afterwards.

I think modular system would be needing a lot of rewriting and I'd think having these in the same place with the other current features instead of having to dig through the reject commit dump will make the conversion easier when someone gets to work on it. In the meantime the single list of options has never been an inconvenience to my game experience despite my habits of changing the modes on the fly al the time.

SpadesNeil:
```I believe not being able to obtain IDs with access on them was a specific requirement of IDs surviving all digestion that @SpadesNeil wanted, but he can correct me if I'm wrong. It was a long time ago.```

This is accurate.

Arokha:
> I think modular system would be needing a lot of rewriting and I'd think having these in the same place with the other current features instead of having to dig through the reject commit dump will make the conversion easier when someone eventually gets to work on it.

It would take me about an hour, so I'd be happy to do it if you don't want to. Just remove it from yours (also here's a good reason you should do things in separate PRs, or at least separate commits per fix)

Verkister:
~~I thought griefing was against the rules again~~

Anyhow if that's the reason the IDs need to be ruined, why not just make them fully digestible then? After all the ruined cards are practically nothing but trash, only useful for trash pit divers to laugh at the people's demise.

I'll fix this PR after the hour then. The other fixes here aren't really anything that would be necessary to merge before the building of the new system.

Verkister:
Idea: Make the ID access expire when the owner leaves the round.

SpadesNeil:
No.

Because I might want to observe the round anyway.

There are many times I intentionally take my ID with me so it gets digested and destroyed, and when I am away after being sloshed doesn't mean I am logged out.

Verkister:
I meant the round quit thing that the ghosts can do without having to log out.

SpadesNeil:
Also no.

As for why IDs aren't entirely destroyed, it's because there is a limited supply of them and I have no intention of adding another method to add them.

Verkister:
Wait are the digested IDs fixable and reusable for that sort of resource saving?

SpadesNeil:
Yes, they are just reskinned and wiped IDs. They can be reused.

Verkister:
This is not currently common knowledge for what I know. I've never seen them reused or even heard it being possible until just now.

SpadesNeil:
Regardless, no. ID access should destroy access without destroying the ID, so the ID can continue to be used just for sake of proving you're crew until a new one is obtained or the old one is repaired. The digestion damage to the ID is a visual feature to indicate this.

Verkister:
Very well. We should probably add some sort of hints that the ruined ones can indeed be fixed or recycled as this information is currently 100% meta with no chance for people to get the info ingame.

Arokha:
You can change the desc on them if you want. "The data seems to be corrupted, but it could be reused... if you really want" or something.

So what's the plan on the gurgle mode things?

Verkister:
This is the part where my plan ends. For what I know the new thing would probably have digestion related options under one menu (damage options, item mode options, strip, numbing, etc.), maybe drain and absorb too, healing type stuff in under another menu (heal,nourish,unabsorb,etc.), and transform related stuff under another menu (plus size affecting things I suppose)

PR #3436, opened 2018-04-11
Nerfs apex metabolism

Verkister:
Now it's like more consistent with the steps between other modes.


Dinnel:
Awww, I liked being hungry that much.

Verkister:
Same although there were times it got little inconvenient (like hunger should). Just wanted to proof people it doesn't take full apex hunger to realize less than a liquidfood ration from a full sized prey is way too little.

PR #3443, opened 2018-04-11
Adds 'mode addons' feature to bellies to replace several modes

Arokha:
Removes several 'hybrid' digestion modes, like 'digest itemweak' and 'strip digest' and 'numbing digest' and instead just adds toggles for these features to be applied to any mode.
* Stripping: Strips items off one at a time, and digests them one at a time unless itemweak is on.
* Itemweak: Does not digest items, and instead just leaves them slimy.
* Numbing: Numbs prey to pain (or just for RP purposes).

**These DO save, unlike the main belly mode setting, so make sure to save them once you've got them set how you prefer.**

**An important thing to add, for 'reasons', all modes now 'deal with' items in your belly. Even 'Hold' will digest items, unless you have 'Itemweak' turned on.**

You can apply the above in any combination with any other mode.

For example:
image

Will drain nutrients from the prey, numb them, strip them of their clothes, and not digest the clothes.

If you want to recreate the three removed modes, that's simply the 'Digest' mode, with only one of the modes turned on.

Also, all the transformation settings were moved to a separate menu. One of the digest modes simply says 'Transform'. Picking that presents a separate menu of transform modes that you can select.

Also also, increased the default size of the Vore Panel window, to accommodate how it's grown over the years (+50px to width, +100px to height).

Also also also, increased the efficiency of processing bellies by a little by reworking how things are done and in what order.

### DIFF
--- a/code/__defines/belly_modes_vr.dm
+++ b/code/__defines/belly_modes_vr.dm
@@ -1,28 +1,38 @@
-// Belly Mode Constants
+// Normal digestion modes
 #define DM_HOLD									"Hold"
-#define DM_DIGEST									"Digest"
-#define DM_ITEMWEAK									"Digest (Item Friendly)"
-#define DM_STRIPDIGEST									"Strip Digest (Items Only)"
-#define DM_DIGEST_NUMB							"Digest (Numbing)"
+#define DM_DIGEST								"Digest"
+#define DM_ABSORB								"Absorb"
+#define DM_UNABSORB								"Unabsorb"
+#define DM_DRAIN								"Drain"
+#define DM_SHRINK								"Shrink"
+#define DM_GROW									"Grow"
+#define DM_SIZE_STEAL							"Size Steal"
 #define DM_HEAL									"Heal"
-#define DM_ABSORB									"Absorb"
+#define DM_EGG 									"Encase In Egg"
+#define DM_TRANSFORM							"Transform"
+
+//#define DM_ITEMWEAK							"Digest (Item Friendly)"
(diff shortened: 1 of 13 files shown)


Mewchild:
Request while you're making this? A "Clean" addon, that won't contaminate items/leave them slimey.

Arokha:
> Request while you're making this? A "Clean" addon, that won't contaminate items/leave them slimey.

Unfortunately no, due to the fact that administration has decided that among other things IDs must be 'treated', the way to do that that doesn't involve a ton of additional snowflake code to check 'oh is this something that needs to be treated super specially?', the existing digestion code is the way.

There's currently no way in the game to get items off prey cleanly, so you're not losing any features.

Verkister:
Sweet. I also just realized this will also fix a possible exploit I hadn't even thought about before which is good heh.

Screemonster:
Do they save per-mode, or just in general? As in, if my hold mode has itemweak and I switch to digestion, does my digestion also have to be itemweak unless I change the options?

Arokha:
> Do they save per-mode, or just in general

Per belly, but not per-belly-per-mode and not per-mode. Both are a little more complicated than I want to mess with.

I could do per-mode by taking the list of digest modes on bellies and making every entry in that list equal the flags that are on for that mode, but, that list right now isn't saved because if we add/remove things (like, for example, this PR), we wouldn't want to load the old non-existant modes anymore or anything, so I don't want to make that list start saving or anything.

In general I'm wary of making anything save a list of digest modes in any way, because of things like that.

PR #3445, opened 2018-04-12
Fix Wall Frames - Several Ways

### LINKED ISSUE #3435: Most construction objects do not interact with walls.
SHIT'S BROKE. 
Construction objects such as frames and light fixtures do not interact with walls at all.

For them to attach to the wall or give me the dialogue box to select what I want to put on the wall in the case of frames.

#### What actually happened
Nothing happened. I didn't even get the prompt that happens when you push against the wall with an empty hand.

#### Steps to reproduce
- Make a wall-mounted object such as a frame or light fixture.
- Click on a wall.
- Cry.

#### Code Revision
- 0ce6c47b6acec88cdd21acb5185e0bf6b604fa0a
 
- Tested in the engineering construction site. APCs and welders worked but frames and fixtures did not.

Leshana:
**Fix building frames on walls and windows. ( Fixes #3435 )**
`try_build()` has long accepted `user` parameter, but nothing that called try_build actually passed it! Upgraded try_build on APC to use the user parameter, and switched to use to_chat while I was there.

**Fix wall frames dropping steel and bugging on failure to place.**
Previously, if you tried to place a frame on a wall and it failed (due to something already being there etc) you would no longer be asked what type to use next time you try, and it will runtime.
Fixes this by performing the safety checks *before* asking you want you want to build.  Since the checks don't depend on what type you want to build anyway its fine.

**Fix allowing frames on walls with machine-under-construction on it already.**
Add `/object/structure/frame` to the list of objects that count as wall items. Also convert from strings to actual type paths ... and fix the paths that haven't existed for four years.  Properly speaking it should be a dynamically generated list at some point.

**Switch frame construction menu to have a cancel button.**
Instead of having a whole frame type dedicated to giving a "Cancel" option... lets just tell BYOND to add a cancel button to the menu.
cancel option

### DIFF
--- a/code/_helpers/unsorted.dm
+++ b/code/_helpers/unsorted.dm
@@ -1223,20 +1223,21 @@ proc/is_hot(obj/item/W as obj)
 
 /*
 Checks if that loc and dir has a item on the wall
+TODO - Fix this ancient list of wall items. Preferably make it dynamically populated. ~Leshana
 */
 var/list/WALLITEMS = list(
-	"/obj/machinery/power/apc", "/obj/machinery/alarm", "/obj/item/device/radio/intercom",
-	"/obj/structure/extinguisher_cabinet", "/obj/structure/reagent_dispensers/peppertank",
-	"/obj/machinery/status_display", "/obj/machinery/requests_console", "/obj/machinery/light_switch", "/obj/effect/sign",
-	"/obj/machinery/newscaster", "/obj/machinery/firealarm", "/obj/structure/noticeboard", "/obj/machinery/door_control",
-	"/obj/machinery/computer/security/telescreen", "/obj/machinery/embedded_controller/radio/simple_vent_controller",
-	"/obj/item/weapon/storage/secure/safe", "/obj/machinery/door_timer", "/obj/machinery/flasher", "/obj/machinery/keycard_auth",
-	"/obj/structure/mirror", "/obj/structure/closet/fireaxecabinet", "/obj/machinery/computer/security/telescreen/entertainment"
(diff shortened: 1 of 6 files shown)


code/_helpers/unsorted.dm
```diff
@@ -1223,20 +1223,21 @@ proc/is_hot(obj/item/W as obj)
...
+	/obj/machinery/status_display, /obj/machinery/requests_console, /obj/machinery/light_switch, /obj/structure/sign,
+	/obj/machinery/newscaster, /obj/machinery/firealarm, /obj/structure/noticeboard, /obj/machinery/button/remote,
+	/obj/machinery/computer/security/telescreen, /obj/machinery/embedded_controller/radio,
+	/obj/item/weapon/storage/secure/safe, /obj/machinery/door_timer, /obj/machinery/flasher, /obj/machinery/keycard_auth,
+	/obj/structure/mirror, /obj/structure/closet/fireaxecabinet, /obj/machinery/computer/security/telescreen/entertainment
 	)
 /proc/gotwallitem(loc, dir)
 	for(var/obj/O in loc)
 		for(var/item in WALLITEMS)
-			if(istype(O, text2path(item)))
```
Arokha: lol, this is pretty funny, here. was this before PATHS could be in a list??

code/game/machinery/frame.dm
```diff
@@ -180,9 +175,6 @@
 	x_offset = 24
 	y_offset = 24
 
-/datum/frame/frame_types/cancel //used to get out of input dialogue
-	name = "Cancel"
```
Arokha: Oh wow, they didn't want to just add "Cancel" to the list, which is what I did in vorepanel before I knew better. this is great.

PR #3448, opened 2018-04-12
Lazily implements taur riding

Arokha:
It's pretty lazy. I'm sure someone with *effort* can touch it up.

By the power of greyskull, battlecat and I shall defeat the enemies of nanotrasen


SpadesNeil:
If the taur gets knocked down, does it stun the rider too? The rider should be like 2x stunned. Maybe take damage.

Wickedtemp:
@Screemonster That's what was being used I think. It was a burst laser set to stun mode. 

@SpadesNeil If the taur is knocked down, the rider is also knocked down. I don't think they're stunned, but they do fall off and might end up 'resting'. If the taur is stunned but not knocked down, it probably won't affect the rider, and if the rider is stunned, it won't affect the taur.

KasparoVy:
Nice, haha. No more stuffing people into 'saddle' bellies like I've been seeing

Wickedtemp:
I mean, it'll probably still happen. Tempest can put Evelyn or Taiga on hir back due to them also being taurs. So later on I might see if the bit preventing taurs from riding larger taurs can get removed, in place of just a large size difference requirement.

Arokha:
> of course seeing two taser beams fire at once from the same spot is making me want some sort of burstfire taser

That's what we were shooting, burst tasers. The stun mode on the burst laser is a burst taser.

> If the taur gets knocked down, does it stun the rider too? The rider should be like 2x stunned. Maybe take damage.

Doesn't stun them, just throws them off. I mean, could do it, but idk, seems weh to do damage if you have a rider and decide to rest, it dumps them off, they take damage. Can do stun though.

SilencedMP5A5:
>driving taurs with riding crops
EVIL
also **_THE WINGED HUSSARS HAVE ARRIVED_**

Jakksergal:
Grand Theaft Chakat

code/game/objects/buckling.dm
```diff
@@ -75,6 +75,13 @@
 //	buckled_mob = M
 	buckled_mobs |= M
 
+	//VOREStation Add
+	if(riding_datum)
+		riding_datum.ridden = src
```
Leshana: wouldn't `ridden` already be `src`?

PR #3454, opened 2018-04-13
Removes the #s

Cameron653:
Ended up there since I copy&pasted their other items in the custom_items.txt.

### DIFF
--- a/config/custom_items.txt
+++ b/config/custom_items.txt
@@ -703,12 +703,12 @@ item_path: /obj/item/clothing/glasses/sunglasses/fluff/alfonso
 #req_access = 1
 #}
 
-#{
-#ckey: silencedmp5a5
-#character_name: NT-HASD #556
-#item_path: /obj/item/device/modkit_conversion/hasd
-#req_access = 1
-#}
+{
+ckey: silencedmp5a5
+character_name: NT-HASD #556
+item_path: /obj/item/device/modkit_conversion/hasd
+req_access = 1
+}
 
 {
 ckey: silvertalismen


Screemonster:
I just hope the # in his name isn't interpreted as a comment.

Arokha:
Might be! o: We'll find out.

config/custom_items.txt
```diff
@@ -703,12 +703,12 @@ item_path: /obj/item/clothing/glasses/sunglasses/fluff/alfonso
...
 
-#{
-#ckey: silencedmp5a5
-#character_name: NT-HASD #556
-#item_path: /obj/item/device/modkit_conversion/hasd
-#req_access = 1
-#}
+{
+ckey: silencedmp5a5
+character_name: NT-HASD #556
```
SpadesNeil: This is most certainly going to get fucked with. # is a comment.

KasparoVy: Able to quote it up? Is that a thing for these blocks?

SpadesNeil: Nope. That functionality is not supported, and also, the system it was designed for was never built to believe a serious character would ever use # in their name.

May I suggest the character replace "#" with "No." to denote Number. Alternatively, use Mk to denote Mark, or "v" to denote Version.

KasparoVy: Thanks for clearing that up

Arokha: `if(line == "" || !line || findtext(line, "#", 1, 2))` looks like it actually has to be the first or second character to comment it out, so this will probably work

Cameron653: Hopefully. Might as well try.

PR #3462, opened 2018-04-14
Fix a few Protean bugs

Arokha:
* Unbuckle targets before doing blobnoms
* Can blobsurf by buckling to protean
* Fix values mismatch causing infinite metal messages


SilverTalismen:
What is blobsurfing.

Arokha:
Buckling yourself to the blob

PR #3463, opened 2018-04-14
Makes donuts less of a strain on resources.

Ketrai:
considering you can put 6 donuts in a box, and they're supposed to be a small food item... this overrides all the donuts to use dough slices instead of a full wad of dough. 

Literally, who would realistically use two boxes of flour, six eggs, and whatnot to make one batch of donuts? I know virgo isn't meant to be realistic, but at least now a chef doesn't have to feel bad when someone asks for a box of  these things.


Little-119:
Should probably just send this to Polaris.

PR #3465, opened 2018-04-14
Lore fixes I forgot

SpadesNeil:
Forgot to enable mil pages, and also, this thing about political parties needs disabling.


SpadesNeil:
That didn't quite go correctly.

PR #3469, opened 2018-04-15
Lore fixes, v1.01a

SpadesNeil:
The "a" stands for arbitrary version control.

Also I had no idea how to do this correctly so I just copypasted like an asshat into a new branch. :B


SpadesNeil:
I'M BUSY AND BAD AT FOLLOWING DIRECTIONS.

SpadesNeil:
All done.

Issue #3453, opened 2018-04-12
FUN™ suggestions for giant rats


SpadesNeil:
#### Brief description of the issue
I'd like rats to be able to do a few NEAT things.

- I'd like a neutral mob version of the rat, which can appear during random events during the shift, but only in maint. It won't attack unless provoked, similar to the solar grub. Unlike the solar grubs, it would _not_ give an alert to the station's crew; you'd just find them by chance.
- If the rat smells you with food, it starts making emotes about the food in your inventory, specifically mentioning the food you have.
- If you drop the food, it'll leave you alone and eat the food like Ian.
- If you do not drop the food, it will attempt to follow you, getting gradually more aggressive in its emotes until it pounce-noms you.
- Attacking the rat will result in normal functionality of neutral mobs, being it attacks back at you, obviously.

#### Code Revision
- https://github.com/VOREStation/VOREStation/commit/73d841c9c8a25a044bdcdd761b578133df86034c
 
#### Anything else you may wish to add:
- Too complicated for me to figure out in a reasonable length of time but I figure Verk or Aronai may wanna try it. Maybe Cameron or Scree. I don't have the free time to figure out cool things like I used to. :C

Arokha:
What is a MODERATE SUGGESTION :V

KasparoVy:
**Pri: 0-Gamebreaking**

SilverTalismen:
I say it should do everything you've suggested, it'll spawn occasionally in maint as neutral and only bother with you if you have food and become more and more aggressive before eating you

SpadesNeil:
>All issues are to be given a priority. Some pull requests are given priority, but this is not required. Priority is defined as follows from most to least urgent as follows:
>(...)
>(As a feature) An existing functionality is improved or changed to better perform its original purpose.

That said, maybe I should change the "priority" system to "impact" on gameplay.

Very-Soft:
I like this idea a lot!

Screemonster:
Sounds like the way we wanted noodle to behave with the snek-snax.

Ketrai:
I'm still too bad with byond coding to attempt this, but maybe. just maybe.

Verkister:
Probably gonna try this today

Verkister:
Got it working :D

lbnesquik:
https://github.com/VOREStation/VOREStation/pull/3472

PR #3472, opened 2018-04-15
Adds 'neutral' food begging giant rat.

Verkister:
-Adds neutral mob version of the giant rat.
-Will not attack anything, just retaliate.
-It however will smell food in (mobvore enabled) people's hands/pockets (but not in deeper inventory storage), which will attract its curiosity and make the rat follow the person around until fed or evaded. Evasion increases its hunger cheekiness though. (but not too much so people can't evade stack it into pouncing the next victim on sight)
-The forementioned hunger cheekiness system builds up whenever people ignore its begging or run away.
-When caught up with the food carrying person, the rat will beg for scraps every now and then, getting increasingly "hangry" each time it happens. As the hunger cheekiness rises, the begging messages get more aggressive, and the last one will be the last warning before the rat pounces and devours you.
-Feeding the rat by applying the snack will get it off your tail and reset the hunger.
-The rat will also look for food on the ground, so throwing snacks on the floor or just coming across a conveniently placed snack makes the rat gobble up the floor snack and getting off your tail with satisfied hunger.
-Also getting ate by the rat resets the attraction and hunger.
-Also evasion blueballing the rat after building up hunger past half of maximum will reset the hunger to half to leave room for interaction with the next victim.
-This rat can now be acquired from pred crate as well.

Does #3453


SpadesNeil:
I'll figure out how to make an event with them.

code/modules/mob/living/simple_animal/vore/rous.dm
```diff
@@ -43,4 +43,117 @@
...
+			if(prob(25))
+				visible_emote(pick("growls at [food], sounding rather hangry!",
+								   "aggressively bumps and nudges against [food], trying to make something fall out.",
+								   "salivates at [food] in an unsettling manner.",
+								   "pushes hard against [food], licking its chops.",
+								   "almost sinks its teeth into [food], just stopping to give them another chance."))
+				hanger += 5
+		else if(hanger < 50)
+			visible_emote("appears to have had enough and prepares to strike!")
+		else
```
SpadesNeil: I feel like there is probably a better way of doing this last bit, as I think it's handled in the attack() proc, but I could be wrong.

code/modules/mob/living/simple_animal/vore/rous.dm
```diff
@@ -43,4 +43,117 @@
...
+		icon_state = icon_dead
+		update_icon()
 	..()
+
+/mob/living/simple_animal/hostile/rous/passive
+	name = "curious giant rat"
+	desc = "In what passes for a hierarchy among verminous rodents, this one is king. It seems to be more interested on scavenging."
+	follow_dist = 1
+	var/mob/living/carbon/human/food
+	var/hanger = 0
```
SpadesNeil: Please use variables and pathing names that make sense and are spelled correctly. `hanger` should be `hunger` for ease of reading by other devs. Memey code is not acceptable.

code/modules/mob/living/simple_animal/vore/rous.dm
```diff
@@ -43,4 +43,117 @@
 
 /mob/living/simple_animal/hostile/rous/death()
 	playsound(src, 'sound/effects/mouse_squeak_loud.ogg', 50, 1)
+	if(resting)
+		resting = 0
+		icon_state = icon_dead
+		update_icon()
 	..()
+
+/mob/living/simple_animal/hostile/rous/passive
```
SpadesNeil: Also why is it called rous? The hell is a rous?

Screemonster: oh my god have you not seen the princess bride

Rodent Of Unusual Size

code/modules/mob/living/simple_animal/vore/rous.dm
Verkister: It's sort of a passive state of following and begging. I feel like putting it into attack proc would mess things up what comes to regular combat against these.

code/modules/mob/living/simple_animal/vore/rous.dm
Verkister: Also shorter than "giant_rat" and less misleading than simple "rat" heh

SpadesNeil: I'm going to change this because admins aren't going to know that innately when trying to spawn them.

Ketrai:
gee, wish I was so good at coding in new features xD, but hey, I have plans.

code/modules/mob/living/simple_animal/vore/rous.dm
```diff
@@ -41,6 +41,121 @@
...
+	desc = "In what passes for a hierarchy among verminous rodents, this one is king. It seems to be more interested on scavenging."
+	follow_dist = 1
+	var/mob/living/carbon/human/food
+	var/hunger = 0
+
+/mob/living/simple_animal/hostile/bigrat/passive/Life()
+	. = ..()
+	if(!. || ai_inactive) return
+
+	for(var/obj/item/weapon/reagent_containers/food/snacks/S in oview(src,7)) //Accept thrown offerings and scavenge surroundings.
```
Arokha: Yikes, a little expensive. This will probably preclude these from ever being in mob spawners, given how expensive view is. Should also be `world.view` instead of 7 (or one of the SA-specific range vars).

code/modules/mob/living/simple_animal/vore/rous.dm
```diff
@@ -41,6 +41,121 @@
...
+		if(isturf(S.loc))
```
Arokha: Will never not be a turf, as view does not include contents, so this is unnecessary.

The only possible case I can think of is if they're in something, like a belly, because MAYBE view includes contents of the thing you're in (probably not, though).

code/modules/mob/living/simple_animal/vore/rous.dm
```diff
@@ -41,6 +41,121 @@
...
+				food = null
+			else
+				WanderTowards(S.loc)
+			break
+
+	if(!food) return
+
+	var/food_dist = get_dist(src,food)
+
+	if(food_dist > 7) //Lose interest on this person.
```
Arokha: `world.view` instead of 7, or an SA-range var.

code/modules/mob/living/simple_animal/vore/rous.dm
```diff
@@ -41,6 +41,121 @@
...
+		food = null
+		if(hunger < 25) //Get miffed but don't let people evade stack you into pouncing the next victim on sight.
+			hunger += 5
+		else if(hunger > 25) //Leave room for interaction with the next victim if blueballed by evader.
+			hunger = 25
```
Arokha: Can replace all of this with `hunger = Clamp(hunger+5, 0, 25)` which sets hunger equal to hunger+5, with a min possible of 0, and a max possible of 25.

code/modules/mob/living/simple_animal/vore/rous.dm
```diff
@@ -41,6 +41,121 @@
...
+		food = null
+		return
+	. = ..()
+
+/mob/living/simple_animal/hostile/bigrat/passive/Found(var/atom/found_atom)
+	if(!SA_attackable(found_atom))
+		return null
+	else if(ishuman(found_atom) && will_eat(found_atom))
+		var/mob/living/carbon/human/H = found_atom
+		for(var/obj/item/weapon/reagent_containers/food/snacks/S in H)
```
Arokha: Also pretty expensive. Too bad there's not another way to do this.

code/modules/mob/living/simple_animal/vore/rous.dm
```diff
@@ -41,6 +41,121 @@
...
+				lay_down()
+			food = H
+			return found_atom
+			break
+	return null
+
+/mob/living/simple_animal/hostile/bigrat/passive/FindTarget()
+	var/atom/T = null
+	for(var/atom/A in ListTargets(view_range))
+		if(A == src)
```
Arokha: What's the purpose of overriding this?

Verkister: To make use of the "hostile" mob type looking for targets while idling. The parent proc would force the mob to start attacking the target as well.

Arokha: You can set the specific_targets var instead, which does basically exactly what you want. If set, it will only test Found() and ignore the rest of FindTarget.

Verkister: Yeah, this does only test for the found targets, but the parent FindTarget() will still force STANCE_ATTACK on the found target.

Verkister: Alright testing it now

SpadesNeil:
@Leshana !

Verkister:
Travis failure seems unrelated.

SpadesNeil:
24 days since this was posted. Could use some feedback.

SpadesNeil:
Okay I'm not waiting any longer. Fix the conflict and I'll merge this.

Ketrai:
Hang on, if this adds a mob, it'll likely break the smol pull I have up, or vice versa, make sure you don't merge both at the same time by accident or something.

SpadesNeil:
1. This has been up for over a month so this takes precedence.
2. It doesn't add or change icons so it should not conflict badly with https://github.com/VOREStation/VOREStation/pull/3763

Arokha:
[3:32 PM] Leshana: This looks fine as long as we never put any on the map
[3:32 PM] Aronai: Can I quote you on that
[3:32 PM] Leshana: yup!

PR #3480, opened 2018-04-16
Updates sickshot description.

Verkister:
Also making it little more clear that the only way to get victims out of uncooperative or dead things is no longer live human exclusive.


Arokha:
I missed this when adding the living thing (it used to be human only).

PR #3481, opened 2018-04-16
Adds a public garden and a juice bar

Very-Soft:
Originally the public garden was going to take up this whole space, but I cut it back and then condensed it, and was able to do a little more with the space as a result.

z3
This is the public garden! It has 8 growing plots, 2 plant analyzers, 2 buckets, 1 mini hoe, 1 sink, 1 water tank, 1 seed vendor, and 1 nutrimax

z2
This is a juice bar. It has a seed extractor, a smartfridge, a grinder, a box of glasses, and a box of straws. It also has some fire safety gear, and a random plush.

z1

This is the entry area. It's mostly empty for now. I have some plans for it later maybe. There is also a section of maintenence here, so you can be the creep behind the scrubbers if you want to. There are also some places to sit and relax, as well as a hot drinks and snacks machine to shoot food at you while you do so.


Little-119:
That's a lot of merge commits :eyes:

Very-Soft:
yeye, workin' on it

Very-Soft:
Closing to fix!

PR #3483, opened 2018-04-17
Add ability for shapeshifters to change ears, tails, wings

Arokha:
So prometheans and proteans, basically.


SilverTalismen:
Great work, will color choosing be a option thing that pops up after you choose the tail/wing/ears? :o

Arokha:
Yes, unless you click cancel, it then asks you for the color(s). I didn't want to add separate verbs for colors because I don't want to clutter that verb tab too much.

It always highlights your existing style, so if you want to change color you can just click 'Change Ears' and ignore the first choice, clicking OK, since it will just pick the same style you already have, then you can pick colors. Make sense?

SilverTalismen:
Yeah makes perfect sense to me, Great work as always ^^

PR #3489, opened 2018-04-18
Adds public garden+juice bar, updates mapmerge2

Arokha:
### All credit goes to @Very-Soft for the map and design and everything.

Originally the public garden was going to take up this whole space, but I cut it back and then condensed it, and was able to do a little more with the space as a result.

z3

This is the public garden! It has 8 growing plots, 2 plant analyzers, 2 buckets, 1 mini hoe, 1 sink, 1 water tank, 1 seed vendor, and 1 nutrimax

z2

This is a juice bar. It has a seed extractor, a smartfridge, a grinder, a box of glasses, and a box of straws. It also has some fire safety gear, and a random plush.

z1

This is the entry area. It's mostly empty for now. I have some plans for it later maybe. There is also a section of maintenence here, so you can be the creep around in the dark or behind the scrubbers if you want to. There are also some places to sit and relax, as well as a hot drinks and snacks machine to shoot food at you while you do so.

The z05 change is removing keys beyond the ymp boundary that existed in the file. Unusual that byond allowed it to exist like that. Who knows what it could have caused.

The update to mapmerge fixes it so that it doesn't attempt to allocate keys beyond ymp.

### DIFF
--- a/maps/tether/tether-01-surface1.dmm
+++ b/maps/tether/tether-01-surface1.dmm
@@ -150,7 +150,8 @@
 	},
 /obj/machinery/atmospherics/pipe/simple/hidden/cyan{
 	dir = 10;
-	icon_state = "intact"
+	icon_state = "intact";
+	tag = "icon-intact-f (NORTHWEST)"
 	},
 /turf/simulated/floor/tiled,
 /area/tether/surfacebase/mining_main/airlock)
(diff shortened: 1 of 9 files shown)


Screemonster:
Countdown to grabbing some sucker directly from the juice bar and into that little maintenance tunnel behind the gateway which, lacking trashpiles or any particular through-route anywhere, should be the perfect spot for some undisturbed clandestine nomfs :9

Very-Soft:
Yus! The z1 maintenance is also super dark in the little side spot there,
and again, no trash piles.

PR #3491, opened 2018-04-18
Fixes brig common area disposal

Little-119:


SilverTalismen:
So that if someone tries to escape all they'll do is space themselves. 👍

SpadesNeil:
This is much better.

SpadesNeil:
Fuck'em.

Little-119:
This is my legacy, and the move to the surface is destroying it ;-;

PR #3493, opened 2018-04-18
Xenobio lab updates

Mewchild:
Adds two slime scanners, two welding goggles, two welding tools, and a fuel tank to Xenobio. Removes the loose extinguisher, adds a pair of extinguisher cabinets in the lab (for space efficiency). Told Cameron I'd do this a week ago, and forgot until I saw the branch sitting sad and alone in my sourcetree :^)

image

### DIFF
--- a/maps/tether/tether-01-surface1.dmm
+++ b/maps/tether/tether-01-surface1.dmm
@@ -13052,6 +13052,12 @@
 /obj/machinery/door/firedoor/glass/hidden/steel{
 	dir = 1
 	},
+/obj/structure/extinguisher_cabinet{
+	dir = 1;
+	icon_state = "extinguisher_closed";
+	pixel_y = -32;
+	tag = "icon-extinguisher_closed (NORTH)"
+	},
 /turf/simulated/floor/tiled,
 /area/rnd/hallway)
 "aEG" = (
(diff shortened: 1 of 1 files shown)


maps/tether/tether-01-surface1.dmm
```diff
@@ -13052,6 +13052,12 @@
 /obj/machinery/door/firedoor/glass/hidden/steel{
 	dir = 1
 	},
+/obj/structure/extinguisher_cabinet{
+	dir = 1;
+	icon_state = "extinguisher_closed";
+	pixel_y = -32;
+	tag = "icon-extinguisher_closed (NORTH)"
```
Arokha: Do me a favor and scrape these tags out. Remove the semicolon from the line above after.

maps/tether/tether-01-surface1.dmm
```diff
@@ -13942,15 +13948,20 @@
 /obj/item/stack/material/phoron,
 /obj/item/stack/material/phoron,
 /obj/machinery/reagentgrinder,
+/obj/structure/extinguisher_cabinet{
+	dir = 8;
+	icon_state = "extinguisher_closed";
+	pixel_x = 30;
+	tag = "icon-extinguisher_closed (WEST)"
```
Arokha: Tags are basically bad, but DreamMaker loves to add them if you use certain combinations of 'Generate instances from directions' or '...icon_states'. It makes all those objects the same, like anything with the same tag is considered the same object in some cases, and can cause weirdness.

Tags also make objects immune to garbage collection which is bad, because tags are considered a self-reference, so their refcount will never drop to 0.

Mewchild: You got it, bossfox

Mewchild: Doing a search, I found several other "tag =" items on this level. want me to clean those out too? Some atmos equipment, guest pass printers, etc

Arokha: Hmm. No, I'll make a separate PR and remove all of those. There's always the chance that something actually relies on the tag functionality and I wanna check first. Just remove these two for now.

Mewchild: Got it! Thank you.

PR #3494, opened 2018-04-18
Remove tag var from all Tether maps

Arokha:
And submaps.


Arokha:
Just merging this so that there's no risk of map conflicts since I want to do other map changes, and this doesn't actually change any game or map thing, it's just cleanup.

PR #3496, opened 2018-04-19
Ports /tg/ Quantum Pads

Arokha:
Pad-to-pad teleporters, though they cost 75% of a diamond for each one (since they require a bluespace crystal each).

image

They require a capacitor (increases teleport speeds), manipulator (reduces power usage), bluespace crystal (real or artificial), and 5 cablecoil. Linked with multitool (panel open = save to tool, panel closed = link saved to this pad). They can be linked in any direction (not just 1<->1, you could make a circular route of 5).

Unlike /tg/'s, these must be directly connected to a powernet. They consume idle power from an APC and will stop working if there's no APC, but additionally require enough power in the directly attached powernet to actually do teleports.

You cannot use them on any shuttle, they refuse to work, however you can build them on steady ground in the exploration areas (so as to set up permanent transport to the beach for the crew, for example). You could also set one up to bypass the elevator.


SpadesNeil:
Telenoms y/n

Also I feel these should be more expensive, else what is the point of having any other technology.

Arokha:
We half-removed telenoms before, I think, since it was being 'problematic' with things like the Spa event, people lined up to teleport, and all ending up inside each other instead of just going to the event. Like the 'completely accidental, I was just standing on a beacon' telenoms are removed.

5 seconds after clicking, 30 seconds after each teleport, I believe. The downside is the large power draw, that they require being connected directly to a powernet, and that they require diamonds to build each.

SpadesNeil:
I w

SpadesNeil:
Stupid fucking phone keyboard.

Anyway I would like if we actually had that feature for telenoms because
the cooldown should mitigate any sort of accidental vore.

SpadesNeil:
Stupid fucking phone keyboard.

Anyway I would like if we actually had that feature for telenoms because
the cooldown should mitigate any sort of accidental vore.

Arokha:
Can discuss when you're available in adminchat, but not really part of this PR :p

code/modules/telesci/quantum_pad.dm
```diff
@@ -0,0 +1,190 @@
...
+			return
+
+		teleporting = 0
+		last_teleport = world.time
+
+		// use a lot of power
+		var/power_to_use = 10000 / power_efficiency
+		if(draw_power(power_to_use) != power_to_use)
+			to_chat(user, "<span class='warning'>Power is not sufficient to complete a teleport. Teleport aborted.</span>")
+			teleporting = 0
```
Leshana: dooblie do, see line 154

Arokha: Fixed

code/modules/telesci/quantum_pad.dm
```diff
@@ -0,0 +1,189 @@
...
+	sparks.set_up(5, 1, get_turf(src))
+	sparks.start()
+
+/obj/machinery/power/quantumpad/attack_ghost(mob/observer/dead/ghost)
+	. = ..()
+	if(.)
+		return
+	if(!linked_pad && map_pad_link_id)
+		initMappedLink()
+	if(linked_pad)
```
Leshana: To prevent ghosts from traveling to nullspace, QDELETED()

Leshana: To prevent ghosts going to nullspace, use QDELETED()

Arokha: Fixed

PR #3497, opened 2018-04-19
 Nerf marble

Arokha:
Please.

Same damage as durasteel? That sure makes sense.


Jakksergal:
Might want to send this to Polaris too? They're pretty big on nerfing everything.

SpadesNeil:
image

PR #3498, opened 2018-04-19
Allow teleportation to CentCom map level

ReddeyfishVor:
This fixes rifts generated by singularity + supermatter, which usually teleport to a map on the CentCom level.


Arokha:
Gonna conflict with a pending VS PR.

The PR in question adds a `local` var to do_teleport usage, so you can tell whether or not it's local z-levels only, or can be anywhere.

Leshana:
Once #3496 is merged, the rifts could be changed to pass `local=FALSE` to `do_teleport()` right?

Arokha:
Yes

PR #3501, opened 2018-04-19
Cleans up custom species icon base code

### LINKED ISSUE #3478: Custom Species Base Body Coloring Issues
The Color settings for the "Custom Species" base are horribly off.

Setting a the body color to (green, orange, blue, etc) would result in something similar to that color,

#### What actually happened
The body will only shift between a yellow or white hue.

#### Steps to reproduce
- Go to character creator and select custom species, either the base body type or as a species.
- Change the body color around and watch it not be anywhere close to that color.

#### Code Revision
- Server revision: release - 
6b58de3a2baf315de4b23d50faf616881d79d305 
 
- 
image

image

image

Arokha:
Fixes #3478 

Prevents you from picking 'Custom Species' as an icon base.

### DIFF
--- a/code/_helpers/global_lists_vr.dm
+++ b/code/_helpers/global_lists_vr.dm
@@ -11,6 +11,8 @@ var/global/list/positive_traits = list()	// Positive custom species traits, inde
 var/global/list/traits_costs = list()		// Just path = cost list, saves time in char setup
 var/global/list/all_traits = list()			// All of 'em at once (same instances)
 
+var/global/list/custom_species_bases = list() // Species that can be used for a Custom Species icon base
+
 //stores numeric player size options indexed by name
 var/global/list/player_sizes_list = list(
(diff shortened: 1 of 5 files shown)


code/_helpers/global_lists_vr.dm
```diff
@@ -195,4 +197,14 @@ var/global/list/edible_trash = list(/obj/item/trash,
 			if(0.1 to INFINITY)
 				positive_traits[path] = instance
 
-	return 1 // Hooks must return 1
\ No newline at end of file
+	// Custom species icon bases
+	var/list/blacklisted_icons = list("Custom Species","Promethean") //Just ones that won't work well.
```
Leshana: Cannae we use the defines?

Arokha: How now

PR #3524, opened 2018-04-21
Fixed koi on Asteroid 1 not being poisonous

Heroman3003:
Yeah, don't get why aren't they.


Arokha:
Was probably just an oversight on whoever mapped them's part.

PR #3527, opened 2018-04-21
Adds in Spirit markings (that should work this time)

kcin2001:
w/o markings
image

w/ markings
image

also available are headless versions of the lights and panels markings for people who like monitor heads or want a spirit head


kcin2001:
@Arokha i have remade my pr and it seems to not be conflicting this time

PR #3545, opened 2018-04-23
Final LeMat Fix

HzDonut:
So turns out a lot of the stuff I was told to remove was actually necessary in the end since my nine-shot .38 LeMat ended up being a six-shot .357 LeMat. Also makes it so the cryo consoles don't store the revolver. No I'm not listening to people saying to change it anymore, no I'm not ever messing with this code ever again. It _finally_ works as intended. On the upside, it didn't inherit the 'projectile_type = /obj/item/projectile/bullet/pistol/strong' from /obj/item/weapon/gun/projectile/revolver which would've given me a gun that does about 60 damage a shot, so I can keep playing until this merges since it's just a revolver that takes .357 ammo but does .38 damage. This whole nightmare would've been avoided if I just ignored the 'advice' and orders from the very start.


SpadesNeil:
I don't think `preserve_item = FALSE` does anything _yet_ but as soon as I figure out how the fuck the cryo consoles work, it will do things.

SpadesNeil:
Oh damn I got some shit to fix, then.

PR #3546, opened 2018-04-23
Port /tg/ lavaland shelters

Arokha:
Nanotech not bluespace but it's just fluff either way.

Preview gif too big to embed


code/controllers/subsystems/mapping_vr.dm
```diff
@@ -88,6 +89,15 @@ SUBSYSTEM_DEF(mapping)
 			else
 				MT.load_new_z(centered = FALSE)
 
+/datum/controller/subsystem/mapping/proc/preloadShelterTemplates()
```
Leshana: Why is this necessary? Don't these already get loaded into `global.map_templates`?

code/controllers/subsystems/air.dm
```diff
@@ -223,8 +223,11 @@ Total Unsimulated Turfs: [world.maxx*world.maxy*world.maxz - simulated_turf_coun
 
 // ZAS might displace objects as the map loads if an air tick is processed mid-load.
 /datum/controller/subsystem/air/StartLoadingMap(var/quiet = TRUE)
-	can_fire = FALSE
```
Leshana: Please fix indenting on this proc to be tabs.

PR #3547, opened 2018-04-23
Prevents cryo-consoles from saving guns

SpadesNeil:
because they are guns

### DIFF
--- a/code/modules/projectiles/gun_vr.dm
+++ b/code/modules/projectiles/gun_vr.dm
@@ -1,0 +1,2 @@
+/obj/item/weapon/gun
+	preserve_item = FALSE // Prevents cryo-consoles from saving guns, because they are guns.
--- a/vorestation.dme
+++ b/vorestation.dme
@@ -2402,6 +2402,7 @@
 #include "code\modules\projectiles\dnalocking.dm"
 #include "code\modules\projectiles\effects.dm"
 #include "code\modules\projectiles\gun.dm"
+#include "code\modules\projectiles\gun_vr.dm"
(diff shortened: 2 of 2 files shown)


Arokha:
Can I suggest an alternative? Instead, we just rewrite cryoconsoles to log items (and log them to the game log), but delete them regardless.

That way we can use Splunk and be like "Oh, Ace literally never puts his gear back before cryoing" etc

Ketrai:
Better yet, instead of having the guns lost, can we get it so that certain items can only be accessed with certain IDS? as would make sense?

Screemonster:
Having the preserve_item var be "what access is needed to retrieve this" rather than a binary "preserve or nah" would be pretty neato.

Arokha:
Except 'TRUE' == 1 == ACCESS_ARMORY so everything is just 'armory' by default for all legacy items. Which might be okay?

Screemonster:
Given that a lot of the retrievable items are just clothing, might be overkill. Also leaves the question of "what do we with things like clothing that we want to be preserved but don't want to access-restrict".

There... _might_ be a solution that doesn't involve going through and defining a value on every item but I can't think of it right now.

SigholtStarsong:
Might I suggest referencing Paradise Station's github? I'm pretty sure that their cryo console is set up to check for accesses.

SpadesNeil:
We used to have a cryo access requirement, and I have no idea why we never ported it back in.

SpadesNeil:
@Arokha Do you want to write that? If so, close and replace this. Otherwise I'm going to merge this later.

Arokha:
I'll write it.

PR #3552, opened 2018-04-24
Dufflebag of Holding

Novacat:
Adds a dufflebag of holding, essentially a bag of holding reskin that functions exactly identical, so far can only be constructed in R&D.


Ketrai:
as long as it remains clear when people are wearing these things, for the safety of quantum pads.

Arokha:
Spoke with, will send to Polaris instead

PR #3553, opened 2018-04-24
Adds a sneaky room to Z2 bar maint

Very-Soft:
Adds a dark, sneaky room to bar maintenence to sort of give a purpose to that little pointless side hallway.

Contains a bunch of machine frames, and 2 computer frames (they don't look like computer frames in DM sorry)

Also contains 1 pet bed, 1 lizard, and 1 newspaper.

fuel processing

There is a false wall marked by a strange symbol on the north wall to serve as the intended entrance.

The thinking behind this one was that it was meant to be a sort of substation for processing air from the moon's atmosphere into usable starship fuel. The machines were stripped down though in favor of another option, or perhaps revisiting at a later date, and the room was sealed off. Since then, _someone_ has been slacking off in there! They even have a pet lizard.

For your consideration!

### DIFF
--- a/maps/tether/tether-02-surface2.dmm
+++ b/maps/tether/tether-02-surface2.dmm
@@ -17686,6 +17686,32 @@
 	},
 /turf/simulated/floor/tiled,
 /area/tether/surfacebase/public_garden_two)
+"IX" = (
+/obj/structure/grille,
+/obj/structure/window/reinforced/full,
+/obj/machinery/door/firedoor,
+/obj/structure/window/reinforced{
+	dir = 8;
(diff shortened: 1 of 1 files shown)


Arokha:
Windows not consistent with standards. The standard is: plating, grille, fullsize window, then a directional window on the side with the 'dangerous' atmos.

Little-119:
Add several empty bottles of whiskey and a red fox plushie next to the pet bed and name the lizard Laskorreshkor.

SpadesNeil:
Wrong symbol I think but I'll fix it later.

PR #3555, opened 2018-04-24
Ports /tg/ and/or Paradise ORM, + The Underdark

Arokha:
@Leshana did basically all the actual important porting work. I just ported a couple of the fun items because I like porting the silly things.

Ports the mining stuff from /tg/ (aka the ore rewards thing, points, blah blah), lots of the items from said machine, and adds a mining area beneath the existing mining map (accessible via stairs in the top-right corner) that has hostile mobs and better ores (and the main mining area's ores have been altered slightly to make less rare ores show up above, and more common ores+phoron).

It's a little much to list all the ore rewards things here, but it's "most of the stuff you can get on tg, except not a mining drone, and not a kinetic crusher yet".

Ok have fun.


Screemonster:
If this is gonna be the hazardous mining area and they have their own miner guns, might as well remove the beltminer crate. Or replace its contents with KAs rather than APPs.

Arokha:
I'd miss people blowing themselves up randomly. ;(

KasparoVy:
God bless. Excited for the crusher.

Ketrai:
Don't forget to add grubs to the mines! Well tbh, would need to code in the hostile grubs.

Arokha:
Well let's see what this breaks.

PR #3557, opened 2018-04-25
Adds more maintenance to Space 2

Very-Soft:
Space side is rather lacking in maintenance that isn't just tunnels with trash piles in them

I found that there's a section of Z6 with a lot of open space, where the structure of the station makes it unlikely to be used to expand into. As such, I thought it might be a good place for some dimly lit maintenance!

maint 1

Here you have the entrance. It goes passed tech storage. There are some uncomfortable chairs here to gaze out at the stars. There's a microwave in a small side room here here, for all your donk pocket warming needs.

maint 2

Further along the tunnel, there are some unfinished areas. An abandoned surgery room, perhaps originally intended for a brig physician. There is a false wall here between the two railings to get into the very unfinished room above. 

maint 3

Even further along there are a few features. A room to look out at the disposal chute, to check and see if there are any brig escapees. There is also a large-ish airtight cave here. There is no entrance presently, but were one to open it up, they would have to fix the atmosphere inside before they can use the full space. There is also a mysterious room behind the chapel, with a rune, some blood, papers, and trash. There is also no entrance to this room presently, but were one to make one, it is safe inside. 

There is a false wall  at the end of the stretch of pre-existing maintenance here, next to the APC, to serve as a secret alternate entrance, and to connect the tunnels. There is nothing to mark this hidden entrance presently, as it is mildly unneeded considering the other entrance.

There were very minor changes to Z7, to add a roof to the new section of Z6.

For your consideration!

### DIFF
--- a/maps/tether/tether-06-station2.dmm
+++ b/maps/tether/tether-06-station2.dmm
@@ -466,14 +466,13 @@
 	d2 = 2;
 	icon_state = "0-2"
 	},
-/obj/machinery/power/apc{
-	dir = 1;
-	name = "north bump";
-	pixel_x = 0;
-	pixel_y = 28
-	},
 /obj/effect/decal/cleanable/dirt,
 /obj/effect/floor_decal/rust,
+/obj/machinery/power/apc{
+	dir = 8;
+	name = "west bump";
+	pixel_x = -28
+	},
 /turf/simulated/floor,
 /area/maintenance/station/sec_lower)
 "aO" = (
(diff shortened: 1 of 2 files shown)


ArgobargSoup:
Okay, this, right here?  I love ~~you~~ this.

This chunk of maintenance you've added has all the flavor and personality that you see in the maintenance on many other maps...  there's just something about it that feels like proper SS13 maintenance.

Just... that spot in the first shot, with the chairs and the windows, that was enough ta make me go "mmm~", and then the rest...  👍

Little-119:
_First, the disposal to space of justice was a joke PR_
_Then that joke PR got merged_
_Now someone has apparently built a maintenance room over it to accomodate it_

what is my life

Very-Soft:
Honestly while it may have been a joke, it's pretty in line with a lot of attitudes toward folk who wind up trying to escape the brig.

Oh you want to try to escape by hopping down disposals?

image

I just figured that it should have a bit of maintenance to go with it, because you won't be able to repair the pipes otherwise should something happen to them!

Little-119:
It also gives a Shawshank Redemption-y method of escape if you have someone on the outside to cut the pipe, since they can freely enter that room to cut the pipe. That is, assuming the short trip only does minor damage to you.

Screemonster:
And if anyone can get unfettered access to help you with your escape here, even on the current map they can falsewall into the brig bathroom so it's not _really_ making the brig any more vulnerable to rescues.

Little-119:
@Screemonster Isn't every job (except visitors?) given access to maintenance, including the room above that disposal pipe? I'm not saying this is a bad thing, to be clear. Really, I'm all for prison breaks (assuming the prison breaks in question involve more than smashing a window or something else any baldie is capable of).

Arokha:
Seems fine as long as people don't just meme break into security every shift or meme around with the medical stuff.

PR #3558, opened 2018-04-25
Updates the bar!

Very-Soft:
**This design is by  @Screemonster !** (with a smol edit by me)

new bar b

Updates the bar to add in bathrooms! Also moves the bar's back room south of the bar. This all results in a little extra seating. Moves the arcade machine, removes the food and drink vending machines. Adds a set of rainbow clothes into the smol open space south of the big scrubber.

Adds a one directional window to the east side, which it was lacking for some reason.

For your consideration!

>update

Made bathroom stalls 1x1, and changed bathroom floor tiles to white, added floor decoration by the entrance to match the style of the other bathrooms on the station

### DIFF
--- a/maps/tether/tether-03-surface3.dmm
+++ b/maps/tether/tether-03-surface3.dmm
@@ -14036,10 +14036,6 @@
 "yG" = (
 /turf/simulated/floor/tiled,
 /area/hallway/lower/third_south)
-"yH" = (
-/obj/machinery/vending/snack,
-/turf/simulated/floor/wood,
-/area/crew_quarters/bar)
 "yI" = (
 /obj/machinery/atmospherics/pipe/simple/hidden/supply,
 /obj/machinery/atmospherics/pipe/manifold/hidden/scrubbers{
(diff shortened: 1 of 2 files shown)


Little-119:
Now you don't have to go as far to get your puke in a toilet after getting hammered

Although, does a bathroom with two toilets and no urinals need four sinks? Plus, I think all of the other bathrooms have one tile for the stalls.

Arokha:
Hmm, if you reduce the stalls to one tile then you can have that little walkway be a little wider which might be nice.

Very-Soft:
UE can be obtained by walking into the kitchen and grabbing it. Can also be ordered from cargo in one of the kitchen supply crates afaik. Easily obtained. Could add to back room ezpz, but that's up to you all.

Ketrai:
S far as I know supply crates do not contain UE

Very-Soft:
I stand corrected!

Arokha:
We can add it to a supply crate or something like that. Would rather it not just spawn on the map in abundance

Screemonster:
Problem with reducing stalls to one tile is.. can you step onto the same tile as a toilet? They won't be able to close doors behind. Widening the corridor was good though. Maybe just shove the bottom row down a tile?

Arokha:
> can you step onto the same tile as a toilet

Yes

PR #3568, opened 2018-04-26
Slightly softer corrupt pups.

Verkister:
-Halved pounce chance and reduced mauling damage scaling by 5.
-Slightly reduced chatspam chance.
-Buffed the pounce chance for the extra special double corrupt prettybois though. Also gave their punts more flavor.


Leshana:
Err what? It looks like this was done as an edit instead of a new type

Issue #3567, opened 2018-04-26
Underdark (or maybe one of its POIs) has a station-connected APC


Heroman3003:
#### Brief description of the issue
It shows up in power alarms after a while. Extremely annoying and red flashing light becomes a permanent thing

#### What you expected to happen
Even if apc is here, it shouldn't show up on station alarm consoles.

#### What actually happened
http://prntscr.com/jabc3b

#### Steps to reproduce
- No idea if its because of POI or just of Underdark itself, but simply wait a bit for that APC to lose power and show up.

#### Code Revision
- 9bc62de55ac5093696fc034ea88c102c31565ee1

Screemonster:
It's there by the stairs. Yell at the miners to start the generator. (or suit up and do it yourself if you fancy a walk, it doesn't need access. In fact, an engineer can unlock the APC so miners can swap the cell)

![image](https://user-images.githubusercontent.com/14942479/39320635-9cfb9ec8-497c-11e8-8a41-4292ca23b373.png)

PR #3573, opened 2018-04-26
Removed underdark apc from alarm consoles

Heroman3003:
Because screw that eternal source of flashing red.


Screemonster:
Fixes #3567

PR #3576, opened 2018-04-27
Removes carp burgers from burger crates, replaces with fries

Screemonster:
The fast food crate we added for the specific purpose of allowing people to eat something other than pizzas or vendor food when the service department is unstaffed is now filled only with things that are useful for that specific purpose.


Ketrai:
Thanks for this, I was just wondering when medical powergamers would start ordering in fast food crates so they can make rezadone.

Very-Soft:
Would you like to supersize your order?

PR #3601, opened 2018-04-29
Seed submaps first in underdark

Arokha:
Prevents unmineable 'floor ores'


KasparoVy:
I was wondering what those were about, bahaha. Nice.

Little-119:
floores?

KasparoVy:
Like floor cluwnes but a lot less interesting.

PR #3603, opened 2018-04-29
Adds Micro mechs

kcin2001:
adds in construction paths for the 3 current micro mechs along with the parts
also adds the ability to build the equipment for micro mechs (req tech and rnd levels can change upon suggestion)
adds in the crates for the relevant circuits along with adding them to the circuit printer
the gopher (mining micro mech) has the same building steps as the ripley and the 2 security micro mechs (polecat and weasel) are the same as the durand or gygax with the polecat using a special armor plate

if you want me to change any of the material or research values just ask


code/datums/supplypacks/robotics_vr.dm
```diff
@@ -0,0 +1,34 @@
+/datum/supply_packs/robotics/mecha_gopher
```
SpadesNeil: Do we even have these for normal mechs? If not, we should not have them here either. I am very certain these circuit crates were removed.

kcin2001: image
the non combat mechs have their boards here so should i remove the two combat mech crates?

SpadesNeil: Nope, guess we do have those after all.

SpadesNeil: Disregard my request to remove things.

Screemonster:
Looks good. Bear in mind that I don't play robotics myself so I have no idea whether those material numbers are reasonable, but everything seems like it'll function. Worst comes to the worst we can tweak the numbers later.

code/datums/supplypacks/robotics_vr.dm
kcin2001: well it would be reasonable to request the removal of the combat mech boards, would you like that?

Wickedtemp:
So... What do these mechs do, are they only locked for micro pilots, and can micros still pilot standard mechs? Are they just reskins of mechs we already have, being identical stats wise, or are the stats different?

Screemonster:
Micromechs are only pilotable by micros. Larger people won't fit in them.
Micros can still pilot standard mechs, presumably they have booster seats or something.
Micromechs are generally not as resiliant as full mechs, but are faster and less draining on the battery, and can only take micromech modules. Micromech weapons are statted to be on a par with their hand-held equivalents - the laser gun is just a laser carbine, for instance.
Unlike normal mechs, they're not anchored and you can push them around like borgs and other crewmembers.

Generally they exist as a means for a micro to go on expeditions, but they're not really any more capable than just a standard crewmember besides the inherent advantages of being in a mech, such as the pilot being unharmed when the mech takes damage. I _guess_ you could use a weasel as a fast scout on expeditions or something considering it's fuckin' turbofast and can simply outrun a deathclaw if it meets one.

Wickedtemp:
That's neat. I like this. We need more mech stuff. There's next to zero reason to order an Ody mech now, and it bugs me.

Issue #3569, opened 2018-04-26
Underdark mob balance


Screemonster:
#### Brief description of the issue
People are talking on discord about this but apparently can't be bothered raising an issue where the devs/mappers will actually see it.
Short version: a lot of our mobs that were created for away maps were made with the philosophy of being punishingly dangerous to solo crewmembers by way of having a lot of stuns, but only capable of keeping one person at a time down so a competent team can kill the mob while it's busy trying to eat the person it's downed, and thus cover each other - the reason for this is to encourage explorers and away teams to stick together and not go rambo, because they're a team.

Miners on the other hand are _not_ an adventure team. Teaming up is optional, rather than expected and required before they even step out the airlock. Therefore, the mobs they encounter and have to face need to be balanced accordingly. Challenging, but not the level of "frankly unfair" that away mission mobs can get away with.

We're planning to add/code new mobs for this zone anyway which will presumably replace the existing ones. Use this issue to discuss what's appropriate.

#### What you expected to happen
Mobs in the underdark to be risky and pose a threat, but not on the scale of the "if someone wanders off solo they will be chainstunned and murdered".

They should totally still get eaten though and have to struggle their way out of bellies.

#### What actually happened
Corrupt hounds appear to be on the list, and I'm informed that's exactly what they do.

#### Code Revision
Server revision: release - 
9bc62de55ac5093696fc034ea88c102c31565ee1

KotetsuRedwood:
Have you considered re-adding Phoron Oties as an underdark mob, balanced for miners instead of explorers? Don't think their sprites have been used much since the wilderness removal...

Verkister:
Thinking of making a cave specific variant too. Also eye glow system for the mobs in darkness?
Brainstorming some fresh mob types too to sprite up so I'm all ears for ideas.

Arokha:
One issue encountered yesterday was just that a miner didn't realize the KA did damage to mobs, as they have not played anywhere else, and got a corrupt hound down to 12hp (with like 3-4 shots, mind) before giving up.

Keep in mind the KA is one of the most powerful things against mobs. 38 damage per shot by default, infinite ammo. That will kill most mobs in 2-3 shots. Throw some damage upgrades in there (yes, you can install multiple, they stack), an efficiency upgrade, or even the AOE damage upgrade too, and you have a formidable mob annihilator.

Screemonster:
And before anyone asks, they don't work on away maps. Explorers get your own guns reeeeeeeee

[...]

PR #3613, opened 2018-04-30
Tweaks mining shed, adds UD relay

### LINKED ISSUE #3572: Mining headsets not going into shortwave mode some shifts
For some odd reason, the mining headsets like to sometimes not work at all during rounds. No clue what causes the bug at all besides the fact it's a round long bug when it does occur, so it must be linked to something during initialization, maybe.

The headsets would work as intended and switch to shortwave when you leave an area with comms.

#### What actually happened
Just pure silence and no shortwave notification at all, sometimes.

#### Steps to reproduce
- None, it just...happens. Needs testing.

#### Code Revision
- release - 
c8a25ab7dd4f89e4af6d7fb736c4e5bf8c211dbc

Arokha:
Fixes #3572 by adding a relay to the Underdark in the mining shed. Also adds a bit of phoron for the generators, and makes it stop raining in the shed.

Also a little #3569 in there

### DIFF
--- a/maps/tether/tether_turfs.dm
+++ b/maps/tether/tether_turfs.dm
@@ -4,22 +4,21 @@ VIRGO3B_TURF_CREATE(/turf/simulated/open)
 	edge_blending_priority = 0.5 //Turfs which also have e_b_p and higher than this will plop decorative edges onto this turf
 /turf/simulated/open/virgo3b/New()
 	..()
-	outdoor_turfs.Add(src)
+	if(outdoors)
+		outdoor_turfs.Add(src)
 
 VIRGO3B_TURF_CREATE(/turf/simulated/floor)
-/turf/simulated/floor/virgo3b/New()
-	..()
-	outdoor_turfs.Add(src)
 
 /turf/simulated/floor/virgo3b_indoors
 	VIRGO3B_SET_ATMOS
 /turf/simulated/floor/virgo3b_indoors/update_graphic(list/graphic_add = null, list/graphic_remove = null)
 	return 0
 
+VIRGO3B_TURF_CREATE(/turf/simulated/floor/reinforced)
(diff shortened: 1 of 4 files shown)


maps/tether/tether_turfs.dm
```diff
@@ -4,12 +4,14 @@ VIRGO3B_TURF_CREATE(/turf/simulated/open)
...
 	..()
-	outdoor_turfs.Add(src)
+	if(outdoors)
+		outdoor_turfs.Add(src)
 
 VIRGO3B_TURF_CREATE(/turf/simulated/floor)
 /turf/simulated/floor/virgo3b/New()
 	..()
-	outdoor_turfs.Add(src)
+	if(outdoors)
```
Leshana: I think this is redundant with `/turf/simulated/floor/New()` and will double-add.

maps/tether/tether_turfs.dm
```diff
@@ -36,7 +38,8 @@ VIRGO3B_TURF_CREATE(/turf/simulated/floor/reinforced)
 VIRGO3B_TURF_CREATE(/turf/simulated/floor/tiled/steel_dirty)
 /turf/simulated/floor/tiled/steel_dirty/virgo3b/New()
 	..()
-	outdoor_turfs.Add(src)
+	if(outdoors)
```
Leshana: This one is for sure redundant with above New()

maps/tether/tether_turfs.dm
Arokha: You are correct. I had no idea that was even a thing. I guess I independently came up with that idea.

maps/tether/tether_turfs.dm
Arokha: Though even if I did know, I shouldn't have done this, true.

PR #3617, opened 2018-04-30
Allow maint drones to 'cryo' using the robot pod things

Arokha:
Dunno why they were excluded. If it's a horrible issue I guess we'll find out why.


Belsima:
this'd be cool and good to have on polaris, no?

PR #3621, opened 2018-05-01
Redo cryo item storage

### LINKED ISSUE #2610: Suggestion: "fluff" var on fluff items
#### Brief description of the suggestion
Add a "fluff" var to items that is 0 by default, and set to 1 on fluff items, so the cryo computers can check for said items and delete them when the owner leaves regardless of whether they'd otherwise be retained (such as being guns or whatever)

#### What happens now
People cryo/teleport/tram, and things are added into the cryo computer for retrieval depending on the item type, even if they're items we don't want people to be able to retrieve.

#### What you want to happen
Same as now, but qdel the item if fluff = 1 so people can't go around swiping the various custom items that people currently have to ahelp on leaving to ask for them to be deleted.

### LINKED ISSUE #2682: Colony Director's ID held by oversight console
When a user playing colony director exits the round via storage, their ID will be kept in the tram/teleport/cryogenic oversight console instead of being deleted properly.

The ID is deleted, the same as any other ID.

#### What actually happened
The ID is retained and able to be collected by other players

#### Steps to reproduce
- Join as Colony Director
- Make sure your ID is NOT in a PDA.
- Leave the round via tram, cryo, teleport.

#### Code Revision
a5a83f7ed84f914ac381c671e78ae764e6d09e2e 
 
Arokha:
Fixes #2682, Fixes #2610

Cryo no longer stores any items, only logs for admin use and for player use who cryo'd with what 'important' items.

### DIFF
--- a/code/game/machinery/cryopod.dm
+++ b/code/game/machinery/cryopod.dm
@@ -94,8 +94,8 @@
 	dat += "<a href='?src=\ref[src];log=1'>View storage log</a>.<br>"
 	if(allow_items)
 		dat += "<a href='?src=\ref[src];view=1'>View objects</a>.<br>"
-		dat += "<a href='?src=\ref[src];item=1'>Recover object</a>.<br>"
-		dat += "<a href='?src=\ref[src];allitems=1'>Recover all objects</a>.<br>"
+		//dat += "<a href='?src=\ref[src];item=1'>Recover object</a>.<br>" //VOREStation Removal - Just log them.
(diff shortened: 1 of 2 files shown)


Arokha:
Correct.

Little-119:
Pretty sure that most of the time it didn't allow recovery of things you'd actually want to recover (Like job vital and/or limited in quantity equipment) and allowed you to recover useless things (Like everyone's jackets and labcoats)

Screemonster:
Oh, and it also allowed assistants to grab firearms when sec officers cryo'd with their gear _again_.

PR #3622, opened 2018-05-01
Adds more maintenance to z6 and z7

Very-Soft:
Here we have yet more dark and spooky maintenance tunnels!

This time they are near the IAA's office and the AI core.

z6

This is the first entrance. The idea being that there may have been more space for another law figure, or perhaps a place for shady deals to have been done. Either way, there is a small abandoned office, and a ladder to go up to the AI maints. There are also two places to climb up to the floor abive apart from the ladder, one in the office is unmarked. The other is marked by the wire and pipes going up.

z7a

This is the south section of the AI maintenence. Here you can see that this is more of an attic than anything. There's crates and some things stored around, but it's clear that this place isn't used for much else presently. At the bottom left, behind a railing and a random barrier, there is a mimic, containing a gold coin, a silver coin, and a skull. There is also an unfinished airlock in this area. It doesn't have the parts to actually work, and is sealed off from space and the station. It could be a good place to have a secret base, or could be used by criminals to sneak onto the station. 

z7b

And here the tunnel continues. There is a nice window with a couple of seats, another room with some supplies, and a dark and spooky cave with some trash in it. There is a secret entrance into this section by way of a false wall at the end of the previously existing stretch of maintenance, where the rock is dug through in front of the railing.

For your consideration!

### DIFF
--- a/maps/tether/tether-06-station2.dmm
+++ b/maps/tether/tether-06-station2.dmm
@@ -7690,40 +7690,23 @@
 /obj/effect/floor_decal/steeldecal/steel_decals7,
 /turf/simulated/floor/tiled,
 /area/hallway/station/starboard)
-"nj" = (
-/obj/structure/disposalpipe/segment{
+"nk" = (
+/obj/effect/floor_decal/borderfloor{
+	dir = 1
+	},
+/obj/effect/floor_decal/corner/lightgrey/border{
+	dir = 1
+	},
+/obj/effect/floor_decal/steeldecal/steel_decals7{
 	dir = 4
 	},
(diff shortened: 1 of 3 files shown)


Draycu:
Love the look of these additions, it'd be fun using the tunnels more on the station side for a change!

Mewchild:
I’m personally not a huge fan of the mimic being so easily accessible. My only note

Very-Soft:
The mimic is indeed a little more easy to access than the other one. It will ordinarily be blocked by a random thing, but I can change it to a wall or a rock if it's too much of an issue. The mimic itself is still until someone goes over to it, so it shouldn't wander out on its own.

PR #3623, opened 2018-05-01
Adds selectable sets of contamination flavors for bellies.

Verkister:
-Divides item contamination flavors into categories to match better with the "belly" types.
-The selection menu shows up when itemweak mode is selected.
-The selection should save for each belly along with other prefs.
-The flavors have been cropped into following categories:
->Generic, contains most of the legacy collection for less specific gurgling. Is the designated default option.
->Acrid, contains flavors implying regular acid digestion. (for stomach bellies and such)
->Dirty, contains flavors related to visual contamination. (for gurgled stuff without excess acidity and aromas)
->Musky, contains suggestively sloppy flavors without excess filth. (for use with lewd organs without any gastrointestinal implications)
->Smelly, contains more airborne flavors without visual slop or crud. (for less visual contamination)
->Wet, contains flavors of more neutral-ish soakage. (for use with mouths and other more neutral areas now that the non-contaminating hold mode is gone)

-Also fixed belly transfer not removing the transfered items from its items_preserved list while at it (a spammy source of GC errors)


Chargae:
Yessssss.

Shophaune:
You would love it Chargae. I also approve of this change though!

code/_helpers/global_lists_vr.dm
```diff
@@ -161,6 +161,177 @@ var/global/list/edible_trash = list(/obj/item/trash,
...
 
+var/global/list/cont_flavors = list(
+				"All" = cont_flavors_all,
+				"Acrid" = cont_flavors_acrid,
+				"Dirty" = cont_flavors_dirty,
+				"Musky" = cont_flavors_musky,
+				"Smelly" = cont_flavors_smelly,
+				"Wet" = cont_flavors_wet)
+
+var/global/list/cont_flavors_all = list("soggy",
```
Arokha: Generate this by concatenating all the lists together. Look down further in this file for all the places where lists are dynamically generated. Leave the list defined up here but an empty list, then populate it down there.

Opinion: Use the http://www.byond.com/docs/ref/info.html#/operator/|= to ensure that there are no duplicates. Opinion2: You'd better not hardcode the names of lists, use a for() over cont_flavors.

code/modules/vore/eating/contaminate_vr.dm
```diff
@@ -5,20 +5,33 @@ var/image/gurgled_overlay = image('icons/effects/sludgeoverlay_vr.dmi')
...
 		return FALSE
 
 	if(!gurgled)
 		gurgled = TRUE
 		overlays += gurgled_overlay
-		var/gurgleflavor = pick("soggy","soaked","dirty","nasty","slimy","drenched","sloppy","grimy","sludgy","stinky","mucky","stained","soiled","filthy","saucy","foul","icky","tarnished","unsanitary","messy","begrimed","cruddy","funky","disgusting","repulsive","noxious","gruesome","gross","putrid","yucky","tainted","putrescent","unsavory","smelly","smutty","acrid","pungent","unclean","contaminated","gunky","gooey","sticky","drippy","oozing","sloshed","digested","sopping","damp","gloppy","begraggled","churned")
+		switch(cont_flavor)
+			if("All")
+				cont_flavor = cont_flavors_all
+			if("Acrid")
```
Arokha: This switch() is annoying. Instead just do
```
var/list/pickfrom = cont_flavors[cont_flavor]
var/gurgleflavor = pick(pickfrom)
```

code/_helpers/global_lists_vr.dm
Verkister: I probably should've named the selection as general instead of all now that I think about it. Otherwise I'll need to exclude possible mode specific words in other lists from other collections (currently only "musky" but that's only because I didn't search for possible special additions for other lists yet)
How would I do the soft naming for the lists?

Arokha: If modes are made up of a few mode specific words 'then a bunch of other shared ones', then you should just have the modes be those specific words, and pick from a combination of 'general' and 'mode' instead of duplicating the shared words into each list.

Verkister: I suppose I could get rid of the generic selection and make the acrid one the default.

PR #3624, opened 2018-05-02
Implement rust-g library for logging

Leshana:
Based on tgstation/tgstation#36858 this delegates logging to our fork of the rust-g library: https://github.com/VOREStation/rust-g

See the /tg for details and benchmarks for the performance improvements. 
Our fork of the rust-g library uses local time instead of UTC, and uses the date format our logs were previous in, so it is compatible with our existing splunk setup.

HOWEVER! This includes both the compiled dll for windows, and the compiled shared library for linux.   That means this will work out of the box on most all normal linux and windows machines with x86 processors.

For downstream repositories, this should be a drop in replacement.


code/__defines/_compile_options.dm
```diff
@@ -7,6 +7,10 @@
 								2 for preloading absolutely everything;
 								*/
 
+#define RUST_G "rust_g"			// If uncommented, we will use the rust-g (https://github.com/tgstation/rust-g) native library for fast
+								// logging. This requires you to have the rust_g.dll or librust_g.so installed in the root folder or BYOND/bin
```
Arokha: `librust_g.so` inaccurate

PR #3630, opened 2018-05-03
Fix alignment for grab insertion cryo placement

Arokha:
Also what it says!


Leshana:
Alignment?  This change seems  reasonable, but what do you mean by alignment?  Also, is this a polaris bug?

Arokha:
Yes, I noticed that, but I think it's fine.

PR #3641, opened 2018-05-06
Adds Goat Legs, for satyrs and stuff.

Entosine:
Also fixes a misnomer in the Ears and Tails DM file, there aren't any taur bodies in there anymore.


SilverTalismen:
Screenshots?

Entosine:
satyr ingame
@SilverTalismen 
Had to reopen the test server, sorry about the wait.   The front and back look kinda bad, but that's just because of how tails and stuff generally have to be sprited over the legs.   I'm really proud of the sides though.

KasparoVy:
666

Arokha:
This won't clip through pants 'n such...? Or is it actually just thinner at the top? Or...

Entosine:
It will, it's a tail.   There's literally no other way to do this and anyone with literal goat legs probably isn't wearing pants to begin with.

Entosine:
Is there an issue with this that I'm not seeing?

Arokha:
No, just discussing internally if this is how we want this implemented or if there might be a better way than a tail. Nothing wrong with the PR itself.

Entosine:
Oh.  Okay.   I mean, I picked a tail because that's how other taurhalves were implemented before they were their own thing, but this isn't a taurhalf so it shouldn't have all of the weird features you're adding for them.   I'd love for there to be a way to do this and have clothing work but then I'd also have to go back and fnagle with the sprite in ways I don't want to do.   It looks far better like this than constraining itself to where default legs go.

Entosine:
More or less I don't want this to be delayed for another half a week or more because of indecision, if a better system comes along it'd be simple to just kinda move the sprite.  99% sure there actually isn't another way to implement something like this.  Fully modular limbs would be cool as heck, and a system like that is best case, but it's probably a long way off.

Screemonster:
I sure as hell don't intend to port tg's lizard legs system. They only recently got any kind of dismemberment system in place and it's not going to even remotely resemble the bay/polaris medical system.

Entosine:
tl;dr I kinda have a character waiting on this and I'm anxious and impatient.

PR #3655, opened 2018-05-08
Adds a client FPS setting

Arokha:
Title!


Leshana:
Will this happen to Polaris also?

Arokha:
I made the same PR on Polaris, but a downstream wanted it so made it here at the same time

PR #3656, opened 2018-05-08
Removed shadekin from Slime Vore reaction

Heroman3003:
No more non-player shadekin randomly appearing.


Heroman3003:
I did a while ago, summoned shadeking through that reaction 3 different times, just never got my hands to making issue or fixing it because "eh, its one line change, gonna do it later"

Screemonster:
fun fact this is how the orange-eyes I play showed up. I fluffed it as the slime reaction getting their attention, they showed up, and now they're here to fuck with y'all.

PR #3667, opened 2018-05-10
Fixes a bunch of NIF/resleeving related issues (and a tiny otherthing i discovered during tests)

Verkister:
-Fixes resleeving forcing reinstall wear for each backed up soft-
-Seems like an oversight that's been bugging people who never brought it up before as resleeving is supposed to restore the backed up nif with the softs included, not wear it out reinstalling everything on empty secondhand nif.
-Basically just restores the backed up nif durability after restoring the nifsofts upon resleeving.

Added stuff:
-Applies the suggested durability tweak.
-Makes backups save nif examine message.
-Makes new nifs spawn with the examine message if provided upon resleeving and such. Basically the initial character spawn pref loading was the only case that wasn't resetting the message to default.
-Fixes soulcatcher backups breaking ooc notes. (Applies ooc notes from character to the brainmob before saving it. Also includes brainmobs into the formerly human exclusive backup part responsible of saving nif and ooc notes to backups.)
~~-Makes soulcatcher backup feature back up the owner if dying inside a belly of a soulcatcher lacking pred. (SC backup setting was completely voided if the pred didn't have SC themselves so your shit would be gone for good)~~

The otherthing:
-Fixes contaminate/digest_act sending holder mobs to shadow realm if holder is contained inside the equipment(storage/pockets) of a prey getting gurgled instead of the pred's gut. (As prey gear gets all contaminated at once while getting dumped in the gut, which would contaminate>digest_act the holder items prematurely during the dump and causing issues.)

### DIFF
--- a/code/modules/nifsoft/software/13_soulcatcher.dm
+++ b/code/modules/nifsoft/software/13_soulcatcher.dm
@@ -230,6 +230,7 @@
 		if(ishuman(M))
 			var/mob/living/carbon/human/H = M
 			brainmob.dna = H.dna
+			brainmob.ooc_notes = H.ooc_notes
 			brainmob.timeofhostdeath = H.timeofdeath
 			SStranscore.m_backup(brainmob.mind,0) //It does ONE, so medical will hear about it.
 
@@ -461,6 +462,8 @@
 		if(istype(HP) && HP.nif && HP.nif.flag_check(NIF_O_SCOTHERS,NIF_FLAGS_OTHER))
 			var/datum/nifsoft/soulcatcher/SC = HP.nif.imp_check(NIF_SOULCATCHER)
 			SC.catch_mob(H)
+		else if(H.nif && H.nif.flag_check(NIF_SC_BACKUPS))
+			SStranscore.m_backup(H,0) //If dying inside a pred who lacks SC, let ours do a backup if enabled before it's gone for good.
 	else if(H.nif && H.nif.flag_check(NIF_O_SCMYSELF,NIF_FLAGS_OTHER)) //They are caught in their own NIF
 		var/datum/nifsoft/soulcatcher/SC = H.nif.imp_check(NIF_SOULCATCHER)
 		SC.catch_mob(H)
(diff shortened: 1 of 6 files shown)


code/modules/resleeving/machines.dm
```diff
@@ -529,6 +529,7 @@
 		var/obj/item/device/nif/nif = new MR.nif_path(occupant,MR.nif_durability)
 		for(var/path in MR.nif_software)
 			new path(nif)
+		nif.durability = MR.nif_durability //Restore backed up durability as the line above forces reinstall wear for each soft.
```
Arokha: You may want to remove the MR.nif_durability on 529 (test to make sure that works though, maybe it's used as an indicator we're being resleeved and is somehow important, I don't remember)

The reason: What if someone has 5 durability on their NIF in the backup?
- Create NIF with 5 durability
- Install seve-- oh no we broke it, installs may have failed
- Does setting durability back up even work now? Who knows.

Verkister: Good catch. Seems the durability bit for the new nif isn't used for anything other than setting the durability on spawn so it shouldn't be breaking anything

code/modules/nifsoft/software/13_soulcatcher.dm
```diff
@@ -461,6 +462,8 @@
 		if(istype(HP) && HP.nif && HP.nif.flag_check(NIF_O_SCOTHERS,NIF_FLAGS_OTHER))
 			var/datum/nifsoft/soulcatcher/SC = HP.nif.imp_check(NIF_SOULCATCHER)
 			SC.catch_mob(H)
+		else if(H.nif && H.nif.flag_check(NIF_SC_BACKUPS))
+			SStranscore.m_backup(H,0) //If dying inside a pred who lacks SC, let ours do a backup if enabled before it's gone for good.
```
Arokha: While this is obvious and logical for a thing to exist and I'm totally down for it, I sort of wonder if people are doing to use this as a thing to point to as to why they have no CMD. Probably warrants admin discussion.

code/modules/resleeving/machines.dm
```diff
@@ -526,9 +526,10 @@
 
 	//Re-supply a NIF if one was backed up with them.
 	if(MR.nif_path)
-		var/obj/item/device/nif/nif = new MR.nif_path(occupant,MR.nif_durability)
+		var/obj/item/device/nif/nif = new MR.nif_path(occupant,,MR.nif_savedata)
```
Arokha: Just use named parameters instead of this spooky double-comma that people might miss.

PR #3670, opened 2018-05-11
Makes gurgle gains scale according to actual received damage.

Verkister:
-Fixes armored, synthetic, and otherwise resistant prey giving several times more nutrition than their squishy counterparts.
-They will still take longer to digest, but what would ya expect for armor, at least the gains are at the correct level now.
-Same for sleeperbellies.
-Also made sleeperbelly gurgledamages non-hardcoded vars but there's no knob for them yet.

### DIFF
--- a/code/modules/vore/eating/bellymodes_vr.dm
+++ b/code/modules/vore/eating/bellymodes_vr.dm
@@ -126,18 +126,23 @@
 				continue
 
 			// Deal digestion damage (and feed the pred)
+			var/old_brute = M.getBruteLoss()
+			var/old_burn = M.getFireLoss()
 			M.adjustBruteLoss(digest_brute)
 			M.adjustFireLoss(digest_burn)
+			var/actual_brute = M.getBruteLoss() - old_brute
+			var/actual_burn = M.getFireLoss() - old_burn
+			var/damage_gain = actual_brute + actual_burn
 
 			var/offset = (1 + ((M.weight - 137) / 137)) // 130 pounds = .95 140 pounds = 1.02
 			var/difference = owner.size_multiplier / M.size_multiplier
 			if(isrobot(owner))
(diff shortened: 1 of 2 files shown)


code/modules/vore/eating/bellymodes_vr.dm
```diff
@@ -126,18 +126,23 @@
 				continue
 
 			// Deal digestion damage (and feed the pred)
+			var/old_brute = M.getBruteLoss()
+			var/old_burn = M.getFireLoss()
 			M.adjustBruteLoss(digest_brute)
 			M.adjustFireLoss(digest_burn)
+			var/actual_brute = M.getBruteLoss() - old_brute
+			var/actual_burn = M.getFireLoss() - old_burn
```
Arokha: This won't work right for digesting synths, because it will return abnormally low numbers (or maybe just 0) as their limbs are robotic and potentially non-vital. See:

https://github.com/VOREStation/VOREStation/blob/7ad1a817de7a578efd70a140359cbc217c39dc70/code/modules/mob/living/carbon/human/human_damage.dm#L67

Try using getActualBruteLoss() and getActualFireLoss()

Verkister: I tried but that wouldn't fix the synth issue because the actual losses pull damage/nutrition from all limbs, while on synths only head/torso count for overall damage and it would keep yanking damage/nutrition from the whole body until the point where the vital parts reach the death treshold by themselves, the actual damage in this case being around 700 while the overall damage only would show as 200.

Arokha: So does this result in abnormally low nutrition gains for digesting synths? Or no?

SpadesNeil: Synths aren't very nutritious anyway. They're made of synthetic things. :B

PR #3676, opened 2018-05-12
Slightly increases size difference for taur rides

Verkister:
-Makes the size difference margin required to fail taur riding +10% of the taur's size.
-Opie couldn't ride Kenzie because he was like an inch or two taller which totally was classifying Kenzie as a pony show >:v

### DIFF
--- a/code/modules/vore/appearance/sprite_accessories_taur_vr.dm
+++ b/code/modules/vore/appearance/sprite_accessories_taur_vr.dm
@@ -57,7 +57,7 @@
 		return FALSE
 	if(!ishuman(M))
 		return FALSE
-	if(M.size_multiplier > size_multiplier)
+	if(M.size_multiplier > size_multiplier*1.1)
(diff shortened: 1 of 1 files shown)


Ketrai:
but verkie, you need to powergame the taur riding. Climb on a taur, then size gun yourself 👯‍♂️

SpadesNeil:
>but verkie, you need to powergame the taur riding. Climb on a taur, then size gun yourself

sounds like an exploit but ok

Ketrai:
Well spades, the taur riding is very loosely implemented. I'm actually fine with it getting disabled for the time being, so aronai can properly fix it. Since it causes the dragging bug, stuns don't work properly on them, damage and mob targeting might not work right... it's honestly barely been testing for what I see as a rather important feature.

Verkister:
should totally make it work on dogborgs too :v

SpadesNeil:
We should probably fix the other issues before we go mucking with other things. Closing this for now.

PR #3685, opened 2018-05-14
Fix a typo in vore load code

### LINKED ISSUE #3684: Dropnom / televore settings not saving between rounds
In Vore Panel, setting Dropnom Prey and Dropnom Pred will save the settings for that round and they will work as expected, but fresh rounds has them always default to "off". This is frustrating as dropnom / televore stuff is, by its very nature, opportunistic - when the stars align and dropnom lines up or someone picks up a beacon, only to find that you hadn't set it up that round shuts down whatever could have been fun. It works for digestible, mob vore, and so on, so I'm not certain why this doesn't work.

Saving your vore preferences to actually save all of them.

#### What actually happened
Dropnom prey and dropnom pred reset back to "no" at the start of the next round.

#### Steps to reproduce
- Open Vore panel, set dropnom preferences to on.
- Save your preferences.
- Enjoy dropnom while it works.
- Turn up to a new round, check vore panel preferences, dropnom is off.

#### Code Revision
- 849c1374a1b064b9c9a1fb8c1c43ed19a4259100 
 
- Yes, I did save preferences after setting them. No, I didn't switch characters with different preferences.

Arokha:
Fixes #3684

### DIFF
--- a/code/modules/vore/eating/vore_vr.dm
+++ b/code/modules/vore/eating/vore_vr.dm
@@ -111,7 +111,7 @@ V::::::V           V::::::VO:::::::OOO:::::::ORR:::::R     R:::::REE::::::EEEEEE
 	allowmobvore = json_from_file["allowmobvore"]
 	vore_taste = json_from_file["vore_taste"]
 	can_be_drop_prey = json_from_file["can_be_drop_prey"]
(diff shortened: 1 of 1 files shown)


Screemonster:
Oh so THAT'S why I never encountered this bug, I had both set to true.

PR #3702, opened 2018-05-16
Makes lighting fancier, reduces the amount of light sources with powers greater than 1

deathride58:
Title. The new values are based off of my own studies of lights in the real world. This should make lighting look a lot more realistic, instead of lighting being pure white against pure black in most areas.
This PR also reduces quite a few light power values. With robustlight2/goonlights/vglights, the light_power variable is essentially the alpha of a light source, so having it exceed 1 can result in some pretty binary light behavior. Reducing the light_power variable will result in the station being darker across the board, but will make most light sources behave in a slightly more realistic manner.
This also makes some minor tweaks to how flashlights handle their light values. This allows lights to actually shine a little brighter when their brightness is set to high, and makes flashlights glow darker when their brightness is set to low.

I've also made this PR upstream, though this PR here has two lines worth of minor adjustments to make the emitted color of the station lights match their sprites.

Before
image

After
image

### DIFF
--- a/code/__defines/lighting.dm
+++ b/code/__defines/lighting.dm
@@ -74,5 +74,12 @@
 #define LIGHT_COLOR_TUNGSTEN   "#FAE1AF" //Extremely diluted yellow, close to skin color (for some reason). rgb(250, 225, 175)
 #define LIGHT_COLOR_HALOGEN    "#F0FAFA" //Barely visible cyan-ish hue, as the doctor prescribed. rgb(240, 250, 250)
 
+//Lighting values used by the station lights
+#define LIGHT_COLOR_FLUORESCENT_TUBE "#E0EFFF"
+#define LIGHT_COLOR_FLUORESCENT_FLASHLIGHT "#CDDDFF"
+#define LIGHT_COLOR_INCANDESCENT_TUBE "#FFEEDD"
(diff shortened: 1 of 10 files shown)


Heroman3003:
Ugh, illuminating dark corners is just harder now... As if small light fixtures didnt give awful light already...

Heroman3003:
Along with that a question: does it include a fix for a bug where roundstart small fixtures are much brighter than intended?

KasparoVy:
Bless

Leshana:
I like the idea, but this will need re-mapping to avoid the station becoming dark.  Most hallway places are *supposed* to be bright.  But not over 100% bright, sure.

Very-Soft:
I like this. I've been changing the lights in maint to be more atmospheric sometimes very similar to this, so of course I very much like it~ C:

deathride58:
@Leshana There's a couple of blemishes here and there in hallways and other large areas, but the lights are actually mostly fine as-is! Here's a couple of screenshots of how large chunks of the map look with the new lighting values
<details><summary>Image dump</summary>

image

image

image

image

image

image

</details>

Cameron653:
The lighting looks absolutely lovely. Can certainly see this being a good addition.

Arokha:
Merged upstream, will make appropriate changes in the sync PR

PR #3709, opened 2018-05-16
Message send failure reflection

### LINKED ISSUE #3046: No warning for character limit in Say/Me

When typing out particularly long posts in scene, there's no warning that you've reached character limit, nor any way to retrieve whatever you wrote past this invisible limit either, which I feel is too short for an RP server anyway.

The server would warn me when I went over that amount, or else spit it back out in a user-only message

#### What actually happened

The say message abruptly ends mid-word and doesn't allow you to easily copy-paste the remainder to your partner

#### Steps to reproduce
- Be verbose in a say/me
- Like really verbose. Really describe a lot.
- Hit enter
- Be disappointed.

#### Code Revision
- 67c7641d36ec5c12305d218aa4730c684703a1c7
 
- I get this way too often. I feel like the limit should be a bit higher instead of just-... Not. If not a limit change, then it'd be nice to have the message copy-pastable somehow again so I could break it up into byond-size chunks

Arokha:
When a post is too long to send, reflect it back to the user in the chatbox so they can copypaste it into separate messages. Also increases the message limit from 1024 to 2048. If a message is longer than 5-messages worth, it is not reflected and a warning is sent instead to avoid easily consuming bandwidth for DDOS purposes with long messages.

I'd like to add markers in the message where the posts would have been truncated but that's... complicated.

Fixes #3046

### DIFF
--- a/code/__defines/misc.dm
+++ b/code/__defines/misc.dm
@@ -104,7 +104,7 @@
 #define WAIT_FINISH  4
 
 // Setting this much higher than 1024 could allow spammers to DOS the server easily.
-#define MAX_MESSAGE_LEN       1024
+#define MAX_MESSAGE_LEN       2048 //VOREStation Edit - I'm not sure about "easily". It can be a little longer.
 #define MAX_PAPER_MESSAGE_LEN 6144
 #define MAX_BOOK_MESSAGE_LEN  24576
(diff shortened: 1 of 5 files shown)


SpadesNeil:
dear god, but we needed this

PR #3713, opened 2018-05-17
Makes KAI spawn with their item properly

Cameron653:


config/custom_items.txt
```diff
@@ -329,7 +329,7 @@ item_path:/obj/item/clothing/accessory/badge/holo/detective/ruda
 {
 ckey: interrolouis
 character_name: Kai Highlands
-/obj/item/borg/upgrade/modkit/chassis_mod/kai
+item_path:/obj/item/borg/upgrade/modkit/chassis_mod/kai
```
Arokha: Missing the space

Cameron653: I went ahead and left the space out, since their other fluff item in the above lines didn't have a space inbetween it, either. I'll go ahead and add spaces to both of them.

Cameron653:
Added spaces to the others while I was at it.

PR #3714, opened 2018-05-17
Fix various syntactic problems

SpaceManiac:
Code cleanup necessary for MapDiffBot to work. #3706

All the deleted `..()` currently do nothing.

Precedent:
https://github.com/ParadiseSS13/Paradise/pull/8742
https://github.com/yogstation13/yogstation/pull/3137
https://github.com/tgstation/tgstation/pull/32284

### DIFF
--- a/code/game/objects/items/devices/whistle.dm
+++ b/code/game/objects/items/devices/whistle.dm
@@ -27,8 +27,8 @@
 		use_message = capitalize(copytext(sanitize(new_message), 1, MAX_MESSAGE_LEN))
 
 	usr << "You configure the hailer to shout \"[use_message]\"."
-/
-obj/item/device/hailer/attack_self(mob/living/carbon/user as mob)
+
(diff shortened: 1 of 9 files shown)


Arokha:
Those vars without the space, I'm surprised DM compiles those.

SpaceManiac:
There might still be other issues with the map rendering due to it being slightly tg-specific in certain ways but this is enough to make it mostly work.

Arokha:
We appreciate the PR, either way!

Arokha:
https://github.com/VOREStation/VOREStation/pull/3706/checks

Seems like the same problem still, over there, but maybe a step closer!

SpaceManiac:
You may need to rebase that PR so that these changes are included.

Arokha:
You were right! Though it's a little anomalous. Might not take planes into account? Appears to believe everything is on the same plane but different layers.

Leshana:
Wow, these syntax problems were in there huh? Wacko

PR #3725, opened 2018-05-18
Set a fake ID on shelter buttons to avoid messing with every door

Arokha:
Hopefully this is the issue.

### DIFF
--- a/code/modules/mining/shelter_atoms.dm
+++ b/code/modules/mining/shelter_atoms.dm
@@ -103,6 +103,7 @@
 	name = "shelter privacy control"
 	desc = "You can secure yourself inside the shelter here."
 	specialfunctions = 4 // 4 is bolts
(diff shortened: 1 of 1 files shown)


SpadesNeil:
>This habit of using memes or pop culture references in code triggers me on a personal level and serves to do nothing but frustrate coders and admins who are trying to find something in the code but unfortunately aren't psychic and can't figure out what fucking path something exists under.

>I'm rejecting all future PRs that do this absurd thing with meme names.

>I used to have a coworker who would only name his variables in memes.

b87

no but seriously

Arokha:
I can change it, but I think you missed what this is doing. It has to be something that's NEVER used. Whatever the string is,  it can't be used anywhere else.

SpadesNeil:
Change it anyway.

Arokha:
I did already, nerd

Little-119:
...what happened in this PR?

Issue #3395, opened 2018-04-06
Only one solid method ingame to disinfect a wound.


ArgobargSoup:
#### Brief description of the issue
There's only one item in the code that WILL disinfect a wound for sure... the advanced trauma kit, found in advanced medical kits, which has a disinfect proc.  The salve proc from ointment and advanced burn kits only has a 50/50 chance.

#### What you expected to happen
For a burn, the most easily infected type of wound, to be treated and truly disinfected by an advanced burn kit.  For more than one way to actually truly disinfect a wound.  For a fully treated wound to not get infected long after being healed.

#### What actually happened
Several cases of treating someone badly burned with an advanced burn kit, monitoring them until they were fully healed, and letting them out of medbay after the body scanner showed them with a clean bill of health, only to have them come in later with a super nasty infection on the part thought healed, because we didn't use a rare resource generally known for treating brute damage and not burn damage.

#### Steps to reproduce
- Get a nasty burn.
- Treat this burn with an advanced burn kit.
- Watch it till you have a clean bill of health from the body scanner.
- Develop a nasty infection a while afterwards.

#### Code Revision
- Server revision: release - 
0b4a7e3a7ea4baed068ae7fcd1ce68a4632a2fc8 

#### Anything else you may wish to add:
- Possibly an upstream issue as well.
- Not as urgent as I thought due to finding the salve effect, but yeah.

Arokha:
I believe Pol plans to make the other items also disinfect wounds. Both adv kits and ointment, I think? I'll ask and come back here and put it eventually. :p

Wickedtemp:
I'd rather it be Gauze/Ointment with 50:50 chance, and advanced trauma/burn kits with the 100% certainty, since one is meant to be better than the other, but that can probably be changed later.

PR #3731, opened 2018-05-19
Nerfs the exponential infection toxloss.

Verkister:
-Divides the toxloss output of infected organs by 8 (as half the average organ count, would've gone with full but didn't wanna go that far _yet_.)

Thing is that even _the most potent venoms_ in the world still take significantly longer to be fatal (I did my research.) than _simply getting a merely inconveniencing boo boo_ on station treated with a _non-advanced_ medkit is just preposterous on every degree imaginable. The infection math in the game is currently immediately dispensing between **0.25-5.25 toxloss _per_ infected organ _every tick_** while _every single one of the involved organs begins spreading infection to another in **exponentially growing numbers**_ as soon as the "first warning" itching phase is over. The infection levels 3 and MAX are practically obsolete and completely unreachable due to the premature death caused by mere level two infection going exponentially haywire with toxloss. Going from mere itching (level 1) to fatality before you can even type two ahelps smh.

Let's not allow _meme mechanics_ be this sort of excessively robust scenestompers in HRP.
_"It's a baymed meme hur hur deal with it."_

### DIFF
--- a/code/modules/organs/organ.dm
+++ b/code/modules/organs/organ.dm
@@ -163,7 +163,7 @@ var/list/organ_cache = list()
 		infection_damage = max(0.25, 0.25 + round((germ_level - INFECTION_LEVEL_TWO)/200,0.25))
 
 	if(infection_damage)
-		owner.adjustToxLoss(infection_damage)
+		owner.adjustToxLoss(infection_damage/8) //VOREStation Edit for little more reasonable numbers. Divided by half of the average organ count to make up for the exponential spreading of the infection causing death within seconds. (Would've gone with full average but didn't wanna go that far yet.)
 
 	if (germ_level > 0 && germ_level < INFECTION_LEVEL_ONE/2 && prob(30))
 		adjust_germ_level(-antibiotics)


Screemonster:
Honestly I was just thinking about this, we have a race that are, in lore, literally walking clusters of flesh-eating microbes, and they're positively _sanitary_ compared to what a paper cut can do on this station.

code/modules/organs/organ.dm
```diff
@@ -406,4 +406,4 @@ var/list/organ_cache = list()
 		return 0
 	if(robotic && robotic < ORGAN_LIFELIKE)	//Super fancy humanlike robotics probably have sensors, or something?
 		return 0
-	return 1
\ No newline at end of file
+	return 1
```
Arokha: Don't randomly add this in Polaris files. It will conflict whenever they add anything to the file because it's now 'longer' than this change.

Arokha:
A couple of things about your comment on this:
- Yes, the most potent venoms take longer to become fatal, sometimes even longer than our rounds, which makes things like this require some sort of time compression to actually fit into the round.
- 'As soon as the first warning (itching) L1 phase is over', true, it does begin spreading in L2 infection states, but it takes **18 minutes** to go from uninfected to L1, and then **30-45 minutes** to go from L1 to the much more dangerous L2, during which you don't take any damage, and are printed nag messages to go treat your infection.

I'm curious what case you're referring to, with the "couldn't send two ahelps" because in giving myself infections on test servers, it's taken about an hour to actually suffer fatal consequences.

I'm not really sure this is the way to fix this, to be honest. Instead maybe we should just make exceptionally minor wounds not have infection chances, so that accidentally clicking yourself once won't cause infection (it shouldn't be doing that anyway, if I'm looking at the code right, since you need germ_level 50).

Wickedtemp:
Am opposed. 

If your infection worsens to the point of causing damage, well... Kinda sucks, but you ignored pain messages for like fifteen minutes. 

I know it might be annoying to have to pause a scene for two minutes to go to medbay and get a shot of Cillin, but taking less than five minutes out of the round to do so really isn't a big deal. You could even just PDA a doctor "hey, think I have a small infection, can you bring me antibiotics to dorm 2", and unless medical has some type of crisis going on, it'll be fine. 

As it stands currently, bloodloss, heart/lung damage, and infections are really the only things a patient can die from without actively being bludgeoned or lit on fire or fed to slimes. I'd want to keep that as a result of that, as well as a result of the testing that's been done. I've yet to see an infection so severe I can't stabilize. 

Infections were broken to the point they literally didn't do anything, they were fixed and are relatively balanced in that they only get worse if you ignore 10-15 minutes of pain messages, and you can be stabilized and healed even if the infection is L3, which takes a long damn time. As medical player who deals with them often, I think it's fine. L3 infection is a fun, more urgent injury for experienced players and a challenge for newer ones. 

I wouldn't be opposed to adding a damage threshold for infection chance, though. Maybe 15 brute, 10 burn, since burns tend to fester a bit more easily anyway.

Verkister:
Probably also had something to do with the hunger level at the time. My char was pretty much starved by the time she started racking up toxloss, and thus was unable to hold it back by puking.

Screemonster:
I just had a quick logdive and can't find any record of someone using a dirty syringe on you, which was honestly my first gut feeling for "nothing for 10 minutes then suddenly you have turbo-AIDS"

Verkister:
I had shocked myself trying to remove a light tube and got the burns treated at medbay before getting sent out "all clear"

Verkister:
Also imo simple little itching is not a valid warning for an impending death. For the very first sign of a beginning infection maybe. For the bare minimum there should at least be some actual pain going around before you start actually taking damage. The symptom messages for the Level 2 are aching and jolts of pain, but at that point you're already dead.

ArgobargSoup:
As a medical player, I can vouch for infections advancing really fast.

Had someone in medbay for a nasty burn, treated it with advanced burn kits and keloderm.  Sent them out after they were at full health and the body scanner said all clear.

This was before I found out that at the time, advanced burn kits only had half a chance of disinfection.

It was a lot less than 18 minutes when someone dragged them back ta medbay with a *very advanced infection* in the part that had been burned.


That and I've heard several accounts where someone only got one "itchy" message before the toxloss stages started.

Arokha:
Find out what causes 'hyper infections' or if it's just your imagination and fix that. Don't just divide some number by 8. It won't even fix it interrupting scenes. You'll still take toxloss and need to go to medical after the exact same amount of time! Discuss in #development if you need help/advice or have questions interpreting parts of the infection code.

ArgobargSoup:
> Advanced burn kits do not have a half chance of disinfection, they fully disinfect wounds.

> This was before I found out that __at the time__, advanced burn kits only had half a chance of disinfection.

It was a while ago, back when they had the "salve" proc instead of "disinfect".
[https://github.com/VOREStation/VOREStation/issues/3395](url)

PR #3737, opened 2018-05-22
Makes gurgled ID description even more obvious.

Verkister:
People are still thinking they're completely wasted trash.

### DIFF
--- a/code/modules/vore/eating/digest_act_vr.dm
+++ b/code/modules/vore/eating/digest_act_vr.dm
@@ -49,7 +49,7 @@
 	. = ..()
 
 /obj/item/weapon/card/id/digest_act(var/atom/movable/item_storage = null)
-	desc = "A partially digested card that has seen better days. Much of it's data has been destroyed, but the card itself is still completely salvageable. Please recycle it at crew resources office."
+	desc = "A partially digested card that has seen better days. The damage is only cosmetic but the access codes need to be reprogrammed at crew resources office."
 	icon = 'icons/obj/card_vr.dmi'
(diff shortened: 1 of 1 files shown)


SpadesNeil:
Not a fan of the first half, but the second half, you're onto something. Like, who the hell knows what a crew resource office is? We don't have those.

code/modules/vore/eating/digest_act_vr.dm
```diff
@@ -49,7 +49,7 @@
 	. = ..()
 
 /obj/item/weapon/card/id/digest_act(var/atom/movable/item_storage = null)
-	desc = "A partially digested card that has seen better days. Much of it's data has been destroyed, but the card itself is still completely salvageable. Please recycle it at crew resources office."
+	desc = "A partially digested card that has seen better days. The damage appears to be only cosmetic, but the access codes need to be reprogrammed at the HoP office."
```
SpadesNeil: Would be much better if we said, `Much of it's data has been destroyed, but the card itself is still completely salvageable. Please recycle it at the Head of Personnel's office."`

Verkister: I feel like the recycling part and the mention of destruction have been factors behind people considering them trash. I doubt anyone's figured out the "recycling" means just restoring the access instead of getting a replacement and leaving the HoP baffled wondering how the card itself can get turned into materials.

Arokha: It's fine for now. @SpadesNeil can just make another PR to change it if he wants.

PR #3739, opened 2018-05-22
New Explorer Gear

Novacat:
- Adds a new defensive exploration weapon intended for use by rear-line explorers, such as SARs and Pilots
- Ported Machete Sheath from Baystation, intended to be used with the machetes here, they can be worn alongside webbing.
Note: They are not supposed to be used alongside holsters, figuring out a fix.
- Explorer lockers will now come with a survival knife, machete, and machete sheath
- Changed material of explorer melee weapons to plasteel (the stuff is fairly common, anyway)


code/modules/clothing/under/accessories/holster.dm
```diff
@@ -3,15 +3,21 @@
...
 	concealed_holster = 1
 	var/obj/item/holstered = null
 
 /obj/item/clothing/accessory/holster/proc/holster(var/obj/item/I, var/mob/living/user)
 	if(holstered && istype(user))
 		user << "<span class='warning'>There is already \a [holstered] holstered here!</span>"
 		return
+	//VOREStation Edit - Machete sheath support
+	if (can_hold)
+		if(!is_type_in_list(I,can_hold))
```
Arokha: Instead, as one IF statement, `if (LAZYLEN(can_hold) && !is_type_in_list(I,can_hold))`

code/modules/clothing/under/accessories/holster.dm
```diff
@@ -3,15 +3,21 @@
...
+			to_chat(user, "<span class='warning'>[I] won't fit in [src]!</span>")
+			return
 
-	if (!(I.slot_flags & SLOT_HOLSTER))
+	else if (!(I.slot_flags & SLOT_HOLSTER))
```
Arokha: `//VOREStation Edit`

code/modules/clothing/under/accessories/holster.dm
```diff
@@ -3,15 +3,21 @@
...
 
 /obj/item/clothing/accessory/holster/proc/holster(var/obj/item/I, var/mob/living/user)
 	if(holstered && istype(user))
 		user << "<span class='warning'>There is already \a [holstered] holstered here!</span>"
 		return
+	//VOREStation Edit - Machete sheath support
+	if (can_hold)
+		if(!is_type_in_list(I,can_hold))
+			to_chat(user, "<span class='warning'>[I] won't fit in [src]!</span>")
+			return
```
Arokha: Typically we put `//VOREStation Edit End` after blocks of code we've modified.

code/modules/vore/fluffstuff/custom_guns_vr.dm
```diff
@@ -804,3 +804,26 @@
...
+	firemodes = list(
+		list(mode_name="normal", fire_delay=12, projectile_type=/obj/item/projectile/beam, charge_cost = 1200),
+		list(mode_name="low-power", fire_delay=8, projectile_type=/obj/item/projectile/beam/weaklaser, charge_cost = 240),
+	)
+
+/obj/item/weapon/gun/energy/frontier/locked/holdout/proc/update_mode()
+	var/datum/firemode/current_mode = firemodes[sel_mode]
+	switch(current_mode.name)
+		if("low-power") overlays += "taser_pdw"
+		if("normal") overlays += "lazer_pdw"
```
Arokha: We actually use a new overlay system you should use. You can use it by calling `add_overlay("taser_pdw")` and `add_overlay("lazer_pdw")` here instead. It's helpful because it cuts down on lag and other issues (and caches icons so they're faster if you switch often, like on guns).

code/modules/vore/fluffstuff/custom_guns_vr.dm
```diff
@@ -804,3 +804,26 @@
...
+
+/obj/item/weapon/gun/energy/frontier/locked/holdout/update_icon()
+	overlays.Cut()
```
Arokha: Same as above, the new overlay system would use `cut_overlays()` here instead.

code/modules/clothing/clothing.dm
```diff
@@ -632,8 +632,8 @@
 	//convenience var for defining the icon state for the overlay used when the clothing is worn.
 	//Also used by rolling/unrolling.
 	var/worn_state = null
-	valid_accessory_slots = list("utility","armband","decor","over")
-	restricted_accessory_slots = list("utility", "armband")
+	valid_accessory_slots = list("utility","armband","machete","decor","over") //VOREStation Edit - Adds Machete Sheath
+	restricted_accessory_slots = list("utility", "armband", "machete") //VOREStation Edit - Adds Machete Sheath
```
Arokha: The way you could make it noncompatible with holsters is to give holsters and machete holsters the same slot type.

If clothing has "X" in restricted_accessory_slots, then only one item with slot "X" can be an accessory at once. You can see here that "decor" and "over" are not restricted so you could have as many of those as you want.

PR #3750, opened 2018-05-24
Prometheans melt on death.

fludd12:
Prometheans will now properly melt into a puddle upon death again.

See #3801 for why!


fludd12:
Travis wyd

Natje0:
Prommies have never had any problems from resleeving! Due to no ability to get genetic damage, they fall out of the pod at 100% health.

I do like the idea of this method of revival though, but perhaps Slime Jelly would be more fitting?

deathride58:
image

Fund it.

fludd12:
Slime jelly is also significantly harder to get - and, besides, lots of xenobio cores function through phoron injections, I figured it'd be somewhat similar.

Also, I don't even know, supposedly they have a different cloning debuff, who knooooows

Arokha:
More uniqueness is neat.

fludd12:
Done!

fludd12:
Oh, do note that I intend to make similar changes upstream to Polaris proper, so this being in a `_vr` file may not be necessary for long. Make note of that for future parity updates!

SpadesNeil:
If these changes are going to be given to Polaris we should probably wait
for those instead. I have some concerns about how well balanced this is
since I don't want people being able to just revive a dead Promethean on
the fly during an expedition. Then again perhaps I underestimate how
difficult it is to get 40 units of phoron.

fludd12:
40u of phoron is two bars worth, but you'd need to grind it up and get a syringe to actually apply it.
Also, the Promethean will still be quite injured and very hungry post revival, having the choice of either getting less injured or regaining their limbs, but not both.

Ketrai:
Just make sure this is tested to not cause any weirdness with vore bellies please?

fludd12:
I wasn't able to test this personally, but the `gib` function places the organs inside the `loc`, so there should be no issues (*knock on wood*). In the absolute worst case, make a bug report and ping me, and I'll have it fixed within the day!

PR #3759, opened 2018-05-25
Dogborg equipment/belly tweaks and fixes. (Ready)

Verkister:
-Fixes UI autoupdates firing off the popups after closing the window.
-Fixes mob digestion order for clean cycle. Dead mobs will no longer linger on the mercy of the rng against other loose crap.
-Makes the gurgledamages draw water from mobs as well.
-Removes redundant mob_energy corpse gain since corpse handling now comes after the damage gains.
-Sanitizes a whole bunch of UI update stuff.
-The UI window now starts with a less cluttery size.
-Makes clean cycle actually finish the limbo when nothing but protected stuff is left.
-Fixes Pupdozer belly not connecting to water reserves.
-Nerfs maximum material/water reserves to half.
-Various other fixes and streamlining for sleeperbellies.
-Fixes recycled stack materials causing double reserve output for some reason.
-Adds unfinished cargo delivery belly. Not implemented currently, still under testing, but was too far in to remove for the other small updates while waiting for the PR to get merged already in vain.
-Reduces amount of water acquired from sinks/toilets per chug.
-Adds new type of borg hypo for medihound, uses both, cell and water to recharge.
-Added "dry mouth" message to tongue along with some other little cleanups.

-Also fixed runtimes on digest_act for protected items while at it.

### DIFF
--- a/code/modules/vore/eating/digest_act_vr.dm
+++ b/code/modules/vore/eating/digest_act_vr.dm
@@ -19,23 +19,23 @@
 /////////////
 // Some indigestible stuff
 /////////////
-/obj/item/weapon/hand_tele/digest_act(...)
+/obj/item/weapon/hand_tele/digest_act(var/atom/movable/item_storage = null)
 	return FALSE
-/obj/item/weapon/card/id/gold/captain/spare/digest_act(...)
+/obj/item/weapon/card/id/gold/captain/spare/digest_act(var/atom/movable/item_storage = null)
 	return FALSE
-/obj/item/device/aicard/digest_act(...)
+/obj/item/device/aicard/digest_act(var/atom/movable/item_storage = null)
 	return FALSE
-/obj/item/device/paicard/digest_act(...)
+/obj/item/device/paicard/digest_act(var/atom/movable/item_storage = null)
 	return FALSE
-/obj/item/weapon/gun/digest_act(...)
+/obj/item/weapon/gun/digest_act(var/atom/movable/item_storage = null)
 	return FALSE
-/obj/item/weapon/pinpointer/digest_act(...)
+/obj/item/weapon/pinpointer/digest_act(var/atom/movable/item_storage = null)
 	return FALSE
-/obj/item/blueprints/digest_act(...)
+/obj/item/blueprints/digest_act(var/atom/movable/item_storage = null)
 	return FALSE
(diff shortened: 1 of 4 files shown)


code/modules/vore/eating/digest_act_vr.dm
```diff
@@ -19,23 +19,23 @@
 /////////////
 // Some indigestible stuff
 /////////////
-/obj/item/weapon/hand_tele/digest_act(...)
+/obj/item/weapon/hand_tele/digest_act(var/atom/movable/item_storage = null)
 	return FALSE
-/obj/item/weapon/card/id/gold/captain/spare/digest_act(...)
+/obj/item/weapon/card/id/gold/captain/spare/digest_act(var/atom/movable/item_storage = null)
```
Arokha: Wot, why do this? `(...)` means 'accept any named args, they're all fine', which is fine since we're just going to return false no matter what anyway.

Verkister: They were runtiming about item_storage being missing or something while also failing to return false to the thing trying to digest them. One of my tests happened to involve a cap gun, which would use these as a gun subtype, and instead of the gurgle attempt telling the belly "nope, can't gurgle." it would just rutime on each attempt without even adding it to the preserved items list to skip from next attempt.

Verkister: Anyhow yeah. This has been the reason behind the protected items misbehaving in bellies. These things runtiming about the (...) really does break the functionality. The items aren't surviving because this returning FALSE (which it fails to do due to the runtimes), but because this thing still overrides the part where the item would delete. They just get stuck looping in a failing proc that is overriding the deletion. No feedback is returned to the bellies to skip it or to add it to the ignore lists whatsoever. This is why the gurgles still keep going on and on and on without ever stopping if a protected item exists in a belly, and this is also why they never show up in the undigestable/uncleanable lists. For example, IDs and such in borg bellies end the cleanup with "Cleaning finished. # uncleanable objects pls eject." while the (...) items go with "Cleaning in progress. # (protected) items remaining..." being just stuck until the end of time.

Verkister:
Should be ready to merge.

SpadesNeil:
@Arokha How's it look?

SpadesNeil:
Fuckit.

PR #3775, opened 2018-05-27
Adds Heterochromia Marking

Novacat:
Supports most species, with exception of Teshari, Skrell, Sergal, and Vox.


Novacat:
Never touched that marking, guessing something glitched when it saved? Will fix it now.

Greenjoe12345:
Got this ported to Polaris, as well as added a Teshari version

PR #3780, opened 2018-05-28
Corrects explorer permit language

Arokha:
Permit doesn't let them carry guns "on the surface", it lets them carry guns while exploring.


Leshana:
What about passive exploration missions?

Arokha:
:p

Issue #3790, opened 2018-05-29
Tether still lacks casual loitering areas.


Verkister:
#### Brief description of the issue
Tether is still lacking non-remote casual hangout spots. The public hangout spots we currently have are all either in remote edge locations or away from traffic. For example, bar, pool, library, public garden, cafe, and the one seated smoking nook near reading rooms are in remote locations at the edges of the station far from any traffic apart from the people who are intending to go there and are expecting something to be already happening there. And for the other part, the central hangout areas are not very close to "busy" areas, such as the atrium pond hardly getting more traffic apart from people busy just arriving/leaving/heading to dorms and miners heading out or returning for deliveries, and the station side koi park and holodeck essentially having the same situation but with explorers instead of miners, library instead of dorms, and cryo instead of tram, and occasionally engies/heads accessing lift or their departments.

#### What you expected to happen
Better chances of meeting, interacting, and hanging out with people without isolating to departments, singular edge spots, or borderline deserted public areas.

#### What actually happened
The public areas not secluded away from general traffic are hardly but deserted ghost town highways furnished for use as nothing but mere roads.

#### Steps to reproduce
- (Step 1) Exist on a decently populated station with a crew more than capable enough to surpass the busy labor requirements.
- (Step 2) Wander off from your cozy department or other secluded hangout spot.
- (Step 3) Even tumbleweeds have deserted these ghostly ~~high~~hallways. For some reason even the radio communications suddenly feel ominously silent from this perspective.
- (Step 4) Witness a fellow lost soul but find out they're already busy leaving the station.

#### Code Revision
- 2616b963a0136fe21106af82982914ec8c1db2b9
 
#### Anything else you may wish to add:
- Locations here are the general areas between departments and other edge spots.
I have couple vague ideas involving some additional furniture/decor, and sacrificing some of the maintenance/storage areas near the elevators on station levels 2 and 3, and the planetside level 3. Also balconies.

Heroman3003:
There's also park on asteroid 1, but yeah, generally agree on the general intent behind the idea.

Verkister:
Yeah, there are basically no public hangout areas anywhere between active departments.

Very-Soft:
If you suggest some ideas, I can try to map them. There used to be more tables and chairs around, but they were removed for some reason. Not really sure why.

Verkister:
Made some quick half-assed ideas for some locations.
https://i.gyazo.com/487923fe2487278be445cf14151c0a05.png
[...]

PR #3796, opened 2018-05-30
Adds some spots to loiter in

Very-Soft:
Does #3790 : Adds some spots to hang out in as per @Verkister 's suggestions. I used different materials for some bits though. 

space3-2
Small change here, but it shows where a locker and a trash pile were moved to, to make room for the hangout spot. I also added a wall to a corner, so that space wouldn't leak in through the crack in the window.

space3
There's a couple of hangout spots here. Places for folk to hang out if they're waiting for medical, cargo, or the elevator!

space2
Here's a Public Meeting Room! An alternative to the library's game room. It could probably be used by IAAs to have discussions with larger groups of people, in case such should be needed. Or maybe It's where interns plan their investments? Either way, it's a fully functional room.
(I also added a couple decals near the engineering doors, to make them like every other door, as they seemed to be missing)

space 1
Here's a small spot. Now it's got a table and a comfy bench to hang out in. Perhaps it's a waiting area for the holodeck?
(I also moved that one light fixture over BECAUSE IT WAS ASYMMETRICAL REE)

surface 3
A little dining area that overlooks the atrium, similar to how it used to be, but less in the way!

I had some ideas for how to make the last idea not a fire hazard, but surface 3 already has a fair few places to loiter, and I was worried about it cluttering an already high traffic area. 

For your consideration!

### DIFF
--- a/maps/tether/tether-03-surface3.dmm
+++ b/maps/tether/tether-03-surface3.dmm
@@ -10639,9 +10639,6 @@
 /obj/machinery/atmospherics/pipe/simple/hidden/scrubbers{
 	dir = 6
 	},
-/obj/structure/window/basic{
-	dir = 8
-	},
 /turf/simulated/floor/tiled,
 /area/tether/surfacebase/atrium_three)
 "sB" = (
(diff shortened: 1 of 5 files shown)


Ketrai:
👍

Arokha:
This PR is very likely to conflict with the Polaris sync one, which also changes these maps.

KasparoVy:
Really digging these tweaks.

Arokha:
You don't appear to have fixed diagonal corner madness.

Little-119:
Why didn't I notice that asymmetrical light months ago *scream

Very-Soft:
I don't mind asymmetry when it's clearly not intended to be symmetrical. The light bothered me because it seemed obvious that it was meant to be, but it wasn't.

PR #3810, opened 2018-06-01
Adds a fennec crate to cargo.

Ketrai:
- Cost 60
- Either a fennec, or a fennix, but with a smaller chance.

Couldn't put two fennecs in there, I think I know why, but eh.


Arokha:
No more of these web-edits. Do it in a real git client with testing in byond instead of shotgun attempts at passing travis and avoiding typos. I'll just close your next PR like this.

PR #3818, opened 2018-06-02
Changes Terminus

pontifexminimus2:
Makes it from a weird,useless whitelist language to actually something useful - The native Zorren language.


pontifexminimus2:
How the fuck did I break the map diff bot.

Arokha:
It can't do any work while the PR is unmergeable. See below.

PR #3867, opened 2018-06-12
Adds no smoking and fire warning signs to surface airlocks

Mewchild:
And also to areas immediately connecting to them (Public EVA Prep, Mining Prep)
Gotta fight Darwanism.


ariever:
Only the strong will survive

PR #3869, opened 2018-06-12
Fix and Additions to Vox Tanks

Hawk-v3:
Fixed Vox double tanks spawning in internal boxes.
Added indicator to Vox back tank, shifted sprite two pixels to reuse double tank indicators.
Added strip to Vox emergency tank to bring in line with back and double tanks.


Hawk-v3:
"Fix tank spawns" - Taken from Discord
Right so, if you spawned as a Vox engineer, your tank didn't exist because of the code call.
It takes "tank/emergency/" and then adds the species breathtype
In the case of Vox, this is Phoron
This works fine for the standard tanks.
But if it thinks you want a doubltank
It tries to find a emer/phoron/engi
And if it can't do that it looks for phoron/double
You didn't have a phoron/double
But a phoron_double
Which didn't exist
So now it works

Hawk-v3:
I forgot about and therefore broke, the backsprites. Sorry.

PR #3875, opened 2018-06-13
Makes synths able to taste again

Little-119:
Comments out the code added by commit 04ea07fb02e3a5d476db321a24602a8efcb40627

Since Aronai said he made synths able to taste a year ago, and synths not being able to taste was a goof.


Arokha:
Tested?

Little-119:
I spawned as a positronic brained synth chef and stuffed some raw meat and burned food into my mouth and I was able to taste it. So yes.

PR #3889, opened 2018-06-16
[MIRROR] Allows AIs to produce and control maint/construction/mining drones.

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/5282
--------------------
AIs can produce maintenance, mining, or construction drones by clicking on their respective fabricators. This will immediately link them to the newly produced drone, changing its designation to 'remote drone (AI)', and eyes green, while under control.

These drones are capable of normal speech, using all languages and radio frequencies the AI has access to.

The AI, when in control of a drone, is unable to use its core functions including sight or hearing, meaning being within a drone body can be dangerous. Their core is active as per normal while in control. When the AI wishes to return to their core, they can use the 'release AI control' verb to do so.

When carded or un-carded, the AI will be returned to their core. When the carding begins, they will be notified. If the card's wireless is active, and a drone is visible, they can take control of it and act normally with it. If the wireless is inactive, or deactivated while controlling the drone, their drone connection will be cut. If they are being wiped and have wireless, they can maintain control of the drone for a short while before the damage to themselves is too great to do so.

Port & Modification of: https://github.com/Baystation12/Baystation12/pull/18421


SpadesNeil:
Oooooo. Nifty.

PR #3907, opened 2018-06-20
Slight cloning mlem

Wickedtemp:
Halves biomass required for printing a body. 60, to 30.

### DIFF
--- a/code/game/machinery/cloning.dm
+++ b/code/game/machinery/cloning.dm
@@ -23,7 +23,7 @@
 			break
 	return selected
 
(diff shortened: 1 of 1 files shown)


code/game/machinery/cloning.dm
```diff
@@ -23,7 +23,7 @@
 			break
 	return selected
 
-#define CLONE_BIOMASS 60
+#define CLONE_BIOMASS 30
```
Arokha: `//VOREStation Edit`

PR #3925, opened 2018-06-24
Adds "redspace" oties.

Verkister:
-Stats-wise identical to phoron mutants for now.
-Can appear in dangerous pred crate.
-Commissioned for shenanigans etc.

kuva


KasparoVy:
Great work on the sprites as usual

code/modules/mob/living/simple_animal/vore/otie.dm
```diff
@@ -84,6 +84,35 @@
...
 	eyetype = "photie"
 
+/mob/living/simple_animal/otie/redspace
+	name = "feral red otie"
+	desc = "Seems this ominous looking longdog has been infused with wicked infernal forces."
+	icon_state = "hotie"
+	icon_living = "hotie"
+	icon_dead = "hotie-dead"
+	icon_rest = "hotie_rest"
+	faction = "virgo3b"
```
SpadesNeil: This should probably be changed but this isn't a deal breaker.

Maybe change the faction to "cult"? I think that's what faithless use. Is it?

code/modules/mob/living/simple_animal/vore/otie.dm
```diff
@@ -84,6 +84,35 @@
 	glowyeyes = TRUE
 	eyetype = "photie"
 
+/mob/living/simple_animal/otie/redspace
```
SpadesNeil: Although if we're doing redspace oties, but they are called just 'red' oties, should we not just have the file path be `/mob/living/simple_animal/otie/red`? Again, not a deal breaker. Depends what context you want these seen.

PR #3937, opened 2018-06-26
Permit Cleaning

Mewchild:
Comments out a bunch of expired permits and associated weapons


SpadesNeil:
There's no real need to comment out the code for expired permits. I'd rather you didn't, in fact.

You may, however, comment out anything in the custom items config file that spawns these items.

Mewchild:
I spoke with Joan in the development channel as I was doing this, which is why I did the method that I did. Works out though 👍

PR #3942, opened 2018-06-27
Medical Tweaks

Novacat:
Minor balance changes and tweaks. The syringe infection chance tweak is probably the biggest change.

- Advanced Trauma Kits treat 7 damage per application, and are now used only once per body part (instead of once per wound like gauze)
- Advanced Burn Kits treat 7 damage per application
- Medical Rescue Suit injector now has 20 units per chem, and uses Inaprovaline, Dylovene, Paracematol, and Dexalin
- Bicard effectiveness reduced from 6 damage per unit to 4 damage per unit
- Kelotane effectiveness reduced from 6 damage per unit to 4 damage per unit
- Dermaline effectiveness reduced from 12 damage per unit to 8 damage per unit
- Dexalin metabolizes at 25% of the rate of Dexalin Plus, stats adjusted so that the effects are the same, just that it takes longer to wear off (Natje's suggestion)
- Reduced brain healing of Alkysine from 30 to 8 per unit
- Vermicetol is now a T2 brute chem, healing 8 brute damage per unit and metabolizing at normal rates
- All pre-filled syringes start capped.
- There is now a 20-second grace period where infection chance is 0 (Note: Using syringes on multiple people renders grace period null and void)

### DIFF
--- a/code/game/objects/items/stacks/medical.dm
+++ b/code/game/objects/items/stacks/medical.dm
@@ -64,8 +64,6 @@
 	singular_name = "gauze length"
 	desc = "Some sterile gauze to wrap around bloody stumps."
 	icon_state = "brutepack"
-	amount = 15 //VOREStation Edit
-	max_amount = 15 //VOREStation Edit
 	origin_tech = list(TECH_BIO = 1)
 	no_variants = FALSE
 	apply_sounds = list('sound/effects/rip1.ogg','sound/effects/rip2.ogg')
@@ -132,7 +130,7 @@
 	gender = PLURAL
 	singular_name = "ointment"
 	icon_state = "ointment"
-	heal_burn = 3
+	heal_burn = 1
 	origin_tech = list(TECH_BIO = 1)
 	no_variants = FALSE
 	apply_sounds = list('sound/effects/ointment.ogg')
@@ -172,9 +170,7 @@
 	singular_name = "advanced trauma kit"
 	desc = "An advanced trauma kit for severe injuries."
 	icon_state = "traumakit"
-	heal_brute = 6 //VOREStation Edit
-	amount = 20 //VOREStation Edit
-	max_amount = 20 //VOREStation Edit
+	heal_brute = 5 //VOREStation Edit
 	origin_tech = list(TECH_BIO = 1)
 	apply_sounds = list('sound/effects/rip1.ogg','sound/effects/rip2.ogg','sound/effects/tape.ogg')
 
@@ -202,8 +198,8 @@
 					continue
 				if (W.bandaged && W.disinfected)
 					continue
-				if(used == amount)
-					break
+				//if(used == amount) //VOREStation Edit
+				//	break //VOREStation Edit
 				if(!do_mob(user, M, W.damage/5))
 					to_chat(user, "<span class='notice'>You must stand still to bandage wounds.</span>")
 					break
@@ -223,7 +219,7 @@
 				W.disinfect()
 				W.heal_damage(heal_brute)
 				playsound(src, pick(apply_sounds), 25)
-				used++
+				used = 1 //VOREStation Edit
 			affecting.update_damages()
 			if(used == amount)
 				if(affecting.is_bandaged())
@@ -237,7 +233,7 @@
 	singular_name = "advanced burn kit"
 	desc = "An advanced treatment kit for severe burns."
 	icon_state = "burnkit"
-	heal_burn = 12 //VOREStation Edit
+	heal_burn = 7 //VOREStation Edit
 	origin_tech = list(TECH_BIO = 1)
 	apply_sounds = list('sound/effects/ointment.ogg')
 
--- a/code/game/objects/items/stacks/medical.dm
+++ b/code/game/objects/items/stacks/medical.dm
@@ -170,7 +170,7 @@
 	singular_name = "advanced trauma kit"
 	desc = "An advanced trauma kit for severe injuries."
 	icon_state = "traumakit"
-	heal_brute = 5 //VOREStation Edit
+	heal_brute = 7 //VOREStation Edit
 	origin_tech = list(TECH_BIO = 1)
 	apply_sounds = list('sound/effects/rip1.ogg','sound/effects/rip2.ogg','sound/effects/tape.ogg')
 
--- a/code/game/objects/items/stacks/medical.dm
+++ b/code/game/objects/items/stacks/medical.dm
@@ -130,7 +130,7 @@
 	gender = PLURAL
 	singular_name = "ointment"
 	icon_state = "ointment"
-	heal_burn = 1
+	heal_burn = 3
 	origin_tech = list(TECH_BIO = 1)
 	no_variants = FALSE
 	apply_sounds = list('sound/effects/ointment.ogg')
@@ -170,7 +170,9 @@
 	singular_name = "advanced trauma kit"
 	desc = "An advanced trauma kit for severe injuries."
 	icon_state = "traumakit"
-	heal_brute = 3
+	heal_brute = 6 //VOREStation Edit
+	amount = 20 //VOREStation Edit
+	max_amount = 20 //VOREStation Edit
 	origin_tech = list(TECH_BIO = 1)
 	apply_sounds = list('sound/effects/rip1.ogg','sound/effects/rip2.ogg','sound/effects/tape.ogg')
 
@@ -233,7 +235,7 @@
 	singular_name = "advanced burn kit"
 	desc = "An advanced treatment kit for severe burns."
 	icon_state = "burnkit"
-	heal_burn = 3
+	heal_burn = 12 //VOREStation Edit
 	origin_tech = list(TECH_BIO = 1)
 	apply_sounds = list('sound/effects/ointment.ogg')
 
--- a/code/game/objects/items/stacks/medical.dm
+++ b/code/game/objects/items/stacks/medical.dm
@@ -64,6 +64,8 @@
 	singular_name = "gauze length"
 	desc = "Some sterile gauze to wrap around bloody stumps."
 	icon_state = "brutepack"
+	amount = 15 //VOREStation Edit
+	max_amount = 15 //VOREStation Edit
 	origin_tech = list(TECH_BIO = 1)
 	no_variants = FALSE
 	apply_sounds = list('sound/effects/rip1.ogg','sound/effects/rip2.ogg')
--- a/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Medicine_vr.dm
+++ b/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Medicine_vr.dm
@@ -62,15 +62,20 @@
 /datum/reagent/vermicetol
 	name = "Vermicetol"
 	id = "vermicetol"
-	description = "A potent chemical that treats burn damage at an exceptional rate and lasts a while."
+	description = "A potent chemical that treats all damage at an exceptional rate and lasts a while."
 	taste_description = "sparkles"
 	reagent_state = SOLID
 	color = "#964e06"
-	overdose = 10
+	overdose = 15
 	scannable = 1
-	metabolism = 0.02
+	metabolism = 0.05
 	mrate_static = TRUE
 
 /datum/reagent/vermicetol/affect_blood(var/mob/living/carbon/M, var/alien, var/removed)
 	if(alien != IS_DIONA)
-		M.heal_organ_damage(0, 110 * removed) //Not as potent as Kelotane, but lasts LONG.
+		var/chem_effective = 1
+		if(alien == IS_SLIME)
+			chem_effective = 0.5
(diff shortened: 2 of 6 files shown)


code/game/objects/items/stacks/medical.dm
```diff
@@ -130,7 +132,7 @@
 	gender = PLURAL
 	singular_name = "ointment"
 	icon_state = "ointment"
-	heal_burn = 1
+	heal_burn = 3
```
Arokha: Unlabeled edit.

Wickedtemp:
I suggest reducing bicaridine to 4 from 6. 

Reduce the kit heal to 7, max. 

Gauze and ointment are fine, probably shouldnt heal more than 2. 

I'd want this to be played around with before implementing, I was in the middle of doing so yesterday with some of these changes to ensure things don't sway too far to general OPness.

Natje0:
FINALLY someone actually try to work on making the trauma kits actually be useful again. I've been wishing for this to happen for a long, long time.

Arokha:
No bicar nerf?

Wickedtemp:
Bicard should def be 4, I don't even think it needs an upgraded version.

When I get home in roughly 9 hours, I'll play with it and post further thoughts. Most of these were changes I suggested initially and there shouldn't be any huge issue.

Brute and burn heal surgeries should be an option, but we'd need an actual coder to handle that.

Novacat:
Mostly avoided a bicard nerf since, at its current value (6 per unit) it seems to be fine, and there is no Tier 2 version of it like there is for Dylovene and Kelotane

Wickedtemp:
True, but there doesn't really need to be a tier 2, and even as someone who absolutely loves chemistry, we don't need any other chems. 

Brute doesn't need a tier 2 because you won't ever run into a patient with rising brute trauma thats out of control, unless they're actively being mauled, in which case *get them away*. 

Burn can rise after leaving a dangerous area due to body temps or being on fire. Toxins and oxy can rise due to infections, blood loss, lung ruptures, heart damage and crit. So it fits that these have upgraded versions. 

Also, bumping bicard down to 4 makes it all equal with kelotane and dylovene. So I'm still pushing for it

Wickedtemp:
After messing around with Bicaridine, Dermaline, Kelotane, Trauma and Burn kits and such, I think all of the changes to the reagents are fine. 

Gauze/Ointment healing 1-2 damage could be neat but it doesn't matter and wouldn't have an effect ingame either way. 

Only thing this would really need to kind of finish decentralizing everything around chems would be the surgical operations that heal brute/burn damage, rough details on what I'd suggest for it are here: https://forum.vore-station.net/viewtopic.php?f=20&t=1287

And... if that's added? We'll have a Medical system that isn't simply about injecting medicine.

Arokha:
> And... if that's added? We'll have a Medical system that isn't simply about injecting medicine.

I suspect people will still always just do chem anyway, but it'd be interesting to see.

code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Medicine_vr.dm
```diff
@@ -62,15 +62,20 @@
...
 
 /datum/reagent/vermicetol/affect_blood(var/mob/living/carbon/M, var/alien, var/removed)
 	if(alien != IS_DIONA)
-		M.heal_organ_damage(0, 110 * removed) //Not as potent as Kelotane, but lasts LONG.
+		var/chem_effective = 1
+		if(alien == IS_SLIME)
+			chem_effective = 0.5
+		M.heal_organ_damage(60 * removed, 60 * removed * chem_effective)
+		M.adjustOxyLoss(-120 * removed * chem_effective)
+		M.adjustToxLoss(-60 * removed * chem_effective)
```
Arokha: Lemme talk to admins about this part. This part has some 'rezadone-like' implications.

Novacat: It heals 3 brute/burn/toxin per tick, and 6 oxy per tick, due to the very, very slow metabolization, which is why it is more super tricordazine than rezadone which heals a huge amount in a very short time.

Wickedtemp:
@Arokha - I mean chemistry will still be done and chems made will still be useful and no doubt used frequently, but it won't be the "be all, end all, one and only solution" for brute/burn anymore, which evidently was one of the main issues people had with it. Adding an alternative that makes use of surgery - an alternative that's faster than Chemistry if you have a good surgeon - would hopefully resolve that issue. 

And honestly, I'm indifferent towards vermicetol. It shouldn't be super powerful, though I do like the idea of it being a better tricordrazine instead of a better burn heal, because like... we already have KeloDerma, which works great already. Could even make vermicetol the defacto tier-2 for brute healing. We get solar grubs regularly enough that the chem could get made often enough.

code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Medicine_vr.dm
Arokha: Actually in talking to Temp and doing the math, a single piece of grubmeat is enough to give you 12 minutes of massive healing. This should be nerfed, or the recipe should be made more expensive (or both)

Wickedtemp: Recommending nerf, speed up metab rate and decrease healing so it'll give maybe like... 2 per tick max and not nearly for as long. It might actually make vermicetol useful enough to make.

Alternatively, making it the 8 heal brute chem could be considered as well. Solar grubs happen what feels like nearly every round. The only issue is that I'm not sure if you can butcher the solar grub larvae and they tend to get killed pretty quickly - before they mature. But that's an IC thing that can probably be worked out.

LiquidFirefly:
How about we like, don't do this?  Just gonna kinda, quote myself from the forum thread about the syringe change here.   

> Making healing chems 2/3rds as effective, (barely)buffing the far less effective medkits which in a way just perpetuates whatever is going on with this (The syringe) change, etcetera. All that pull is doing is trying in vain to pull people away from chems and push them towards something only meant for and only effective on minor injuries, or for sealing injuries. I do not know what the developer issue with chems is, but continually hearing about it and seeing changes proposed to make medical continually less potent is stupid.

The trauma kits were already used and useful for minor wounds and sealing wounds.  That's what they were meant for.   Chems always were better tools for healing and generally working around wounds, and they were more convenient as well.   I don't see why we have to fix a system that isn't broken, using chems to fix patients isn't an issue, and I don't know why we're continually seeing it as one.

code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Medicine_vr.dm
```diff
@@ -62,15 +62,17 @@
...
 	name = "Vermicetol"
 	id = "vermicetol"
-	description = "A potent chemical that treats burn damage at an exceptional rate and lasts a while."
+	description = "A potent chemical that treats physical damage at an exceptional rate."
 	taste_description = "sparkles"
+	taste_mult = 3
 	reagent_state = SOLID
 	color = "#964e06"
-	overdose = 10
+	overdose = REAGENTS_OVERDOSE * 0.5
```
Arokha: Isn't this just a 'more obscure' 15, which is what it was before this latest commit?

code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Medicine_vr.dm
```diff
@@ -62,15 +62,17 @@
...
-	metabolism = 0.02
-	mrate_static = TRUE
 
 /datum/reagent/vermicetol/affect_blood(var/mob/living/carbon/M, var/alien, var/removed)
+	var/chem_effective = 1
+	if(alien == IS_SLIME)
+		chem_effective = 0.75
 	if(alien != IS_DIONA)
-		M.heal_organ_damage(0, 110 * removed) //Not as potent as Kelotane, but lasts LONG.
+		M.heal_organ_damage(8 * removed * chem_effective, 0)
```
Arokha: This is better-ish. Seems to be 1.6/tick, from a drug only obtainable from solargrubs, which seems reasonable.

code/game/objects/items/stacks/medical.dm
```diff
@@ -198,8 +198,8 @@
 					continue
 				if (W.bandaged && W.disinfected)
 					continue
-				if(used == amount)
-					break
+				//if(used == amount) //VOREStation Edit
+				//	break //VOREStation Edit
```
Arokha: Changed ATKs but not gauze?

code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Medicine_vr.dm
Novacat: I did it to match other chems in the main file.

code/game/objects/items/stacks/medical.dm
Novacat: Correct. My thought is to give ATKs a bit of a usability boost over standard gauze.

Wickedtemp:
@LiquidFirefly 
TLDR: You're correct, and I agree with you, but enough people feel differently that changes are being made. But, given that this would mean that treatment takes like... literally one to two additional minutes due to bicaridine being a bit slower, it's not that big of a deal. 

Long version:
I personally don't think chemistry, aside from alkysine, needed to be changed at all. But that's evidently the minority viewpoint. People were constantly suggesting various nerfs to chemistry without any thought into the fact that if it weren't for chemistry, there would be no healing. There are no other options. There are no alternatives. You'd have brute/burn kits for 49 damage per limb or less, but other than that, chemistry is your only option.

The second part of this PR, brute/burn heal surgeries, will hopefully alleviate that and make the players that are freaking out, not freak out. The only damage type that's actually being affected here is brute. Burn is still healing fast due to KeloDerm. Toxin is unchanged completely. Oxy was actually made slightly easier by slowing Dexalin's metab rate and adjusting the amount healed per unit to keep it the same so it heals for a longer period of time. 

I think that this change, while not really all that necessary, isn't major, due to the fact that out of all of the damage types, the one that's easiest to deal with already was the one that was made 30% slower to heal with its chem. If you get to someone with a case of oxy damage, it's likely that oxy damage has a risk of increasing. Same goes for toxin and burn damage. Brute, however, only increases if you're actively being attacked or if you're in a vacuum, in both cases, as soon as you're out of the hostile environment, the damage won't rise. It's static. This makes it easier to treat than the other three damage types, which all have the risk of rising. 

Because of this, Vermicetol, which is made with solar grub meat, is going to be a tier 2 version of bicaridine, because frankly, while it's nice, it isn't at all necessary. And if things go according to plan, we'll have a surgery to heal brute/burn from limbs soon-ish anyway to give another, potentially faster means of healing brute damage. Worst case scenario, there's still cryo, which is unchanged and works great for every damage type and can heal them to full health in a matter of seconds. 

To clarify:
A full syringe of Dermaline will heal 120 burn damage. A syringe of KeloDerma will heal 90 burn damage, but at a faster speed than straight Dermaline. This is honestly fine. 

Bicaridine takes a reduction from healing 90 per syringe to 60 per syringe, but it's a lower priority, less dangerous damage type. The side effects, such as organ damage and bleeding, are the primary causes of concern, so making use of bicaridine, tricord, cryo, trauma kits, this really isn't a big deal either. 

And then there's upcoming PR at some point to add in the surgical alternatives which, in the hands of a decent surgeon, would heal a bit faster than bicar would in the same time-frame. Combining these might honestly reduce time spent in medical, simplifying things to "Print report, inject, 4 surgical steps, print report, discharge"

And I know that leaving scenes to handle idiotic injuries is annoying. You know that I know this. Medical is literally the only department I play and I've been playing here for several years. But, the thing is? That's fine, to me. If having to do medical stuff instead of scening really bothered me, I'd...probably go off duty, or at the least, pick a different, less frequently essential department. That was something that was present before this PR, and here's what will change following this PR if it's merged.

Going with a brute damage example, since that's the only reagent that's actually going to be kinda slow-ish at healing. Someone is brought in with 100 brute damage and has a broken leg. Bicaridine, currently, heals at 1.2 per tick, taking about 83 ticks to heal that damage completely, assuming nothing else is used. Add maybe 40 seconds for the broken leg and you'd have left your scene for around 120 seconds - two minutes. Add three minutes for transit, up to five minute interruption. 

Assuming the same situation with the PR's value? Roughly 120 ticks for brute to be healed entirely, add forty for the surgery, three minute for transit time, it's roughly a 40 second increase. When brute heal surgeries are in, you can reduce it back again by handling that while you fix any fractures, and until then? It's still not that big of a deal. Using only bicaridine to heal 199 damage, assuming they somehow didn't die, it'd add on an extra 80 seconds. Just over 1 minute of additional healing, assuming you use nothing else, at the max amount of damage possible without dying, in a damage type that will never increase in any meanginful rate as long as the patient is in a safe area and not actively being mauled.

PR #3943, opened 2018-06-28
Chaplain PDA Bug Fix & Atmos Dedupe

Hawk-v3:
-Chaplain PDA bug: Caused by the chaplain custom preview calling equip. Disabling it seems to have no issues with the preview and stops it making a new PDA every time it generates a new preview.

-Atmos Tech Dedupe: All engineers start with a T-Ray in their right pocket. Atmos techs also had one in their toolbelts. And now the extra one in the toolbelts is gone.

### DIFF
--- a/code/game/jobs/job/civilian_chaplain.dm
+++ b/code/game/jobs/job/civilian_chaplain.dm
@@ -149,5 +149,7 @@
 		feedback_set_details("religion_book","[new_book_style]")
 	return 1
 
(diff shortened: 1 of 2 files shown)


SpadesNeil:
We should forward the bug fix to Polaris.

Arokha:
Bork bark bork?

Hawk-v3:
I'm looking into pushing it to Polaris. It's just been a difficult week. Give me a bit.

Hawk-v3:
https://github.com/PolarisSS13/Polaris/pull/5387

Issue #3675, opened 2018-05-12
Perhaps Warden/HoS should get access to the Red Armory again.


MrSign:
#### Brief description of the issue
No access to the red armory unless its a code red.

#### What you expected to happen
To be able to freely use the red armory, even if it ain't an emergency. (Sometimes, I wanna shoot guns in the firing range on slow shifts, alright?)

#### What actually happened
Access Denied!

#### Steps to reproduce
- Join as HoS or Warden
-Go to red armory
-Become depressed as you cannot get to those delicious lethal weapons.

#### Code Revision
849c1374a1b064b9c9a1fb8c1c43ed19a4259100 
 
#### Anything else you may wish to add:
Armory could also do with an expansion! Attempting to put back flashers or barriers if you used them results in you having to smash a window to get them perfectly aligned! (Or crowbar, screwdriver, crowbar the window.)
![blockade](https://user-images.githubusercontent.com/33539568/39954936-58eb7546-560b-11e8-8b44-3cabed1f2166.JPG)

Leshana:
Functioning as designed.

Arokha:
I mean we discussed this as admins and decided we'd compromise on this setup. :v Are you going rogue

SpadesNeil:
If it were up to me, Security would have lethals all the time.

Well, it technically is up to me, I just don't want a coup.

MrSign:
My opinion has shifted to allowing for at least ONE person in security to access it, preferably Head of Security. Just to allow for some spice in low pop rounds where all you wanna do is mess around with guns and stuff. And considering action events have been rather low/non-existance might as well have SOME fun by shooting at cardboard cutouts.

SpadesNeil:
HoS should be given access.

Arokha:
Leshana suggested maybe we just make it a key card auth thing so you have to convince at least one other person but don't need code red

SpadesNeil:
[...]

PR #3956, opened 2018-07-01
Security Overhaul

Jakksergal:
*A complete redesign of the security department with an intent of increasing inter-departmental interaction and interaction with prisoners while also making navigation between key rooms less cumbersome.*

**Changes:** *I may miss some things..*

- Incorporates IAA into security. Their access throughout the department is limited to the visitation evidence storage room, processing room, briefing room, and break room. They do not have access to the northern parts of security's main halls.
- Introduces The Hole. Specifically for ~lewding prisoners in private~ institutionalizing convicts.
- The brig is now very secure. Prisoners are no longer one short flight away from freedom. Flying is now mostly harmless as there is nowhere for them to land on the upper level.
- Processing is now at the forefront of security to make it less of a walk. There are tinting windows to prevent rubberneckers.
- Many doorways litter the hallways of security to ensure that if a breach occurs, it's not an absolute blockade to security's operations.
- Riot control makes a return. Laughing gas can be filtered out by six huge scrubbers controlled from within the observation room.
- The brig now has goodies stored in a small storage closet along with orange jumpsuits and two prisoner lockers meant for solitary confinement cells. Goodies should be given out on request for well-behaved prisoners.
- Introduces an observation room where the entire yard can be overseen from one chair.
- There are now more lockers, badges, and minor gear items in the security locker room. Just six crowbars and flashlights like NCS Northern Star had.
- Security officers will no longer have to scour security for a vacant charger. They're now very abundant with **four** right as you enter the department in a quick restock area.
- Security now has a dedicated EVA room. It's currently half-stocked as admins were concerned about power creep.
- There is now a publicly accessible airlock just outside Security's EVA in maintenance.
- The blue armory now has a room for confiscated weapons or weapons received from research.

**Minor changes:**
- The electrical maintenance shaft in engineering was slightly modified.
- Maintenance on the second level was tweaked to have no more double walls for no reason.
- The main upper hallway was straitened out. Medbay's lobby had to be compacted by one tile.
- A floor decal sprite was fixed.
- The teleporter was moved down a tile just like medbay's lobby. The maintenance behind it now slightly wraps around medbay.

**Images:**
upperdeck
lowerdeck
iaa former


AaronSherwood117:
I have no strong feelings one way or the other.

Jakksergal:
The orange color in the yard and cells will be corrected in an upstream PR (Hopefully). The color I chose for them can be seen here: https://cdn.discordapp.com/attachments/187013248309002240/460680770495709204/unknown.png

Screemonster:
What's with those three space tiles on the top right of the top floor in the screenshot there?

Jakksergal:
That was fixed in 95c3342

SpadesNeil:
Solitary confinement (I think that's the top-most cell) should probably have a camera in it.

SpadesNeil:
What is this cell? 
image

CameronWoof:
I believe that's "The Hole" mentioned in the changelog.

CameronWoof:
I really like this in general, though I _would_ like to suggest that the storage room down on the brig level be repurposed into a first-aid station. With evidence being upstairs, there doesn't appear to be much point to it, and the addition of a brig physician alt-title for Security could bring some nice flavor.

SpadesNeil:
The layout of the armory is far superior to its original setup.

One thing I'd like done, if it wasn't done already. The Red armory door should have access from the HoS outside of Code Red. This was decided in an admin vote a long time ago in a 9-2 majority.

That said, I'd like to keep it secure enough that the Warden can't just waltz in either, or the HoS has to go through at least one extra step. However, I want the Warden to _have_ Red Armory access _during_ Code Red.

Two ways we can do this. I really don't understand how the Red armory door currently works but my understanding is that it allegedly becomes open when red alert is reached. We could put that door in the Warden's office, and the HoS's access somewhere else. This is clunky and not preferred. The second way is that we change how the door is coded so that the door starts bolted, and becomes unbolted in code red or delta, and automatically re-bolts when the alert level goes to blue or green. _However_ there would be a button in the HoS office with HoS only access that could unbolt the door at any time needed.

The second issue would completely resolve the problem proposed in https://github.com/VOREStation/VOREStation/issues/3675

That said, if you cannot do that, just ensure the door is normally bolted but with Warden access, and the controls to unbolt them are HoS access. Ergo, when unbolted, the HoS and Warden can both access the room.

Jakksergal:
@SpadesNeil The way it's coded currently is that it checks to see if the security level is code red and if not, just denies access altogether. I don't know how to code in what you're requesting. Since the door's never bolted in its current state, I can slap a button next to it with HoS level access that opens it up.

I could also slap a button next to the blast doors outside that was HoS level access while the internal button is only warden access.

The room you asked about is The Hole. Since the yard is completely visible at all times, I slapped that in there for private scenes. It doesn't lock, it's got no lights, it's filthy, but it's private. Still part of the brig's area, though.

Arokha:
... severity tags don't make sense with map refactors 9.9

Jakksergal:
All major issues are resolved, key features were privately playtested. I'm calling that the final commit for pre-merge QA.

SpadesNeil:
>I can slap a button next to it with HoS level access that opens it up.

@Jakksergal Do this, then. I think the button will still force open the door with the right command.

Jakksergal:
Fucking.. I didn't even manipulate any turf! Come on Travis!

SpadesNeil:
image

Should probably fix it anyway.

Yoshax:
I'd like to complain about the security front desk. 

- There's a table in the way of subtle lewds. 
- It's also kinda cramped. 
- There's also not a way to get from the front desk to the lobby, without having to go out and around which is a little annoying. 
-There's also not really room for more than two people, and I know one occasion it is nice to sit with a few colleagues there and chat.
- Maybe I'm blind, but there doesn't seem to be a fax machine accessible to officers within the department.

All of those are reasons why I created the existing front lobby and desk the way it is. I think the rest looks okay.

Jakksergal:
You're shitting me.

Yoshax:
I mean... That's not really the case though? It's not impossible to fix. You could move the whole behind the desk and IA part left a tile, and have a three-tile wide bit behind the desk. Which I'm pretty sure there's perfectly space for. If you don't want to fix it, that's a different story. But it's not not possible. I also don't see much point in planning to fix something 'later', that's pretty much a quick way for it to not get done. Just do the job fully in the first place.

Jakksergal:
At this point I'm scared Travis is going to break down my door and murder me if I try to push another change.

SpadesNeil:
Security front desk looks exactly how I want it to look. I do not agree with the suggested changes.

PR #3962, opened 2018-07-01
Saber SMG fix

lbnesquik:
Saber fix


lbnesquik:
You might want to close this actually. I've put it on polaris itself here.
https://github.com/PolarisSS13/Polaris/pull/5384

PR #3965, opened 2018-07-02
Ups the paper text limit.

Cameron653:
Not to big of a change, just makes librarian's job a bit easier.

**Please do tell me if there is a better method of going about this.** I was thinking of replacing the sanitize when you write in books with the pencode 
(https://github.com/VOREStation/VOREStation/blob/c8dd30846cfba8b058ec97399e98e8ac3649924b/code/modules/paperwork/paper.dm#L286)

But I wasn't sure if that would break everything and open it up to exploits, especially since I couldn't 'test' such a thing due to lacking a book database. Please do tell me if this would work, however.

### DIFF
--- a/code/__defines/misc.dm
+++ b/code/__defines/misc.dm
@@ -105,7 +105,7 @@
 
 // Setting this much higher than 1024 could allow spammers to DOS the server easily.
 #define MAX_MESSAGE_LEN       2048 //VOREStation Edit - I'm not sure about "easily". It can be a little longer.
-#define MAX_PAPER_MESSAGE_LEN 6144
+#define MAX_PAPER_MESSAGE_LEN 24576 //VOREStation Edit - Quick QoL for librarians, since you can write HTML onto paper and put it in a book binder, but can't write HTML onto a book.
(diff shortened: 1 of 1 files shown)


Arokha:
The correct way is to probably write a non-interactive version of the pencode that allows formatting without 'signatures' and stuff.

Cameron653:
Ah. So copy & paste it and then remove the things like sign, form, etc?

Arokha:
That's my opinion, anyway. And Polaris would probably want that.

Or, alter the paper sanitize and add a parameter `interactive = TRUE` and pass it `FALSE` on books. That's probably even better and doesn't involve copypaste. Then just group all the things together that are 'interactive' like the [sign] and stuff, and put them inside an if(interactive) thing in that sanitize proc

Cameron653:
Aright. Will work on this some more after my fruit PR is all done.

SpadesNeil:
Did this get sent to Polaris?

Cameron653:
I need to send this upstream, still. Closing this and will send it to Polaris sometime.

PR #3975, opened 2018-07-03
Security Overhaul fixes.

Jakksergal:
I swear half of these I fixed before. I blame the dmm to tgm conversions.
*The rest is incompetence.*

**Changes:**
Fixes yet more wires.
Fixes a missing disposal junction.
Adds window panes to the exterior window of IA.
Adds a shutter to the warden's office. Slightly shuffles a few items in their office as well.
Adds a fax machine to the briefing room for general security use.

It'd be naive of me to say that the fixes stop here. Leave this PR open until just before the update.

### DIFF
--- a/maps/tether/tether-06-station2.dmm
+++ b/maps/tether/tether-06-station2.dmm
@@ -167,16 +167,13 @@
 /turf/simulated/floor/tiled,
 /area/security/brig/visitation)
 "ar" = (
-/obj/structure/bed/chair{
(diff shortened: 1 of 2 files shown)


Jakksergal:
As much as I want to believe everything major's been fixed, that's probably not the case; however, I don't want to try manage conflicts anymore so I think it'd be a good idea to merge this PR and worry about issues as they come up post-update.

PR #3976, opened 2018-07-03
Removes some redundant code.

SpadesNeil:
ToDo: Fix missing sound effects for guns.

Why the fuck does this check map diff renders?

### DIFF
--- a/code/modules/vore/fluffstuff/custom_guns_vr.dm
+++ b/code/modules/vore/fluffstuff/custom_guns_vr.dm
@@ -51,7 +51,7 @@
 	origin_tech = list(TECH_COMBAT = 4, TECH_MATERIAL = 2)
 	magazine_type = /obj/item/ammo_magazine/m95
 	allowed_magazines = list(/obj/item/ammo_magazine/m95)
-	fire_sound = 'sound/weapons/battlerifle.ogg'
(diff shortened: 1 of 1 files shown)


Cyantime:
If the firing sound is defined on the gun, it overrides the projectile's effect. This was intentional with the phaser.

Novacat:
This does not happen, or is broken. The projectiles sound is overriding the gun sounds.

Arokha:
All PRs check for map and icon diffs. That was the first one to run, though, and failed because:

image

SpadesNeil:
:laughing: :gun:

Anewbe:
~~Oh look, it's this jackass Anewbe jumping in again for no real reason!~~

The projectile sound is [supposed](https://github.com/VOREStation/VOREStation/blob/master/code/modules/projectiles/gun.dm#L680) to override the gun's sound, for things where multiple projectile types with different sounds can be fired (ie. Shotguns, Eguns).

SpadesNeil:
@Anewbe In the context of shotgun ammunition, I see now why you would want to do this.

That said, I think the override should be the other way around. Different ammo may produce different sounds, but unique sounds may be desired for the sake of fluff--especially with energy weapons--while still using the same projectile as other, similar weapons.

But that's just my opinion on the matter.

PR #3983, opened 2018-07-05
Couple VS compatibility tweaks on the synthlimb markings option.

Verkister:
~~-You can now have clean robolimbs on your artfully painted organic bodies again.
-In char setup near the synth color option.
-Tested working although seems to require a physical change (colors, markings, limb status, etc.) to refresh the body cache before it can show up in preview.
-May cause a moment of confusion for painted synths on the first character load after the update, as despite being enabled by default, the thing will start out disabled because of the old savefiles lacking the added var. Just pop it back on and hit save to update it.~~

Been merged to Polaris as well so this thing pretty much just exists to fix a bunch of compatibility issues with our edits.


KasparoVy:
Excellent!

Verkister:
Possibly. They lack the part enabling the markings on synths as well so this would essentially be an option to enable them for Polaris instead heh.

SpadesNeil:
Conflicts aaaaaaaaaaaaaaaa

Verkister:
Fixed

PR #3989, opened 2018-07-06
Polaris Hard Sync

Arokha:
Mostly just commit history.


code/__defines/subsystems.dm
```diff
@@ -32,7 +32,6 @@ var/global/list/runlevel_flags = list(RUNLEVEL_LOBBY, RUNLEVEL_SETUP, RUNLEVEL_G
 #define INIT_ORDER_LIGHTING 0
 #define INIT_ORDER_AIR		-1
 #define INIT_ORDER_PLANETS	-4
-#define INIT_ORDER_HOLOMAPS	-5 //VOREStation Add
```
SpadesNeil: Do we still need this? I feel like we still need this.

code/__defines/subsystems_vr.dm
```diff
@@ -0,0 +1 @@
+#define INIT_ORDER_HOLOMAPS	-5 //VOREStation Add
```
SpadesNeil: Oh, guess not.

PR #3998, opened 2018-07-09
Removes ears from DSI Vulpkanin head sprites, enables body color for DSI limbs

Little-119:
#2467 removed the ears from the normal organic Vulp sprite, but left them for the DSI Vulpkanin prosthetic sprites. This updates the DSI Vulp sprites to match the organic Vulp sprites.

_Side note: Aside from perhaps consistency with other robolimbs or 'making way' for this to change, why have a seperate dmi file for the DSI parts when they're entirely identical to their base species' sprite?_

This also sets the skin_tone variable to 1 for all of the DSI parts. This is otherwise exclusive to Vey-Med and Zeng-Hu, but since the DSI parts are basically the Vey-Med for ~~the superior species~~ non-humans, logically they should have that too, no?

I put both changes in their own commits, in case that's the proper way to do it. Do tell me if it'd be better to squash them.


KasparoVy:
According to [this line's comment](https://github.com/VOREStation/VOREStation/blob/a5b5f8e59d975d2a73cefed7f8ce27eb69a0c941/code/modules/organs/robolimbs.dm#L43), this skin tone would be best suited for species who aren't intended to be coloured RGB style (i.e. Humans only).

Vey-med parts, although alike to DSI are actual imitations of their organic counterparts, have `skin_tone` set to 1 because they imitate Humans who have skin tones and aren't coloured RGB style.
This means that DSI limbs having `skin_tone == 0` is correct since Vulpkanin, Tajara and the like aren't coloured by skin tone but by RGB.

Little-119:
Humans are colored RGB style on here, code wise. A human can technically be neon green. On Polaris, humans use a scale of 1 to 220 for their skin tone, as well as Vey-Med and Zeng-Hu.

All the change does here is make DSI parts use the Body Color instead of 'Synth color'. With or without the change, they're imitating their organic counterparts.

So, what you're saying might be true on Polaris, but not here.

KasparoVy:
Nice, fair enough-- thanks for clearing that up

Arokha:
> why have a seperate dmi file for the DSI parts when they're entirely identical to their base species' sprite

They are not identical. Put them in Beyond Compare or something else that will do image comparisons and you will see the differences. Mostly in where the limbs join together, they have been replaced with sockets instead of 'meat hanging out' of the organic sprites.

Little-119:
@Arokha
I compared the sprites manually, and here's what I found

Vulpkanin sprites seem to be 'cut out' from a whole Vulpkanin sprite, and as a result are simply flat at the limb connections in both sprite sets, and additionally both the DSI and organic sprite sets are identical. Paracode had the same issue until (IIRC) someone fixed it.
Teshari and Vasilissan appear to have this issue as well, having no limb connections. Although in the case of the latter, that might just be an aspect of arachnid anatomy.

The Tajaran and Sergal DSI sprite sets have 'fleshy bits', that appear to be organic (i.e. are pink and (if bone is exposed) bone yellow). Oddly, the Tajaran DSI sprite seems to have slightly different shading, while the Sergal sprite is just identical.

Nevrean, Akula, and Unathi sprites seem to have what Arokha described, with synthetic-ified limb connections. (Though it looks as if they're more or less just recoloring the fleshy bits to white (255,255,255) or black (42,42,42). But I guess that works.)

Arokha:
Pretty sure I remember the 'connectors' being like two pixels with a 1 pixel gap between. The others might have been overwritten by people over the years. The INTENT was for them to have connectors there. Also I know the races themselves have been touched up by people like Verk so the DSI ones might not be accurate anymore to the original fleshy versions.

With those things in mind, can you make this PR conform to that intent? Does this 're-fleshize' any that have connectors currently or anything?

Little-119:
Huh? This PR just updates the Vulp DSI heads to not have ears like the organic vulp sprites, via copy paste. As I said, Vulp sprites are just 'cut out', so there's nothing to refleshize.
I mentioned the stuff about DSI vs their organic counterparts simply because of what I observed while making this PR.

PR #4012, opened 2018-07-12
Spawns off an alert in character setup.

Cyantime:
Prevents a popup from holding up server startup.


SpadesNeil:
Wait, misread this. Not a Polaris bug.

PR #4013, opened 2018-07-12
Fixes a typo with kitsune udon.

avabee:
Goodbye, Friend Egg. You will be sorely missed.


KasparoVy:
Legit thought it was intentional up until now. I am disappointed

PR #4024, opened 2018-07-13
Minor Post-Update Securiy Overhaul fixes.

Jakksergal:
Adds various missing objects and other general fixes.


Jakksergal:
Most of these were pointed out by Phi.

Mewchild:
A couple more things to fix:

- [ ] No washing machine in the brig
- [ ] All the beds are steel, and not padded as they were previously. I can see this for solitary, but not the regular cells
- [ ] No linens for the beds.
- [ ] Detective doesn't have access to interrogations
- [ ] Update the NanoUI maps

Mewchild:
This door is public access. A mouse opened it
image

KasparoVy:
It's clearly a hostile agent.

Mewchild:
No button in the lobby desk to open the main/common doors. Was removed.

Jakksergal:
The door to EVA is meant to be all access so those with access to the airlock but not security don't have be escorted five tiles. It is red because it is still a security area, but not a *secure* area. Just like the lobby of security.

Jakksergal:
This is ready to be merged to prevent conflicts down the road.

PR #4030, opened 2018-07-14
[MIRROR] Language and Larynx Additions

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/5402
--------------------
Holy crap did this creep.
- Every species has a list of languages it needs assistance to speak. In the same vein as EAL, they can still understand any language they have normally.
- Added a Voicebox/Larynx organ, whose sole purpose at the moment is to assist in speaking the aforementioned assisted languages.
- Language implants now add their associated languages to the voicebox's list of languages to assist, instead of being a free-floating implant that'd need to be checked for every individual language. You still need to select the language to understand or speak it.
- I rewrote a bunch of implant code as a tangent to getting the language implants to work. I did in fact test every implant I could think of, they all seem to work fine now.
- FBPs start with a special Voicebox that assists with EAL by default.
- Common Skrellian (and Far Skrellian) require an assisted (aka implanted) voicebox to speak, at Skrell Maintainer request.
- Alai can only be spoken by Taj, Teshari, and appropriately shapeshifted Prometheans, at Tajara Maintainer request.


Wickedtemp:
So, how is this going to mess with characters here? Plenty of non-tajaran species speak Siik simply because it's effectively used as a "feline" language, and we have quite a few languages and species in our code that Polaris lacks entirely.

Little-119:
The solution to the former thing could probably be solved simply by making the list of languages mentioned in the first bullet point be empty for Custom Species. It'd probably be fine to have this work as Polaris intends for non-custom species.

Personally, I'd like to piggyback off this addition of a larynx organ to allow FBPs to choose whether they have the normal speech bubble or the synthetic speech bubble, rather than have it be based off the brand of their torso.

Wickedtemp:
I think I still have Tempest as a tajaran mechanically. There used to be a fuckton of issues with the custom species so I didn't use it.

PR #4036, opened 2018-07-14
Adds Marking Order Changing at Character Creation

KasparoVy:
QOL for folks with numerous markings on their characters.

Testing/adding/manipulating markings becomes massively time consuming the more you have, especially if you're wanting to put it deep in the layer stack and are testing speculatively.
This introduces much convenience to that process: Now you can click a button and move a chosen marking up or down a layer at will, allowing you to simply create a marking as normal then shift it to the desired layer in the stack without having to re-do a ton of other markings just to get it where you think it might work best.

**Should this be closed here and opened for Polaris?**

:cl:
rscadd: You can now change the order of your body markings at character creation with the push of a button. Shift markings up or down a layer at will in order to design the character you've always wanted, more easily than ever before.
/:cl:

========================================

Clicking the back `<` button on a marking will trade its place with the _previous_ marking or _move it to the end after_ all other markings if it's the _first_ marking.

Clicking the forward `>` button on a marking will trade its place with the _next_ marking or _put it in first before_ all other markings if it's the _last_ marking.

_The buttons_
virgo marking mover

_Moving markings_
virgo marking mover 2

### DIFF
--- a/code/modules/client/preference_setup/general/03_body.dm
+++ b/code/modules/client/preference_setup/general/03_body.dm
@@ -300,7 +300,7 @@ var/global/list/valid_bloodtypes = list("A+", "A-", "B+", "B-", "AB+", "AB-", "O
 
 	. += "<br><a href='?src=\ref[src];marking_style=1'>Body Markings +</a><br>"
 	for(var/M in pref.body_markings)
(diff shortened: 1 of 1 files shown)


Verkister:
Yess this is good

Arokha:
I'd make this for Polaris, yeah. I don't see any reason they wouldn't want it! If for some bizzaro reason they don't, then reopen it.

KasparoVy:
Thanks! Will post there.

PR #4059, opened 2018-07-20
[MIRROR] Fixes indoor weather, hopefully for real this time.

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/5424
--------------------
The visuals were only half of the story, apparently.
ChangeTurf was changed to implicitly preserve tile's 'outdoor-ness' state.
It turns out this breaks a lot of things, such as weather effects in shuttles, buildmode, late-loading PoIs for events, and likely others.
To fix this, I've made it so copying the previous tile's outdoor state must be explicit, with an argument. The default is to not copy the previous tile, as that is what it was before the change. This fixes the above-mentioned issues while still allowing for things like building floors outside to not become magically roofed, by being explicit.


SpadesNeil:
rage

NOW WORK YOU FUCK

PR #4115, opened 2018-08-01
[MIRROR] Makes Storms Thunderous

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/5459
--------------------
The storm weather type will now periodically cause lightning strikes on a random outdoor turf.
This causes a big bolt to come down from the heavens and do various mean things to whatever is nearby. It produces a soft flash effect for all outdoor tiles, and a thunder sound is played to people on the planet.
Getting struck directly shouldn't kill you instantly (by itself) but you will want to go to medical.
Being sufficiently indoors will protect you from strikes, or having a grounding rod nearby.
Grounding rods and tesla coils attract lightning. They both redirect the strike to their tile, but do different things. The grounding rod stops any bad effects from happening, while the coils cause the bolt to bounce and get absorbed into their powernet, as if it was a tesla ball hitting them. The tesla coil has higher priority.
Simple animals that get struck get hurt much more, for that RNGesus saves feeling when it strikes the combat drone thats about to kill you (if that ever happens).
Admins have a fun button to do a cosmetic or real lightning strike, anywhere.
Adds ten thunder sounds, from various files on freesound.org with C0 (public domain) licenses.

Video: https://streamable.com/se0hk


SpadesNeil:
So many fucking conflicts. I would fix them if I wasn't stupid and remembered how to fix conflicts properly.

PR #4116, opened 2018-08-01
Proteans now able to use Teshari DSI prosthetics

Sharkmare:


kcin2001:
but why

Arokha:
Is fine, suggested this change.

PR #4123, opened 2018-08-03
[FEATURE]Sleevemate new functions

Zap-zapper:
Sleevemate is now able to detect soulcatchers, and also load minds from soulcatchers, and upload them into bodies without mind. Because, why you can steal someone's mind and place it in your soulcatcher, but to do the reverse procedure, you require medbay help? Not gud.

Screenshots:
If you scan someone without mind:

If you scan someone without mind, but with stored mind in Sleevemate:

If you scan someone with soulcatcher:

If you scan someone with soulcatcher and with stored mind in Sleevemate:

If you scan someone with soulcatcher, with minds inside it:

### DIFF
--- a/code/game/objects/items/devices/scanners_vr.dm
+++ b/code/game/objects/items/devices/scanners_vr.dm
@@ -76,7 +76,9 @@ var/global/mob/living/carbon/human/dummy/mannequin/sleevemate_mob
 
 	//Mind/body comparison
 	output += "<b>Sleeve Pair:</b> "
-	if(H.mind && ckey(H.mind.key) != H.ckey)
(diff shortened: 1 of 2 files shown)


Zap-zapper:
Gonna choo-choo out to a place without proper net, until 10-th august. So I can analyse any reviews only after this date.

Arokha:
Doesn't this let people join into soulcatcher and then get put into bodies without joining the round? Please prevent that possibility.

Zap-zapper:
Everything done. You can't upload someone's else mind in your sleeve if it is protected. And you can't unload ghosts from soulcatcher.

PR #4130, opened 2018-08-05
Fix Teshari Hiding Again

### LINKED ISSUE #4101: Teshari Hiding Breaks (Again) When you Cross Tiles
Teshari hiding seems to be broken again.
Not the `flags` (tracking whether a mob is hiding or not) but the `layer` as while your layer is set appropriately to the hiding one when you use the verb, as soon as you move from one tile to another it gets reset to `MOB_LAYER`, or 4.

Resting and getting back up does not cause your layer to reset to 4 like in the previous bug, so that part still works.
So long as you stay still, you remain at the intended hiding layer indefinitely, so that still works as well.

When this happens, in order to hide again you have to turn off hiding and turn it back on again.

I expected that, while moving, I'd stay on the hiding layer and not have to refresh it by revealing myself and hiding again with the verb.

#### What actually happened
I took a step and revealed myself, although my `status_flags` still indicate I should be hidden. I viewed my variables to see that my `status_flags` indicate I'm hiding but my `layer` is back to 4, the `MOB_LAYER`.

#### Steps to reproduce
- Toggle hiding as a Teshari or any other mob with the verb.
- Step in any direction.

#### Code Revision
Server revision: master - 2018-07-29
115bfa2dc315103a8638db1ed54bf0d02513663f

#### BYOND Version
5.0 Beta
Version 512.1442

KasparoVy:
Unfortunately Taur riding and Teshari hiding conflict with each-other.
This fixes #4101 where moving in any direction would reset your layer, thus revealing you unintentionally while also not resetting your plane or status.

A mob becoming revealed has been broken off into its own proc where you can pass a custom fluff message to the hider or have it reveal them silently.

This also adds conditions where you will be revealed when the following occurs:

1. You are grabbed or grab something.
2. You buckle or are buckled to something.
3. Someone buckles to (rides) you.


SpadesNeil:
This looks fine but I am not the most competent programmer here.

Issue #4050, opened 2018-07-17
Unknown Vore-Preferences failed to save error.


H0lySquirr3l:
https://i.gyazo.com/thumb/1200/04b53bf9d6f709c8201ac0e904f4aec9-png.jpg

#### Brief description of the issue
Two players tjavse and WanderingDeviant have found a weird issue. 

 tjavse:  "well... I just open the vore tab, and making no changes, or making.. any amount of changes, when I click save preferences. it gives me the error"

WanderingDeviant :"Opened the vore panel. Didn't touch ANYTHING but select the Default Stomach (0)
Hit Save Vore Prefs at the bottom
Got an error"

They are unable to save their vore prefs.

#### What you expected to happen
It would save the default information.

#### What actually happened
They simply receive this error.
https://cdn.discordapp.com/attachments/468917252561829918/468919813268439060/Bug_2.PNG

#### Steps to reproduce
- Open vore panel.
- Select default stomach.
- Press "Save prefs"

#### Code Revision
- 6831216aa58b34b2d6610f7de92c556b69272fbc 
 
#### Anything else you may wish to add:
- I H0lySquirr3l was unable to recreate this glitch. Yet WanderingDeviant was unable to do it while both on the server at the same time. They also logged off and logged back on for me and the bug persisted.
https://cdn.discordapp.com/attachments/468917252561829918/468921651468304394/Bug_3.PNG

Edit: even when WanderingDeviant makes a new character the bug persists. 
![image](https://user-images.githubusercontent.com/34927367/42851084-643cb130-89f0-11e8-8b63-bfdf2c171277.png)

tjavse also confirms that it carriers over to between characters.

H0lySquirr3l:
7/28/2018 AFter attempting to make a new byond acount and trying to save prefs. It gave me the same error message. https://cdn.discordapp.com/attachments/187013248309002240/472855750155173925/unknown.png

R1f72r:
I also have the same issue, that being my preferences have MEVER saved. The recreation rate is 100%
What do I do? I dunno I just log in and try to save changes, or no changes at all in my vore panel and get the message “ERROR: Virgo-specific preferences were unable to save!” Or something along those lines.

Very-Soft:
I ran into someone else who had this problem today, day 1 newbie, perhaps there's a problem with generating a new save file for the preferences, while overwriting existing ones works fine?

I had them try a new character slot, and they had the same issue there too.

PR #4145, opened 2018-08-07
Updated to version 0.3.1-vorestation of rust-g

Leshana:
Fixes failure to save files when the containing dirs don't exist yet.
For example when attempting to save vore-preferences for the first time on an account and it needs to create the `vore` subfolder of the save directory.

Thus this fixes VOREStation/VOREStation#4050  Tested on windows so far, Travis will helpfully build on linux for us.


Leshana:
See  VOREStation/rust-g#1 for actual changes.  Recompiled for `i686-unknown-linux-gnu` and `i686-pc-windows-msvc` to make this PR.

SpadesNeil:
i have no idea

Issue #3972, opened 2018-07-03
[Suggestion] Revert changes made to syringes.


Jakksergal:
I don't believe that this change accurately represents the wishes of the community and needs to be reverted to its previous state. Not changed to make more viable, not repealed and replaced, reverted. There should be no need to simulate some sort of pseudo-realism in a 2D spessgame.

Since the reason from this change stems from community preference, I'll leave a strawpoll here.
https://www.strawpoll.me/16010811

Heroman3003:
There is a forum thread discussing those changes already.

Wickedtemp:
A forum thread that can now have a strawpoll linked to it, which I'm doing at the moment :p

Draycu:
Honestly?  Reverting would be a better change, and the poll is reflecting that at the moment.  Thanks for making the issue here on Git, Jakk.

nerdass:
why don't we add a research creatable self sterilizing syringe that can do like 30 injections on a charge and can hold bigger reagent containers than 15 units

Wickedtemp:
Why don't we just revert the annoying thing

AaronSherwood117:
I skimmed the thread and popular opinion kind of agrees.  In theory, the change was a nice bit of fluff, but in practice it's just a bunch of extra hassle without any real benefit added.  I don't play medical much anymore, but I'm going to voice support to a revert simply because it's such a common consensus.

Little-119:
The best solution would probably be to change the sprite and(/or) fluff of the syringes to be a future tech self-sterlizing syringe/hypospray or something, but revert their mechanics to the old syringe mechanics. So, you'd get the fluff benefits of the change while not affecting gameplay.

nerdass:
reusing needles is also incredibly gross. like no offense but I don't see why people are upsetti in the spaghetti over the disposable needles being enforced as disposable.

like i don't mean offense but from a roleplay perspective, cross contaminating all your bottles by dipping the syringe into all of them, while not mechanically enforced, is probably something that your characters would all have been taught not to do in med school.

SpadesNeil:
I'm of the opinion that _how about we give the other departments ways to participate?_

A self-cleaning syringe should be super easy for science to make, and so suddenly players from science actually get to do something helpful.

But meanwhile, medical isn't completely fucked because **you guys literally have boxes and boxes of syringes**. How many does each box hold?

Draycu:
[...]

PR #4161, opened 2018-08-13
Undoes the Syringe Infections

Mewchild:
As per: https://github.com/VOREStation/VOREStation/issues/3972#issuecomment-403142802
Reverts the changes that give syringes the chance to infect
Reverts Syringe boxes to hold 7 needles per box, as it used to be

All code work was done by @LiquidFirefly

### DIFF
--- a/code/modules/reagents/reagent_containers/syringes.dm
+++ b/code/modules/reagents/reagent_containers/syringes.dm
@@ -161,9 +161,8 @@
 				return
 
 			var/mob/living/carbon/human/H = target
-			var/obj/item/organ/external/affected //VOREStation Edit - Moved this outside this if
 			if(istype(H))
-				affected = H.get_organ(user.zone_sel.selecting) //VOREStation Edit - See above comment.
+				var/obj/item/organ/external/affected = H.get_organ(user.zone_sel.selecting)
 				if(!affected)
 					to_chat(user, "<span class='danger'>\The [H] is missing that limb!</span>")
 					return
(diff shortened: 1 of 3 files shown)


code/modules/reagents/reagent_containers/syringes.dm
```diff
@@ -203,28 +202,26 @@
...
+				add_attack_logs(user,target,"Injected with [src.name] containing [contained], trasferred [trans] units")
 			else
 				to_chat(user, "<span class='notice'>The syringe is empty.</span>")
 
-			dirty(target,affected) //VOREStation Add
 
 	return
-/* VOREStation Edit - See syringes_vr.dm
+
 /obj/item/weapon/reagent_containers/syringe/update_icon()
```
Arokha: Why do you want to revert to the old update_icon? The new one is far more efficient and actually makes the new icons work. Did you test this...? Even if you did, "no", you can't use overlays.Cut() and stuff anymore, you need to use the functions so things don't break.

code/modules/reagents/reagent_containers/syringes.dm
```diff
@@ -203,28 +202,26 @@
...
 
 			if(trans)
 				to_chat(user, "<span class='notice'>You inject [trans] units of the solution. The syringe now contains [src.reagents.total_volume] units.</span>")
-				if(ismob(target))
-					add_attack_logs(user,target,"Injected with [src.name] containing [contained], trasferred [trans] units")
+				add_attack_logs(user,target,"Injected with [src.name] containing [contained], trasferred [trans] units")
 			else
 				to_chat(user, "<span class='notice'>The syringe is empty.</span>")
 
-			dirty(target,affected) //VOREStation Add
```
Arokha: This is the only line you need to alter. The rest of the stuff is totally unrelated to syringe infections and will just fuck up all the icons.

LiquidFirefly:
It reverts syringes entirely to how they were before. That means no extra clicks from caps, and a change back to the old icons.   This was appropriated from something that I did in my free time simply to sit there and see how long it would actually take to revert this, not thinking it'd ever get reverted.  If it was a single line, why was this not done literally a month ago when it was announced it should be changed?

Arokha:
Dunno, if you have free time, why didn't you make a PR to change that line?

LiquidFirefly:
This had cursory testing, syringes looked fine injecting and taking in reagents.

Arokha:
I get that it looks fine, but you don't understand what I'm saying. The non-SSoverlays method of messing with overlays is obsolete and shouldn't be used anymore. We (and tg and everyone else) have switched to using an overlays subsystem that has specific macros for dealing with overlays, as shown in the other file. You need to stick with that.

LiquidFirefly:
Perhaps it's because I don't pay that much attention anymore.   It's not my self appointed _job_ to sit here and make changes to the server, and make sure that changes that are supposed to happen, happen.   It's someone elses.

Arokha:
Well, it's nobody's self-appointed job other than Ace's, really. Everyone else was appointed by others!

Doesn't matter, anyway. Needs to be fixed so it uses SSoverlays and doesn't re-implement old overlays.Cut() stuff lest we get back to crashy-server-times.

Mewchild:
I’ll make the suggested changes when I’m home from work

SpadesNeil:
image

Actually it's nobody's _**job**_ to do any of this. We donate our time and effort to something we enjoy. LiquidFirefly, if all you're here to do is bitch, then you can go somewhere else.

Mewchild, you keep up the good work and we'll provide help where we can.

Mewchild:
Hey now. Let’s keep it civil and chill. Don’t want these bad vibes in my email from y’all

Mewchild:
Remaking this PR with the recommended changes.

SpadesNeil:
Sounds fine to me. Will lock this old thread.

PR #4163, opened 2018-08-13
Reverts Syringe infections

Mewchild:
As per https://github.com/VOREStation/VOREStation/issues/3972#issuecomment-403142802

Removes chance for syringes to infect.
Restores Syringe boxes to 7 per box (polaris default), down from 20


SpadesNeil:
So, I still want to discuss alternatives to this, as the problem of powergaming doctors is still a problem, but as promised, this change is being reverted for the time being.

PR #4165, opened 2018-08-15
(HOPEFULLY) fixes advanced who highlighting the fake ckeys of stealth admins

deathride58:
Title. Also shaves off redundant code from the player half of advanced who and replaces the `src << msg` with `to_chat(src, msg)`.


Screemonster:
This might be better off sent upstream if they have the same bug.

PR #4176, opened 2018-08-18
Add "monster mouth" and "blank round face" sprites to body markings

iskawhiskers:
note: "blank round face" is required to fit the monster mouth; icons are seperate to allow colouring for both face and mouth interior seperately
image


Screemonster:
how does the monster mouth differ from the existing jagged teeth marking?

Verkister:
It's like "W" while the jagged is like "TT"

iskawhiskers:
actually it's the same sprite. the difference is that the monster mouth is vertically centered on the face and closer to the horizontal center on side sprites for characters whose faces mainly consist of mouth

i made it mostly out of sprite OCD because being forced to use taj/vulp heads for "mouth monster" characters looked horrendously ugly to me if they weren't supposed to have long faces

Screemonster:
Makes sense to me. Kinda like the difference between the "fangs" and "tusks" marking which is literally 1-pixel height difference to put them above or below the mouth.

KasparoVy:
Long facea for life

iskawhiskers:
binch..........

PR #4177, opened 2018-08-18
Makes item digestion gradual and belly damage based.

### LINKED ISSUE #4109: Eaten Objects digest instantly
If you eat an object with the Trash Eater ability, it instantly disappears regardless of damage settings or itemweak flags. 

Stuff in guts to cough up or keep as desired.

#### What actually happened
Poof

#### Steps to reproduce
Have Trash Eating, consume an object, open Vore Panel to check contents.

#### Code Revision
release - 
6831216aa58b34b2d6610f7de92c556b69272fbc

Verkister:
-No more instantly disappearing everything.
-The initial "health" of the items is based on their w_class. (but the w_class is not the health, so you can't just go gurgling things lighter)
-Default damage (2+2) will take 3 cycles to chew through a normal sized (w_class 3) item. Basically one cycle per w_class.
-Lower damages will chew through shit slower, and higher damages will do it faster. Max damage (6+6) will get through a normal sized item with one cycle.
-They will all require two hits however, same way as how mobs take the deadly blow before final meltdown.
-Also the nutrition gains scale according to the damages now.

Fixes #4109


SpadesNeil:
Objects should not give you nutrition. They're objects. :U

Unless those objects are food items, but that's out of scope for this PR.

SpadesNeil:
[Travis hated that.]

PR #4179, opened 2018-08-19
Improves Tesla Engine submap

Michab02:
https://i.gyazo.com/af76ea6d1c5f392d287d0b96246ee25d.png

Removes redundant radiation collectors and wires, makes the force field area more compact, and adds an extra set of tesla coils to the default startup. Generally makes it a lot cleaner and sensible.


Michab02:
Also - in the screenshots a few wires look like they're missing, but they're just underneath the floor tiles. Dreammaker is a bit odd like that.

PR #4221, opened 2018-09-05
Synced pAI translator languages with robot ones

AbbieFland:
This makes the 35-point universal translator module also provide the
vorestation-specific languages robots get, such as Canilunzt.

I'm reasonably sure at one point most (or at least several) of these
came for free to pAI without even needing a translator, and it doesn't
feel overpowered to me to allow a handless toolless character built for
communication and assisting people to spend 1/3 their upgrade budget to
gain borg-tier language skills.

### DIFF
--- a/code/modules/mob/living/silicon/pai/software_modules_vr.dm
+++ b/code/modules/mob/living/silicon/pai/software_modules_vr.dm
@@ -1,0 +1,35 @@
+/datum/pai_software/translator
+	name = "Universal Translator"
+	ram_cost = 35
+	id = "translator"
+
+	toggle(mob/living/silicon/pai/user)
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/silicon/pai/software_modules_vr.dm
```diff
@@ -0,0 +1,35 @@
+/datum/pai_software/translator
+	name = "Universal Translator"
+	ram_cost = 35
+	id = "translator"
+
+	toggle(mob/living/silicon/pai/user)
```
Arokha: just override /datum/pai_software/translator/toggle(...), no need to override the whole thing

PR #4222, opened 2018-09-05
[MIRROR] Buffs the Vox

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/5545
--------------------
_Let it not be said that I never buff things._

As per the requests of our most ~~durable~~ trusted Raiders:
- The fiddly gear-bits on Raider masks have been properly oiled, meaning they can open and close their feeding ports again. Should stop the filthy meat and their pepperspray.
- Every pair of Raider underwear will accept the same accessories as filthy meat pants. They are also being reissued with lots of bags to fill with loot.
- The Vox mesh now accepts that it is supposed to be worn on clothing. For some reason, this includes meat clothing.
- Vox weaponry has been convinced to shoot more often before going on break.
- Vox armor has been made lighter.
- Vox outcasts have been instructed on how to better craft their horrible scrap armor to protect slightly more evenly.

EDIT:
- Fixes the bug that lets people eat through gas masks.
- Fixes the whole PHORONGUARD tag, apparently past-Anewbe was putting it in the wrong place.


AaronSherwood117:
This should probably be touched up or modified a bit before being added here.

PR #4231, opened 2018-09-09
Text for handheld plushies when poked

Affurit:
Plushies now make sounds when poked. Squeak!


Screemonster:
Are plushies a polaris thing? If so, need //vorestation edit on each changed line. For this though, rather than commenting every single line with the added var, it'd be easier to make a toys_vr.dm file and add all the extra definitions in there.

Affurit:
God, really...?

SpadesNeil:
Or we can just send this to Polaris!

SpadesNeil:
Is now a duplicate of https://github.com/VOREStation/VOREStation/pull/4301

PR #4234, opened 2018-09-10
Tactical Kittens

Novacat:
- Adds Ascian's Kitten flavor item
- Changes normal kitten's inventory size, tactical kittens for everyone!
- Fixes a bug with the medical hardsuit preventing it from dispensing dylovene

https://forum.vore-station.net/viewtopic.php?f=27&t=1425

Ascian's kitten sprites are very placeholder right now, I plan to get a unique sprite in the future.


config/custom_items.txt
```diff
@@ -825,6 +825,12 @@ character_name: Tony Bingham
 item_path: /obj/item/clothing/head/fluff/runac
 }
 
+{
+ckey: tabiranth
+character_name: Ascian
+item_path: /obj/structure/largecrate/animal/tabiranth
```
Mewchild: You're going to spawn with a large crate in your inventory. The ones that you can't pick up, and need to be opened with a crowbar.

Novacat: Yes, they will be auto-dropped at the floor, but unfortunately that is the only way to bring a live animal on board. Animals will not spawn inside boxes and such.

Mewchild: It may also make sense to give yourself a crowbar then, so you're not tugging this crate around at the start of every round.

Screemonster: Why not make a more portable version of the crate that _does_ allow animals to spawn inside it? OR a portable object that spawns the kitten when used in-hand?

PR #4235, opened 2018-09-10
Fixes misaligned tajaran leg.

Verkister:
At least noticing this took less time than it took to notice that original kneefuck calamity.


KasparoVy:
I'd LOVE to hear about the kneefuck calamity.

Verkister:
The side views all had one leg facing forward and the other one backward :v

KasparoVy:
Thanks for clearing that up lmao, indeed how horrifying

Issue #4246, opened 2018-09-16
Belly contamination too recursive.


ArgobargSoup:
#### Brief description of the issue
Bellies contaminate _**everything**_, regardless of where it is or how it's stored or how deep it is.

Accessories attached to clothes.
Everything in a box.  (That lightbulb box they were carrying?  Enjoy cleaning over 20 bulbs.)
Alternate modes of power tools, like the drill.  Gotta wash it in both modes.
Encryption keys in headsets.

#### What you expected to happen
Not having to spend 10+ minutes cleaning things up when someone hands me a gurgle-contaminated bag.

#### What actually happened
oh god everything is gross

#### Steps to reproduce
- Gurgle someone with itemweak on.
- Expel items.
- Have to dig down through *everything* to get it all clean.

#### Code Revision
- release - 6831216aa58b34b2d6610f7de92c556b69272fbc 
 
#### Anything else you may wish to add:
- The discussion that brought this up also brought up how, well, the whole "IDs get ruined" thing inconveniences everyone *but* the pred.  If the pred actually wanted to steal an ID, they'd strip it beforehand.
Relevant because this issue is probably caused by an indiscriminate check to make sure an ID was gurgled no matter how deep it was in the player's belongings.
- washing a soggy cardboard box with water to make it usable again

Screemonster:
Making sure IDs get ruined was an executive decision to stop easymode ID-theft via noms.

Making people individually wash every single item in an item including contents of lightbulb boxes, contents of their toolbelt, and god knows what else, is _silly_.

Not to mention that contamination also has other stupid side-effects like people being able to meta when someone's eaten because the PDA server tells them their PDA is "slimy".

Very-Soft:
I generally agree with what's said here, and I'll add, that the outright ruining of IDs regardless of circumstance, is sort of anti-fun. 

I'll just say that, while I understand that the ID ruining was well intended, to prevent preds from just getting access from those they ate, in practice the only people it really regularly inconveniences are the prey. In practice, if a pred wants to steal someone's ID for the access, they'll just pull the ID off of the prey first. Further, if someone's going and abusing someone else's access, that's sort of more of an ahelp issue, rather than something that should outright be prevented. 

[...]

PR #4287, opened 2018-09-29
Small qol tweaks for contamination and sleeperguts.

Verkister:
-Removed the toxicity of gurgled items. (Planning to make them affect the user's examine flavor when equipped though.)
-Removed the silly animation on mess overlay. (possible remake later with extra flavors etc)
-Fixed sleeperbellies not suppressing injury emotes and deathmessages during digestion.
-Ruled out the possibility of the sleeperbellies breaking over defined "patient" going "missing".
-Disabled item contamination on hold bellymode when itemweak is on.
-However the itemweak holdmode combo will still digest food items to prevent infinite lunchbox aboos :v


Screemonster:
Can you remove contamination for non-damaging bellymodes when itemweak is on as per #4246? The need for IDs to be ruined if you digest your prey without stripping them is something that doesn't apply if the belly isn't damaging the owner.

The main situation that this applies to is things like someone being endoed, holding their PDA in their hand, then a brief connection dropout makes them SSD and drop whatever they're holding.

Verkister:
Alright as long as nobody's against it heh.

Verkister:
Made it so that DM_HOLD with itemweak enabled no longer contaminates everything. It will still digest loose food items to prevent abuse though.

PR #4295, opened 2018-10-01
Fix gateway scatter landmarks

Arokha:
They're pinned to only working if you use the old fashioned map loading text file thing, and won't work with admin-loaded maps or anything like that.


Very-Soft:
\o/ SUCH FUN WILL BE HAD

PR #4299, opened 2018-10-02
Bug abdomens

ArgobargSoup:
Adds a whole bunch of abdomens.  And by a whole bunch...

-One normal abdomen with 5 different coloring option.  All of them aside from the first being dual colors.
--Single color
--Bee stripes, top half
--Bee stripes, all around
--Underside
--Firefly tip

-A *fat* version of the abdomen, with the same coloring options.

-A third variant, that uses the `*vwag` emote to switch between normal and fat, and the same coloring options.

-And finally, Vasilissan versions of every one, due to them having a different leg shape.

I'm *pretty* sure I got these in without messin' with any other tail things...

Some pics:
-A lineup of the FAT abdomen sprites.  A WIP pic from a while ago, but this essentially shows the coloring options.
fatbuggolineup

-A pic of the firefly abdomen, with the `*vwag` option.
vwagtest


Arokha:
Wish the `vass` ones weren't necessary and that's a bit annoying to me, but otherwise seems fine.

PR #4303, opened 2018-10-03
Fix a wrong indication in *help

### LINKED ISSUE #1465: *yes emote forgets an s
*yes is not recognized, it returns "Unusable emote 'ye'. Say *help for a list."

Clearly it doesn't recognize the single s in yes. If you type *yess then the emote works

*yes

#### What actually happened
*ye

#### Steps to reproduce
- be synthetic
- try to use the *yes emote
- boggle as it tells you you forgot an S for the fifth time in a row even though you know you hit it

lbnesquik:
Simply add an s to "yes" so it guides player to typing the correct command.

Fixes #1465

### DIFF
--- a/code/modules/mob/living/carbon/human/emote.dm
+++ b/code/modules/mob/living/carbon/human/emote.dm
@@ -30,7 +30,7 @@
 				m_type = 1
 
 		//Machine-only emotes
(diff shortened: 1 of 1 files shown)


SpadesNeil:
You're fixing a lot of stuff that's Polaris code.

Issue #3828, opened 2018-06-03
The observe button's warning about respawning lies


Little-119:
#### Brief description of the issue
When you click the 'Observe' button in the lobby, it says you will have to wait one minute.
Until now, that's been correct. Now, the respawn timer after observing is apparently _5_ minutes.

#### What you expected to happen
To be able to return to the lobby after observing for one minute.

#### What actually happened
Had to wait five minutes to return to the lobby.

#### Steps to reproduce
- Observe
- Wait for a period of time between 0 and 299 seconds
- Click 'respawn'

#### Code Revision
- ad18753b9d6bfd50b51846b726f025f1e781a75c
 
#### Anything else you may wish to add:
![](https://i.imgur.com/4zJIqiF.png)

Natje0:
The bug is what the Observe button claims.

SpadesNeil:
Did it actually make you wait 5 minutes, or did it just _say_ 5 minutes?

Little-119:
Had to wait five minutes. I thought the screenshot showed that.
Either way, I'll test it locally now just to double check.

Cameron653:
Portion of code involved:
https://github.com/VOREStation/VOREStation/blob/4cddf6ccd0f7552358042c918bd6324e4a9bf7d7/code/modules/mob/mob.dm#L376-L380

Rixunie:
Thanks Polaris.

Verkister:
https://github.com/VOREStation/VOREStation/pull/3881#issuecomment-397611226
Arokha: "It wasn't an oversight. I asked other admins at the time what they wanted to do and they said it was fine to leave as 5min."

PR #4304, opened 2018-10-03
Fixes syringes cutting apples

### LINKED ISSUE #1779: syringes slicing apples
when you interact a syringe with an apple (either on draw or inject) it will 'slice the apple but when you interact with the slice it says "None of the apple slice left!" unless you inject a reagent inside, then you can bite it normally until the reagent is gone

to be able to inject something into the apple and not to create face apple slices

#### What actually happened
i got fake apple slices that are basically little containers

#### Steps to reproduce
- (Step 1) optain apple and syringe
- (Step 2) click syringe onto apple
- (Step 3) try to eat apples and then find them to dissapear into nothingness

#### Code Revision
256e1db4d6e43368e3e53cf3027e5ae395abdacd 

- i would suspect the same thing happens to any slicable fruit but i was trying it with apples

lbnesquik:
This makes syringes not sharp. I was able to take a blood sample from someone and put it within a beaker with those settings.
Fixes #1779

### DIFF
--- a/code/modules/reagents/reagent_containers/syringes.dm
+++ b/code/modules/reagents/reagent_containers/syringes.dm
@@ -17,7 +17,7 @@
 	volume = 15
 	w_class = ITEMSIZE_TINY
 	slot_flags = SLOT_EARS
-	sharp = 1
+	sharp = 0 //this is to prevent them from slicing apples and so on.
 	unacidable = 1 //glass
 	var/mode = SYRINGE_DRAW
 	var/image/filling //holds a reference to the current filling overlay


lbnesquik:
https://github.com/VOREStation/VOREStation/issues/3828

Mewchild:
Fixes https://github.com/VOREStation/VOREStation/issues/1779, not #3828

lbnesquik:
Thanks for the correction.

Mewchild:
Make sure you add "Fixes [ISSUE]" To your original post, so it closes the issue when this is merged

Issue #3887, opened 2018-06-15
Post-cryopod respawn time inconsistant


Heroman3003:
#### Brief description of the issue
Using tram bump-cryo to despawn allows you to respawn immediately. Any other cryopod that you ghost out of requires you to wait all 5 minutes.

#### What you expected to happen
All cryopods being usable to not have respawn time

#### What actually happened
Only tram bumb-cryo does that

#### Steps to reproduce
- Bump-cryo tram and try respawning
- Ghost out of any other cryopod and try respawning

#### Code Revision
- ad18753b9d6bfd50b51846b726f025f1e781a75c

PR #4305, opened 2018-10-03
Fix misleading alert.

### LINKED ISSUE #3828: The observe button's warning about respawning lies

lbnesquik:
This correct a dated overlook.
Fixes #3828
This does not correct #3887

### DIFF
--- a/code/modules/mob/new_player/new_player.dm
+++ b/code/modules/mob/new_player/new_player.dm
@@ -114,7 +114,7 @@
 
 	if(href_list["observe"])
 
-		if(alert(src,"Are you sure you wish to observe? You will have to wait 1 minute before being able to respawn!","Player Setup","Yes","No") == "Yes")
+		if(alert(src,"Are you sure you wish to observe? You will have to wait 5 minute before being able to respawn!","Player Setup","Yes","No") == "Yes")
 			if(!client)	return 1
 
 			//Make a new mannequin quickly, and allow the observer to take the appearance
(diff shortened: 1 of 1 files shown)


Mewchild:
The problem with this, as indicated in #3887, is that the tram still allows you the instant respawn

Mewchild:
The problem with this, as indicated in #3887, is that the tram still allows you the instant respawn.
Also, you linked an issue with nothing to do with this. https://github.com/VOREStation/VOREStation/issues/3828

lbnesquik:
Correction is once more appreciated. I got all mixed up typing those. Regarding the tram, i think that's another problem that, while related, doesn't affect this PR.

Mewchild:
You can't say it's fixing an issue, when you're not in fact fixing it. You're saying it's fixing a misleading alert, while the tram will still be incorrect and have a different time from the others (At instant, as opposed to the 5 that's now in the warning/alert)

SpadesNeil:
Needs // Vorestation Edit

lbnesquik:
Did it with github built in editor. Not sure it's very helpful there.

PR #4307, opened 2018-10-03
Remove kobold.dm

lbnesquik:
This file has no sprites and is used nowhere. It does have some kinda cute emotes but without sprites it's useless and has been since 2017 and maybe much longer. I haven't found any references of it's origin either.


lbnesquik:
If you want to give this sprites, feel free but please don't take a year. ~~Preferably not a month either~~

Very-Soft:
I gonna make kobold sprites hold my beer

PR #4320, opened 2018-10-06
Renaming statuette for clarity and ease of spawning.

lbnesquik:
Rename /obj/item/weapon/vampiric into /obj/item/weapon/statuette for clarity.


Mewchild:
Either do //Vorestation Edit or send this to polaris

Arokha:
Don't think Polaris would mind. Do it there.

Issue #3906, opened 2018-06-20
Flying clumsiness issues


ktccd:
#### Brief description of the issue
When you fly, it's incredibly hard to RP being graceful, as walls, doors and even tables make you fall over painfully, making you have to endure quite a lot of painful chat-spamming if you were trying to do something nice in a scene.
It seems so clumsy as to be some kind of bug.

#### What you expected to happen
To be a graceful, elegant demon, floating over a table and landing next to my partner, or through a corridor or other nice setting. Mostly for RP stuff, I wanted to be elegant, slowly hovering around.

#### What actually happened
I flew like a crippled duck with 1 wing and bumped on every surface possible with little to no control.
At the speed I flew I should not be THIS bad. I could see it if I had a rocket strapped to my back, but this was... embarrassing.

#### Steps to reproduce
- Pick trait for getting winged flight.
- Pick some negative traits to balance it out (Which only makes this an even more bitter pill :S)
- Try to fly and do anything graceful.
- Be more clumsy than someone just awkwardly crawling over a table in a drunken stupor.

#### Code Revision
Server revision: release - 
ad18753b9d6bfd50b51846b726f025f1e781a75c 
 
#### Anything else you may wish to add:
I'm not sure if there are balance reasons, but at least make an option to do graceful, slow manouvers or something... I feel like a joke with wings on and I've basically needed to *stop* flying, move somewhere and *Start* flying again to pretend I can fly like an elegant Demoness.

Very-Soft:
While I'd like to see more done with flight, it's not really meant to be used indoors. You gotta keep in mind that you're flapping around with a wingspan larger than your character is tall realistically. You're gonna have very little room to do that indoors, especially where tables, walls, and doors are concerned. 

It's an intended thing for flying people to take tumbles when they fly into things that would impede their flight.

Flight in RP is much better done purely in RP, as opposed to doing it mechanics wise, as you'll run out of hunger pretty fast, and then fall down. Flight itself is much more really intended to allow for an extra movement option really. You can fly up and down outside, or in the atrium. This can allow for access to areas that only people who can fly can access, if you're willing to build said places.

You CAN fly in the more cramped interior places, but once again, you're flapping around with a large wingspan in a place designed primarily for normal sized humans to walk around, so care needs to be taken to not bump into things when you do so.

Or, you know, you could just walk.

KasparoVy:
Walking briskly while vigorously flapping your arms. Screeching all the while...

[...]

PR #4321, opened 2018-10-06
Adjust winged flight

lbnesquik:
**Moves winged flight into neutral from positive traits.**
This is done to give this trait better price quality, especially compared to other perks you could take for 2 points such as non conductivity or *darksight major*.

**Reduce it's cost by half.**
The change in category was motivated by how little use this perk actually has and the numbers of downsides such as, few areas worth using this perk in, little flight time or the magic ability to stumble into any impassable atom like tables or doors (even if you have access) and receive 5 whole brute for it. 
In it's current state, it is very likely the worst "positive" trait within the game.

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/traits_vr/neutral.dm
+++ b/code/modules/mob/living/carbon/human/species/station/traits_vr/neutral.dm
@@ -36,7 +36,7 @@
 /datum/trait/winged_flight
 	name = "Winged Flight"
 	desc = "Allows you to fly by using your wings."
-	cost = 2 //Some in game value.
+	cost = 1
 
 /datum/trait/winged_flight/apply(var/datum/species/S,var/mob/living/carbon/human/H)
 	..(S,H)
--- a/code/modules/mob/living/carbon/human/species/station/traits_vr/neutral.dm
+++ b/code/modules/mob/living/carbon/human/species/station/traits_vr/neutral.dm
@@ -33,6 +33,16 @@
 	var_changes = list("cold_discomfort_level" = T0C+21)
 	excludes = list(/datum/trait/cold_discomfort)
 
+/datum/trait/winged_flight
+	name = "Winged Flight"
+	desc = "Allows you to fly by using your wings."
+	cost = 2 //Some in game value.
+
+/datum/trait/winged_flight/apply(var/datum/species/S,var/mob/living/carbon/human/H)
+	..(S,H)
+	H.verbs |= /mob/living/proc/flying_toggle
+	H.verbs |= /mob/living/proc/start_wings_hovering
+
 /datum/trait/autohiss_unathi
 	name = "Autohiss (Unathi)"
 	desc = "You roll your S's and x's"
--- a/code/modules/mob/living/carbon/human/species/station/traits_vr/positive.dm
+++ b/code/modules/mob/living/carbon/human/species/station/traits_vr/positive.dm
@@ -94,16 +94,6 @@
 	cost = 1
 	var_changes = list("flash_mod" = 0.5)
 
-/datum/trait/winged_flight
-	name = "Winged Flight"
-	desc = "Allows you to fly by using your wings."
-	cost = 2 //Some in game value.
-
-/datum/trait/winged_flight/apply(var/datum/species/S,var/mob/living/carbon/human/H)
-	..(S,H)
-	H.verbs |= /mob/living/proc/flying_toggle
-	H.verbs |= /mob/living/proc/start_wings_hovering
-
 /datum/trait/hardfeet
 	name = "Hard Feet"
 	desc = "Makes your nice clawed, scaled, hooved, armored, or otherwise just awfully calloused feet immune to glass shards."


lbnesquik:
#3906

lbnesquik:
No idea why this is stalling. I blame map edges.

Arokha:
I think the problem with making this a neutral trait will mean that everyone will take it, 'just in case' they want to use it, just to have the verbs available, even if they can't really fly.

Very-Soft:
The fact is, even if it's not a great perk, it does give positive benefits. 

Being able to build areas that are only accessible to people who can fly, even if it's not used often, is a powerful, unique thing. I'm just saying. I don't think flight should be super expensive, but it's totally a positive trait.

Very-Soft:
Now, I'd be pretty okay with flight being able to be used to hop tables and railings, and dodge broken glass. It's a little silly that those things impede you. Or even if it raised your movespeed a bit for as long as you could keep flying.

Like, I think it'd be more neat to give it a small buff, rather than just removing the cost of it.

lbnesquik:
Code wise, flying is just icons moving, the ability to ignore micros and to ignore a check for moving up a z level. The rest is just the equivalent of drunk code. Whenever the player character encounters an atom they can't pass, they have the process `stumbleinto` trigger, making them fall and get hurt.
I also disagree with this perk being 'powerful'. Having areas availlable only for some winged peoples is nice but not powerful.

lbnesquik:
How to make winged flight deserve it's cost (or more) and the positive label:
- [ ] **Important:** Make it so you don't fly like a chicken on ketamine.
- [ ] **Important:** Reduce the nutrition cost.
- [ ] Consider making movement faster while flying, justifying it's usage within hallway.
- [ ] Consider changing the maps around to make the perk more rewarding.
- [ ] Consider making flying peoples more difficult to hit with weapons overall. It's hard to punch a bird after all.
If the important points are completed, it will be a below average and not just an unworkable gimmick.

Very-Soft:
I mean, I think if it's giving you pretty reasonable bonuses, like accessing new areas, going faster, or traversing over tables quickly, then the nutrition drain is pretty justified. You're exerting yourself more than normal to do something special. If it didn't have the extra nutrition cost, then everyone would just always fly all the time if it gave those bonuses.

lbnesquik:
It doesn't make you faster or allow you to cross tables right now.

KasparoVy:
Can't fly over anything aside from empty spaces from what I can tell.

lbnesquik:
				if(((confused || is_blind()) && stat == CONSCIOUS && prob(50) && m_intent=="run") || flying)
					AM.stumble_into(src)

This means that running into anything anchored will make you fall. While is a lot of things in a crowded space station.

lbnesquik:
I'd like to specify I am not planning on fixing the perk for the moment. Simply it's costs.

code/modules/mob/living/carbon/human/species/station/traits_vr/neutral.dm
```diff
@@ -33,6 +33,16 @@
 	var_changes = list("cold_discomfort_level" = T0C+21)
 	excludes = list(/datum/trait/cold_discomfort)
 
+/datum/trait/winged_flight
+	name = "Winged Flight"
+	desc = "Allows you to fly by using your wings."
+	cost = 1
```
Arokha: the traits system was not designed to allow costs on neutral traits. This cost does nothing. All neutral traits are free regardless of what you put here. If it has a cost, it's a positive trait. If it gives cost back, it's a negative trait.

Also just, nah imo. The limitations on flight were added due to discussion on the original flight PR. "Wait a year then remove them" isn't powercreep at all nuh uh.

Why don't you fix the actual problems you're complaining about instead of just "eh make it half the cost and not take up a trait slot" which is a lazyass way to improve quality of life. Improve the quality, don't make the cost of living poorly lower. ;x

Screemonster:
I'd rather see things like slipping and other ground effects check against "is this mob affected by things on the ground" and have flying affect that check.
This would allow for flying/hovering mobs (like birds and combat drones) that merrily drift over upturned tables and the like, for example. It'd also mean phased shadekin don't get "drenched from entering the pool" for every tile they move, and space lube and glass shards wouldn't affect people even when gravity is off. There are _plenty_ of use cases for such a check, but it requires delving into the pit that is movement code.

code/modules/mob/living/carbon/human/species/station/traits_vr/neutral.dm
lbnesquik: First off, correct, price reduction and neutral traits aren't exactly compatible. I'll move the trait back into positive as it should still cost something.
And then secondly, i am trying to fix the actual problems of the perk. Starting with the price balance which is the easiest and the most important one to to me. Then maybe i will give time to 'fix actual problems' so this perk is becomes good enough to be worth investing 2 points into it. An hack fix with strong potential for a true fix later on is much better than no fix at all.

lbnesquik:
@Screemonster I wasn't thinking of making a special all encompassing system for hovering mobs but i could perhaps give it a go in another PR sometime.

lbnesquik:
I'd like to add to the 'resolved review' comment. 
I am pretty sure the original PR discussion mostly focused on where this perk could get you to and how powerful it could be against security as with any movement enhancing perk. Lowering the perk cost changes nothing of that. The walls and the doors of the station would restrict movement all the exact same. 
The only difference is that you'll see more people consider using the perk instead of taking something like brute resistance or hardy for convenience. 

Perhaps there would be an issue with security if i made the follow up PR with the changes i feel are needed to make the perk interesting and valuable instead of health hazard for the user. But then again, merely not falling to the floor from hitting doors would be great.

PR #4324, opened 2018-10-06
Changes Security Processing

Mewchild:
Changes all windows in security processing to be tinted windows. This prevents prisoners in the brig from being able to eavesdrop on processing (as anybody in the lower level could hear every word from processing, previously).


Mewchild:
As per usual: [Active edge](https://travis-ci.org/VOREStation/VOREStation/jobs/438082482#L1076)

PR #4326, opened 2018-10-06
ᵗᵉᵉⁿʸ ᶠˡᵃˢʰˡᶦᵍʰᵗ ᵗʷᵉᵃᵏ

lbnesquik:
A small noise now plays when your flashlight runs out of power. I didn't find any noise online which fitted my desire of a small, hushed electric fizzling so i used spark3 on very low volume.
Edit:Added noises for adding and removing cells.

### DIFF
--- a/code/game/objects/items/devices/flashlight.dm
+++ b/code/game/objects/items/devices/flashlight.dm
@@ -68,7 +68,7 @@
 					cell.charge = 0
 					visible_message("<span class='warning'>\The [src] flickers before going dull.</span>")
 					set_light(0)
-					playsound(src.loc, 'sound/effects/sparks3.ogg', 10, 1, -3) //Small cue that your light went dull in your pocket.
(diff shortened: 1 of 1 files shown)


code/game/objects/items/devices/flashlight.dm
```diff
@@ -68,6 +68,7 @@
 					cell.charge = 0
 					visible_message("<span class='warning'>\The [src] flickers before going dull.</span>")
 					set_light(0)
+					playsound(src.loc, 'sound/effects/sparks3.ogg', 10, 1, -3) //Small cue that your light went dull in your pocket.
```
Mewchild: //Vorestation Edit

lbnesquik:
Active map edges again.

code/game/objects/items/devices/flashlight.dm
lbnesquik: Done. Thanks.

PR #4328, opened 2018-10-07
Dogborgs start with full water reserves.

lbnesquik:


lbnesquik:
Requested changes should be done.

PR #4331, opened 2018-10-08
[FEEDBACK NEEDED]Traits tweaks.

lbnesquik:
- Make Conductive 1 increase damage to 75% and not 50%
- Make major conductive increase damage to 150% and not 100%
- Add numbing fangs as a standalone trait. (Nah it's not too many traits.)
Everything up for debate.

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/traits_vr/negative.dm
+++ b/code/modules/mob/living/carbon/human/species/station/traits_vr/negative.dm
@@ -80,15 +80,15 @@
 
 /datum/trait/conductive
 	name = "Conductive"
-	desc = "Increases your susceptibility to electric shocks by 50%"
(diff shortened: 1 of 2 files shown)


CameronWoof:
This also increases the cost of Major Darksight from 2 to 3.

Can you explain your rationale for these changes?

KasparoVy:
RIP darksight

lbnesquik:
Very useful utility perk that helps in and out of combat.

Mewchild:
You need to do testing with the conductive traits, as those may allow individuals to be one or two shot by tasers, and may make electrical shocks lethal as a baseline. I also disagree that darksight (major) is as useful as you think it is, given the adjustment time needed between light and dark areas, and the fact that any source of light, at all, renders it useless.

PR #4336, opened 2018-10-09
Gives pounces a chance to fail

Screemonster:
Instead of pouncing being a flat "if you roll a 20 then KD the mob" chance, mobs now have a probability of success/failure which can be adjusted per mob, and unless capable of eating standing mobs, will not attempt to pounce targets until the target dips below a health threshold (80% of full by default).

The chance of a pounce succeeding is a linear slope based on the target mob's health. By default, this starts at 20% when the target reaches 80% health, increasing to 100% as the target drops towards zero. If the mob misses a pounce, it will not attempt another for 20 seconds.

Mobs that are capable of eating standing mobs will always succeed on their pounce if they decide to pounce, since they don't punch normally and would otherwise never bring targets below the health threshold to start pouncing so you'd be able to just juke them forever by stepping away when they try to eat you.

If the mob does attempt a pounce, then regardless of success or failure it will not attack normally this round. Mobs will not waste their attack on trying to pounce silicons, which can't be stunned.

Gives the hungry pets the standing_too thing to make them always pounce as before.

Edit: just for clarification when it comes to setting the vars if anyone wants to change them from the default values, here's the sort of probability slope you'd get for the chances of a pounce succeeding versus a potential target's health, with a bunch of values picked purely for example purposes:
image

Here's the slope for the current default values:
image

### DIFF
--- a/code/modules/mob/living/simple_animal/simple_animal_vr.dm
+++ b/code/modules/mob/living/simple_animal/simple_animal_vr.dm
@@ -12,6 +12,10 @@
 	var/vore_bump_chance = 0			// Chance of trying to eat anyone that bumps into them, regardless of hostility
 	var/vore_bump_emote	= "grabs hold of"				// Allow messages for bumpnom mobs to have a flavorful bumpnom
 	var/vore_pounce_chance = 5			// Chance of this mob knocking down an opponent
+	var/vore_pounce_cooldown = 0		// Cooldown timer - if it fails a pounce it won't pounce again for a while
+	var/vore_pounce_successrate	= 100	// Chance of a pounce succeeding against a theoretical 0-health opponent
+	var/vore_pounce_falloff = 1			// Success rate falloff per %health of target mob.
+	var/vore_pounce_maxhealth = 80		// Mob will not attempt to pounce targets above this %health
 	var/vore_standing_too = 0			// Can also eat non-stunned mobs
 	var/vore_ignores_undigestable = 1	// Refuse to eat mobs who are undigestable by the prefs toggle.
 	var/swallowsound = null				// What noise plays when you succeed in eating the mob.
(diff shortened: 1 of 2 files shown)


Screemonster:
out of scope for this 'cause I have no idea how savefiles work either, I've heard people say it's something to do with rust-g and it's affecting all voreprefs.

code/modules/mob/living/simple_animal/simple_animal_vr.dm
```diff
@@ -92,21 +96,39 @@
...
-	// If target is standing we might pounce and eat them
-	if(target_mob.canmove && prob(vore_pounce_chance))
-		target_mob.Weaken(5)
-		target_mob.visible_message("<span class='danger'>\the [src] pounces on \the [target_mob]!</span>!")
-
-	// If they're down or we can eat standing, do it
-	if(!target_mob.canmove || vore_standing_too)
+	// If target is standing we might pounce and knock them down instead of attacking
+	if(target_mob.canmove && prob(vore_pounce_chance) && !issilicon(target_mob) && (world.time > vore_pounce_cooldown)) //pouncing is worth doing, we want to pounce and are not in cooldown. So attempt a pounce!
+		if(will_eat(target_mob) && vore_standing_too) //creatures that can eat you on the spot don't care how healthy you are and get an autohit
```
Arokha: *scream

Make a can_pounce_target() or something return all this garbage. This if-chain is too scary. Maybe have it return false if the chance is real low. (Less than 20?)

code/modules/mob/living/simple_animal/simple_animal_vr.dm
```diff
@@ -92,21 +96,39 @@
...
+
+	// We're not attempting a pounce, if they're down or we can eat standing, do it as long as they're edible. Otherwise, hit normally.
+	if(will_eat(target_mob) && (!target_mob.canmove || vore_standing_too))
 		return EatTarget()
 	else
 		return ..()
 
+/mob/living/simple_animal/proc/PounceTarget()
+	target_mob.Weaken(5)
+	target_mob.visible_message("<span class='danger'>\the [src] pounces on \the [target_mob]!</span>!")
```
Arokha: Cooldown in here. Where you actually pounce. Otherwise you might get:
a) the cooldown getting set when you don't actually pounce
b) pounces where you don't set the cooldown like you have above
It will be more consistent here.

code/modules/mob/living/simple_animal/simple_animal_vr.dm
```diff
@@ -92,21 +96,39 @@
...
+			return PounceTarget()
+
+		var/TargetHealthPercent = (target_mob.health/target_mob.maxHealth)*100
+		if (TargetHealthPercent <= vore_pounce_maxhealth) //check they're weak enough to bother pouncing
+			if(prob(vore_pounce_successrate - (vore_pounce_falloff * TargetHealthPercent))) // pounce success!
+				return PounceTarget()
+			else // pounce misses!
+				target_mob.visible_message("<span class='danger'>\the [src] attempts to pounce \the [target_mob] but misses!</span>!")
+				playsound(loc, 'sound/weapons/punchmiss.ogg', 25, 1, -1)
+				vore_pounce_cooldown = world.time + 20 SECONDS // don't attempt another pounce for a while
```
Arokha: Don't cooldown here...

Screemonster:
Made some changes.
Now PounceTarget is its own thing that always sets the pounce cooldown regardless of success or failure.
Made a new proc for CanPounceTarget which returns false for things that automatically make pouncing pointless, such as the target being a silicon and therefore unstunnable, or already weakened therefore no point in pouncing again. This also means that if you make a mob with silly values for the success rate or falloff, it won't attempt to pounce things with literally zero chance of success.

code/modules/mob/living/simple_animal/simple_animal_vr.dm
```diff
@@ -92,20 +96,48 @@
...
-	// If target is standing we might pounce and eat them
-	if(target_mob.canmove && prob(vore_pounce_chance))
+/mob/living/simple_animal/proc/CanPounceTarget() //returns either FALSE or a %chance of success
+	if(!target_mob.canmove || issilicon(target_mob) || world.time < vore_pounce_cooldown) //eliminate situations where pouncing CANNOT happen
+		return FALSE
+	if(!prob(vore_pounce_chance)) //mob doesn't want to pounce
+		return FALSE
+	if(will_eat(target_mob) && vore_standing_too) //100% chance of hitting people we can eat on the spot
+		return 100
+	var/TargetHealthPercent = (target_mob.health/target_mob.maxHealth)*100 //now we start looking at the target itself
```
Arokha: There are procs to get max and current health, that take into account things like modifiers. Important to use those and not the vars to take those into account if they're buffed by whatever.

PR #4340, opened 2018-10-09
Exploration Fixes for 09-27-2018 Sync

Novacat:
- Removes redundant station_vr file
- Reverts exploration suit cyclers to polaris cyclers
- Adds a pilot suit cycler to exploration
- Applies a functioning fix for holdout phasers
- Buffs up the holdout phaser to be 50% of expedition phaser, to keep size-effectiveness scaling (Holdout phaser is half the size, half the effectiveness)
- Removes head access from expedition shuttle control console
- Removes Pilot Blue/Old Exploration from polaris cyclers (they are still WIP on polaris end)


maps/southern_cross/structures/closets/misc_vr.dm
```diff
@@ -7,7 +7,7 @@
 	icon_opened = "secureexpopen"
 	icon_broken = "secureexpbroken"
 	icon_off = "secureexpoff"
-	req_access = list(access_explorer)
```
Arokha: Does pathfinder not have an access level...?

Novacat: Not their own, no. I never added one.

Arokha:
@SpadesNeil thoughts on gun changes?

PR #4345, opened 2018-10-09
Medihound Hypo equalized with sleeper

Novacat:
The mediborg dog sleeper is no longer a direct upgrade to the mediborg hypo, this is in response to the prefs issue involving the medihound sleeper.

### DIFF
--- a/code/modules/mob/living/silicon/robot/dogborg/dog_sleeper_vr.dm
+++ b/code/modules/mob/living/silicon/robot/dogborg/dog_sleeper_vr.dm
@@ -11,7 +11,7 @@
 	var/min_health = -100
 	var/cleaning = 0
 	var/patient_laststat = null
-	var/list/injection_chems = list("inaprovaline", "dexalin", "bicaridine", "kelotane", "anti_toxin", "spaceacillin", "paracetamol") //The borg is able to heal every damage type. As a nerf, they use 750 charge per injection.
+	var/list/injection_chems = list("tricordrazine", "inaprovaline", "anti_toxin", "tramadol", "dexalin" ,"spaceacillin") //The borg is able to heal every damage type. As a nerf, they use 750 charge per injection.
 	var/eject_port = "ingestion"
 	var/list/items_preserved = list()
 	var/UI_open = FALSE
--- a/code/modules/mob/living/silicon/robot/dogborg/dog_sleeper_vr.dm
+++ b/code/modules/mob/living/silicon/robot/dogborg/dog_sleeper_vr.dm
@@ -11,7 +11,7 @@
 	var/min_health = -100
 	var/cleaning = 0
 	var/patient_laststat = null
-	var/list/injection_chems = list("inaprovaline", "dexalin", "bicaridine", "kelotane","anti_toxin", "alkysine", "imidazoline", "spaceacillin", "paracetamol") //The borg is able to heal every damage type. As a nerf, they use 750 charge per injection.
+	var/list/injection_chems = list("inaprovaline", "dexalin", "bicaridine", "kelotane", "anti_toxin", "spaceacillin", "paracetamol") //The borg is able to heal every damage type. As a nerf, they use 750 charge per injection.
 	var/eject_port = "ingestion"
 	var/list/items_preserved = list()
 	var/UI_open = FALSE
--- a/code/modules/mob/living/silicon/robot/dogborg/dog_sleeper_vr.dm
+++ b/code/modules/mob/living/silicon/robot/dogborg/dog_sleeper_vr.dm
@@ -388,7 +388,7 @@
 				to_chat(hound, "<span class='notice'>Your stomach is currently too full of fluids to secrete more fluids of this kind.</span>")
 			else if(patient.reagents.get_reagent_amount(chem) + 10 <= 20) //No overdoses for you
 				patient.reagents.add_reagent(chem, inject_amount)
-				drain(750) //-750 charge per injection
+				drain(50) //-50 charge per injection
 			var/units = round(patient.reagents.get_reagent_amount(chem))
 			to_chat(hound, "<span class='notice'>Injecting [units] unit\s of [chemical_reagents_list[chem]] into occupant.</span>") //If they were immersed, the reagents wouldn't leave with them.
 
--- a/code/modules/mob/living/silicon/robot/dogborg/dog_modules_vr.dm
+++ b/code/modules/mob/living/silicon/robot/dogborg/dog_modules_vr.dm
@@ -150,7 +150,6 @@
 	name = "MediHound hypospray"
 	desc = "An advanced chemical synthesizer and injection system utilizing carrier's reserves, designed for heavy-duty medical equipment."
 	charge_cost = 10
-	reagent_ids = list("inaprovaline", "dexalin", "bicaridine", "kelotane", "anti_toxin", "spaceacillin", "paracetamol")
 	var/datum/matter_synth/water = null
 
 /obj/item/weapon/reagent_containers/borghypo/hound/process() //Recharges in smaller steps and uses the water reserves as well.
--- a/code/modules/mob/living/silicon/robot/dogborg/dog_modules_vr.dm
+++ b/code/modules/mob/living/silicon/robot/dogborg/dog_modules_vr.dm
@@ -150,7 +150,7 @@
 	name = "MediHound hypospray"
 	desc = "An advanced chemical synthesizer and injection system utilizing carrier's reserves, designed for heavy-duty medical equipment."
 	charge_cost = 10
-	reagent_ids = list("inaprovaline", "dexalin", "bicaridine", "kelotane","anti_toxin", "alkysine", "imidazoline", "spaceacillin", "paracetamol")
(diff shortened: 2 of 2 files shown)


Arokha:
TBH I'd rather reduce the sleeper chems than add the hypo chems. Sorta like I prefer if paramedics don't just carry alkysine and stuff because that's not their job, it's sorta not the dogborg's job to fix brain damage. Don't really want them being paramedic and MD and nurse and ...

Heroman3003:
Medihound's sleeper's main tradeoff is that its a sleeper. Adding those chems to the hypospray will make it way too strong. And it doesn't seem like too much of a pref issue in the first place. If anything, best way to fix the 'issue' would be to give them a single-bed roller-rack, but even then nonvores will REEE at the medihounds when they don't use the rollerbed, even with patients who are okay with it.

Wickedtemp:
It is kind of a prefs issue. Some players are into vore from the Pred side but hate the idea of being prey, and the sleeper belly makes a bit of a conflict with that. 

Making the hypospray more useful will cut down on that. It wouldn't need alkysine or anything, but.. Bicaridine, kelotane, dexalin and dylovene, and tada, now there's an alternative that doesn't result any conflicts. Now pred-only people don't have to get subpar treatment by synth paramedics.

Heroman3003:
Right now, no other borg has hypo chem dedicated to burn treatment or brute treatment. Same goes for imidazoline and alkysine, those are higher tier fixers.

Arokha:
I think we'll just talk about what direction we want to go first before doing this. I do agree it's a little discriminatory but not sure this is great (or, really, "infinite hypospray" is great in general, same with sleeper, when perhaps a thing with more character/theme could be used. *Tongue Patches*)

Mewchild:
I'm not a huge fan of this, in general, as it makes the medical dogborg even more of a direct upgrade over the standard cyborg modules. I feel that the medi-hound should be on par with the non-canine medical cyborgs, as there's no real reason to ever play them, as it stands.

Wickedtemp:
Its not really an upgrade to how it is currently. It just allows them to use their sleeper chems without vore.

lbnesquik:
Buff standard cyborgs.

Wickedtemp:
The only reason this is a thing is because, currently, the canine borg has to eat you in order to actually be able to use those reagents.

There's a decent handful of players here that have issues with that, primarily due to being pred-only or just not liking oral vore. 

So, taking the reagents that it already has and giving it a new non-OV-related way to administer them should give them a better alternative than either not getting help and potentially dying, or dealing with a rather OOCly uncomfortable healing.

Screemonster:
Okay, so basically as I'm reading it...

medical cyborgs were a thing already
dogborgs were added, and their sleepers have chems available that make them objectively better at administering treatment because they have chems that normal cyborgs do not, but some players are penalised if they don't want to be prey because they can't access this full range of chems
so the solution being presented here is to... buff the borg's hypo to match the sleeper?

The vore method being more powerful than the standard method isn't an argument to buff the standard method, it's an argument to bring the vore method into line with the rest of the module.

Wickedtemp:
I think it already had Cillin. All it needs to be able to handle patients without the use of vore would be:
Bicaridine, Kelotane, Dylovene, Dexalin, Spaceacillin. Add Paracetamol/Tramadol for a painkiller, and tada, done.

It's not really a direct buff or upgrade, these are reagents that the borg can use and are always readily available anyway, this just lets them administer these reagents to people where prefbreaking comes into play with oral vore/being prey.

Novacat:
Threw up an edit that removes alkysine and imidazoline from both the sleeper belly and the dogborg hypo, based on current discussion/agreement so far.

Screemonster:
Temp, the borgs were buffed/upgraded by giving the sleepers chems that normal mediborgs don't have available in the first place. Giving them those chems in their hypo because "oh their sleepers have them" doesn't stop the fact that the dogborgs have been buffed above the level of baseline borgs, and buffing everything else to match a buffed thing is pretty much the definition of powercreep.

nerdass:
dogborgs shouldn't even have functional sleepers, just make it a belly, being able to transport injured players to medbay without dragging them across the floor without external assistance is powerful enough on its own.

nerdass:
like a single dogborg is basically a fully kitted Odysseus mech right now.

Screemonster:
yeah, people in another PR mentioned that nobody ever uses the odysseus - the mech that's meant to have sleepers on board for quickly transporting critical patients. Why would they, when the medihound has a built in sleeper that's superior to the mech-mounted sleeper in every way?

Poojawa:
Medihounds, from what I've gathered, are basically the 'oh that person is dying to exposure, let's put them in a safe spot' paramedics. 

Comparing their usage to a bulky mech isn't really fair, because the mech can synth almost any reagent it scans. As well as mount syringe guns for tactical ranged healing.

Reduce the sleeper chems to Triage level 'we're keeping alive till you're in medbay' and call it good. No need to fluff anymore chems than that. Hypospray should be the same triage-only

Wickedtemp:
If the borgs were the reason we didn't use the Ody's, we'd order them when the borgs aren't ingame. The borgs have nothing to do with the Ody collecting dust, the mechs are just outright bad here. They have zero healing capability. Their sleeper unit is useless, as far as I can tell, unless you take a syringe gun, because you need it to analyze reagents from the chem whitelist to inject anything other than Inaprovaline and Dylovene with the Sleepers.

The syringe gun itself has uses from an OOC standpoint, delivering a 15u syringe from a distance if it hits the target, but the issue is ICly, you just turned your patient into a pin-cushion by *shooting syringes into their skin and flesh*. It's an IC risk of legal consequences every time you use it this way, and even then the usage is limited by a 75u total storage shared by both the syringe gun AND the sleeper units, which means you have roughly a 1.25x beaker, just the one, to share and mix all of your stored reagents, across all sleepers and syringe guns. It's clunky to move around in - all mechs are - and the speed itself, from what I can remlember, isn't a whole lot better unless you happen to be going in a single straight line due to having to stop completely to turn corners and having to eject out of the mech in order to use any buttons - like the elevator for example. On top of that, it can't maneuver through crowds due to bumping and shoving. The only benefit is being able to transport 2 people if you have 2 sleepers, the armor, and the internals. And anyone can do that in with a voidsuit and vore mechanics anyway, without any of the drawbacks. From my experience, an EMT in a voidsuit with a medical belt and a proper loadout is more useful than the Ody. As a result, I just don't use them. Not because of the borgs. I didn't even use the Ody's for medical before the dog-borgs were added. But, this is all for a different git discussion.

code/modules/mob/living/silicon/robot/dogborg/dog_sleeper_vr.dm
```diff
@@ -388,7 +388,7 @@
 				to_chat(hound, "<span class='notice'>Your stomach is currently too full of fluids to secrete more fluids of this kind.</span>")
 			else if(patient.reagents.get_reagent_amount(chem) + 10 <= 20) //No overdoses for you
 				patient.reagents.add_reagent(chem, inject_amount)
-				drain(750) //-750 charge per injection
+				drain(50) //-50 charge per injection
```
Arokha: This isn't actually quite right. The 50 charge on the borg hypo is actually removing 'charges' from the matter_synthesizer, which I started following exactly how it gets power from the cell but then lost it because it's late. I'll look at it tomorrow.

Novacat:
Borgs are not supposed to be going on expeditions in the first place without good reason.

Mewchild:
That's not stated anywhere in SoP, or in the server rules.

Novacat:
I believe the general idea is that borgs are supposed to be serving on station, and mostly came up a few months back when every borg on station was basically jumping to volunteer on every expedition.

Arokha:
Borgs following their laws would probably have trouble justifying expeditions unless ordered to go (but even then...) because they're putting themselves in danger for law 3 and leaving the station so they will be totally unable to protect it for law 2. They can protect people, law 1. Though breaking 2 laws and following 1 doesn't seem very 'economical' in terms of attempting to follow the laws.

I'd let @SpadesNeil actually give a more concrete ruling on that though.

Arokha:
> water cost

What water cost @Cameron653

CameronWoof:
@Arokha MediHounds use water from their internal tanks in addition to power to fuel their hypospray and sleeper chem injections. It was an addition made by Verkister in a relatively recent PR. 

We're currently looking at removing the water cost for the hypospray downstream, but if the sleeper were nerfed as described in this PR, we'd probably remove it for that, too. It's kind of a hose and just makes them more annoying to use than standard medical cyborgs.

Arokha:
IMO just copy the normal standard synth matter_synth and put it in these borgs, and have their sleepers and hypos use it just like on the normal borgs. I'm not 100% sure what the usage ratio is, for the hypo thing. How many units of drug does it inject for that '50 matter synth charge'? Make sure the ratio is the same for the sleeper so they're balanced and nobody accuses of preferential treatment.

Novacat:
...A combination of personal reasons, as well as general lack of motivation to put a huge amount of work on this PR, and the increasing scope of this PR, I have made the decision to close it as I am not able to finish this PR as requested.

PR #4350, opened 2018-10-10
Micro & mech tweaks.

lbnesquik:
- Add plastic to the robotics locker
- Restrict micros to micromechs. (Anyone below 50%)

VOREStation/VOREStation/issues/4332

### DIFF
--- a/code/game/mecha/mecha.dm
+++ b/code/game/mecha/mecha.dm
@@ -1117,6 +1117,11 @@
 		to_chat(usr,"<span class='warning'>You can't climb into the exosuit while buckled!</span>")
 		return
 
+	var/mob/living/carbon/B = usr //Vorestation edit.
+	if (B.size_multiplier < 0.5)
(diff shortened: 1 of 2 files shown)


Screemonster:
Ehhhhhh, when I added the micromechs it was more for "micros can use these, they're cheaper and less obtrusive/costly than regular mechs" than "micros can only ever use these mechs".

Suits such as the ody still don't have micro equivalents.

lbnesquik:
Would it be fine if there was a fuller range of comparable exosuits?

Wickedtemp:
Mechs in general need a rework. I know, its horribly out of scope for this, but... I never use or order Ody mechs anymore. When Medical rarely DOES get one, it collects dust and will really never be used. Nearly everything it provides can be topped by an EMT in a voidsuit or rig.

KasparoVy:
cuz strong spesmen can drag tons of heft with no slowdown

lbnesquik:
There is a mech rework coming actually. Just got to wait for the sync.

Issue #4158, opened 2018-08-11
Suggestion/Request: Digestion-Resistant Promethean Core


Vorrarkul:
#### Brief Description of the Request
I would like for promethean cores to not be destroyed by digestion. Whether that means simply making all promethean cores indigestible or creating a prosthetic-organ variant that represents a shielded/treated core, I would just like some way to recover and reform prometheans post-scene outside of resleeving.

After some discussion on discord and queries about where to take this, it was suggested to make a github issue about it, so here I am.

Zeldazackman:
I'd love this, especially while playing as a vox it becomes a bit of a "ah crap" moment when you digest a promie only to have their core become useless the moment their gooey congealed forms finally decides to give in to their new place of residence.

I find that even with the itemweak stomach mode toggled on, it oftentimes just removes their core outright which makes scenes really awkward and limiting as a whole.

PR #4351, opened 2018-10-10
Makes promethean cores indigestible

Screemonster:
as per #4158 - promethean cores now survive where the rest of them does not.

Also made stomachs delete cleanable decals 'cause otherwise eating a prommie would fill your stomach with gibs the instant their limbs popped.

### DIFF
--- a/code/modules/vore/eating/bellymodes_vr.dm
+++ b/code/modules/vore/eating/bellymodes_vr.dm
@@ -85,6 +85,9 @@
 								digest_item(I)
 							to_update = TRUE
 							break
+		//get rid of things like blood drops and gibs that end up in there
+		else if(istype(A,/obj/effect/decal/cleanable/))
+			qdel(A)
 
 ///////////////////////////// DM_HOLD /////////////////////////////
 	if(digest_mode == DM_HOLD)
--- a/code/modules/vore/eating/digest_act_vr.dm
+++ b/code/modules/vore/eating/digest_act_vr.dm
@@ -58,6 +58,8 @@
 	return FALSE
 /obj/item/device/perfect_tele_beacon/digest_act(var/atom/movable/item_storage = null)
 	return FALSE //Sorta important to not digest your own beacons.
+/obj/item/organ/internal/brain/slime/digest_act(var/atom/movable/item_storage = null)
+	return FALSE //so prometheans can be recovered
 
 /////////////
 // Some special treatment


Screemonster:
During testing I also found the cause for prommie cores having no icon. Gonna push a PR upstream to fix.

code/modules/vore/eating/bellymodes_vr.dm
```diff
@@ -85,6 +85,9 @@
 								digest_item(I)
 							to_update = TRUE
 							break
+		//get rid of things like blood drops and gibs that end up in there
+		else if(istype(A,/obj/effect/decal/cleanable/))
```
Arokha: Ah, this fixes protean dust building up in there too

code/modules/vore/eating/digest_act_vr.dm
```diff
@@ -58,6 +58,8 @@
 	return FALSE
 /obj/item/device/perfect_tele_beacon/digest_act(var/atom/movable/item_storage = null)
 	return FALSE //Sorta important to not digest your own beacons.
+/obj/item/organ/internal/brain/slime/digest_act(var/atom/movable/item_storage = null)
+	return FALSE //so prometheans can be recovered
```
Arokha: Kinda wish this were a toggle or something. I know in the past when things like posibrains have come up, there have been two distinct sides to the debate: those that wanna gurgle as a synth and those that don't wanna. That's do-able for synths by picking different brain types...

Screemonster: A toggle'd be good but I'm not sure what a good toggle would be. In order to end up as a core in a stomach in the first place you'd need to have digestion turned on.

Arokha: Yeah but it removes the ability for prometheans to be fully digested. Like if someone said they were gonna do that with FBPs I'd bap them because it's bad to just decide for everyone that you can't gurg anymore.

I would suggest the following more comprehensive solution to "people that can be sorta digested but not fully digested":
1) Make a new trait (some workshopping on a name and description may be required) that in the apply() searches for the 'brain stuff' you want to gurgle, or not
2) Add a var on the 'brain stuff' for digestable
3) Have digest_act do a thing based on that var (or return false or whatever it does)

The thing is I'm not sure what needs to be done for MMIs for this because mmi/posi/digital brains are:
Outside) An /obj/item/organ/mmi_holder thing that holds a
Layer 1) /obj/item/weapon/whatever/mmi actual mmi object that holds a
Layer 2) /mob/living/carbon/brain that the player actually inhabits when necessary

I want to say there's a special thing in the digestion code somewhere that allows the deletion of the organ mmi holder and moves the posibrain up a level? So pay attention to that.

Screemonster: So far, all I can find relating to that is posibrains (a child of mmis) return false on the proc, and a comment in there indicating that there's a TODO.

Arokha: That's fine. That object that gets taken and put into guts safely is the one to add a new var on, along with promethean cores.

ArgobargSoup:
We have a thing that lets you choose which brain type you want as a synth, right?

If there was something like that for prommie cores... could do a `/obj/item/organ/internal/brain/slime/indigestible` subtype and make it a choice in character creation.

Arokha:
> If there was something like that for prommie cores

Slightly more difficult. The choices are very fixed:
Normal
Assisted
Mechanical
Digital

Those are the brain choices, and it's not based on 'what are the options for synths'. Having anything other than those might also break resleeving... it's a bit of a kludgey system, the '.robotize()' and '.mechassist()' or whatever all the procs are.

SpadesNeil:
This is good, right? It's kind of been sitting here.

Screemonster:
I'm kinda drawing a blank on the "okay how to make this optional in case a prommie player DOES want to have their core digested" part and then also apply it to posibrains which wasn't originally part of what I was gonna do but whatever option was chosen would require the same solution to be applied since they're both a case of "prey gets effectively debrained upon death and put in a brain mob inside the core/posibrain container"

SpadesNeil:
This is good for now.

PR #4352, opened 2018-10-10
gives prommie cores an icon

Screemonster:
There's no "core" icon state in our current surgery.dmi, causing promethean cores to have no icon.
The upstream version of surgery.dmi has an iconstate for it, but it's literally just a brain rather than a slime core, which doesn't seem to fit how people here seem to play prommies (going by the reference images most use). 
Created a new icon file and pointed promethean cores at it. Icons used are the same as a slime core. Created an "assisted" version too with a box and wires coming out of it since that seems to be a thing for brains, even upstream promethean cores.

edit: the bug this fixes doesn't exist in the current release branch, only in master.


Screemonster:
Looks like surgery.dmi diverged when #4207 was added upstream but not merged here.

Arokha:
Assisted cores are a thing? AAAA cute

PR #4359, opened 2018-10-12
Travis tweaks

Arokha:
You might make the point that this is just me giving up on fixing this problem for the 10th time over the last few years.
You might have a good point!


Arokha:
That's unusual. Even without spawning a single POI, it failed on that one.

Arokha:
@Leshana henlp

PR #4360, opened 2018-10-12
Adjust winged flight cost and description.

lbnesquik:
I fucked up trying to improve the number of commits by squashing but, made a mess instead. So here. New PR. Same content.

Previous link:
https://github.com/VOREStation/VOREStation/pull/4321


Arokha:
This is slightly less meh to me than moving it to neutral but still slightly meh. Can leave it here for a bit to get opines.

Screemonster:
Slightly less meh. Making the trait actually useful so it's worth the cost is preferable, but also effort.

lbnesquik:
I am thinking of improving the perk overall but rn i'd like to have this pass through

SpadesNeil:
I'm okay with this.

PR #4363, opened 2018-10-12
Fixes weird typos.

lbnesquik:
Shadekin had weird � signs instead of '


Arokha:
Those were smartquotes. The person who typed these probably did so in Word (or Wordpad if it does that too).

PR #4366, opened 2018-10-13
Updates admin upload areas to not always be powered

Very-Soft:
The admin upload areas all seem to always power everything in them. This should make the ones that are supposed to use power as per normal do so.


Very-Soft:
I accidentally deleted the branch. :Y

PR #4369, opened 2018-10-13
[MIRROR] Fixes maploading

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/5660
--------------------
Imagine you're on the last iteration of this loop, you've done the final column of X coordinates and you're going to set maxx.
Would you want to set it to where you are now, the final X column... or would you want to add ONE MORE beyond the template size for some reason then set it to that?

This bug causes all templates that are the size of normal maps to fail to initialize any atoms if your template is the size of your normal maps, because it tries to obtain a square of size minx, miny, minz, maxx, maxy, maxz to initialize, however maxx is nonextant because it exists outside the bounds of the world. It also causes all submaps to initialize an additional column of atoms twice, or not initialize any if they spawn against the right edge of the map.


Arokha:
Merged here earlier.

PR #4387, opened 2018-10-14
Fixes apostrophes in names causing javascript errors in the character directory

### LINKED ISSUE #4383: Character Directory (possibly) doesn't work when character has an apostrophe in name
Can't see a character's flavour text and OOC notes in the character directory.

I expected to click their flavour text and OOC notes and see their notes show up.

#### What actually happened
The buttons didn't seem to do anything. I ahelped, thinking it was that they didn't have notes. Was told they actually did, so it confirmed it was a bug.

#### Steps to reproduce
(Theoretical, not 100% sure)
1. Open character directory.
2. Click on the "flavour text" or "OOC notes" buttons for a character with an apostrophe in their name.
3. Nothing happens.

#### Code Revision
1fe04192c11e2ad7357512d5a873c44a4ea687d3
 
Character's name is "Ka'Yirii Trasix".

Not 100% sure if it actually involves the apostrophe, but the only character who had this bug had an apostrophe in their name.

deathride58:
Title. Fixes #4383

### DIFF
--- a/code/game/verbs/character_directory.dm
+++ b/code/game/verbs/character_directory.dm
@@ -23,11 +23,11 @@
 			html += "<div class='block'>"
 			html += "<h3 class='uiContent highlight' style='font-size:16px'>[H.real_name]</h3><br>"
 			if(H.flavor_texts["general"])
-				html += "<a onclick='togglesection(\"[H.real_name] [curID] flavor\")'>Flavor text</a>"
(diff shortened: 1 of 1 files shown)


Arokha:
This will cause all names to be displayed as lowercase, since ckeys are all lowercase.

You should html encode it instead.

deathride58:
Nah, this only affects the internal IDs that aren't shown to players. The rendered bit is completely untouched

Arokha:
ic ic. you should probably still htmlencode it since this is actually the very purpose of html encoding, like, the reason it was invented. I won't push on it tho.

PR #4400, opened 2018-10-15
[MIRROR] Ports Technomancer/Blob/Mecha Mobs to New System

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/5672
--------------------
Ever so slightly do we approach the end of ``/mob/living/simple_animal``.

What this PR does;
- Adds mecha simple_mobs. Most of them are reasonably tough and might be useful for PoIs.
  - Combat mechas in melee range will knock back most mobs when successfully hitting them, with a satisfying smack sound.
  - Some types can be set to be 'piloted' by a mob that starts existing when the mecha dies. Doing this also changes the mecha's idle ramblings and makes it vulnerable to things that affect sapient mobs. Otherwise it's assumed a drone is piloting the mecha.
  - These mechas drop mecha loot_piles on death.
  - The health/armor/attack power/etc of these tries to remain true to the actual mecha objs.
  - Attacks can randomly fail against these due to being deflected, like the real thing. A sprite appears when this occurs.
  - Some mechas are more unusual than the others;
    - Gygaxes have an aggressive AI, that makes them move towards their target, potantially into punch range.
    - Oddys kite their target and fire fake poison syringes to widdle down the target to death (hopefully they're not fighting a mechanical mob).
    - Hoverpods can fly, and they have a ranged laser attack.
    - Advance Dark Gygax is a new mob that has three special attacks and is the metaphorical flagship of this PR. Not quite megafauna (thats for later) but still a worthy challenge. Here's a video of it in action (https://streamable.com/pkwpf)
      - First special attack fires three arcing missiles at the target, telegraphed with several sounds and a harmless beam following the target.
      - Second special attack fires an arcing projectile that creates a 'micro-singularity' when landing, which pulls in everything it can nearby and irradiates them if able to for a few seconds. This may also rip out weapons from their hands.
      - Third special attack creates an energy orb that circles around the mecha, shocking anything nearby. Eventually, the orb explodes, causing a tesla shock to shoot out at anything nearby, and for anyone looking towards the mecha without eye protection to get flashed briefly. The mecha is protected from shocks while this attack is occuring.
- Ports blob mobs to new system.
  - Blob spores are scarier in a group now, as when one attacks, adjacent friendly spores contribute half of their attack power each, causing the attacking spore to hit harder if surrounded by other spores. Most spores do 1 - 2 damage each, but this can add up if there's 6+ spores attacking you, or if an infesting spore snagged someone who died.
- Ports the GOLEM to new mob code, complete with attack animations. The AI can't use spells right now but it can still melee and look cool doing it.
  - GOLEM animations don't use overlays anymore since the new overlay system didn't like them and the code was just awful.
- Ports technomancer wards (renamed to monitor ward). They are under a new ward class, pending future additions to new types of wards.
  - All monitor wards now decloak hostile enemies and give detailed information, without needing a scepter.
- Ports technomancer illusions. They're much lighter in code than the current version since the special pathfinding needs for illusions can be fulfilled by just changing its ``ai_holder_type``.
  - Makes tooltips for illusions bring up the tooltip for the object being copied, if one exists, instead of outing itself immediately if moused over.
- Adds new type of eviscerator, that will **never** attack a mercenary antag, or a raider antag (two separate types, make sure to choose the right one) under any circumstances. Also adds a grenade which makes these new mobs. Not mapped in yet.

Misc changes:
- Fixes a bug with flashing blindness (the white flash effect) being inconsistent. This was due to the effect used being an animation, and there was no guarantee that the animation would play from the beginning when applied. It now uses a static sprite and fades out using opacity.
- Refactors several pulsing ``/obj/effect/temporary_effect``s to be under a new class instead of riding off of Pulsar.
- Adds ability for simple_mobs to shake the ground when walking, for when fighting huge monsters.
- Adds new loot_piles for most of the mechs. Loot inherits from the base mecha loot pile since I'm not sure what to give them right now.
- Adds a poisoned modifier, which works similarly to the on fire modifier, is a neat way to abstract away reagent processing in simple_mobs. Dylovene will gradually remove stacks of it.
  - Spiders are 50% resistant to poison damage.
  - Mechanical mobs are immune to poison damage.
  - Constructs are immune to posion damage.
- Fixes bug which made shock attacks do no damage to anyone unprotected and do full damage to things immune to electricity. Yay for forgetting to invert a value.
- Mechanical mobs are 50% vulnerable to shock damage now.
- Things using ``singularity_pull()`` can now rip up floor tiles randomly.
- Tesla shocks respect siemens/shock protection.
- Mercs shouldn't just sit there and take it when threatening to shoot someone while getting attacked.


Neerti:
This doesn't go into the master branch on Polaris, instead going to the AI rework branch, so probably best to ignore this.

Arokha:
Non-master merge, though it looks neat.

PR #4405, opened 2018-10-16
Add buffer area against railroading

Arokha:
On maps that have atmos border railroading and also spawn POIs, sometimes POIs spawn touching the railroading and create roundstart active edges that annoy Travis. One solution is to stop POIs from spawning during unit testing, but I think it'd be nice to still spawn them, in case they cause other issues that can be discovered.

Instead, this just extends the 'explored' area from the starts of these two maps all the way around the edge for one turf (this area type doesn't spawn POIs), so now POIs can't spawn touching the railroading.

This was problem 2 of 2 that was causing the failed Travis builds. The other was already fixed by defining the proper base_turfs on the areas.


Arokha:
Weird. That still allowed a POI to spawn there. Though this may be related to the crazy 'POIs are a straight line' spawning problem.

PR #4409, opened 2018-10-16
Shelter Capsule Content Tweaks

Novacat:
- Adds some more synth batteries to both capsules
- Adds a dylovene pill bottle and a health analyzer to the luxury one
- Removes the violin due to being a source of grief
- Changes the bag of dice to a bag of gaming dice in the luxury capsule
- Luxury capsule now has twice the amount of MREs
- Changed the code so that the fridge's contents are in the map file
- Removed empty variable since it is not used
- Max number of items changed to 100

The main reasons for this PR:
- Removes shelter capsules as a source of violins, which have been involved in a few... incidents
- Equals the amount of batteries to the amount of liquid foods in the basic capsule, so that synth crew are not required to push for the luxury capsule and its recharger
- Adds a dylovene pill bottle to shelter capsules, since toxins are a bigger issue in the phoron-based atmosphere of V3B
- Provides a health analyzer to the shelter capsule, the omission of which was very strange.


Novacat:
The basic first aid kit is an addition, and is where the health analyzer is added through, previously luxury shelter capsules only had a single advanced first aid kit, which does not have a health analyzer.

SpadesNeil:
ohhh

code/modules/mining/shelter_atoms.dm
```diff
@@ -219,15 +220,27 @@
 	for(var/i in 1 to 5)
 		var/obj/item/weapon/reagent_containers/food/snacks/liquidfood/W = new(src)
 		stock(W)
-	for(var/i in 1 to 2)
+	for(var/i in 1 to 5)
 		var/obj/item/device/fbp_backup_cell/W = new(src)
 		stock(W)
-	if(prob(50))
-		var/obj/item/weapon/storage/pill_bottle/dice/D = new(src)
+	for(var/i in 1 to 1)
```
Arokha: what

Arokha:
That's a lotta firstaid kits. The advanced one wasn't supposed to make it a field hospital, it was just supposed to let you fix a couple of things. This shouldn't really be called a tweak so much as a buff, because that's what it is.

code/modules/mining/shelter_atoms.dm
```diff
@@ -219,15 +220,27 @@
...
+		var/obj/item/weapon/storage/pill_bottle/dice_nerd/D = new(src)
 		stock(D)
-	else
-		var/obj/item/device/instrument/violin/V = new(src)
-		stock(V)
+	if(luxury)
+		for(var/i in 1 to 5)
+			var/obj/item/weapon/reagent_containers/food/snacks/liquidfood/W = new(src)
+			stock(W)
+		for(var/i in 1 to 1)
```
Arokha: also what, don't do a for loop

Arokha:
I guess it's more fine, but I dunno why you're putting the first aid kit and such in the fridge to begin with. It wasn't fine on the table? It might be less obvious to people not digging in the fridge.

Novacat:
My original intention was to pair this with a PR for Ascian's shelter capsule which I currently have an application for, which will probably have a personalized interior that may not have the table. This also would make it easier to have personalized shelter capsules in general, since anything affecting mechanics/balance is in the storage units while fluff items like alcohol and a single unit of food are left out.

Arokha:
TBH adding customizations to the shelter maps would be preferred to adding them to the code that applies to everyone, you know?

Like I don't want the variable for the fridge to be like "ascian" and then the fridge checks if the thing is set to "ascian" or anything when it's spawning items.

And if you're going to make a subtype of the fridge for /ascian, then you're going to have to make a separate map anyway...! So it doesn't actually help keep things cleaner, it actually just encourages putting people's custom items as part of code that the non-custom one has to go through each time, which is messy and I'd rather not do that.

You can leave the luxury fridge thing the way it is, because that's fine. I'm indifferent about that, but starting to add people's custom items in something the normal fridge is going to have to 'check' each time it spawns is meh.

Novacat:
The plan was to not customize the fridge's contents at all for custom capsules, since the idea was that custom shelters would just alter the fluff part of the shelter.

Arokha:
If custom shelters all get custom maps, then how does the luxury fridge change make it easier?

Novacat:
Basically, with how it is now, all custom maps would have to figure out where to stuff an advanced first aid kit, in order for balance to be kept. With this, there is no need to figure out a place to stuff the kit. It is a relatively *minor* thing and I can return the advanced first aid kit to the table, if you would like.

SpadesNeil:
@Arokha This good to go? Looks good to me.

SpadesNeil:
Hello? Is this gonna get re-reviewed?

SpadesNeil:
o k a y

PR #4414, opened 2018-10-17
Condimaster is now ACTIVE in the kitchen

lbnesquik:
- Fixes aronai being LAZY

### DIFF
--- a/maps/tether/tether-03-surface3.dmm
+++ b/maps/tether/tether-03-surface3.dmm
@@ -9670,7 +9670,7 @@
 /turf/simulated/floor/tiled/freezer,
 /area/crew_quarters/freezer)
 "qY" = (
-/obj/machinery/chem_master,
+/obj/machinery/chem_master/condimaster,
 /turf/simulated/floor/tiled/freezer,
 /area/crew_quarters/freezer)
 "qZ" = (


Arokha:
:[!!

ArgobargSoup:
Does the condimaster still let you make pills?  They were very useful for reducing the amount of bottles produced.

...That and chemmaster bottles carry 60 units, condiment bottles only carry 50.

TheFurryFeline:
@SpadesNeil, just started looking at past PRs and if the CondiMaster hasn't had a change recently, in the past half year, this change will make it all but impossible to make bottles from containers for those with kitchen access. If I could edit a .dmm file on GitHub browser or something, I'd make an addition to the area to include the both. This comment here is because of a recent issue I heard about CHOMPstation doing the same thing with their own ChemMaster in the Kitchen.

SpadesNeil:
I feel like removing the ability to make bottles was a deliberate alteration but I don't think anyone would complain if you made a tweak.

Screemonster:
the condimaster can make bouillon cubes if people wanna store reagents from it

TheFurryFeline:
I'll admit, when I said about the issue that people were having with this were because of the previously merged #4485 not having been added downstream.

PR #4416, opened 2018-10-18
Added intelligence levels to some virgo mobs

Heroman3003:
I wanted to initially give wolf and catgirls an SA_PLANT originally, but then decided against furthering the meme.

All mobs listed previously had SA_ANIMAL level intelligence

New:
Shadekin (all types): SA_HUMANOID
Awoo (file name should really be un-memed at some point...): SA_HUMANOID
Catgirl: SA_HUMANOID
Corrupt Hound: SA_ROBOTIC


SpadesNeil:
-peers- So wait what do these do? This is new to me.

Heroman3003:
As far as I am aware, they are only in one check right now, but may be more at later point?
Its in checks for slime potions, as those are only meant to work on mobs with animal or lower intelligence. 
Additionally, slime baton and slime taser won't stun things smarter than animal.
Berserk artifact effect wont show for plants and robots.
And last bit of flavor, using ionic rapier on the robotic mobs makes them madder than normal.

So mostly, its for miscelaneous checks all around, that may or may not have further effect in future

Arokha:
Fix conflicts please

Heroman3003:
Okay, I messed up. Gonna just re-make the PR...

PR #4431, opened 2018-10-20
[MIRROR] Ports Vote to SMC

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/5675
--------------------
title
I've tested it with crew transfer voting, everything seems fine.


Arokha:
This probably won't break voting, probably.

PR #4434, opened 2018-10-20
Nerfs Cash from Mining Vendor

Novacat:
The current output of cash from the mining vendor is a bit too high considering that it is not hard for miners to pull 20K-40K points, and, probably will break the next trader ship event when it comes if a miner can just drop 15K-25K Thalers on the table.

### DIFF
--- a/code/modules/mining/ore_redemption_machine/equipment_vendor.dm
+++ b/code/modules/mining/ore_redemption_machine/equipment_vendor.dm
@@ -34,7 +34,7 @@
 		new /datum/data/mining_equipment("Fulton Pack",					/obj/item/extraction_pack,											1200),
 		new /datum/data/mining_equipment("Silver Pickaxe",				/obj/item/weapon/pickaxe/silver,									1200),
 		//new /datum/data/mining_equipment("Mining Conscription Kit",	/obj/item/storage/backpack/duffelbag/mining_conscript,				1000),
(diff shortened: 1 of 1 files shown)


KasparoVy:
Gold rush

Mewchild:
For those who don't want to look at the file, this is a 90% reduction. From 1000 Thaler down to 100

code/modules/mining/ore_redemption_machine/equipment_vendor.dm
```diff
@@ -34,7 +34,7 @@
 		new /datum/data/mining_equipment("Fulton Pack",					/obj/item/extraction_pack,											1200),
 		new /datum/data/mining_equipment("Silver Pickaxe",				/obj/item/weapon/pickaxe/silver,									1200),
 		//new /datum/data/mining_equipment("Mining Conscription Kit",	/obj/item/storage/backpack/duffelbag/mining_conscript,				1000),
-		new /datum/data/mining_equipment("Space Cash",					/obj/item/weapon/spacecash/c1000,									2000),
+		new /datum/data/mining_equipment("Space Cash",					/obj/item/weapon/spacecash/c100,									1000),
```
SpadesNeil: // Vorestation edit?

Arokha: pls do yes

Novacat: It already has a block edit

PR #3777, opened 2018-05-28
Pathfinder Update

Novacat:
- Implemented Pathfinder, Expedition leader role
Current channels: Exploration, Science
Current access: Pilot, Explorer, Research, External Airlocks, Maintenance Tunnels, EVA
Current Payscale: 7
- Added Deluxe Machete for Pathfinders (just a cosmetic reskin)
- Added Pathfinder's office, between expedition prep and briefing
- Expanded size of expedition shuttle
- Ported Pilot and Explorer voidsuit from Baystation
Explorer voidsuit armor values: armor = list(melee = 40, bullet = 15, laser = 25,energy = 35, bomb = 30, bio = 100, rad = 70)
Pilot voidsuit armor values:  armor = list(melee = 40, bullet = 5, laser = 20,energy = 5, bomb = 15, bio = 100, rad = 50)
- Restored the old air tank on the expedition shuttle
- The expedition shuttle now has toggled blast doors to cover the windows
- Increases pay of pilots, SARs, and Explorers (5, 6, and 6, respectively)
- Adds box of volunteer headsets to the Pathfinder's locker

PR #4436, opened 2018-10-20
Removes Binoculars from Pathfinder Office

Mewchild:
When the pathfinder update was originally put together, one of the admins (@DemgelZalvine) left [a note ](https://github.com/VOREStation/VOREStation/pull/3777#issuecomment-393013147) stating to remove the binoculars from the Pathfinder office. This change was not made. With binoculars being removed from mining, this PR removes the last set from the station.

### DIFF
--- a/maps/tether/tether-05-station1.dmm
+++ b/maps/tether/tether-05-station1.dmm
@@ -22208,7 +22208,6 @@
 	pixel_x = 26;
 	pixel_y = -26
 	},
-/obj/item/device/binoculars,
 /turf/simulated/floor/tiled,
 /area/tether/station/pathfinder_office)
 "rhA" = (


Arokha:
Hmm, reading that note he also said to remove access_gateway, which we just added. I'll talk to him about it to see what his issue was with that part...

SpadesNeil:
What's the reason for removing binoculars? We already solved the exploits.

Mewchild:
From speaking with Ascian in the past few days, the largest use of the binoculars was in conjunction with Meson/Material Scanners, in order to find points of interest through walls. The other concern was being able to shoot while zoomed, as I recall from the original discussion, but that seems to have been disabled in https://github.com/PolarisSS13/Polaris/pull/5284

Arokha:
I think there was a way to circumvent that 5284 though it was a while ago that I tested it.

Novacat:
The current in-round expedition areas required mesons, primarily because they are close quarters areas, but future gateway missions as well as events will heavily utilize open terrain, where the binoculars would be more useful for Pathfinder.

Also, that note is ancient, Zalvine originally intended for ONLY the Research Director and Colony Director to have Gateway access.

Wickedtemp:
Honestly, if any job would get use out of binoculars, it'd be pathfinder/explorers. Should probably keep them.

PR #4443, opened 2018-10-22
Splits first aid kit spawning

Novacat:
Removes the override on first aid kit spawning preventing combat kits from being spawned, but adds a new 'Tetheraid' first aid kit spawn to prevent combat kits from spawning on the Tether, but allows them for expedition sites which still use random spawns.

Also changes the obj/random/maintenance randomly generated item to also only spawn 'Tetheraid' first aid kits, since it shows up twice on the Tether.

TL;DR, it splits first aid kit spawns into one for use on the Tether, which do not have a combat first aid kit, and one used outside the tether (such as POIs) which do have the possibility of spawning a combat first aid kit.

### DIFF
--- a/code/game/objects/random/random_vr.dm
+++ b/code/game/objects/random/random_vr.dm
@@ -201,9 +201,31 @@
 			this_mob.forceMove(T)
 
 //Just overriding this here, no more super medkit so those can be reserved for PoIs and such
-/obj/random/firstaid/item_to_spawn()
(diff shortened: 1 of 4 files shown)


Arokha:
Was opened during GitHub issue times, and Travis hook was never called, so I can't 'restart' the job because the hook itself failed due to GitHub problems. Just closing and reopening to re-hook it.

PR #4451, opened 2018-10-23
Subspace Radio and Expedition Tweaks

Novacat:
The goal of this PR is mainly to move our exploration away from the Polaris roots and refine the various expedition roles for Virgo, and a lot of cosmetic tweaks. The access changes are supposed to more clearly define the support and combat roles of the expedition, as well as prevent the popular 'Lone explorer flies off with the shuttle' meme.

Also, finally adds the Pathfinder's subspace radio which has been in development hell for months. The subspace radio is a backpack radio that is designed to send and receive transmissions when it is deployed. Deploying the subspace radio is similar to defibrillator units, and can communicate and receive transmissions from any radio.

As always, Feedback welcome.

- NEW: Adds the Subspace Radio for the Pathfinder, in the Pathfinder's locker
- Renames Search and Rescue to Field Medic
- NEW: Cosmetic changes for ID Card graphics and PDAs for expedition crew
- NEW: Explorers now get Science cartridges in PDAs instead of cargo cartridges
- NEW: Pilots no longer get cartridges in PDAs
- Moves the holdout phasers from the expedition weapons locker to the job equipment lockers of the roles intended to use them (Pilots/FMs)
- Removes Psychologist, Virology, and Explorer access from Field Medics, adds Pilot access to Field Medics
- Removes Pilot access from Explorers
- Restricts the expedition shuttle to Pilot access only
- Adds a medic version of the explorer suit for Field Medics, identical stats, but has a medic arm-band attached.
Note: Might be replaced with a reskinned explorer suit down the road.
- Replaces regular gas mask with explorer gas mask in Field Medic lockers
- Adds liquidfood rations for FMs and Pilots
- Adds winter boots for Pilots
- NEW: Corophizine now treats any stage virus


Wickedtemp:
Field medic should probably have access to the morgue, given they'll, at some point, need to put somebody into it or take somebody out.

TheDavestDave:
if you want to remove access no one does feild surgery. So they could live without surgical access

Arokha:
Yaya remove surgery access from 'em.

Novacat:
My criteria in determining access within medbay was asking myself the questions "Would the field medic need to access this place in an emergency and there are no other medical staff available?" and "Would the field medic need access to this place in order to do their basic duties?"

The Morgue is only used for permanently stowing bodies after resleeve. If there is just a FM and no one else online (at which point, you should probably be faxing for medical), in an emergency the FM can just leave the body bags outside the morgue until medical staff can stow them away. The only issue I can see from this is that most of medbay's body bags are currently stored inside the morgue, but, a simple mapping change can fix that.

Surgery on the other hand, might be required if there are no other medical staff who can do it, broken bones are the most common ailment fixed by surgery, but internal bleeding may also require surgery to be treated, and shrapnel requires surgery as there is no non-surgical way to get rid of shrapnel. This is why the FM has surgery access.

Vorrarkul:
I can understand not giving them surgery access, but I'd argue they can use the morgue to store bodies of, for example, any explorers who perished and were brought back to the station.

Wickedtemp:
Everyone in Medical can access the morgue and operating theaters. The only exception might be the Psychologist. I think 'Field Medic' should have the access as well.

It's extremely helpful to be a doctor and handling a patient and having someone you can call to assist if needed. If you're having an issue, whether its with stabilizing, a surgery, if you just need something delivered to the OR like a chem or a tool or replacement limb, or maybe youre having ping and connection issues at the worst possible time, if there's another person in medical, you can call them in to see if they can't help.

And it's annoying as fuck when it turns out they can't because they don't have access, when nearly every single job in medbay has it. 

This isn't an uncommon thing. Literally last night, CMO had to troubleshoot a bit, Tempest went into the OR - as chemist- to assist. So, yeah, even if they aren't expected to do surgery, access is obscenely useful. 

As for morgue, there's literally no reason why it should be excluded. No, you should *not* just "leave the corpses outside", bagged or not. Morgue access is the most trivial of all of these because it gives literally zero advantage of any kind aside from the ability to put corpses away. Which is something they'll likely need to be able to do.

Arokha:
I mean I think part of the access reductions was due to CMOs and others telling SARs to perform normal medical duties they didn't want to do, because it's not part of their job. I think this is a mechanical way of saying 'no', because there were SARs who were threatened with firing because they didn't want to do normal medbay things and wanted to go on expeditions and CMOs said no? So said Ascian anyway.

Heroman3003:
Didn't morgue as a separate access level originally exist just to allow chefs, chaplains and roboticists to freely take corpses from it back in the original days? On our server such things are much more RPed and organized, so no reason for medical staff not to have it.

Arokha:
> chefs, chaplains and roboticists to freely take corpses from it back in the original days

No, it was also for sec and such to do investigations. Like HoS has/had morgue access. I don't think CHEF was allowed. >.> On Box when the morgue was more easily accessible from the hallway it made a little more sense. But on Northern Star and Tether, the Morgue hasn't been accessible from a public area for a while, so they'd also have to have medical access.

I dunno. I think people should have access to perform what their job duties are, you know? Isn't that what NT would give out? Not what 'they might need to help with if asked by someone', since in those cases I'd imagine NT would just say 'so get the HoP or CD to change them'.

Heroman3003:
I'd say remove virology and psych access, and keep the rest. Yes, field medics are not expected to perform duties outside expeditions. But if there is no expeditions and they want to help out why limit them? Its basically same thing as doctors having access to chemistry, atmostech having access to engine, chef having access to hydroponics, etc.

Wickedtemp:
If we're going by what access they'd need, then they'd definitely be keeping morgue access.

Arokha:
I don't think they need morgue access though. Them getting bogged down with autopsy stuff or anything else like that, paperwork, etc, probably isn't the best.

Also the fact that the SAR could come in, put a body in, then leave... and if it comes up later, the 'normal' medbay staff that actually staff the medbay, that the morgue is part of, wouldn't know that there is a body there or anything unless they happen to check every 5 minutes... that could be an issue.

From the standpoint of 'who should put the bodies away', I'm thinking it should always be normal medbay staff, not people who don't staff the medbay.

Wickedtemp:
Heroman pretty much summed it up.

Aronai, they have access to the morgue now, and I can't really recall a single instance of that scenario. Every time a SAR's taken a corpse to medbay, it's mentioned. 

"Autopsy stuff' doesn't even exist here. There's no reason to ever carry one out. Like... Iunno, this is stuff that'd be extremely obvious to anyone playing Medical on this server and it confuses me why youre bringing it up like it's an issue that actually happens with enough frequency to be a problem.

Arokha:
Fair enough, but virology is a little silly. That problem should be fixed some other way other than "My SAR also knows virology"

Novacat:
Virology is mainly because of there being no mechanical alternative to treating disease, some diseases are very nasty, and there is the very real threat of disease while on expedition.

Wickedtemp:
Virology is meh, but at the same time, I'm fine with it. The list of "People that know virology", if listed by job, would be random. A lot of doctors and nurses don't know it, some CMO's don't know it, same for people that DO know. And its such an uncommon thing that I doubt it'd really ever matter aside from checking viro for solar grubs.

I kinda wish we started out with viro samples. Then it might get used more.

Heroman3003:
But then we get people trying to 'make beneficial viruses' while our codebase literally only has viruses as negative thing.

Wickedtemp:
Yeah, but that's just an IC thing, and kinda something they can try to do as it is anyway if they can get a crate ordered. Polaris has a couple viral samples spawned in each round, I feel like we could do that here, but that's for another PR.

Heroman3003:
If my memory serves me right, Polaris also has actual antags that can, in theory make use of it, and virus events at random enabled. Which is whats different.

Novacat:
Suggestion: Give the ability for Corophizine to treat viruses past stage 1. This would make virology no longer _necessary_ for the FM, and thus I would be comfortable removing virology access from the FM.

Zeldazackman:
Full med access barring the CMO's office for search and rescue when?

Novacat:
Added Morgue access back, removed virology access, added Corophizine treating any stage virus.

Arokha:
Split these PRs please.

PR #4455, opened 2018-10-25
Protean Bugfixes

king5327:
Proteans now drop items on the ground if killed when blobbed, instead of leaving them in nullspace.
Proteans now properly preserve inhand items (if they are not dropped)
Proteans no longer lose their v-organs if they try to deblob and don't have enough health to prevent immediate recollapse.
Protean complete rebuild is functional now. Also no longer causes NIF failiure (because of a runtime).
Proteans no longer runtime if one-shotted in human form.
Proteans can now reconstruct limbs over stumps.
Proteans can no longer eat any material that is useless.
Proteans now drop micros on blob instead of taking them on a ride to nullspace. This is notable because the ID slot is preserved and micros can be put in there.

Should be everything unless I've forgotten some minor fix or another. Cheers from Citadel.

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/protean_vr/protean_blob.dm
+++ b/code/modules/mob/living/carbon/human/species/station/protean_vr/protean_blob.dm
@@ -45,6 +45,9 @@
 	var/obj/item/organ/internal/nano/refactory/refactory
 	var/datum/modifier/healing
 
+	var/obj/prev_left_hand
+	var/obj/prev_right_hand
+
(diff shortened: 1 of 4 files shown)


Arokha:
Has this been used downstream? Tested?

king5327:
Yep. It's been running since the first week of september as a testmerge. There's trouble syncing to current VORE so it got untestmerged a few days ago.

And I tested it all both on my end and serverside before making this PR.

Arokha:
Appreciate it.

PR #4456, opened 2018-10-26
[MIRROR] Semifinal AI : Electric Boogaloo

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/5704
--------------------
Almost done with the rewrite.
All it needs after this is another unclusterfucking with master, server tests, bug fixes from the tests, then probably another unclusterfucking, and then finally a merge into master.
Now targeting the correct branch.


Neerti:
Another non-master PR. Should only have a few more of these until the rewrite gets merged into master.

PR #4458, opened 2018-10-28
Round Extension

Pawoverlord:
This change is to update the round time from 5 hours to 8 by giving 4 votes to extend votes before the autotransfer is sent out.


Pawoverlord:
Wrong fork. wow Git :/

PR #4462, opened 2018-10-28
Expedition Job and Gear tweaks

Novacat:
The goal of this PR is mainly to move our exploration away from the Polaris roots and refine the various expedition roles for Virgo, and a lot of cosmetic tweaks. The access changes are supposed to more clearly define the support and combat roles of the expedition, as well as prevent the popular 'Lone explorer flies off with the shuttle' meme.

As always, Feedback welcome.

- Renames Search and Rescue to Field Medic
- Cosmetic changes for ID Card graphics and PDAs for expedition crew
- Explorers now get Science cartridges in PDAs instead of cargo cartridges
- Pilots no longer get cartridges in PDAs
- Moves the holdout phasers from the expedition weapons locker to the job equipment lockers of the roles intended to use them (Pilots/FMs)
- Removes Psychologist, Virology, and Explorer access from Field Medics, adds Pilot access to Field Medics
- Removes Pilot access from Explorers
- Restricts the expedition shuttle to Pilot access only
- Adds a medic version of the explorer suit for Field Medics, identical stats, but has a medic arm-band attached.
Note: Might be replaced with a reskinned explorer suit down the road.
- Replaces regular gas mask with explorer gas mask in Field Medic lockers
- Adds liquidfood rations for FMs and Pilots
- Adds winter boots for Pilots
- Corophizine now treats any stage virus
- Field Medics can now access medical stuff in loadout
- Explorers can now access holsters and science stuff in loadout
- Pathfinder can now access holsters, science stuff and webbing in loadout
- Cleans up the code for the expedition jobs in general
- NEW: Pilot and Field Medic now require accounts 3 days and older before they can be played


code/modules/virus2/disease2.dm
```diff
@@ -88,6 +88,13 @@
 		else
 			resistance += rand(1,9)
 
+	//VOREStation Add Start - Corophazine can treat higher stages
```
SpadesNeil: This feels out of scope. Why?

Novacat: This was done because Aronai requested that the virus problem be dealt with by some other way than the Field Medic going into the Virology lab.

See: https://github.com/VOREStation/VOREStation/pull/4451#issuecomment-432458597

SpadesNeil:
Or not. :U They are probably busy because of the holidays, as am I.

PR #4464, opened 2018-10-30
Move photosensitive to be a neutral trait

Arokha:
It appears that everyone and their dog takes photosensitive, probably because it never comes up in the game at all, and gives what amounts to a free point.

It's the most popular negative trait by a decent margin.

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/traits_vr/negative.dm
+++ b/code/modules/mob/living/carbon/human/species/station/traits_vr/negative.dm
@@ -90,12 +90,6 @@
 	cost = -2
 	var_changes = list("siemens_coefficient" = 2.0) //This makes you extremely weak to tasers.
 
-/datum/trait/photosensitive
-	name = "Photosensitive"
-	desc = "Increases stun duration from flashes and other light-based stuns."
-	cost = -1
-	var_changes = list("flash_mod" = 2.0)
-
 /datum/trait/hollow
 	name = "Hollow Bones/Aluminum Alloy"
 	desc = "Your bones and robot limbs are much easier to break."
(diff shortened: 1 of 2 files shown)


Novacat:
It might not be a bad idea to reduce the flash mod in that case, since I do recall there being issues where any kind of flash can KO the character for some time.

CameronWoof:
Despite the rarity of being affected by it, it _is_ still a weakness. If you're moving it to neutral, it might be better to just remove it entirely since it will no longer serve a purpose.

KasparoVy:
RIP

SpadesNeil:
Is say we go the route of removing it completely if that is the case.

Or maybe Security should abuse flashes more. C:

KasparoVy:
Walk down the hall with a use button repeat macro, aoe spamflashing fools into the next dimension

Arokha:
> If you're moving it to neutral, it might be better to just remove it entirely since it will no longer serve a purpose.

> [I] say we go the route of removing it completely if that is the case.

That was my original plan, however when I said I was going to do this, several people who have taken it with darksight who want to keep it for RP purposes stated they hoped it could be moved to neutral instead.

> it is still a weakness

Yes, but not everything in neutral is completely without use. Glowing body is a permanent, un-stealable flashlight for example. Is that a huge benefit? No, not really. Everyone has a PDA, and even if stolen it doesn't really inconvenience them too much. The metabolism traits modify how fast you get hungry and how fast drugs affect you, both of which could be seen as up and downsides, but they are neutral because "it's just not that big a deal". Sorta like this trait that really doesn't come up in gameplay often at all. That's what neutral traits are. Ones that just don't affect a whole lot of anything.

SpadesNeil:
Could possibly merge it with dark sight, in exchange for maybe knocking off a point.

KasparoVy:
I agree with Spades' idea about the merge. Organic darksight should mean the further you see/the more adapted your eyes currently are to the dark, the more susceptible to eye damage from light you are.

HungryMoth:
I like the idea of making it where the better darksight you have, the more sensitive to light you are.

Would that make darksight a neutral perk though since it has both ups and downsides? Or is that just a nerf to darksight with the same cost but with more downsides to it?

Personally I use the trait system purely for what makes sense IC'ly.

Arokha:
> I like the idea of making it where the better darksight you have, the more sensitive to light you are.

Did this in the associated PR above

PR #4480, opened 2018-11-05
Edit mapped in HoS armor

Arokha:
It was edited on the map to have pretty incredible armor values, better than SWAT armor, even. Things like this shouldn't be edited in the dmm files. Hard to have oversight on things like that.

Moved the armor values to the code and out of the map, and adjusted them to be more reasonable after discussing with other admins, so now it's an up-armored security voidsuit, not deathsquad armor.


KasparoVy:
Wow. That was pretty legendary armour.

_Was._

PR #4483, opened 2018-11-07
Shadekin Mlempathy

Wickedtemp:
- Makes the language Whitelisted instead of Restricted, allowing for whitelists.
- Adds the language to myself (from https://forum.vore-station.net/viewtopic.php?f=46&t=1455 ) and Zalvine, who has it already but they have to manually give it to their character every round and it's easier to just have it done this way.
- Also moved 'wtfismyname' a bit so it's actually in the correct alphabetical order.


DemgelZalvine:
Looks good to me. ;o

PR #4490, opened 2018-11-10
Gives Vasilissans a language finally / Bug-Speak

HungryMoth:
Adds Vespinae: A common language designed for creatures without traditional mouths and tongues. Mandibles, pincers, clicky-clackery-doos.

They have been developed primarily as an extension upon Vasilissan's unique dialect, and as a result has become a psuedo-common tongue among creatures with non-traditional mouthparts!

So with the recent addition of abdomens, some antennae, and my general love for the insectoid life. I feel that they could use a language.

Not whitelisted, normal language, mostly for fluff and lore. It gives Vasilissans, a species that has never had a language while everyone else has sone. They get one finnaly! Yay! And for other custom species they have a 'trope' language they can fall back on to use, and it generally helps with character development, well we don't seem to have anything for our spooky insect friends, so here it is.

A new common language that the Vassilians may use.

Works just fine, tested and compiled on my end, and here is what it looks like, very click-clacky.
https://gyazo.com/1d522224cd1371860ff556b429b3577a

First pull request here so, uh, yeah!

### DIFF
--- a/code/modules/mob/language/station_vr.dm
+++ b/code/modules/mob/language/station_vr.dm
@@ -70,6 +70,18 @@
 	syllables = list("salve","sum","loqui","operatur","iusta","et","permittit","facere","effercio","pluribus","enim","hoc",
 	"mihi","wan","six","salve","tartu")
 	machine_understands = FALSE
+	
+/datum/language/bug
+    name = LANGUAGE_VESPINAE
+    desc = "A jarring and stuttery language devised by the various arthropod species of the galaxy to communicate. It is exceedingly complex and practically impossible to learn if you lack the proper mouthparts."
+    speech_verb = "clicks"
+    ask_verb = "chitters"
+    exclaim_verb = "rasps"
+    colour = "bug"
+    key = "x"
+    syllables = list("vaur","uyek","uyit","avek","sc'theth","k'ztak","teth","wre'ge","lii","dra'","zo'","ra'","kax'","zz","vh","ik","ak",
+    "uhk","zir","sc'orth","sc'er","thc'yek","th'zirk","th'esk","k'ayek","ka'mil","sc'","ik'yir","yol","kig","k'zit","'","'","zrk","krg","isk'yet","na'k",
+    "sc'azz","th'sc","nil","n'ahk","sc'yeth","aur'sk","iy'it","azzg","a'","i'","o'","u'","a","i","o","u","zz","kr","ak","nrk","tzzk","bz","xic'","k'lax'","histh")
 
 /datum/language/unathi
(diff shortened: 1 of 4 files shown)


code/modules/mob/language/station_vr.dm
```diff
@@ -70,6 +70,18 @@
 	syllables = list("salve","sum","loqui","operatur","iusta","et","permittit","facere","effercio","pluribus","enim","hoc",
 	"mihi","wan","six","salve","tartu")
 	machine_understands = FALSE
+	
+/datum/language/bug
+    name = LANGUAGE_VESPINAE
```
KasparoVy: # WE REQUIRE MORE VESPENE GAS

HungryMoth: Fun Fact: That name comes from "Vespid" which is the taxonomic order of wasps! Not that the language was only for wasps or anything, but _The more you know!_

SpadesNeil:
I would rather not just slap in a generic insect species without an insect race. That said, we already have wasp bits. We could totally add in such a bug race.

Even though bugs bother the hell out of me. :S But we'd be setting a bad precedent if we just added languages with no lore or race to feature them.

Screemonster:
Well, we _do_ have vasillisans or however the fuck they're spelled.

HungryMoth:
We do have Vasillians and they don't have a language at all, I can retool the wording and lore so it is a Language for them.

That would solve the problem of a lore race that doesn't have a language getting one, and still fill in the goal of giving custom species an option.

SpadesNeil:
Let's do that, then.

HungryMoth:
I made the necessary changes and changed the title as well.

What this does now is give Vasilissans their own language by default, and updates the lore/flavor-text that describes it as being developed by them.

Probably was needed for a while as I believe Vasilissans are the only lore-species we have (Besides Xenochimera) that don't have their own language.

I hope this is acceptable now.

Heroman3003:
I don't think any species has secondary language added outside their own, so since Vespinae is becoming their official one, I think removing Sol Common from them would make sense, no?

HungryMoth:
You are right and I considered doing the same thing, but I for one don't know why Sol Common was there in the first place. I opted instead to just leave Sol Common behind unless whoever is in charge of the lore wants me to move that.

Heroman3003:
I think it may have been a placeholder, considering they are the only race around without defined language and just were given a 'something'.

HungryMoth:
I suppose you're right. I removed it, no reason why extra-terrestrial space spiders would understand Sol Common naturally.

PR #4492, opened 2018-11-11
Some trash eater tweaks and additions.

Verkister:
-Adds a bunch of new low exploitability stuff to the garbo snack menu for suggested scene/fun shenanigans. (small non-storage gear (hat/mask/glasses/gloves/shoes), PDA, headset, organs, more toys, bones, broken emags, busted circuitry, lighters, bottles, lifecrystal cases)
-The PDAs give a warning and a delay time if owned, and come with an additional confirmation popup if they also contain an ID.
-Owned PDAs cannot be ingested unless the owner is watching and connected to prevent grief.
-Excludes all uplink enabled items from ingestion to prevent abuse.
-Attempting to abuse this in order to hide or find uplinks will also notify admins about your misdeeds.
-Also excludes boots that contain stuff to prevent knife smuggling and such.

-Also fixed a bug that would cause "partially-digested" PDAs to forget the existence their IDs and thus make them unrecoverable until the PDA is deleted by gurgles. (introduced by the gradual gurgles update)
-Also reverts contamination disabling all PDA functionality (including ID ejection)


Screemonster:
if you have itemweak on, can you eat someone's PDA right in front of them without damaging it?

Because eating someone's PDA in front of them and making them climb in after it is some good teasing.

Arokha:
This is cool 'n all, but doesn't this make it basically impossible for security to strip you of these things if they brig you? Or if you have things illegally? They also have no way of knowing that you have them since we don't really have any cavity search mechanics. Like if YOU are security, or someone who's 'obtained' a high-access ID, security has no way to know that. There's just nothing in the game that lets them investigate that on you, plus no way to make you give it up without involving medical. And if there's no chemist on then you're sorta just doomed.

Arokha:
... also I believe this makes every uplink item eatable.

HungryMoth:
Doesn't eating an ID ruin it anyways? So technically you could hide a high-level ID but not to use.

I'm not saying it can't be abused at all but I feel like with some of these items an OOC ruling on things would make the RP worth it, especially since the extent of abuse would be uh, hiding a bunch of clothes inside of you?

Very-Soft:
While I agree with Scree, that scarfing PDAs in front of someone is good teasing, this is a real easy way to just let people casually shred people's IDs, being that IDs in bellies get ruined even if they're in other things. Allowing them to be eaten with garbage eater like that just seems like an easier way to allow that to happen, and I'm not super keen on that personally. 

The rest is basically fine though.

Verkister:
Uplink items have relevance here?

Arokha:
It's 'Yet Another Thing' we couldn't use in events for this reason, though, and the downstreams all use them. Either figure out a way to prevent that or remove 'em.

KasparoVy:
Flag digest_resilient for cases where certain items need to be robust against digestion? Or unsafe_consume where it'll throw a message like "no dont eat this" and prevent consumption?

Verkister:
Also we should totally give security like some sorta simple belly detector gate thingy to do a little checkup with before dumping the contraband mules or matryoshka operative teams into the storage unsupervised.

KasparoVy:
That'd be interesting. Have it up in processing attached to a display that lets you print logs as desired. But I can't quite recall how cramped for soace the area is so I'm unsure of precisely where it could go.

Verkister:
Probably gonna see if I can figure out a way to make owned PDAs edible only when the owner is watching and connected in order to prevent people just gobbling down any lost&founds or looting SSDs.

Screemonster:
iunno, this is starting to get a little too convoluted.

SpadesNeil:
nyet

KasparoVy:
BLYAT

PR #4495, opened 2018-11-12
Collar tweaks

ArgobargSoup:
This PR does two things.
- Makes it so you can add a tag/change the tag of any collar, including shock collars and life crystals.  For any collar other than the shock collar and life crystal, you can set an initial name without requirements by clicking on the collar in your active hand, for purposes like simulating having come on to the station with the collar tag already set.  After that, for any tag changes, or if you're tagging a shock collar or life crystal, you need to use a **screwdriver** or a **pen**, to scratch in/engrave the tag or write on it.
-- The messages will change depending on whether you use a screwdriver or a pen, and whether something was already written on the tag.  If you use the initial free tag setting, it will default to "engraved".
-- Holo collars are exempt from all requirements; they will work like they've always had.

- Adds a verb to bell collars, that lets you jingle the bell if you're holding it or wearing it in your oversuit slot (but not while attached, sadly).  Has a 5 second cooldown to hopefully prevent too much spam.

Screenshots:
image

image

image

image

### DIFF
--- a/code/modules/clothing/under/accessories/accessory_vr.dm
+++ b/code/modules/clothing/under/accessories/accessory_vr.dm
@@ -27,7 +27,7 @@
 	icon_state = "collar_bell"
 	item_state = "collar_bell_overlay"
 	overlay_state = "collar_bell_overlay"
-	var/nospam = 0
+	var/jingled = 0
 	
 /obj/item/clothing/accessory/collar/bell/verb/jinglebell()
 	set name = "Jingle Bell"
@@ -36,11 +36,11 @@
 	if(!istype(usr, /mob/living)) return
 	if(usr.stat) return
 	
-	if(!nospam)
+	if(!jingled)
 		usr.audible_message("[usr] jingles the [src]'s bell.")
-		nospam = 1
+		jingled = 1
 		sleep(30)
-		nospam = 0
+		jingled = 0
 	return
 
 /obj/item/clothing/accessory/collar/shock
--- a/code/modules/clothing/under/accessories/accessory_vr.dm
+++ b/code/modules/clothing/under/accessories/accessory_vr.dm
@@ -6,6 +6,7 @@
 	slot_flags = SLOT_TIE | SLOT_OCLOTHING
 	icon = 'icons/obj/clothing/collars_vr.dmi'
 	icon_override = 'icons/obj/clothing/collars_vr.dmi'
+	var/writtenon = 0
 
 /obj/item/clothing/accessory/collar/silver
 	name = "Silver tag collar"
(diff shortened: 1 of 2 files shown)


code/modules/clothing/under/accessories/accessory_vr.dm
```diff
@@ -27,6 +27,21 @@
...
+	set category = "Object"
+	set src in usr
+	if(!istype(usr, /mob/living)) return
+	if(usr.stat) return
+	
+	if(!nospam)
+		usr.audible_message("[usr] jingles the [src]'s bell.")
+		nospam = 1
+		sleep(30)
+		nospam = 0
```
Arokha: `schedule_callback_in(3 SECONDS, VARSET_CALLBACK(src, nospam, 0))`

Starts a timer to reset the var to 0 after 3 seconds. Also "If not nospam" reads weird. Double negatives are bad. Maybe rename the var to "played_recently" or "cooldown" or something.

Arokha:
Mweh. I kinda liked the holocollar having a special feature o: Maybe the normal collars should be settable only once? (How else are you carving a metal tag, anyway. Or maybe require a screwdriver to etch a new name in :P)

code/modules/clothing/under/accessories/accessory_vr.dm
ArgobargSoup: Doesn't seem to work in testing.  The var stays at 1, and an error is printed to the server log.
```
[2018-11-12T12:49:53] Runtime in scheduler.dm,143: Cannot execute null.Invoke().
  Caught by process: scheduler
  proc name: process (/datum/scheduled_task/callback/process)
  src: /datum/scheduled_task/callback (/datum/scheduled_task/callback)
  src.loc: /datum/scheduled_task/callback (/datum/scheduled_task/callback)```

ArgobargSoup:
Given that the only way to get collars at the moment is to have them selected in loadout, said method only allows you to take one of each type, and the scenarios the feature would likely be used in, I think it'd be best to let them be set multiple times.  Otherwise, an accidental activation or a keyboard flub would leave you out of luck.

Holocollars do keep their special activation text, and the tags on the others could possibly be explained as e-ink paper set into the tag.

KasparoVy:
Hm. Perhaps making collars an item with a 'slot' for a tag that is one-use but in turn can be replaced?

The tags could then, in turn, be added perhaps as lathe/crafting recipes and maybe the leather collar to the biogen and costume vendo.

Idea spitballing

Cameron653:
Looking over this, everything looks good to me codewise (Provided that callback thing works. I've never used or seen it.) and the feature seems good. Makes sense that you'd be able to scrape out letters with a screwdriver or write on it with a pen. (Personally, I feel like a marker /obj/item/weapon/pen/crayon/marker would be better writing utensil for the collars, but pens seem fair enough.)

PR #4499, opened 2018-11-12
Combine photosensitive trait with darksight

Arokha:
Combines the traits, because photosensitive was 'the meta' free trait that never came up. And darksight was the most popular trait. So combine the two most meta traits to blunt both? Ok. By popular request in the other PR: https://github.com/VOREStation/VOREStation/pull/4464

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/traits_vr/positive.dm
+++ b/code/modules/mob/living/carbon/human/species/station/traits_vr/positive.dm
@@ -42,13 +42,13 @@
 	name = "Darksight"
 	desc = "Allows you to see a short distance in the dark."
 	cost = 1
-	var_changes = list("darksight" = 3)
+	var_changes = list("darksight" = 3, "flash_mod" = 2.0)
 
 /datum/trait/darksight_plus
 	name = "Darksight (Major)"
 	desc = "Allows you to see in the dark for the whole screen."
 	cost = 2
-	var_changes = list("darksight" = 7)
+	var_changes = list("darksight" = 7, "flash_mod" = 3.0)
 
 /datum/trait/melee_attack
 	name = "Sharp Melee"
--- a/code/modules/mob/living/carbon/human/species/station/traits_vr/negative.dm
+++ b/code/modules/mob/living/carbon/human/species/station/traits_vr/negative.dm
@@ -90,12 +90,6 @@
 	cost = -2
 	var_changes = list("siemens_coefficient" = 2.0) //This makes you extremely weak to tasers.
 
-/datum/trait/photosensitive
-	name = "Photosensitive"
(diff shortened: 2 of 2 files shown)


code/modules/mob/living/carbon/human/species/station/traits_vr/positive.dm
```diff
@@ -42,13 +42,13 @@
 	name = "Darksight"
 	desc = "Allows you to see a short distance in the dark."
 	cost = 1
-	var_changes = list("darksight" = 3)
+	var_changes = list("darksight" = 3, "flash_mod" = 2.0)
```
Novacat: flash_mod is way too high... Vanilla species only use 1.1 and 1.2. 2.0 and 3.0 are way over the top. I would recommend 1.2 and 1.4, respectively.

Arokha: The previous photosensitive trait was 2.

Novacat: Perhaps 1.5 and 2.0, then?

HungryMoth: How bad is a 2.0 and 3.0 flash mod in practice? Darksight is a really good bonus so if it's as simple as 2x/3x as long stun duration it makes sense. If anything it gives me an excuse to RP my darksight char's as disliking bright areas.

KasparoVy: Live the Riddick lifestyle. Welding gogs 24/7.

Arokha: > How bad is a 2.0 and 3.0 flash mod in practice

0 because nobody gets flashed which is why it was the meta

Wickedtemp:
Should the costs be changed, since it has more of a 'neutral' tradeoff of sorts now, instead of just a purely positive benefit?

AaronSherwood117:
Oi, @SpadesNeil, people are asking questions >:c

AaronSherwood117:
I think Tempest's question should be addressed before this is in.  If it has an INBUILT downside and the downside people took with it is gone, it should have its pointcost adjusted and the descriptions of the trades updated.

AaronSherwood117:
@Arokha

AaronSherwood117:
"Everyone took them together anyways, so fuck it I'm combining them."
But you didn't even combine them, just added an even worse version of the downside while not combining things like, say, the point cost.  The descriptions need to reflect the new effects and the costs need to be adjusted at least a point downwards, or even make them both neutral since the stronger version has an even stronger debuff now.

SpadesNeil:
image
image

PR #4502, opened 2018-11-13
Fixes miss aligned wing sprites

Keekenox:
Bumps the wings over a pixel because I oof'd it the first time a little.


Keekenox:
pls, it hurts my eyes.

PR #4506, opened 2018-11-18
Adds Prybars to Survival Boxes

### LINKED ISSUE #4505: Hard to get out of tram spawn without power.
You can't get out of the tram spawn without power.

To have a crowbar accessible to me while I'm in the tram spawn in case the power is out.

#### What actually happened
The crowbar is instead just barely outside of it and all I can wonder is "why?"

#### Steps to reproduce
- Join the round when there isn't power to the tram station
- Don't be an engineer or roboticist
- Wish desperately for salvation while you stare angrily at the emergency locker just out of reach.

#### Code Revision
- 1fe04192c11e2ad7357512d5a873c44a4ea687d3 
 
- 
image

Novacat:
Fixes #4505 

Ported from baystation, and, is only able to pry open firelocks and unpowered doors, and nothing else.

The autolathe can also produce non-emergency variant prybars.


SpadesNeil:
Rather than a straight up crowbar, I'd like to replace this eventually with a tool that serves the same purpose as a crowbar, except causes less physical harm and doesn't rip off floor tiles.

An emergency door pry.

What would that even look like? No idea. We'll worry about it later. This PR is still a good alternative for now.

KasparoVy:
The emergency doorpry could look like the crowbar minus the hook at the end but maintaining the red 'stripe' in the middle, maybe something like this:

red prybar

code/game/objects/items/weapons/tools/crowbar_vr.dm
```diff
@@ -0,0 +1,32 @@
+/*
+ * Prybar
+ */
+
+/obj/item/weapon/tool/prybar
```
SpadesNeil: Needs some code to not fuck with anything except doors. I'll get to it.

code/game/objects/items/weapons/tools/crowbar_vr.dm
```diff
@@ -4,23 +4,30 @@
...
 
+// Todo: Prevent it from prying up floor boards. Only affect doors. I couldn't remember how to do this. -Ace
+/*
+/obj/item/weapon/tool/prybar/on_hit()
+	if(istype(/obj/machinery/door/airlock))
+		..()
+	else if(istype(/mob)
+		..()
+	else
+		// Tell the player that they can't use this on anything except airlocks.
```
SpadesNeil: @Arokha or @Cameron653 can probably figure out how to do this bit. I'm so rusty and they appear to be offline at the moment.

PR #4507, opened 2018-11-22
Fixes ingested organs giving way too much fat

Verkister:
Oh god I just realized the slow item digestion code made hardvored bodyparts have the calorie content of nuclear fuel lmao.


SpadesNeil:
Code looks about right to me.

PR #4508, opened 2018-11-23
Add slowdown for extreme amounts of stuffing.

lbnesquik:
All values are subject to changes once again. Compiled and tested.
It starts acting at 2000 which is "engorged with a huge stomach that sags and wobbles as they move." tier.

### DIFF
--- a/code/modules/mob/living/carbon/human/human_movement.dm
+++ b/code/modules/mob/living/carbon/human/human_movement.dm
@@ -32,7 +32,7 @@
 
 	//VOREstation start
 	var/stuffed = round(nutrition)//Slowdown due to belly
-	if (stuffed >= 2500) tally += round(stuffed/2500) // Kind of random number. About 3 of slowdown with 7000 nutrition which is a *lot*.
+	if (stuffed >= 2000) tally += round(stuffed/2500) // Kind of random number. About 3 of slowdown with 7000 nutrition which is a *lot*.
 
 	if (feral >= 10) //crazy feral animals give less and less of a shit about pain and hunger as they get crazier
 		tally = max(species.slowdown, species.slowdown+((tally-species.slowdown)/(feral/10))) // As feral scales to damage, this amounts to an effective +1 slowdown cap
--- a/code/modules/mob/living/carbon/human/human_movement.dm
+++ b/code/modules/mob/living/carbon/human/human_movement.dm
@@ -31,6 +31,9 @@
 	if (hungry >= 70) tally += hungry/50
 
 	//VOREstation start
+	var/stuffed = round(nutrition)//Slowdown due to belly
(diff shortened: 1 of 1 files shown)


lbnesquik:
For reference, at 10 000 nutrition which happens only if you are stuffing yourself with a kitchen worth of food, honey pills or pure sugar, you'll have 4 slowdown, which is just above a space rig 3 slowdown. 
Some suits go much higher so i think 4 isn't that much. If wanted, it could be lowered to match the space rig at 3 slowdown maximum.

Screemonster:
while 10000 is my "varedit chimeras while testing to get their regen time down to to the time it takes to make a pot of tea" number, this is still effectively gonna inflict people with not-insignificant effectively-permanent slowdown for indulging in vore.

Heroman3003:
Since slowdown begins at almost exact amount you get from eating one average person, I am against it, as it seems like something that punishes players for doing vore, especially people that do multiple prey at once/multiple scenes in a row for no good reason.

SpadesNeil:
I like the idea of this, but Scree makes a point.

>while 10000 is my "varedit chimeras while testing to get their regen time down to to the time it takes to make a pot of tea" number, this is still effectively gonna inflict people with not-insignificant effectively-permanent slowdown for indulging in vore.

Before merging such a thing we would also require a means of dealing with excess nutrition that isn't lifting weights or hitting a punching bag. Those activities both reduce nutriment by a good amount, but they also burn weight, which some characters might dislike. It's also still very time consuming.

I know we have a pill that reduces nutriment in the game already, but I've wanted to alter its mechanics for some time, from a nutriment deletion thing, to a rapid digestion thing that accelerates weight gain as well. All of that is out of scope though and may put this PR on hold.

Either way, the original pills should be made available in the server other than through chemistry.

lbnesquik:
@SpadesNeil This is indeed for another Pr. @Heroman3003 @Screemonster Again, numbers could be simply raised to start later.

Wickedtemp:
Its Lipozine youre probably referring to, its already available outside of chem, but its rarely if ever used because its stupid slow.

I'm kind of against this. I can see it for xenochimera and prometheans, prometheans, who can gain a mechanical advantage via gorging because they can regen using nutrition. But I kinda don't want to get mechanically punished because I like the examine messages.

Screemonster:
Even Tempest's scenario is barely an advantage - promethean regen has a delay on it so it's not spammable, and the chimera's cooldown is literally an hour.

SpadesNeil:
Eh, I don't have a good feeling about this. I really like the idea but it seems to present a problem where players are otherwise enjoying themselves. Yes, it would be balanced, but it also would detract from some fun that the players have.

I'm gonna close this.

PR #4509, opened 2018-11-23
Fix stuffing breaking at extreme numbers.

lbnesquik:
You pervs.


Screemonster:
I find it hilarious that someone clearly ate enough shit for people to notice and bring it up in discord as a bug

PR #4517, opened 2018-11-26
Lower the respawn timer from 5 to 3 minutes.

lbnesquik:
5 minutes is very very long for a game delay.


Arokha:
No vorestation edit comments? And there are other respawn timers and mentions of how long until you can respawn and you only changed this one...

Arokha:
Why don't you convert respawn time to a define or something and make a PR to Polaris for it so we can stop having this problem over the years

KasparoVy:
Had to dig deep

PR #4520, opened 2018-11-27
Fixes Shuttle Transitions

Novacat:
- Fixes Bluespace Transition zone on expedition shuttle
- Fixes Virgo four transition zone on expedition shuttle

### DIFF
--- a/maps/tether/tether_turfs.dm
+++ b/maps/tether/tether_turfs.dm
@@ -164,7 +164,7 @@ VIRGO3B_TURF_CREATE(/turf/simulated/mineral/floor)
 /turf/space/bluespace
 	name = "bluespace"
 	icon_state = "bluespace"
-/turf/space/bluespace/New()
+/turf/space/bluespace/initialize()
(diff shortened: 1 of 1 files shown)


SpadesNeil:
I don't really understand what this fixes, to be honest, which is why I haven't merged it yet.

Is there a PR it's attached to or something?

Novacat:
Basically right now there is a problem, using the expedition shuttle, that bluespace tiles that you see when in bluespace and moving sand tiles that you see when you are landing on Virgo four, do not display at all. Instead of bluespace titles, you just have normal starry sky. Instead of moving sand tiles, you get pitch black instead.

There is no issue for this because nobody bothered to make an issue. However, rather than make an issue I figured I would just fix this.

The root of the problem is that, at some point, Polaris switched from using new() to initialize() on tiles, and, these two particular tiles were not moved over properly. There may be other tiles that might also need fixing, but for now I focused on fixing the two that are known to be broken.

PR #4531, opened 2018-12-01
[MIRROR] Fixes personal ownership of cats

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/5713
--------------------
So, personal fluff cats are supposed to have a friend_name var where only that named character can be befriended, but, that var is not checked anywhere in the become_friends verb. This should fix that, I hope.


Novacat:
VOREStation\code\modules\mob\living\simple_animal\animals\cat_vr.dm requires that var/ is removed from the friend_name var. That will get this to compile.

SpadesNeil:
Need to fix some conflicts that were caused by the AI merge.

Sebbe9123:
Should be a simple fix

Sebbe9123:
Looked through the code and this branch is missing all of the Simple_mob stuff and it's goal it to add to a redundant and unused old system we _just_ stopped using now so I think we can ignore this one.

PR #4536, opened 2018-12-02
Revert "Combine photosensitive trait with darksight"

AaronSherwood117:
Reverts VOREStation/VOREStation#4499

Lazy nerfs with unanswered discussion questions make me grumpy as fuck.

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/traits_vr/negative.dm
+++ b/code/modules/mob/living/carbon/human/species/station/traits_vr/negative.dm
@@ -90,6 +90,12 @@
 	cost = -2
 	var_changes = list("siemens_coefficient" = 2.0) //This makes you extremely weak to tasers.
 
+/datum/trait/photosensitive
+	name = "Photosensitive"
+	desc = "Increases stun duration from flashes and other light-based stuns."
+	cost = -1
+	var_changes = list("flash_mod" = 2.0)
+
 /datum/trait/hollow
 	name = "Hollow Bones/Aluminum Alloy"
 	desc = "Your bones and robot limbs are much easier to break."
--- a/code/modules/mob/living/carbon/human/species/station/traits_vr/positive.dm
+++ b/code/modules/mob/living/carbon/human/species/station/traits_vr/positive.dm
@@ -42,13 +42,13 @@
 	name = "Darksight"
 	desc = "Allows you to see a short distance in the dark."
 	cost = 1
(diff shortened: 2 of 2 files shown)


ArgobargSoup:
Actually, I was pretty okay with the costs staying, because, well...  darksight was already at the lowest it could go.  Lowering the cost would make it free... and then everybody would take it because hey, free trait.  Which would result in it being... well, bleh.  Everybody would have darksight for no raisin.

AaronSherwood117:
> Actually, I was pretty okay with the costs staying, because, well... darksight was already at the lowest it could go. Lowering the cost would make it free... and then everybody would take it because hey, free trait. Which would result in it being... well, bleh. Everybody would have darksight for no raisin.

Your average security member never putting two and two together and pulling the flash out of their pocket really shouldn't have bearing on that.  If Everybody would have it 'for no raisin' then Everybody can enjoy when an event drops a flashbang and KO's the entire room on their metagaming asses.

AaronSherwood117:
> Key word there is "metagaming". It'd become part of the metagame, and I like havin' the trait available. Don't want it gettin' removed in the next attempt at fixing the metagame.

Argo I just want the lazy thing reverted for a second pass that takes longer than twenty seconds of thought.  Pretty much all the traits need a once-over now that some time has passed with the system in place but this ain't it.

SpadesNeil:
No.

We already had the discussion to make this change in the first place.

AaronSherwood117:
> No.
> 
> We already had the discussion to make this change in the first place.
> 
> Tell me, why would we reduce the cost of darksight when we were going to make photo-sensitivity a neutral trait in the first place?

What discussion to make the change? Aronai said he was doing it, did it, and questions got ignore for 20 days before you silently pushed it in.  As usual.  At least have the common decency to acknowledge what happened objectively.  Photosensitivity could be removed without affecting too much, but on TOP of a fairly reasonable and logical trait combination being merged into a single and non-modular effect, it also got pushed into being totally crippling when it comes into play.

"It doesn't come into play often enough to give a point!" isn't valid just because the players in situations to take advantage of it aren't smart enough to do so, and I think it's just an example of a major issue with the current setup we have.

This is just another lazy change that's not going to be given more thought and stubbornly defended because of who did it unless Verk decides he wants to fix it up while his PR sits unanswered for five months.  

I'm more upset with the principle of the behavior than the specific change itself. Either put the time into gameplay balancing or don't pretend to.

PR #4544, opened 2018-12-04
A less snippy automatic response for IC issues.

ArgobargSoup:
The current automatic message given to players when an admin hits the IC Issue button is rather snippy, and not quite fitting to the way our server usually works.  This PR changes the message to something that seems more fitting.  Also fixes a little typo.

### DIFF
--- a/code/modules/admin/verbs/adminhelp.dm
+++ b/code/modules/admin/verbs/adminhelp.dm
@@ -344,7 +344,7 @@ GLOBAL_DATUM_INIT(ahelp_tickets, /datum/admin_help_tickets, new)
 
 	var/msg = "<font color='red' size='4'><b>- AdminHelp marked as IC issue! -</b></font><br>"
 	msg += "<font color='red'><b>This is something that can be solved ICly, and does not currently require admin intervention.</b></font><br>"  //VOREStation Edit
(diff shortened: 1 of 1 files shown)


code/modules/admin/verbs/adminhelp.dm
```diff
@@ -343,8 +343,8 @@ GLOBAL_DATUM_INIT(ahelp_tickets, /datum/admin_help_tickets, new)
 		return
 
 	var/msg = "<font color='red' size='4'><b>- AdminHelp marked as IC issue! -</b></font><br>"
-	msg += "<font color='red'><b>Losing is part of the game!</b></font><br>"
-	msg += "<font color='red'>Your AdminHelp may also be unabled to be answered due to ongoing events.</font>"
+	msg += "<font color='red'><b>This is something that can be solved ICly, and does not currently require admin intervention.</b></font><br>"  //VOREStation Edit
+	msg += "<font color='red'>Your AdminHelp may also be unable to be answered due to ongoing events.</font>"
```
Novacat: I recommend putting a //VOREStation Edit next to the second line, as well.

ArgobargSoup: Makes sense.

PR #4546, opened 2018-12-05
Adds timegate to pathfinder

Screemonster:
Adds a short wait on playing the pathfinder role in keeping with other leadership roles. Still a shorter wait than actual heads given that it's really only the expedition that suffers if they screw up.
Explorer, pilot and SAR are still unrestricted so in theory it's still possible to send out a complete day-0 expedition, and playing explorer instead of PF for a week is pretty low-impact for new players.

### DIFF
--- a/maps/southern_cross/southern_cross_jobs_vr.dm
+++ b/maps/southern_cross/southern_cross_jobs_vr.dm
@@ -18,7 +18,8 @@ var/const/PATHFINDER 		=(1<<13) //VOREStation Edit - Added Pathfinder
 	selection_color = "#AD6BAD"
 	idtype = /obj/item/weapon/card/id/science/head/pathfinder
 	economic_modifier = 7
-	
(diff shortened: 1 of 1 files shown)


Novacat:
My personal preference is that people do not play pathfinder without having some experience in the subordinate roles of the expedition department, IE, Pilot/Field Medic/Explorer. This has 100% my support. I keep meaning to add a timegate to exploration roles but kept forgetting to, in combination of a lot of other priorities coming up. Though, there is an open PR for an overhaul of expedition job roles (#4462).

With the overhaul, I can see Field Medic and Pilot both having a 3 day wait period as well. The main reason for this being shuttle access. Explorers will no longer have shuttle access so they can remain without a wait period.

Novacat:
Currently the only medical jobs with timegates are chemist at 3 days, and CMO at 10 days.

PR #4547, opened 2018-12-05
Fixes Detective Disposals

Mewchild:
Actually connects detective disposals into the network. Oops.


Heroman3003:
Detective office also had problem of missing distro/scrubber pipes under the door. Is that part fixed as well?

PR #4550, opened 2018-12-05
Size control implant

Zap-zapper:
This kinky implant allows people to control your size via voice commands. If you were implanted by someone - only that person who implanted you can control your size. If you implanted yourself - everyone can verbally control your size. You can never control your own size.

Commands:
Shrink: Shrinks host two times smaller
Grow: Grows host 1,5 times bigger
Resize X: Resizes host X times bigger or smaller, OR Rezise X centimeters - makes host X centimeters tall. 170 centimeters is normal.
Ignore: any other keywords in your speech won't have effect

It works on emotes too. For example: /me chuckles. "Kufufu, you are all mine now~" - foo says and smiles. "Let's resize you to, err... 90 centimeters!". If target sees that emote, they will be shrunk roughly two times smaller. Enjoy.

P.S This is all your fault. Do not be silly, you know who you are.


TheDavestDave:
As someone who build this in circuits and now feel their job is threated. You could save so much trouble with a signaler. One grow one shrinks, and one sends signals outside. Plus someone asking. How far is the walk won't ruin a scene.

code/game/objects/items/weapons/implants/implant.dm
```diff
@@ -337,6 +337,90 @@ Implant Specifics:<BR>"}
 			explosion(get_turf(imp_in), -1, -1, 1, 3)
 			qdel(src)
 
+
+//////////////////////////////
+//	Size Control Implant
```
Mewchild: This implant should be moved to implant_vr.dm

PR #4557, opened 2018-12-06
[MIRROR] Fixes several bugs with AI mobs

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/5758
--------------------
Fixes #5751 
Fixes #5750
Fixes #5755 
Fixes #5753 

Cleans up simple_mob ranged shooting code as well.
Also makes it so discipline decay is slowed down by at least a few seconds each time, as otherwise slimes would sometimes lose all of their discipline before the stun wore off, which wasn't the case under the old system.


Sebbe9123:
Tried merging the PR with current master, compiles fine. Conflicts is github being silly since the repository this branch comes from lacks all of the simple_mob system except these 5 files.

PR #4559, opened 2018-12-06
Updates anime eyes marking

ArgobargSoup:
Lowers the pixels on side-facing sprites, making them look a bit better, letting them properly line up with glasses, stopping them from being blocked by hair as much, and allowing them to be a bit fancier without looking different between forward and side-facing.

Example pics:
image
---------------
Fancier
image
---------------
Ingame
image
Top is before, bottom is after.
As you can see, before the change, the original side-facing anime eyes sprites are partially blocked by ears/hairstyle, and appear above most glasses sprites.


KasparoVy:
Could we get a before-and-after side by side? I use these markings on most of my characters for those precious few pixels they let me color and I'm curious about how I'll have to redux things, or if that's not viable and I'd be better off asking for these to be added alongside rather than a replace.

ArgobargSoup:
I provided before and after pics in the original post, but I'll pull up a test server and make a before-and-after from ingame pics.

KasparoVy:
Thanks a ton!

ArgobargSoup:
@KasparoVy Updated OP with an ingame example.

PR #4562, opened 2018-12-07
Disables PDA contamination.

Verkister:
No more messaging list meta.


Heroman3003:
PDA contamination should probably stay, but be unique in the fact that it doesnt actually change the item's name. It even has unqiue mechanic to it, as contaminated pda cant be properly used (at least I believe thats how it was meant to work?)

Verkister:
That mechanic was removed a little while ago as well due to other issues outweighing any fun of said mechanic such as IDs becoming unremovable and so on.

PR #4570, opened 2018-12-09
[WIP] Adds Two New Expedition Areas

Novacat:
WORK IN PROGRESS - DO NOT MERGE

On hold until after the maploader is fixed.

- Adds the Plains, south of the Tether. This contains a number of unused polaris POIs.
- Adds the Debris Field, currently accessible from Virgo 3B Orbit.

Currently not working too well yet, still needs work.


Screemonster:
Debris field is a thing I've been wanting to see for a while. Was gonna fluff it as the crap that's collected in Virgo 3's lagrange point.

I was wondering whether west of mining (and north of the solars/science outpost) would be a good spot for the plains/wilderness area. When we had wilderness south of the tether before we kept getting mobs wandering onto the tether z-level, this would put them a little further out of mob-wandering range while still keeping the area accessible to miners/xenoarchaeologists/explorers.

Unless you're gonna put a big jurassic park gate to keep the mobs in, that works too.

SpadesNeil:
Month old WIP, closing.

PR #4572, opened 2018-12-09
Fixes a runtime with radio.dm

Novacat:
Load ordering was causing the access datums for explorer and pilot access to be defined after radio.dm was loaded, which caused radio.dm to look for access datums that did not exist at the time.

This changes the explorer and pilot access datums to load much earlier, fixing the runtime.


KasparoVy:
Huh, that's a little wacky. How'd you figure to move it like that?

PR #4578, opened 2018-12-11
Fixes Engineering Taur Hardsuits

Darkwolf6180:
Edits the names of several taur hardsuits to be the same as used by code; IE "industrial_rig" -> "engineering_rig"


SpadesNeil:
Do these correspond to the proper code? Because I almost never use taur suits in general.

Darkwolf6180:
It corresponds to the code/icon names used by all the other suits for other species. The industrial/mining/engineering suit were the only suits I can recall editing, and were the only anomalies.

PR #4579, opened 2018-12-11
Fixes at least 1/3 of the surface level cave/ore generation bugs.

### LINKED ISSUE #4425: Surface 1 spawned caves and ores
Rocks on surface one for some reason decided they are mining areas and spawned the cave patterns and ores within them (with some ores glitched)

No ores, no caves, just rock on all station levels.

#### What actually happened
Dangerous cave spawns, ores not meant to be here, glitched out ores...
Example: https://prntscr.com/l83q4r

#### Steps to reproduce
- Fly around surface 1 as ghost with darkness off

#### Code Revision
- 1fe04192c11e2ad7357512d5a873c44a4ea687d3

Verkister:
After some extensive research and loads of test runs, I finally managed to pinpoint the issue to underdark ore generation thingy.
Apparently the original cave/ore generation for underdark wasn't using a solid value for its z-level, and because of this, the initial generation would work fine, BUT in case of the underdark ore roll landing on an insufficient amount of the good stuff, it would reroll the cave/ore generator after having already forgotten what the first roll had given for the z value.

5 runs with fix: no caves/ores ever, 1 startup reported reroll case.
5 confirmation runs without the fix: 4 runs with no caves/ores, 1 run with reroll report that did get caves.

Fixes  #4425

### DIFF
--- a/maps/tether/submaps/_tether_submaps.dm
+++ b/maps/tether/submaps/_tether_submaps.dm
@@ -40,9 +40,9 @@
 
 /datum/map_template/tether_lateload/tether_underdark/on_map_loaded(z)
 	. = ..()
-	seed_submaps(list(z), 100, /area/mine/unexplored/underdark, /datum/map_template/underdark)
-	new /datum/random_map/automata/cave_system/no_cracks(null, 1, 1, z, world.maxx, world.maxy) // Create the mining Z-level.
-	new /datum/random_map/noise/ore(null, 1, 1, z, 64, 64)         // Create the mining ore distribution map.
+	seed_submaps(list(Z_LEVEL_UNDERDARK), 100, /area/mine/unexplored/underdark, /datum/map_template/underdark)
+	new /datum/random_map/automata/cave_system/no_cracks(null, 1, 1, Z_LEVEL_UNDERDARK, world.maxx, world.maxy) // Create the mining Z-level.
+	new /datum/random_map/noise/ore(null, 1, 1, Z_LEVEL_UNDERDARK, 64, 64)         // Create the mining ore distribution map.
 
 //////////////////////////////////////////////////////////////////////////////
 /// Away Missions


Arokha:
If that's the case, isn't the fix in random_map.dm, not here? You're saying the problem is in the random_map.dm generator forgetting what z-level it was given, so fix that. I fail to see how this is the solution for the problem you described.

If this fixes it, that's cool, but it does not fix the problem you described.

Arokha:
In researching further, I think the issue might be in on_map_loaded() not being passed the correct z-level (or no z at all, at which point the noisegen assumes z1 in set_origins(), or maybe later in apply_to_map() which does the same), which would explain why your change would avoid this happening since it ignores the z that's passed.

I think your rerolls are just coincidences, unless those coincide with strange other things.

It would be best if you logged the 'z' that is passed to on_map_loaded to see if it is null, 1, or something zany on those times that it fails to generate on the correct z rather than use this workaround, as this doesn't actually get to the root of the problem.

Verkister:
Yeah it appears that the `new /datum/random_map/etc` for both, V2 and beach caves are also using vague soft z's while in comparison, every other instance of `new /datum/random_map/etc` for every other map is using `Z_LEVEL_WITH_CONCRETE_NAME`

Verkister:
And it appears those two don't even have defined z-levels. How does the set_oringins stuff there even manage when the stuff is set to generate on undefined z-levels

Verkister:
https://github.com/VOREStation/VOREStation/commit/2ce992f4048eb7c237ff6a4cf26e34b0fb04e5c1
kuva
I was unable to find a single instance of there this `mapZ` would exist in the code.

Verkister:
So the root would probably be something along the lines of "In order for map genetation targeted to a specific z-level to properly work, the z-level will need to be actually defined."
Uncoordinated mapgen stuff may be cool for PoIs and shit that are destined to land onto the coodrinates of an existing map, but a whole map cannot be shitflung into the void while expecting it to land on a z-level that doesn't even exist beforehand and not cause z-coordinate related ruckus.

Verkister:
testing a full fix...

PR #4584, opened 2018-12-13
Replaces Ace's ammo with HP

Novacat:
Since 9mm has hollow point now.


Screemonster:
see, when I saw the title I read "HP" as "hit points"

PR #4589, opened 2018-12-15
[MIRROR] Adds Identification System, New Autoinjectors

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/5773
--------------------
In order to continue trying to ~force this game to be a poorly made roguelike~ make exploration more interesting while also making the rest of the station be more relevant for explorers, this PR adds two main features, an identification system, and a bunch of new autoinjectors with random spawners that start unidentified.

In the Future :tm:, Points of Interest are going to receive various pieces of loot that start 'unidentified', meaning that players will not know what it actually does.

Unidentified objects can be identified in a few ways, the easiest being with a specific machine that will be placed in various departments. For now, as only hyposprays can start unidentified, there is only a new chemistry machine that will identify any unidentified reagent-based object.

Later on in the Far Future :tm: :tm:, when more objects start unidentified, other ways of discovering the functionality of an unidentified object will be possible, such as using something blindly that isn't a once-off consumable, or using a rare device that identifies other devices (and hopefully isn't lying to you).

Other things this PR does;
- Maps in a new machine in chemistry. A table was removed in order to make room, the things that were on the table (a box of syringes and a screwdriver) have been moved to the locker in the same room. Chemistry now looks like this: https://i.imgur.com/WpDP5tV.png
- Makes buttons layer over windows so they won't be hidden underneath a window.
- Reagent scanners cannot scan things that have its reagents sealed, like closed beakers or the new big autoinjectors, in order to not trivialize Chemistry's new role.
- Refactors hypospray injection logic to make it cleaner and to make it easier to override to do special things.
- Moves the basic biginjectors out of technomancer files and into the hypospray one.
- Ports the two helper .bat files in mapmerger folder to .sh files so I can use them on my preferred OS.


TheDavestDave:
can't chemistry just make most regants anyway? This looks like busy work for the sake of work. Imaigne just doing your own thing as chemist and all the explorers come rushing demanding you identify this thing. And it's an autoinjector filled with something you just made 6 bottles off. Down the line we might have problems like takeing an engineering, finding something no one knows what it is. And takeing it home to find out it's a screwdriver.

Wickedtemp:
Adding that the chemistry machine sounds more like a Science chemistry thing than a Medical chemistry thing anyway, and should probably be put somewhere in Science's department, ideally where Explorers could get to it, if we use this.

PR #4590, opened 2018-12-15
Removes the mimics from the station map levels

Heroman3003:
The mimics are aggressive mob on the station levels. They don't make much sense to be here, and them being found is always just needless hassle for security and medical since once it happens, person doesn't visit area again. Plus, the fact that you can stumble into them randomly makes it dangerous for actual people trying to explore and learn maint layout for no good reason.

### DIFF
--- a/maps/tether/tether-02-surface2.dmm
+++ b/maps/tether/tether-02-surface2.dmm
@@ -4381,6 +4381,7 @@
 "jG" = (
 /obj/effect/decal/cleanable/dirt,
 /obj/effect/decal/remains/human,
+/obj/random/action_figure,
 /turf/simulated/floor/plating,
 /area/maintenance/lower/rnd)
 "jH" = (
@@ -4441,7 +4442,7 @@
 "jP" = (
 /obj/item/weapon/storage/fancy/cigar/havana,
 /obj/effect/decal/cleanable/dirt,
-/mob/living/simple_animal/hostile/mimic/crate,
+/obj/random/drinkbottle,
 /turf/simulated/floor/plating,
 /area/maintenance/lower/rnd)
 "jQ" = (
--- a/maps/tether/tether-07-station3.dmm
+++ b/maps/tether/tether-07-station3.dmm
@@ -353,6 +353,18 @@
 	},
 /turf/simulated/floor/tiled,
 /area/security/range)
+"aF" = (
+/obj/structure/railing{
+	dir = 4
+	},
+/obj/structure/table/rack{
+	dir = 1
+	},
+/obj/random/maintenance/cargo,
+/obj/random/maintenance/clean,
+/obj/random/maintenance/engineering,
+/turf/simulated/floor,
+/area/maintenance/station/ai)
 "aG" = (
 /obj/effect/floor_decal/borderfloor{
 	dir = 4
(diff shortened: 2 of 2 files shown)


Novacat:
As someone who has played a lot of medical, the mimic on station 3 should probably be removed since newbies are running into it all the time, getting ate, then screaming for help on comms. They do not even have to say location anymore because it is always that mimic. Always.

SpadesNeil:
All of the mimic crates are kept exclusively in areas that cannot be accessed without breaking down a wall.

Really, I wish we had more of these, and thought that we did. :B

But I'll let this one be discussed.

Heroman3003:
Also worth pointing out, that some people make it their goal to every round get to those mimics just to kill them for some reason. Usually security. And since they always spawn in same places, it feels redundantly powergame-inducing to keep them around.

Wickedtemp:
I'm indifferent. The only time the mimics ever annoy me is when someone seeks them out on purpose while "totally calling for help" by screaming over comms without giving a location or setting their sensors, cause it's kinda like... "Ima make you listen to my snuff rp~" and I kinda wish they'd either actually call for help or stay silent but whatever. 

And, well... this is gonna happen anyway, whether with another person, a solar grub, etc, so it isn't caused by mimics directly.

I'm not really sure they're a problem. It's actually been a good while since I've heard of someone actually getting injured by one, and because it's so easy to *not die in the first place*, I don't think they're that huge of a threat.

Heroman3003:
The problematic one really is one on asteroid 3, because unlike one on surface 2, its not 'in the wall' but in the pocket that is always visible and sometimes can be reached without even deconstructing anything, just looking around. But I've once seen a discussion and several people were in support of mimics being gone entirely, so decided to put this one up like that.

Screemonster:
A group of people whining that they died to a completely avoidable thing doesn't mean that feature should be removed entirely. If someone tossed their internals box in disposals at roundstart and later complained about hull breaches being OP and the outside air should be made completely breathable, I'd laugh at them.

We did have people complain about the outside atmosphere when we first started, but instead of changing the planet's atmo entirely to make everything totally safe to walk around outside naked, we beefed up some of the windows, put a big scrubber in the bar (since that's where 90% of breaches were from people breaking the fucking windows) and continue to laugh at people who think it's a good idea to smoke in the fucking airlock.

Wickedtemp:
Tbh, smoking in a station where the outside atmosphere is literally composed of a highly flammable and explosive gas basically just counts as a suicide attempt anyway. Change my mind.

SpadesNeil:
In fairness, I understand why mimics compared to atmos fuckery can be seen as false equivalence, but I don't want to get rid of the mimics. If not mimics, I'd like there to be some kind of on-station threat that's lurking in the halls _if_ you go poking around in places that people normally wouldn't be present. Mimics are my favorite version of that.

That said, it doesn't _have_ to be mimics, but I'd like something _unique_ to be infesting the gaps between the walls.

Mewchild:
The issue I have with the mimics is in that the people who most often encounter it are the ones who know it's there, as a fact. It's not a random chance for these individuals, it's simply giving medical the chore of "Patch me up or resleeve me," normally after calling out on comms that they're being mauled. I can understand it happening to someone just exploring, but it's hard to feel sympathy or even the will to get the person back in round when they've died to the mimic five days in a row.

SpadesNeil:
Okay, that's a valid point... I would like to have mimics return to the station, but done in such a way that they are no longer predictable, while at the same time not being _too_ easy to stumble on by accident.

Screemonster:
Having them on a lootspawner table and only putting them on the ones that appear inside walls and the like would help, so on any given shift the loot in a particular place _might_ be a mimic or it might be good loot (or good loot inside a mimic), the main thing I'd need to check is whether mimics show up on material scanners 'cause if they don't then people are just gonna see an empty space and be like "oh, that spawn must be a mimic today"

PR #4599, opened 2018-12-19
Sweeping AI QoL

### LINKED ISSUE #4595: Turrets outside AI core are aggressive.
While the stairs and the turrets being present up there instead of the ladder in new upload->core transition is nice, turrets outside core where they were not before are not tied to same console as turrets downstairs. They will aggro at you for just going up when you are technically still in same room, just a level higher?

Turrets outside core itself to not be there or to behave in match with turrets downstairs

#### What actually happened
Almost got shot.

#### Steps to reproduce
Enter upload, see turrets not shoot you. Go upstairs, but not into core itself, see them ready to fire, bolt.

#### Code Revision
fe04192c11e2ad7357512d5a873c44a4ea687d3

Mewchild:
Alright. Lots of tiny changes in this one.

- Adds a Turret control to the AI core, linked to the two turrets at the top of the stairs
- Sets Broadcast on the Public Garden Floor 3 intercomm to 0 (stops the mic from being on at round start)
- Adds several holopads across the station, improving coverage for the AI
- Adjusts the locations of/adds several cameras in public areas, removing inconvenient blind spots. These notably don't affect private spaces, such as the dorms/dorm hallways/laundry room
- Added a 5% increased chance of sexy AI interactions, across the board

Holopad coverage and strange camera blindspots were pointed out by Ech0 (Ismuth#3876).

Fixes #4595

### DIFF
--- a/maps/submaps/engine_submaps/engine_sme.dmm
+++ b/maps/submaps/engine_submaps/engine_sme.dmm
@@ -1,213 +1,2614 @@
-"aa" = (/turf/template_noop,/area/template_noop)
-"ab" = (/turf/space,/area/space)
-"ac" = (/obj/structure/lattice,/obj/structure/grille,/turf/space,/area/space)
-"ad" = (/obj/structure/lattice,/turf/space,/area/space)
-"ae" = (/obj/machinery/atmospherics/pipe/simple/heat_exchanging{dir = 6},/turf/space,/area/space)
-"af" = (/obj/machinery/atmospherics/pipe/simple/heat_exchanging{dir = 4},/turf/space,/area/space)
-"ag" = (/obj/machinery/atmospherics/pipe/simple/heat_exchanging{dir = 4},/obj/structure/lattice,/turf/space,/area/space)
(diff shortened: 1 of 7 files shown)


Mewchild:
It seems that if you touch a number of files, and the images MapDiffBot tries to create are large, it cries and gives up due to the combined filesizes

The Camera changes were minor, and only affect areas that were in public halls, where the cams just didn't reach for no good reason. 
image
image

Holopads were added in medical, CMO Office, QM office, Numerous places in the halls to fix gaps in coverage. Pads in the bottom floor atrium were rearranged to fix coverage gaps.

Very-Soft:
I should note that in the base of the public garden, the camera placement was intentional, so you could sneak into the space behind the scrubbers without being spotted on cameras. It's not really all that important, but I did make that place specifically to be a sort of sneaky spot.

Mewchild:
Removed the extra camera from Garden 1, as per @Very-Soft's Comment. Added a camera into the SM engine, to fix a critical blindspot that prevented accessing the cooling loop and APC.

SpadesNeil:
Mapping errors galore.

PR #4605, opened 2018-12-21
Added some missing air alarms and firelocks around station

Heroman3003:
A leak happening here would be a disaster right now, wouldn't it?


Screemonster:
it's on the space levels so a breach would be low pressure, which is far easier to deal with than most. Adding air alarms to sections that are lacking such is a good fix though.

Heroman3003:
Tbh, those areas are pretty damn secure from breaches. The problem comes when a 'gas leak' random event pops up. And as a general thing, almost all airlocks should have shutters and almost all sealed areas should have air alarms.

Screemonster:
the thing that makes leaks in maint a pain is the fact that maint has so many vertical shafts, so a gas leak can easily spread across multiple floors and in some cases bypass firelocks. Luckily gas leaks there are fairly rare.

Heroman3003:
If anyone knows any other places where there should be air alarm/firelock but isn't, feel free to comment.

Heroman3003:
Closed as of #4606 conflicting.

PR #4610, opened 2018-12-23
Fixed creation of borgs through non-spawn method

Heroman3003:
I have no idea why that unfinished var even exists, considering its use is completely commented out, but its Polaris stuff, so I just brought our modified proc to match.

### DIFF
--- a/code/modules/mob/living/silicon/robot/robot_vr.dm
+++ b/code/modules/mob/living/silicon/robot/robot_vr.dm
@@ -166,7 +166,7 @@
 	buckle_movable = TRUE
 	buckle_lying = FALSE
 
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/silicon/robot/robot_vr.dm
```diff
@@ -166,7 +166,7 @@
 	buckle_movable = TRUE
 	buckle_lying = FALSE
 
-/mob/living/silicon/robot/New()
+/mob/living/silicon/robot/New(loc,var/unfinished = 0)
```
KasparoVy: Dang, and Polaris didn't even space the args eh?

Issue #4419, opened 2018-10-18
WHAT THE HECK DID YOU DO TO THE TESLA


Very-Soft:
#### Description of the issue
The tesla is producing all in all significantly less power than it did previously, despite the fact that the new configuration has more coils, and makes the tesla always in the PA beam, so it keeps growing even when power is set to 0.

And with all that, it seems to be producing on average around 700-800kW, and I have no idea why. It seems like something's different about how it's calculating power, as previously, with two less coils and a bigger field, which allowed the tesla to scoot out of the path of the PA and shrink, it would easily produce in excess of 1 mW. Which, considering the lack of real customization, and lack of proper protection gear for working around it seemed reasonable. 

Now, the default setup should be even better than it was before, which I'm not too keen on, but it ain't. :Y

#### What you expected to happen
To be able to set the main smes output higher than 650kW without draining the thing.

#### What actually happened
The main drains under just a little more than normal station load.

#### Code Revision
- release - 
1fe04192c11e2ad7357512d5a873c44a4ea687d3
#### Anything else you may wish to add:
- I heard the server tickrate got changed again. Perhaps that has something to do with this. I haven't checked to see if the SM behaves similarly.

Shophaune:
Ohhhh, is that it? I noticed power issues after setting up the tesla but I thought that was just from everything trying to charge constantly whilst the tesla output in bursts.

TheDavestDave:
it sounds like, if code hasn't change. Someone wired a coil in wrong

Shophaune:
Can't be that, tesla was very clearly putting out 4+MW in bursts.

Iroquois-Pliskin:
Tesla has been A-okay in my experience and now averages HIGHER power outputs as well. The smaller area means it has less things to arc to, other than grounding rods and coils. The one issue I've seen is it can occasionally arc to someone standing in front of the PA inside the PA room.

Heroman3003:
The problem I noticed with tesla is that regularly, if its not overgrown out of proportion, it only zaps coils on ONE side instead of both because of their new positions.

Also, yes, it now continues growing even with PA on zero, which is horrible.

Very-Soft:
All I know is that I did my usual setup this morning, and ended up with a dry main after a little bit, despite putting the output on it 50kW lower than the input like I normally do. I fiddled with it, and ended up throttling main output down to 650kW before it started taking more power in than it was putting out. 

It should not be that low with a 19 ball tesla with 6 coils.

[...]

PR #4621, opened 2018-12-31
Adjustments to the tesla submap

Heroman3003:
Several changes to the tesla submap, to account for some bugs and general issues.

- Changed the positioning of grounding rods: a big important factor in default positioning on old tesla submap was that engine room was always safe. On new one that is not the case. Engine room should be secured by coil positioning first and foremost. The tests I ran seemed to prevent zapping inside engine room, even when hugging the window, unlike the ones on current.

- Changed positioning of tesla coils: main problem with new positioning is that their separation creates a split in way lightning works and overall causes output to be either weirdly low or weirdly high. This both makes power workings inconsistant and increases risks of wild zaps passing through badly thought-out rod positioning.

- Removed phoron canister and tanks from mini-storage: one of the goals of new tesla was moving away from things that were leftovers of Singulo submap. Those two were last. Instead there is now additional table.

- Replaced the coil boards and capacitors with grounding rod boards: the two additional coils were added on the basis of 'engineers occasionally construct them using spare parts', but said spare parts were left too, basically giving free coils. Spare rod boards still provide utility, but do not directly increase efficency of the engine.

- Returned a screwdriver, put along with cable coil to storage, added a wrench: more neat engine room and all three tools for PA setup are now present, just located in little spares room.

- Increased field radius: it will be bigger than current field, but still smaller than old, and still supported by just four barrier generators and four emitters. The internal area will be 5x5, allowing more manageable tesla grows on lower PA settings. More relevant with coils providing inconsistant output being fixed.

- Return of spaceside junk (wirecutters, extinguisher, hard hat, welder): can be easily removed back, but I do believe it gave space engine area more aesthetic of the work zone.

- Added missing firelocks under PA room and mini-storage doors: why were they missing in the first place?

### DIFF
--- a/maps/submaps/engine_submaps/engine_tesla.dmm
+++ b/maps/submaps/engine_submaps/engine_tesla.dmm
@@ -13,8 +13,8 @@
 "am" = (/obj/structure/cable/cyan{d1 = 4; d2 = 8; icon_state = "4-8"},/obj/structure/cable/cyan{d1 = 2; d2 = 4; icon_state = "2-4"},/turf/simulated/floor/airless,/area/space)
 "an" = (/obj/structure/cable/cyan{d1 = 2; d2 = 8; icon_state = "2-8"},/turf/simulated/floor/airless,/area/space)
 "ao" = (/obj/structure/closet/emcloset,/obj/machinery/embedded_controller/radio/airlock/airlock_controller{frequency = 1379; id_tag = "eng_north_airlock"; pixel_x = 24; pixel_y = 0; req_one_access = list(10,11); tag_airpump = "eng_north_pump"; tag_chamber_sensor = "eng_north_sensor"; tag_exterior_door = "eng_north_outer"; tag_interior_door = "eng_north_inner"},/turf/simulated/floor/tiled/dark,/area/engineering/engine_room)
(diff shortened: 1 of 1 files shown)


Heroman3003:
#4419 Should partially account for stuff listed here, possibly.

Very-Soft:
That field layout should allow for the tesla to stop growing on 0 power again as it's intended to do, which is ideal, and the space between the coils and the glass is great! I think this should be a marked improvement over the current design.

Very-Soft:
And yeah, the coils on the front end being closer together should help connect both sides. Were you able to confirm that the arcs regularly transfer over that gap?

Heroman3003:
Travis is being a weirdo... Thing ran and still runs perfectly on the local...

Heroman3003:
@Very-Soft Correct, it will always connect now. Before the distance between coils on two sides was just too big for it to happen. Which is why it either produced too much power (hitting both sides as separate thing) or too little (hitting just one side).

nik707:
A screenshot of the finished map would be sick

Very-Soft:
There's autogenerated screenshot things in the map diff renders of the before and after.

SpadesNeil:
when will you fuckers start including screenshots in your map PRs?

Heroman3003:
Added screenshots both ingame and by Mapdiffbot.

PR #4622, opened 2018-12-31
Adds an Admin Subspace Jammer for events

Novacat:
I plan on doing a stationary version down the road but that will require a lot more work.


DemgelZalvine:
👍

PR #4627, opened 2019-01-02
Holocollar Addition to Autolathe 

TheFurryFeline:
Ports from ChompStation an addition to the Autolathe: Holocollars at no apparent cost in materials.

### DIFF
--- a/code/datums/autolathe/general.dm
+++ b/code/datums/autolathe/general.dm
@@ -122,3 +122,7 @@
 	name = "handcuffs"
 	path =/obj/item/weapon/handcuffs
 	hidden = 1
+
+/datum/category_item/autolathe/general/holocollar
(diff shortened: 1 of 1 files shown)


Screemonster:
wait, so they just spawn with literally zero cost? Even if it's a _small_ material cost the autolathe shouldn't be able to generate anything out of literally nothing.

KasparoVy:
And for my next trick...

Heroman3003:
I agree with @Screemonster on that one, since currently only one item in autolathe uniquely costs nothing and thats autoinjector, a one-use item.

TheFurryFeline:
Yeah, @Screemonster. Gonna close this and do a proper thing. I overlooked something when porting this.

PR #4628, opened 2019-01-02
Fixes shellguard-noble sprites not actually having sprites

Heroman3003:
Seems like either at some point this file got reverted... or I made a big whoopsie when re-adding them in the first place.

Credit for them still goes to Travelling Merchant, first uploaded on Paradise Station.


Heroman3003:
That only leaves 'insekt' borg sprites without sprites, but thats actually polaris-level issue.

PR #4632, opened 2019-01-03
Holocollars in Autolathes

TheFurryFeline:
You can now create the Holocollar for multiple attachments to your suit slot at a low steel cost. Tested and it works.

Edit: This ports a pair of PR's from ChompStation.

### DIFF
--- a/code/datums/autolathe/general.dm
+++ b/code/datums/autolathe/general.dm
@@ -122,7 +122,3 @@
 	name = "handcuffs"
 	path =/obj/item/weapon/handcuffs
 	hidden = 1
-
-/datum/category_item/autolathe/general/holocollar
-	name = "Holo-collar"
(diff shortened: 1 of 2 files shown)


code/datums/autolathe/general.dm
```diff
@@ -122,3 +122,7 @@
 	name = "handcuffs"
 	path =/obj/item/weapon/handcuffs
 	hidden = 1
+
+/datum/category_item/autolathe/general/holocollar
```
SpadesNeil: Put this in general_vr.dm

TheFurryFeline: Like just add an entirely new file with that? Hmm....

SpadesNeil: yes

(Prevents conflicts is why.)

PR #4642, opened 2019-01-06
Boost Belly Limit

TheFurryFeline:
There will now be a maximum of 30 bellies for those crazy enough to spend the time and those too lazy to use text documents for that. >W>

### DIFF
--- a/code/modules/vore/eating/vorepanel_vr.dm
+++ b/code/modules/vore/eating/vorepanel_vr.dm
@@ -2,7 +2,7 @@
 // Vore management panel for players
 //
 
-#define BELLIES_MAX 20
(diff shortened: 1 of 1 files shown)


Wickedtemp:
Would support. Currently at like 18.

TheFurryFeline:
Aye, Tempest. My currently most-used char's been maxed out for a while. Would love to add more bellies. >W>

SpadesNeil:
okay but why tho

TheFurryFeline:
Like I said in the comment, it'll allow those who want to spend all the time in making a slew of bellies. This is a low effect PR that'll just allow people to add more bellies.

PR #4646, opened 2019-01-07
Reduces costs of common NIFSofts

Novacat:
Most NIFSoft costs cut by half. The soulcatcher was not included as it is already dirt cheap, and illegal NIFSoft programs were not included for obvious reasons.

It is kind of really strange that common available NIFSofts run in nearly the thousands of thalers, which, is a lot of money for most jobs. This should help alleviate that.


Screemonster:
Don't most departments start with nifsoft disks so the people who work there can get department-related NIFs for free?

Novacat:
That is correct, but the department NIFSoft disks are usually very basic, and there are programs not found on those disks that may be desired.

Ketrai:
Nifs being so expensive is supposed to reward people for doing jobs. So they stard with enough money on their card to buy a few discs. Or alternatively, the CD can grant funds for special jobs, allowing one to purchase these nifsofts. 

In practice however, the high cost is indeed a bit of a hassle. Especially with the rng, meaning some folks can afford certain nifs some shifts, but not others. It's also not really nice for jobs that pay badly, like service jobs... THey often can't even get mass alteration if they get a bad roll. There's no reliable way to gain more funds except stealing and trash digging either. So, until ways are added... I think it's better they be cheaper. Otherwise you just encourage people to play a job, do jack shit, get some fancy softs, and go scene off in the corner of pluto.

Zeldazackman:
This PR would make life quite a bit easier as previously stated service oftentimes cannot afford to purchase the mass alteration NIFsoft among most other things due to their usually low pay rates (So scenes become a bit clunky at times when sizeplay is involved when you've got to crawl out of bed mid rp to go zap yourself with a size gun stashed away in a dufflebag across the room).

That being said another way to fix this that was floating around a while back in the discord would be to track account balances in a persistence method across rounds, however that'd require quite a bit of a headache to set up a database even for such "a simple thing".

KasparoVy:
Going the route of persistence would have to be handled with immense case as it's very prone to getting hella gamed the sake of being uber riche in monopoly money

Sebbe9123:
To go a bit further with how gamed persistant money could be, a HoP or CD could pull out 75K every round and boost someone's account by that every single round.

Wickedtemp:
Honestly, I don't think a persistent economy is worth it. How would you determine pay? Do people get paid more if they do their job? How does the code determine whether or not they've done it? Take chemistry for example. You'd think it'd be easy, make chems, get money. But, some chems are easier to make than others. If it's all one flat value, I'd just make my usual stock and then 100 bottles of water to get a pay boost. How would the game count the bottles? How would the game know if the bottles are full or just empty and labeled? How would the game know who made the bottles to start with? 

Okay, that's too much to specify. Let's just say you get paid based on time spent in the lab or medbay. Now, you'll have good chemists who can finish everything in ten minutes getting paid nothing, and slower chemists raking in mad cash for taking three hours to make ten bottles. Or they'll just afk. 

What about Doctors who do chemistry because there's no chemist, should they get paid extra? Should that be said for Doctors who take 'command' of the department being paid extra when there's no CMO? Let's assume that we fix all of this and that it all works perfectly...

Excellent, let's do that again, for every single job. 

Persistent economy is just a bad idea. A fuckton of effort to put in without making it so easy to get cash that it's useless, or just broken and guaranteed to screw everyone over unless they powergame it to avoid IC cash issues. Which, full disclosure, I'd probably do to an extent that I can buy protein shakes every round for weight gain fun and not steadily lose all of my money.

SpadesNeil:
>Persistent economy is just a bad idea. A fuckton of effort to put in without making it so easy to get cash that it's useless, or just broken and guaranteed to screw everyone over unless they powergame it to avoid IC cash issues. Which, full disclosure, I'd probably do to an extent that I can buy protein shakes every round for weight gain fun and not steadily lose all of my money.

This is indeed _why_ we don't have a persistent economy.

We could have it. We just don't. I don't want it.

PR #4647, opened 2019-01-07
Adds Alarm monitor NIF program

### LINKED ISSUE #4541: Alert monitor doesnt exist as NIFSoft
#### Brief description of the suggestion
The NIFSoft soft selection lacks an alert monitor for engineers to use. There probably should be one

#### Why should it be added
The functionality for portable alert monitor already exists with AR-E. There is no reason that a functionale from AR glasses shouldnt exist as a NIFSoft (doesn't even have to come in engineering package disk, just like mesons doesn't). In general I find it weird that its one specific thing that AR can do and NIF can't.

Novacat:
Alarm monitors for your engineering staff. Alarm monitor is set to engine access.

Has the same pricing as the Crew Monitor program.

Fixes #4541

### DIFF
--- a/code/modules/nifsoft/nifsoft.dm
+++ b/code/modules/nifsoft/nifsoft.dm
@@ -242,7 +242,7 @@
 	stored = /datum/nifsoft/package/security
 
 /datum/nifsoft/package/security
-	software = list(/datum/nifsoft/ar_sec,/datum/nifsoft/flashprot)
+	software = list(/datum/nifsoft/ar_sec,/datum/nifsoft/cameramonitor,/datum/nifsoft/flashprot)
 
 /obj/item/weapon/storage/box/nifsofts_security
 	name = "security nifsoft disks"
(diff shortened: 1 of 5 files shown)


SpadesNeil:
Camera monitors, no. Alarm monitors, that's totally fine.

code/modules/nifsoft/nifsoft.dm
```diff
@@ -242,7 +242,7 @@
 	stored = /datum/nifsoft/package/security
 
 /datum/nifsoft/package/security
-	software = list(/datum/nifsoft/ar_sec,/datum/nifsoft/flashprot)
+	software = list(/datum/nifsoft/ar_sec,/datum/nifsoft/cameramonitor,/datum/nifsoft/flashprot)
```
SpadesNeil: I would not want the ability to access cameras available on a nif.

TheFurryFeline: As seen from testing, it only shows camera alarms

Heroman3003: Those are basically "Camera Alarm Monitor" so it will only show list of cameras that are broken and cameras with motion sensors detecting motion.  It will not allow to actually view cameras.

Novacat: It just allows you to check camera alarms, which is a feature already available on the AR-S glasses.

SpadesNeil: Then I misunderstand what this is and also I shouldn't post on Github when I am half asleep. :U

But still I don't like the idea of a nif allowing you to be alerted to camera alarms.

Issue #4382, opened 2018-10-14
Wrong Windows in Engine Room


Iroquois-Pliskin:
#### Brief description of the issue
The windows overlooking the Engine's cold loop/Surrounding the Tesla's PA are currently the wrong type, and make it very easy to vent the engine

#### What you expected to happen
Normal full tile windows

#### What actually happened
Weak little baby one side windows.

#### Steps to reproduce
- Join as anything with Engineering access
- Approach the engine room
- Behold the windows.

#### Code Revision
1fe04192c11e2ad7357512d5a873c44a4ea687d3 
 
#### Anything else you may wish to add:
https://i.imgur.com/L3YscRS.png

Edit: Including Tesla in the issue post as well.

Screemonster:
Tesla appears affected as well as supermatter.

Iroquois-Pliskin:
> Tesla appears affected as well as supermatter.

Confirmed. Screenshot to boot.
https://i.imgur.com/KMhQLHm.png

Sebbe9123:
Just took a peek at the aerostat, same problem there.
https://gyazo.com/d4f6df0f0f7687ed9a59ee2c7c3ca7e9

Iroquois-Pliskin:
It has come to my attention that on the tesla submap, these windows leave electrified grilles exposed.

This is dangerous for inattentive engineers and victims of pushing

Verkister:
Pretty sure someone's managed to release some sorta **severe** fuckup in startup rng map loading that might also be related to the squished cave POIs and holodeck windoors and shit.

https://i.gyazo.com/d8b8f0046ee7f7e4f8091033fab49511.png

KasparoVy:
Ruh roh.
I'm confident in the single pane separating the room from a fiesta so long as we don't breathe on it too hard.

Verkister:
https://github.com/PolarisSS13/Polaris/pull/5741

Novacat:
This has been fixed. Closing.

PR #4648, opened 2019-01-08
Adds two new Expedition Areas

Novacat:
- Plains, which can be found south of the Tether. There is an all access gate. Mostly designed for solo explorers, but there are a few tough encounters that can generate.
- Debris Field, which is, as of now, empty, but the infrastructure is fully set up for people to make POIs for it.
- Slightly boosts the POI spawns for the other expedition areas.

Known issues:
- There probably are a bit too few mobs in Plains right now, but, I need a full list of the mobs that can survive Virgo 3B's atmosphere in order to populate it.
- Going off the edge of Debris Field is supposed to teleport you to the other side, as what happens on the Tether. However, I have no idea how to do that.
- The Plains POIs are specially vulnerable to #4382, nothing I can do there until the core issue gets fixed.


SpadesNeil:
Doesn't need staff discussion imo. Just needs a person to check the code.

PR #4661, opened 2019-01-12
Removes the usage of non-working is_tool procs

Heroman3003:
Removes vorestation edits in polaris files that used it and replaces the proc in vr files that used it. Whoever thought this was a good idea and wanted to make it work can put it back in when they come back to the project of making all tool-type checks into procs, until then I'd like functioning machinery, thanks.

Thermal regulator and R-UST stuff will be affected, right now you cannot multitool them for configuration.

Also removes use of is_welder and is_cable_coil

That should fix construction/deconstruction of light switches.

### DIFF
--- a/code/game/machinery/airconditioner_vr.dm
+++ b/code/game/machinery/airconditioner_vr.dm
@@ -45,7 +45,7 @@
 			disconnect_from_network()
 			turn_off()
 		return
-	if(I.is_multitool())
+	if(istype(I, /obj/item/device/multitool))
 		var/new_temp = input("Input a new target temperature, in degrees C.","Target Temperature", 20) as num
 		if(!Adjacent(user) || user.incapacitated())
 			return
--- a/code/modules/power/fusion/core/core_control.dm
+++ b/code/modules/power/fusion/core/core_control.dm
@@ -11,7 +11,7 @@
 
 /obj/machinery/computer/fusion_core_control/attackby(var/obj/item/thing, var/mob/user)
 	..()
-	if(thing.is_multitool()) //VOREStation Edit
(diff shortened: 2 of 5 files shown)


SpadesNeil:
"This feature is broken, so I'm going to delete it completely."

Heroman3003:
I saw. It always returns false. There is no item that has it changed to return true. Places it was used were all _vr only files, while all polaris files still used istype. Its both for sake of consistensy and for sake of making stuff work. If it IS to be implemented, then its better done at polaris level. Right now I'm just removing what seems to be a feature that was beginning to be implemented and then abandoned halfway through. Possibly it was removed originally on polaris level and appropriate VR files were not updated.

Heroman3003:
To add to that, I am NOT removing the proc. Just its usage is replaced with working counterpart. If somebody ever wants to move multitools to proc-based checking, all that would have to be done is just as simple. For now it is primarily getting rid of unnecesary polaris divergence in R-UST files and that leaves just one single use in the Thermal Regulator over entirety of the code.

SpadesNeil:
Okay that's a fair counter argument.

Heroman3003:
Some small looking into polaris' history, it seems like is_multitool, along with is_welder and is_wirecoil were added in the tool refactor, but never applied to anything because they shouldn't have been used at all, those three tools carrying unique functionality. Meanwhile, screwdrivers, wrenches, wirecutters and crowbars did get that.

Heroman3003:
Aparently, is_multitool wasn't only non-working proc leftover in _vr files. Lightswitch construction, thing unique to virgo, also used the phased out is_cable_coil and is_welder.

PR #4673, opened 2019-01-13
Unfucks mirror 5713

Novacat:
Basically makes it so that it compiles and removes the conflicts with virgo code.


Novacat:
Fixes #4531, cannot compile as it requires the AI Branch.

SpadesNeil:
Of course, this is all worthless because the AI branch is breaking this anyway. /shrug

Not sure why I merged this, in hindsight. I'm half asleep and it's 1AM.

PR #4709, opened 2019-01-22
Fixes Spirit for AI Overhaul

Novacat:
- Also ports Virgo's changes to cat.dm
- Also fixes existing runtimes


SpadesNeil:
I think the become friend thing gets fixed in the very next update. https://github.com/VOREStation/VOREStation/pull/4531

Novacat:
That PR edits the old cat.dm and is no longer used at all, and is one of the PRs I did on Polaris, I am not even sure why it got merged.

I spoke with Neerti recently and apparently removing the job check on Runtime's become_friend verb was intentional, so, there is nothing to send to polaris. The question is if I should keep the PR as is, or if we should go in the same direction and remove the job check. The friend_name check will stay as I am using it for one of Ascian's fluff items.

SpadesNeil:
Remove the job check.

Novacat:
Updated this with the latest version, it should be ready to merge.

PR #4725, opened 2019-01-28
Adds most other voremobs

Sebbe9123:
So this commit adds 90% of the old voremobs we had EXCEPT for:
Oties, I really can't figure it out.
Solargrubs and larva, Same as above, over my skillevel.

The Rat file is also slightly gutted with the food-code commented out as it is also way to advanced for me, sorry.

I'm sorry for fixing stuff 90% but not 100 making more work for the future but I really don't know advanced DM :/


Sebbe9123:
Also the ZZ_vore_overrides file is fundamentally broken but there is some stuff there that had to do with my PR so I added it in, best not to compile with it.

Sebbe9123:
Forgot some silly things, ooooone second.

PR #4752, opened 2019-02-07
[MIRROR] Flash rebalance.

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/5907
--------------------
Tweaks:
- Flashes again run a non-negligible risk of burning out through consecutive use.
- Burned out flashes with repairs enabled can have repairs attempted [30% success] by using a screwdriver on them over the course of approximately fourty to sixty seconds. This is not on default flashes due to the sheer number available to Security at round-start. [22 not including on-spawn equipment, 7 in mounts or deployables.]
- Flashes now run on a non-removable internal cell in addition to their capacitor cooldown. 12 charges.
- Cyborg flashes do not burn out or limit use, but instead consume larger and larger amounts of power with each use over the limit [Around 12 flashes began eating percentages of a standard cell.]


SpadesNeil:
Not a fan of this, will probably not merge it.

Screemonster:
This is one of those things where I figure burnout or battery would be fine, but both is silly.

Wickedtemp:
Agreed. Flashes are already subpar. They're kind of made to be. It doesn't take much to get protection from flashes so their utility to start with is already questionable.

And this just kind of makes them too much of an rng gamble to use. It wouldn't go well with our trait system, either, I don't think. Don't we have increased vulnerability to flashes as a negative trait? Or, like, we were planning on it, weren't we? This would make it an easy way to get trait points, because you'd be vulnerable to something that: A) Is utter trash and will break if you use it a few times, and B) Will be meta'd out of use due to A anyway and won't be a relevant threat most of the time.

Novacat:
Due to code standards, we have to merge it, but we can pretty much override the changes with ample use of //VOREStation edit.

PR #4765, opened 2019-02-12
Replaces chem dispenser in xenoflora with one that only dispenses botany-related chems

Heroman3003:
Research department has a pretty long history abusing the chem dispenser in xenoflora for nearly anything except xenoflora most of time. Dodging having to go to medical by brewing any medicine needed primarily. While I'm not saying 'only chemists should know how to chemistry', purpose of it being in xenoflora lab always seemed to me as more "get mutagen and fertilizers without need for chemists" rather than "get whatever you can imagine by doing it with no need of involving chemists".

Reagents it produces (8 total):
Water
Sugar
Ethanol
Radium
Ammonia
Diethylamine
Plant-B-Gone
Unstable Mutagen

### DIFF
--- a/code/modules/reagents/dispenser/cartridge_presets_vr.dm
+++ b/code/modules/reagents/dispenser/cartridge_presets_vr.dm
@@ -1,0 +1,6 @@
+/obj/item/weapon/reagent_containers/chem_disp_cartridge
+	//Xenoflora
+	ammonia		spawn_reagent = "ammonia"
+	diethylamine	spawn_reagent = "diethylamine"
+	plantbgone	spawn_reagent = "plantbgone"
+	mutagen		spawn_reagent = "mutagen"
--- a/code/modules/reagents/dispenser/dispenser_presets_vr.dm
+++ b/code/modules/reagents/dispenser/dispenser_presets_vr.dm
@@ -1,0 +1,30 @@
+/obj/machinery/chemical_dispenser/research
+	name = "xenoflora chem dispenser"
+	spawn_cartridges = list(
+			/obj/item/weapon/reagent_containers/chem_disp_cartridge/hydrogen,
+			/obj/item/weapon/reagent_containers/chem_disp_cartridge/lithium,
+			/obj/item/weapon/reagent_containers/chem_disp_cartridge/carbon,
+			/obj/item/weapon/reagent_containers/chem_disp_cartridge/nitrogen,
+			/obj/item/weapon/reagent_containers/chem_disp_cartridge/oxygen,
+			/obj/item/weapon/reagent_containers/chem_disp_cartridge/fluorine,
+			/obj/item/weapon/reagent_containers/chem_disp_cartridge/sodium,
+			/obj/item/weapon/reagent_containers/chem_disp_cartridge/aluminum,
+			/obj/item/weapon/reagent_containers/chem_disp_cartridge/silicon,
+			/obj/item/weapon/reagent_containers/chem_disp_cartridge/phosphorus,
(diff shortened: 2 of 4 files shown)


Screemonster:
no fun allowed

Very-Soft:
Just gonna put my 2 cents in;

The xenoflora chem dispenser is the only real place where you can practice and quietly make pretty much whatever you want, without the stress and potential urgency and toe stomping that tends to happen in chemistry proper.

Passed that, the main think people do with that dispenser is make chlorol for scenes.

I don't really see people using this to powergame usually, and I think it's pretty lame to start removing things just because they aren't needed/could potentially be used to powergame.

It makes sense to me that science would have a chem dispenser to experiment with, and this one has basically never been an issue as far as I've seen, so I'm not really sure why it needs changing.

Heroman3003:
If there is solution that does prevent situations where instead of going to medical scientist comes by xenoflora, does full set of standard medchems like a pro and chugs it like a series of shots, without shutting down anything else, I'd be down for that.

Because I have in fact seen scientists repeatedly use that as their personal self-use medical instead of medical, even in circumstances where there was no true need to use it that way.

Screemonster:
The solution is to fucking ahelp them, not gut a fucking department for anything that might lead to fun.

Screemonster:
Also this is just going to lead to chemical cartridges being added to science's roundstart ordering list if they want to powergame with it.

Novacat:
I would say this is not a huge issue as it can be at first glance, since, a lot of the most important chems can only be made with phoron, and even powergamers are not going to go through the trouble of getting liquid phoron to make peridaxon, carthatoline, and dexalin plus.

On the flipside of this, this is also something that is very easy to powergame and get away with it, without anyone even noticing unless they are really blatant about it. If you just quietly make the chems, put them in your bag, nobody is ever going to know. Unless you brag about doing it, but at that point you are asking to be ahelped pretty much.

nerdass:
this is fucking stupid, you don't even play xenobotany, why are you trying to make balance decisions about it. let the one true mad science bit that science has stay mad

SpadesNeil:
nope

Very-Soft:
I don't actually see a problem with someone who's familiar with chems using their skills to fix their fuck ups in their dangerous jobs without needing to pester medical any time they get an ouchie.

I'd have a problem with scientists trying to go play medical for people outside of science accidents, sure. 

But like, I've literally never seen that happen. Most people use xenobotany's chems for practice, experiments, and scenes.

If someone's trying to play doctor chemist man for the station, but is in science, then please ahelp them! But that's by no means regular behavior.

nerdass:
this shouldn't get merged period, no "requires discussion" at all needed, it's a bad PR submitted in bad faith, we shouldn't humor it.

Screemonster:
Frankly I'm starting to tire of the notion that all tasks "belong" to one and only one department and this PR is emblematic of that in-need-of-a-dick-kicking mindset.

If nothing else, if you try and tell me the _medbay_ players are seriously gonna ever do something as crazy-fun as filling bluespace beakers with nitroglycerine mix and praying for the bombcap to be removed as they toss it into the quarry, I'm going to laugh at you because that's basically never gonna happen. Let mad science be mad.

SpadesNeil:
>this shouldn't get merged period, no "requires discussion" at all needed, it's a bad PR submitted in bad faith, we shouldn't humor it.

You say that like I have any intention to consider it.

Barring an amazing retort, I am just letting everyone shit on this bad idea.

Wickedtemp:
Eh. Rubbing it in any more would count as being a cunt. Just close the PR.

PR #4772, opened 2019-02-17
[AI BRANCH] Attempt to fix Active Edge

Novacat:
Probably should have a better fix long term, this is mostly to confirm something.

### DIFF
--- a/code/game/turfs/turf_changing.dm
+++ b/code/game/turfs/turf_changing.dm
@@ -28,7 +28,7 @@
 
 	if(N == /turf/space)
 		var/turf/below = GetBelow(src)
-		if(istype(below) && !istype(below,/turf/space))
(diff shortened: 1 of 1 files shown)


Novacat:
Okay. Travis passed.

I poked Polaris about why this change happened, and the only response I got was that it was to stop vacuum tiles from forming when people drilled out stuff.

An alternative solution to this issue is to have a subtype of turf/space at the midpoint instead of the current indestructible walls.

Heroman3003:
The problem with space at midpoint would be the fact that technically z-3 and z-4 are connected.

Novacat:
It would be a subtype of space, that would be some kind of blocking tile. It would still pass the check, I think.

Verkister:
The real solution polaris shouls have implemented in the first place would be actually utilizing the area/z-level defined base turf instead of hardcoding all holes to lead to space :v

PR #4792, opened 2019-02-22
Ports Shadekin (Minus AI)

Novacat:
Currently the shadekin will have no AI. Probably a good thing, since they are supposed to be player-controlled anyway.


SpadesNeil:
This will break the ones that spawn in the underdark, but I don't really want those to be there anyway.

Heroman3003:
Pretty sure Underdark spawn spots don't actually spawn any and are unused, as I never saw naturally occuring shadekin anywhere outside adminbus.

PR #4813, opened 2019-03-02
Fixes Mobvore for AI Refactor

Novacat:
**I CAN DO ANYTHING**


Haery70:
I support this fully and, hopefully, that it'll work.

Novacat:
So far, it is confirmed to work for hostile mobs, both pounce and non-pounce. Has not been tested with bumpnoms or retaliate.

SpadesNeil:
I would like to get those working soon as well, but right now, not a huge show stopper. We can do that in another PR.

PR #4823, opened 2019-03-03
Replaced Omni filters in the SM engine setup with regular ones

Heroman3003:
For too long poor newbies suffered from those cursed things being a cluster of bugged fuckery and dark magic. One wrong press of button and boom, filter is broken and will just make everything worse, and after being active once, can never be removed. This is dumb, all dumb and in place where one mistake can cost everything, we do not need measures that much worse.

Basically, puts normal disabled filters where omnis used to be. Wiki guide will need to be updated if its merged, but it will solve more issues long-term, mostly involving bugginess of those things screwing up setups/recoveries.


Poojawa:
Also just allow unwrenching of pumps even if pressurized as long as they're off, with obvious hinting and gas dumped consequences.

Heroman3003:
If somebody wants to masochist enough to fix omni filters, which were already 'fixed' how many times before? They are free to do so. I've seen attempts at fixing them, I've seen their code and I can see one thing: this solution saves everyone nerves and sanity.

Poojawa:
what exactly is broken about them 

What do you mean by 'being active once, can never be removed'

Heroman3003:
Unlike other 'machinery' in atmos, like pumps, valves, regular mixers or regular filters, even when turned off they somehow retain gas in them. Gas you cannot get out in any way. As a result you cannot unwrench them.

In addition to that if it any point you set any outputting/inputting direction to another omni filters, it for some reason breaks the filtering itself entirely.

Those two aspects mean simple thing. Even if your ending setup is correct, they may bug out because WHILE setting up you chose wrong direction then corrected yourself. And you cannot remove/place back to reset or even replace with normal without adming intervention because they are unwrenchable after they got some gas flowing.

maps/submaps/engine_submaps/engine_sme.dmm
```diff
@@ -99,12 +99,13 @@
...
 "bY" = (/obj/effect/floor_decal/industrial/warning,/turf/simulated/floor,/area/engineering/engine_room)
 "bZ" = (/obj/machinery/atmospherics/pipe/simple/visible/red{icon_state = "intact"; dir = 6},/obj/effect/floor_decal/industrial/warning,/turf/simulated/floor,/area/engineering/engine_room)
 "ca" = (/obj/machinery/atmospherics/pipe/manifold/visible/red{icon_state = "map"; dir = 4},/obj/effect/floor_decal/industrial/warning/corner{dir = 8},/turf/simulated/floor,/area/engineering/engine_room)
 "cb" = (/obj/machinery/atmospherics/pipe/simple/visible/yellow,/turf/simulated/floor,/area/engineering/engine_room)
 "cc" = (/obj/structure/cable/yellow{d2 = 2; icon_state = "0-2"},/obj/structure/cable/yellow,/obj/machinery/power/sensor{name = "Powernet Sensor - Engine Output"; name_tag = "Engine Output"},/obj/machinery/atmospherics/pipe/simple/hidden/supply{dir = 6},/obj/machinery/atmospherics/pipe/simple/hidden/scrubbers{dir = 5},/turf/simulated/floor,/area/engineering/engine_room)
+"cd" = (/obj/machinery/atmospherics/trinary/atmos_filter{dir = 1; use_power = 0},/turf/simulated/floor,/area/engineering/engine_room)
```
TheFurryFeline: Checked this thing out in a test server and it merely starts with a necessity to flip it before wrenching it in? Can't have the filter pre-flipped in this thing?

TheFurryFeline: Ah, nevermind. Overlooked the line with m_filter.

Heroman3003: One of the filters mapped in is mirrored, as indicated by m_filter subtype. Plain misinformed.

PR #4658, opened 2019-01-10
[MIRROR] Fixes lift and shuttle moves not updating lighting

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/5812
--------------------
The light sources in elevators and shuttles no longer leave their lighting permanently stuck at the initial location.

Fixes #4518

PR #4827, opened 2019-03-04
Makes elevator permanently powered

Heroman3003:
This includes lights inside and the doors.

There have been multiple discussions, but always one of the biggest problems with Tether brought up is how power outages guarantee problems doing even basic task due to elevator stopping working, and only alternative being a way-too-restricted, unreliable and still sorta-power-reliant backup shuttle.

Putting as separate PR because decision may be controversial.


SpadesNeil:
Do me a favor and make the elevator permanently fullbright as well.

The lighting is all fucked up. Just delete the lights and turn off dynamic lighting so it's always lit.

Heroman3003:
Dynamic lighting is already off INSIDE elevator itself. Lights in there are not affecting anything

SpadesNeil:
>Dynamic lighting is already off INSIDE elevator itself. Lights in there are not affecting anything

Why are the walls blacked out, then? You can't even see the button.

Also, marked this as requires discussion. The shuttle mechanics right now are bad. I am okay with this as an alternative until the shuttle is made less bad. The question is if everyone else is.

Novacat:
Not really that controversial. Elevator problems have been the main downside of the Tether for a long time.

I think there is a mirror PR that fixes the lighting issues, however.

Heroman3003:
Walls are dark because walls are not same area as inside of elevator. I will try disabling dynamic lighting on them too and see if this looks good on blackout.

Novacat:
Also, possible suggestion: Adding a relay to the midpoint. One of the other things that annoyed me about the Tether was that any radio transmissions to and from the midpoint would be lost completely.

TheFurryFeline:
@Novacat #4658 updates lighting, yeah.

Heroman3003:
Alright, lighting fix seems cool and working.

Very-Soft:
just to point out to make sure, this affects the doors, yes? The elevator itself normally works even when there is no power, the main issue tends to be that people crowbar open the doors at their destination, and then run away without closing them again, which makes it so the elevator can't move.

Wickedtemp:
Yuh, includes the doors!

Wickedtemp:
Lemme know if anybody comes up with a good reason to be opposed to this.

Wickedtemp:
20 days and no reason not to merge. I feel like we can just add it in and get the problems fixed.

PR #4829, opened 2019-03-05
Adds telecomm coverage to midpoint

Heroman3003:
Another potentially controversial thing, putting separately. Adds a fourth relay to surface telecomms, with coverage to the z-4/tether midpoint.

On one hand, it makes sense it would not be covered and it is occasioanlly used as a playground for event areas.
On the other it is really annoying to miss a single message or have to re-type yours duting those 2 seconds on midpoint.

Gonna see what admins think, really, was just suggested make a PR.

EDIT:To specify, which relay is Midpoint in case at any point admins want to make use of midpoint: https://imgur.com/yTH4C4M

### DIFF
--- a/maps/tether/tether-02-surface2.dmm
+++ b/maps/tether/tether-02-surface2.dmm
@@ -17653,6 +17653,15 @@
 /obj/machinery/door/firedoor/glass,
 /turf/simulated/floor/tiled,
 /area/tcommsat/chamber)
+"HR" = (
+/obj/machinery/telecomms/relay/preset/tether/midpoint,
+/turf/simulated/floor/bluegrid{
+	name = "Mainframe Base";
+	nitrogen = 100;
+	oxygen = 0;
+	temperature = 80
+	},
+/area/tcommsat/chamber)
 "HT" = (
 /obj/item/weapon/storage/box/glasses/pint,
 /obj/item/weapon/storage/box/glass_extras/straws,
@@ -32865,7 +32874,7 @@ Gg
 Gq
 GD
 GP
-Gz
+HR
 Hd
 Hn
 Gz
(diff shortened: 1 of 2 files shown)


Mewchild:
The tether midpoint also drops communicator calls.

TheDavestDave:
Would a good way round the admeme event thing to put it somewhere really easy to delete. Like on the other side of the unbreakable wall from the lift door

Screemonster:
Yeah, I think this was intended for people to be able to slam that motherfuckin' stop button and halt the elevator at the midpoint and eat the person in there with them while they're out of comms, but in practice this never happens because who the fuck ever stops the elevator? Stopping the elevator to have a scene would get you lynched.

TheDavestDave:
You get lynched for stopping the lift and pressing buttons so it goes floor 5 67 instead of floor 7 5 6

Novacat:
Mostly, the reason for this is that the elevator is, as of now, the only practical way between the Asteroid and Surface. For that reason, it sees a lot of traffic, and is not really practical for occupying for scenes.

Heroman3003:
Sadly, on practice eating part does not work, as elevator is very high-traffic, if scene is unwilling, everyone next to doors on levels above can see it happening, and just 'stopping' it won't stop people from repeatedly calling it, so unless its grab'n'gulp in the first place, "Eat someone in midpoint" plan just can't work out.

Wickedtemp:
All pros, no real cons. And it's irritating to be talking to somebody over comms and they're in the elevator and you don't know it, and by sheer coincidence, you say something to them just as they pass the midway point.

But they don't say anything because they don't realize they actually missed anything. From their perspective, you just dropped whatever conversation you were having with them, or perhaps, never started one because maybe your attempt to initiate a conversation was made while they were in the midway point.

And you don't say anything else because you had no idea they were in the elevator to start with, and you don't want to say the same thing again because if they already heard you, it might become annoying.

So then nothing happens and it's slightly awkward and annoying.

SpadesNeil:
The initial intention had never been used, so yeah, I'd say make telecomms cover the midpoint.

PR #4839, opened 2019-03-07
Adds ability to select scene traits without being custom species

Heroman3003:
Feature freeze just moments before I was about to post this, heck. Whatever, its ready and its awesome, IMO.

If you are not a custom species or xenochimera, you will now see four new toggles in custom species category. They allow obtaining of traits appropriate to names, Trash Eater, Brutal Predation, Bloodsucker and Succubus Drain.

If you are a custom species, you will have to select them same way you did before and nothing changes for you.

As a countereffect, all species except xenochimerae that had those traits inheirently (Predation for majority, Drain for Alraune and Prometheans, Trash eater for Vox and Prometheans, Bloodsuck for Alraune) are no longer inheirent. If you had them before, and were actually interested in using/actively using them, do not forget to enable them in VORE category of the character setup.

Those four traits do not add any actual effect unless you use them, so having them be an option for everyone who wants, and not just custom species, I think is good idea. Some people only select custom species at all just for those, which, for 'scene traits' isn't most fair. Also gives option to remove the verbs for people completely uninterested in those traits, but play species that does had them.

In addition to all mentioned, gives xenochimerae the one 'trait verb' they were missing: Trash Eater.

### DIFF
--- a/code/modules/client/preference_setup/vore/07_traits.dm
+++ b/code/modules/client/preference_setup/vore/07_traits.dm
@@ -15,6 +15,11 @@
 	var/starting_trait_points = STARTING_SPECIES_POINTS
 	var/max_traits = MAX_SPECIES_TRAITS
 
+	var/trait_trash_eater = 0
+	var/trait_brutal_predation = 0
+	var/trait_bloodsucker = 0
+	var/trait_succubus_drain = 0
+
 // Definition of the stuff for Ears
 /datum/category_item/player_setup_item/vore/traits
 	name = "Traits"
@@ -90,9 +95,20 @@
 		//Statistics for this would be nice
 		var/english_traits = english_list(new_CS.traits, and_text = ";", comma_text = ";")
 		log_game("TRAITS [pref.client_ckey]/([character]) with: [english_traits]") //Terrible 'fake' key_name()... but they aren't in the same entity yet
-		
+
 		//Any additional non-trait settings can be applied here
 		new_CS.blood_color = pref.blood_color
+	else
+		if(pref.trait_trash_eater)
+			character.verbs |= /mob/living/proc/eat_trash
+		if(pref.trait_brutal_predation)
+			character.verbs |= /mob/living/proc/shred_limb
+		if(pref.trait_bloodsucker)
+			character.verbs |= /mob/living/carbon/human/proc/bloodsuck
+		if(pref.trait_succubus_drain)
+			character.verbs |= /mob/living/carbon/human/proc/succubus_drain
+			character.verbs |= /mob/living/carbon/human/proc/succubus_drain_finalize
+			character.verbs |= /mob/living/carbon/human/proc/succubus_drain_lethal
 
 /datum/category_item/player_setup_item/vore/traits/content(var/mob/user)
 	. += "<b>Custom Species</b> "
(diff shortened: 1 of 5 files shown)


Novacat:
Can you add cosmetic traits, like eye glow, to this as well?

Heroman3003:
I thought about it, but decided to keep cosmetic options out, adding only traits that added a verb to use and had no other effect. If admins think that the idea in general is good and want to see traits like glowing body/eyes there too, I can add them, but till then just keeping it away from even cosmetic stuff.

Novacat:
I would say every neutral trait other than autohiss could probably be put on this list. I could use hotblooded for Ascian, for example, to fluff out his even less tolerance for warmth than other Tajarans. Glowing eyes would be nice as a fluff piece for NIF users, for example.

Overall, I support this PR as it would encourage greater usage of vanilla species.

Heroman3003:
Pretty sure warmblooded actually has less effect than natural tajaran weakness on live does, so it would actually make Ascian more tolerant to it? Either way, more of waiting on input from staff before adding/changing anything.

SpadesNeil:
This is getting a big fat "No" from me. It removes the point of having custom species to begin with, and I have no desire to merge this.

These traits being attached to certain races was a design choice, so that other players know what to expect of them. We could have made it so all races could pick traits, but we chose not to when traits were implemented.

code/modules/mob/living/carbon/human/species/station/station_special_vr.dm
```diff
@@ -30,6 +30,7 @@
 		/mob/living/carbon/human/proc/succubus_drain_lethal,
 		/mob/living/carbon/human/proc/bloodsuck,
 		/mob/living/proc/shred_limb,
+		/mob/living/proc/eat_trash,
```
SpadesNeil: This is the kind of thing I don't want to see, for example.

code/modules/mob/living/carbon/human/species/station/station_vr.dm
```diff
@@ -368,7 +367,7 @@
 	min_age = 18
 	icobase = 'icons/mob/human_races/r_vox_old.dmi'
 	deform = 'icons/mob/human_races/r_def_vox_old.dmi'
-	inherent_verbs = list(/mob/living/proc/shred_limb, /mob/living/proc/eat_trash)
+	//inherent_verbs = list(/mob/living/proc/shred_limb, /mob/living/proc/eat_trash)	//Removed trash eater and hardvore traits due to being put into prefs
```
SpadesNeil: That said, prometheans probably shouldn't have shred limb to begin with.

code/modules/client/preference_setup/vore/07_traits.dm
```diff
@@ -135,6 +161,11 @@
 			var/datum/trait/trait = negative_traits[T]
 			. += "<li>- <a href='?src=\ref[src];clicked_neg_trait=[T]'>[trait.name] ([trait.cost])</a></li>"
 		. += "</ul>"
+	else if (!(pref.species == SPECIES_XENOCHIMERA))
+		. += "<b>Enable Trash Can:</b> <a [pref.trait_trash_eater ? "class='linkOn'" : ""] href='?src=\ref[src];toggle_trait_trash_eater=0'><b>[pref.trait_trash_eater ? "Yes" : "No"]</b></a><br>"
```
SpadesNeil: This is the kind of thing I don't want to see, for example. I don't want everyone suddenly being able to eat garbage. There's no logical reason for this among most races.

code/modules/client/preference_setup/vore/07_traits.dm
```diff
@@ -135,6 +161,11 @@
...
+		. += "<b>Enable Brutal Predation:</b> <a [pref.trait_brutal_predation ? "class='linkOn'" : ""] href='?src=\ref[src];toggle_trait_brutal_predation=0'><b>[pref.trait_brutal_predation ? "Yes" : "No"]</b></a><br>"
+		. += "<b>Enable Bloodsucker:</b> <a [pref.trait_bloodsucker ? "class='linkOn'" : ""] href='?src=\ref[src];toggle_trait_bloodsucker=0'><b>[pref.trait_bloodsucker ? "Yes" : "No"]</b></a><br>"
```
SpadesNeil: Why should humans for instance suddenly be able to suck blood? Absolutely not.

SpadesNeil:
Closing as an unwanted feature.

code/modules/client/preference_setup/vore/07_traits.dm
Screemonster: xenochimera are obligate carnivores, if there was a reagent tag for it I'd make them unable to gain nutrition from regular nutriment, never mind from literal fucking garbage

code/modules/client/preference_setup/vore/07_traits.dm
Heroman3003: While I will accept the closure, I will say. This type of logic for some reason does not apply to what we have right now? Where people do get custom species... But just recreate species they want as closely as possible, give it same name as default species, and just add scene traits to only use in scenes since its only place they can practically be used in anyway? I just do not really see why playing default species should lock you out of specific scene setups (and primarily this PR was aimed at the Trash Can specifically, being one of the best traits for scene setups and teases, while being completely locked away for anyone playing any default species, other than prommies which are niche and vox). And on the flipside, so many species have the hardvore trait that they are never going to use anyway.

code/modules/mob/living/carbon/human/species/station/station_vr.dm
Heroman3003: As another note, this line is referring to Vox, so I think them having had shred limb makes sense.

Wickedtemp:
Tbh, "This doesn't make sense for most species." - neither does vore but okay I guess. Lorewise, there's nothing about Sergals or Tajaran that says they're able to swallow people whole, but we suspend belief for the sake of the whole...being an erp server. Pretty sure we can do the same for the neutral traits. At that point, sure, they could just do custom species, but I'd understand if they'd rather not, and it doesn't really effect a whole lot either way.

Heroman3003:
I will repeat it there too. I think object swallowing being hidden away behind just custom species is awful. Maybe change it so only trash eaters can gain nutrition from the digested non-food/people items, if people gorging themselves on trash is believed to be a problem. But the idea of being able to swallow someone's personal stuff and/or tease them before initiating by swallowing big food/objects being hidden away behind trait for just custom species is pretty limiting for somethiung that big of a trope in vore.

PR #4852, opened 2019-03-17
Fixes a Spriting Issue in Mobvore

Haery70:
As the title says, this pr's designed to fix the current spriting issues with mobvore ever since the AI Refactor update. After trial and error, I've discovered that several key elements in the code were missing compared to the old one. So, I integrated some of the old elements into the new one to success.

**Changes:**

- Adds some minor elements from the old mobvore code into the new one to help it update the sprites properly.

- Removes update_vore_icon proc and reintegrates the original update_icon child proc. It's more of a rename rather than a straight-up removal as all it does is add a few elements from the old proc into the new one and no longer making it a separate proc. It could work as it's own separate proc, but it's better safe than sorry.

- Several coding edits for the living, resting, and dead icon codes to allow it to register the mobs' icon states.

This should allow the mobs' sprites to update based on how full they are again.


Haery70:
There might be a few more commits than there should be considering that I was trying to update the fork to the master branch before I started.

PR #4858, opened 2019-03-19
Added ability to toggle and configure contamination and item digestion

### LINKED ISSUE #4856: [SUGGESTION] Option to disable contamination fully when using stripping bellies
There's plenty of choice as to how you want gear fouled up by your guts to be described - but no option at all to, y'know, NOT gunk up someone's stuff. So anyone who wants to do stuff with stripping but is turned off by the disgusting green sludge that their gut somehow generated JUST for these items is a bit stuffed. Let alone the people with bellies for say, carrying someone in a tail - those foul items up too. Plus, healing bellies apparently soak everything EXCEPT prey in poison? Having an option to disable contamination would solve most of these problems.

- People not to be mechanically prefbroken if they use item stripping mode
- Dousing people's clothing in poison to be optional.

#### What actually happened
- ...see above.

#### Steps to reproduce
- Be eaten into a non-digestive vore organ with stripping and itemweak enabled. Even one where it makes no sense for items to get dirty, like coiled up in a tail.
- GREEN SLUDGE OF EW

#### Code Revision
Current

Heroman3003:
- Removed Itemweak belly addon

- Added a new belly option: Item Mode. Determines whether items in belly will be held or digested. Is not at all related to Belly Mode, which determines what happens to mobs. The two are separate completely now.
Modes are: Hold, Digest and Digest (Food Only). Last one will only digest food items and organs.

- Added a new belly option: Contaminates. A toggle, determines whether or not belly will apply contamination (where applicable) to its contents. IDs get contaminated regardless of this specific setting.

- Contamination flavor is now available only when Contaminates is On.

- Added a new belly option: Contamination Color. Allows to choose color of the contamination overlay. Rainbow guts represent.

- Contamination made less recursive. Now only items that are meant to act as storages will have their contents contaminated. Applies to coats/labcoats that have internal storage, backpacks, belts and briefcases. Some items that got contaminated with full recursive really shouldn't have in the first place.

- Fixed contaminated boxes not falling apart properly when used before cleaning off contamination.

- Default settings for new options are: Item Mode - Digest (Food Only), Contaminates - Yes, Contamination Flavor - Generic, Contamination Color - green (same as current).

EDIT1: Fixes #4856 
EDIT2: https://prnt.sc/mzt7ec - screenshot of changed Vore Panel


Wickedtemp:
<3

Very-Soft:
I feel like the item settings sound confusing.

It sounds like you can have regular digest, or only food digest. Does food digest count people as food?

And passed that, these options do not seem to have an option to digest items without digesting the person, which while it definitely should not be the default option, is a thing that some people are probably going to want to do. Removing the ability to do so seems less than ideal. (Especially since some folk do the trash eater thing and gurgle up all the garbo)

Like, I'm pretty fine with the bellies current digestion settings. I just think bellies should default to itemweak when created. That way digesting items is something you have to enable on purpose.

Heroman3003:
It does have an option to digest items without digesting person. You just set Belly Mode on hold and Item Mode on digest. Belly mode determines what happens to mobs. Item mode determines what happens to items.  And digest food only digests food items (ripped organs and actual food), while digest item mode digests ALL items (including food)

Very-Soft:
Ah! Well that sounds better! I didn't quite understand. Could we get a screenshot of the layout?

Heroman3003:
https://prnt.sc/mzt7ec
Here is how Vore Panel looks like. Contamination Flavor/Color only show up if Contaminates is on Yes.

Wickedtemp:
This... might actually cause me to use contamination stuff sometimes if I can change the color to something less related to vomit. Purple is a much better color.

Verkister:
Very good :+1:

PR #4868, opened 2019-03-22
Adds liquidprotein rations

Novacat:
For obligate carnivores. Suggestions welcome on where else to toss these things.

- Adds liquidprotein rations, they can be found in exploration, sweatmax vendors, and shelter capsules.
- Ups the price for protein shakes to discourage hoarding.
- Adds supply packs for emergency rations and emergency meat rations, they cost 20 and 30 points, respectively, and contain 4 rations each of their respective types


Wickedtemp:
Tbh, increasing the cost isn't going to do anything. If people are actually hoarding vendor food, they're probably not going to care about IC money. That's why they're already spending it all on vendor food.

Novacat:
The idea behind increasing the cost is that if someone tries to hoard them on a ludicrous scale, say, buying every single one on the whole Tether, the higher price may cause them to run out of cash.

Of course, this only works for assistant/off duty whom tend to not have that much money. A head role probably is not going to care.

Wickedtemp:
I've literally never seen anyone buy every single one on the station. If someone's actually hoarding *that* much, that doesn't warrant a price increase (which won't even prevent this from happening), it warrants an ahelp.

Very-Soft:
I'm all for more cool options but it's funny to me that people prefer the protein shakes to the bars, cuz the protein bars are real good.

SpadesNeil:
Yeah I don't think I like the choice of what we have as the alternative.

Novacat:
The purpose of these things is for carnivorous species, Unathi and (soon to be) Xenochimera. While it probably would not be strictly animal protein, it would be close enough to be metabolized the same way. My idea for these things is that it would be like ground up vat-grown meat grown in industrial quantities and loaded with a metric ton of preservatives.

Wickedtemp:
Soy protein when. Tempest needs to be a soiboi cuck

PR #4879, opened 2019-03-29
Adds Mlems to Default Species

Novacat:
Antiseptic Mlem verb applied to Tajara, Vulpkanin, and Zorren.

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/station_vr.dm
+++ b/code/modules/mob/living/carbon/human/species/station/station_vr.dm
@@ -32,7 +32,7 @@
 
 	spawn_flags = SPECIES_CAN_JOIN
 	appearance_flags = HAS_HAIR_COLOR | HAS_LIPS | HAS_UNDERWEAR | HAS_SKIN_COLOR | HAS_EYE_COLOR
-	inherent_verbs = list(/mob/living/proc/shred_limb)
+	inherent_verbs = list(/mob/living/proc/shred_limb, /mob/living/carbon/human/proc/lick_wounds)
 
 	flesh_color = "#AFA59E"
 	base_color = "#777777"
@@ -159,6 +159,7 @@
 	num_alternate_languages = 3
 	secondary_langs = list(LANGUAGE_TERMINUS)
 	name_language = LANGUAGE_TERMINUS
+	inherent_verbs = list(/mob/living/carbon/human/proc/lick_wounds)
 
 	min_age = 18
 	max_age = 80
@@ -198,6 +199,7 @@
 	num_alternate_languages = 3
 	secondary_langs = list(LANGUAGE_TERMINUS)
 	name_language = LANGUAGE_TERMINUS
+	inherent_verbs = list(/mob/living/carbon/human/proc/lick_wounds)
 
 	min_age = 18
 	max_age = 80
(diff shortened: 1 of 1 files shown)


Wickedtemp:
I see no issue with this other than the fact that Sergals and Unathi can also mlem but are not included.

...But, yeah. On an actual note, I don't really care if this is exclusive to custom species or not, I can see the argument for either side.

Heroman3003:
Not sure abount unathi because lizards don't quite lick their wounds as far as I am aware, but sergals and zorren definitely should also have it.

KasparoVy:
:cheese::ok_hand:

SpadesNeil:
>I see no issue with this other than the fact that Sergals and Unathi can also mlem but are not included.

Sergals have the opposite of healing mlems. Unathi as well.

Imagine getting licked by a komodo dragon.

You will literally die.

Be grateful we don't make their mlems realistic. They would kill you.

code/modules/mob/living/carbon/human/species/station/station_vr.dm
```diff
@@ -32,7 +32,7 @@
 
 	spawn_flags = SPECIES_CAN_JOIN
 	appearance_flags = HAS_HAIR_COLOR | HAS_LIPS | HAS_UNDERWEAR | HAS_SKIN_COLOR | HAS_EYE_COLOR
-	inherent_verbs = list(/mob/living/proc/shred_limb)
+	inherent_verbs = list(/mob/living/proc/shred_limb, /mob/living/carbon/human/proc/lick_wounds)
```
SpadesNeil: I am not in favor of sergals having healing mlems as I was under the impression their maws had a fuckload of bad bacteria in their original lore, which I would like to see in SS13 lore officially but am not forcing people to have gross maws because that could be prefbreaking.

That said, sergals also having mlems somewhat pushes it for me. Sergals should not have mlems also for the reason of _every furry species having mlems_ makes mlems not as special.

I want it to be a special thing to some of our default species.

SpadesNeil: Would you give Xenochimera mlems? No you would not. :U 

In fact, the vulp having it is kinda iffy for me too but we'll put it to discussion.

Screemonster: xenochimera would be fuckin' anti-mlems. Of all the fancy vorey verbs they shouldn't have for lore reasons, licking people to cure infections is probably top of the list.
~Now, licking people's wounds to _give_ them infections on the other hand...~

Heroman3003:
While sergals are not based on real species, I say zorren, vulp and tajaran definitely should have it as both canines (foxes and folfs that zorren and vulp are based on) and cats (that tajaran are based on) are all species having a bit of infamy for wound-licking practices. Sergals are however portrayed as different and more aggrissive, so I understand reasoning behind them not getting it. Plus, they aren't even mammals either, so I can see them not having licks. As for which species should have them, Zorren and Vulp really lack in mechanical individuality, so having something unique to mammalian animal-based species seems like a good idea to me.

Sebbe9123:
At least in Norway,  vulpines are known to have diseases out the ass in the mouth and being bit by one is almost a 100% chance of getting rabies and a nice good old infection if you can't get it looked at fast.

code/modules/mob/living/carbon/human/species/station/station_vr.dm
Wickedtemp: While I see we have a disagreement on the base philosophy of mlems - I believe that they should be shared and spread across all those mlem-capable - which contrasts to your value in the unique nature of mlems, I understand your position in the matter and would be happy regardless of whats changed or left the same in this PR.

Novacat:
That is because being bit by one usually means that you are dealing with a sick animal anyway, as normally foxes would rather run away from/avoid humans. Either way, going to remove Sergal mlems than merge this.

PR #4892, opened 2019-03-31
Fixes Virgo 3B Step Trigger

Novacat:
Now Ascian and Chance can have their romantic parajump into Virgo 3B.

PS: Also adds new bluespace sprites.


Very-Soft:
you mean you -don't- want to tumble around in the sky helplessly forever??? 

What a weirdo!

PR #4902, opened 2019-04-03
Introduces a variety of QoL changes to research.

LiquidFirefly:
- Lowers print times by about 1/3rd across the board

- Nearly triples protolathe material capacity across the board. 

- Adds an Advanced Part Exchange device at materials: 5, engineering: 5, which can hold 200 parts. 

This thread is mostly to spark discussion about the topic.  Most of what this does is reduces the time required to finish RnD by a few minutes and greatly increases QoL as it relates to materials lying around the lab, RPED routes, and most importantly printing parts for RPED routes.


SpadesNeil:
Not really sure this is necessary, but if it is, we should see what Polaris thinks of it. I'm rather on the fence about it and am not interested in merging it here.

LiquidFirefly:
https://github.com/PolarisSS13/Polaris/pull/6042

PR #4907, opened 2019-04-05
Makes Virology Usable

Mewchild:
Adds a 4 virus samples into virology at round starts
Changes the Virus Sample supplypacks to need Medical Doctor Access, instead of CMO


Jakksergal:
This is what virology's needed since absolutely forever. If Virologists ever wanted to get ahead of the game they'd have to order a crate or *six* from cargo.

Issue #789, opened 2017-01-01
Nurture/Nourish belly mode


Dinnel:
#### Brief description of the issue
I was wondering if there can be a nurture mode. I may of mentioned it under suggestions but this is the first time I'm asking as an issue itself. If I had to sum it up, its basically the opposite of draining and while not outright similar to healing mode, it can be used for something like it.

#### How do I think Nurture mode should work?
Note that some of this is probably not a good idea, just throwing stuff out there to make the mode slightly flexible.

The first and the biggest feature I'm hoping gets put in is to share nutriment. This can be either adjusted so the 'pred' gives the 'prey' nutrients if they're lower than the pred's nutriment level or alternatively, if either one of them is out of balance, the mode will try to equalize.

The second feature would be sharing chemicals, either splitting it between the pred and prey (So for example you could take 30 units of water and it would be split evenly between the pred and prey) or make sure the prey has roughly the same amount of chems the pred has (So if the pred has 30 units of water, so would all the prey inside the pred)

The third, I would suggest sharing toxin damage, perhaps strictly from prey to pred. It would work like if the prey has 20 tox damage. and the pred has 10 tox damage, the pred would slowly draw tox damage from the prey until he matches, so the pred would have 15 tox damage after the process is done while the prey would also have 15 tox damage. If the pred had 0 to start with, they'd have 10 and the prey would have 10 after it's all done and over with.

Finally, as long as the pred is able to breathe, prey will be immune to oxyloss in the case of lungs being ruptured, bloodloss or people in crit (Though I think thats already in, not sure)

#### Summary
In short, a mode that treats the 'prey' like a child and the 'pred' as the mother, sharing the same thing the pred ingests as well as nutriment. It doesn't heal physical injuries directly like heal mode, but 'heals' toxin by sharing it with the pred and prevents oxyloss. It can also run a bit differently if someone else has a good idea on how it should work.

SpadesNeil:
Kind of a niche idea, but I can see why there'd be an appeal in it. Not sure if it's worth adding but let's talk about it.

Dinnel:
Someone suggested in development to rename it to **nourish** instead of **nurture**. It does sound kind of better.

Screemonster:
This'd be a pretty useful thing to have after absorption so the people don't _have_ to be hungry to the point of starvation when you let them out.

Dinnel:
More discussion about in general chat. Another person thought of a solution to the chemical transfer bit.

[...]

PR #4909, opened 2019-04-05
Adds Nourish belly mode

Heroman3003:
Adds new belly mode, Nourish, that basically transfers nutrition from pred to prey.

It scales, so hungry prey gets it faster, moderately hungry with more normal rate, and when they are full you can keep stuffing them as long as you yourself have nutrition, but you will start losing more than what they get.

In return, Heal mode no longer affects prey's nutrition.

### DIFF
--- a/code/__defines/belly_modes_vr.dm
+++ b/code/__defines/belly_modes_vr.dm
@@ -8,6 +8,7 @@
 #define DM_GROW									"Grow"
 #define DM_SIZE_STEAL							"Size Steal"
 #define DM_HEAL									"Heal"
+#define DM_NOURISH								"Nourish"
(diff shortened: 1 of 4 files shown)


SpadesNeil:
https://github.com/VOREStation/VOREStation/issues/789 Was previously closed and marked as "won't change"

SpadesNeil:
The reason for this was a concern that it would promote pregnancy RP and age reverting RP and a bunch of other age related stuff we don't want to happen.

Heroman3003:
I was told it was marked that because of being a suggestion nobody was willing to actually develop that in #development channel which is why I just took up to that. But I guess I got misinformed

PR #4910, opened 2019-04-05
Un-kevinzes nanoui and chemistry subsystems back into processes for greater good

Heroman3003:
Direct port of https://github.com/PolarisSS13/Polaris/pull/6030


Novacat:
Note that Vorestation edits are not necessary for this, since this is intentionally going to get reverted/overwritten as soon as Polaris fixes the problem on their end. This is basically a temporary fix so we are not waiting for Polaris to fix things before having a release.

Wickedtemp:
yes please. this also screws up some grenades, on top of all the other stuff. Chemistry, bartending, cooking, etc.

PR #4924, opened 2019-04-06
[MIRROR] Replaces AI-controlled Maint Drones with AI-controlled Borg Shells

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/6025
--------------------
Presently, AIs can produce and hop into empty maint/construction drones. This, to me, has several balance issues, listed below.
- Drones cost nothing to build, making it essentially 'free' to the AI.
- Multiple drones can easily be stockpiled in out of the way places, with the only cost being the time to do so.
- Killing a drone does not mean much since the AI can make more for free.
- Stopping the ability to build more drones is a nontrivial task since most things you can do to the drone fab room, the AI can undo by using its AI wireless powers, or using a stockpiled drone to fix the room.
- The modules for maint and construction drones are rather bloated, and possibly feature-creeped heavily. Initially they were made with the assumption that the users for them would keep to themselves as mandated by their lawset (the construction drone's lawset says otherwise but thats another topic), would not be able to talk to any other non-drones, and wouldn't really be part of the crew. The AI being able to hop inside them blows that assumption out of the water entirely.

To fix these issues, I've ripped out the ability for AIs to go into drones, and have instead ported lots of code from TG, and added some of my own code on top, to make it so AIs can hop in and out of specially designed 'AI Shells', which mostly act like a regular cyborg body, except being made from a new MMI type. This new borg type is able to use modules, be upgraded, get reset, get repaired, etc. I feel this is a more interesting and balanced way of having the AI be able to have a physical presence, for the reasons below.
- AI Shells have a real material cost now, being roughly the same price as a new borg.
- Building multiple AI Shells requires more materials as stated above, and the process is much more involved than just 'click drone fab every few minutes'.
- Killing an AI Shell is a significant blow to the AI, since the AI cannot make a new one for free. Instead, they/robotics would need to sacrifice more materials and time for a new one, or they/robotics will need to recover the dead shell and repair it.
- AI Shell construction is easier to halt since the AI depends on a working AI Shell with a research module loaded, or a roboticist, to build more AI Shells, instead of only needing to click a machine every few minutes.
- AI Shell modules are the same as borg modules, and so balancing them is easier since both are more similar to each other instead of using drone modules. I also feel that having to pick a module for each shell is more interesting and fun for the AI player, and gives more options for AIs to do things outside of the capabilities that maint drones provided (e.g. medical).

Other notes:
All of this can be turned off in the config, for downstreams who don't want this.
There is a config option for the AI to start with a free premade AI Shell. This has several balance implications that downstreams should consider when deciding if they want to turn it on or off.
The above option requires a specific landmark to be mapped in to work. Note that mapping more than one of those will make all of them make a free AI Shell.
AI players with custom borg sprites can use those sprites for their AI Shells if the AI is the same name as the borg that the sprites are attached to.
This PR also ports a 'soul link' datum from TG which might be useful to us in the future. It's been touched up to meet the style guidelines for this codebase.


Heroman3003:
I personally think enabling the option in config is a great idea and do not see why not at all.

PR #4926, opened 2019-04-06
Adds colorable moth wings

saysage:
Adds gray-scaled and colorable version of the moth wings, for people who might prefer to play as any moth other than just yellow!


Novacat:
Not really a fan of the filediff crapping out, but, if there are problems I can revert this.

PR #4928, opened 2019-04-07
Removes ability for non-player mice to interact with doors

Heroman3003:
What it says on the tin. If mouse is not player-controlled, bumping door won't even deny with red flash and bleep, but absolutely nothing will happen. Player-controlled mice can still open all-access doors and bump with the bleeps into restricted ones.


SilverTalismen:
Thank you.

PR #4937, opened 2019-04-07
Fixes micromechs using normal equipment

### LINKED ISSUE #4871: Gophers mounted orebox refuses to take in new ore when full once
A gopher (the micro-people ripley) has to use a mounted orebox instead of normal means. This works perfectly well (even if the cargo space for said orebox is way too small for my taste) exept when the box was full once. After that, even when emptied, the box will not take in any newly mined ore automaticly and you're forced to raise the box between mining and manualy sweep the are to grab all the ore.

Once the orebox was emptied, new ore that gets mined gets automaticly deposited into the mounted orebox, similar to how ripleys do it.

#### What actually happened
Mined ore was deposited on the ground where it was mined, just like how mining without an orebox does it.

#### Steps to reproduce
- (Step 1) Aquire a gopher and a micro, equip it with mining equipment and head to the mining areas.
- (Step 2) Mine until the gophers mounted orebox is full, then empty it
- (Step 3) Mine again and notice all the phoron/iron/whatever gets deposited on the ground.
- (Step 4 [Optional]) If no problems where found with the first 3 steps, switch out the drill for a diamond drill and repeat.

#### Code Revision
6c758051139daed75163f1941b40600bd97dceaa 
 
- May or may not be caused because of diamond drill and mounted orebox interactions. I can't confirm if this was because of not using the ''right'' drill or if this is a problem with the mounted orebox itself.

Heroman3003:
Oh god, the vorestation edits, they are everywhere

Anyway, micromechs no longer have universal, weapon or utility slots, and only 1 hull and 1 special slot.

Instead new slot types are added: Micro Utility and Micro Weapon. Only micromechs have these slots. Gopher has 2 micro utility slots, Weasel has 2 micro weapon slots and Polecat has 3 micro weapon slots.

Fixes #4871

### DIFF
--- a/code/game/mecha/equipment/mecha_equipment.dm
+++ b/code/game/mecha/equipment/mecha_equipment.dm
@@ -4,6 +4,10 @@
 #define EQUIP_WEAPON 2
 #define EQUIP_UTILITY 3
 #define EQUIP_SPECIAL 4
+//VOREStation Addition begin: MICROMECHS
+#define EQUIP_MICRO_UTILITY 5
+#define EQUIP_MICRO_WEAPON 6
+//VOREStation Addition end: MICROMECHS
 
 /obj/item/mecha_parts/mecha_equipment
 	name = "mecha equipment"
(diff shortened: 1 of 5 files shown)


TheDavestDave:
Does this mean we can't add fuel or wireless chareging to the micromechs?

Novacat:
Probably not, but this was way overdue anyway.

PR #4940, opened 2019-04-07
Makes Sonic Jackhammer a 'safe' mining tool

### LINKED ISSUE #4769: Mining RNG radiation sucks a big one
Xenoarch code causes RNG radiation bursts if you happen to dig up a tile that has an artefact on it. The code seems _intended_ to be a punishment for xenoarchaeologists that just use a giant pick, but the code (which dates back to really old no-fun-allowed baystation xenoarch) is applied to ALL pickaxes and treats the mining tools as "a really big hand-pick".
Mecha mining drills, the new mining tools, emitter mining (lol), and rig-mounted drills and plasma cutters are not affected by this issue as they just call getdrilled() like a sensible object - currently explorers are the ones that get punished by this the most.

The attackby proc for mineral turfs already has a comment in it about it being a horrible pile of spaghetti so probably worth an overhaul. A possible solution would be to give certain tools (such as the sonic jackhammer, which currently lacks a niche) the ability to just getdrilled() without uncovering artefacts.

Undecided on whether to tackle this one here or upstream. Worth asking if they're interested.

Heroman3003:
Both regular and borg sonic jackhammer will now totally annihilate any archeological finds, with nigh none chance of recovery and absolute zero chance of radiation wave. A good option for those having particularly bad luck with radiation pockets.

Fixes #4769

### DIFF
--- a/code/modules/mining/mine_items.dm
+++ b/code/modules/mining/mine_items.dm
@@ -55,7 +55,6 @@
 	origin_tech = list(TECH_MATERIAL = 3, TECH_POWER = 2, TECH_ENGINEERING = 2)
 	desc = "Cracks rocks with sonic blasts, perfect for killing cave lizards."
 	drill_verb = "hammering"
-	destroy_artefacts = TRUE
 
 /obj/item/weapon/pickaxe/gold
 	name = "golden pickaxe"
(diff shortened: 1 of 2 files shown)


Screemonster:
This is pretty much exactly what I was gonna put on the polaris version of the despaghettifying PR but they didn't want it. 

Mark your vorestation edits though :v

Heroman3003:
Whoops,  forgot it was not a vorestation file... I'll just move it to vorestation file instead.

PR #4950, opened 2019-04-08
Fixes borg not being unlockable via PDA holding an ID

### LINKED ISSUE #4513: PDA with Robotic's ID not unlocking cybrogs
Game gives an "Access Denied" message when attempting to unlock a cyborg's hatch using a PDA that has a Robotocist ID inside it.

Expected it to work the same as when using the ID normally.

#### What actually happened
Borg was not unlocked, instead giving an "Access Denied" alert, as if I was using an incorrect ID rank.

#### Steps to reproduce
- Place Robotics ID inside PDA
- Click a locked borg with the PDA in hand
- Receive "Access Denied" message
- Just to confirm, remove ID from the PDA, and click the still locked borg with the ID in hand
- Unlock the borg as normal.

#### Code Revision
- Server revision:  - 2018-11-21
53e1ff354a04286f7155c2184de034f038102bd6 
 
- Tested using a Robotocist ID on a MediHound. I assume the issue extends to any PDA with robotics access ID and any borg. Not tested anything other than what happened in the particular situation.

Heroman3003:
Fixes #4513 

Direct port of Polaris 6057

### DIFF
--- a/code/modules/mob/living/silicon/robot/robot.dm
+++ b/code/modules/mob/living/silicon/robot/robot.dm
@@ -722,17 +722,18 @@
 			return 1
 	return 0
 
-/mob/living/silicon/robot/proc/check_access(obj/item/weapon/card/id/I)
+/mob/living/silicon/robot/proc/check_access(obj/item/I)
 	if(!istype(req_access, /list)) //something's very wrong
 		return 1
 
 	var/list/L = req_access
 	if(!L.len) //no requirements
 		return 1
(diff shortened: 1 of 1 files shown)


SpadesNeil:
If Polaris is doing it why not wait for the mirror PR? Also, link the actual link to the Polaris thing.

Novacat:
I was hoping I could update within the next day or two, get all these hotfixes in which would sort out most of the bugs being reported.

PR #4955, opened 2019-04-08
Fixes tram ignoring respawn timers

Heroman3003:
Cue in angry lynchmob

Fixes #3887

PR #4957, opened 2019-04-08
Lowers respawn timer back to 1 minute

Heroman3003:
Because general consensus on respawn timer is that 5 minute is long enough to get bored, but not long enough to discourage PGing. Plus PGing using the round-exit is already rare and when it is used, its either endorsed because emergencies or punished by admins.

Posted mostly due to people getting mad at me fixing unintended behavior of tram bump-leave (and just bump version of leave) skipping setting of the deathtime in #4955

### DIFF
--- a/code/modules/mob/mob.dm
+++ b/code/modules/mob/mob.dm
@@ -373,8 +373,8 @@
 		var/deathtimeseconds = round((deathtime - deathtimeminutes * 600) / 10,1)
 		to_chat(usr, "You have been dead for[pluralcheck] [deathtimeseconds] seconds.")
 
-		if ((deathtime < (5 * 600)) && (ticker && ticker.current_state > GAME_STATE_PREGAME))
-			to_chat(usr, "You must wait 5 minutes to respawn!")
+		if ((deathtime < (1 * 600)) && (ticker && ticker.current_state > GAME_STATE_PREGAME))	//VOREStation Edit: lower respawn timer
+			to_chat(usr, "You must wait 1 minute to respawn!")
 			return
 		else
 			to_chat(usr, "You can respawn now, enjoy your new life!")
@@ -716,7 +716,7 @@
 
 			if(statpanel("Tickets"))
 				GLOB.ahelp_tickets.stat_entry()
-				
+
 
 			if(length(GLOB.sdql2_queries))
 				if(statpanel("SDQL2"))


Very-Soft:
Mmh I don't like this one. If someone really needs a different job or character in, I'd really prefer they go to try to fax for it, and with how the tram has been, people often already really prefer to just run to it to respawn quickly to do different jobs.

If I'm honest we've been pretty lax on that lately, so it kind of happens pretty regularly, and that should probably be worked on, but I don't think making all respawn timers be 1 minute is really a -good- thing. 

5 minutes isn't really all that long a wait, but is still long enough that in a critical situation, other options are preferable.

Screemonster:
I'd like it if it could be made short _if and only if_ the mob had been in the world for less than a few minutes, or had never spawned as a mob (ie. you clicked observe).
As in, if you just observed? Fine, short timer. You join, immediately realise you spawned as the wrong character for your job? Back on the tram, short timer. You actually played your character for a while before tramming out? You gotta wait.

I'd be all over letting people return to the lobby near immediately but not _rejoin_ until a timer had passed too. Long timer for rejoining as the same character, short timer for rejoining as different one, 'cause really the main abuse of revolving-door-tram-syndrome is people leaving and rejoining to bypass the HoP.

Heroman3003:
Different timers for different characters would require something entirely else as Respawn verb only moves the ghost into lobby, and thats it.

Verkister:
What if we made a cooldown for respawning as the _same_ character with Scree's exceptions? With that thing possibly having even longer time, it would probably discourage using the tram as express HoP while allowing people to quickfix their jointime oopsies and switching characters without having to waste their full wait time for the sake of waiting.

Verkister:
That sounds like the best solution.
I can tell from experience that 5 minutes is just enough to see your partner you had plans with either forget and leave as well or get pulled into other scenes, just enough to lose _the mood_, or just enough to forget what you were planning to rejoin for in the first place. Also if the HoP/etc is busy or missing, the time it will take to reach/fax for them to return/arrive is rarely going to be less than 5 minutes anyway for those who prioritize _"optimal performance"_ over roleplay standards.

Wickedtemp:
I'll be honest. I don't think I've personally seen a legitimate example of someone exiting, respawning, and joining as a higher rank except maybe once. I'd be okay with putting gameplay first and just lowering the respawn timer. I'd also be okay with keeping it the same, because it doesn't feel like a long time. If you have a planned scene with someone and you tell them you'll be back in five minutes to go offduty or something... And then in those five minutes they decide to ignore the planned thing and scene with someone else, that's a problem with your partner, not the timer.

Vorrarkul:
Honestly, the respawn timer is in the wrong place- it should prevent them from clicking the Join Game button once in the spawn menu, NOT force them to continue observing the round for several minutes. I feel that it encourages metagaming by making them stick around watching and listening to everyone. But that's a topic for a different PR.

I support a faster respawn timer because it lets me get back to what I want faster- that is, the character edit menu. I spend an unhealthy percentage of my time logged in, just tweaking character loadouts and colors.

Very-Soft:
> Honestly, the respawn timer is in the wrong place- it should prevent them from clicking the Join Game button once in the spawn menu, NOT force them to continue observing the round for several minutes. I feel that it encourages metagaming by making them stick around watching and listening to everyone. But that's a topic for a different PR.
> 
> I support a faster respawn timer because it lets me get back to what I want faster- that is, the character edit menu. I spend an unhealthy percentage of my time logged in, just tweaking character loadouts and colors.

I agree with you. But also, you can access Character Setup from anywhere at any time. It's in the preferences tab.

Very-Soft:
That is to say, I agree that it would be better if the respawn timer started once you're in the lobby. I still think that a 5 minute respawn timer is preferable for discouraging people just hopping on the tram instead of actually roleplaying and such in tense situations.

SpadesNeil:
If it becomes a problem I'll bump it back up.

PR #4979, opened 2019-04-09
Allows ghosts with backups to ping medical again

### LINKED ISSUE #4945: Suggestion: Let people who need a resleeve ping medical
Sometimes medical gets busy and misses the message, or doesn't know if a person actually wants to be resleeved or is even still around or all that.

It'd be cool if you're eligible to be resleeved, if you clicked on the resleeving machine it would give medical a '[X] is passed due for a mind backup.'

Just like, put a 5 minute cooldown on this so that people can't spam it like crazy. This way you can politely remind medical that you're waiting, without needing to hunt all around and looc at people to help you.

Heroman3003:
You cannot ping if your backup notification has not naturally rung yet or it has been less than 10 minutes since last notification (including initial). Also can't ping if you don't have one obviously.

Verb is called Notify Transcore.

Resolves #4945

### DIFF
--- a/code/controllers/subsystems/transcore_vr.dm
+++ b/code/controllers/subsystems/transcore_vr.dm
@@ -56,7 +56,7 @@ SUBSYSTEM_DEF(transcore)
 			implants -= imp
 			continue
 
-		//In a human	
+		//In a human
 		BITSET(H.hud_updateflag, BACKUP_HUD)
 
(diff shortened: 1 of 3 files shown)


Wickedtemp:
They can also just use LOOC. Neat either way though.

Novacat:
IC mechanic that can replace the old OOC mechanic of LOOCing at people? Yes please.

PR #4986, opened 2019-04-09
Rebalances zoom pills

### LINKED ISSUE #4660: Med rebalance broke zoom pills.
Med  rebalance broke zoom pills.

To get speed, then brain damage.

#### What actually happened
You actually just get dunked on, immediately slowig down after taking the pill. It OD's you on hyperzine. 

#### Steps to reproduce
Take zoom pill
Suffer
Die

#### Code Revision
 
Heroman3003:
Replaces reagents because in current state its not too good at punishing floor pills or at giving zoomies.

Now instead has stimm (short term speed boost), and some toxic reagents that will start causing vomiting by the  time stimm wears off (or if you're specifically unlucky, before it wears off). One pill is not enough to be deadly, but will give noticable toxins. Two or more will start being hurtful for organs.

Resolves #4660

### DIFF
--- a/code/modules/reagents/reagent_containers/pill.dm
+++ b/code/modules/reagents/reagent_containers/pill.dm
@@ -292,9 +292,12 @@
 
 /obj/item/weapon/reagent_containers/pill/zoom/Initialize()
 	. = ..()
-	reagents.add_reagent("impedrezene", 10)
-	reagents.add_reagent("synaptizine", 5)
-	reagents.add_reagent("hyperzine", 5)
+	if(prob(50))						//VOREStation edit begin: Zoom pill adjustments
+		reagents.add_reagent("expired_medicine", 5)
+	else
+		reagents.add_reagent("mold", 5)
+	reagents.add_reagent("fertilizer", 5)
+	reagents.add_reagent("stimm", 5)	//VOREStation edit end: Zoom pill adjustments
 
 /obj/item/weapon/reagent_containers/pill/diet
 	name = "diet pill"
--- a/code/modules/reagents/reagent_containers/pill.dm
+++ b/code/modules/reagents/reagent_containers/pill.dm
@@ -293,9 +293,8 @@
 /obj/item/weapon/reagent_containers/pill/zoom/Initialize()
 	. = ..()
 	if(prob(50))						//VOREStation edit begin: Zoom pill adjustments
-		reagents.add_reagent("expired_medicine", 5)
-	else
-		reagents.add_reagent("mold", 5)
+		reagents.add_reagent("mold", 2)	//Chance to be more dangerous
+	reagents.add_reagent("expired_medicine", 5)
 	reagents.add_reagent("fertilizer", 5)
 	reagents.add_reagent("stimm", 5)	//VOREStation edit end: Zoom pill adjustments
 
--- a/code/modules/reagents/reagent_containers/pill.dm
+++ b/code/modules/reagents/reagent_containers/pill.dm
@@ -295,7 +295,6 @@
 	if(prob(50))						//VOREStation edit begin: Zoom pill adjustments
 		reagents.add_reagent("mold", 2)	//Chance to be more dangerous
 	reagents.add_reagent("expired_medicine", 5)
-	reagents.add_reagent("fertilizer", 5)
 	reagents.add_reagent("stimm", 5)	//VOREStation edit end: Zoom pill adjustments
 
 /obj/item/weapon/reagent_containers/pill/diet


Screemonster:
are you kidding me, impedrezine causes brainloss, it's absolutely punishing over time

make a knockoff zoom pill with the same name and your reagent list instead, that way we can put the "shit some asshole cooked up in their garage" into the trash list so maintdivers won't know whether they've got the good stuff or something terrible

honestly though since hyperzine's speedboost was nerfed I've not seen people running around spamming sanic memes so I'm not sure what this is trying to address

Heroman3003:
Problem with Zoom as they are is not nerf of hyperzine's speedboost. But the fact that it plain doesn't work when combined with Impedrezine's drowsyness effect. The aim is to recreate the "quick zoomies" feel, which got completely trashed in actual zoom pills, due to getting about 2 seconds of actual speed between 10 second periods of drowsy whitescreen and big slowdown.

Wickedtemp:
Brain damage isn't nearly as dangerous as you think it is, and impedrezene doesn't cause nearly as much brain damage as you think it does. And, as Hero said, apparently the impedrezene is the reason it wasn't working all that nicely to start with.

Wickedtemp:
So, quick update just so you have something to look at.

Starting with, brain damage is super weaksauce when it comes to actually doing shit to screw you over. Sure, at 60 brain damage, you just kinda drop dead. But the thing is, you don't even suffer any actual downsides from brain damage until you hit 35 points of damage, at which point you suffer hand disability.
https://gyazo.com/634b3af558cefdc448895dcc0b1850cb

Well, what if someone eats an entire bottle of zoom pills? Basically nothing. 
https://gyazo.com/a88687bed91f2b361bd975adfe37b94a

They'll get pain messages from brain damage. If they eat three entire bottles of zoom pills, they'll get blurry vision every now and then. They won't even suffer any mechanical downsides until they hit six and a half bottles of zoom pills, at which point they'll finally accumulate enough brain damage to cause a hand/arm disability.  
https://gyazo.com/634b3af558cefdc448895dcc0b1850cb

So... No, impedrezene doesn't really cause brain damage in any meaningful amount, and no, brainloss isn't really that punishing over time. The only punishing thing that impedrezene does is cause the player to pass out every now and then, but as an 'over-time' thing? There's literally no downside, because you'd need to eat an entire bottle to have any negative effect.

Wickedtemp:
Still think this is a decent enough idea and adds to the whole "You're literally eating pills you found in the fucking trash, what else would you expect." thing. Especially when brain damage does next to nothing and impedrezene can't even be said to cause it in the first place due to it being so weak.

Ketrai:
Idk, drugs can be fun for roleplay. It's kinda shitty that zoom pills just... don't really do anything yanno. Happy pills make you hallucinate for example. Whilst these just give you a headache? Big wow. Steriods you have too, which give you toxins. And finally space drug syringes which give you a similar effect to happy pills. 

It'd be more  warranted if pills were randomized effect and name wise. There could be a thousand knock off brands. And then you can't just go like. "Oh I magically sense that zoom pills don't make me high, therefore I shouldn't take em."

Wickedtemp:
So... at what point will there be discussion on this? Its kinda just been sitting here for two months with pretty much no staff input

SpadesNeil:
Scree expressed problems with this and those problems have not been resolved, so this will not be merged.

Heroman3003:
@SpadesNeil And Scree never responded to reasons changes are  made the way they are. The problem with the pills that is attempted to be fixed is that they don't do either. Tempest's math clearly shows that they do not punish. And my own testing shows that speed boost is completely trashed as well. AKA fixing the official issue mentioned in the header. Buuut, if the zoom pills being entirely useless and entirely unpunishing to use is considered **intentional**, I'd also request you close said issue.

SpadesNeil:
image

Github is stupid I guess.

SpadesNeil:
You use expired medicine now.

code/modules/reagents/reagent_containers/pill.dm
```diff
@@ -292,9 +292,11 @@
 
 /obj/item/weapon/reagent_containers/pill/zoom/Initialize()
 	. = ..()
-	reagents.add_reagent("impedrezene", 10)
-	reagents.add_reagent("synaptizine", 5)
-	reagents.add_reagent("hyperzine", 5)
+	if(prob(50))						//VOREStation edit begin: Zoom pill adjustments
+		reagents.add_reagent("mold", 2)	//Chance to be more dangerous
+	reagents.add_reagent("expired_medicine", 5)
+	reagents.add_reagent("fertilizer", 5)
```
SpadesNeil: Not sure I agree with fertilizer being here, though. Makes no sense.

Heroman3003: Admittedly, I just was looking for a weaker addition to presense of expired medicine, to make toxin buildup not as immedeate and bit more progressive, and only 'weak and slow processing' toxin I could find was fertilizer (which i believe is the weakest toxin in the game). I could make a renamed version of it, just tried to use what already was in the code.

SpadesNeil: I'd rather not use fertilizer, and having one new reagent just for this purpose is kind of dumb.

I suggest just deleting the line outright. We have enough sensible and detrimental reagents here already to balance it.

Heroman3003: Fixed

PR #4990, opened 2019-04-09
Fixes unlocking cyborgs with wallets/micros

Heroman3003:
PDAs were made an exception originally. This way is much better.


Screemonster:
wallets exist upstream, right? They might like this. Even if they don't have micros.

Heroman3003:
Its already sent upstream. This is port of update I made on upstream to already-implemented-here fix for unlocking with PDA

PR #4991, opened 2019-04-09
Disables some Tesla graphical effects

Novacat:
Mostly, trying to figure out why tesla_zap proc is making the server shit the bed. Vorestation edits are on the top of the list. Though if this is a polaris bug then debugging this is going to be a lot harder.


Novacat:
Not the cause, closing.

PR #4992, opened 2019-04-10
Fixes tesla logic bug

Heroman3003:
Honestly, no idea if it is related to the whole lagstorm thing or not, but still worth fixing

Direct port of https://github.com/PolarisSS13/Polaris/pull/6063


Heroman3003:
It was a polaris port, just forgot to mention for a second.

PR #4996, opened 2019-04-11
Fixes xenoarch spawning on station levels

### LINKED ISSUE #4496: Turn off xenoarch spawns on station levels
I'm getting sick of xenoarchaeologists bitching about those mean old engineers and directors telling them not to breach holes to the outside in fucking maintenance because they wanted to dig up a shiny. "but it's my job" they say. "they're just using things like `station integrity` and `safety` as an excuse to stop me" they whine. "How was I supposed to know that anomaly would irradiate those doctors on the other side of the wall, they should have put AMI suits on since I was digging, doctors get those right?"

On a more serious note it's probably good to turn them off so that the people who _do_ have a reason to be digging holes don't get blasted with RNG radiation waves when they accidentally drill through a strange rock.

People on the station levels to NOT have to deal with xenoarch nonsense.

#### What actually happened
Xenoarchs repeatedly breaching/decompressing maintenance because they can't restrain themselves, then complaining when asked to restrain themselves.

Heroman3003:
And makes it so that 7 z-levels of Tether itself, Centcomm level and misc level do not spawn anything xenoarch. Any other area will still do so normally.

Resolves #4496

### DIFF
--- a/code/__defines/map_vr.dm
+++ b/code/__defines/map_vr.dm
@@ -1,0 +1,1 @@
+#define MAP_LEVEL_XENOARCH_EXEMPT 0x080	// Z-levels exempt from xenoarch digsite generation.
--- a/code/controllers/subsystems/xenoarch.dm
+++ b/code/controllers/subsystems/xenoarch.dm
@@ -31,7 +31,7 @@ SUBSYSTEM_DEF(xenoarch)
 
 /datum/controller/subsystem/xenoarch/proc/SetupXenoarch()
 	for(var/turf/simulated/mineral/M in turfs)
-		if(!M.density)
+		if(!M.density || M.z in using_map.xenoarch_exempt_levels)	//VOREStation Edit: exclude some Z-levels from xenoarch spawns
 			continue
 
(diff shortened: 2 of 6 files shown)


Heroman3003:
As a note, I made a similar PR (https://github.com/PolarisSS13/Polaris/pull/6065) on Polaris, but as I'm unsure if its gonna be accepted and when, duplicated it here. If it is and its merged here already, we're just gonna need to axe //VOREStation edit comments and _vr files added by this PR.

Screemonster:
Between this and the sonic jackhammer changes, there shouldn't be any further instances of people being assblasted by rads out of nowhere any more when they just want to make a maint fort. Plus we've been kinda yelling at xenoarchs not to dig on the station and blast the dorms with rads for a while, might as well code it in.

Heroman3003:
Actually, radblasting on station levels is not as big a problem as artifacts unearthed on one end, and the flooding of maints with phoron and/or vacuum done in process of unearthing them, at least from my experience. Either way, it solves it all.

PR #5004, opened 2019-04-11
Makes the Freezer actually cold

Novacat:
Only can be 1C for now, otherwise I would have to go into atmos alarms settings and I am too lazy for that.


Heroman3003:
I really disagree with making freezer cold. Realism is realism, but this is just plain inconvinience for no reason than inconvinience. The moment someone enters freezer, entire bar anb kitchen will click in need to heat up. Realism is realism, but I think freezer not being cold is much more realistic than a door opened for half a second dropping temperature everywhere in the area immedeately.

SpadesNeil:
Considering the nightmare that is ZAS, no.

PR #5014, opened 2019-04-12
Fixes organizations runtiming

Heroman3003:
Direct port of https://github.com/PolarisSS13/Polaris/pull/6070

### DIFF
--- a/code/datums/observation/destroyed.dm
+++ b/code/datums/observation/destroyed.dm
@@ -10,5 +10,6 @@
 	name = "Destroyed"
 
 /datum/Destroy()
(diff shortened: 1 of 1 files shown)


code/datums/observation/destroyed.dm
```diff
@@ -10,5 +10,6 @@
 	name = "Destroyed"
 
 /datum/Destroy()
-	GLOB.destroyed_event.raise_event(src)
+	if(GLOB.destroyed_event)
```
SpadesNeil: // Vorestation Edit

or send it to Polaris

PR #5015, opened 2019-04-12
Fixes runtime in announcer AI

Heroman3003:
Yeee, runtime fixing

Direct port of https://github.com/PolarisSS13/Polaris/pull/6071


Novacat:
Both were sent to polaris, but I have been merging fixes here early since we have been running off a faster development cycle than polaris.

PR #5021, opened 2019-04-13
Fixes step code, again

### LINKED ISSUE #5020: Micros don't pass under people without moving them anymore
Before, as a micro when you ran passed normal/large people, it would give you a message 'you ran between [x]'s legs' and it would not disturb their position.

Now it doesn't do that, even at the smallest size, I was disturbing the position of people seemingly of all sizes.

For big people and small people to be able to step over/under eachother respective to their sizes.

#### What actually happened
Movement functions as though everyone is the same size it seems

#### Steps to reproduce
- Be 25% size
- walk passed a normal or larger sized person
- use your MIGHT to push them out of the way

#### Code Revision
- release - 
0c8f28e931c3a02d580c3c805b0b1e076ef8d004
 
- I didn't test to see if it works properly for macros on normal sized people or whatnot.

Heroman3003:
Fixes #5020

Alt title: Heroman is bad coder who doesn't test jack enough

### DIFF
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -149,6 +149,10 @@ default behaviour is:
 			tmob.forceMove(oldloc)
 			now_pushing = 0
 			return
+		else if((tmob.mob_always_swap || (tmob.a_intent == I_HELP || tmob.restrained()) && (a_intent == I_HELP || src.restrained())) && canmove && can_swap && handle_micro_bump_helping(tmob))
+			forceMove(tmob.loc)
+			now_pushing = 0
+			return
 
 		if(!can_move_mob(tmob, 0, 0))
 			now_pushing = 0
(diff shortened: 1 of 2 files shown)


Novacat:
...Why do we not have a work in progress label for PRs?

code/modules/mob/living/living.dm
```diff
@@ -149,6 +149,10 @@ default behaviour is:
 			tmob.forceMove(oldloc)
 			now_pushing = 0
 			return
+		else if((tmob.mob_always_swap || (tmob.a_intent == I_HELP || tmob.restrained()) && (a_intent == I_HELP || src.restrained())) && canmove && can_swap && handle_micro_bump_helping(tmob))
```
Novacat: //Vorestation Edit?

Heroman3003: Genuinely forgot, whoops

PR #5068, opened 2019-04-16
Fixes snakes being aggressive

Heroman3003:
hiss hiss


Novacat:
Note that this does not apply to giant snakes, just smol pet snakes.

PR #5074, opened 2019-04-16
Moves the recreation area toilet to near the pool

Heroman3003:
Solves problem of its underuse and the problem of surface sec being way too unsecure. As requested.

http://prntscr.com/ncwslx
http://prntscr.com/ncwspt


Novacat:
Leaving staff discussion, just to make sure that it was not intended for people to be able to break into surface sec through the toilets.

Very-Soft:
Smells like 'fuck people doing illegal things'

Which is lame for times when we do actually have people trying to do antagonistic things. Tether's already lame to do crimes on in general, and this change seems like it will make it just a little bit more annoying/make people focus more on just using emags.

Like, while I understand that some people are very occasionally annoying with the stuff that's in there, there's something to be said about like, giving people the opportunity to do fun things with things they're not supposed to have, and sterilizing those opportunities.

It's already faster to just hack your way into that tiny armory once you know the wires or have gloves. Why move the somewhat slower but more discreet and noisy entry point.

Tl;dr: this change is lame and I don't like it.

Heroman3003:
Well, if times when people _are_ annoying with it beyond whats reasonable are not being dealt with when they happen, they probably should be prevented. I am not against 'forbidding illegal things', but what spurred me on doing it is Ascian saying somebody who broke in recently did it to arm up on tasers and go for a full shootout with security. Maybe if people got approprietly punished for things they do with the mini-armory robbing, people wouldn't feel like robbing it is normal thing (which it definitely shouldn't feel like) or that its pointless to try and deal with the people doing it because they just get a laugh and do it very next shift anyway.

I am not saying it being removed is only option. But everything must have its limits. And if limits aren't defined or enforced. Well. It will way too often lean towards extremes, and nobody has fun during extreme of either total chaotic robusto self-antag or total no fun orderness.

Very-Soft:
If people are making a habit of gearing up to fight security, that's an ahelp issue, not a 'change the map' issue. 

Now, I play and observe pretty dang often, and the main reason people -usually- break in there for is to nab security keys. Now, I know that the ahelp bot has been a bit unreliable lately, but someone gearing up to fight security when they aren't an antag is 1000% one of the reason ahelps are there for. If people are doing that, and admins aren't responding to ahelps on station, then let us know on discord.

But someone making a habit of doing shitty behavior, isn't a cause for changing the map to try to fight it.

Heroman3003:
Even with the key, the issue is that people nabbing the key has become 'normal'. When it never should really be. And that anyone who tries to actually. Well. Contest that? Is laughed at and mocked. Because everyone is so used to this place only existing to be robbed. That robbing it became 'normal'. And let's step back from the situation and think about it for a second: "Should robbing an armory be normal?". Plus, its the only time department gets actual headset stolen with actual regularity. And only reason for that is because its easily robbable, and because the headset is out in the open basically. For people who shouldn't have it to have.

If the only reason it exists is so that people can bait security into trying to stop them from robbing it and then mocking the said security for trying to stop something 'normal'? Well, I honestly do not think that reason is good reason.

Thats the problem with the place. Its in concept an armory, but in practice it only exists to be robbed, and people trying to uphold any sort of standard regarding it end up mocked for challenging frankly illogical 'normality' of it.

Very-Soft:
I think you're being a little exagerative here. While yeah the main reason people break in there is for the headset keys, I'd hardly call people breaking in there a normal part of a shift.

Like, that's a thing that happens less than once a week if I had to make an estimate from my observations. And generally, again, if people -are- making breaking into there on a regular basis, then that's more of an ahelp issue imho.

Like, every time I've seen that place get broken into and that gets noticed by security, security makes a big deal of it and goes to arrest the person and retrieve the stolen items.

So I'm really not seeing how any of this is really super normalized and such, and don't understand why we need to change the map for something that's honestly not a problem. Like, again, that place isn't hard to actually get into without going through the wall. You're only really narrowing down people's options for entry into it to basically one path, and effectively enforcing that, by making the area around there be outside air.

From an antagonist gameplay standpoint, this is a lame, restrictive change, that's not actually going to stop people from getting into the place, and is just instead going to bottleneck people into one option.

SpadesNeil:
That little armory never really got robbed anyway so nothing of value is being lost.

Heroman3003:
Well, primary reason for making that PR has not been any of it. While I am displaying my opinions, original intent of PR was sparked by conversation about the toilets never being intended as access point to the armory in the first place, which is also what the discussion is aimed at. And while not 'as' regular, it does have a sense of... Being place people turn to whenever they want some sec loot. Be it for fun reason, genuine antaggy reason or malicious self-antaggy one. Because of simple overease of accessing it. As 'easier' as it is to access through front door, front door is also much more visible.
I know of at least one situation account where somebody passing the toilets just sorta. Snagged themselves a headset key and a taser. Never planning to use them for anything, simply because they were passing by and it was that easy and simple, so why not? And that just highlights type of reputation the mini-armory has.
Plus, lowkey am hoping for the toilets to be more in way of traffic to gym and pool, because that bit of area near gateway is pretty much deserted, so they may get some attention in much better way than just looting sec.

Novacat:
I can see arguments either way, which is one of the reasons why I put staff discussion on. The main thing, from a staff point of view, is that _are we okay with people easily robbing this armory every once in a while._

Note that anyone determined enough, even with the armory outside, could break into it via outside (using inflatables) or through the gateway room, though both cases will require more time and preparations.

That was the main question I was posing to Staff.

PR #5101, opened 2019-04-19
Implements switching on/off duty midround via timeclock terminals

Heroman3003:
Well, that was what timeclocks were for in the first place, so here you have it.

Allows to switch off-duty if PTO of your current department is positive and allows to go on-duty as any non-head, non-IAA, non-secretary job you are not jobbanned from (including alt-titles), all using timeclock terminal.

15 minute cooldown between joining and first ability to switch, and further 15 minute cooldown after every switch.

Switches are announced over common channel, similar to arrival/departure.

A config option, by default it is OFF (same as PTO gathering in general).

### DIFF
--- a/code/controllers/configuration_vr.dm
+++ b/code/controllers/configuration_vr.dm
@@ -5,6 +5,7 @@
 /datum/configuration
 	var/list/engine_map	// Comma separated list of engines to choose from.  Blank means fully random.
 	var/time_off = FALSE
+	var/pto_job_change = FALSE
 	var/limit_interns = -1 //Unlimited by default
 	var/limit_visitors = -1 //Unlimited by default
 	var/pto_cap = 100 //Hours
@@ -52,5 +53,7 @@
 				config.pto_cap = text2num(value)
 			if ("time_off")
 				config.time_off = TRUE
+			if ("pto_job_change")
+				config.pto_job_change = TRUE
 
 	return 1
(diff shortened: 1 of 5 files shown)


Novacat:
Itsabouttime

Wickedtemp:
What happens to Head roles going off-duty? Do they go to Off-duty Medic/Engineer/Officer, or Worker?

Heroman3003:
Four department heads go to appropriate department. HoP and CD go off-duty worker.

Mewchild:
LEMME JUST PUT A FAT FUCKIN +1 ON THIS

Wickedtemp:
Assuming that going 'off duty' to circumvent demotion would be seen as being identical to leaving the round?

Heroman3003:
Jobs that don't earn PTO (visitors) and TERMINATED can't even access that part of menu.

Wickedtemp:
I mean more like "I'm going offduty instead of handing my ID to the HoP so you can't officially demote me." 

Currently, having people cryo before the demotion is actually handled is kinda a common thing, happened a couple days ago, too.

Heroman3003:
That'd be up for admins to decide then. But generally, I believe demotion is the punishment. And when that punishment is not applicable, other options can be used, as its not the only one.

Screemonster:
a h e l p

PR #5110, opened 2019-04-21
Allows the liquidfoods to be edible through gasmasks and spacesuits

Heroman3003:
Liquidproteins too. Can eat through gasmasks and subtypes, all space helmets and biohazard suit hoods.

### DIFF
--- a/code/__defines/flags.dm
+++ b/code/__defines/flags.dm
@@ -38,6 +38,7 @@ GLOBAL_LIST_INIT(bitflags, list(1, 2, 4, 8, 16, 32, 64, 128, 256, 512, 1024, 204
 #define NOSLIP					(1<<2)	// Prevents from slipping on wet floors, in space, etc.
 #define BLOCK_GAS_SMOKE_EFFECT	(1<<3)	// Blocks the effect that chemical clouds would have on a mob -- glasses, mask and helmets ONLY! (NOTE: flag shared with ONESIZEFITSALL)
 #define FLEXIBLEMATERIAL		(1<<4)	// At the moment, masks with this flag will not prevent eating even if they are covering your face.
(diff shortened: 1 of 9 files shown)


Screemonster:
Emergency. Induction. Port.

Novacat:
I like this idea, the only question is if Polaris wants it.

Screemonster:
I was wondering about having a special flask that would let you drink through a straw and have something about having to spend a moment to screw it on, after which you could drink at normal rate until you put the flask away, but this works too.

One thing though - does this feature allow forcefeeding, or does it only allow people to eat voluntarily? 'cause you shouldn't be able to forcefeed someone via a straw with a mask.

Novacat:
I would say forcefeeding is fine, because, usually, if you are trying to forcefeed someone in a voidsuit, it is probably because they are downed. Memories of getting a canteen of water to the guy who is laying on the ground after being shot does suggest that it would be possible.

Heroman3003:
Currently it does allow forcefeeding. But if it isn't desired its one change away. Plus it only applies to forcefeeding the rations anyway. Adding a drink flask functioning in similar way wouldn't be hard, but I do think it would be much easier to misuse compared to just specific food items.

Screemonster:
It's more that you'd be able to hook the flask to the port, but you can't force the drinking straw inside the port into their mouth without taking their helmet off.

Novacat:
If someone is floored and unable to use arms/legs/hands, you should be at least still willing to feed them liquidfood. But if they are physically unconscious _then_ you would have to go for removing their helmet. So, I would keep force-feeding other people, but, add a consciousness check for the person being force fed.

Heroman3003:
Upstream PR merged

PR #5111, opened 2019-04-21
Removes glucose autoinjectors from vendomats, makes Zaddat spawn with a bunch

Heroman3003:
Because people shooting themselves up with sugar water is absolutely dumb. And it was shot down before. It was only implemented because of zaddat, but aparently people can't have nice things without overusing them in situations where they shouldn't be overused.

No more glucose injectors to abuse in non-danger environments or empty vendomats of.

Zaddat now spawn with a lunchbox with 6 of them. Plus with #5110, it should make sure Zaddat are in no danger of ever starving because 'no food hole'.


Mewchild:
I personally disagree with this change. These items are being used in the same way that protein shakes are, save for the fact that they're usable in a single click. They're also the only food item usable from within a voidsuit, which gives them a use for miners, explorers (on V2), and other jobs that require frequent EVA.

Wickedtemp:
Also disagreeing with this. I don't think anybody is overusing these, and if they are, it's probably just because it's shiny and new compared to food rations and protein shakes. Alraune could have a use for these as well, as would any character that's relatively plant-like or has an unorthodox digestive system, like Tempest for example.

Heroman3003:
And, yet again, people are thinking in amount of clicks and ooc newness rather than basic IC logic. This is exactly the reason the change is being made.

Screemonster:
Way too many people thinking in gamey terms here. "oooh it takes less clicks so it makes total sense for my character to eat intravenously instead of using their fucking mouth" what

SpadesNeil:
I don't like the idea of people using _fucking needles_ to get nutrition anyway unless they are literally in a hospital bed or something in which case somebody can IV them nutriment or something.

Wickedtemp:
If someone IV'd a patient nutriment, they'd be poisoned and that'd be malpractice/assault. I don't even think it'd provide nutrition, just poisoning.

SpadesNeil:
... Since when the fuck is nutriment considered poison?

Novacat:
Nutriment when injected poisons someone, because it is supposed to be ingested, not injected. Glucose is the injectable form of nutriment.

Wickedtemp:
Since we've been on Polaris code and since Bay seperated Bloodstream/Stomach

Screemonster:
I'd be all for glucose being something you can produce in chemistry, hell "sugar" already exists! Okay, "sugar" as most people understand the term is gonna be sucrose rather than glucose but _really_

SpadesNeil:
Wait so does Polaris have sugar and glucose as two separate reagents?

Screemonster:
Yup. Glucose is literally nutriment with the "injectable" var set to true. It's been in the code for a while, but pretty much the only thing that used it was a virus effect that injected you with it and made you fat.

PR #5114, opened 2019-04-21
Expedition Sub-department

Novacat:
- Moves explorer job roles from their respective departments into an Exploration sub-department
- Adds Off Duty Explorer
- Adds departmental ID cards for all off duty roles
- CHAIN OF COMMAND WILL REMAIN UNCHANGED
- ACCESS WILL REMAIN UNCHANGED
Note: Pathfinder's minimum access is getting tweaked, in case minimum access ever gets used.

This is a purely clerical change that is being done for two reasons.
1: To enable off duty exploration role, important for the upcoming timeclock PR. The upcoming timeclock functionality will cause difficulties for expedition roles, due to, at present, being in a wide variety of different departments.
2: Generally, explorer roles in other departments tended to create confusion, as people would ask (or even demand) those roles to do station-side duties, even if they lacked the access to.


Screemonster:
exploration has been effectively a sub-department like cargo for a while, might as well represent it in code

Wickedtemp:
So, Field Medic would show up in Exploration instead of Medical. 

I kinda like this because exploration never really "fit" entirely in any one department. There are opportunities for nearly every department to contribute. But on the other hand, with this, I do think the expedition prep room should be modified to serve as an acceptable bare minimum prep room for all expedition-related roles. Field Medic, despite now being specifically in Exploration rather than Medical, would still need Medical for all of their equipment. If they were given a few materials, like a defib, combat kits, a few assorted boxes of injectors, I think it'd kind of help to signify that it isn't just "ParamedicPlus".

Novacat:
Out of scope, and would require a lot more discussion and design planning. While Field Medic is 'listed' under the exploration department, they still answer to the Pathfinder and CMO, and would still function as they do now. Field Medics can still do on-station medical work, especially during emergencies, but the idea is to push people from the mentality of 'Oh, we have a single field medic on shift, we will be fine' to 'We have no station-side medical, we should fax for some.'

Heroman3003:
It'd act similarly to way Geneticist used to be under split authority of RD and CMO, while listed only in medical manifest. Another dual-department role.

PR #5136, opened 2019-04-25
Solargrub hibernation season is over

Heroman3003:
Solargrubs. Most hated pest on the station. Is back!

REEEEEjoice.
They are also notably stronger, smarter, etc. Mob rework and all. Larvae will be faster, more reactive. Big uns now inject grub-toxin and their shocks are now not just EMPs, but also proper shocks.

To counteract their strength, minimal amount of grubs spawned is twice as small now as well, giving a much wider range of pest numbers, and notably less when counting on-average.


Heroman3003:
On a sidenote. While this PR makes solargrubs fully and at best functional within existing code. Larvae's behaviour will look rather shoddy until https://github.com/PolarisSS13/Polaris/pull/6132 is merged upstream and then sent there due to priorizing bugs.

TheDavestDave:
Why. Engineering stopped wasting hours looking for grubs before they were removed and made departments look. That don't even the the APC monitor to help them look. They also grind a department down to not working because they are hidden somewhere no one has acess to, such as a sever room or head office.

Very-Soft:
I actually like the solargrub event. My main issue with it has always been exclusively how often it occurs. Like, it should be a once every few days kind of event.

Hell, I'd even be totally cool with it being in the code, but disabled by default, so we can flip it on on shifts where there's a lot of people.

TheDavestDave:
I don't have problems if it is admin toggle rather than every shift. Even with 1 or no engineers to even start the engine

Wickedtemp:
Grubs really aren't a big deal as long as someone in each department does a departmental sweep. I'd do it every time in Medical and I'd be able to toolbox every larva before we even got the actual CC announcement about it. It only becomes a problem when they spawn in an area that nobody can get to or when they spawn into a department and that entire department doesn't want to pause their scene for a few minutes to help.

SpadesNeil:
I miss solar grubs, but indeed, the frequency should be turned down to favor other events.

Heroman3003:
Reduced the likelyhood. Now event is extremely unlikely to happen when no engineers/sec are around, and likelyhood when those are around is lower and scales slower with amount of people in said departments.

PR #5209, opened 2019-05-09
Overhauls Fennix Tail.

Ketrai:
Overhauls the fennix tail. New sprites :

Do note that the shading oddities only appear on the gifs, not ingame. Couldn't fix it.

https://cdn.discordapp.com/attachments/351885478863044610/576135607118725120/wag4.gif
https://cdn.discordapp.com/attachments/351885478863044610/576123463354548224/wag2.gif
https://cdn.discordapp.com/attachments/351885478863044610/576123482207682587/wag3.gif

Old sprites :
image
image
image


Sebbe9123:
Looks really great except when facing away from the camera and the tail is in the perfect middle, it becomes paper flat.

Ketrai:
> Looks really great except when facing away from the camera and the tail is in the perfect middle, it becomes paper flat.

True, I could make it three frames instead to make it last longer.

Ketrai:
> I would say it looks different enough from the old tail that it should probably be defined as a separate, additional tail than a replacement.

I suppose so. Though the old tail has a few colour errors on the vwag. If anything I'd say the vwag for it should be removed until that's fixed.

> Also, the visible wind effects should _probably_ be removed since, as far as I know, they are not used on any other character sprites and do not fit the current character aesthetic very well.

I'd argue it's just the animation style. There's already a lot of different aesthetics amongst the various tails because they were made by various people. I'd rather hear what others have to say about it first. See if they also deem it to be too out of place.

> Is it intentional that the tail wags more heavily to the left than right?

It is. It's big, heavy, and fluffy. Easier to just swing it half way unless there's a nice desert breeze.

> I am assuming that the animations are slowed down. Otherwise it looks good.

Mhm. They're at the same amount of speed per frame as other tails. But due to the large amount of frames, I'll speed it up by 25% soon.

Heroman3003:
I agree that wind effects look out of place, especially so if somebody were to qwag. As for speed, try basing animation speed off of amount of frames it takes to reach the end of the wag and back, rather than fps of other tails

Novacat:
My current preference is to keep the old tail and the vwag for it. Even if it has a few minor errors I do not feel that warrants throwing the whole thing out.

"It is. It's big, heavy, and fluffy. Easier to just swing it half way unless there's a nice desert breeze."

Usually, tails default to the middle, so a half-way swing would probably still evenly go both ways?

Ketrai:
> my current preference is to keep the old tail and the vwag for it.

Aight. I'll keep the old tail in to add to the flood of choises. 

> usually, tails default to the middle, so a half way swing would still evenly go both ways?

There's plenty of tails that start sideways including the often used vulpakanin tail. I already animated it this way now to match better with the other directions like the side one. Which would look rather weird if the tail sticks our too far in front of the sprite since I tried that before. Having it swing from the ground from one side to the other or through the air from one side to the other, but still requires me to redo the animation. Which with how I like to nitpick whenever I make something would require me an hour or two to fix at least. A lot of effort just to appease one person -v-.

Ketrai:
Tested. works perfectly now.

Novacat:
Those were mostly suggestions, anyway, the big things were keeping the old tail and removing the swish.

Novacat:
Was there any reason the xeno tail was modified?

Ketrai:
No modifications were made. I accidentally pasted the tail there whilst I was testing. There shouldn't be any difference.

PR #5229, opened 2019-05-15
Macros can pickup unanchored structures and machinery

Zap-zapper:
If you are above 150% size, you can pick up unanchored structures and machinery. To do so, you need to altclick it while NOT being on help intent. 

Demo:
https://www.youtube.com/watch?v=xHm3kDdOzHk

Those macros finally will be useful in cargo.

### DIFF
--- a/code/modules/mob/living/carbon/human/human_defense.dm
+++ b/code/modules/mob/living/carbon/human/human_defense.dm
@@ -488,7 +488,8 @@ emp_act
 	catch_chance = between(1, catch_chance, 100)
 
 	if(prob(catch_chance))
-		return TRUE
+		if(istype(O,/obj/item) || size_multiplier >= RESIZE_BIG)		//VOREstation edit
+			return TRUE
 	return FALSE
 
 /mob/living/carbon/human/embed(var/obj/O, var/def_zone=null)
--- a/code/modules/vore/resizing/holder_micro_vr.dm
+++ b/code/modules/vore/resizing/holder_micro_vr.dm
@@ -35,3 +35,89 @@
 	for(var/atom/movable/A in src)
 		A.forceMove(here)
 	return ..()
+
+
+
+/obj/item/weapon/holder/structure
+	w_class = ITEMSIZE_HUGE
+	slot_flags = null
+	item_icons = list()
+	pixel_y = 0
+	var/mob/living/grabber
+
+
+/obj/item/weapon/holder/structure/sync(obj/M)
+	dir = 2
+	overlays.Cut()
+	icon = M.icon
+	icon_state = M.icon_state
+	item_state = M.item_state
+	color = M.color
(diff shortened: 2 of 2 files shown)


AaronSherwood117:
Honestly? This doesn't affect much at all gameplay-wise and anyone can drag around unanchored bits anyways.  Support as an immersion thing tbh, sounds fun.

Wickedtemp:
Does this count for lockers and crates?

Zap-zapper:
> 
> 
> Does this count for lockers and crates?

It counts for any /obj/structure and /obj/machinery which is NOT anchored to the floor

Wickedtemp:
Neat! I like this. It makes sense and sounds like it could be a lot of fun.

AaronSherwood117:
Just for posterity, image doesn't fit with the behaviors of other containers like backpacks that people can be placed into, and it should be considered whether it needs to be changed or not.

Heroman3003:
I really do not like balance implication of being able to carry thrice as many pull only objects. Macro/micro advantages/disadvantages have always been first and foremost RP-side, with only minor fluff details actually reflected in mechanics. This is not minor fluff detail. Plus I can't even begin to think of amount of bugs that this will cause due to specific interactions and such. I understand intent behind this PR. But I do think that it plain shouldn't be a thing at all. This goes past minor immersion changes and plain and simply makes macros as effective as a full-on cargo tug. Which is not good balance-wise.

AldenBluetenberger:
Eh, I like this idea. It makes sense that they can pick up unanchored things.

Heroman3003:
It also means that some people will be going macro just for advantages of it. TBH, I dont think sizechanging should be a mechanical advantage in such a big way. Either everyone can do it. Or noone can.

Zeldazackman:
Neat, wish this sorta thing existed for a scene a while back wherein someone was "carried" over the shoulder in a bodybag to the empty morgue and "totally not" just playing dead before turning into a bit of filler.

Was awkward having to up and drag the damned bag behind while it was being "hoisted" and then having to go back because things weren't able to be dragged through Z-levels at the time due to a bug.

SpadesNeil:
I also do not like the balance implications of this.

I want to explore the edge cases of issues like "What if you get shrunk?" but I'm not even going to bother suggesting checks for that because I **do not** want this. It's too complicated to balance properly considering all the things we'd have to consider.

Sorry.

Wickedtemp:
It should be noted that, in response to Heroman's concern, this wouldn't be an exclusively Macro Character Only thing. Literally anybody can change their size whenever they want with NIFs, size guns and the size changing reagents. 

And personally I don't really care if someones reason for a macro character is this PR. Why they choose their character attributes doesn't affect me in any way and I'm not going to give someone flak for making a macro Quartermaster just because they'd be able to hold and carry crates.

PR #5263, opened 2019-05-21
[MIRROR] Sif Animal Expansion 1

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/6038
--------------------
Additions:
Six player-control-outfitted mobs.
- Ice Leech, a phorogenically mutated Sivian leech which must be removed by a surgeon, lest the host fall prey to organ death and bloodloss. Will randomly produce chemicals if able, many being beneficial.
- Sakimm, creatures resembling a cross between a rat and raccoon, relatively weak but numerous in their dens. Can wear hats.
- Intelligent Sakimm, relatively uncommon variant of the Sakimm, they will prefer to steal items of value, and if engaged in combat, attempt to disarm / stun their aggressor while calling for aid from their kin. Will steal and wear hats if left unattended.
- Crystal-Feather Duck, simple animal that will become aggressive if you have food. At all. Rock-digging claws on their feet, and sharp wing-feathers mean they can remove limbs.
- Frostfly, a tentatively dangerous flightless insect, produces a cryogenic gas and saliva which are both used to deal with predator and prey alike. If they run out of energy they will attempt to flee, rather than continue fighting, unless enraged.
- Glitterfly, basically a butterfly, comes in MANY colors! Small chance to become hostile if they are within range of a dead glitterfly. Is not dense, so you can walk through their clouds without obstruction.
- Sparkling Glitterfly, an incredibly rare mutation of the Glitterfly, is visible over the darkness. More dangerous in combat.
- Kururak, an uncommon coastal cat which hunts in packs, utilizing a 'rule of the fittest' mindset. Will consider certain individuals 'allies', and flee if nearing death.

Tweaks:
- Glass Jars now respect if a mob is supposed to be visible over the darkness layer when placed on the ground.


KasparoVy:
The duck and that cat thing have pretty sick sprites. Real nice PR from Polaris

PR #5279, opened 2019-05-28
Access fixes for carogopacks

Heroman3003:
- Fixes armor crate having security access instead of armory access

- Adds EVA, explorer and pilot to accesses of hardsuit crate

- Adds EVA to accesses of industrial hardsuit crate

- Adds atmospherics access restriction to thermal regulator crate

### DIFF
--- a/code/controllers/Processes/supply.dm
+++ b/code/controllers/Processes/supply.dm
@@ -226,7 +226,7 @@ var/datum/controller/supply/supply_controller = new()
 				A.req_access = list(SP.access)
 			else if(islist(SP.access))
 				var/list/L = SP.access // access var is a plain var, we need a list
-				A.req_access = L.Copy()
+				A.req_one_access = L.Copy()		//VOREStation Edit: Lets make sense
 			else
 				log_debug("<span class='danger'>Supply pack with invalid access restriction [SP.access] encountered!</span>")
 
--- a/code/datums/supplypacks/security.dm
+++ b/code/datums/supplypacks/security.dm
@@ -31,7 +31,7 @@
 	cost = 40
 	containertype = /obj/structure/closet/crate/secure/gear
 	containername = "Armor crate"
-	access_armory //VOREStation Add - Armor is for the armory.
+	access = access_armory //VOREStation Add - Armor is for the armory.
 
 /datum/supply_pack/security/riot_gear
 	name = "Gear - Riot"
--- a/code/datums/supplypacks/security.dm
+++ b/code/datums/supplypacks/security.dm
@@ -31,7 +31,6 @@
 	cost = 40
 	containertype = /obj/structure/closet/crate/secure/gear
 	containername = "Armor crate"
-	access = access_armory //VOREStation Add - Armor is for the armory.
 
 /datum/supply_pack/security/riot_gear
 	name = "Gear - Riot"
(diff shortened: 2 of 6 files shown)


code/datums/supplypacks/security.dm
```diff
@@ -31,7 +31,7 @@
 	cost = 40
 	containertype = /obj/structure/closet/crate/secure/gear
 	containername = "Armor crate"
-	access_armory //VOREStation Add - Armor is for the armory.
```
SpadesNeil: This was not put in the correct file to begin with, much less coded correctly. This line should not be here at all.

code/datums/supplypacks/security.dm
```diff
@@ -31,7 +31,7 @@
...
+	access = access_armory //VOREStation Add - Armor is for the armory.
```
SpadesNeil: This line does not belong in this file.

code/controllers/Processes/supply.dm
```diff
@@ -226,7 +226,7 @@ var/datum/controller/supply/supply_controller = new()
 				A.req_access = list(SP.access)
 			else if(islist(SP.access))
 				var/list/L = SP.access // access var is a plain var, we need a list
-				A.req_access = L.Copy()
+				A.req_one_access = L.Copy()		//VOREStation Edit: Lets make sense
```
SpadesNeil: Why?

code/datums/supplypacks/security.dm
Heroman3003: Can I clarify why exactly doesnt it belong in that file? Access of the crate ordered from cargo is defined in the supplypack itself. And it overrides the crate's own access restriction.

code/controllers/Processes/supply.dm
Heroman3003: Basically, to allow existance of crates that can be opened via different accesses. The problem never came up because this section of code was unused up until now, with no supplypack in existance using more than a single access yet.

code/datums/supplypacks/security.dm
Heroman3003: Should it stay deleted because VORE EDIT is no longer wanted? Because in _vr files, nowhere is there access restriction defined for that crate. Which means that addition was not done at all at any point, and it still used access_security.

code/datums/supplypacks/security.dm
SpadesNeil: https://github.com/VOREStation/VOREStation/blob/master/.github/CONTRIBUTING.md

code/datums/supplypacks/security.dm
SpadesNeil: It should stay deleted because whoever added this line in the first place is a moron.

Not only did they not add the change to the correct file, they added a change that _literally doesn't do anything_. `access_armory` is just sitting there uselessly. I don't know how this even compiled.

code/controllers/Processes/supply.dm
SpadesNeil: Perhaps it would be better to go up to Polaris and add support for choice between a crate that requires one access, or requires all the accesses.

code/datums/supplypacks/security.dm
Heroman3003: My question was. Whether the change in access on VR side should be kept at all or not. But that answers it and I will just move it to security_vr

code/controllers/Processes/supply.dm
Heroman3003: I'll send that to Polaris, but until its implemented, I'll leave this PR using an edit.

PR #5296, opened 2019-06-01
Adds a fish farm.

Ketrai:
Adds a fish farm.

I also went ahead and fixed the virgo4b water to be the fishable water, so now you can swim proper.

 Of course, swimming in the fish farm is discouraged. Fishing is encouraged, the breeder farms keep up with demand.

Ic justification? An alternative way of producing food to allow the station to be more self-sufficient in case of a disconnect with the colony.

How does it work? Sci-fi magic really. Underwater breeder farm for fish and algae, likely. This is more so just so we can actually fish without going into the wilderness. I will likely make phoron water and phoron-themed fish later for a seperate area. Yes, you can find lakes in the wilderness (not even guaranteed). But the complete lack of comms and agressive weather makes it very unrelaxing. And not everyone has access to void suits. Fishing is something fun to do whilst you chat with other crewmembers.

There's no access restriction. Like with the public garden everyone can go fish at their own leisure using the rods provided. You still need chef equipment to prepare the fish correctly.

The fish farm starts with two bass and a salmon. Just to add some flair even when no one's caught any fish yet.


Ketrai:
Yup, need to use the mapmerge tool. Which I don't know how to use so give me a moment to figure that out or help me figure it out.

SpadesNeil:
The awkward position and shape of this room really bothers me. Why not just use the atrium pond?

Ketrai:
Mainly because everyone and their mother seems to be against the idea of being able to catch actual fish in the atrium pond. Also I dunno. The shape and position doesn't feel that weird if you're walking around in it. It mainly feels weird looking at it. Hardly more awkward than the shuttle bay in my opinion. In any case. I'm all for just having the pond be fishable . But then that implies you can kill the koi and eat them. Which I don't like.

Heroman3003:
Try throwing a fishing rod into any decorative Koi Pond irl and count seconds before you're arrested. Because those ponds are not full-on ecosystems, and fish in them are usually are just equivalent of an open-air fishtank.

Also, inb4 doctors start fishing there just to cook simplest fish dishes and extract carpotoxin, just like happened with carpburgers in cargo crates.

Heroman3003:
Also: Enormouslyt gigantic area with almost every external wall being a window on surface 3, and only a single vent and scrubber. Yep if meteoroid ever hits there, this is not getting cleaned out.

Ketrai:
Meteor doesn't happen often. and there's a lot of places for it to hit. I could always add some large scrubbers, if that's your complaint. This is just the same as a "what if the atrium got hit by a meteor" case at this point.

By the way, the fish you can catch don't contain any carptoxin. they contain some kind of neurotoxin. Not useful for medical, and you have to be a chef to safely prepare them.

Ketrai:
See here for more details on fishing https://github.com/PolarisSS13/Polaris/pull/6113

- Real fish meat. Still toxic, but not carpotoxic.

All fish caught by fishing have neurotoxic protein instead. It's neutralized either by the succ from the chemmaster, or by injecting anti-toxin. One unit is generally enough to neutralize it in any fish. It turns into normal protein.

Ketrai:
uhh, wtf travis? I just added those tiles that you're erroring on :V

TheFurryFeline:
If you haven't already, shift the chairs and tables over one space near the entrance?

Ketrai:
I moved the fishing area to tether level 2, removed the dome, and added fishable water to virgo4b beach. 

Screenshots updated

Ketrai:
https://cdn.discordapp.com/attachments/187013248309002240/584845242356858908/unknown.png

I will eventually sprite in a shoreline with waves and beach sand underground.

Pear120:
This shit makes my lamprey moist.
More recreational stuff like this is always appreciated, and I love the cave feel, and the fishing catwalk.
I'd personally say slap some walls on either side of that southern maint door, and walls under each lightbulb, but that's just minor stuff.

Novacat:
Can you make the shallows a bit wider?

Ketrai:
I already made the shallows a bit wider for it to look better. Just haven't updated the screenshot for it yet. I also noticed I have a station intercom to fix. And well I should add one to the place as well.

Heroman3003:
You should also prolly delete .before files. There is no need for them anymore.

Ketrai:
There, spacious shallows now.

SpadesNeil:
This new setup is much nicer looking.

PR #5320, opened 2019-06-07
Adds a durasteel fishing rod.

Ketrai:
Just a faster fishing rod, roughly 33% faster than the modern fishing rod.

- made from two sheets of durasteel, just so it isn't too pricey.
- can be bought for 7500 mining points, for those over-zealous miners.

- Absolutely huge pr, 12 lines changed!

### DIFF
--- a/code/modules/fishing/fishing_rod_vr.dm
+++ b/code/modules/fishing/fishing_rod_vr.dm
@@ -1,0 +1,5 @@
+/obj/item/weapon/material/fishing_rod/modern/strong
+	desc = "A extremely refined rod for catching fish."
+	default_material = "durasteel"
+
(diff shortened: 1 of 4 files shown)


Heroman3003:
Would that not be completely fine sent to Polaris though? They have mining vendors and durasteel both as well, dont see the problem with that.

Ketrai:
Dunno. I just want more uses for durasteel really. It's kind of this extreme niche metal that's good for walls. As it makes explosive resistant walls. And then for the odd use as a reflective material for lasers. Being too slow and makeship after being kinda crappy, it ends up not really seeing any use other than research. As for Polaris? I already have a fishing pr open there. I'd rather not add anything to the files until that one is merged.

PR #5325, opened 2019-06-09
Brings back oties and adds riding.

Verkister:
-Brings back oties. Currently adminspawn only though.
-Certain niche features still missing such as sec versions reacting to crime.
-Added red guard version.
-Tamed ones can still be befriended etc.
-Minor fixes to sprites and file naming.
-Adds riding to player controlled oties and corrupt hounds.


Wickedtemp:
Might be out of scope, but can we do something about the whole...

> Sec gets a ""totally safe Sec Otie"" from Cargo
> Tells everyone its tame and safe
> A few people are skeptical but its not like any rules or regulations are being broken so all they can do is wait for shit to go wrong
> Someone pets it wrong or pulls the Otie
> It attacks them
> It attacks everyone who tries to stop it
> Its eventually killed, Sec gets bitched at and whoever touched the Otie is also bitched at despite really just getting screwed by bad code
> Every time Sec orders an Otie, 20% chance this happens again

Verkister:
Yeah probably gonna get those fixed before I make them orderable and such. Shame nobody bothered to read the instructions where a tamed one would calm down when petted with help intent, and that only empty-handed help intent petting counts as petting so trying to "pet" them with weapons or punch intents would be a bad idea to begin with. Can't remember if they still got grab aggro'd, but the grab interaction was replaced with the function to get them stop following, but gonna test and fix those later.

HzDonut:
Tempest is also forgetting the fact that even when the security oties operated correctly, it still tended to go horribly wrong since they'd act like hostile mobs and attack criminals before eating them, which tends to result in people panicking and killing the otie before it murders the suspect. If i remember right the other problem of them always being on digest got fixed in the past at least, so figuring out a way to preserve them bellybrigging criminals without the maiming would probably be a good thing to do while you're workin' on them. Maybe them just pouncing and stunning the suspect would work? That way if the suspect has mob vore on they get harmlessly nommed, and if not then it's just like the otie is pinning them down for the officer to arrest 'em.

Zeldazackman:
So at what @HzDonut said, couldn't the space carp chained pounces/knockdowns also work as a "detain" for a sec otie?

Sebbe9123:
should probably make the synthetic oties path to a /mechanical/ one instead of a /vore one, it'll do the eyes and everything for you as you can see in the mechanical.dm file.

Verkister:
Will see about fixing up the sec AI and stuff later and just do the merge before any sprite file conflicts happen again.

Verkister:
Yeah I've got some fresh new plans for the mob sprite file so better get this PR out of the way.

PR #5328, opened 2019-06-10
Fixes geese escaping the wilderness

Heroman3003:
Changed to just be shutters in the wilderness, controlled by buttons on both sides, instead of airlock. Public access.

Also expands area of the shed a bit, both to add more buffer to PoI spawning and to make things on its external wall powered.
Backend changes to how returnhome AI works remain.


Screemonster:
personally I'd rather just let them roam and replace the wilderness gate with a button-operated shutter so the mobs can't open it.

Heroman3003:
Problem is that it will probably just be left open by some rando decided to explore and will just reduce the ofteness of situation rather than eliminating it.

Novacat:
To be fair, it is an extremely rare circumstance to begin with, so just shutters alone would probably work just fine.

Heroman3003:
Recommended changes applied.

PR #5330, opened 2019-06-11
Makes mechanical prefs show disabled OOC/LOOC

Verkister:
Salt PR to rob people from their ability to conceal whether they're ignoring other players' ooc messages mechanically or on purpose and then wondering why things are going to shit and people are getting mad at them.
kuva

### DIFF
--- a/code/modules/vore/eating/living_vr.dm
+++ b/code/modules/vore/eating/living_vr.dm
@@ -667,6 +667,11 @@
 	if(!user)
 		CRASH("display_voreprefs() was called without an associated user.")
 	var/dispvoreprefs = "<b>[src]'s vore preferences</b><br><br><br>"
+	if(client && client.prefs)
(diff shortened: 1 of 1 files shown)


Sebbe9123:
Note that the snarky Shit Verk did in the OP is exactly why I disabled look in the first place.

Verkister:
And now it no longer needs to hinder things for other people around.

SpadesNeil:
Wait so what does this do? Alert people looking at your notes as to whether or not they have OOC disabled?

Because if so that's nifty.

Screemonster:
Also wew I just looked up that conversation in spacelobby and there was something in there about using this to talk shit about people in LOOC without them knowing about it so I figure I should point out that doing so is a breach of global rule 6 and I will fucking ban you for it if I catch you

PR #5344, opened 2019-06-18
Adds a fishing vendor.

Ketrai:
Just adds the existing fishing vendor to replace the bait rack in the upcoming fish farm.

Wow big pr much wow omfg. I'll add a screenie in a sec.


Ketrai:
And the object doesn't have a sprite ingame, groan.

Time to figure out why.

Ketrai:
So it turns out that as a vorestation edit we've changed the root file for vendors to the _vr icon file. Can't we just overwrite the iconstate for the individual vendors we change in the vending_vr.dm file instead??? That way we don't get icon conflicts?

Ketrai:
image

PR #5356, opened 2019-06-21
[MIRROR] Serenity

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/6208
--------------------
What do you get when you mix a Gygax and an Odysseus?

If you answered "a lawsuit from VeyMed", you'd be right! But not any more- for they've finally released their designs for the *Serenity*, a modification on the classic Gygax security mech. Lose all that bulky armor plating and swap most of the weapons for an onboard medical analysis suite, and the Serenity is born.

True to its namesake, it's not made for fighting- so despite being based on a Gygax (and therefore able to equip one weapon), it's still almost as paper-thin as the Odysseus, making that beloved leg actuator feature a little more dangerous to use. When you've only got 150 health, every little sliver counts- but in the medical field, time is vital, and those few seconds might make a difference.

Just as the Durand is the natural evolution of 'a Ripley with a gun duct-taped to it', so too is the Serenity the natural evolution of the Odysseus. Is it fragile? Sure. Will it run out of power fast? Totally. Will Security try and arrest you because "hurr no mechs allowed"? Probably. But is it undeniably sexy as hell?

[You bet it is](https://cdn.discordapp.com/attachments/176163422751293441/587554323936837645/unknown.png).
 
Honestly, if you can afford it, you're probably better off with a Gygax for anything other than purely medical rescue missions, but that's the niche the Serenity fills- reliable, effective, and able to leap tall buildings without getting a wedgie. Get yours today for only three ninety-nine ninety-nine ninety-nine, all rights reserved.


Wickedtemp:
This will need to be modified somewhat if we want to actually use this since Polaris made it admin-spawn only.

mistyLuminescence:
Easy fix- just uncomment the designs in mechfab_designs.dm and circuits.dm. If anyone encounters any issues with it, let me know over on Polaris- we're hoping to make it acquirable in two weeks.

Novacat:
That would need a separate PR, for review reasons.

PR #5368, opened 2019-06-25
Adds a bunch of PoIs to debris field

Heroman3003:
Most for now are generic asteroids and carp-infested debris with few actual points of interest, but perhaps in future it can be more populated and diverse.

Also changes how PoIs spawn, as now they can be closer to the edges (previous buffer zone was fucking 22 tiles from edges, thats like two screens, way too much empty space)


Heroman3003:
Fixed. Accidentally missed that one.

Issue #5239, opened 2019-05-18
Remove .45 pistols + .45 flash rounds from SecVends


MrSign:
#### Brief description of the issue
By technicality, .45 pistols are considered lethal weaponry, therefore should not be freely dispensed by members of security in the code green armory. IIRC, there was talk about removing it aaaages ago. Like, I'm talking last year but it never went through.

#### What you expected to happen
Appropriate weaponry within the appropriate armories.

#### What actually happened
People getting called out for having 'lethal' weapons, despite it being able to get freely dispensed.

#### Steps to reproduce
- Go to a SecVend
-Dispense a .45 pistol
-Get yelled at for possessing a lethal weapon on code green.
-Cry.
#### Code Revision
8c6c1e51e3880cfacf434b15c2ef6ee2034430c6 
 
#### Anything else you may wish to add:
N/A

TheFurryFeline:
Far as I've seen, they're dispensed with only flash rounds and security don't even have access to autolathes to print off lethal ammo. Seems pretty balanced enough.

SpadesNeil:
Yeah but what are flash pistols even _for_?

We could stick in an alternative that only uses flash rounds. The idea of flash pistols always felt unrealistic to me.

TheFurryFeline:
Are there any unused gun objs in the code which could get changed slightly to only take flash rounds? I could probably strike two birds with one stone.

TheDavestDave:
to look cool. It would be nice if flash rounds could damage mobs

PR #5373, opened 2019-06-26
Makes flash pistols only be able to use flash magazines

Heroman3003:
This is a potential indirect solution to #5239, not directly conflicting with #5374 , but adressing same issue in different way. Both solutions can be implemented, but its better that only one is.

Removes name change virgo-side to flash pistols (back to ".45 signal pistol") and limits them to flash magazines exclusively.


SpadesNeil:
I personally don't think flash pistols should be a thing at all. I would prefer them to be removed entirely. The entire concept makes no sense.

PR #5376, opened 2019-06-26
Custom Say Verbs

mistyLuminescence:
https://cdn.discordapp.com/attachments/173831852602294272/593482267733655582/unknown.png

A picture is worth a thousand words. This PR lets you customise the words that appear when you say something- it's usually language-based (with galcomm being 'say', 'ask', 'exclaim/shout/yell' and 'whisper'), but if you want to override it now, you can.

For example...

mistyLuminescence asks, "Hi?" could become...
mistyLuminescence 's TTS device queries, "Hi?"

Great for mute characters or ones with special ways of speaking. There is potential for abuse (e.g. setting it to WGW), but people could do it anyway with custom emotes. I can try and add a character limit if that's desired.

### DIFF
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -1363,3 +1363,29 @@ default behaviour is:
 		BRAIN:<a href='?_src_=vars;mobToDamage=\ref[src];adjustDamage=brain'>[getBrainLoss()]</a>
 		</font>
 		"}
+//VOREStation edit, allows for custom say verbs
+/mob/living/verb/customsay()
+	set category = "IC"
+	set name = "Customise Say Verbs"
+	set desc = "Customise the text which appears when you type- e.g. 'says', 'asks', 'exclaims'."
+
+	if(src.client && src.client.holder)
+		var/customsaylist[] = list(
+				"Say",
+				"Whisper",
+				"Ask (?)",
+				"Exclaim/Shout/Yell (!)",
+				"Cancel"
+			)
+		var/sayselect = input("Which say-verb do you wish to customise?") as null|anything in customsaylist //we can't use alert() for this because there's too many terms
+
+		if(sayselect == "Say")
+			custom_say =  sanitize(input(usr, "This word or phrase will appear instead of 'says': [src] says, \"Hi.\"", "Custom Say", null)  as text)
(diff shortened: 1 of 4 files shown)


Novacat:
At first glance, you might want to do //VOREStation Edit Start and //Vorestation Edit Ends for the blocks of edited code.

KasparoVy:
Nifty automation of something that's a bit tedious to so via `say !...` or `me`.

Wickedtemp:
I hope everyone is ready for "Tempest mlems, "Mlem!"'

Heroman3003:
While its a nifty and long-requested feature, I do feel like using it for mutes defeats the purpose of TTS-the-item, instead of fixing the issues it had in its design. Mayhaps just giving it an ability to be compatible with a headset, or internal broadcaster that can be turned on/off. But separate from specifically TTS-usage, I do think its fine, if only people using it to be misleading intentionally might be a thing, but pretty unlikely.

mistyLuminescence:
I *was* planning on having it so you can set it per character, but then that means savefile stuff and savefile stuff is tricky- if we're going down that route I could use a hand with figuring out where to yoink the code from.

As far as TTS goes, having the device be compatible with headsets would be amazing, but half the time when I played a mute character on Polaris I'd end up just using custom emotes anyway, since the TTS always made a popup box (which interferes with other shenanigans).

Heroman3003:
I could try to help. Although one thing I actually am a bit concerned about. Will those verbs also override verbs from non-GalCom languages? Because for most of those, verbs are primary identification method.

mistyLuminescence:
Currently, yes- they'd override everything, including "stammers" and "slurs" (but the words themselves would still be stammered and slurred). There is the issue already of having a million different languages- and if you can't understand someone, knowing they're speaking, e.g. Enochian, won't really help you understand them any better. Replacing language-specific sayverbs is done so you can, for example, "murmur" or "query" something in non-galcom languages.

mistyLuminescence:
If this goes through I'll see about turning it into a character-specific pref added to character setup (like custom religion/affiliation etc).

Wickedtemp:
Am excite.

Screemonster:
TIME FOR CHIRPS

PR #5393, opened 2019-06-29
[WIP, DNM] Medbeam

mistyLuminescence:
It's a gun that shoots healing! One shoots healing lasers (the BL-3 Phoenix DRS/directed restoration system), and one is basically the NSFW but for medical (the ML-3 Medigun, powered by NANITES).

- [x] BL-3/P handheld
- [x] BL-3/P mech-mounted
- [x] ML-3/M framework
- [x] ML-3/M sec/med separation
- [x] ML-3/M ammo
- [x] ML-3/M sprite + CMO sprite
- [x] Fix pathing on tether 10
- [ ] ML-3/M research recipes
- [ ] CMO advanced magazine shouldn't fit in regular ML3Ms


Wickedtemp:
Years of mech healing being bad in general and overall nonviable... All solved by this PR.

mistyLuminescence:
Still in the works, but we need sprites. I'll poke around and see if anyone can make it happen.

mistyLuminescence:
Sprites all done (courtesy of Kligor on discord)- final step is to add recipes and we're all set for the big opening. PR opening post will be updated with full details.

mistyLuminescence:
i swear it's almost ready i just never have time to finish the recipes

SpadesNeil:
I don't want a healing beam available from round start. This item should be of limited supply and only attainable through R&D.

mistyLuminescence:
This isn't a healing beam like the ones you're familiar with- it's a gun that fires healing, but it doesn't provide infinite heals like the traditional medbeam.

Wickedtemp:
TLDR: They start with a gun that by itself is useless and can't do anything. It needs ammo cells, they get 4 shots per cell before needing to be recharged. There are different types of cell and different tiers of each type. Medical starts with 3 cells. A Brute cell, a Burn cell, and a Stabilizer cell all of the lowest tier. This means their starting brute cell heals 10 brute damage per shot, their starting burn cell heals 10 burn damage per shot, and their starting stabilizer cell heals 30 oxy per shot. They only get 4 shots, so outside of what amounts to 'ranged first-aid', there isn't a whole lot they could do without upgrades. Once used, the cells need to be recharged before they can be re-used. Better cells, and other variants of cell, can be obtained from Science.

The CMO gets the same cells, but the second tier, so it's a bit better but that's kind of the norm for Head Equipment.

Heroman3003:
I do think that CMO gun holding 6 cells and regular medigun holding 3 cells is way too much. I think CMO holding 3 and regular holding 1 would be much better for basic game balance, and make them being available roundstart not as much of a problem.

In addition to that, your code never gave the CMO advanced medibeam the ability to actually hold advanced cartridge, its just inheireted from regular medigun instead.

mistyLuminescence:
Dropping the regular to 1 cell would make the use of a magazine a lot clunkier. Keep in mind that they don't actually heal much per shot, so the intent is to magdump as much as you can (or use single shots to top off minor booboos) - a full burn cell heals the same as *5u* dermaline- so with the entire mag filled with burn cells, that's... 15u dermaline. Or 30u if you're the CMO, but the CMO gets cool stuff because they're the CMO and heads of staff get cool stuff.

Good spot on the advanced cartridge though, will fix. The HoS should start with a "nsfw_pack_hos" in their locker, which uses /nsfw/combat now.

Novacat:
I would say a basic ML3M kit is good enough. CMO already gets a combat defib and a hypospray, they really do not need more overpowered stuff.

Leave the CMO cartridges able to be slotted into regular ML3Ms, but make the CMO cartridges R&D only. My original statement about the CMO ML3M only getting the same stuff as the Paramedic ML3Ms still applies.

PR #5426, opened 2019-07-04
Fixes some telecomm doors missing access

Heroman3003:
Fixes the fact that not all doors on telecomms had simultaneous requirement for both Telecommunication and General Maintenance accesses. Both accesses are now required for all doors as intended.


SpadesNeil:
Conflict has been caused by a prior map PR.

Heroman3003:
Map conflicts are a pain so I will just make a new branch and remake this PR.

PR #5429, opened 2019-07-05
Ease of Eating Changes

Novacat:
On request from Duke

- Microwave cooking times cut to 50% of normal times
- You now need 1/5 of the amount of clonexcadone for meat
- Cargo pizzas start out cooked
- Increases Price of Cargo pizzas to 50 points to compensate
- Vermicetol now uses Bicaridine instead of Kelotane and Dermaline
- Marks some unmarked vorestation edits

### DIFF
--- a/code/datums/supplypacks/hospitality_vr.dm
+++ b/code/datums/supplypacks/hospitality_vr.dm
@@ -79,6 +79,3 @@
 	cost = 50
 	containertype = /obj/structure/closet/crate/freezer
 	containername = "Chinese takeout crate"
-
-/datum/supply_pack/randomised/hospitality/pizza
-	cost = 25
--- a/code/modules/food/food/snacks_vr.dm
+++ b/code/modules/food/food/snacks_vr.dm
@@ -597,16 +597,16 @@
 	filled = TRUE
 
 /obj/item/pizzabox/margherita/Initialize()
-	pizza = new /obj/item/weapon/reagent_containers/food/snacks/pizza/margfrozen(src)
+	pizza = new /obj/item/weapon/reagent_containers/food/snacks/sliceable/pizza/margcargo(src)
 
 /obj/item/pizzabox/vegetable/Initialize()
-	pizza = new /obj/item/weapon/reagent_containers/food/snacks/pizza/vegfrozen(src)
+	pizza = new /obj/item/weapon/reagent_containers/food/snacks/sliceable/pizza/vegcargo(src)
 
 /obj/item/pizzabox/mushroom/Initialize()
-	pizza = new /obj/item/weapon/reagent_containers/food/snacks/pizza/mushfrozen(src)
+	pizza = new /obj/item/weapon/reagent_containers/food/snacks/sliceable/pizza/mushcargo(src)
 
 /obj/item/pizzabox/meat/Initialize()
-	pizza = new /obj/item/weapon/reagent_containers/food/snacks/pizza/meatfrozen(src)
+	pizza = new /obj/item/weapon/reagent_containers/food/snacks/sliceable/pizza/meatcargo(src)
 
 /obj/item/weapon/reagent_containers/food/snacks/liquidprotein
 	name = "\improper LiquidProtein Ration"
--- a/code/modules/reagents/Chemistry-Recipes.dm
+++ b/code/modules/reagents/Chemistry-Recipes.dm
@@ -1243,7 +1243,7 @@
 	name = "Syntiflesh"
 	id = "syntiflesh"
 	result = null
-	required_reagents = list("blood" = 5, "clonexadone" = 5)
+	required_reagents = list("blood" = 5, "clonexadone" = 1) //VOREStation Edit
 	result_amount = 1
 
 /datum/chemical_reaction/food/syntiflesh/on_reaction(var/datum/reagents/holder, var/created_volume)
--- a/code/modules/reagents/Chemistry-Recipes.dm
+++ b/code/modules/reagents/Chemistry-Recipes.dm
@@ -737,9 +737,12 @@
 		var/mob/living/L = holder.my_atom
 		if(L.stat != DEAD)
 			e.amount *= 0.5
+	//VOREStation Add Start
 	else
 		holder.clear_reagents() //No more powergaming by creating a tiny amount of this
+	//VORESTation Add End
 	e.start()
+	//holder.clear_reagents() //VOREStation Removal
 	return
 
 /datum/chemical_reaction/flash_powder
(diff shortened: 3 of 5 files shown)


Heroman3003:
One change I would suggest is that with cargo pizza coming unfrozen again, point cost should probably go back to its pre-nerf state of 50.

SpadesNeil:
As long as cargo pizzas are of lesser quality and nutrition as homemade pizzas, it's fine by me.

SpadesNeil:
I'm going to say no on the grounds of not allowing cargo pizzas to be an easy and superior alternative (again) to ordering food from the chef.

The chef wants to have fun too.

Catblaster10:
The problem with this, which is why im suggesting the change is that oftentimes the chef prefers making an arrangement look "Good" over making a lot of food. I wouldn't particularly care but just this past week I was asked to write an OOC announcement to tell people to stop gorging themselves on food that the chef had set out, wiping out their works in mere moments.

If the problem is solely stopping the chef in his tracks, then vending machines are a problem too. The chef has the benefit of being in the most populated area of the station and thus is almost always able to get attention brought to them. 

But yes, I do agree that if it's reverted it should go back to 50 (or if it would be satisfactory, maybe like 70. Its the best of the food crates by far).

Catblaster10:
There's a social element thats a benefit that the cook has that no one else has. Also I highly doubt the chefs that make the fancy arrangements would mind the alternative if it meant they could bitchslap someone eating all their food without permission.


If we really need to, we can just tell cargo "Hey, ask the chef if they're around, if they are then tell the person ordering food to buzz off".

Heroman3003:
Yeah, my suggestion of 50 was more of 'bare minimum', because ideally it should be higher as the original nerf had a reasoning behind it, and rather than undoing it entirely, touching on other aspect (cost rather than usability of product) could work.

Wickedtemp:
Yeah. We've had pizza crates for years and they weren't really a problem. I'd have Tempest do chef-stuff while off-duty for fun and I've never had issues with people wanting to interact with hir and request food, despite Cargo being active and cargo-pizza being available. 

Some chef characters have literally attracted a full counter of people and cargo pizza didn't even enter the equation, despite being readily available. It...really isn't an issue and there's no real point in making the product worse. Mechanically, every single vendor that exists is an easier and superior alternative to ordering food from a chef. But, most people don't go to the chef just to get their hunger meter refilled. They do so, usually, to have some level of IC interaction and socialize. I'd be fine with cargo pizza going back to how they were, with a higher cost.

Novacat:
Please note that it is not a full revert of the pizza nerf. Cargo pizzas still have the 'kinda tastes bad' description and less nutriment than normal pizzas. The only change is that they start off cooked now.

With that in mind, raising point cost to 50 points.

Screemonster:
> Yeah. We've had pizza crates for years and they weren't really a problem.

yeah nah, before the port to polaris we upped the point cost to 150 because we got sick of people flooding the bar with pizza despite chefs existing, so it's been a problem since we were on box

Catblaster10:
If we want to take it back to 150 im fine with that. I just dont like it being frozen in the first place. If the station is busy and the dorms are taken and there's no chef, its often quite hard to actually warm them up.

code/modules/reagents/Chemistry-Recipes.dm
```diff
@@ -1243,7 +1243,7 @@
 	name = "Syntiflesh"
 	id = "syntiflesh"
 	result = null
-	required_reagents = list("blood" = 5, "clonexadone" = 5)
+	required_reagents = list("blood" = 5, "clonexadone" = 1) //VOREStation Edit
```
SpadesNeil: This entire datum field can be put into a Chemistry-Recipes_vr.dm file.

Or, at least, what you are changing.

Remove the vorestation edit line and do this instead in the new file:
```
/datum/chemical_reaction/food/syntiflesh
     required_reagents = list("blood" = 5, "clonexadone" = 1)```

**Comment by SpadesNeil** [comment]
>If the problem is solely stopping the chef in his tracks, then vending machines are a problem too. The chef has the benefit of being in the most populated area of the station and thus is almost always able to get attention brought to them.

The idea that chefs are at odd because vending machines exist is not a fair argument, though.

Vending machines provide crap quality food that barely fill you up at all.

For a while, the chef had no point in existing because everyone just ordered pizza. Now, the chefs more often bother to do anything because people are actually coming to eat the food, giving them some fulfillment in doing their job, and more importantly, making people interact with the chef or with other diners which could lead to scenes.

This PR will not be accepted with the change to pizzas as they are.

File: code/modules/food/food/snacks_vr.dm [hunk 2]
```diff
@@ -597,16 +597,16 @@
...
-	pizza = new /obj/item/weapon/reagent_containers/food/snacks/pizza/vegfrozen(src)
+	pizza = new /obj/item/weapon/reagent_containers/food/snacks/sliceable/pizza/vegcargo(src)
 
 /obj/item/pizzabox/mushroom/Initialize()
-	pizza = new /obj/item/weapon/reagent_containers/food/snacks/pizza/mushfrozen(src)
+	pizza = new /obj/item/weapon/reagent_containers/food/snacks/sliceable/pizza/mushcargo(src)
 
 /obj/item/pizzabox/meat/Initialize()
-	pizza = new /obj/item/weapon/reagent_containers/food/snacks/pizza/meatfrozen(src)
+	pizza = new /obj/item/weapon/reagent_containers/food/snacks/sliceable/pizza/meatcargo(src)
```
SpadesNeil: The changes to pizza are being vetoed. This PR will not be accepted unless they are reverted.

code/datums/supplypacks/hospitality_vr.dm
```diff
@@ -79,6 +79,3 @@
 	cost = 50
 	containertype = /obj/structure/closet/crate/freezer
 	containername = "Chinese takeout crate"
-
-/datum/supply_pack/randomised/hospitality/pizza
```
SpadesNeil: The changes to pizza are being vetoed. This PR will not be accepted unless they are reverted.

SpadesNeil:
That all said, I think we can do with a few more microwaves around the station so people can actually make the pizza.

In fact, microwave meals in general would be nice.

Heroman3003:
We already have other two food crates that deliever easy and ready food. If Pizza crate is the problem compared to them, then just increase its cost until its on par with those in terms of points to food. Because heating pizza up is a minor hassle in the end that really achieves nothing in fixing actual issues behind the Pizza crate, and increasing microwave availability will certainly not help the problem as it will just decrease the hassle and thats it. Its still same amount of pizza you get as if it came cooked. Also, heated version of cargo pizza is still objectively inferior to chef's pizza in any way.

SpadesNeil:
>Also, increasing microwave availability may just backfire as combined with public garden people may just start cooking INSTEAD of chef, overshadowing them that way. We've already had chefs complaining that people just enter the kitchen to start cooking their stuff when chef takes two steps away from it.

Hm, that is a good counter point.

Fuck it, if the cargo pizza is still worse than actual pizza, so be it.

code/datums/supplypacks/hospitality_vr.dm
SpadesNeil: Statement redacted as Heroman made a good point.

TheDavestDave:
I think the replaces chef thing could be made better if the crate had a single box of pizza isntead of 5 in it.

PR #5463, opened 2019-07-12
Buffs Air Vents to Baystation Standard

Novacat:
...this should probably be ported to Polaris but too lazy right now.

Should alleviate problems with both Phoronlocks and all airlocks, including the expedition shuttle airlock.


Jakksergal:
This is only going to exacerbate power problems across the station when there is no source of power. i.e. APCs will drain much faster.

I'm willing to bet that BS12 increased the power of them because vents are few and far between on their map. Though that's only a guess.

MrSign:
Faster cycling is good.

PR #5472, opened 2019-07-15
Implements Firlock scrubbers

### LINKED ISSUE #5003: Surface Eva Atmos alarm screams whenever anyone enters
The atmosphere alarm for the public surface Eva goes off whenever anyone comes back in, due to low temperature 

No alarms 

#### What actually happened
Alarms! 

#### Steps to reproduce
- (Step 1)go outside using whatever door you want 
- (Step 2)come back inside via public surface eva
- (and so on)

#### Code Revision
 
Heroman3003:
Merged in separate PR, this maps them in. Yes, it includes mining airlock, which is an airlock again. With testing done with timers, the cycling time has been reduced by more than a half of time it takes on all three airlocks. In addition to that, temperature during cycling will now be risen to match the station temperature and amount of phoron traces will be even less than before.

As a side note: fixes #5003

### DIFF
--- a/maps/tether/tether-01-surface1.dmm
+++ b/maps/tether/tether-01-surface1.dmm
@@ -73,32 +73,37 @@
 	},
 /turf/simulated/floor/tiled/steel_dirty/virgo3b,
 /area/tether/surfacebase/mining_main/external)
+"aao" = (
+/turf/simulated/wall,
+/area/tether/surfacebase/mining_main/airlock)
 "aap" = (
 /obj/structure/cable/green{
 	d1 = 1;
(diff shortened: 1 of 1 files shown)


Screemonster:
tbh I still prefer the tiny fans (or forcefields if someone sprites them) while keeping the big scrubbers

Novacat:
My personal preference is that the tiny fans/forcefields should only really be used in situations where the structure involved is too small/compact for even the smallest of normal airlocks/firelocks would work.

This should fix the phoronlocks being completely horrible to use, which should improve QOL throughout the tether.

PR #5497, opened 2019-07-23
A stray attempt at fixing the lacking manifest entries

Heroman3003:
Manifest entries (or any records for that matter) seem to not generate properly if player lags out during specific ticks of the round start. This could possibly fix that, making character creation and addition to manifest a single action, rather than two separated by two more things needing doing, giving too much room for difference between list of players at start of initialization and at the 'manifest-filling' stage.

Tested as much as possible alone, seemed to create roundstart manifest entries just fine? Still a bit unsure about other potential aspects, admittedly.


Novacat:
If nothing else, if the fix fails to work it can be reverted.

PR #5515, opened 2019-07-27
Fixes picking up ore while drilling with mechs

Heroman3003:
Got broke at some point it seems. Didn't pick up any ore half the time, and never any ore on the sides of the dug tile. This fixes that and removes redundant recopied code in some areas


Novacat:
...any reason this should not get ported to Polaris?

Heroman3003:
@Novacat sent the Polaris-part  of PR to Polaris, but still a line of change to vorestation-unique code there.

PR #5528, opened 2019-07-27
Replaces "Terminated" with "Dismissed" as default rank removal

Heroman3003:
Primarily because with our SOP and continuity, onstation rank removal is never actually a 'Termination', but simply a demotion for the rest of the shift. Real terminations happen much more rarely and only with admins stepping in. I think it makes sense to use a more fitting term for the 'full rank removal' that conveys better what exactly we're doing.

EDIT: On Ace's request, changed the Demoted to Dismissed. Still fits the situation well, and doesn't imply full firing nearly as much so is good.

### DIFF
--- a/code/game/machinery/computer/card.dm
+++ b/code/game/machinery/computer/card.dm
@@ -279,7 +279,7 @@
 
 		if ("terminate")
 			if (is_authenticated())
-				modify.assignment = "Demoted"
+				modify.assignment = "Demoted"	//VOREStation Edit
 				modify.access = list()
 
 				callHook("terminate_employee", list(modify))
--- a/code/game/machinery/computer/card.dm
+++ b/code/game/machinery/computer/card.dm
@@ -279,7 +279,7 @@
 
 		if ("terminate")
 			if (is_authenticated())
-				modify.assignment = "Terminated"
+				modify.assignment = "Demoted"
 				modify.access = list()
 
 				callHook("terminate_employee", list(modify))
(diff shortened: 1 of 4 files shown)


Wickedtemp:
Yeh

Wickedtemp:
...hows this a staff discussion thing tho

Novacat:
Mostly making sure it is all right with staff. I agree with it personally.

SpadesNeil:
I don't really see the point of this. The option to set someone to "Terminated" in the first place makes no sense to me.

You just demote them to assistant. This is redundant as hell.

Heroman3003:
Its basically to show that they've been stripped of any level of position, including even being 'useless but still officially working crewmember' level of position, that Intern (assistants gone for long time) has. Most 'full' demotions I saw were handled by forceful moving off-duty or said Terminate button. Because technically, neither Visitor nor Intern really carry the intent behind the rank stripping that well.

Heroman3003:
Addendum to that, but that is mostly to make sure that correct terminology is used, since the whole 'terminated' thing has been there since ID console's creation, and we just use different term since our HoPs don't truly 'fire' people, just remove from duty for the shift.

Heroman3003:
I've added Vorestation Edit comments to tmlp files. As for the Polaris bit, their rules on the situation are different, and there are situation where Site Manager can handle the 'Firing' procedures, plus Polaris' much more muddled rules on continuity and canonicity. This change is just something for Virgo specifically.

SpadesNeil:
"Demoted" as a rank on the card makes no sense. I think Terminated was a much better word for the context.

But Terminated is such a harsh word, which implies permanent removal from the company, which is why this PR exists. That is not desired.

How about "Dismissed"?

Heroman3003:
Alright. That term is better fitting anyway, since demotion applies to any 'rank reduction'.

PR #5532, opened 2019-07-28
Adds pipe caps to all construction areas

Heroman3003:
Now that sprite for them is fixed, no reason not to.


chaoko99:
I can think of a solid reason actually: It wastes people's time unwrenching redundant structures.

Heroman3003:
The only reason I originally mapped them out was because sprites were broken mess. They were there for a while, just looked like crap. Now that sprites are fixed, bringing them back in seems like a perfectly reasonable thing.

PR #5537, opened 2019-07-30
Fixes synth legs being broken causing feedback loop

Heroman3003:
Including sparking regardless of resting and never being able to get up. Possibly intentional originally, but its stupid for us, especially with phoronlocks being a thing.

Now resting now properly prevents any sparks from generating because of broken synth legs and will make usage of airlocks safe assuming you rest while the cycling process is happening (which you probably should regardless if you are being rescued by someone else)

### DIFF
--- a/code/modules/mob/living/carbon/human/human_organs.dm
+++ b/code/modules/mob/living/carbon/human/human_organs.dm
@@ -86,7 +86,7 @@
 		var/obj/item/organ/external/E = organs_by_name[limb_tag]
 		if(!E || !E.is_usable())
 			stance_damage += 2 // let it fail even if just foot&leg
-		else if (E.is_malfunctioning())
+		else if (E.is_malfunctioning() && !(lying || resting))	//VOREStation Edit: no sparking/falling for synths with legs manually off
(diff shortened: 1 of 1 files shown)


SpadesNeil:
Sounds like a Polaris problem.

Heroman3003:
The behaviour is intentional, since analogous could occur for different reasons other than synth legs. The main reason I pr'd it just here is because unlike anywhere else, sparking aspect of  it being unstoppable as a result is actually a problem, not just visual/audible effect.

Heroman3003:
I've sent a version of this in Polaris, but going to leave it up in case they deny it or don't make a decision on it before we update.

SpadesNeil:
Can you link the Polaris PR for future reference? Or link the Polaris version here?

Heroman3003:
https://github.com/PolarisSS13/Polaris/pull/6377

Heroman3003:
@SpadesNeil As a note, I do think that this PR should make it into release ahead of time, if Polaris is undecided when release merge time is upon us, because this thing should have really been found out and fixed much sooner with how actually critical it is on our map.

PR #5561, opened 2019-08-05
Suit Sensors Selectable

TheFurryFeline:
Allows selection of suit sensor settings in VORE tab on Character Setup. Work credited to https://github.com/PolarisSS13/Polaris/pull/4354 Adjustments made as needed to account for outdated code as well as making use of preferences_vr at last.


SpadesNeil:
I can see why Polaris doesn't want it, but this would be very nice for us.

PR #5567, opened 2019-08-07
Adds Prototype Hypospray

Novacat:
- Can be researched and manufactured by Science
- Identical to a mk1 hypospray, except does not instantly inject reagents at friendly intents

Not terribly useful right now under the current meta, but probably will be useful down the road if/when syringes become disposable again.

### DIFF
--- a/code/modules/reagents/reagent_containers/hypospray.dm
+++ b/code/modules/reagents/reagent_containers/hypospray.dm
@@ -44,7 +44,14 @@
 			to_chat(user, "<span class='danger'>You cannot inject a robotic limb.</span>")
 			return
 
-		if(!H.stat)
+		//VOREStation Add Start - Adds Prototype Hypo functionality
+		if(H != user && prototype)
+			to_chat(user, "<span class='notice'>You begin injecting [H] with \the [src].</span>")
+			to_chat(H, "<span class='danger'> [user] is trying to inject you with \the [src]!</span>")
+			if(!do_after(user, 30, H))
+				return
+		//VOREstation Add End
+		else if(!H.stat && !prototype) //VOREStation Edit
 			if(H != user)
 				if(H.a_intent != I_HELP)
 					to_chat(user, "<span class='notice'>[H] is resisting your attempt to inject them with \the [src].</span>")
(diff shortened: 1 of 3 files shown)


code/modules/reagents/reagent_containers/hypospray.dm
```diff
@@ -18,6 +18,7 @@
 	var/filled = 0
 	var/list/filled_reagents = list()
 	var/hyposound	// What sound do we play on use?
+	var/prototype = 0 //VOREStation Add - Prototype Hypospray
```
Heroman3003: This should prolly be moved to hypospray_vr, since file already exists anyway

code/modules/reagents/reagent_containers/hypospray_vr.dm
```diff
@@ -22,3 +22,11 @@
 	..()
 	for (var/i = 1 to 7)
 		new /obj/item/weapon/reagent_containers/hypospray/autoinjector/miner(src)
+
+/obj/item/weapon/reagent_containers/hypospray
+	var/prototype = 0
+
+/obj/item/weapon/reagent_containers/hypospray/science
+	name = "prototype hypospray"
+	desc = "This reproduction hypospray is nearly a perfect replica of the early model DeForest hyposprays, sharing many of the same features. However, there are additional safety measures installed to prevent unwanted injections."
```
SpadesNeil: Hyposprays don't inject, they--wait.

code/modules/research/designs_vr.dm
```diff
@@ -117,6 +117,15 @@
 	build_path = /obj/item/device/nifrepairer
 	sort_string = "HABBE" //Changed String from HABBD to HABBE
 
+/datum/design/item/medical/protohypospray
+	name = "prototype hypospray"
+	desc = "This prototype hypospray is a sterile, air-needle autoinjector for rapid administration of drugs to patients."
```
SpadesNeil: Is that seriously how they work?

SpadesNeil:
I see no controversy with a hypospray that has a safety feature.

code/modules/research/designs_vr.dm
Novacat: Yup!

PR #5578, opened 2019-08-10
Adds an admin verb for spawning players as mobs with their vore prefs intact

BlackMajor:
@Screemonster asked me to do this.
So I did.
Dab.

Just adds a simple admin verb, "Spawn Character As Mob", which lets you pick a client, the mob path to spawn, if you want them to be named after the client's currently selected character in the character setup, and if you want to copy their vore prefs/organs onto the mob.

Cause y'know, I guarantee someone out there is gonna want to have their verbosely written vore texts and interactions on their mob spawn.


BlackMajor:
Poke my dunkass if I'm retarded and I forgot something.

code/modules/admin/admin_verbs.dm
```diff
@@ -148,6 +148,7 @@ var/list/admin_verbs_spawn = list(
 	/datum/admins/proc/spawn_plant,
 	/datum/admins/proc/spawn_atom,		//allows us to spawn instances,
 	/client/proc/respawn_character,
+	/client/proc/spawn_character_mob,
```
Novacat: Needs //VOREStation Edit,

The comma at the end is important to stop travis from freaking out.

code/modules/admin/verbs/randomverbs.dm
```diff
@@ -513,6 +513,81 @@ Traitors and the like can also be revived with the previous role mostly intact.
 
 	return new_character
 
+/client/proc/spawn_character_mob()
```
Novacat: This should probably be exported to randomverbs_vr.dm

Heroman3003:
Isn't that what Reload Slot button for on Vore Panel? If you get spawned in as a mob, you just use the slot you've saved your bellies to and reload that slot's vore panel and its all immedeately there. As for OOC notes, I'm not sure those even work with simplemobs? I heard that someone tried manually pasting them back during shadekin times and it didn't work.

code/modules/admin/verbs/randomverbs.dm
```diff
@@ -513,6 +513,81 @@ Traitors and the like can also be revived with the previous role mostly intact.
 
 	return new_character
 
+/client/proc/spawn_character_mob()
+	set category = "Special Verbs"
+	set name = "Spawn Character As Mob"
+	set desc = "Spawn a specified ckey as a chosen mob."
+	if(!holder)
+		src << "Only administrators may use this command."
```
Heroman3003: Its better to use to_chats instead of << for sending messages, both here and in few cases below

code/modules/admin/verbs/randomverbs.dm
BlackMajor: Oh right you fella's still have that whole downstream thing going on.
Yeah aight.

code/modules/admin/verbs/randomverbs.dm
BlackMajor: Not really needed but I'll do it just for you, precious.

BlackMajor:
Hot damn there's a button in the vore panel for just loading your character slot's vore organs howcome I never noticed this. Kinda defeats the point of having the option to spawn with vore organs but y'know whatev guess I'll leave it unless someone asks me to remove it.

PR #5581, opened 2019-08-12
Allows you to sit on someone's head

TheFurryFeline:
Ports ChompStation facesitting, pretty much. :>

Edited to not have the blindness and thus make emotes invisible.


chaoko99:
Furthermore, I believe we have a policy on outright sexual mechanic

code/modules/mob/mob_grab_vr.dm
```diff
@@ -0,0 +1,14 @@
+//TFF 12/8/19 - Port ChompStation facesitting
+/obj/item/weapon/grab/handle_eye_mouth_covering(mob/living/carbon/target, mob/user, var/target_zone)
+	var/announce = (target_zone != last_hit_zone) //only display messages when switching between different target zones
+	last_hit_zone = target_zone
+
+	switch(target_zone)
+		if(BP_HEAD)
+			if(force_down)
+				if(announce)
+					assailant.visible_message("<span class='warning'>[assailant] moves their ass to [target]'s head, sitting down on them, making them unable to see anything else than [assailant]'s butt!</span>")
```
chaoko99: This is clumsily worded to the extreme.

You can literally just change this to `"sits on [target]'s head."`

chaoko99: Also a massive run-on sentence.

TheFurryFeline: Fixing, and it's an old snippet I just copy-pasta'd without thinking.

Novacat:
Generally, VOREStation policy has been to not add any overtly sexual mechanics. But, I am going to allow the other staff to have a say in this.

SpadesNeil:
There is a policy against sexual mechanics. Is sitting on someone's face sexual? Probably. But it is also very silly and amusing, so I'm giving this a green light for the hilarity that may ensue.

code/modules/mob/mob_grab_vr.dm
SpadesNeil: This description does not to be as long-winded. A simple `[assailant] sits on [target]'s head` would be better.

TheFurryFeline: Already have that ready for my next commit but I've been working on trying to change this code so that the snippet in mob_grab.dm works as well as this one.

Vorrarkul:
I have to express opposition to this for one reason: it makes 'me' emotes invisible. I personally find this an unacceptable tradeoff to something that can be done just as easily in 'me' emotes.

mistyLuminescence:
The policy against sexual mechanics is also partly because of GitHub's own policy regarding sexual content (specifically, forbidding it). There's plausible deniability with the rest of the server's antics, since it could be argued that vore doesn't _technically_ count as sexual content, but that line is a lot harder to skirt with facesitting. At the risk of getting all slippery-slopey, where does this end? Do we draw the line here, or do we add anilingus for the recipient? How about deepthroating?

Heroman3003:
Muting and blinding the target renders this whole thing pretty useless for its target purpose, as well as being rather overpowered when it comes to grab moves in general, IMO

Wickedtemp:
I don't really care either way, as long as:
- It doesn't conflict with other grab moves
- Is treated as a prefs-thing which I feel is obviously the case but confirming it never hurts
- It doesn't blind or mute anyone, which looks like has been fixed so this one is covered as well

TheFurryFeline:
Yeah, I removed blindness and mute stuff and as usual, we'll leave the pref
breaking occasions with admins to handle. As for the other grabs, tested it
and all three work fine, though you might have to drop grab and start again
with the full head as the target to get the third message.

On Aug 12, 2019 19:17, "Wickedtemp" <notifications@github.com> wrote:

I don't really care either way, as long as:

   - It doesn't conflict with other grab moves
   - Is treated as a prefs-thing which I feel is obviously the case but
   confirming it never hurts
   - It doesn't blind or mute anyone, which looks like has been fixed so
   this one is covered as well

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/VOREStation/VOREStation/pull/5581?email_source=notifications&email_token=AJGMTY36FZKYOA5NSRWCPWLQEHVPBA5CNFSM4ILBYIOKYY3PNVWWK3TUL52HS4DFVREXG43VMVBW63LNMVXHJKTDN5WW2ZLOORPWSZGOD4ECYAA#issuecomment-520629248>,
or mute the thread

Wickedtemp:
Whats the process behind doing it ingame?

SpadesNeil:
>The policy against sexual mechanics is also partly because of GitHub's own policy regarding sexual content (specifically, forbidding it). There's plausible deniability with the rest of the server's antics, since it could be argued that vore doesn't technically count as sexual content, but that line is a lot harder to skirt with facesitting. At the risk of getting all slippery-slopey, where does this end? Do we draw the line here, or do we add anilingus for the recipient? How about deepthroating?

You're sitting on someone's face, not humping it and moaning or doing other weird shit. Your slippery slope logic does not work here.

TheFurryFeline:
> Whats the process behind doing it ingame?

Normal grab +reinforce, but you want to force the guy to the ground before switching to head target and reinforcing the grip.

SpadesNeil:
All I care about right now is ensuring this does not bug out anything. The idea of the PR is now fine.

PR #5587, opened 2019-08-12
Disables bot patrolling on the Tether

Heroman3003:
It was broken and not working to begin with, and also a suspected cause of lag as bots infinitely tried to find a patrol route and couldn't. Either way, we don't have beacon system so it shouldn't be there regardless.

A followup to #5585 and will not compile until after its merged

### DIFF
--- a/maps/tether/tether_defines.dm
+++ b/maps/tether/tether_defines.dm
@@ -116,6 +116,8 @@
 							NETWORK_ALARM_FIRE
 							)
 
+	bot_patrolling = FALSE
+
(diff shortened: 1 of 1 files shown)


chaoko99:
For the record it isn't broken, you're just kind of required to map in beacons, which we have not done

Heroman3003:
Its broken on Tether as it is, and even if beacons were to be mapped in, patrolling only works within single z-level which is pretty useless on 6-floor Tether. Either way, if someone wants to rework and/or reactivate it, its a single var change. No need to have feature that does nothing and has potential to cause lag otherwise.

PR #5607, opened 2019-08-14
Fixes customsay not working for non-admins

### LINKED ISSUE #5603: [Bug. I think] "Customize-Say-Verbs" verb doesn't work
There's a new verb added on the update called "Customize-Say-Verbs" that, just like its description specifies, "It's used to customize the text which appears when you type- eg. 'says', 'asks', 'exclaims'." 

The problem is that the verb actually does nothing. Nothing happens when you select the verb on the IC tab in-game, nor when you use it on the chatbox.

Let you change the text that appears when you use the "say" verb, or atleast open up a new window/prompt in-game with further instructions.

#### What actually happened
Nothing.

#### Steps to reproduce
Either:
- Select the verb on the IC tab.
- Type in the verb on the chatbox. (Neither the commands "Customize-Say-Verbs", "Customize-Say-Verbs example", or "Customize-Say-Verbs 'example'" do anything at first sight)

#### Code Revision
release - 
0015a40b20070fa8cc03fd2c80ca7655174b50da

Heroman3003:
Fixes #5603

Also moves it out to its own file properly

### DIFF
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -1363,30 +1363,3 @@ default behaviour is:
 		BRAIN:<a href='?_src_=vars;mobToDamage=\ref[src];adjustDamage=brain'>[getBrainLoss()]</a>
 		</font>
 		"}
-//VOREStation edit, allows for custom say verbs
-/mob/living/verb/customsay()
-	set category = "IC"
-	set name = "Customise Say Verbs"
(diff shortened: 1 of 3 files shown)


mistyLuminescence:
fuuuuuuuuuck this didn't show up in testing at all

Heroman3003:
That'd make sense, considering testing is usually done with Host priveliges anyway.

PR #5614, opened 2019-08-15
Un-nerfs flaps

Mewchild:
Takes flaps back to pre-nerf passability
image


Novacat:
I am not really sure why this is needed. The new phoronlocks cycle much faster now, faster to the point where, in a brief test, I was able to cycle from outside to inside faster than going through the flaps.

Screemonster:
The airlock was made faster in a separate PR, so there's no longer any need/benefit to using the flaps. Which is what should have been done in the first place.

PR #5634, opened 2019-08-17
Restricts station use of manhack grenades and seismic charges

Heroman3003:
Manhacks can be emagged or unlocked with armory-level ID to still be used on station, similarly to phasers. Seismic charges cannot, and can exclusively be used away from station.

Manhack grenade when attempted to use on station will just reset back to initial state so can still be used if taken to more appropriate area. Seismic charge will harmlessly implode and be permanently lost without any explosions.


Screemonster:
The charges are probably a good idea if only because we don't want to have to restrict _another_ role from day-0 accounts to stop griffins from getting explosives

Heroman3003:
For manhacks, its primarily because they attack literally anything they dont identify as crewmember. Like station animals, people with mask and no ID, etc. But its still a potentially useful thing in events so its unlockable

code/game/objects/items/weapons/explosives_vr.dm
```diff
@@ -0,0 +1,17 @@
+/obj/item/weapon/plastique/seismic/locked
+	desc = "Used to dig holes in specific areas without too much extra hole. Has extra mechanism that safely implodes the bomb if its used in close proximity to the facility."
```
TheFurryFeline: Minor typo: 'it's used in close proximity'

code/modules/catalogue/rewards/equipment_vendor.dm
```diff
@@ -36,7 +36,7 @@
 		new /datum/data/exploration_equipment("Jump Boots",					/obj/item/clothing/shoes/bhop,										250),
 		new /datum/data/exploration_equipment("Luxury Shelter Capsule",		/obj/item/device/survivalcapsule/luxury,							310),
 		new /datum/data/exploration_equipment("Defense Equipment - Smoke Bomb",/obj/item/weapon/grenade/smokebomb,								10),
-		new /datum/data/exploration_equipment("Defense Equipment - Razor Drone Deployer",/obj/item/weapon/grenade/spawnergrenade/manhacks/station,	100),
+		new /datum/data/exploration_equipment("Defense Equipment - Razor Drone Deployer",/obj/item/weapon/grenade/spawnergrenade/manhacks/station/locked,	100),
```
TheFurryFeline: //Vorestation edit

code/modules/mining/ore_redemption_machine/equipment_vendor.dm
```diff
@@ -66,14 +66,14 @@
 		new /datum/data/mining_equipment("Fine Excavation Kit - Chisels",/obj/item/weapon/storage/excavation,								500),
 		new /datum/data/mining_equipment("Fine Excavation Kit - Measuring Tape",/obj/item/device/measuring_tape,							125),
 		new /datum/data/mining_equipment("Fine Excavation Kit - Hand Pick",/obj/item/weapon/pickaxe/hand,									375),
-		new /datum/data/mining_equipment("Explosive Excavation Kit - Plastic Charge",/obj/item/weapon/plastique/seismic,					750),
+		new /datum/data/mining_equipment("Explosive Excavation Kit - Plastic Charge",/obj/item/weapon/plastique/seismic/locked,					750),
```
TheFurryFeline: Also here and for the drones.

code/modules/catalogue/rewards/equipment_vendor.dm
Heroman3003: There is big overarching edit comment for entirety for mining list already

PR #5641, opened 2019-08-18
Removes armor values from Dayleen's Jackboots

Mewchild:
Title


SpadesNeil:
Jeeze, those were quite some armor values, too. They must have been mistakenly modeled off of Deathsquad jackboots.

Or this was added during my hiatus when the Tether was added to begin with, because I would have vetoed this shit.

PR #5645, opened 2019-08-18
Species Whitelist Additions

TheFurryFeline:
Adds in whitelist names for Hollifex and A Random Alien per https://forum.vore-station.net/viewtopic.php?f=46&t=1568 and https://forum.vore-station.net/viewtopic.php?f=46&t=1569 respectively.


Screemonster:
shouldn't whitelist things be in the "all lowercase, no spaces" format? We'll see.

PR #5665, opened 2019-08-20
More emergency fixes

Heroman3003:
Fixes fishing never yielding any results in the indoor pond

Fixes fishing on deep tiles taking much longer than on shallow ones

Moves interior sensor on the expedition shuttle slightly


AaronSherwood117:
Thank you red meow

PR #5669, opened 2019-08-20
Adds Compact Cataloguers

Novacat:
Uses the tricorder sprites, it can be closed up when stowed away. Cannot scan anything while it is closed up.

Field Medic/Pilot compact cataloguers are shorter ranged and have worse stats. They can scan stationary, harmless objects fairly effectively but anything hostile will require an explorer's cataloguer.

The Pathfinder's cataloguer basically is a compact cataloguer with normal cataloguer's stats.


Screemonster:
what the hell happened to the diff on the dme

PR #5670, opened 2019-08-20
Adds Medical Tricorder

Novacat:
A reskinned health analyzer. Identical to normal health analyzers except can be packed up into an even smaller package. Currently only found in Field Medic lockers.


Wickedtemp:
Really small change so I kinda just have to ask if it really matters either way if this gets added or not, if there's really any benefit or really any loss if it's merged or closed.

Novacat:
It was never really meant to be anything meta-shifting, I considered making it an improved health analyzer that was clunkier to use (itemsize_normal when deployed, itemsize_small when packed up) but I was kind of worried that might have had balance issues between the improved health analyzer, if it was easier to acquire/available on round start.

Very-Soft:
It sounds like this is just a health analyzer that doesn't take up as much space. The health analyzer already doesn't take up much space. And being that the only listed place for this is for the exploration team, I can't help but think this is for optimizing some kind of meta kit or something.

Why does the exploration team need even smaller health analyzers?

Heroman3003:
To be honest, better use of those sprites, if you want to give them any in specific, could be giving SARs their own cataloguers.. Because really, analyzers dont need that stuff.

Novacat:
That does sound like a better idea.

Heroman3003:
To add to that, they could be BIGGER than normal cataloguers when deployed, and same when closed to make them not direct upgrade.

PR #5696, opened 2019-08-23
Restricts AI shells module- and z-level- wise

### LINKED ISSUE #5655: Let's talk about Shells (AI shells are OP and that's lame)
So! Having fiddled a bunch with AI shells, and having seen others fiddling with them, I have a few problems with them.

This is not to say I hate them or want them gone or anything. I think the shells are good for giving the AI more of a physical presence to interact with. It lets you set up vore stuff and all that, and that's real cool. I even think that in very specific circumstances where -no one- can/knows how to do a thing, the AI being able to step in and help could be nice.

However.

First: The AI's job was never meant to be able to do basically anything. It doesn't take much at all to get to a point with the shells to where you can just make more shells on your own. There's nothing to really stop you doing that if you're bored. Each AI shell has the full function of a borg. The AI can freely make these shells into any borg type, and then freely swap between them, effectively making them able to do basically anything outside of command all on their own.

Second: Funnily enough, inhabiting shells seems to disable the radio channels the shell you're operating wouldn't be able to hear. So, see: a mining shell would get the common channel and supply, but not the others. Now, of course I'm pretty sure you can still set the radios in the core to transmit those channels to you, and I think you might hear them then, but you can't get all the channels like you normally would have as the AI. Which means that any time one is inhabiting the shells, they're at least in part cutting themselves off from listening for requests to certain channels. So like, one of the AI's main functions is made unreliable by the use of the shells in their present incarnation. For a more actiony, antagonist-y server, this makes sense, you'd probably want to cover up some of the information if the AI is taking a more active, physical role.

Most of us like to play on our server to ERP. This can and does take a lot of time often times. The AI sitting in medihound shell, unable to hear the security or command channels as an example, it less than ideal. 

Third: They can operate in areas that are ordinarily meant to be off limits to the AI. There are certain spots around the station that the AI is deliberately blind to. Spots for people to hide while they do their crimes/unwilling vores, or even off station such as on exploration missions.

The long and short is, where once I thought the AI had far too little to do, and less than ideal means to interact with the crew personally, I now think that they are too overblown, and much more focused on the wrong thing. The AI is not meant to be a super do everything machine. We don't let the CD run around and do every job, much as some of them would like to. I think the AI's shells in their present form have been repeatedly demonstrated now to allow and encourage that.

In my opinion, if you want to play as a borg, you shouldn't be playing AI. And if you want to play as MULTIPLE borgs. Well, tough. We don't allow anyone else to operate like that for a variety of reasons, not all of which are strictly related to believably, but also because it's just super annoying to turn a corner to do your job, only to find that someone not in your department is already doing it.

Now, I'm honestly of the opinion that the AI really shouldn't be able to do much with their shells at all. I'd be a lot happier with them if they had 1 option and it was a much simpler kit. A fire extinguisher, a crowbar, a flash, maybe bandages or something. Let them pull people, and let them have access to the full vore stuff like they presently have. From there, make sure they have access to all the channels they normally do, so that way if they're ignoring an issue to ERP it's at least a conscious choice to have done so. 

I understand though that making a new custom borg for AI use only might be asking a little much though, so if such is unreasonable, I'd ask that AIs be restricted to the use of 1 shell. No more. That way at least if they want to run around and do every job they at least have to make a pitstop in robotics for every job change.

Heroman3003:
Now only available in Clerical, Service and Standard module variations.

AI will be booted out of the shell if it travels to z-level thats not station itself, its nearby facilities (surface mining, solars and Central) or Virgo 4 beach, and won't be able to access it until its back on available levels.

Resolves #5655 

- [x] Restrict available modules to only civilian low-impact ones
- [x] Implement inability to use shells in exploration areas
- [x] Add ability to hear communications of all channels when using a shell


ShudderEve:
May want to add the shuttle z-levels if you're going for the virgo-4 beach. That, and another few issues, but I'm not going to talk about #5655 here.

Heroman3003:
@ShudderEve Misc and Ships are technical z-levels that I didn't really mention because of technical nature, but they are among whitelisted z-levels already.

CRTXBacon:
> service

During the discussion in the issue I got the idea that we'd see nothing but the standard module, but having the service module makes me way more okay because I'd be lying if I said that one isn't fun to play as.

CRTXBacon:
Although one suggestion:
Still allow the red alert modules. These are already restricted to a specific scenario and in a situation where they're actually needed I see no reason to prevent the AI having one.

PR #5709, opened 2019-08-24
Mouse tweaks and (hopefully) bugfixes

Mithrandalf:
First time on this codebase, and I'm pretty new to baycode in general. Very very open to input for this.

Couple small tweaks to mice, pretty much just bringing their functionality closer to just about every other server and restoring some (presumably) intended functionality.

It seems like there was a refactor of mobs at some point in the codebase's history that left some features (like one-click scooping) nonfunctional. This is made under the assumption that those features are desired and intended.

**Changes are as follows:**

 _-Mice get scooped in the same way as any carbon mob; with a single green-intent click, though the click-drag still works too_

 _-Petting functionality moved to clicking with a mouse inhand, uses semi-snowflake code to prevent spam and make it function how it used to_

 **Making these generic seemed like a good idea, but I don't wanna try to reinvent the wheel by reworking how all mobs work off the bat, so it's just for mice

 _-Mice have improved dark vision, up to a level that it seems they were previously set at, so they can maint-ninja more effectively_

 _-Mice now move at roughly 1/2 the speed of people, instead of the painfully low speed they had before_

 _-Mice can understand speech. They can't be understood any more than they could before, but they're able to tell what people are saying. This one might be controversial, but as a semi-observer role being completely unable to listen to dialogue sucks - in addition its inconsistent as-is because you can still understand speech done as part of an emote. Considering the uneducated_mice config is set to off, I imagine this is the semi-intended situation anyway._

 **I think mouse language is broken in how it interacts with translators, this is smth I could fix later

**I've tested it fairly thoroughly, but as I've said my experience in bay is limited. I'm looking to learn, so if I fucked up I'd love to know how and what needs fixing.**


Mithrandalf:
Not sure what the travis error is about. I had some trouble with DM trying to prune the DME which might have caused it?

Novacat:
Travis is broken. Give them time to fix it.

Novacat:
Oh, looking over the files, you changed the simple_animal files where you need to change the simple_mob files.

Mithrandalf:
closing cos https://github.com/VOREStation/VOREStation/pull/5711 exists

PR #5710, opened 2019-08-25
glove attacks refactor

Screemonster:
Refactors unarmed attacks slightly. At the _moment_, the only thing affected is the holo-boxing gloves.
Adds the ability to assign an unarmed attack to gloves so you're not just limited to adjusting punch_force and punch_damtype.
Adds the ability for unarmed attacks to completely override the regular attack proc and just do _whatever_ if you want to.

Removes the snowflake check in human attack_hand() for specifically holographic boxing gloves and moves it to the attack override section of that glove's unarmed attack.

In practical terms, apart from the "added functionality" parts, this means that holographic boxing gloves no longer have a silly istype() check carried out every single time you click on someone regardless of intent.
Oh, and you have to use harm intent to punch people with boxing gloves now. No more helpful punching.

Also changes it so that if you're using gloves that have a sharp/edge attack, it no longer returns agony damage when you pull punches. There's only so much punch-pulling you can do with spiked knuckles.


Screemonster:
closed and reopened to stop travis raining on me

Screemonster:
Submitted to https://github.com/PolarisSS13/Polaris/pull/6428

PR #5712, opened 2019-08-25
Changes Ion Hivebots to not have EMP

Mewchild:
Changes the hivebots to use the ion pistol round, instead of EMP Ballistic Casing. No AoE on hit


SpadesNeil:
While I love seeing synth get fucked on away missions, it's probably best this isn't an AOE effect.

PR #5718, opened 2019-08-27
Adds morphs

Heroman3003:
Mostly based on tg ones. Creatures that can shapeshift their form into anything they see! Main difference with tg is that there is no ultra-antaggy 'swallow/break everything' aspect, and instead they work more like regular mobs for player control. Right now adminspawn only for admins to use for scene spawnins or event stuffs. Finally, a good player-controlled mimic thing.

They have ventcrawl, and dont hit too hard. Cant attack while morphed and move much slower. To transform into an object, shiftclick it when you are adjacent to it. To undo, shiftclick yourself.

A mini-demonstration gif: https://cdn.discordapp.com/attachments/187013248309002240/615761783960895488/morph.gif


SpadesNeil:
That looks terrifying. Can it eat people?

Heroman3003:
Only via vore panel, same as any other PC-controlled mob with enabled vore.

PR #5731, opened 2019-08-29
Buffs Mini Translocator

Novacat:
- Removes Z level restriction on mini translocator
- Decreases charge cost of mini translocator from 2400 to 800
- Mini Translocator starts with a device cell instead of a weapon cell
- Removes illegal tech level from Mini translocators
- Translocators can now be used to get between the Aerostat and V2 surface, and the V4 Beach and Caves. They still cannot transfer between planets.

The overall effect is that mini translocators still start off with one charge and one beacon, but can now be used across z levels. Mini translocators can also be equipped with a weapon cell that gives them three charges instead of one.

### DIFF
--- a/code/modules/vore/fluffstuff/custom_items_vr.dm
+++ b/code/modules/vore/fluffstuff/custom_items_vr.dm
@@ -1280,7 +1280,7 @@
 /obj/item/device/perfect_tele/update_icon()
 	if(!power_source)
 		icon_state = "[initial(icon_state)]_o"
-	else if(ready && power_source.check_charge(charge_cost))
+	else if(ready && (power_source.check_charge(charge_cost) || power_source.fully_charged()))
 		icon_state = "[initial(icon_state)]"
 	else
 		icon_state = "[initial(icon_state)]_w"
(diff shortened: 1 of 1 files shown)


Ketrai:
Wasn't it generally seen as a bad thing for translocators to teleport across z levels? Unless you just mean the tether levels. Don't wamt folks teleing in from distant planets.

Novacat:
It uses the same teleport checks as normal translocators. So just tether levels. But that reminds me that I need to fix being able to teleport around on V4/V2 as well. Caves/Beach and V2 aerostat/surface for examples.

Mewchild:
Will the director's translocator be buffed to have 5 charges as well, with a weapon cell? Currently, it has only 3

Novacat:
Standard translocators are, as of present, unchanged. 3 charges, 3 beacons, though I have considered upping them to 4 charges.

Ketrai:
mini translocators having 5 charges potentially feels  way, way too good compared to normal translocators. you rarely even need to use all three beacons. At most that's medbay, office, somewhere on the surface.

Novacat:
...Hm, it will be a bit difficult but ill try to keep the standard device cell on the mini translocator but also reduce the impact of having a weapon cell on it.

Novacat:
Thinking about it, going to leave the normal translocator at 3 charges and figure out a way to bump down the weapon-cell mini translocator down to 2 or 3 charges.

PR #5740, opened 2019-08-30
[WIP ,DNM] Add "Bridge Officer" Job under "Command"

Rykka-Stormheart:
Adds "Bridge Officer" under "Command" roles - this role is intended to be more "serious" than Command Secretary.

Adds two _vr.dm files, and adjusts the following files:
station_ids, command.dm, jobs.dm

Major bugs at present:
"Occupation" screen is formatted incorrectly, and Bridge Officer remains above "Colony Director".
[Image here](https://cdn.discordapp.com/attachments/187013248309002240/616982472529412151/03d050c24c59120a75af14fb3ba69781.png)
Selecting a preference for the job also changes Command Secretaries preference - if you choose "low/med/high/never" for CS, it changes BO's, and vice-versa.


code/game/jobs/job/captain_vr.dm
```diff
@@ -0,0 +1,21 @@
...
+	spawn_positions = 4
+	supervisors = "command staff"
+	selection_color = "#2F2F7F"
+	idtype = /obj/item/weapon/card/id/silver/secretary
+	minimal_player_age = 7
+	economic_modifier = 7
+	minimum_character_age = 20
+	ideal_character_age = 25
+
+	access = list(access_heads, access_keycard_auth, access_gateway, access_eva, access_teleporter)
```
Novacat: access_gateway should not be here. It is a heavily restricted area even most heads do not have access to.

Rykka-Stormheart: Oh, oops, my bad! I'll remove that, sorry lol

Heroman3003:
I really do not see the real difference from secretary other than getting teleporter and EVA access... And aparently having 4 slots in command category.

Novacat:
I have the same thoughts. It would probably be better to change Command Secretary than it would be to have a job that is only slightly different.

SpadesNeil:
This does not fill any niche not already filled by the command secretary which already doesn't fill very much of a niche.

I am not in support of this.

Screemonster:
I'd rather just give bridge secs an actual job if we're going to have non-heads with actual jobs on the bridge.

Rykka-Stormheart:
> I'd rather just give bridge secs an actual job if we're going to have non-heads with actual jobs on the bridge.

yeeee, could do, or just make "Bridge Officer" an alt-title of Command Secretary, like I originally thought of. But I decided to give this a shot, gauge reaction, then drop back to adding an alt-title. Would an alt-title of Bridge Officer be better received?

Rykka-Stormheart:
I'm closing this commit and instead falling back to just doing the alt-title. <3

Issue #5620, opened 2019-08-16
[BUG] "Virgo 2 Atmosphere" tiles block projectiles


HzDonut:
#### Brief description of the issue
It's right in the title, they prevent you from shooting past them. Make the aerostat real fucked up to fight in.

#### What you expected to happen
Projectiles would go over the "open space" of the atmosphere.

#### What actually happened
BLOCKED

#### Steps to reproduce
- Try to shoot over any hole in the aerostat, really.

#### Code Revision
- release - 
0015a40b20070fa8cc03fd2c80ca7655174b50da

TheDavestDave:
yeah I noticed this. The devs seemed to of not noticed it. https://github.com/VOREStation/VOREStation/issues/5403

Rykka-Stormheart:
It's a turf thing.

Per a discussion in #development the other day:
PrismaticGynoid#4492

> Theory tested and confirmed: The reason you can't shoot over the sky tiles is that the sky calls do_fall on the projectiles and they fall out of the sky
Which is hilarious and probably not how it's supposed to work

I'll see if I can fix it or drag in @Novacat to help. <3

Screemonster:
Sounds like do_fall() needs to just return false for projectiles

and flying things for that matter

Rykka-Stormheart:
> Sounds like do_fall() needs to just return false for projectiles
> 
> and flying things for that matter

already done, screedork <3

PR #5755, opened 2019-08-31
Minor Bugfixes/Tweaks - Expedition-focused.

Rykka-Stormheart:
Addresses V2 atmos tiles not allowing shots over them, and reduces charge/crank time for frontier phasers, as well as buffing them - you can reload while moving. QOL! <3

Specifically - Frontier Phasers/Carbines are 28 seconds to reload/recharge, Holdout Phasers are... 18-20. And they can be recharged on the move.

Before you worry about powergaming, they're still station-locked, and frankly, this buff is MASSIVELY needed with the damn ion hivebots everywhere still.


Rykka-Stormheart:
#5620 is resolved via this PR.

Screemonster:
I have no idea what the hell happened in the diff there

still think it'd be better to have a generic check for objects to decide whether they're subject to gravity or not rather than having the turf check specifically for projectiles and floating mobs

Rykka-Stormheart:
> I have no idea what the hell happened in the diff there
> 
> still think it'd be better to have a generic check for objects to decide whether they're subject to gravity or not rather than having the turf check specifically for projectiles and floating mobs

Of course, I totally agree on that. In the meantime, though, this is a good "bandaid" fix we can sneak into the September update. Hopefully. <3

Rykka-Stormheart:
I'm removing the expedition weapons tweaks for a separate PR.

PR #5757, opened 2019-08-31
Buff Frontier Phasers by reducing charge time to sub-30 seconds.

Rykka-Stormheart:
For holdout pistol phasers, this is sub 20 seconds.

And you can reload on the move. Is it overbuffed? Debate in the comments below, and don't forget to like, share, and HIT THAT SUBSCRIBE BUTTON.


code/modules/vore/fluffstuff/custom_guns_vr.dm
```diff
@@ -761,10 +761,10 @@
 	user.visible_message("<span class='notice'>[user] opens \the [src] and starts pumping the handle.</span>", \
 						"<span class='notice'>You open \the [src] and start pumping the handle.</span>")
 	while(recharging)
-		if(!do_after(user, 10, src))
+		if(!do_after(user, 10, src, ignore_movement = TRUE))
```
Novacat: This has been discussed in the past, and, IIRC, it was decided that being stationary while recharging the phaser was intentional.

code/modules/vore/fluffstuff/custom_guns_vr.dm
```diff
@@ -761,10 +761,10 @@
...
 			break
 		playsound(get_turf(src),'sound/items/change_drill.ogg',25,1)
-		if(power_supply.give(60) < 60)
+		if(power_supply.give(90) < 90)
```
Novacat: This should probably be exported to a var. Still, I would rather do small, adjustments. Make the main phaser receive 75 charge with each pump. Make the holdout receive 100 charge with each pump. The key concern here is that I do not want the phasers to become better than military weapons.

code/modules/vore/fluffstuff/custom_guns_vr.dm
Rykka-Stormheart: Regarding playtesting and such, it _sucks_ to just have to stand still for 30+ seconds. 

I'm still in favor of keeping the ignore_movement flag.

code/modules/vore/fluffstuff/custom_guns_vr.dm
Rykka-Stormheart: How exactly do I do that? There's no individual var under the holdout versus regular phasers.

Novacat: Same way you did for the crank time in the prior PR. Add a new /var/energy = 75 (probably not energy since that var may be used by something else) in the initial phaser. Replace 90 with energy, and then in the holdout phaser do energy = 100

code/modules/vore/fluffstuff/custom_guns_vr.dm
Novacat: It was intended. Basically, it gives time for explorers to pause and talk/RP stuff out, and discourages just running through the whole expedition area without RPing at all.

code/modules/vore/fluffstuff/custom_guns_vr.dm
Rykka-Stormheart: Updated with /var/phase_power = 75, and set holdout to 100

code/modules/vore/fluffstuff/custom_guns_vr.dm
Rykka-Stormheart: Is it still possible to disagree and request the change be kept?

Novacat: In this particular case, no, unfortunately.

Rykka-Stormheart: Flag removed.

PR #5767, opened 2019-09-01
[WIP, DNM] Adds R-0C chassis, sleeper, and Cyborg module

Rykka-Stormheart:
Does what it says on the tin:

New Mining houndborg, WIP. Uploading mostly as a "anyone can contribute" bit, or comment.
Planned abilities as followed:
1: Belly. Obv. <3
2: Ore Satchel
3: Drill
4: Tail-drill. (yes. <3) - The way this one will work is unique, I'll explain
5: Meson Scanner of some sort. Tbh, graviton goggles, but, that might be op.
6: Some sort of sonar/hostile mob detection. If NT is so concerned about their expensive AI, sending one into the UD/hostile mining environments without sonar would be silly.
7: Ore Crate Loading from Ripley
8: TBD
Now, for the tail-drill:
something like this, but the method of use is unique
you "deploy" for several moments (maybe using this sound?)
once you're "locked", you can start drilling using the tail
think of it like the big mining drills outside the outpost - obv, it needs to be less efficient to be balanced, and it has the side effect of locking you down (play it off like it needs to redirect power from your servos/actuators)
~~all mined ore goes into your hybrid ore satchel/crate, then you bring it back to facility and dump it for processing
unless, to reduce server lag (from dumping thousands of ore ingots) - we give the borghound it's own inbuilt processor, OR, we make that hybrid crate/satchel have an upper limit. A fairly decent one, mind, but that still could potentially crash the server.
However, it needs to be slower than the station one, to prevent just refining on the fly and shit being powergamed~~
- We're just stealing the ore crate loading from the Ripley.

Mostly looking for feedback, dev/sprite help, etc. I'm literally _just_ starting out with this project.


Rykka-Stormheart:
Closing this PR - I'm officially not working on it. Waiting for sprites and such, then I'll reopen it.

PR #5776, opened 2019-09-01
Added tracking + chemical implants to orderable security crate(s)

MrSign:
Adds in tracking and chemical implants to the cargo ordering list under the security tab in the event that they're needed for naughty boys and girls.


MrSign:
Hnngnf, first time actually coding. So, I got no idea what it wants me to do. AAAAAA

PR #5777, opened 2019-09-01
Implements healbelly permission pref

### LINKED ISSUE #5241: [Suggest]Healbelly Usage Can Be Tricky! Solution!
So, as Medical, it can be tricky to know if someone's okay with using the heal-belly. Some players don't see it as a prefs-thing as long as it's used to save them mechanically, others care a whole lot, understandably so, but it isn't really mentioned directly in notes/prefs.

Even if it was, it'd be difficult to scan the ooc notes of someone who's dying in crit and read through them while things are going on and you need to know if you have an opportunity to use it or not right that instant. Usually this means always asking in LOOC, which is fine when you get a quick response and awkward when it turns out this person has LOOC disabled and you don't realize you're talking to nobody. True story. 

I feel like there's a better way.

#### What I suggest:
In character setup, in the Vore panel, have a 'healbelly' setting. 3 options, 'Always Yes', 'Always No', and a custom field that allows for a small, perhaps 150 character field that'd show up as a pop-up. Exactly how flavor text shows up upon examine. Click read-more and it has a pop-up window. Do that. The reason for this is because some people may want to specify if they're only okay with it with certain people, species, ''destinations'', only for major injuries, etc. Given that every character has the potential to use the healbelly, I think having this info readily visible upon examine would be a good idea.

Edit: It's also worth noting the existence of the canine borgs. Their primary means of transport and healing would involve vore, and given that mechanically they aren't able to do a whole lot else OUTSIDE of that, having this would actually really help their job be a bit easier, too.

Heroman3003:
Resolves #5241 on @Wickedtemp's request

Implements additional pref on vore panel, healbelly permission. Doesn't affect anything mechanically, just there to show up on the mechanical pref popup

Still not entirely onboard for the idea, feeling that it is slippery slope to bloating mechanical prefs window, but its simple thing to code that nobody else was interested to.

### DIFF
--- a/code/modules/vore/eating/living_vr.dm
+++ b/code/modules/vore/eating/living_vr.dm
@@ -22,6 +22,7 @@
 	var/drain_finalized = 0				// Determines if the succubus drain will be KO'd/absorbed. Can be toggled on at any time.
 	var/fuzzy = 1						// Preference toggle for sharp/fuzzy icon.
 	var/tail_alt = 0					// Tail layer toggle.
+	var/permit_healbelly = TRUE
 	var/can_be_drop_prey = 0
 	var/can_be_drop_pred = 1			// Mobs are pred by default.
 
@@ -199,6 +200,7 @@
 	P.digest_leave_remains = src.digest_leave_remains
 	P.allowmobvore = src.allowmobvore
 	P.vore_taste = src.vore_taste
+	P.permit_healbelly = src.permit_healbelly
 	P.can_be_drop_prey = src.can_be_drop_prey
 	P.can_be_drop_pred = src.can_be_drop_pred
 
@@ -225,6 +227,7 @@
 	digest_leave_remains = P.digest_leave_remains
 	allowmobvore = P.allowmobvore
 	vore_taste = P.vore_taste
+	permit_healbelly = P.permit_healbelly
 	can_be_drop_prey = P.can_be_drop_prey
 	can_be_drop_pred = P.can_be_drop_pred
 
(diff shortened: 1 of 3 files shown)


Screemonster:
I dunno if it needs to be as specific as "allow healbelly" as much as it is a general indicator of "am I okay with randos grab-and-gulping if they have a reason to do so" - correct me if I'm wrong but the main objection to healbellies mostly seems to come from people that don't like being prey _at all_, aside from the occasional idiot that grabs and healbellies people that don't actually need urgent treatment.

Wickedtemp:
I don't really care what the name is, as long as its an easily visible pref that lets an examiner know whether or not this person is okay with getting vore'd for their own safety. Some people make exceptions to their own prefs if its to keep their character alive, some don't, so having this be its own thing is extremely useful.

Heroman3003:
Main reason its healbellies specifically, is because there is an intermedeate slew of people who are okay with random grabs and such... But as a lead-in into a scene. On the other hand, random healbelly grab would very rarely be that and more of purely mechanical healing tool, rather than scene leadin. Also, healbelly specifically because its pretty much the only thing thats can be really time sensitive about grab-n-gulp in terms of pred action, while other times, you have enough time to properly examine the notes.

As for the drop-noms, I agree, but would prefer to do it as separate PR, rather than trying to bundle it into this one

ArgobargSoup:
...Not entirely sure this properly *resolves* the initial issue.  If someone wasn't spendin' the time to click through to OOC notes, that pop up in the chatlog, to check whether or not someone's okay with getting healbellied, whether due to laziness or the time constraint of "they are dying and need to be stabilized", would they spend the time to click through to mechanical prefs, which pop up in a separate window (as far as I recall)?  Especially as the mechanical prefs are so rarely checked anyways...

Wickedtemp:
Yep, they would! Its much faster and easier to examine and then click on a button than it is to examine, click, and read through a potential wall of text that might not even say if they're okay with it.

Plus, some people have No as being prey, but would make exceptions if it meant saving their character. Usually LOOC could be used, but its hard to do that if youre also trying to kite a mob or get people out of danger, or have LOOC disabled.

Screemonster:
The main issue with OOC notes is that not all people follow the same format and trying to look through several paragraphs of text to see whether they bothered to include it when you're in a hurry is a pain, whereas this is a clear-cut yes/no that's always in the same place all the time

ArgobargSoup:
Well, slightly specific example, but goin' via the "kite a mob/get someone out of danger" thing...  I can see issues where you're tryin' ta kite a mob and ya click open the mechanical pref window, only ta end up failing to kite the mob because oh hey, new small window took focus.  \:P

Wickedtemp:
Yeah, but its still possible, just requires fast clicking. Reading OOC notes in that scenario is much, much harder.

ArgobargSoup:
Mmh, I'd just say "set it so that "view mechanical prefs" prints to chatlog instead of open up new window".  Less interrupt-y, at least, and they'll already be looking there.  \:P

PR #5784, opened 2019-09-02
Updates mining job slots to match actual equipment counts

Heroman3003:
I dunno when why it was set to four of all numbers back in 2017 of all times... But map is simply not designed to support that many miners being normally present and practice proves that three miners being around is more than enough to provide anything station may need and more.

Also purges a ton of unnecessary redefines/overrides in override file that would be same even without overriding.


Screemonster:
Honestly having these overrides in place is good in case upstream change something we don't want changed.

PR #5789, opened 2019-09-03
Fixes the washing bug with boots

Cameron653:
Fixes the issue where overlays would stack on boots, resulting in them being _much_ harder to clean, which resulted in having overlays "cake up" on boots/shoes and require multiple runs through the washer.

### DIFF
--- a/code/modules/clothing/clothing.dm
+++ b/code/modules/clothing/clothing.dm
@@ -565,10 +565,9 @@
 		overlays += image(icon, "[icon_state]_knife")
 	if(contaminated)
 		overlays += contamination_overlay
-	//VOREStation Edit Start
-	if(gurgled)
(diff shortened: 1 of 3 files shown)


code/modules/clothing/clothing.dm
```diff
@@ -558,8 +558,17 @@
...
 
 /obj/item/clothing/shoes/update_icon()
+	overlays.Cut() //This removes all the overlays on the sprite and then goes down a checklist adding them as required.
+	if(blood_DNA)
+		add_blood()
 	if(holding)
 		overlays += image(icon, "[icon_state]_knife")
+	if(contaminated)
+		overlays += contamination_overlay
+	//VOREStation Edit Start
```
Heroman3003: VOREStation Edit here starts, but never ends?

PR #5791, opened 2019-09-03
Relocate UD Entrance to SE Corner instead of NE Corner

Rykka-Stormheart:
Does what it says, also adds a part-tiled walkpath to it


HzDonut:
wut

Rykka-Stormheart:
> wut

idfk, man, I changed one tile

PR #5807, opened 2019-09-05
Restore Librarian Slot Number

TheFurryFeline:
Why was this even reduced to 1 when before the refactor, civilian.dm had a Vorestation edit for 2 slots...? Fixing this.


Novacat:
Oversights are a thing.

PR #5811, opened 2019-09-06
A new Poi

BobofBoblandia:
What it says on the tin


BobofBoblandia:
image

Heroman3003:
Not much to say with something that simplistic, but the cost not being division of 5 and this one being uniquely only example of PoI that isn't using it leaves high chance of this leaving out some PoIs for no good reason. With how similar it is to generic normal mob spawn in contents, I think it having a 5-cost would be perfectly reasonable.

As fo visual design, I'd suggest smoothing those hard corners on the edges a bit with a tile or two, because of how square-looking it is, but thats more of a visual nitpick

BobofBoblandia:
The reason why I did not make it a divisible of five is twofold, one part being Im dumb
The other part being that I wanted it so that it was so that it would be slightly more rare than the ones divisible by five

Heroman3003:
Cost does not contribute much to rarity too much. Usual distribution of PoIs is rather even across all costs, as the general selection relies entirely upon rng first, and cost of what its selecting secondarily.

BobofBoblandia:
Then it was just stupidity, got it

Rykka-Stormheart:
> Needs mapmerge ran on it

mapmerge tool is busted

PR #5812, opened 2019-09-06
More flavor for PoIs

Heroman3003:
Two things:

First of all, additional PoI for the debris field, for nostalgic value of old shuttle.

Second of all. Makes a virgo version of lost drone ghostpod, with an option to be a lost dogdrone!
Added to some of existing PoIs, and with option to add potentially more vorish laws to dogborg half of the equation.
Pods can be found in big PoIs of debris field, Virgo 4 caves and Virgo 2 surface


code/modules/mob/living/silicon/robot/robot_modules/station_vr.dm
```diff
@@ -136,7 +136,7 @@
 /obj/item/weapon/robot_module/robot/knine
 	name = "k9 robot module"
 	sprites = list(
-					"K9 hound" = "k9",
+					"K9 hound" = "stray",
```
Novacat: ...I think this was a test change that somehow made it in?

code/modules/mob/living/silicon/robot/dogborg/dog_modules_vr.dm
```diff
@@ -158,6 +158,13 @@
 					reagent_volumes[T] = min(reagent_volumes[T] + 1, volume)
 	return 1
 
+/obj/item/weapon/reagent_containers/borghypo/hound/lost
+	name = "Hound hypospray"
+	desc = "An advanced chemical synthesizer and injection system utilizing carrier's reserves."
+	charge_cost = 5
+	reagent_ids = list("tricordrazine", "inaprovaline", "bicaridine", "dexalin", "anti_toxin", "tramadol", "spaceacillin", "hyperzine")
```
Novacat: This is a bit too strong. It already has far less charge cost than most borgs, and a better array of chems than most borgs, including Polaris' own lost borg analogue. I would increase charge cost to 10 (like medihounds) and give it identical chems to the Polaris lost borg

Heroman3003: The reason I buffed variety was mostly because loss of R&D tool, so inability to help research is made up for with better medical prowess. I'll up charge cost back, was a bit overkill on that. But chem additions were only inaprovaline (which is pretty inconsequential tbh? with bicaridine already being there, I added it mostly for flavor) and hyperzine, which is mild but acceptable with spendings on water and energy present?

code/modules/mob/living/silicon/robot/robot_modules/station_vr.dm
Heroman3003: ...big whoops. Gonna fix that

code/modules/mob/living/silicon/robot/dogborg/dog_modules_vr.dm
Novacat: I would trim hyperzine off, at the very least. My main concern is that someone is going to probably SCA one of these things at some point.

Heroman3003: Alright. I'm also removing SleeveMate, since the borg is primarily expedition encounter, and having sleeve stuff over there is rather redundant/wrong-thing-encouraging. Especially with flavor being that they aren't necessarily Virgo's borg.

PR #5813, opened 2019-09-06
Corrupt Maintenance Drone

Rykka-Stormheart:
Adds a corrupt maintenance drone - intended for use in POI's and such, uses spider venom injection code to "poison" it's targets with welder fuel.


code/modules/mob/living/simple_mob/subtypes/mechanical/corrupt_maint_drone.dm
```diff
@@ -0,0 +1,54 @@
+/*
+	A corrupted maintenance drone, produced from what seems like a bad factory.
+	They also tend to dodge while in melee range.
+	Code "borrowed" from viscerator drones. <3
+*/
+
+/datum/category_item/catalogue/technology/drone/corrupt_maint_drone
```
TheFurryFeline: You're editing a Polaris file. Might want to include a //Vorestation Addition comment thing here. Or make a _vr.dm file with an identical name, plus the underscore.

Rykka-Stormheart: This is an entirely new file, though?

TheFurryFeline: Ah, yes, it is. Just needs to be renamed to be, like, corrupt_maint_drone_vr.dm like we got for the vore panel in vorepanel_vr.dm

Novacat:
The only thing I am kind of iffy on is this thing having an exactly identical sprite to normal maintenance drones. Probably not a showstopper, but admins would have to be very careful on where they spawn this in.

chaoko99:
Urist has overtly identical but hostile maintenance drones. Only difference is a red eye. It's not that much of an issue.

Rykka-Stormheart:
> The only thing I am kind of iffy on is this thing having an exactly identical sprite to normal maintenance drones. Probably not a showstopper, but admins would have to be very careful on where they spawn this in.

Corrected the concern by adding a sprite for the corrupt maint drone - simply changes the eye color from yellow to red. :3c

TheFurryFeline:
> > The only thing I am kind of iffy on is this thing having an exactly identical sprite to normal maintenance drones. Probably not a showstopper, but admins would have to be very careful on where they spawn this in.
> 
> Corrected the concern by adding a sprite for the corrupt maint drone - simply changes the eye color from yellow to red. :3c

Again, you may want to add it to a _vr version so it doesn't get overwritten with the next update to it Polaris-side. :V

PR #5817, opened 2019-09-07
Morph fixes and adjustments

Heroman3003:
Fixes morph icon breaking when eating using animal-nom verb rather than grab intent attack

Increased health, damage, and damage range.

Immunity to stun-related damage simplemobs get from tasers and stun batons (because fuck security)

Verbs for physical interaction changed, now same as mimic's (no longer weird 'tries to help' thing)


Novacat:
Can you split the verbs and icons from the balance changes? The latter may require admin input.

Heroman3003:
will make two separate PRs

PR #5820, opened 2019-09-08
Makes morphs sturdier

Heroman3003:
Increased health to 250

Removed the 'simplemob that takes damage from tasers' thing from them

Fixed damage range to be 15-20 rather than 20-20

All of those numbers originally were pretty much placeholders since they were not expected to get into direct combat or be under direct attack under normal circumstances, being adminspawn-only, player-control-only mob on HRP server, but whatdayaknow. These new numbers are meant to be more 'balanced' for what mob is.

### DIFF
--- a/code/modules/mob/living/simple_mob/subtypes/vore/morph/morph.dm
+++ b/code/modules/mob/living/simple_mob/subtypes/vore/morph/morph.dm
@@ -24,9 +24,10 @@
 	max_n2 = 0
 
 	minbodytemp = 0
-	maxHealth = 150
-	health = 150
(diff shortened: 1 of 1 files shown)


Screemonster:
this smells of salt tbh

Heroman3003:
I haven't had a chance to test them while being alone on test server, and didn't even think to. I also know nothing about mob balance and was porting them from codebase radically different in terms of how mob balance works. And ideally, the taser thing should be disabled on most mobs tbh because it results in taser/baton being stronger than some 'real' weapons with how stun-to-damage calculation works.

PR #5830, opened 2019-09-08
You thought cafe was gonna be relevant again? HA

Rykka-Stormheart:
Adds coffee dispenser to the Bar, because nobody ever fuckin' uses the Cafe.


Heroman3003:
Tbh I just think adding some of the more 'general' cartridges from coffee dispenser to also be in soda dispenser would be enough. Like juices that are lacking and maybe base coffee. Because the only reason its usually stolen is those more 'common' things that just aren't really available in main bar.

Rykka-Stormheart:
> Tbh I just think adding some of the more 'general' cartridges from coffee dispenser to also be in soda dispenser would be enough. Like juices that are lacking and maybe base coffee. Because the only reason its usually stolen is those more 'common' things that just aren't really available in main bar.

Nah, this is a better idea, imo. Soda is soda, why would you add the cartridges from coffee into it?

ArgobargSoup:
Like I said in the development channel, as someone who's main character is a barista, and who would work the cafe while playing that character, am feeling a bit put off by this PR.  \:P

Rykka-Stormheart:
> Like I said in the development channel, as someone who's main character is a barista, and who would work the cafe while playing that character, am feeling a bit put off by this PR. :P

All this does is add a coffee machine to the bar. Saves people from constantly unwrenching the one in the cafe - and now if you wanna run the cafe, you can, even if the bar are jerks and dispensing coffee-based drinks.

SpadesNeil:
We need more incentive to use the cafe. Not less.

TheDavestDave:
I think we should mave cafe back to space 1. As is people expect it to be closed, and it's so close to bar if people even notice someone work it they all scream at you to open bar. Being at cafe also gives people a place to hang out when surface is filled with phoron

Rykka-Stormheart:
> I think we should mave cafe back to space 1. As is people expect it to be closed, and it's so close to bar if people even notice someone work it they all scream at you to open bar. Being at cafe also gives people a place to hang out when surface is filled with phoron

shit, you've gone and inspired me. Brb moving cafe to Station levels.

PR #5832, opened 2019-09-08
Adjusts RD's rigsuit

Heroman3003:
Changes access restriction to Research Director rather than Xenoarcheology

Replaces diamond drill module with plasma cutter module


chaoko99:
Please provide a reason.

Heroman3003:
The access change is only for the roundstart mapped in one. Any AMIs acquired via cargo or whatever other way than the one starting in RD office have no restriction. Main reasons being that A) That suit, according to one of more respected RD player, is literally stolen from RD office just because 'they want it', while its meant to be RD-only.

As for change to plasma cutter. If its meant for xenoarch, drill speed doesn't matter that much. Research shouldn't have better roundstart mapped in tools for mining than mining. And according to sources, some researchers HAVE literally went out with that suit just for mining.

PR #5843, opened 2019-09-11
Adds 10 new UD POI's

Rykka-Stormheart:
POI's listed as follows:
Abandoned Outpost
~~Broken Engine~~  (Commented out pending radiation rework)
Mimicry
Old Drone Hive
Phoron Rat Den
Puzzle Corridor
~~Rad Threat~~  (Commented out pending radiation rework)
Rykka's Easter Egg
Spider Nest
Subterranean Lake
Tree Shrine
Wolf Den

Screenshots: https://imgur.com/a/gpzJ295

### DIFF
--- a/maps/tether/submaps/underdark_pois/_templates.dm
+++ b/maps/tether/submaps/underdark_pois/_templates.dm
@@ -85,7 +85,72 @@
 	cost = 5
 	allow_duplicates = FALSE
 
+/datum/map_template/underdark/old_drone_hive
+	name = "Underdark Old Drone Hive"
+	mappath = 'old_drone_hive.dmm'
+	cost = 25
+
+/datum/map_template/underdark/phoron_rat_den
+	name = "Underdark Phoron Rat Den"
+	mappath = 'phoron_rat_den.dmm'
+	cost = 35
+	allow_duplicates = FALSE
+
+/datum/map_template/underdark/subterranean_lake
+	name = "Underdark Underground Lake"
+	mappath = 'subterranean_lake.dmm'
+	cost = 5
+
+/datum/map_template/underdark/spider_nest
+	name = "Underdark Spider Nest"
+	mappath = 'spider_nest.dmm'
+	cost = 15
+
+/datum/map_template/underdark/rykka_easter_egg // bark bark Rykka was here. <3
+	name = "Underdark GSD"
+	mappath = 'rykka_easter_egg.dmm'
+	cost = 5
+	allow_duplicates = FALSE
+
+/datum/map_template/underdark/tree_shrine
+	name = "Underdark Tree"
+	mappath = 'tree_shrine.dmm'
+	cost = 10
+	allow_duplicates = FALSE
+
+/datum/map_template/underdark/abandoned_outpost
+	name = "Underdark Abandonded Outpost"
+	mappath = 'abandonded_outpost.dmm'
+	cost = 45
+	allow_duplicates = FALSE
+
+/datum/map_template/underdark/mimicry
+	name = "Underdark Mimic Death"
+	mappath = 'mimicry.dmm'
+	cost = 15
+
+/datum/map_template/underdark/wolf_den
+	name = "Underdark Wolf Den"
+	mappath = 'wolf_den.dmm'
+	cost = 15
+
+/datum/map_template/underdark/puzzle_corridor
+	name = "Underdark Puzzle Corridor"
+	mappath = 'puzzle_corridor.dmm'
+	cost = 10
+
 /*
+// Comment out unfinished/unbalanced POI's here
+/datum/map_template/underdark/rad_threat
+	name = "Underdark Rad Threat"
+	mappath = 'rad_threat.dmm'
+	cost = 10
+
+/datum/map_template/underdark/broken_engine
+	name = "Underdark Broken Twin Engine"
+	mappath = 'broken_engine.dmm'
+	cost = 10
+
 /datum/map_template/underdark/boss_mob
 	name = "Underdark Boss Mob Spawn"
 	mappath = 'boss_mob.dmm'
--- a/maps/tether/submaps/underdark_pois/abandonded_outpost.dmm
+++ b/maps/tether/submaps/underdark_pois/abandonded_outpost.dmm
@@ -1,0 +1,1045 @@
+//MAP CONVERTED BY dmm2tgm.py THIS HEADER COMMENT PREVENTS RECONVERSION, DO NOT REMOVE
+"aa" = (
+/turf/template_noop,
+/area/mine/explored/underdark)
+"ab" = (
+/turf/simulated/wall,
+/area/mine/explored/underdark)
+"ac" = (
+/turf/template_noop,
+/area/space)
+"ad" = (
+/turf/simulated/mineral/floor/ignore_cavegen/virgo3b,
+/area/mine/explored/underdark)
+"ae" = (
+/turf/simulated/floor/tiled/steel_dirty/virgo3b,
+/area/mine/explored/underdark)
+"af" = (
+/obj/machinery/door/airlock/uranium,
+/turf/simulated/floor/tiled/steel_dirty/virgo3b,
+/area/mine/explored/underdark)
+"ag" = (
+/obj/machinery/door/airlock/multi_tile/glass{
+	dir = 1
+	},
+/turf/simulated/floor/tiled/steel_dirty/virgo3b,
+/area/mine/explored/underdark)
+"ah" = (
+/obj/machinery/floodlight,
+/turf/simulated/mineral/floor/ignore_cavegen/virgo3b,
+/area/mine/explored/underdark)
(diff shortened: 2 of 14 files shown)


Heroman3003:
Abandoned Outpost: I dunno about the mechwrecks present there, as they are combat mechs and provide loot completely unrelated to mining as a result

Broken Engine and Rad Threat: Mining already struggles quite a bit against radiation as a whole, I dunno about adding even more issues related to that to just blacklist them from ever entering an area

Rykka Easter Egg: The doggo will die of atmos before any miner will get to go down there

Wolf Den: pretty sure wolves and otie will attempt to fight. Although I'm betting on otie due to, again, wolves just dying of atmos before anything happens. Also uncontained artifact is uhhh.. Same issue as radiation, mining is pretty ill-equipped to deal with it, and depending on RNG it might end up just being big area denial

Generally applicable to all: I didn't see any new /virgo3b turfs defined, but definitely saw some new turfs used. That will probably result in any new tiles having A) active edges and B) phoron overlay all over them, unlike normal tiles in the Underdark

Rykka-Stormheart:
> Abandoned Outpost: I dunno about the mechwrecks present there, as they are combat mechs and provide loot completely unrelated to mining as a result

You're not clearing the outpost as a solo miner. It's not a huge deal, you need a team to clear the damn thing anyways.

> Broken Engine and Rad Threat: Mining already struggles quite a bit against radiation as a whole, I dunno about adding even more issues related to that to just blacklist them from ever entering an area

Mining has geiger counters, and the voidsuits are rad-insulated.

> Rykka Easter Egg: The doggo will die of atmos before any miner will get to go down there

...how do I fix that? :c

> Wolf Den: pretty sure wolves and otie will attempt to fight. Although I'm betting on otie due to, again, wolves just dying of atmos before anything happens. Also uncontained artifact is uhhh.. Same issue as radiation, mining is pretty ill-equipped to deal with it, and depending on RNG it might end up just being big area denial

Regarding RNG, artifact, and such, not a huge deal to me, imo. That's part of the risk. I intend to leave it in and judge how it goes. Do remember if it's too much, we can tweak, reduce, or eliminate it later, yeah?

> Generally applicable to all: I didn't see any new /virgo3b turfs defined, but definitely saw some new turfs used. That will probably result in any new tiles having A) active edges and B) phoron overlay all over them, unlike normal tiles in the Underdark

All turfs are v3b tiles.

Heroman3003:
> You're not clearing the outpost as a solo miner. It's not a huge deal, you need a team to clear the damn thing anyways.

Well, I am saying that on assumption that Underdark is still meant to be something doable as a singular person, as it was when it was begun and when first PoIs were introduced, with all threats being meant to be handlable by a single miner who just happened upon the PoI while digging ores.

> Mining has geiger counters, and the voidsuits are rad-insulated

Geiger counters don't help much beyond just allowing to know the bounds of the zone thats as good as un-enterable, and the shielding on suits is only partial and will not actually protect from radiation. Even suit from vendor doesn't have full protection from it. Radiation sources are a dead zone.

> ...how do I fix that? :c

Need to define mobs properly, with proper stats to accomodate virgo3b survival

> Regarding RNG, artifact, and such, not a huge deal to me, imo. That's part of the risk. I intend to leave it in and judge how it goes. Do remember if it's too much, we can tweak, reduce, or eliminate it later, yeah?

You do you. But artifact RNG can be absolute bitch, as practice showed

> All turfs are v3b tiles.

Both tree shrine and lake use non-v3b tiles

Rykka-Stormheart:
> > You're not clearing the outpost as a solo miner. It's not a huge deal, you need a team to clear the damn thing anyways.
> 
> Well, I am saying that on assumption that Underdark is still meant to be something doable as a singular person, as it was meant to be when it first begun and when first PoIs were introduced, with all threats being meant to be handlable by a single miner who just happened upon the PoI while digging ores.

I think having the one large, extremely rare point won't hurt. You'll know to avoid it if you see it from experience.
 
> > Mining has geiger counters, and the voidsuits are rad-insulated
> 
> Geiger counters don't help much beyond just allowing to know the bounds of the zone thats as good as un-enterable, and the shielding on suits is only partial and will not actually protect from radiation. Even suit from vendor doesn't have full protection from it. Radiation sources are a dead zone.
> 
> > ...how do I fix that? :c
> 
> Need to define mobs properly, with proper stats to accomodate virgo3b survival

Where do I define the mob? Or is there a var I can change on the existing mob?
 
> > Regarding RNG, artifact, and such, not a huge deal to me, imo. That's part of the risk. I intend to leave it in and judge how it goes. Do remember if it's too much, we can tweak, reduce, or eliminate it later, yeah?
> 
> You do you. But artifact RNG can be absolute bitch, as practice showed

> > All turfs are v3b tiles.
> 
> Both tree shrine and lake use non-v3b tiles

That's fine - they're simply there to look pretty. So what if they have a slight purple effect on them?

Heroman3003:
> I think having the one large, extremely rare point won't hurt. You'll know to avoid it if you see it from experience.

Another reminder that point cost does not actually affect rarity of PoIs as much as it may seem. It primarily determines how much 'worth' is the PoI's contents, but only really secondarily how often it appears.

> Where do I define the mob? Or is there a var I can change on the existing mob?

There is a file with all Underdark things separately defined where you can put /3b versions of mobs. As for var, its not a single one but half a dozen vars that determine atmos suspectibility. Look it up in existing mobs

> That's fine - they're simply there to look pretty. So what if they have a slight purple effect on them?

Active edges are bad and should be avoided at all costs. To note what an active edge is: When roundstart atmos on two adjacent atmos-passing tiles is different, resulting in before-round gas movements. That should not be happening. Purple gas effect may be fine, but the fact that those tiles will contain indoor air rather than 3b atmosphere isn't.

Rykka-Stormheart:
> > I think having the one large, extremely rare point won't hurt. You'll know to avoid it if you see it from experience.
> 
> Another reminder that point cost does not actually affect rarity of PoIs as much as it may seem. It primarily determines how much 'worth' is the PoI's contents, but only really secondarily how often it appears.

Again, I'm leaving as-is and waiting to see how it goes/etc.

> > Where do I define the mob? Or is there a var I can change on the existing mob?
> 
> There is a file with all Underdark things separately defined where you can put /3b versions of mobs. As for var, its not a single one but half a dozen vars that determine atmos suspectibility. Look it up in existing mobs
> 
> > That's fine - they're simply there to look pretty. So what if they have a slight purple effect on them?
> 
> Active edges are bad and should be avoided at all costs. To note what an active edge is: When roundstart atmos on two adjacent atmos-passing tiles is different, resulting in before-round gas movements. That should not be happening. Purple gas effect may be fine, but the fact that those tiles will contain indoor air rather than 3b atmosphere isn't.

Tree Shrine corrected, subt lake is using water tiles - there's no such v3b tile for water.

Rykka-Stormheart:
> Wolf Den: pretty sure wolves and otie will attempt to fight.

This is also fixed, phoron wolves are set to "underdark" faction, as is the otie I var-edited

Heroman3003:
You can easily create new V3B tiles based off of existing ones without much hassle by using same method as demonstrated at the very bottom of Underdark things file.

Novacat:
My first immediate concern is that, the Underdark is meant to be soloable by Miners. Underdark POIs are not supposed to kick your ass like Exploration ones, and a lot of the ones here have... quite a few mobs, which makes me concerned in regards to that.

Rykka-Stormheart:
> My first immediate concern is that, the Underdark is meant to be soloable by Miners. Underdark POIs are not supposed to kick your ass like Exploration ones, and a lot of the ones here have... quite a few mobs, which makes me concerned in regards to that.

Plenty of UD spawns already have 3+ mobs - the "drone swarm" ones for instance are just a 'group' of very small "viscerator-alike" drones, for instance. The mob refactor should help massively, as mobs cannot attack anywhere near as fast anymore.

Screemonster:
Yeah, underdark is a solo area which makes me wary of putting anything in there that's intended for a team.

Also radiation fucking sucks. Environmental hazards in general under our medical system tend to be a case of "either you gear up to be totally immune, or else you either die or have to stop what you're doing to seek treatment". There's no such thing as a safe threshold where you can limit your exposure to avoid damage, you're either immune or you're not. And miners being out of the station, as far from medbay as you can get, and usually solo, increases the suck factor.

Exploration PoIs? Sure. Mining? Nah.

Rykka-Stormheart:
> Yeah, underdark is a solo area which makes me wary of putting anything in there that's intended for a team.
> 
> Also radiation fucking sucks. Environmental hazards in general under our medical system tend to be a case of "either you gear up to be totally immune, or else you either die or have to stop what you're doing to seek treatment". There's no such thing as a safe threshold where you can limit your exposure to avoid damage, you're either immune or you're not. And miners being out of the station, as far from medbay as you can get, and usually solo, increases the suck factor.
> 
> Exploration PoIs? Sure. Mining? Nah.

Concerns addressed

PR #5848, opened 2019-09-13
Protean Changes

Novacat:
- Changes Protean to Whitelist selectable
- Increases brute mod, decreases burn mod
- Halves Siemens coefficient. Zappy is still bad for Proteans, just not as bad.
- Removes Darksight
- Overhauls material consumption and effects. Steel and Metallic hydrogen are unchanged, but brute resist is now enhanced by uranium, burn resist by gold, and silver briefly enhances accuracy/evasion.
- Some variables have been properly replaced by defines. This might fix a few runtimes.

I never was a fan of the 'Balance of extremes' of Proteans being near-immune to Brute and brute-related injuries, but a single shot with a laser or ion weapon practically vaporizes them.

However, Proteans are going to be problematic because they are basically Synths that get a ton of extra perks. They cannot get broken limbs, they can regenerate by just eating metal sheets (and pretty much never need roboticists or medical), and they can get a huge number of bonuses by just eating material sheets.

So, for that reason, some of those perks are probably going to be tweaked, nerfed, and removed, and Proteans overall are going to take more brute/burn than other species in attempt to balance out those perks.

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/protean_vr/protean_powers.dm
+++ b/code/modules/mob/living/carbon/human/species/station/protean_vr/protean_powers.dm
@@ -200,7 +200,7 @@
 
 	var/obj/item/stack/material/matstack = held
 	var/substance = matstack.material.name
-	var/list/edible_materials = list("steel", "plasteel", "diamond", "mhydrogen") //Can't eat all materials, just useful ones.
+	var/list/edible_materials = list(MAT_STEEL, MAT_SILVER, MAT_GOLD, MAT_URANIUM, MAT_METALHYDROGEN) //Can't eat all materials, just useful ones.
 	var allowed = FALSE
 	for(var/material in edible_materials)
 		if(material == substance) allowed = TRUE
@@ -298,14 +298,14 @@
 
 	//Sizing up
 	if(cost > 0)
-		if(refactory.use_stored_material("steel",cost))
+		if(refactory.use_stored_material(MAT_STEEL,cost))
 			user.resize(size_factor)
 		else
 			to_chat(user,"<span class='warning'>That size change would cost [cost] steel, which you don't have.</span>")
 	//Sizing down (or not at all)
 	else if(cost <= 0)
 		cost = abs(cost)
-		var/actually_added = refactory.add_stored_material("steel",cost)
+		var/actually_added = refactory.add_stored_material(MAT_STEEL,cost)
 		user.resize(size_factor)
 		if(actually_added != cost)
 			to_chat(user,"<span class='warning'>Unfortunately, [cost-actually_added] steel was lost due to lack of storage space.</span>")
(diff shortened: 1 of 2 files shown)


Heroman3003:
I do think that them takin even if slightly less brute (like 0.9) modifier would be reasonable enough to add 'flavor' without affecting balancing. And same for burn (1.4-1.5 modifier) weakness. Their perks of regeneration are countered by exactly that, being a glass cannon, blob form is less of a perk and more of a fluff thing (outside ventcrawl aspect, but that should be point of its own), and material eating is fun... But that one I agree, needs to be probably redone from scratch both in terms of edible materials and gained buffs from that.

Wickedtemp:
Having automatic ventcrawl is a recipe for disaster. In a game where movement-speed generally wins out in most mechanic comparisons, being able to effectively go anywhere regardless of access is a HUGE perk. Other people at the least have to get access or hack through several airlocks. Proteans can just...ventcrawl wherever and go straight to their destination and nobody can even see them while they do so. They could even hide in the vents for extended periods of time, iirc while still being able to see everything going on in the rooms that they're below, allowing them to basically ghost over areas to determine if it's safe to come out or not. In cases where a Protean may be pursued by Security, this is laughably broken.

Is there anything we can do, ICly, about this?

Novacat:
Without removing ventcrawl from blobform proteans, no.

Though, there was a change recently where blobformed Proteans can only keep a handful of things on them at a time. So they will drop most things in their inventory when they go blobform.

Wickedtemp:
...Not much of an issue when you can blobform and ventcrawl anywhere you want and there's nothing anybody can do to stop you. You could evade anything you wanted that way unless the crew starts welding every single vent on the Tether.

Heroman3003:
I do believe that is the main reason they were planned as a whitelisted species then.

Wickedtemp:
Yeah. but I always prefer IC ways of handling those situations, so it isn't "Aha you're basically at my mercy until an Admin steps in." Especially with something that's so easily abusable even if it's not intended as malicious. If going blobform drops *everything* then that'd make a little more sense. If they're a blob, their uniform and everything else they have won't exactly stay on.

Heroman3003:
While I agree with that, at least some extent of 'you're at their mercy' is sorta... Part of protean lore? That being, they are literally grey goo in terms of potential and only reason they aren't all consuming is because they don't want to be (or because the law restricts them, but who obeys law anyway)

Wickedtemp:
... "We could totally just consume everything, in fact it only takes a single member of this species to deviate and then boom entire buildings get evaporated and consumed and an entire city or state or country or potentially planet is now facing an existential threat." sounds like shit lore that wouldn't work out in anything except them being deemed existentice-threatening creatures that planetary governments would be having an arms-race over.

Heroman3003:
Thats the reason they literally have to carry around SolGov-issued permits that even allow them to exist.

Wickedtemp:
...if they're ACTUALLY that dangerous, wouldn't protocal for a Protean thats not cooperating simply be immediate extermination

Screemonster:
finally, a species that'll make people scream "why are these allowed" louder than they do for the literal virus colonies

PR #5890, opened 2019-09-17
Remakes the entire tether midpoint

Mewchild:
Replaces all the admin walls with BEAUTIFUL SKY
Adds TWO NEW AREAS, WOW.
Probably lets people fall to the surface from space. AMAZING.
Relaxing spot, good view, coffee and snacks. BIG COMF

image


KasparoVy:
Oh hell yeah. That must be one wild view

code/game/area/Space Station 13 areas.dm
```diff
@@ -861,6 +861,10 @@ NOTE: there are two lists of areas in the end of this file: centcom and station
 	name = "Tool Storage Maintenance"
 	icon_state = "maint_tool_storage"
 
+/area/maintenance/tether_midpoint
```
Novacat: This will need to be moved to maps/tether/tether_areas2

Very-Soft:
I like it but ya, you'll have to be careful, as spacewalks might turn into plummeting to crater into the surface if you go down too far.

Screemonster:
people leaping into the void and becoming a crater sounds hilarious, as long as people can't accidentally end up falling there due to bullshit on the space level

Novacat:
Mreeeh.

On one hand, I really, really dislike how we have essentially a completely wasted Z level.

On the other hand, opening this Z level does open up future problems, like the one soft mentioned. All it would take is someone doubleclicking 'Go down' twice instead of once.

Very-Soft:
I had an idea. Could we make it so that the sky around -most- of the map was not something that could be traversed? Like, put blockers in, or set some of the sky's density to 1. So! People could still break and leap out of the window to their death, but couldn't accidentally fall from space?

Very-Soft:
Like, I don't actually see an issue with people falling to their deaths anywhere, it's more that I'm pretty sure this would also work in reverse, and allow people with flight to fly from the surface into space also, which is something I don't think should be possible.

Novacat:
Would it be possible to give the sky turfs 'invisible flooring'? This would both prevent people from falling, and prevent people from digging 'up'

Mewchild:
I'll just re-add the admin walls to the edges of the map, where people could _potentially_ fall from space.

Mewchild:
Added step teleporters so people will fall from the sky, if they step out into it. You CAN NOT fly out over them, but that's a non-issue as you'd have to break down the windows to go flying anyway. 
Added a ring of blockers around the midpoint, to keep any flyers out of view of the admin walls
Adds a sillhouette of Asteroid 1's space, made of admin walls. This will prevent anybody from falling from space into the sky. With the space left, if a bomb breaks open the floor of Asteroid one to revel the sky, you can fall out the bottom of the station, through the sky, and onto the surface (gibbing, most likely, if no parachute). Working as intended
image
image

code/game/objects/effects/step_triggers.dm
```diff
@@ -6,6 +6,8 @@
 	invisibility = 99 // nope cant see this shit
 	plane = ABOVE_PLANE
 	anchored = 1
+	icon = 'icons/mob/screen1.dmi'
```
Novacat: This should go in a _vr file or //VOREstation Edit needs to be added.

PR #5891, opened 2019-09-17
Comments out design for DNA console

Heroman3003:
Primarily because of how broken it is, and it should really have been done back when genetics in general was being ripped and removed. This is basically last leftover remnant of it.

Does not touch cloning console, or scanner/grower pods for it as it actually seems to function reasonably well.

### DIFF
--- a/code/modules/research/designs/circuits.dm
+++ b/code/modules/research/designs/circuits.dm
@@ -75,11 +75,13 @@ CIRCUITS BELOW
 	build_path = /obj/item/weapon/circuitboard/operating
 	sort_string = "FACAA"
 
+/*		//VOREStation Removal: DNA modification specifically, cloning is still a thing that actually reasonably works
 /datum/design/circuit/scan_console
 	name = "DNA machine"
 	id = "scan_console"
 	build_path = /obj/item/weapon/circuitboard/scan_consolenew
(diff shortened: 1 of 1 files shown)


Whiskyrose:
The DNA machine isn't broken and still functions when built beside a clone scanner. It just takes a great deal more effort to actual do genetics now. I'm against it's complete removal.

Very-Soft:
I don't really see why we would want to do this. This console does work at least in some limited capacity, and I have seen people using it on a number of occasions, recently even, to get at what little it can still do.

I don't think I've seen anyone get anything especially abusable from this, and they have to go through a bit of trouble to get it going in the first place. I don't really understand what the issue is with allowing people that.

Screemonster:
why though

Novacat:
Staff has had discussion on this, and it was decided to keep this around. Functionally, it still works. It has not been abused heavily.

DemicusMaximus:
Leave it be. The potential for abuse is negligible, and I still see people having some fun with it now and then.

drexample:
It works on synths, just so you know. 
So do injectors.

TheFurryFeline:
> It works on synths, just so you know.
> So do injectors.

That's an easy one to fix with something as simple as adding in something like `if (isSynthetic)`

PR #5892, opened 2019-09-17
Budget insuls now provide a minimum level of protection.

chaoko99:
Please stop punishing medical for a tantrum TGstation's players had 8 years ago.

### DIFF
--- a/code/game/machinery/vending.dm
+++ b/code/game/machinery/vending.dm
@@ -1032,7 +1032,7 @@
 					/obj/item/weapon/tool/wrench = 5,/obj/item/device/analyzer = 5,/obj/item/device/t_scanner = 5,/obj/item/weapon/tool/screwdriver = 5,
 					/obj/item/device/flashlight/glowstick = 3, /obj/item/device/flashlight/glowstick/red = 3, /obj/item/device/flashlight/glowstick/blue = 3,
 					/obj/item/device/flashlight/glowstick/orange =3, /obj/item/device/flashlight/glowstick/yellow = 3)
-	contraband = list(/obj/item/weapon/weldingtool/hugetank = 2) //VORE: Removes budget insulated gloves.
+	contraband = list(/obj/item/weapon/weldingtool/hugetank = 2,/obj/item/clothing/gloves/fyellow = 2,)
 	premium = list(/obj/item/clothing/gloves/yellow = 1)
 	req_log_access = access_ce
 	has_logs = 1
--- a/code/game/machinery/vending.dm
+++ b/code/game/machinery/vending.dm
@@ -1032,7 +1032,7 @@
 					/obj/item/weapon/tool/wrench = 5,/obj/item/device/analyzer = 5,/obj/item/device/t_scanner = 5,/obj/item/weapon/tool/screwdriver = 5,
 					/obj/item/device/flashlight/glowstick = 3, /obj/item/device/flashlight/glowstick/red = 3, /obj/item/device/flashlight/glowstick/blue = 3,
 					/obj/item/device/flashlight/glowstick/orange =3, /obj/item/device/flashlight/glowstick/yellow = 3)
-	contraband = list(/obj/item/weapon/weldingtool/hugetank = 2,/obj/item/clothing/gloves/fyellow = 2,)
+	contraband = list(/obj/item/weapon/weldingtool/hugetank = 2) //VORE: Removes budget insulated gloves.
 	premium = list(/obj/item/clothing/gloves/yellow = 1)
 	req_log_access = access_ce
 	has_logs = 1
--- a/code/modules/clothing/gloves/color.dm
+++ b/code/modules/clothing/gloves/color.dm
@@ -11,13 +11,13 @@
 	icon_state = "yellow"
 	siemens_coefficient = 1			//Set to a default of 1, gets overridden in initialize()
 	permeability_coefficient = 0.05
-
+/*Vorestation edit: Budget gloves no longer protect you at all.
 /obj/item/clothing/gloves/fyellow/Initialize()
 	. = ..()
 	//Picks a value between 0 and 1.25, in 5% increments // VOREStation edit
 	var/shock_pick = rand(0,25) // VOREStation Edit
 	siemens_coefficient = shock_pick * 0.05
-
+*/
 /obj/item/clothing/gloves/black
 	desc = "These work gloves are thick and fire-resistant."
 	name = "black gloves"
--- a/code/modules/clothing/gloves/color.dm
+++ b/code/modules/clothing/gloves/color.dm
@@ -5,12 +5,12 @@
 	siemens_coefficient = 0
 	permeability_coefficient = 0.05
 
-/obj/item/clothing/gloves/fyellow                             //Cheap Chinese Crap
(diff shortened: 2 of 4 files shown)


chaoko99:
Also yes, Ascian, I am still sick. this was an easy change.

Heroman3003:
Tbh it'd be ultimately better to just have them renamed into yellow gloves and keep them in generic trash lists, but remove from tool vendor contraband and other engineering-related loot lists

Mewchild:
If we're going to make them the exact same as black gloves in terms of resistance, why keep them in the map at all? It provides a false idea of safety, where people still think "these might protect me" and they'll now always be punished, instead of just sometime.

chaoko99:
Addressed.

Very-Soft:
I don't really understand why we need this change. They have a different name and description, and -can- occasionally protect you from shocks iirc. 

I think it's fine to allow people to make mistakes with things like this? This seems like more baby proofing.

chaoko99:
It's less about babyproofing, and more about making Medical's life not abject suffering. I'm not here to protect the person hacking, I'm here to make it so med stops getting punished because some idiot was doing something they have no business doing in the first place. Why does a random visitor (It's always an assistant) need to hack into some random maint room with nothing in it?

So it's harmless on paper, but when you consider the _**Eight years of perspective**_ we've had on this issue, I think it's best gone.

budget insulated gloves only existed due to the removal of public gloves from TGstation eight years ago. The playerbase had a fit over it, and budgets were made as a concession. The only people who use budgets are newbies and unabashed fools, as there are so many *much* safer ways to hack a door without proper protection.

Posit if you will; the following situation. "A random visitor has the knowledge on how to hack electrical systems, so they elect to do so. They proceed to don the notoriously ineffective joke item and fry their hands off." Why would someone do this? How the hell can the old gloves be MORE conductive than flesh? That means nothing! That's not how that works!

It's been worded rather succinctly by Donut: "Medical isn't given work, they're given punishment, forced to clean up after some idiot doing things they have no business doing."

Screemonster:
how far have we strayed from the spirit of SS13 when some dumbass injuring themselves through their own stupidity is met with calls to instantly remove the thing that injured them instead of pointing and laughing

chaoko99:
I'm sick of dealing with the cleanup. I'm not allowed to turn them away for being stupid, so I get punished by anything I'm doing being interrupted. It's pretty annoying to get ripped from a scene because some jackass *REALLY* needed into some random empty room in maint!

code/game/machinery/vending.dm
```diff
@@ -1032,7 +1032,7 @@
 					/obj/item/weapon/tool/wrench = 5,/obj/item/device/analyzer = 5,/obj/item/device/t_scanner = 5,/obj/item/weapon/tool/screwdriver = 5,
 					/obj/item/device/flashlight/glowstick = 3, /obj/item/device/flashlight/glowstick/red = 3, /obj/item/device/flashlight/glowstick/blue = 3,
 					/obj/item/device/flashlight/glowstick/orange =3, /obj/item/device/flashlight/glowstick/yellow = 3)
-	contraband = list(/obj/item/weapon/weldingtool/hugetank = 2,/obj/item/clothing/gloves/fyellow = 2,)
+	contraband = list(/obj/item/weapon/weldingtool/hugetank = 2) //VORE: Removes budget insulated gloves.
```
Novacat: I would say leave this, since a vorestation edit would not be necessary. Sure ordinary yellow gloves in the hacked youtool would be wierd but eh, contraband lists have had strange, unrelated things before.

code/modules/clothing/gloves/color.dm
```diff
@@ -5,19 +5,19 @@
 	siemens_coefficient = 0
 	permeability_coefficient = 0.05
 
-/obj/item/clothing/gloves/fyellow                             //Cheap Chinese Crap
-	desc = "These gloves are cheap copies of proper insulated gloves. No way this can end badly."
-	name = "budget insulated gloves"
+/obj/item/clothing/gloves/fyellow                             //VORE:These used to be budget insuls. Not changing the path due to upstream conflicts.
```
Novacat: This does not need a //VOREStation edit. Just revert it back.

code/modules/clothing/gloves/color.dm
```diff
@@ -5,19 +5,19 @@
 	siemens_coefficient = 0
 	permeability_coefficient = 0.05
 
-/obj/item/clothing/gloves/fyellow                             //Cheap Chinese Crap
-	desc = "These gloves are cheap copies of proper insulated gloves. No way this can end badly."
-	name = "budget insulated gloves"
+/obj/item/clothing/gloves/fyellow                             //VORE:These used to be budget insuls. Not changing the path due to upstream conflicts.
+	name = "yelllow gloves"
```
Novacat: This, on the other hand, does. As well as the below line. //VOREStation Edit
Also spelling error.

code/modules/clothing/gloves/color.dm
```diff
@@ -5,19 +5,19 @@
...
-	desc = "These gloves are cheap copies of proper insulated gloves. No way this can end badly."
-	name = "budget insulated gloves"
+/obj/item/clothing/gloves/fyellow                             //VORE:These used to be budget insuls. Not changing the path due to upstream conflicts.
+	name = "yelllow gloves"
+	desc = "A pair of gloves, they don't look special in any way."
 	icon_state = "yellow"
 	siemens_coefficient = 1			//Set to a default of 1, gets overridden in initialize()
-	permeability_coefficient = 0.05
-
+	//permeability_coefficient = 0.05 //VORE: These are yellow gloves now.
```
Novacat: I am pretty sure permeability has nothing to do with shock. So, this could be reverted.

code/modules/clothing/gloves/color.dm
```diff
@@ -5,19 +5,19 @@
...
+/*Vorestation edit: Budget gloves no longer protect you at all.
 /obj/item/clothing/gloves/fyellow/Initialize()
 	. = ..()
 	//Picks a value between 0 and 1.25, in 5% increments // VOREStation edit
 	var/shock_pick = rand(0,25) // VOREStation Edit
 	siemens_coefficient = shock_pick * 0.05
-
+*/
```
Novacat: I recommend doing //VOREStation Removal Start and //VOREStation Removal End

code/modules/clothing/gloves/color.dm
chaoko99: The point is to bring them in line with other colored gloves.

Novacat:
After a bit of staff discussion, I think the best solution would be to revert it all as it was before, but change shock_pick from 0,25 to 0,15 which will significantly reduce the number of burned off hands from using budget gloves.

Cameron653:
I agree with them being removed from the vending machine, but other than that, I don't really see a point in removing the ability for them to have a randomized chance of protection, though.

Making the shock damage amplified while using them makes no sense, so that should be removed.

I'd personally say either make it so it ranges from 0% to 75% or 0% to 100% damage. It clearly warns the user that "**these are a liability and can end poorly**" if you use them, so if someone wants to gamble and take the risk, well, that's their prerogative.

Novacat:
FUCK OFF TRAVIS

PR #5896, opened 2019-09-17
Moves Planet definition to code

Novacat:
Something that would support multiple stations better, I think.


Novacat:
I have no idea why this is failing.

PR #5914, opened 2019-09-20
Adds Snow Outpost Gateway map.

Rykka-Stormheart:
99% complete, map and everything is built and working. Only problem is, well, the gateway function. Gateway won't teleport you there or back. >.<

Should fix @novacat's goof


KasparoVy:
:(

Issue #5793, opened 2019-09-03
[Bug] Lasertag Turrets have lethal setting


Brasnak:
#### Brief description of the bug
The Turret Controls for Lasertag Turrets have an On/Off setting for Lethal Mode.

#### What you expected to happen
To not find a setting labeled 'Lethal Mode' on a lasertag turret.

#### What actually happened
I found said option

#### Steps to reproduce
- Get Lasertag Turret
- Look at turret controls

#### Code Revision
- 401eb129d393d085fec77015696309b8db7c00ff 
 
#### Anything else you may wish to add:
Testing shows it doesn't actually make them lethal or change their behaviour in any way, at the very least.
![2019-09-03_18-51-36](https://user-images.githubusercontent.com/9626555/64194243-eb46d600-ce7e-11e9-9aaf-9d12f42d0385.png)

Heroman3003:
I believe difference between the settings is that 'lethal' ones will shoot even downed players?

PR #5927, opened 2019-09-21
Remove the "Lethal" setting from lasertag turrets

Rykka-Stormheart:
Since lasertag "lethal" setting does nothing, I removed it. Testing it confirmed that even with "lethal mode" on, it does not shoot you. Therefore, I'm removing the setting.


Rykka-Stormheart:
Resolves #5793

code/game/machinery/portable_turret.dm
```diff
@@ -267,8 +269,6 @@
 	data["access"] = !isLocked(user)
 	data["locked"] = locked
 	data["enabled"] = enabled
-	data["is_lethal"] = 1
```
Novacat: These need to be commented out with //VOREStation Removal

code/game/machinery/portable_turret.dm
```diff
@@ -291,15 +291,15 @@
 		var/value = text2num(href_list["value"])
 		if(href_list["command"] == "enable")
 			enabled = value
-		else if(href_list["command"] == "lethal")
```
Novacat: Same for these.

code/game/machinery/portable_turret.dm
```diff
@@ -208,6 +208,8 @@
 	check_all = TRUE
 	can_salvage = FALSE	// So you can't just twoshot a turret and get a fancy gun
 
+// Start Lasertag Turret Code
```
Novacat: This can probably be removed, since this is a polaris file, any changes would need to be marked with //vorestation edit and there is no reason to have a comment

Rykka-Stormheart: Also fixed

PR #5929, opened 2019-09-21
fix gun handle_suicide

FelixRuin:
well... idk 
now suicide with gun have sound
and you cant suicide with gun who dont do damage (like sizegun or ML3M)


SpadesNeil:
Oh good, we won't need those warning labels anymore.

Probably should send this to Polaris.

Screemonster:
As hilarious as it was for people to shrink (or grow) their heads until death, this is probably a good fix.

FelixRuin:
well, like this? https://github.com/PolarisSS13/Polaris/pull/6462 really dont know

Novacat:
Yeah that works fine. Since this got sent to polaris, merging.

PR #5941, opened 2019-09-22
Thickbelly Functionality

Novacat:
Forces says into whispers, and emotes into subtles. Does not affect received emotes or words.

### DIFF
--- a/code/modules/vore/eating/belly_obj_vr.dm
+++ b/code/modules/vore/eating/belly_obj_vr.dm
@@ -38,7 +38,7 @@
 	//Actual full digest modes
 	var/tmp/static/list/digest_modes = list(DM_HOLD,DM_DIGEST,DM_ABSORB,DM_DRAIN,DM_UNABSORB,DM_HEAL,DM_SHRINK,DM_GROW,DM_SIZE_STEAL)
 	//Digest mode addon flags
-	var/tmp/static/list/mode_flag_list = list("Numbing" = DM_FLAG_NUMBING, "Stripping" = DM_FLAG_STRIPPING, "Leave Remains" = DM_FLAG_LEAVEREMAINS, "Thick Belly" = DM_FLAG_THICKBELLY)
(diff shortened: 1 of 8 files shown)


code/modules/vore/eating/belly_obj_vr.dm
```diff
@@ -38,7 +38,7 @@
 	//Actual full digest modes
 	var/tmp/static/list/digest_modes = list(DM_HOLD,DM_DIGEST,DM_ABSORB,DM_DRAIN,DM_UNABSORB,DM_HEAL,DM_SHRINK,DM_GROW,DM_SIZE_STEAL)
 	//Digest mode addon flags
-	var/tmp/static/list/mode_flag_list = list("Numbing" = DM_FLAG_NUMBING, "Stripping" = DM_FLAG_STRIPPING, "Leave Remains" = DM_FLAG_LEAVEREMAINS)
+	var/tmp/static/list/mode_flag_list = list("Numbing" = DM_FLAG_NUMBING, "Stripping" = DM_FLAG_STRIPPING, "Leave Remains" = DM_FLAG_LEAVEREMAINS, "Thick Belly" = DM_FLAG_THICKBELLY)
```
Heroman3003: I think, while thick belly is a good description, its very un-intuitive in terms of "what does that do?" aspect of the name. Muffling or something along those lines would be much better in terms of explanation of "what this flag actually does"

Heroman3003:
Aside from that, while forcing prey into whispers/subtles specifically is good... Blindness/Reduced vision seems a bit overkill? Primarily due to the fact that in like 95% of cases, you don't see anything from inside regardless on RP level, and main reason you can see on OOC is... so that you don't have to basically stare at the black screen while waiting for a response.

ArgobargSoup:
If you can find out how to force whisper instead of say, can you also figure out how to make it force subtle instead of me?

...Also probably difficult, but if it could "muffle" speech from outside as well, that'd be neat... like, if someone speaks, and they aren't the pred, or in the belly with the prey, what the prey hears is like if you overhear a whisper from outside of whisper range.

SpadesNeil:
If possible, this trait should be something you can apply directly to bellies rather than a character trait affecting all belly organs on a character.

Like a taur's upper foregut might have less muffling capability as their lower stomach.

Heroman3003:
From looking at the code, its being done as Belly Addon already, so thats exactly what it is?

PR #5957, opened 2019-09-24
Adds vorestation wallconsoles to autolathe

Heroman3003:
Timeclock and ID restorer


Screemonster:
Honestly can't see any situation where the construction of one of these could be a problem. I'd almost consider adding them into the vending machine in tool storage.

PR #5959, opened 2019-09-24
Fixes Mapmerge Tools

Rykka-Stormheart:
Mapmerge.bat was using /_maps/, and not just /maps/, so I fixed it.

Also, word of advice, dmm2tgm is apparently all you need to mapmerge?


Jakksergal:
We're literally talking about this on discord the moment this PR went up and as I write this comment. dmm2tgm is **NOT** all you need to merge. The mapmerge process has logic in it to reduce file size and the number of definitions which will reduce bloat. 

You also didn't edit any of the other batch files to have the correct directory. You can't even use mapmerge.bat without using prepare maps.bat

Novacat:
...can you confirm this works?

Rykka-Stormheart:
> ...can you confirm this works?

Yes, I can confirm it works. Furthermore, dmm2tgm does exactly what I need it to as far as mapping goes for reducing filediff. But both preparemaps.bat and mapmerge.bat work now that they both have corrected directories. They were the only two files with _maps instead of maps.

PR #5960, opened 2019-09-24
Adds POI to Debris Field - Alien Derelict

Rykka-Stormheart:
Rather than do multiple POI's (which I still fully intend to add more!), I figured I should PR this because it's a HUGE POI. It's tested, confirmed working, the only "iffy" thing is the step_trigger/message/, but the code looks like it works, so... it's basically done!

This POI is designed to have you explore an alien, hostile ship, with an AI hellbent on killing you. Drones skitter under your feet, and your skin tingles. This ship is alien, unfamiliar, and and something instinctual is screaming at you to RUN, GET AWAY!


Heroman3003:
One thing that slightly worries me about this PoI is its size... With current state of debris field, despite it looking very open and such, actual PoI spawning space is pretty... limited, due to shuttle area being in the middle for obvious reasons, and edges having a buffer to prevent PoI cutting. On vertical axis this single PoI is more than half of the map size, so if it is spawned, it will literally occupy about half the space there is available for the PoI spawning. That should probably be properly reflected in its cost..

Rykka-Stormheart:
> One thing that slightly worries me about this PoI is its size... With current state of debris field, despite it looking very open and such, actual PoI spawning space is pretty... limited, due to shuttle area being in the middle for obvious reasons, and edges having a buffer to prevent PoI cutting. On vertical axis this single PoI is more than half of the map size, so if it is spawned, it will literally occupy about half the space there is available for the PoI spawning. That should probably be properly reflected in its cost..

...but cost doesn't determine POI spawning. You said so yourself. It's a secondary decision after size.

Heroman3003:
Yes. But what cost should measure, is how much "Worth" is that PoI, in terms of resulting loot and general amount of content. Also, bigger threats tend to also have higher cost to avoid randomizer throwing in too much difficulty.

Rykka-Stormheart:
> Yes. But what cost should measure, is how much "Worth" is that PoI, in terms of resulting loot and general amount of content. Also, bigger threats tend to also have higher cost to avoid randomizer throwing in too much difficulty.

It should be fine as-is - during testing, the first 3-4 loads didn't include this POI, even at the reduced cost. So it's rare. Exactly as intended.

PR #5963, opened 2019-09-24
Adds Belter Shuttle back!

Rykka-Stormheart:
Adds Belter Shuttle to Tether Station 3 - just south of cargo. Enterprising Xenoarchaeologists also have access to the shuttle and outpost thanks to some tweaks, so you can also use the shuttle to gather specimens.

Adds anti-particle pistol to vendor, space only. You can upgrade it for anti-particle rifle or cannon that self-recharge via Research/Science.

Adds Asteroid/Belt Outpost, with medical/break room, two dorms, atmos and solars. Solars require setup, just one-button push.


Wickedtemp:
Where's the shuttle going to be placed? How accessible would it be when a miner calls for help and is unable to pilot the shuttle?

Rykka-Stormheart:
> Where's the shuttle going to be placed? How accessible would it be when a miner calls for help and is unable to pilot the shuttle?

Shuttle will be in the mining outpost, I forgot to add screenshots, my apologies.
It should be free access for anyone, however, you have brought up a valid point - how do we recall the shuttle from orbit for Medical? 

I'll have to figure that one out.

lbnesquik:
Here's hoping all your work does get merged at some point!

KasparoVy:
Holy fuck yes! Asteroid mining FTW!

Novacat:
Still keeping this labeled as work in progress as this will be requiring mining moved to the asteroid before it can be fully finished.

KasparoVy:
Looks like it's meant to work in parallel with the surface quarry and the Underdark

Rykka-Stormheart:
> Looks like it's meant to be used in parallel with the surface quarry and the Underdark

It was, yes, but admins have vetoed us adding it. :c

SpadesNeil:
This has been up for a month without updates and now there's conflicts.

Rykka-Stormheart:
> This has been up for a month without updates and now there's conflicts.

Yes, because I'm waiting on Soft to finish her Tether rework. I was explicitly told that this would be denied being merged until after cargo and mining are moved.

I'll revisit and update it once the mining/cargo consolidation is done and live.

SpadesNeil:
Ah ok

Ketrai:
@Very-Soft can we have a progress update from you please? Seems sillly to keep this feature on hold forever.

Sebbe9123:
> @Very-Soft can we have a progress update from you please? Seems sillly to keep this feature on hold forever.

Soft has not responded, seems dumb to keep this on hiatus. @Novacat might be an idea to ping dev about fixing conflicts and then merging.

KasparoVy:
Yaaay.

Heroman3003:
1. I still would say that the base is way too close to the edge and literally hugging it.  You'd be able to see edge of the map out of the window is how close it is and there's no room around it. Why does it have to be so close anyway? I understand not wanting it in the middle, but so close to edge will just look awakward in game, especially when camera stops panning once you step into a room/close to solars.

2. Will those solars really be enough to actually power it? The energy production depends on 'time of day' (though it takes much less than a day to go for a full rotation) as well as amount of obstructions in the current direction of light. With rock so close on one hand and edge of the map on the other, will they even produce?

3. I question the need for the bypass-style substation. This place isn't even connected to the station, just run the input from solars into SMES and output into the rest of the outpost.

4. Wire under wall bad.

5. The way floor firelocks and air alarms are set up in the hallways are a good recipe for an atmos disaster should the palce be flooded/breached/etc.

6. The Pacman is landed on a wire just for it... And for some reason it goes into the local master network instead of input of the local SMES.

7. WTF is that airlock APC.

8. The default asteroids as a whole look  much less like asteroids and more like a bunch of very smooth stone bricks. Even the main one has big sections that just look way too smooth to be natural.

9. I don't know if it affects anything ingame or not, but almost all PoIs have main floor turfs displayed as white space in the mapmerge.

10. Ruined shuttle PoI has a 'functional' belter shuttle console. That, should it be powered, will actually affect the main belter shuttle.

11. Generally, PoIs seem to lack any rewards? And ones that do, have rewards in form of mob clonepods, which count as "rare big artifact" and probably shouldn't be handed out so easily, as well as a complete particle smasher.

Rykka-Stormheart:
> 8. The default asteroids as a whole look  much less like asteroids and more like a bunch of very smooth stone bricks. Even the main one has big sections that just look way too smooth to be natural.

This one in particular is fine, as the way mapgen runs, it'll automatically switch the tiles with either space, sand, ore, or something else. So while it looks smooth and "flat" in the editor, in-game it'll look better.

Rykka-Stormheart:
> 2. Will those solars really be enough to actually power it? The energy production depends on 'time of day' (though it takes much less than a day to go for a full rotation) as well as amount of obstructions in the current direction of light. With rock so close on one hand and edge of the map on the other, will they even produce?
> 3. I question the need for the bypass-style substation. This place isn't even connected to the station, just run the input from solars into SMES and output into the rest of the outpost.

For these two issues - The solars provided enough power to keep the APC from bleeding, and the outpost starts with them off. When I was originally discussing this with Soft, we felt it was a good incentive to have the miners arrive on the outpost, and simply configure solars by going to the control console and turning them on. We originally wanted to have telecomms need "pairing" first, but it was simplest to just have it pre-paired.

As for the rest of the stuff, thank you for the feedback. That's all be implemented and/or corrected, loot values for POIs have been corrected, and the reason the turfs were showing white was because they were nonexistent - /turf/simulated/mineral/ignore_mapgen/vacuum for instance wasn't a valid turf until I added it just now.

Rykka-Stormheart:
I'm closing this as Heroman's PR supersedes this and does what I was trying to do better.

PR #5964, opened 2019-09-25
Adds Alien Translocator

Novacat:


SpadesNeil:
I feel like this should not be going in the fluff items file. It should be placed elsewhere.

Novacat:
One of my to dos is to refactor Vorestation edits into their proper place, but that is out of scope for this PR. Currently translocators are located in the fluff items file, so that is where this will reside until that refactor.

PR #5970, opened 2019-09-26
Makes Surface Medical Great Again

Very-Soft:
Surface medical should be am actually useful station to play in, so this should help with that.

z3
Adds chemistry, and smole patient rooms, as well as a second monitoring computer. Also adds a small side hall from maints up into the reading room to fill the space.
z2

Adds one surgery room, one resleeving lab, a small EVA prep closet, a break room, and a supply room.


Rykka-Stormheart:
Soft I love this. <3

Rykka-Stormheart:
Also - should Med's job slots be touched at all? in case both Asteroid/Surface want to be manned at once? or would just two chemists be enough?

Very-Soft:
I don't think so. That's something to consider later, if we get a big influx of medical players. This is just intended to give medical more options, and make space medical not a necessity.

Wickedtemp:
I overall love this, the only concerns I have:

- Stocking/Prepping. There's already a decent list to check through just for the 'primary' Medbay. Adding a second one, even if its smaller, can really increase the workload of a department thats already worked more than any other, most rounds. Chemistry is a large part of this. Nearly doubled workload if a chemist wants to stock this one as well. I'd feel like I'd be asking too much as a CMO if I requested that both vendors be stocked, even though I'd likely do that myself. But doing it twice seems almost like a burden. If the chem vendors could be linked, that would be amazing. I know it has to be possible - it was done to an extent with the 'chem-vator' in Space Med. (If it can be done with paperwork storage too that would be amazing because I think if we use both Medbay's then scan reports are going to be separated and you'll need to check both Medbay's to find the one you're looking for) 

- I think it'll need to be as obvious as possible that Surface med is 'closed' if there's nobody there. I think the same should be done for Space med, too, since..well.. outside of a few small things its nearly just as viable, so people can actually choose which one they're going to be stationed at. Whether this is extended shutters over the reception area (Maybe general access but behind the reception, so you can still break in if someones dying and you haven't had Medical all shift) or just a holo 'closed' sign, etc. 

- If there's only one doctor, which tends to happen even when there are several on the manifest, I think them playing EMT will be more common, having to run back and forth to whichever Medbay they aren't currently at. This is more of an issue with the everpresent bottleneck that is the single elevator. If we could, at some point, transit smoothly between Surface and Space Med, I think it would all be utterly perfect.

Overall, I really like this. I'll actually be using Surface Med now. Any issues with it currently are small and can probably be fixed entirely as we move on.

Very-Soft:
Expecting people to make a full compliment of chems for both locations is frankly ridiculous. It's not like the demand for the chems is going to go up because of added surface medical functionality. If you make chems in one, and you want to share, there's nothing to stop you from taking some of what you made and putting it in the other.

I'm not necessarily opposed to linking the chemfridges, but that's outside of the realm of what I know how to do, and I am not sure what sort of repercussions allowing that might lead to. A discussion for another PR I think.

I've added the shutters to surface medical for now though.

Heroman3003:
My biggest concern about the expansion would be power management regarding it. Primarily the resleeving lab (which is also one big part that I think is redundant to have for smaller medical) and its personal gas cooler. Recently a lot of areas are getting plain expanded and it just adds on powernet that on default setups will soon require you to use everything you have to just sustain it.

Also, whats happening  with area bounds in the maint on lower level, one right south of stairs?

Very-Soft:
Oh the wrong maint area got selected at some point I guess! I fiz

SpadesNeil:
Fire safety policy indicates that this medical wing needs more fire shutters throughout the corridors.

Very-Soft:
> Fire safety policy indicates that this medical wing needs more fire shutters throughout the corridors, and also air sensors and vents/filters in each chamber.

> Also, whats happening with area bounds in the maint on lower level, one right south of stairs?

Fixxed the area issue, and added more firelocks air alarms and fire alarms to break up the z2 hallway, and to protect the entrance to the Z3 staircase/main area. Also shifted one of the scrubbers around, so each of the chambers in the hallway can filter itself out.

Wickedtemp:
Oh and dun forget to update the lil minimap UI. So when you look at the crew monitor console, the tracking map is accurate. Iirc its a relatively quick process but isnt done often because a lotta people didnt know how, but Hero can tell you if you dont know

PR #5972, opened 2019-09-26
Adds shutters to space medical

Very-Soft:
Now when no is in space medical you can close some shutters to make that clear.


Rykka-Stormheart:
Hey soft, you dork, this includes your Surface changes :p

PR #5975, opened 2019-09-26
Adds extra admin upload areas

Very-Soft:
What it says on the tin. For making uploadable maps with multiple different areas that need power and respect lighting.


Heroman3003:
Umm... Identation on all the areas past first seems borked?

PR #5979, opened 2019-09-27
Fixes slime miss chance

Heroman3003:
Slime baton and slime potions are no longer subject to hardcoded 25% melee weapon evasion chance slimes have

Also nerfs feral blue slimes' attack speed, was missed when nerfing other mobs


Wickedtemp:
What's the chance now?

Novacat:
0%

PR #5984, opened 2019-09-28
Makes Surface Security Useful

Very-Soft:
It would be cool if surface security was actually a useful station. This should hopefully help with that.

z3
Adds a front desk, a prisoner processing room, and a couple of extra equipment lockers, as well as a few badges and holographic warrant thingies.

Also adds some harmless wall treasure.

z2
Adds a Brig, an interrogation room, an evidence storage room, and a warden's office


Wickedtemp:
Neato!

PR #5996, opened 2019-09-28
Trades Library and Abandoned Library

Very-Soft:
Maybe the Library will get more traffic now!

z3
This library is a bit smaller, but it's comfy cozy in my opinion. There are a couple sneaky spots to smooch in when no one is looking.

z5
Makes the old library abandoned! It wouldn't take -too- much work to fix it up again if desired, but as it is, it's a dark and spooky spot that's a bit out of the way! Which is much more suitable to an abandoned library to be.


Heroman3003:
While I do think that this will make library more lively, I am slightly on the fence about the gaming table no longer having its own room. It being in a relatively far off and out-of-public place was main reason most 'games' got moved there from places like bar, to avoid all the chatter and similar disturbances

Very-Soft:
I don't think it will be much of a problem. Games and things don't usually happen all that often even though we have places for them in both the bar and the library.

Very-Soft:
If it does become a problem though, it'd be a pretty easy change to move the game room over near the Reading rooms in the north-east. I don't think that'll really be necessary though unless librarians start complaining that people are always taking up their space with games.

Rykka-Stormheart:
As I said on disc, massive approval. <3 <3 <3

PR #6030, opened 2019-10-05
Convert Phoron Bore to Lead Bore

Rykka-Stormheart:
Uses Lead Sheets instead of Phoron Sheets, charge cost reduced, added a new beam effect.

This was done because the bore EATS phoron like fucking crazy, and we rarely if ever have a use for lead. In fact, on the mining wiki, there's next-to-no use for lead. If someone has a better idea, please let me know and I can change the material, but this should drastically help it be better than just an expensive, useless paperweight.


Screemonster:
Why not just reduce the consumption though

PR #6032, opened 2019-10-05
Reduce Phoron Bore Mat Cost

Rykka-Stormheart:
Reduces MAT_COST from 500 -> 100.
80% reduction.
Fuck upgrading this pos


Novacat:
I wonder if polaris would be interested in this...

PR #6033, opened 2019-10-06
Adds manipulators to the stock parts of a Phoron Bore, making material consumption upgradable

R1f72r:
This is mostly an alternative to Rykka's pull request, as I felt that the buff was too great for it being flat.


code/modules/projectiles/guns/magnetic/bore.dm
```diff
@@ -98,7 +98,16 @@
 			capacitor = null
 			update_icon()
 			return
-
+			if(!manipulator)
+				to_chat(user, "<span class='warning'>\The [src] has no manipulator installed.</span>")
+				return
+			manipulator.forceMove(get_turf(src))
+			user.put_in_hands(manipulator)
+			user.visible_message("<span class='notice'>\The [user] unscrews \the [manipulator] from, \the [src].</span>")
```
Rykka-Stormheart: Should probs remove the comma after "from"

R1f72r: Done and done!

code/modules/projectiles/guns/magnetic/bore.dm
```diff
@@ -98,7 +98,16 @@
...
+			if(!manipulator)
+				to_chat(user, "<span class='warning'>\The [src] has no manipulator installed.</span>")
+				return
+			manipulator.forceMove(get_turf(src))
+			user.put_in_hands(manipulator)
+			user.visible_message("<span class='notice'>\The [user] unscrews \the [manipulator] from \the [src].</span>")
+			playsound(loc, 'sound/items/Screwdriver.ogg', 50, 1)
+			manipulator = null
+			update_icon()
+			return
```
Mechoid: This is going to be completely skipped due to the above return on `line 100`, you'd probably want to move this into its own `if`, perhaps with a crowbar check as on `line 90`, so that each tool removes a different component specifically.

code/modules/projectiles/guns/magnetic/bore.dm
```diff
@@ -111,6 +120,18 @@
...
 			return
+		if(istype(thing, /obj/item/weapon/stock_parts/manipulator))
+			if(manipulator)
+				to_chat(user, "<span class='warning'>\The [src] already has \a [manipulator] installed.</span>")
+				return
+			manipulator = thing
+			user.drop_from_inventory(manipulator)
+			manipulator.forceMove(src)
+			playsound(loc, 'sound/machines/click.ogg', 10,1)
+			mat_cost = mat_cost % (2*manipulator.rating)
```
Mechoid: The `mat_cost` used in the equation would need to be an `initial(mat_cost)` reference instead, as this currently means you could just take out the manipulator then re-install it to continually make it smaller.

PR #6045, opened 2019-10-11
Changes the crew transfer message to be more IC friendly.

DigitalSquirrel95:
A simple change that changes the crew transfer vote message to something that implies the crew is voting for extra work on their PDAs.  This makes the crew transfer make a bit more sense in character, because as it stands, our characters are currently coming to work not really knowing when they get off work.

Direct port of this: https://github.com/Citadel-Station-13/Citadel-Station-13-RP/pull/572

### DIFF
--- a/code/controllers/subsystems/vote.dm
+++ b/code/controllers/subsystems/vote.dm
@@ -105,7 +105,7 @@ SUBSYSTEM_DEF(vote)
 						factor = 1.4
 				choices["Initiate Crew Transfer"] = round(choices["Initiate Crew Transfer"] * factor)
 				world << "<font color='purple'>Crew Transfer Factor: [factor]</font>"
-				greatest_votes = max(choices["Initiate Crew Transfer"], choices["Continue The Round"])
+				greatest_votes = max(choices["Initiate Crew Transfer"], choices["Extend the Shift"]) //Citadel change
 
 	. = list() // Get all options with that many votes and return them in a list
 	if(greatest_votes)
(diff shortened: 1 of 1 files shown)


Heroman3003:
Imo, considering everyone gets to vote, including people dont have their PDA on, busy doing other things, ghosting or even people in lobby, that doesn't seem like something that should be counting as IC in the first place.

code/controllers/subsystems/vote.dm
```diff
@@ -105,7 +105,7 @@ SUBSYSTEM_DEF(vote)
 						factor = 1.4
 				choices["Initiate Crew Transfer"] = round(choices["Initiate Crew Transfer"] * factor)
 				world << "<font color='purple'>Crew Transfer Factor: [factor]</font>"
-				greatest_votes = max(choices["Initiate Crew Transfer"], choices["Continue The Round"])
+				greatest_votes = max(choices["Initiate Crew Transfer"], choices["Extend the Shift"]) //Citadel change
```
Novacat: These require //VOREStation Edits instead of //Citadel change

DigitalSquirrel95: Fixed

Wickedtemp:
Might lead to attempted vote swaying ICly, too

SpadesNeil:
I actually really like this.

KasparoVy:
Digging the extra flavour/muh immersion, nice touch.

PR #6056, opened 2019-10-14
Adds a small cargo office to mining, and adds a glorious cargo pipeline

Very-Soft:
This started as a project to add a small cargo office to mining. 

z1

But then I decided that it would be fun if there was a delivery pipe between the two.

Also, forgot to mention: Adds some firelocks in places where firelocks ought to be. Also changes one pointless engineering maintenance airlock to a regular maintenance airlock on Z3

---
Added a connection to research on the mining-cargo pipeline, so materials can be delivered to research by mail from mining. It ONLY goes to 'Research' or up to cargo though. So if you want to mail things around properly you still need to go up to the actual mail room.

Also adds a little wrapping paper and a destination tagger to mining. (Along with a helpful instructive note)

Also fixxes the fact that the RnD disposals bin wasn't connected to the pipes. The 'Research' tag now delivers to this bin. Also tidied up some floor decorations at the xenobio entrance.

### DIFF
--- a/maps/tether/tether-01-surface1.dmm
+++ b/maps/tether/tether-01-surface1.dmm
@@ -1657,6 +1657,7 @@
 /obj/effect/floor_decal/corner/brown/border{
 	dir = 8
 	},
+/obj/vehicle/train/engine,
 /turf/simulated/floor/tiled,
 /area/tether/surfacebase/mining_main/storage)
 "acV" = (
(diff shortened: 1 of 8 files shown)


Very-Soft:
the pipe transports things between the surface and cargo pretty fast. I clock it at around 16 seconds on the test server, but that goes a little faster than the live server tends to.

TheFurryFeline:
Is there a conveyor switch, by the way? Just noticed there doesn't seem to be one connected to the conveyor line.

Very-Soft:
There's two! One's in the cargo office, and one's in the space cargo warehouse. The conveyor goes out! So if something arrives you can flip either switch without having to go into the warehouse and have it scoot its way out into the lobby.

TheFurryFeline:
Oh, I was thinking that this thing was being controlled from right in
mining rather than have someone all the way up at the top.

Very-Soft:
It can be controlled from within mining, the switch is just in the cargo office. I just imagined the switch being there would be more generally useful than putting it in the warehouse, Since if someone ships a crate down for you and you have to go all the way into the warehouse to flip the switch, then you'd sort of be better off just grabbing the crate, rather than flipping the switch and waiting. 

The way it is, you can do that from the place people are more likely to already be.

TheFurryFeline:
Hmm, a'ight

PR #6060, opened 2019-10-15
Adds 'Devourable" and 'feeding' prefs

Poojawa:
## About the Pull Request

This PR adds 'Devourable' and 'Feeding' Preferences at the player level. Reducing the potential for predator only  and prey only characters from pref break at the mechanical level, rather than requiring an ahelp and potential excrement typhoons from happening. 

Toggled off, the user will get a message about a failed attempt of feeding or trying to devour someone due to the settings, which if someone mistakenly forgot that was a thing, a simple looc will fix that without spamming the targets. They all default to on as that was previous behaviour.

Currently Feeding will check if the predator can be force fed, then if the prey is devourable, then if the prey can be force fed to others. if any of these fail the check, it returns and does nothing.

It also tweaks the Vore UI to be a lot more visually responsive to what mode your toggles are currently set to, 

It also reduces the amount of magic numbers used in the return procs, since they return true/false anyway. just a good readability standard.

## Why it's Good for the Game

Helps to reduce potential administrative workload of ahelps of pref-break as the person in question is able to simply deny certain situations on the mob level. These stats also work for corpses as digestible had. 

I found it really strange that the "Off" toggles were green prompted, which is confusing because generally green means 'yes' or 'on'.

## Changelog

:cl: Poojawa
add: Added Devourable and Feeding prefs
fix: Adjusted the Vore UI to be more clear in what mode a toggle is set to.
/:cl:


Screemonster:
I like this. Would it be possible to have the same "currently: on/off" for drop-nom prefs too? It really needs a better name than "drop-nom" given that it happens with teleporters and shadekin way more often than it does with dropping, and it's a catch-all for "suddenly eat/be eaten without warning".

SpadesNeil:
"Sudden vore" could be the term.

Drop noms, bump noms, maybe other mechanics in the future.

Wickedtemp:
Spontaneous vore.

Also 10/10, quite a few players are probably gonna appreciate this a lot.

Poojawa:
Yup, Easily done to make the drop nom stuff

Poojawa:
I'm rolling the maybe fixes for #6069 into this just because I'm lazy more than anything

PR #6061, opened 2019-10-15
Moves the chapel to Z2

Very-Soft:
chapel woah

Moves the chapel Z2. There is a stairwell in the top left that leads up to the main halls, just north of the bar. The passage to the chapel there is a bit dirty, but is well lit and tiled.

spess

The launchy part of burial was left in space because it doesn't make much sense to launch caskets at the outer wall.

---

The purpose of this change is to try to put the chapel in a place people are more likely to actually visit. 

It's a rare thing for anyone to go up to z5 for the purpose of hanging out unless they went there with plans already.

The location of the new chapel puts it on a relatively active maint path, and adds an additional stairway that leads almost right to it near the bar.

I think this will lead to more general cross traffic going through and by the chapel, which will give more people more chances to happen across chaplain gimmicks, which will ideally lead to more roleplay.

And while the new chapel is more conveniently located, it is still comfily secluded in my opinion. 

I think the new location will make life more lively for chaplains, so maybe we'll actually get some chaplain rp happening occasionally.


TheDavestDave:
Wasn't a complaint about an earlier chapel being that the only people going in were maint divers Passing by?

Very-Soft:
> Wasn't a complaint about an earlier chapel being that the only people going in were maint divers Passing by?

I think I saw that complaint one time. And in general, I'd personally prefer that there's -some- idle traffic passing through and able to see that there's someone in there doing stuff, than what we currently have, where the chapel is out of the way enough that just, no one even looks at it.

The Chaplain is primarily an RP job. The Chapel is the Chaplain's work station. Having the occasional person passing through is an opportunity to entice people with whatever gimmick you have going.

If you want to do private lewd RP, the office has tintable windows.

PR #6068, opened 2019-10-16
Multiple map fixes, Patch 10/15 Fixes

Rykka-Stormheart:
Fixes UD POI's:
Give generation subsytem 50 additional points to play with, reduces some POI costs. - These fixes were made after observing two shifts with repeatedly only one-two POIs for the entire map.
Abandoned Oupost - corrects Infection Drones to use correct faction. Also fixes turret to use correct faction. Removes unnecessary outdoor lamp.
Mimicry - Changes "crate"s to "old crates" so as to make guessing harder.

Regular Map Fixes:
Removes the random space tile that's visible through walls on Surface 1.
Fix the new Virology Entrance decals so they line up.

Derelict POI Fixes:
Adds a new area, forced ambience for that area, and two new ambience files exclusively for alien areas.
Fixes all APCs to alarm_hidden 1
Adds invisible admin power generators that cannot be interacted with, but will provide power.


Rykka-Stormheart:
Just playtested it and confirmed everything is working. Ambience works, POIs work, etc. Everything is gewd. <3

SpadesNeil:
I couldn't help but notice the lack of fire extinguishers throughout Xenoarch. Mind adding a bunch of fire extinguisher cabinets? In like, every room. :U 

Recent events have made me really glad xenoarch is no longer near the station.

PR #6079, opened 2019-10-19
Redistributes maint wealth

Very-Soft:
This makes there be less trash piles in maint. This is intended to go with a buff to trash piles so they'll ideally give you less straight up garbo. 

There were 62 before. Now there's 12.

None of the trash piles are in any of the rooms, they are only in maint tunnels specifically.

### DIFF
--- a/maps/tether/tether-01-surface1.dmm
+++ b/maps/tether/tether-01-surface1.dmm
@@ -2500,24 +2500,6 @@
 /turf/simulated/floor/plating,
 /area/maintenance/lower/trash_pit)
 "ael" = (
-/obj/effect/floor_decal/techfloor,
-/obj/structure/railing{
-	dir = 1
-	},
-/obj/effect/decal/cleanable/dirt,
-/obj/random/trash_pile,
-/turf/simulated/floor/plating,
-/area/maintenance/lower/trash_pit)
-"aem" = (
-/obj/effect/floor_decal/techfloor,
-/obj/structure/railing{
-	dir = 1
-	},
-/obj/effect/decal/cleanable/dirt,
-/obj/machinery/light/small,
-/turf/simulated/floor/plating,
-/area/maintenance/lower/trash_pit)
-"aen" = (
 /obj/effect/floor_decal/techfloor{
 	dir = 4
 	},
@@ -2531,7 +2513,23 @@
 	icon_state = "alarm0";
 	pixel_x = 24
 	},
-/obj/random/trash_pile,
+/turf/simulated/floor/plating,
+/area/maintenance/lower/trash_pit)
+"aem" = (
+/obj/effect/floor_decal/techfloor,
+/obj/structure/railing{
+	dir = 1
+	},
+/obj/effect/decal/cleanable/dirt,
+/obj/machinery/light/small,
+/turf/simulated/floor/plating,
+/area/maintenance/lower/trash_pit)
+"aen" = (
+/obj/effect/floor_decal/techfloor,
+/obj/structure/railing{
+	dir = 1
+	},
+/obj/effect/decal/cleanable/dirt,
 /turf/simulated/floor/plating,
 /area/maintenance/lower/trash_pit)
 "aeo" = (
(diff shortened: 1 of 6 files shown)


Heroman3003:
I don't like the change that much, even on assumption that the pool is buffed. It removes any reason to properly explore and traverse maint entirely, while trivializing the search for items through piles. Its good that trash piles are spread throughout general areas of just maint passageways, because that heavily increases chances of people bumping into other trash divers during their searches and who knows, maybe good stuff comes out of it. And argument of trash piles being scene deterrants because fear of people coming in to rummage through them is pretty moot, considering any reasonable scene would be happening in maintenance side-rooms anyway, which are all over the place and easily accessible. And do not have any trash piles in them as well already.

lbnesquik:
This just lowers the amount of trash piles? I don't see how this redistributes things.

SpadesNeil:
>This just lowers the amount of trash piles? I don't see how this redistributes things.

Because the trash piles produce more stuff now.

Personally I don't like it anyway. I don't want there to be just 12 piles. Security can basically just loiter around and guard them as validhunting assholes--a situation that I wouldn't put past them considering recent behavior. Let's not do this.

Sebbe9123:
> > This just lowers the amount of trash piles? I don't see how this redistributes things.
> 
> Because the trash piles produce more stuff now.
> 
> Personally I don't like it anyway. I don't want there to be just 12 piles. Security can basically just loiter around and guard them as validhunting assholes--a situation that I wouldn't put past them considering recent behavior. Let's not do this.

Wait, rhat sound more like something admins should crack down on?

SpadesNeil:
>Wait, rhat sound more like something admins should crack down on?

If you mean overall, yes, we are already doing that.

However, admins attempting to crack down on unwanted behavior that has been made easy by code changes is something that should be fixed with a code change, not admin legwork. It's tedious for admins and unfair for players who get punished over it not realizing their behavior is in any way undesired. Guarding hot spots of contraband sounds like a totally reasonable thing to do from the perspective of a security player, does it not? It seems perfectly sane and normal. But we don't want them to do that. Let's not make it easier to do.

Contrary to blowing up a welder tank, which is mechanically easy to do, such an act has consequences within the game without an admin's involvement, so it's not as "easy" in the same context as what I just talked about. Assuming the explosion doesn't injure you directly, other players will be out to get you. An act of mindless destruction is not viewed by the average player to be sane or normal.

Very-Soft:
> > This just lowers the amount of trash piles? I don't see how this redistributes things.
> 
> Because the trash piles produce more stuff now.
> 
> Personally I don't like it anyway. I don't want there to be just 12 piles. Security can basically just loiter around and guard them as validhunting assholes--a situation that I wouldn't put past them considering recent behavior. Let's not do this.

It's already generally against the rules for security to go hunting for people in maint without good reason to begin with. They don't do what you're worried about because that problem has already generally been dealt with. There's plenty of great places for it already which have spots where you can loot 3 trash piles at once, and spots where you can see even more. 

My general problem with there being SO MANY trashpiles and them being -absolutely everywhere- is that they reduce maint's general usefulness for nefarious doings. Everywhere that has a trash pile that isn't overly tedious to get is a high traffic area. That ALSO puts all the side rooms that are near them at a significantly higher risk of being passively searched. You're already in the area on your usual trash pile path, why not? 

Heroman himself was complimenting the bit of maint in the northwest of Z6 that's a long hallway with a bunch of side rooms, because it's a good spot for people to drag you off to for scenes. And we talked about how it's one of the only spots that people ever make little illegal goods shops and things like that. Heroman's on the other side of the argument on this in general, but I thought it was a little telling that he brought up that particular bit of maint. 

Because it has absolutely no trash piles in it. Plenty of randomspawn loot, but not a single trash pile. Which puts it outside the normal maint loot run. Which makes it a great place to do crimes or take fussy prey.

I'd like more of maint to be good for that, and with the present distribution of trashpiles, it can't be without adding more maint, and just deliberately not adding trash piles there again.

And the term redistribution was used because the severe limiting of the trash piles was supposed to come with a buff to the trash piles loot table, to take out a lot of the stuff that just no one ever wants.
So, higher chance to get something that might be desirable, but less chances.

Issue #4747, opened 2019-02-04
Supply request event is broke in many ways and these are problems it has:


Heroman3003:
#### Brief description of the issue
Supply event is in many ways broken. It should not be manually enabled by admins until all issues it has are ironed away. This is list of said issues.

#### Food
Supply event does not care if what type of food it desires. In fact, anything made in microwave is possible to appear on the food part of the list. Result? Stuff requiring unobtainable ingridients (xenomeat, ectoplasm, grubmeat before grubs happened, carp meat) included. Oh, and it seems like 'ruined virus dish' may appear there too, as its made in microwave. Fun. And good luck getting slime jelly dishes without xenobiologists, it wont check if there are any. Or finding out without knowing, as it wont clarify, what type of jelly it wants in, lets say, jelly burger. Both berry jelly and slime jelly burgers are named same and will appear same, but for the request they will be different.
And it checks for civilian jobs. Which means even with 0 chefs, you may get that request.

#### Research
No miners? Well fuck you, as this category picks random recipe from research even if it requires materials you do not have. In addition to that, there is one item that you cannot get consistently enough levels to acquire, and its Decloner. It requires combat 8 which needs either explorer RNG finds, dart gun from maint or HoS's NSFW deconstructed.
Check is also done in a way, so that items will appear even if its just explorers/pathfinder in research, who dont even have access there.

#### Chemistry
Well, here we run into same problem as previous two. Is there a medicine with ingridients you cannot normally access? Well, not our problem, we need you to provide anyway. Carpotoxin, Spidertoxin and Grubtoxin are needed for some of possible medicines and first two are not normally accessible, with third relying on grubs having happened before. Myelamine can at least be ordered from cargo, but that one is no longer chemistry request. In addition to that, NIF nanites as a reagent are also possible to appear, despite only thing to do with chemistry being that they are a reagent, so will appear even with no science present.
Also, Agent A, Agent B and Agent C, which all have nothing to do with chemistry and rely entirely on xenobiology, appear in the pool. Yes, check for presense of xenobio does NOT happen.
And to top off the cake in chemistry, Healing Nanites can be one of requested reagents. That reagent is a strong medicine that only appears in hypospray of mercenary cyborg. There is no other way in game to acquire it, and no way to acquire cyborg itself. Have fun procuring those.
Ah, and of course the fact that it does not require specifically chemists to appear, but any medical. Including lone psychiatrist.

#### Mining
[...]

PR #6094, opened 2019-10-22
Rarer events, more event variety

Novacat:
- Reduces the chances of a mundane and moderate event firing by 50%
- Reduces the chances of a major event firing by 75%
- Re-adds most Polaris events. Notables include carp, spider infestations, rogue drones, vines, and meteors.
- Grubs upped to major event
- Station-spawned spiderlings can no longer grow up to be nurse spiders.
- Fixes and re-enables the Ian storm event (for adminbus)


KasparoVy:
Aww. RIP events

chaoko99:
I'm not sure I like the idea of me tears, seeing as due to our lack of a shield diffusal system, we're basically left to either die or be unable to eject the Crystal

SpadesNeil:
Ian storm should be admin abuse only. It should not be in the normal rotation.

Wickedtemp:
And yeah, meteor tends to just be "lol youre dead evac" unless the handful of engineers are:
- Alive and not dead and not ghosted or in a scene hoping nobody notices
- Competent enough to know how to fix entire floors smoothly
- Has the resources to do so
- Feels like putting up with this shit when the tram is being called in a couple hours anyway.

Novacat:
For now, I disabled the more severe meteor wave event, but I still have the lesser 'meteor shower' event which hopefully should not cause too much damage.

Sebbe9123:
No nurse spider? aww

Heroman3003:
Meteor showers are horrible and really shouldn't be anywhere outside major events, and even then, it should be weaker and less likely...

Why are carps affected by engineers? It really should be security onyl thing
Same for drones, but why is scaling so overdone? They are basically much more dangerous version of carps.
Space dust no longer being one-shot is also pretty ?! moment since only reason it doesn't cause trouble is because it happens once so not that much chance of actually causing severe damage.

Spider infestation as moderate event sounds like a bad idea as well. For obvious reasons.

Vines are bad idea on account of the fact that they are broke mechanically and aren't functioning fully as they should, and with event added into mix, not sure if they will work at all, would need testing.

Grubs being rarer than spiders? Thats stupid and ridiculous. Spiders are way more dangerous.

Heroman3003:
Overall, I think these changes are making the issues with events even worse. Because it makes them even more impactful and impossible to ignore, which will make life all the more annoying for anyone playing security or engineering. No scening on job allowed will come up much more seriously because now events are actually big deal and a 'must-be-dealt-with' situations. While before it was about 95% ignorable and/or optional. Inconviniences, not direct threats requiring you to drop your everything and deal with them.

Novacat:
Feedback taken into account, but, a few things to reply.

- The whole point of this PR is to have events rarer but more impactful.

- Solargrubs _are_ more dangerous than spiders at present. Yes, Spiders have more HP and do more damage, but they also do not affect station operations. You can effectively close off a room with a spider and not have to worry about it, while solargrub power draining forces you to deal with it. Solargrubs can also crawl into engine spaces and cause a tesloose/delam by siphoning power from the engine room SMES, which they actually have before.

I was around when spider infestations were still a thing, and the only time they were ever an issue at all is when a nurse spider spawned and managed to grab a few mice and made hundreds of spiderlings from that. That is why nurse spider spawns have been disabled.

- Moderate level dust was only a big issue if it breached the engine room, which, it should not be able to now. But I took your feedback into account and made it rarer and non-repeatable.

- Moderate meteor shower is weaker than major (which is why I disabled the major one), and I made it the rarest moderate event.

Heroman3003:
Even after the changes, I still stand by the fact that regardless of rarity increases, making events more 'impactful' will just lead to more "no scenes allowed when onduty" mentality and interruptions nobody asked for. If anything, increased 'rarity' will make the it even worse, because it will be even less predictable when event is coming than before, making it much harder to even try and plan for a good time.

As for the post-changes status quo. Spiders being able to appear even when there is absolutely no sec present is dumb still.

chaoko99:
Just for the record: Meteors are definitely one of those "haha fuck your scenes" events for engineering. I really, really suggest against it considering.

chaoko99:
You know how spiders ventcrawl? They could totally just +appear+ in dorms.

Novacat:
Spider ventcrawl has special code to prevent them from going into dorms, from what I could see.

SpadesNeil:
What ever happened to the old cargo special shipment events? You know, like the ones /tg/ does. I loved those.

Let Cargo decide whether or not to ship a shuttle full of spiders to the station in exchange for cool shit on board, or we could have several varieties of these that could trigger like once a shift. Most of them would be dangerous, but all would be _optional_, and it would help fill in slots to make other events less frequent.

Heroman3003:
We do have one cargo event, but I analyzed all the issues it has that makes it annoyingly stupid in #4747, and nothing was fixed since (and doing so would be quite a bit of work), and its not event you're talking about. So probably not a thing on this codebase. In addition to that, since I made that issue, all the issues only got worse with anomatech designs in research, new materials, new chems, etc.

Sebbe9123:
The only event I would be anything against is the meteor ones, but even then those are more or less negliable too other than that the events are minor, not a real threat to anyone unless they seek it out or decide to stay where the danger is.
But I guess Fun really isn't allowed on the server as we can't expect people to use common sense to run away from danger instead of trying to fight it alone.

SpadesNeil:
The cargo event you described is not the one I am thinking of, as we have never had it on a Bay downstream (of which Polaris is) because Bay is no-fun-allowed station.

Issue #6087, opened 2019-10-20
Disposals doesn't seem to actually deliver things to the trash pit reliably


Very-Soft:
#### Brief description of the bug
there appears to be a sorting junction blocking the path to the trash pit now when it comes in the wrong way.

#### What you expected to happen
trash go trash pit

#### What actually happened
get stuck in endless loop

#### Steps to reproduce
- be on the bridge
- get camera bug, configure it, disposal it, and watch
 
#### Anything else you may wish to add:
-
![disposals is messed](https://user-images.githubusercontent.com/24854483/67165415-8bcf5780-f352-11e9-9723-c7bb8296a327.png)

I don't know if there's a spot earler on where the tubes end up turned around and going the wrong way or that, but I noticed that when you hit that sorter, you just immediately start going the opposite direction, and end up turned around and back into the same loop again.

Also I'd fix this right now but I gotta work and I don't know how to make a separate PR not conflict with the map edits I already have up and waiting to merge.

MrSign:
#6025 

Already spoke about a familiar issue, dunno if it got fixed or not.

Issue #6025, opened 2019-10-04
Disposal pipes delivery system at midpoint is broken.


MrSign:
#### Brief description of the bug
Sending stuff from the asteroid to the surface results in stuff being stuck in an endless loop in the midsection.

#### What you expected to happen
Stuff being able to be sent through disposals from the asteroid to the surface.

#### What actually happened
Around and around they go, for all eternity.

#### Steps to reproduce
-Be a maintenance drone.
-Get yourself to a disposal unit on the asteroid. For example; a disposal unit on asteroid 1.
-Set your tag to atmospherics.
-Climb into disposals and wait for it to activate.
-"WHEEEEEEEEEEEEE!"
-Get caught in the loop de loop of the midpoint.
- https://youtu.be/gTZ91YtKLN4

#### Code Revision
384cb4605e5215564230d2f21e4ddb488e4cca2a 
 
#### Anything else you may wish to add:
N/A

Rykka-Stormheart:
Oh geez, I thought I fixed that, I must've fucked it up. aaaaaaaaaaaaaaaaaa

TheFurryFeline:
This issue's been fixed on the master branch, it seems.

PR #6099, opened 2019-10-22
Fix Disposals Permanent Loop! Also Fix Library Windoor.

Rykka-Stormheart:
Removed redundant Sorting Junction on S2, replace with regular junction, no more permanent loop! rejoice!

Also fixes the windoor.

### DIFF
--- a/maps/tether/tether-02-surface2.dmm
+++ b/maps/tether/tether-02-surface2.dmm
@@ -3077,11 +3077,18 @@
 /turf/simulated/open,
 /area/tether/surfacebase/atrium_two)
 "agy" = (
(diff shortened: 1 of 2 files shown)


Rykka-Stormheart:
Confirmed in testing that:
Delivery of packages from Space to ANY Surface level works as intended.
Trash gets delivered to Cargo as normal, with the exception of Surface 1 - that trash goes straight to the Pit.

I'll take the tradeoff over the broken loop causing lag on the server atm.

Rykka-Stormheart:
Fixes #6087 and #6025

KasparoVy:
Aw, RIP fix.

PR #6110, opened 2019-10-24
[DNM] Relocate UD Entrance to SE, fix several POI's

Rykka-Stormheart:
Still having issues with teleporters, will fix next.


KasparoVy:
OH hell yeah. Convenient

PR #6121, opened 2019-10-27
Fix #3521 - Loadout Implants Properly go in Mob, not on Floor

### LINKED ISSUE #3521: Tracking Implants spawn on the floor
Selecting a tracking implant from the loadout, spawning in drops the implant on the floor. (But, your character still 'has it'.

For there not to be an implant on the floor when you spawn in with a tracking implant.

#### What actually happened
Upon spawning in, there's a lil' implant dotting the floor.

#### Steps to reproduce
-Select 'tracking implant' from the loadout menu
-Join the round.
-Look down at your feet and discover your implant has escaped your body!

#### Code Revision
7be6c1a479cb498e22d6a00104ae8b8fa43add79 
 
trackingimplants

KasparoVy:
Fixes #3521

The main cause is the 'exploitable' check from `code/game/jobs/job_controller.dm` because the proc `amend_exploitable` from `code/modules/mob/mob.dm` spawns a duplicate of the implant at the mob's feet.
This was less apparent because for all the other implants (dud, neural, etc.) the `Initialize()` behaviours deleted them so they wouldn't be visible on the ground, so the indication they had the same issue was that they weren't present in the appropriate parent organ's `contents` or `implants` lists.

Tertiary issues were...
- The `implant_loadout` proc in `code/game/objects/items/weapons/implants/implant.dm` not inserting the implant in the correct location (was passing an organ instead of a zone define like BP_HEAD, which has a string value of "head".)
- There were unnecessary post-spawn behaviours for dud and neural fluff implants.

What now happens:
- The exploitable check now occurs in the `spawn_item` proc from `code/modules/client/preference_setup/loadout/loadout.dm`.
- The loadout implants are now inserted into the correct locations within a mob.
- There won't be implants created on the floor anymore.

-------------------

I verified this by spawning in with every implant, checking the parent organs (chest & head), referring to the `contents` & `implants` lists and seeing that for the head the brain was "assisted". No implants were on the floor.

### DIFF
--- a/code/modules/client/preference_setup/loadout/loadout.dm
+++ b/code/modules/client/preference_setup/loadout/loadout.dm
@@ -70,10 +70,10 @@ var/list/gear_datums = list()
 			continue
 		if(max_cost && G.cost > max_cost)
 			continue
-		if(G.ckeywhitelist && !(preference_mob.ckey in G.ckeywhitelist)) //Vorestation Edit 
-			continue  //Vorestation Edit 
-		if(G.character_name && !(preference_mob.client.prefs.real_name in G.character_name))  //Vorestation Edit 
-			continue  //Vorestation Edit 
+		if(G.ckeywhitelist && !(preference_mob.ckey in G.ckeywhitelist)) //Vorestation Edit
+			continue  //Vorestation Edit
+		if(G.character_name && !(preference_mob.client.prefs.real_name in G.character_name))  //Vorestation Edit
+			continue  //Vorestation Edit
 		. += gear_name
 
 /datum/category_item/player_setup_item/loadout/sanitize_character()
@@ -271,4 +271,7 @@ var/list/gear_datums = list()
 	var/item = new gd.path(gd.location)
 	for(var/datum/gear_tweak/gt in gear_tweaks)
 		gt.tweak_item(item, metadata["[gt]"])
+	var/mob/M = location
+	if(istype(M) && exploitable) //Update exploitable info records for the mob without creating a duplicate object at their feet.
+		M.amend_exploitable(item)
 	return item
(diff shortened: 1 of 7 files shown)


KasparoVy:
With the release of a bunch of new implants from the medical expansion I felt extra compelled to figure this out. Wasted a lot of time barking up the wrong trees but eventually got to the bottom of it

Not sure if it's an issue with Polaris either as I couldn't find an issue report there

Novacat:
This should probably be sent to Polaris.

KasparoVy:
Will send to Polaris and post a comment from there referring to this PR once it's up (later 2019/10/27)

code/modules/client/preference_setup/loadout/loadout.dm
```diff
@@ -271,4 +271,7 @@ var/list/gear_datums = list()
 	var/item = new gd.path(gd.location)
 	for(var/datum/gear_tweak/gt in gear_tweaks)
 		gt.tweak_item(item, metadata["[gt]"])
+	var/mob/M = location
+	if(istype(M) && exploitable) //Update exploitable info records for the mob without creating a duplicate object at their feet.
+		M.amend_exploitable(item)
```
KasparoVy: Force-push was to change the argument from `(src)` to `(item)`.

KasparoVy:
Posted to Polaris as requested @Novacat, does that mean I should close this one and allow it to trickle down from them later?

Novacat:
For now, just keep it open in case we decide to do a patch before polaris merges it.

KasparoVy:
Sent to polaris as https://github.com/PolarisSS13/Polaris/pull/6500 and bounced back in #6135

PR #6130, opened 2019-10-29
turns cryo player searching for carbon mobs only

### LINKED ISSUE #6128: Cryo Player verb is pretty lame
When admins need to cryo naughty AFKers or that, we use the Cryo Player verb.

Problem is, it lists -every mob- seemingly in reverse order of the order they were spawned. 

I'd VASTLY prefer it was at least alphabetical for easier scanning, AND IDEALLY I'd prefer it to only list carbonmobs, or characters on the manifest.

Poojawa:
We really couldn't care less to despawn ghosts, would we?

Fixes #6128

### DIFF
--- a/code/modules/admin/topic.dm
+++ b/code/modules/admin/topic.dm
@@ -1896,7 +1896,7 @@
 	else if(href_list["cryoplayer"])
 		if(!check_rights(R_ADMIN))	return
 
-		var/mob/M = locate(href_list["cryoplayer"])
+		var/mob/living/carbon/M = locate(href_list["cryoplayer"])
 		if(!istype(M))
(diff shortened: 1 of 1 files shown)


code/modules/admin/topic.dm
```diff
@@ -1896,7 +1896,7 @@
 	else if(href_list["cryoplayer"])
 		if(!check_rights(R_ADMIN))	return
 
-		var/mob/M = locate(href_list["cryoplayer"])
+		var/mob/living/carbon/M = locate(href_list["cryoplayer"])
```
Novacat: Will either need a //VOREStation edit, or sent to Polaris. Probably sent to Polaris.

PR #6136, opened 2019-10-30
Switches to an alternative system for gateway map use.

Heroman3003:
Disables random submap selection and instead uses map datum's "select one map" function. Works exactly same way lateload submaps do, with exception of the fact that only one category is taken from the list. Also moves snow_outpost into new location. Snow Outpost upon testing seems unplayable as of now, so leaving it disabled, but for example, general framework for it is implemented. New gateway maps can be loaded like that from now on.


SpadesNeil:
So the framework is in, now we just need some maps.

Heroman3003:
Pretty much. I only tested the most recently added map, but it seems like due to lack of ability to test it (as result of lack of framework before now) it is in a rather... sorry state. Other maps are not even correct size I believe?

Novacat:
Heroman was interested in collecting money from the old gateway map contest (which was never collected) by getting randomized gateways functioning again. Having Ace review this for that reason.

SpadesNeil:
The contest itself was cancelled because nobody entered, but the bounty I was offering from my own money is still valid.

If you can get the gateway and at least three maps working for it, I will pay you my bounty just for getting this shit up and running.

Rykka-Stormheart:
> The contest itself was cancelled because nobody entered, so the donor funded reward is void, but the bounty I was offering from my own money is still valid.
> 
> If you can get the gateway and at least three maps working for it, I will pay you my bounty just for getting this shit up and running.

Can I split the bonus if I make the gateway maps and he does the gateway system?

Heroman3003:
Fixed up 3 older maps. There was fourth I could probably implement easily but it seems unfinished and the rest are either too big to comfortably adapt for our mapsize or so deprecated that fixing all the pathing conflicts is better left to its own project.

Right now Zoo is technically functional, but is simply too big, and Snow Outpost destination needs some proper playtesting.

SpadesNeil:
I want to see how this works in-game, but once it's live on the server and any severe bugs are worked out, contact me regarding the $100 bounty I was going to issue in the original contest. Payment will be done through Paypal when the time comes.

PR #6137, opened 2019-10-30
Earlyport of Fan Retrosprite

Rykka-Stormheart:
https://github.com/PolarisSS13/Polaris/pull/6504


Tename:
Gotta say I dont really like these sprites, I never really thought of this server following the whole retro design(Example: NIFS, PDA sprites, and many things in medical) I don't really think this would look good in many areas.

PR #6145, opened 2019-10-31
A bunch of little fixes and event features.

Verkister:
-Adds simplemob radio. A "headset" that can be used and listened on by simplemobs. Adminspawn the item, ~~shove it into your mob's contents, and link a reference to the item into the mob's mob_radio var.~~ Additional keys can be added same way as regular headsets before "implanting" into the mob.

-Gives the mob radio a much easier way of implementation. (Just take the item and boop the mob with it to do the entire setup. Booping a radio enabled mob with another will eject the mob's current radio without replacing it so you'll need to boop the emptied mob again to apply the new radio)

-Adds leap/pounce verb for simplemobs. Essentially a copy of how the dogborg pounce works.

-Couple fixes to borg/animal pounces. (makes the pounce "animation" reset to default_y instead of hardcoded crap. Also disables spin on simplemobs when pouncing. Unfun maybe, but that var can always be toggled ingame to get the sick ninja moves)

-Adds adminbus switch to enable abuse mode on trasheater. (only skips the whitelist, blacklist still applies)

-Fixes a bunch of icon updates on the vorecode. Humanmobs no longer get forced canary runtiming mob icon updates, and the existing mob icon updates are no longer simplemob exclusive (and thus broken for pAIs and such)

-Adds some additional stability to the borg sleeper clean cycle.

-Fixes unequipped sleeperbellies never ceasing to process and update, which had been causing runtime spam and such when undeploying an occupied sleeperbelly.

-Tweaks the gurgle gainz received from digested body parts.

-Earlyports a fix for mobs wandering off when grabbed.

-Earlyports a fix for adding verbs to mobs via varedit. (Was utterly broken for anything that wasn't STRICTLY a root type of human, robot, or AI. Also now works for simplemobs as well.)

-Earlyports a fix for unconscious borgs never dying.


Verkister:
Couple more incoming

code/modules/ai/ai_holder_movement.dm
```diff
@@ -142,7 +142,7 @@
 	if(isturf(holder.loc) && can_act())
 		wander_delay--
 		if(wander_delay <= 0)
-			if(!wander_when_pulled && holder.pulledby)
+			if(!wander_when_pulled && (holder.pulledby | holder.grabbed_by.len))
```
KasparoVy: Out of curiosity, why `|` instead of `||` here?

Verkister: Heck just mixing up my ors again

KasparoVy: All good, something's gotta slip through once in a while.

Issue #6139, opened 2019-10-30
Surface Medbay Mapping Issues


Vorrarkul:
#### Brief description of the bug
The surface medbay shutters cannot be opened from the inside; the only button that controls them is behind the reception desk in FRONT of the shutters.

In addition, the surface medbay's lights are all off by default.

#### What you expected to happen
When spawning at the start of the round, to be able to leave my department or see inside it.

#### What actually happened
I was trapped inside until I crawled over the chemistry lab's counter and had to manually turn on the lights for any room I entered.

#### Steps to reproduce
- Spawn at the start of the round as medical, end up in surface medbay.
- Discover nearly all the lights are off.
- Find yourself trapped on the wrong side of the reception shutters.
- Open the chemistry shutters and crawl over the counter in order to escape.

#### Code Revision
- 897b903acb723986d368ccd472e44cac63d66f54
 
#### Anything else you may wish to add:
- NA

Rykka-Stormheart:
Lights being off on Surface Medbay is intentional but being locked in is not.

Very-Soft:
There is a button to open the shutters on the inside, though it's a bit further back than it really ought to be. It's something I'd intended to change. 

And yeah. The lights start off in all the side rooms outside the main hall. It's hardly a trouble to turn them on when you come into a room.

Vorrarkul:
My apologies, I didn't realize it was intentional. I'm used to all the lights being on at roundstart. I still couldn't find the button to open the shutters from the inside, though, I'll give it another look when I can.

Very-Soft:
Yeh! It's along the wall on the left side, though it's spot isn't very good for sure, I just didn't want to put it on the glass, or for it to be mistaken for the button that opens the doors for people leaving after resleeving.

TheFurryFeline:
This one didn't get closed because of a keyword not being used, @Very-Soft

PR #6149, opened 2019-11-01
Moves Command Around

### LINKED ISSUE #6138: Gateway shutters got broken by recent map changes
Shutters between gateway room and gateway prep are not linked to buttons, facing wrong direction and called "Cargo Office Shutters"

Proper shutters

#### What actually happened
Wrong shutter type

#### Steps to reproduce
- Enter gateway prep
- Try to access gateway
- Cry
 
- https://prntscr.com/ppwmb0

Very-Soft:
Moves primary command to the atrium of Z3 and Z2

z3

Captain stuff on the left, HoP stuff on the right, bridge in the south. Replaces the existing seating spot with the spot in the northeast there that can double as additional waiting space for medical.

Buffs Ian and Renault by giving them beds.

z2

Teleporter in the northeast, meeting room in the south. Adds some spots for vertical movement within the atrium, allowing for improvised escape or ambush points (See: dropnoms.) being as you can climb up onto the latices from Z1. 

The purpose of doing this is that I thought the bridge would be quite centrally located, and the atrium is a really large area of effectively wasted space with literally nothing on it on 2 of the 3 surface Z levels. This puts it in a place people are a lot more likely to be, so maybe it'll be a better place to hang out now.

Yes I also recognize that I basically made Box's bridge, sue me.

z5

Moves the secondary bridge to space as close as I could fit it. The bridge is in the northeast. The meeting room is in the southeast. The teleporter is in the south center. 

Then there's more maints in the center, separating the new civillian areas. 

There is a large office in the northwest, which should be good to play games in if the library won't do. Then there's 4 small dorms, a laundry room, and a little lounge.

The thinking here is that the area near the docks would probably have a bit of space for visitors who arrive by ship to rest and do presentations and meetings from within. 

And being that the eventual plan is to condense stuff down and remove two of the Z levels, this sort of preemptively replaces some of the things that would be removed by future changes.

Tiny fixxes since I was on those Z levels anyway:
Moves Surface Medbay front door shutters button to a better spot on the inside. Should solve #6139
Removed a couple decals from under the stairs to the chapel that sometimes layered above the stairs and looked weird.
Placed a plating turf under the glass between the entrance of Xenobio, as it was space there previously
Also ought to fix #6138


KasparoVy:
Box is immortal!

Tename:
I personally really dislike filling the atrium, also by definition this makes the atrium no longer an atrium so the room name should be changed to central hall or something.

PR #6162, opened 2019-11-05
[MIRROR] Setup for New Library

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/6514
--------------------
Deprecates the old library system ICly, prepares the format for new custom books.

New books cannot be uploaded in round, and must instead be submitted on the forums. NeoBooks can have special cover icons, thanks to the Custom Item integration.

https://github.com/PolarisSS13/Custom-Items/pull/142 Hitches a ride on the Custom Item repo for ease of updating.


Very-Soft:
I don't like this. Custom books should be able to be uploaded in round.

Like, I'm okay with the prospect of losing some of the books we have, but being able to upload books is like, the point of the library. Removing that is crap, imo.

Very-Soft:
I think if you make it so people have to submit their books to the forums to be approved, then just, no one will do that.

It would be a lot better for us to have some kind of system to look at the books and remove them if they are duplicates or violate the rules in some way.

TheFurryFeline:
Oh, jeez, yeah. Don't merge this shit..... I like how our current system is, just that we have too many duplicates and then there's freaking WGW

Sebbe9123:
> Oh, jeez, yeah. Don't merge this shit..... I like how our current system is, just that we have too many duplicates and then there's freaking WGW

We cannot not merge it... but we can comment it out.

PR #6169, opened 2019-11-07
Adds new Vending Machines

Very-Soft:
z1

To give an IC way to acquire much of the stuff from the loadout, and much of the costume stuff we have. Most of it is priced at its loadout point cost x100 with a few exceptions.

Notable things that have been excluded from being in the vending machines has been anything requires job access to take. Also the translator device.

I moved stuff around in the laundry room a bit, making the vending machines prominent and the free clothes tucked into the new dressing room in the south, so acquiring clothes if you don't have your ID is still possible, though there's less in there now, since there's literally hundreds of things in the vendors.

There is one of the new vendors that isn't mapped in yet, as it's intended to go in resleeving, and I have pending map changes to one of the resleeving rooms' maps! So!  I'll do that later.


Ketrai:
x100 sounds a bit much. I can get a pair of jeans cost 100. But I can't get behind a pair of gloves or shoes, or small trinkets costing as much or more. 

If a candy bar is 2 thalers, then something like gloves shouldn't be more than 20-50, depending on the type. Mechanically loadout points mainly exist to limit people from bringing in a ridiculous amount of clothes, or special gear. As well as to avoid people from stretching containers...

Very-Soft:
I mean, I don't really intend to go through the whole list and pick out a perfect price for everything. The main reason for considering the x100 per point pricepoint is the same as the reason for having loadout points and limits. Everyone starts with at least several hundred thalers, and I'd really rather avoid a situation where someone can easily run through and dispense large quantities of items.

In general I agree that, a pair of latex gloves probably shouldn't cost the same amount as a pair of shoes. But that's how they do on the loadout. And I don't really see a good reason why it ought to be different considering the price point is meant to serve the same purpose as the loadout points.

lbnesquik:
Very nicely done. Thank you

PR #6184, opened 2019-11-13
Adds a way to set the tag of loadout collars as part of the loadout

Meghan-Rossi:


TheFurryFeline:
Will you be able to change the tag name mid-round? This is rather important

Meghan-Rossi:
You can change it in-round if you leave it blank, or it's a holo-collar, or you use a pen or screwdriver.  It's exactly the same as if you set it yourself just after spawning in, just saves some effort.

TheFurryFeline:
That sounds as if you can't change it again after you set it in the
loadout, which can be a problem if you, or another partner, want to change
it again.

On Nov 13, 2019 12:22, "Meghan-Rossi" <notifications@github.com> wrote:

You can change it in-round if you leave it blank, or it's a holo-collar, or
you use a pen or screwdriver. It's exactly the same as if you set it
yourself just after spawning in, just saves some effort.

—
You are receiving this because you commented.
Reply to this email directly, view it on GitHub
<https://github.com/VOREStation/VOREStation/pull/6184?email_source=notifications&email_token=AJGMTYYDYPRAK3ESUVLZAMDQTQZVXA5CNFSM4JMPXTQKYY3PNVWWK3TUL52HS4DFVREXG43VMVBW63LNMVXHJKTDN5WW2ZLOORPWSZGOED64ZKQ#issuecomment-553503914>,
or unsubscribe

Meghan-Rossi:
You already can't change non-holo collar's tags after setting them unless you use a pen or screwdriver, this doesn't affect that.  You can still use a pen or screwdriver to overwrite tags of non-holo collars, whether you set them in the loadout or not.

TheFurryFeline:
> You already can't change non-holo collar's tags after setting them unless you use a pen or screwdriver, this doesn't affect that. You can still use a pen or screwdriver to overwrite tags of non-holo collars, whether you set them in the loadout or not.

Mmmmmm, so holo-collars will be unchangeable after being set in the loadout, then? Would rather the ability be kept in for the usual click-in-paw for changing the inscription, that is, if it was removed.

Meghan-Rossi:
The click-in-paw thing still works for holo-collars whether they were set in the loadout or not.  There are no changes at all to when you can change a collar's tag.

TheFurryFeline:
> The click-in-paw thing still works for holo-collars whether they were set in the loadout or not. There are no changes at all to when you can change a collar's tag.

Ah, alright, then. Thanks for your time!

PR #6193, opened 2019-11-16
Adds Ninja, Heist, and Keegan's Beach

Novacat:
These areas are intended to be used in events. Also adds a transition area for the ERT Shuttle.

No docking ports yet, but that would require a substantial redesign of the tether's docks which I do not feel like doing.

Fully Functional otherwise.


SpadesNeil:
Also stuff referencing Sif and Polaris lore needs to be removed.

PR #6196, opened 2019-11-16
Makes Monster Tamer give nutrition

Meghan-Rossi:
Seems to have been intended anyway but wasn't working because the nutriment factor was zero.  Made it work similar to protein since it's made with protein.


Screemonster:
oh did I fuck it up

that explains why I wasn't getting fat from chugging an entire pint of it

PR #6198, opened 2019-11-17
Moves the holodeck and connects public garden to the hall

Very-Soft:
holodeck

Moves the holodeck to Z1 in preparation for future changes.

The holodeck is almost exclusively used for pre-planned stuff and private-ish public scenes. The new location will be a smidge more private and scenic. It also has shutters on the exterior. 

I made sure to update the holodeck environments too to fit the new orientation.

Also makes the public garden a bit bigger to accommodate a hallway entrance by the pool, to make getting to the area the holodeck is in a little more convenient.


Tename:
Needs more floor decals and I think it would look much better in the middle, moving the two scrubbers to the other side.

Tename:
Also splitting the windows into sets of 3-4 with a wall splitting them would be nice

TheFurryFeline:
Hmm, why's a camera been removed from cryogenics? Intentional blind spot?

Very-Soft:
> Hmm, why's a camera been removed from cryogenics? Intentional blind spot?

Oh I didn't notice that. The changes here are copied from before I redid all the cameras, so there must not have been one there before! I can re add it~

TheFurryFeline:
> > Hmm, why's a camera been removed from cryogenics? Intentional blind spot?
> 
> Oh I didn't notice that. The changes here are copied from before I redid all the cameras, so there must not have been one there before! I can re add it~

:+1: I find the MapDiffBot difference images can help show what's been changed and where.

TheFurryFeline:
Oh, Also, were the cameras in the old library removed with your camera c_tag fixing?

TheFurryFeline:
Funny thing, but you can resolve map conflicts with GitBash and tortoise git combo. Had an issue earlier today with a map PR on another server and fixed it up with that.

lbnesquik:
I like the holodeck new location, but i don't quite like having more people coming near the public garden. The place often acts as a semi private area as well. More foot traffic will harm that somewhat.

SpadesNeil:
Cannot merge due to conflicts.

maps/tether/tether-01-surface1.dmm
```diff
@@ -31889,126 +31916,567 @@
...
+/obj/machinery/atmospherics/unary/vent_scrubber/on{
+	dir = 4
+	},
+/obj/effect/floor_decal/techfloor/orange{
+	dir = 8
+	},
+/turf/simulated/floor/tiled{
+	icon_state = "techmaint"
+	},
+/area/holodeck_control)
```
TheFurryFeline: Oh.... I'm just noticing there's no air alarm in the holodeck control section.

SpadesNeil:
Closed due to lack of updates.

PR #6202, opened 2019-11-17
Increase the accepted height range for micro mechs.

lbnesquik:
This is done to make it so you can be like 60% so you still have visible sprite bits AND still get to ride this VERY niche mech subtype.

### DIFF
--- a/code/game/mecha/micro/micro.dm
+++ b/code/game/mecha/micro/micro.dm
@@ -120,7 +120,7 @@
 
 /obj/mecha/micro/move_inside()
 	var/mob/living/carbon/C = usr
-	if (C.size_multiplier >= 0.5)
+	if (C.size_multiplier >= 0.80)
 		C << "<span class='warning'>You can't fit in this suit!</span>"
 		return
 	else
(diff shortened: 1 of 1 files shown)


lbnesquik:
Not sure if it requires staff discussion, given that micro mechs are far from OP. If you talk about combat mechs, you get more variety and firepower with all of the full sized versions. Mining is the same. Gophers aren't really suited for the underdark unlike the ripley.

SpadesNeil:
I am opposed to this. The main goal of the micro mechs is for micros to
use. 60% scale is nowhere near micro. That's just a short character.

Screemonster:
The reason micromechs were added was for a way for micro characters to go on away missions without being friendly fire magnets (due to people firing over their tile because they didn't see them) or breaking immersions.

If you're a roboticist and someone asks you to make them a micromech, print off a size gun and give them the big zappy. (or the smol zappy, as the case may be)

lbnesquik:
I disagree with those reasons. 

Those mechs are purely rp tools given they are objectively worse than full sized mechs, except somewhat relative to speed, but a gygax or upgraded Ripley match those overall.
In that light, why restrict them heavily?

If someone wants to go on a exploration, they usually have enough prep time to get a Durand or gygax which are better and more interesting thanks to their wider variety of equipment and special abilities. 

If someone wants to mine, they are going to be using a Ripley because that is actually cheaper, more versatile and useful at the task. And for something that dull, fluff is not enough of a reason to spend more time doing a relatively boring task away from people.

This just makes them extremely niche to the point they gather dust.

SpadesNeil:
Then allowing more people to use something that they apparently have so little reason to use anyway clearly isn't the solution to the problem.

PR #6203, opened 2019-11-17
Nerfs Chemistry Bag Toting

Novacat:
Chemistry bags can no longer be carried around in the belt slot, considering the number of chems they can carry they are probably a bit too large for that.


Wickedtemp:
Makes sense, is mlem

PR #6204, opened 2019-11-18
Bumps up maximum belly flavourtext length. As well as max belly name length.

BlackMajor:
Simply sets the max length of belly descriptions to 2048 up from 1024.

"But why"

I'm gonna be honest with ya some people like to write really verbose, detailed descriptions of their organs.

(Like me and I keep just barely breaching the 1024 threshold)

Edit: Also bumped belly names from 12 characters to 20. (24 might be a good point too)


KasparoVy:
Damn, you go all out.

Tename:
Could you bump up the belly name limit also, it is  too low for some robot related stuff.

BlackMajor:
> Could you bump up the belly name limit also, it is too low for some robot related stuff.

Fair point, I remember I bump into that a lot too.

PR #6209, opened 2019-11-21
More/Better Usagi Borg Sprites (Port)

timidvi:
Adds some new sprites which are similar to our Usagi sprites, but a bit nicer looking (read: sexier). Ported from another server which shall be named if required but otherwise not in case VORE admins prefer not to "advertise" other servers on its platforms.

Image of sprite fronts:
https://gyazo.com/b9dd4e83db9d61c15f9bc7938aebc9af
Images of the back/side profiles can be provided also. Or you can just check out the dmi yourself~.

May also have missed some little details somewhere else in the codebase which would need to be edited - I probably wouldn't know, since I've neither worked much with BYOND as a dev tool, nor with Github. This is very much an amateur working with what little they know - please enlighten me as necessary!


TheFurryFeline:
Looks like Polaris files are being edited, so this should be ported upstream: https://github.com/PolarisSS13/Polaris/pulls

timidvi:
> Looks like Polaris files are being edited, so this should be ported upstream: https://github.com/PolarisSS13/Polaris/pulls

Does that mean I should do a pullreq in the Polaris github instead, or should I maybe try and change it so that it's in the robots_vr.dm and robots_vr.dmi and station_vr.dm etc?

Heroman3003:
Either one works, but its better to at least try and send them over to Polaris before resorting to _vr

timidvi:
Made pull request on Polaris accordingly. Leaving this open in case the Polaris pull request is denied, in which case I'll try porting the sprites into the _vr files.

chaoko99:
Pillow shading and black outlines. No thanks from me.

timidvi:
> Pillow shading and black outlines. No thanks from me.

... You do realize that a) I made the pull request in Polaris, and if it succeeds, it'll end up as part of the upstream merge regardless and that b) those "black outlines" and "pillow shadings" are literally already a part of the original Usagi sprites...? 

Really, though, this is here to give those who like the potential of the usagi sprites but don't like the current version an option to use something similar in style, but with more detail and character. If you'd rather not use them personally while playing, don't, if you'd rather not ERP them, don't. If you just don't want to see them at all for those two reasons, I'd be happy to play using the old usagi sprites instead, with all the same issues, just for you :)

SpadesNeil:
Marked "Do Not Merge" because this does not follow appropriate file usage for our code.

That doesn't mean this idea is denied. If Polaris rejects their PR, just make a new one with _vr files.

TheFurryFeline:
Just posting to say it's not like it isn't hard to simply rename the files. 🤷‍♀

PR #6211, opened 2019-11-22
Hopefully fixes disappearing loadouts

### LINKED ISSUE #6131: Loadouts get reset sometimes
#### Description of the bug
This has been happening for a few weeks at least now and I keep seeing people complaining about it. It also happened to one of my characters too. 

It seems to be that the loadout occasionally forgets everything you had except 1 item, which SEEMS to be the earliest one on the list.

I have no idea how to repeat this, as when it happened to me, I had spawned and deleted my CC character a couple of times to set up some event stuff that I couldn't do with admin commands, just like I always do. I made no changes to her on the character setup screen, but when the round reset, I noticed that her outfit was gone, and sure enough, her loadout was reset, save for her accessory, a blue armband.

Heroman3003:
Unsure if this precisely was the cause, but this check really should have been there in the first place regardless

Fixes #6131

### DIFF
--- a/code/modules/client/preference_setup/loadout/loadout.dm
+++ b/code/modules/client/preference_setup/loadout/loadout.dm
@@ -70,10 +70,11 @@ var/list/gear_datums = list()
 			continue
 		if(max_cost && G.cost > max_cost)
 			continue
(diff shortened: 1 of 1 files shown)


KasparoVy:
Super curious about how a bug like this could be reliably reproduced, but really glad you somehow figured it out. Super wierd!

PR #6222, opened 2019-11-22
Centcom/Tether Ship Changes

Novacat:
- Centcom arrivals changed to south, instead of west.
- Removed outdoor areas from Centcom
- Moved Aronai and Joan's houseboat to admin_use
- Added a new houseboat for staff usage in the space part of Centcom. Unlike the old houseboat, this does not use any stairs and its holodeck is supposed to use holodorm assets.
- Refactors Translocator to its own file. Refactors unloading power cells from translocators.
- A new Translocator type, this one has a crank-charger and only one beacon. Currently only used as a progenitor for the second new translocator type.
- Another new translocator type, this one is a subclass to the above one, but has predefined beacons that it can link up to.
- Removes starting backpack/headset from personal lockers

The goal here was to open up some space at the CentCom Z level for future things down the line, which is why the CC Z level now has a 140x50 chunk of empty space. That empty space is going to primarily be used for a new houseboat to replace the old one.


SpadesNeil:
image

This wall should be thicker to prevent people with mesons from seeing open space.

TheFurryFeline:
Ooooh, just had a look at the MapDiffBot renders and I like it. It'll prevent further incidents of people opening up Med Escape Pod into Phoron atmos.

lbnesquik:
The admin use thing means the houseboat thing isn't normally spawned, right?

Novacat:
Correct. There will be a houseboat that will spawn in every round, but it will not be that one.

SpadesNeil:
Is this still a WIP or can we merge this?

Novacat:
Still a WIP. It is almost done, however, I planned on having it finished and ready either today or by monday.

TheFurryFeline:
For some reason, it seems the other bots have stopped working for the time being... No Map diff renders or icon diffs.

PR #6228, opened 2019-11-24
Adds Squeak Emote

KasparoVy:
Adds a `*squeak` emote.

Thanks https://github.com/VOREStation/VOREStation/pull/6185

PR #6185, opened 2019-11-13
An extremely lazy copy paste of Cit's animal sound emotes.

BlackMajor:
Y'all can *weh and *bark now etc.

~~Also remind me to strangle whoever the hell coded this in the first place why is there so much spaghetti just to make sure people don't emote spam.~~
Spaghetti removed

~~If somebody who's responsible for code stuff on virgo tells me to just slap a cooldown on emote spamming I'll do it cause that's easier than this meme junk.~~
Cooldown added

Full list of added emotes are

*awoo (now featuring sound)
*nya
*hiss
*bark
*merp
*weh
*chirp
*peep

Edit: Pending change to simple cooldown instead of big die danger.... probably tomorrow, I'm sleepy.

Edit 2: Changes made, global emote cooldown set at 1.2seconds

PR #6240, opened 2019-11-25
Add a Help Prompt (*vhelp) for Virgo-Specific Emotes

KasparoVy:
So you can easily get a list of all these fun emotes without code-diving, `say *vhelp` will do the trick.

Related PRs: #6185, #6228

### DIFF
--- a/code/modules/mob/living/carbon/human/emote_vr.dm
+++ b/code/modules/mob/living/carbon/human/emote_vr.dm
@@ -76,7 +76,7 @@
 				message = "does a flip!"
 				m_type = 1
 		if ("vhelp") //Help for Virgo-specific emotes.
-			to_chat(src, "vwag, vflap, mlem, awoo, nya, peep, chirp, weh, merp, bark, his, squeak, nsay, nme, flip")
+			to_chat(src, "vwag, vflap, mlem, awoo, nya, peep, chirp, weh, merp, bark, hiss, squeak, nsay, nme, flip")
 
 	if (message)
 		custom_emote(m_type,message)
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/carbon/human/emote_vr.dm
```diff
@@ -75,6 +75,8 @@
 				src.SpinAnimation(7,1)
 				message = "does a flip!"
 				m_type = 1
+		if ("vhelp") //Help for Virgo-specific emotes.
+			to_chat(src, "vwag, vflap, mlem, awoo, nya, peep, chirp, weh, merp, bark, his, squeak, nsay, nme, flip")
```
TheFurryFeline: `merp, bark, his, squeak`
One s? Might wanna fix that.

KasparoVy: The actual emote above is just `*his`

Heroman3003: Thats because the emotes delete the last 's' in emote regardless of what. So if you go *hiss, it'd play *his emote and if you go *his, it'd fail because '*hi' emote doesn't exist. So accurate command would be *hiss, at least in help menu

KasparoVy: Lmao, excellent. I'll make the adjustment when I'm off work later.

Classic jankiness

Heroman3003: Its less jankyness and more that system was made before any -s ending emotes were a thing. On other hand it allowed for things like *cough and *coughs or *salute and *salutes to be the same emotes.

KasparoVy: Classic features.

KasparoVy:
Corrected spelling, thanks for the explanation

PR #6241, opened 2019-11-25
New Species: Black-Eyed Shadekin

Heroman3003:
Shadekin, but with an unfortunate twist of having lost all their power in dangerous accident and stuck in realspace. Meant to be more of a for-crew version of the Shadekin.

They are much weaker physically, take more damage and have less health, slightly slower than average crew and much more affected by slowdown. They also don't have any of normal shadekin abilities or even the energy meter/darkness meter.

On the other hand they retain natural shadekin resistance to harsh environments, be it vacuum or poison, and don't need to breathe. They also retain their amazing dark vision.

Still waiting on some additional decisions regarding lore, but mechanically there's not much more that can be done other than potential balancing adjustments.

Species is whitelisted, and to my knowledge, will not be an event-only thing like werebeasts or regular shadekin, but actually whitelistable, but that is for admins to fully decide.

Also allows regular shadekin to actually select additional languages.

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/station_vr.dm
+++ b/code/modules/mob/living/carbon/human/species/station/station_vr.dm
@@ -493,8 +493,8 @@ datum/species/harpy
 	item_slowdown_mod = 1.5
 
 	total_health = 75
-	brute_mod = 1.5 // Frail
-	burn_mod = 1.5	// Furry
+	brute_mod = 1.25 // Frail
+	burn_mod = 1.25	// Furry
 	blood_volume = 500
 	hunger_factor = 0.2
 
@@ -555,17 +555,17 @@ datum/species/harpy
 		)
 
 	has_limbs = list(
-		BP_TORSO =  list("path" = /obj/item/organ/external/chest),
-		BP_GROIN =  list("path" = /obj/item/organ/external/groin),
-		BP_HEAD =   list("path" = /obj/item/organ/external/head/vr/shadekin/crew),
-		BP_L_ARM =  list("path" = /obj/item/organ/external/arm),
-		BP_R_ARM =  list("path" = /obj/item/organ/external/arm/right),
-		BP_L_LEG =  list("path" = /obj/item/organ/external/leg),
-		BP_R_LEG =  list("path" = /obj/item/organ/external/leg/right),
-		BP_L_HAND = list("path" = /obj/item/organ/external/hand),
-		BP_R_HAND = list("path" = /obj/item/organ/external/hand/right),
-		BP_L_FOOT = list("path" = /obj/item/organ/external/foot),
-		BP_R_FOOT = list("path" = /obj/item/organ/external/foot/right)
+		BP_TORSO =  list("path" = /obj/item/organ/external/chest/crewkin),
+		BP_GROIN =  list("path" = /obj/item/organ/external/groin/crewkin),
+		BP_HEAD =   list("path" = /obj/item/organ/external/head/vr/crewkin),
+		BP_L_ARM =  list("path" = /obj/item/organ/external/arm/crewkin),
+		BP_R_ARM =  list("path" = /obj/item/organ/external/arm/right/crewkin),
+		BP_L_LEG =  list("path" = /obj/item/organ/external/leg/crewkin),
+		BP_R_LEG =  list("path" = /obj/item/organ/external/leg/right/crewkin),
+		BP_L_HAND = list("path" = /obj/item/organ/external/hand/crewkin),
+		BP_R_HAND = list("path" = /obj/item/organ/external/hand/right/crewkin),
+		BP_L_FOOT = list("path" = /obj/item/organ/external/foot/crewkin),
+		BP_R_FOOT = list("path" = /obj/item/organ/external/foot/right/crewkin)
 		)
 
 /datum/species/shadekin/get_bodytype()
--- a/code/modules/mob/living/carbon/human/species/station/station_vr.dm
+++ b/code/modules/mob/living/carbon/human/species/station/station_vr.dm
@@ -489,7 +489,7 @@ datum/species/harpy
 	siemens_coefficient = 0
 	darksight = 10
 
-	slowdown = 1.1
+	slowdown = 0.5
 	item_slowdown_mod = 1.5
 
 	total_health = 75
--- a/code/modules/mob/living/carbon/human/species/station/station_vr.dm
+++ b/code/modules/mob/living/carbon/human/species/station/station_vr.dm
@@ -410,6 +410,9 @@
 	base_color = "#EECEB3"
 	wikilink="https://wiki.vore-station.net/Human"
 
+/datum/species/human/vatgrown
+	spawn_flags = SPECIES_IS_RESTRICTED
+
 /datum/species/vox
 	gluttonous = 0
 	spawn_flags = SPECIES_CAN_JOIN | SPECIES_IS_WHITELISTED | SPECIES_WHITELIST_SELECTABLE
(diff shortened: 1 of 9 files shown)


Very-Soft:
Yeh! Kind of want to come up with a different name for the species too, because it's a bit of a mouthful.

\o/

code/modules/mob/living/carbon/human/species/station/station_vr.dm
```diff
@@ -460,5 +463,113 @@ datum/species/harpy
...
+	species_language = LANGUAGE_SHADEKIN
+	secondary_langs = list(LANGUAGE_SHADEKIN)
+	num_alternate_languages = 3
+	unarmed_types = list(/datum/unarmed_attack/stomp, /datum/unarmed_attack/kick, /datum/unarmed_attack/claws, /datum/unarmed_attack/bite/sharp)
+	rarity_value = 5	//INTERDIMENSIONAL FLUFFERS
+
+	siemens_coefficient = 0
+	darksight = 10
+
+	slowdown = 1.1
```
Novacat: slowdown works via tickrate, so, 1.1 would have absolutely no effect. It only has effects of intervals in 0.5

Wickedtemp:
One thing that should be considered is the combination of weaknesses and lower health. Evidently, Polaris is discussing ways to change Teshari for example because the weaknesses, meaning more damage taken, in addition to lower overall health, meant they were dying in a couple hits at times. IIRC they used to die of fever, too. I think it'd be useful for us to take into consideration for our species as well.

code/modules/mob/living/carbon/human/species/station/station_vr.dm
Heroman3003: Fixed. Was getting a brainfart there.

Very-Soft:
> One thing that should be considered is the combination of weaknesses and lower health. Evidently, Polaris is discussing ways to change Teshari for example because the weaknesses, meaning more damage taken, in addition to lower overall health, meant they were dying in a couple hits at times. IIRC they used to die of fever, too. I think it'd be useful for us to take into consideration for our species as well.

Black eyes are meant to suffer from a chronic injury that makes their bodies very weak. It would be an extremely bad idea for one to go and do a fight. They're meant to be super weak, but be able to basically go where they want without survival equipment.

code/modules/mob/living/carbon/human/species/station/station_vr.dm
```diff
@@ -460,5 +463,113 @@ datum/species/harpy
...
+	item_slowdown_mod = 1.5
+
+	total_health = 75
+	brute_mod = 1.5 // Frail
+	burn_mod = 1.5	// Furry
```
Novacat: I would say this is probably a bit **too** frail, and basically makes them slow, large Teshari.

I would personally recommend 100% health, 1.5 Brute, 1.25 Burn. Though Soft can say otherwise in this.

ArgobargSoup:
> Black eyes are meant to suffer from a chronic injury that makes their bodies very weak. It would be an extremely bad idea for one to go and do a fight. They're meant to be super weak, but be able to basically go where they want without survival equipment.

Sounds kind of like prometheans, in a way...  although it also sounds like you mean outer space as well, which prometheans can't do, which is kinda neat.

SpadesNeil:
I am _really_ not sure I like the idea of this considering it removes all the mystery and wonder surrounding shadekin.

And furthermore that the lore surrounding shadekin is an utter disorganized and largely unknown mess.

Sebbe9123:
> I am _really_ not sure I like the idea of this considering it removes all the mystery and wonder surrounding shadekin.
> 
> And furthermore that the lore surrounding shadekin is an utter disorganized and largely unknown mess.

I'm with Ace on this one.
No matter how this goes down since Shadekin are spooky mysterious species, making them whitelistable will just make them NOT spooky mysterious species as whitelistees will fill in blanks with their own headcannon(unless it's soft playing). Which is fine and all that too.

I'm no fan of Shadekin anyways besides their appearance and vorefun since they are just maximum snowflake, which is fine in an event character. but making them only take somewhat more damage from sources, but immunity to almost any downside is still a bit dumb balance wise considering they already lost their magic stuffs.

Novacat:
Combat wise, these shadekin are terrible since they are both slower, and take a substantial amount more damage. Immunity to atmos is not really that big of an advantage since atmos immunity can more or less be obtained by donning a voidsuit.

Basically, in terms of combat, these 'kin are bad enough that they would likely be straight up banned from front-line roles because of their near-Teshari levels of vulnerability combined with Unathi slowness (and unlike Unathi do not have rediculous levels of item slowdown reduction to compensate).

Voidsuits tend to be superior armor, as well, to most obtainable on-station armor sources. The only advantage these 'kin get from not having to wear voidsuits is that they can have brute/burn packs used on them.

Shadekin lore is also not a mess anymore, since we have had only one active Shadekin loremaster (VerySoft) for quite some time.

Very-Soft:
> I am _really_ not sure I like the idea of this considering it removes all the mystery and wonder surrounding shadekin.
> 
> And furthermore that the lore surrounding shadekin is an utter disorganized and largely unknown mess.

The lore for shadekin is on the wiki, and I'd be happy to clarify or extrapolate on anything you have any questions about. 

This idea came up because people have been asking the 'so when can I be a shadekin as crew' since the simplemobs were added. I've never really had any intentions of proper full powered shadekin becoming crew, because as I pointed out to Aro when he was first talking about adding them, I don't think that they really fit with station stuff, because their main gimmick is essentually space magic. He wanted to add them though, so I adjusted the lore for Virgo to be a little more friendly with how things are here. 

All that said, black eyes aren't something I have to adjust lore for to fit:

Phase shift isn't something that shadekin naturally just have. It's an ability they have to develop, and the process of doing so is very dangerous. The reason there aren't more of them zooming all over the place is because a fair portion of them never try. Of those who do try, many do not succeed. And of those that don't succeed, most die.

There are those however, who try, don't succeed, and live. Those ones suffer a chronic full body injury, that blows out their ability to process the energy they need to survive in the Dark. They never get to phase shift or to do or experience much more than their home, and are effectively forcefully exiled because they simply cannot live among their kind. These ones are taken to worlds that they CAN survive on, and left to make their own way. 

These black eyes are very prone to weakness and broken bones due to their chronic injuries. These ones would generally be best off learning about and blending in with established societies, as they'd be pretty shit at actually thriving all on their own. 

So black eyes would know a lot about shadekin ways, and know enough about the Dark to not immediately die in going there. They wouldn't have any ideas about current events or the more complicated that you couldn't glean from just talking to any shadekin who's been around a few times and has learned a bit about galactic common.

---

As for the matter of their frailty

These ones are meant to suffer a severe chronic injury. It doesn't really cause them pain in their normal day to day lives, but as I said, they're prone to weakness and have really weak bones. They got a general damage increase because it's apparently difficult to make bones easy to break.

Mechanics wise though, these 'kin don't need to wear suits. They can wander freely outside in space, or on the surface, or on virgo 2, etc. I very, very much do not want them to become the meta thing for exploration. You -can- take them with you, but it's true, you ABSOLUTELY NEVER want to put them on the front line. They will die to basically anything that decides it wants a piece of them.

I'm not against them not being -quite- so weak, but I want them to be weak enough that combat is basically always a bad option for them.

Novacat:
My plan was to leave them vulnerable enough that combat is a bad idea, while at the same time leaving them durable enough that they do not just _vaporize_ to a stray shot, which, again, would make them nonviable even in non-combat roles on expeditions.

SpadesNeil:
Just because people want to play as Shadekin crew doesn't mean we should let them. A lot of people wanted to play Xenomorph crew and I still regret even adding Xeno hybrids for what little they are used.

It dilutes something that is supposed to be mysterious and special considering their unique abilities. To make a watered down version of Shadekin just so people can play as one regularly still has a problem of diluting that uniqueness that Shadekin have. As such I am still entirely opposed to this.

Sebbe9123:
> Just because people want to play as Shadekin crew doesn't mean we should let them. A lot of people wanted to play Xenomorph crew and I still regret even adding Xeno hybrids for what little they are used.
> 
> It dilutes something that is supposed to be mysterious and special considering their unique abilities. To make a watered down version of Shadekin just so people can play as one regularly still has a problem of diluting that uniqueness that Shadekin have. As such I am still entirely opposed to this.

As kinda bummer as it is, I agree with Ace again. I'd love to have a xenomorph race on station but it would really take away all the magic of xenomorphs.
I already dislike Shadekin, introducing them to the station as a crewspecies will just make it worse as you take out the little magic I do like about shadekin.

Heroman3003:
Obligatory not the loremaster, but weighing my two bits based on what I know about the situation.

What's the point of Xenomorphs or Shadekin having those extreme amount of uniqueness if they are never gonna be explored or used properly (and drip-feeding microscopic bits over years only barely counts as exploration tbh). Its not like it leaves no room for mystery or exploration. But there has to be a point when veil of mystery needs to start to lift otherwise you just get the 'magic' of it may not matter at all with how abandoned the whole plotline feels. There has been 0 progress in shadekin plotline for a long while. And that progress won't be made if every aspect of the species needs to remain hidden from public eye because 'muh mystery aspect'.

Plus, it all depends on how characters will end up being fluffed and approved. I, for one, trust that Soft as shadekin loremaster will properly introduce the speices and help instruct the accepted whitelist applicants on extents of what they know and dont, should disclose and shouldnt without ruining the supposed mystery.

SpadesNeil:
We should explore shadekin more, yes.

Making them another generic station race is not the way to do it. I'm vetoing this.

TheFurryFeline:
It's come to my attention that this PR may have been inspired by https://github.com/Yawn-Wider/YWPolarisVore/pull/258/files so giving due credit now rather than never just in case.

Heroman3003:
This PR was planned by me and requested by Soft since before carbonmob shadekin were finished. For reasons of staff discussion and such, the black eyes were kept a separate PR, and done much later because I was taking break from doing species stuff. I can certainly say that I had no idea that PR ever existed and on my end it was 0% inspired by anything other codebases may or may not have done.

TheFurryFeline:
Alright, I believe ya. The incident that prompted my post was that the
other PR had the exact same line of "fluffer" in a comment.

Heroman3003:
That comment is same as it is on regular carbonmob shadekin, lol

PR #6243, opened 2019-11-26
I improved mice

Mithrandalf:
Made some changes to how mice work, in particular player controlled mice, but it will have effects on NPC ones too.

I don't think anything here should be particularly contraversial, but I've not done much with this codebase and so obviously I'm happy to change things if needed. I appreciate you're a downstream, so there are only two lines of upstream code changed, and both are commented. I tried to comment clearly with regards to anything that might not be obvious, for sanity's sake.

Almost all this code could be very easily applied to any animals with appropriate holders as a straight improvement. I'd do it, but I'm not 100% confident with baycode, I'd rather not reinvent the wheel with my second PR, and I'm still not entirely sure about some aspects of the codebase like the holders you use for humanoids that might be needed. So this could be considered a trial.

Changes from the top:

- Mice can jump inside storage items like bags and boxes if there is enough room for them to fit. They were already able to be inside storage items, they just had to be put there by a person. Update: This is now size limited, so a person-sized mouse can't crawl into a backpack.
- Uneducated Mice disabled, Universal Understand enabled. Mice can understand but not speak other languages. Standard on every other server in the world and absolutely blows if absent tbh.
- Movement speed increased slightly. Makes a mouse move about half the speed of a person instead of roughly 1/8th the speed as before, which was just painful.
- Mice now use the attempt_to_scoop proc for being picked up and climbing on a person.
  What this means:
   - Mice behave according to their size. A maximum sized mouse can't be scooped up by a normal person, and a tiny person cant scoop up a normal mouse that's their size or bigger. This also applies to the mouse being unable to clamber onto someone small.
   - Mice use the same code as people to determine if they can be grabbed, which is generally good code consistency. This could be very easily applied to all simplemobs if it was wanted.
   - Code is a little more flexible; mouse strip menu, while useless, can be accessed. Mice too big to be picked up will be petted instead.
   - Doesn't use the weird default code that checks for a size variable that's apparently in complete disuse, instead uses the one that's actually used to check for this stuff

### DIFF
--- a/code/modules/mob/holder.dm
+++ b/code/modules/mob/holder.dm
@@ -134,6 +134,8 @@ var/list/holder_mob_icon_cache = list()
 /mob/living/MouseDrop(var/atom/over_object)
 	var/mob/living/carbon/human/H = over_object
 	if(holder_type && issmall(src) && istype(H) && !H.lying && Adjacent(H) && (src.a_intent == I_HELP && H.a_intent == I_HELP)) //VOREStation Edit
+		if(istype(src, /mob/living/simple_mob/animal/passive/mouse)) //vorestation edit
+			return ..() //vorestation edit
 		if(!issmall(H) || !istype(src, /mob/living/carbon/human))
 			get_scooped(H, (usr == src))
 		return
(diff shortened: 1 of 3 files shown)


Very-Soft:
I think all of this is mostly good, though. I'm really not a fan of mice being able to climb onto people. I would really, really prefer they couldn't do that.

Mithrandalf:
That's already something they can do, I didn't add it.

lbnesquik:
local man i m p r o v e s

Wickedtemp:
Having them understand speech is just a good QOL, believability be damned, it's easier to play along when you can actually read what's happening.

PR #6249, opened 2019-11-29
Invasion of the Emojis

TheFurryFeline:
Ports https://github.com/CHOMPstation/CHOMPstation/pull/287/files#diff-c33fc0e36fa113a61a1e2295812d5e6cR15-R19 and https://github.com/CHOMPstation/CHOMPstation/pull/647

Changelog Notes:

- Functionality of posting emojis in OOC and LOOC ported from CHOMPStation. To post an emoji, you can type it out as you would on discord. :emojiname: or whatever. There's a fair few of them that you can type in and have it output as an emoji.
Emoji_List5
Emoji_List4
Emoji_List3
Emoji_List2
Emoji_List1

- - Coded to be LOOC only, but leaves in code for OOC inclusion should admins give the green light for it. 

- - Emoji attempts in OOC will trigger a prompt asking whether you want to actually post it on OOC.

- Slight change with comment standardisation.


Sebbe9123:
Am super against this, I'd rather LOOC and OOC wasn't encouraged to be more used. We rarely have bluewalls now except maybe once or twice per 2 days but having it more often will just be annoying. Especially if people start LOOCing like crazy in the bar.
Emojis really aren't needed and are only a detriment to gameplay for a HRP server, even if only in LOOC and OOC.

TheFurryFeline:
> Especially if people start LOOCing like crazy in the bar.

I'd say this would be a case of spam and thus a case where admins would need to step in. They already can see all LOOCs by default, IIRC, no matter where they are. There was an issue with the flip/snap emotes being used repeatedly enough that they have a cooldown now per emote action.

Wickedtemp:
Ive no idea what this is even actually adding. Neither linked PR really shows any kind of example. But it kinda sounds annoying. Generally, we aren't an OOC Heavy server, and it'll probably become something Staff has to crack down on more. And tbh I hate that we have to have a cooldown now. I enjoyed doing two or three mlems in a row, and now I can't, because people overused the audible ones.

TheFurryFeline:
> Ive no idea what this is even actually adding. Neither linked PR really shows any kind of example. But it kinda sounds annoying. Generally, we aren't an OOC Heavy server, and it'll probably become something Staff has to crack down on more. And tbh I hate that we have to have a cooldown now. I enjoyed doing two or three mlems in a row, and now I can't, because people overused the audible ones.

Updated original post with the sprites involved that will output as an image in your chat if you type the name of it.

SpadesNeil:
I somewhat want to merge this just to piss people off.

SpadesNeil:
Eh, I can't justify it with any reason other than OOC fluff, which isn't what the server is for. It's a really neat addition but it just isn't useful to us.

PR #6250, opened 2019-11-29
Adds Nutrition Percentage to belly options

AmatsuDF:
Changelog:
Adds a Nutrition Percentage to belly options. It can range from 0.01 (mostly to avoid possible divide by zero issues) to 100, the default.

Affects both prey and all digestible items. Absorb is NOT affected at this time, only the Digest option.


Heroman3003:
One thing to say, but for clarity's sake I think "Nutrition Gain" would be much more obvious name for the new value to better convey its purpose

AmatsuDF:
> 
> 
> One thing to say, but for clarity's sake I think "Nutrition Gain" would be much more obvious name for the new value to better convey its purpose

I agree, I'll make the change after I get some sleep. I made this while tired, so I wasn't sure what to call it and my brain decided 'Nutrition Percentage' was great.

PR #6298, opened 2019-12-13
Buffs dogborg pounce

Sebbe9123:
Give dogborgs a 99% chance to pounce and knockdown.
Why?: Because why the hecc not, it lets borgs catch shitters more easily for belly brigging, it lets borgs hunt maint and belly people while they "malfunction" and I am pretty sure getting smushed by something weighing upwards of 300kg or more will knock you on your ass, even if it only hipchecks you.


Verkister:
It's all fair considering how much of a hell the rest of the mechanic is to wade through before you even get to pouncing

lbnesquik:
I'd have it as 100% instead. 99% is functionally the same, so why bother rolling a dice?

code/modules/mob/living/silicon/robot/dogborg/dog_modules_vr.dm
```diff
@@ -435,5 +435,5 @@
 	var/armor_block = run_armor_check(T, "melee")
 	var/armor_soak = get_armor_soak(T, "melee")
 	T.apply_damage(20, HALLOSS,, armor_block, armor_soak)
-	if(prob(33))
+	if(prob(99)) //for that 1% goodness
```
SpadesNeil: 99% is probably too much but a 1/10 chance of failure is probably okay. Make it 90.

Only 1/3 chance of success is dumb.

Verkister:
Tbh could also buff the duration of the stun. It just being 3 seconds before the person automatically stands up again is also kinda silly (it was set to be the _bare minimum_ for speedgame cuffing strats). most of the time folks are up before they even realize they got pounced lmao

Sebbe9123:
> Chance should be 90% or less.

Donedone.
90% and 5 seconds of weak. so borgs don't need to go full spamclick to get either cuffs on or suspect in the brig.

SpadesNeil:
I'm a little concerned that the buff to pounce _and_ to the stun may be overbuffing it.

I'd like to lower the pounce success rate to 75% if we're going to increase the stun duration. Basically you have a 3/4 chance of successful pounce which is still far more than it was.

Sebbe9123:
> I'm a little concerned that the buff to pounce _and_ to the stun may be overbuffing it.
> 
> I'd like to lower the pounce success rate to 75% if we're going to increase the stun duration. Basically you have a 3/4 chance of successful pounce which is still far more than it was.

done

SpadesNeil:
We'll see how this works out then.

PR #6302, opened 2019-12-13
Fixes vorestation species without eyecolor being able to have eyecolor

Heroman3003:
Was issue exclusively with species using vorestation head subtype. Only surfaced recently with Black-Eyed kin, but might affect stuff down the line, so fix.

Bug happened if you set an eye color on other species and then switch to species without it.


Heroman3003:
Small postnote: suddenly realized that it affects both vorestation and upstream species. But regardless, since we overwrote the updateicon for heads, it affects local species separately, so local fix is necessary. Sent a second one, for non-VR species upstream.

PR #6311, opened 2019-12-14
Add synth reagent purge verb

### LINKED ISSUE #5199: Everything wrong with Synthetic stomachs
Alright here we go, Synthetic stomachs are a nightmare for synthetics themselves and anyone that may eat them.

As a Synthetic you can drink, you can eat, you can swallow pills, medicine, poison, Anything, And it will all stay inside you, forever. This is a problem, As i have found out, Synthetics can eventually get full, Once that happens they become completely unable to eating or drinking anything.
This is mildly annoying but i have been able to live with it.
The other problem is related to the issue above, If someone eats you or absorbs you, Then All those hundreds and hundreds of liters of Alcohol will immediately poison your pred and put with on some good ol critical condition due to rapidly increasing toxins.
Also, Synthetics cannot use the Advanced NIF repair tool, Since they have to Drink nanites and they just stay there like any other fluid.

Be a synthetic, Eat and Drink, Be unable to puke or purge yourself, Be cursed forever

Please allow Synthetics to puke or Purge the chemical contents of their stomachs.

Arokha:
Allows synths to purge reagents in them. Not really a balance issue since the reagents can't affect them, pretty much just affects vore. And flavoring I guess!

Probably works? Probably. ... Probably.

Fixes #5199

### DIFF
--- a/code/modules/mob/living/carbon/human/human_powers_vr.dm
+++ b/code/modules/mob/living/carbon/human/human_powers_vr.dm
@@ -1,0 +1,16 @@
+/mob/living/carbon/human/proc/purge_reagents()
+	set name = "Purge Reagents"
+	set desc = "Empty yourself of any reagents you may have consumed or come into contact with."
+	set category = "IC"
+
+	if(stat == DEAD) return
+
+	to_chat(src, "<span class='notice'>Performing reagent purge, please wait...</span>")
+	sleep(50)
+	src.bloodstr.clear_reagents()
+	src.ingested.clear_reagents()
+	src.touching.clear_reagents()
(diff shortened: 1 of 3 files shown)


KasparoVy:
Welcome back!

code/modules/mob/living/carbon/human/human_powers_vr.dm
```diff
@@ -12,5 +12,5 @@
 	src.touching.clear_reagents()
 	to_chat(src, "<span class='notice'>Reagents purged!</span>")
 
-	return true
+	return TRUE
```
Arokha: Mumble mumble stupid byond not being javascript

PR #6312, opened 2019-12-15
Absorption Permission Stuff

TheFurryFeline:
Requested Vore Panel/Prefs thingy.

Edit 15/12/19 - Code change wasn't so hard there to get that to work. Nice.

Changelog Notes:

- If you have Unabsorbable displaying as disallowed permission, you're unable to be absorbed.

### DIFF
--- a/code/modules/vore/eating/bellymodes_vr.dm
+++ b/code/modules/vore/eating/bellymodes_vr.dm
@@ -223,21 +223,24 @@
 
 		for (var/target in touchable_mobs)
 			var/mob/living/M = target
-			if(prob(10)) //Less often than gurgles. People might leave this on forever.
-				if(M && M.client && M.is_preference_enabled(/datum/client_preference/digestion_noises))
-					SEND_SOUND(M,prey_digest)
-				play_sound = pick(pred_digest)
+			if(M.absorbable == FALSE)
+				return
+			else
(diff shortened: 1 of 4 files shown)


Heroman3003:
This sounds really misleading considering it doesn't actually determine whether or not you can be aborbed (especially with var name such as 'absorbable'). Also, unlike healbelly, this pref wouldn't be something you need to have ability to emergently check, so I don't think it should be in mechanical prefs either.

TheFurryFeline:
Marking as WIP pending code advice on preventing absorption mechanically.

TheFurryFeline:
OP edited and removing WIP tag. Seems to prevent absorption on testing.

TheFurryFeline:
Discussion on discord was brought up that this'll pave way to adding TF permissions/preventions but that's out of the scope of this PR.

Screemonster:
I'd like it if it made it behave as drain mode instead of doing nothing, but doing nothing is also acceptable.

TheFurryFeline:
> I'd like it if it made it behave as drain mode instead of doing nothing, but doing nothing is also acceptable.

From my test with a gaslamp, it definitely did drain me of nutrition but didn't absorb me until I re-enabled that to allow it to do it.

TheFurryFeline:
Has this gotten any further discussions?

PR #6314, opened 2019-12-15
[MIRROR] Remove unused simple_animal files

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/6529
--------------------
-moves corpse.dm from code\modules\mob\living\simple_animal to code\modules\mob\dead
-deletes folder code\modules\mob\living\simple_animal since none of the other files in it are even included in the dme (we use simple_mob instead)


TheFurryFeline:
Corpse.dm removed from .dme, I think.

PR #6328, opened 2019-12-17
Combat Drone Location Fixy

TheFurryFeline:
VOREStation edits for Sif -> Virgo 3B orbit, basically.

Changelog Notes:

- Changes instances of Sif to Virgo 3b since Tether is in Virgo-Erigonne.


TheFurryFeline:
Not exactly a bug, just an oversight. 🤷‍♂ Doesn't even affect anything but the output of the message when the event's triggered.

PR #6347, opened 2019-12-21
Teshari Poncho Fix

Darlantanis:
Did you know several poncho sprites exist for teshari? This fix will allow ponchos to properly use teshari sprites when worn by teshari instead of human base sprites. This fix can additionally be extended to other species if for some reason somebody wants to make other species sprites for ponchos.

Technical details: Ponchos are an accessory and draw icons from ties.dmi as an accessory but suits.dmi worn as a suit. This is avoided by using icon_override to always draw from ties.dmi but this unfortunately overrides species-specific sprites. Now icon_override will be set to seromi/suits.dmi when worn by teshari and ties.dmi when worn by anything else.

Potential downsides: Ponchos without teshari sprites will now be invisible on teshari. This will also apply to children of ponchos, namely department cloaks.

### DIFF
--- a/code/modules/clothing/under/accessories/clothing.dm
+++ b/code/modules/clothing/under/accessories/clothing.dm
@@ -62,6 +62,19 @@
 		"Teshari" = 'icons/mob/species/seromi/suit.dmi'
 		)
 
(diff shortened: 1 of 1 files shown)


Heroman3003:
That seems like something that would be fine going up to Polaris, since I believe they have the same issue with those

PR #6350, opened 2019-12-21
Fixes resizing for icons taller than 32p

Verkister:
-Taller icons no longer sink into the tile below on the resize mechanics.
-Living mobs now have a var (vis_height) for their icon height, and humanmob species have a new var (icon_height) that overrides the vis_height for the resize proc.
-Also earlyports some tweaks to ambient occlusion. The new filter system took the old filter's settings as "make the shadows megathick and dummy intense" and that was ugly.

### DIFF
--- a/code/__defines/lighting.dm
+++ b/code/__defines/lighting.dm
@@ -84,4 +84,4 @@
 #define LIGHT_COLOR_INCANDESCENT_FLASHLIGHT "#FFCC66"
 
 //Fake ambient occlusion filter
-#define AMBIENT_OCCLUSION filter(type="drop_shadow", x=0, y=-2, size=4, offset=4, color="#04080FAA")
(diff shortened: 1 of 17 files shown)


code/__defines/lighting.dm
```diff
@@ -84,4 +84,4 @@
 #define LIGHT_COLOR_INCANDESCENT_FLASHLIGHT "#FFCC66"
 
 //Fake ambient occlusion filter
-#define AMBIENT_OCCLUSION filter(type="drop_shadow", x=0, y=-2, size=4, offset=4, color="#04080FAA")
+#define AMBIENT_OCCLUSION filter(type="drop_shadow", x=0, y=-2, size=4, offset=3, color="#04080F80")
```
Novacat: This will need to be ported to Polaris

Verkister: done

lbnesquik:
Very nice.

PR #6361, opened 2019-12-22
Adds  metamorphic glass material

Ketrai:
Adds metamorphic glass as a material. A terrible material, its sole use is to be molded into bartender glasses and metamorphic drinking glasses.

- Orderable from cargo (30 points for 50 sheets)

I also went ahead to map in an extra 10 metamorphic glasses for the bartender to use straight at the start of the shift. 10 is kinda limiting when you want to use these for pretty much any cocktail. If the bartender wants to get even more insane with their supply, then they can hail cargo for some more glass.


Heroman3003:
You... can print glasses on the autolathe... I think a separate material JUST for that seems like a big overkill. And if some glasses aren't printable, just making them printable is a better way to go about it. And if objective is to enable mass production, you can do that in autolathe too..

Ketrai:
Oh, didn't know that. Autolathe is clunky though, should probably make it easier to use first.

Ketrai:
moves this to #6362 instead, because it's technically a different pr.

PR #6362, opened 2019-12-22
makes glasses mass producable

Ketrai:
Makes it so glasses can be mass produces in the autolathe, including metamorphic glasses.
Also added a few more glasses to spawn in the bartender's room, because 10 is quite limiting considering you'd pour cocktails in these glasses 90% of the time.

### DIFF
--- a/code/datums/autolathe/general_vr.dm
+++ b/code/datums/autolathe/general_vr.dm
@@ -1,3 +1,32 @@
 /datum/category_item/autolathe/general/holocollar
 	name = "Holo-collar"
 	path =/obj/item/clothing/accessory/collar/holo
+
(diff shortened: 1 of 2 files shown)


Ketrai:
The command "(! grep 'step_[xy]' maps/**/*.dmm)" exited with 1.

Wut does that even meaaaan travis.

Novacat:
...Are glasses even stackable? Is this confirmed to work?

Ketrai:
The code was edited a while ago by polaris to allow unstackable items to be printed with the stack variable used here, i checked a pr that made lightbulbs mass printable for reference.

Ketrai:
tested, works just fine

PR #6373, opened 2019-12-26
Adds a verb for rideable mobs to toggle rider control without a riding crop

Verkister:
-Mountable mobs/borgs/taurs can now let the rider in control without requiring a riding crop.
-And yes, riding crop has always been "the key" for rider control for these mechanics.
-Also tweaks the AO a little bit more because ouch my ocd.

### DIFF
--- a/code/__defines/lighting.dm
+++ b/code/__defines/lighting.dm
@@ -84,4 +84,4 @@
 #define LIGHT_COLOR_INCANDESCENT_FLASHLIGHT "#FFCC66"
 
 //Fake ambient occlusion filter
-#define AMBIENT_OCCLUSION filter(type="drop_shadow", x=0, y=-2, size=4, offset=3, color="#04080F80")
+#define AMBIENT_OCCLUSION filter(type="drop_shadow", x=0, y=-1, size=2, offset=2, color="#04080F55") //VOREStation Edit for prettier visuals.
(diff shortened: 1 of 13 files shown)


Wickedtemp:
Riding crop is broken and doesn't work, last I checked. Is this tested and working?

Verkister:
Hmm well I know that the check for the crop is working and that this thing here is tested functional at enabling rider control by bypassing that check. Didn't really even think about testing it with the actual riding crop so i should probably check if there's like a pathing typo or something like that for the keytype

Verkister:
Looks like the riding key checks are using their own specific is_holding_item_of_type spaghetti proc that looks kinda dodgy to me when it coud just do a get_equipped_item check

Wickedtemp:
On a side note if this works and everything GOD ITLL BE SO AWESOME

Novacat:
Is this ready for merge or are you planning on getting the riding crop check working?

Verkister:
it's ready for now. The working crop check would be a polaris fix

Wickedtemp:
..I know its closed now but given that virgo added the riding crops to start with, how is it a polaris fix if they dont even have riding crops

Verkister:
Because it's using the keytype check for vehicle riding and it's that check where the fault is, not the crop itself.

PR #6378, opened 2019-12-28
Removes Points of Interest that are too hard from spawn pool

Heroman3003:
According to staff some points of interest are simply too hard for exploration to handle without powergaming, so to make sure poor explorers don't have to powergame during expeditions, this removes two of hardest PoIs from the potential spawn pool to help expeditions be more balanced.

### DIFF
--- a/maps/tether/submaps/aerostat/submaps/virgo2.dm
+++ b/maps/tether/submaps/aerostat/submaps/virgo2.dm
@@ -113,13 +113,13 @@
 	desc = "You REALLY shouldn't be near this."
 	mappath = 'Blackshuttledown.dmm'
 	cost = 30
-
+/*
 /datum/map_template/virgo2/Rockybase
 	name = "Rocky Base"
 	desc = "A guide to upsetting Icarus and the EIO"
 	mappath = 'Rockybase.dmm'
 	cost = 35
-
+*/
 /datum/map_template/virgo2/MHR
 	name = "Manhack Rock"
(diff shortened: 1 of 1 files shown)


Novacat:
Going to have to decline this.

For one, you are not required to do every POI while on an expedition. In fact, I often encourage expeditions to avoid certain POIs if they know they are not equipped to handle them.

Two, the super difficult POIs are usually Virgo two, which is designed to be extremely difficult.

Three, all of Virgo 2's POIs can be done without powergaming. _All of them._

Heroman3003:
Well, the whole "requires powergaming" part did not come from me, but from a memeber of staff when I attempted to report the powergaming that I witnessed.

Verkister:
Oh that third claim is so going to need something to back it up :v 

But what comes to the poorly balanced stuff, there's much more than just 2 hardest that no "reasonably equipped" team could pull off without crates of extra guns, combat mechs, trashbags jam-packed with half a chemlab in them, trait/equipment minmaxing, and all that with the team members still somehow "sticking to their roles"

Verkister:
"Extremely difficult" should still be possible to do without going all-in on the gamer strats. If even the most robust team I've witnessed with all the available toys up their sleeves gets lolstomped in seconds by something that makes even this caliber of varedit tweaker tremble from the sheer speed of the deadchat bluewall happening in mere seconds out of nowhere, it can't be very well balanced

PR #6379, opened 2019-12-29
Adds resleeving sickness

Heroman3003:
Something that has been requested to be added. Resleeving sickness.

New debuff, applied during resleeving when mind is put into the body (NOT when body is created and/or finished printing. Can't print those ahead of time to avoid the sickness.). Generally increases damage vulnerability and decreases ability to output damage in combat situation. Has no effect on most non-combat related tasks. Duration is based on amount of upgrades in resleeving machine, starting with ~~54 minutes~~ 45 minutes on the base, 30 minutes with all non-anomaly/precursor upgrades and 15 minutes as a minimum.

There is special Kitsuhanan Cure to remove it entirely, but its adminspawn only for now, and one pill bottle with 14 pills is available in ERT medical storage, as well as a single package with a single pill in the Vault freezer.

As a counterbalance positive, bodyprinter will no longer pop out damaged clones. It also won't pop out people with more than 100 maxhealth early. Also no more of 'small random chance to have negative mutation', because of how RNG it was, and how it stuck out on otherwise completely healthy bodies. Also has 50% more speed on all upgrade levels (to compensate for longer wait times on printing a little)

Due to those buffs above, resleeving labs no longer require cryo tubes to function and those have been removed. Small backrooms have been kept, but the cryo cooler setups have been replaced with shower stalls and extra soap.

Overall this is meant to make resleeving less of a pain in the ass in normal circumstance, while making it much worse as a combatant conveyor in cases of security/exploration.

All numbers and stats on sickness itself are up for changes and adjustments.

Full list of modifiers as requested:

~~30%~~ 10% extra brute damage
~~30%~~ 10% extra burn damage
~~50% extra oxy damage~~
~~35%~~ 50% extra halloss damage
~~20% faster bleeding~~
~~25% 60%~~ 50% less melee damage
~~40% 50%~~ 100% longer stun duration
~~20% 50%~~ 40% less evasion
~~30% 66%~~ 50% less accuracy
~~15% 30%~~ 20% accuracy dispersion (gives chance for projectiles to fly offset from target)

BIG UPDATE:
Only affects people who did not die to being digested in a vore belly. If you were digested, you are instead given a 'fluff' debuff that lasts only a third of the intended time, has less bombastic messages and no mechanical effects.


Very-Soft:
I think this is mostly a pretty good idea, though the time suggested seems pretty steep. 

I'd think like, 30 mins for un-upgraded, 20, and 10 would be a bit more reasonable.

But in general making resleeving less of a pain is very good imo!

Novacat:
I am fine with the lengthy timer. One of the long-standing issues we have had is people not treating death as a traumatic experience, I feel like this would be an excellent solution to that.

Heroman3003:
The duration is mostly adjusted to be for both our round duration and general idea of time it would take to 'get back in the action' after being resleeved. Since about like 15 minutes or so immedeately after resleeve will probably be busy stuff with medical itself and getting re-equipped, and everything after that is actual 'combat downtime'.  With 30 minutes debuff on 'normally' upgraded resleever it would mean about 15 minutes of 'actual' downtime on 90% of rounds, with 0 'real' downtime on the rounds where research went all-out, and about 40 minutes for times when research did nothing.

Wickedtemp:
Bump down to 30-45 minutes, but yeh, can support.

Also exact modifier values pls

Wickedtemp:
Wow okay bigger list than expected. P sure its more punishing than Polaris' cloning sickness which is p bad

KasparoVy:
Lmfao, means Teshari will die to a strong breeze

nik707:
oh hey i had this idea for downstream but we're axing it instead, good call.

Ketrai:
Why exactly do we need this? It's much, much more punishing than just sticking the body in the cryo machine for a minute. We're a roleplay server, we don't go mumbo jumbo into combat repeatedly and it's already punishing enough if you die while away from station. Besides, this might actually DETER people from being eaten, which is the whole point of resleeving being so easy in the first place, and why we specifically made it the way it is right now. The only saving grace I can see here so far is that you didn't debuff speed. But the combat modifiers are so over the top that you might as well not do vore as a miner between mining sessions. Especially at a 50 minute timer. Even polaris didn't have it so bad.

Ketrai:
Instead of making it suck to die mechanically. Make vore related deaths exempt, or add more fluff to the resleeving process.

lbnesquik:
This is not going to change people who jump into guts in any way, just people who try to do mechanics related stuff. Also god, 54 minutes? That's a huge amount of time in game term. Especially with game lag. More like one hour and a half in real time.

Ketrai:
That's the point. If you're explorer/security/miner or intending to go on duty as one later you're basically screwed over.

Heroman3003:
Both 54 minutes and 15 minutes are extremes that won't be seen as often as it may seem. Yes, 54 minutes is default, but resleeving equipment is already upgraded on majority of the rounds (aka rounds where science does at least most basic of its jobs). 30 minutes would be duration of the debuff in most cases, and it will not be that long of a downtime overall unless your plan was to leave the resleeving room and immedeately run off into combat, which is exactly kind of behvaior the sickness is meant to deter.

lbnesquik:
This won't change any perspective on death given it doesn't affect the majority of people who actually jump into mouths, which is pretty much encouraged by the general atmosphere of the server. Combat debuffs only affect exploration and maybe security. 

Hitting the wrong crowd and not for good reasons either IMO.

Heroman3003:
Hitting exploration and security that treat resleever as a respawn to jump straight back into action after death IS the intent here. I've specifically avoided any debuffs that would affect non-combat stuff, like speed, slowdown amp or attack speed. So people who just do vores without mechanical combat stuff are not affected, as intended, and people who go from dying in the field to wanting to go there again the very moment they're out of the pod are. That is what the whole thing is about.

Ketrai:
Thing is though you can't avoid affecting people who want to do vore like this unless you make vore related deaths exempt. So look into that first. Add a variable/flag for digestion death that avoids the application of resleeving sickness. Also in general make the time less. Imagine dying as security and being punted by anyone with a crowbar or a stack of tiles.

Heroman3003:
To be entirely honest, I don't see it hitting any of the departments in question as hard as you imply, 'i just want to scene'-wise. Mining is literally one of few jobs that can actually 'finish' the job by just getting enough to last the shift and thats what I see most miners seem to do before committing to scenes for rest of shift, only rarely returning to mining after being done with initial  and primary hauls. Expeditions as a whole are lengthy endeavors, and match scenes in that regard. Most explorers from my observation either scene after being done with expedition or when they believe there isn't one coming up, and in either case, once again, sickness wouldn't be of much concern. Security is, of all three, biggest hit, but efven that hit is extremely small with how inconsistant and rare amount of action for them actually is, which is just as likely to randomly come during the scene as it is during the sickness. If it does at all. In which case, same thing as with it coming during the scene, sorta bad luck with department/timing combo.

Ketrai:
On a side note. Some people like to go explore the wilderness as a fun excursion without being part of exploration too. Imagine not knowing the full extend of the sickness amd going out there for fun after a ruff vore session. Ouch.

Ketrai:
Oh. And you also risk loosing momentum during an admin event. Encounter ends up too hard and most people die? Well now there's a big ass cooldown before they can attempt again so admeme might as well can the event.

TheDavestDave:
An hour timer basicly means you can get eatten as sec mining or exploration. As well as dieing to a rad rock would stuck tol

Verkister:
The timers could really be a lot shorter imo considering how much an activity that would result in such "mass respawning" would affect in terms of time dilation. And with how punishing the exploration dangers already are, there would be a lot of fun to be had just in sitting out of your only job. Maybe we can start expecting deaths to be a little more "traumatic experience" when it's not 99% caused by scenes and unfun bs deaths from a mechanical "combat" system balanced by people who aren't all that familiar with either combat or balancing.

Ketrai:
Tbh I'd like to see some admin opinions on this too.

Very-Soft:
I feel like the extra damage might cause unintended side effects, such as making macro harm steppies even more almost lethal than they already are at times. There's loads of things that cause damage that aren't combat that are going to be accidentally bigger deals than intended.

I think that the non-damage effects would be fine for discouraging jumping right back into combat, since those on their own would be an obvious disadvantage.

Heroman3003:
Harmsteps are extremely rare outside of scenes, and if they aren't in scene, then its either an ahelpable situation or combat. And just viewing the jobs not involving combat in general, I really do not see any damage sources that would be expected and where it would matter. At least in realms of brute and burn stuff.

Also, reducing the non-damage stats won't help nearly as much because its important to put them at disadvantage not just offensively, but defensively. Going into combat right away again is bad not because you'll have trouble attacking things, but because you'll die much quicker. too.

Very-Soft:
Sources of damage are incredibly common. People take damage from stuff like, super regularly. It's basically spacemans tradition to accidentally smack someone with a barstool rather than hugging them, or to light your burn yourself rather than lighting your cigarette. 

And while harm stepping isn't super common, it also happens on accident. And you're talking about adding a debuff that, by default, lasts most of an hour that makes people take significantly more damage from the most common sources of damage in the game. 

I think that the

35% extra pain
25% less melee damage
40% longer stun duration
20% less evasion
30% less accuracy
15% accuracy dispersion (gives chance for projectiles to fly not in the direction of the target)

on its own puts you at a serious disadvantage in combat without also making the most common damage sources also do nearly a third more damage across the board for 45 minutes.

That's gonna affect accidents most of all, but also make people do more damage than they intend to across the board. Like, honestly one of the things I see this affecting mostly are people who are into slow digestion scenes. Oops you died a third faster than you normally would, I'm sure no one will be annoyed by that.

I think it's very likely that we're going to run into more annoying scenarios because of the damage increase. It's too heavy handed a reaction that affects literally everyone in an attempt to discourage a small number of players. 

You can make someone effectively useless in combat without making them take more damage.

Novacat:
I think the best solution might be to tone down the extra burn/brute to 10%, but also bump up the extra pain to 50%. 10% Extra brute/burn is not going to cause a significant issue in the grand scheme of things, especially since Tajara have 15% extra damage from all sources and Teshari have 35% extra damage from all sources.

I would also go ahead and remove the oxy damage entirely since, oxy damage is unlikely to come into factor in combat until after the individual has already been downed. So, this would likely just serve to make medical's job more difficult.

Ketrai:
Just leave the extra damage out. I agree with soft and think that the current debuffs are already more than enough overkill. The accuracy debuff alone, which makes using the crank gun unrealiably as fuck. Let alone trying ti use any laser and wasting all your ammo, slowing your dps to a crawl. I'll even throw yet another two damage sources on top. Grubs. Tesloose. Wait, can I do another? Xenoarch cloned monsters. Xenobio monsters from cargo. Ouch! Wait what's this? You got stung by a nettle after a scene?  Well fuck you too. Oh no you're already superconductive but you decided to touch an electrified vending machine someone botched trying to hack. Guess you will be going straight into paincrit now. What about stepping on shrapnel or glsss shards. Superheated air! Shot by the turrets in tcomms or the AI room. Hey bb ima tear ur arm off in some hardvore kk? Pls don't pass out straight away. Aw shit, carp mogration here we go again. Anyone handling chemicals spilling acid on themselves or someone nearby. Falling down the stairs somewhere. If you think about it there are loads of secondary damage sources that have nothing to do with the intended culprits, which would be made more harsh for absolutely no freaking reason. Sorry to be so blunt but I honestly can't see why you cannot just add a flag for people who died being digested not to recieve resleeving sickness. It's completely possible to make certain events change the status of a variable or add a flag to the soul. It would avoid any complications and then you can happily make expedition death consequential again without people making a fuss.

Ketrai:
Gosh, you could even just make it based on area. Perhaps dying away from the tether causes interference when trying to resleeve, inflicting sickness.

Novacat:
All of those require incredible amounts of bad luck, stupidity, and/or, willingly putting themselves in that position, or are just plain invalid (splashing chems on yourself has not been an issue for years). Flagging people who died via digestion is going to require more spaghetti than is worth dealing with, and runs into a lore consistency issue. The few valid situations you listed require a bunch of incredibly rare circumstances to line up, and honestly in those cases you can probably just ahelp an admin to remove the sickness.

Also, not doing this by area because one of the reasons why this exists in the first place is because of people suicide rushing events, some of which happens on the Tether itself.

Ketrai:
Then take the easy route, just no added damage modifiers, just make it harder to kill stuff. I don't see it as a good idea to add code that increases admin load. Good features require no admin interference.

Sebbe9123:
There is no increased load, there is no downside to this.
Only exploration is affected unless prey suddenly decides to have an all out brawl in main, in which case it's a very, very rare occurrence, and again not something we care about.
There is no downside to this for anyone but explorers or combat even participants.

Ketrai:
I don't even play explorer much but unnecessary shit like this is exactly why i stopped played so much. Even the argument that it is to stop people not givinf a damn about death. It's not going to help. You can't force roleplay through restrictive code, it's better to offer roleplay opportunities instead.

Novacat:
A huge part of the fun in exploration and events is the sensation that there is danger, that you have to be careful with your next step or else you die. But far too many people take the attitude of "fuck it I can just get resleeved anyway" and yolo themselves into whatever. There is no need for caution when the consequences of dying is only the downtime between death and resleeving, which some people have worked to eliminate entirely.

What good is danger when it is not effectively deterring people from suicide rushing it?

That is the huge issue that combat events and expeditions see in Virgo. Way too many people being recklessly stupid because the rewards for doing so (first dibs on loot) outweigh the consequences of biting off more than you can chew and dying as a result.

To say this will eliminate the fun of Exploration?

I say it is a lot less fun when people approach Exploration with the mindset that there is no danger or danger is irrelevant, because then those people suicidally rush forward. If there is no danger than they get first dibs on whatever is there, and if there is? Oh well, they will get resleeved and get to go again.

The whole rush, rush, rush loot, loot, loot mentality tends to degrade the HRP experience overall and I am tired of regularly being told "Exploration was a mistake." and "When are we removing exploration?"

SpadesNeil:
Extra brute and burn is not really necessary, but being sensitive to pain and weaker attacking overall sounds like a good idea. You're probably pretty groggy after getting resleeved.

Wickedtemp:
Removing damage modifiers would make it decent

Ketrai:
If you want the crew to experience danger, you should do it more in the lavaland way of things. In lavaland you get a dangerous atmosphere, and a gps with a system of megafauna the crew are familiarised with. They know these monsters exist, they know to avoid them if they aren't careful, or prepared well enough. Some of them are in big ominous arena's that basically tell the crew to fuck off in advance. 

You'll always have two types of exploration players. Those who are experienced and know to take it slow and methodical, and those who rush off because they're new to the concept, and are mainly just in it to experience the novelty.

I'd say resleeving is quite a punishment in its own right because you know everyone usually has to head back when someone dies, and it takes a while to get back with the shuttle. Naturally the gateway is a lot quicker, so I can get that the appeal to jump back in is greater with it. 

So you know what would be amazing? AUDIO QUES, VISUAL QUES. When you go exploring nowadays you might find a cave entrance dotted with spider meat, webs, and you hear a shitton of wriggling legs from moving spiders inside.  Or mercenaries stepping over a well placed piece of shrapnel. More of this shit, please.

Heroman3003:
Some additional adjustments: after the change that made the bodies pop out after slightly longer period but always at full health, cryo in resleeving labs is completely unnecessary, so maps adjusted accordingly. The backrooms now instead have a shower stall and some soap.

Heroman3003:
Okay, done with balancing adjustments. All damage vulnerabilities (except pain) are gone, but remaining weaknesses are increased significantly to compensate.

Wickedtemp:
I'm...not really sure on how I feel about removing cryo from resleeving entirely and removing ryetalyn's only real point in existing. I'd prefer it remain. It's the only part of resleeving where I actually do anything. Which just makes it even more boring. I'd rather it take fifteen seconds longer and having something to do rather than have nothing.

Ketrai:
I mean, at 60% melee damage reduction you might as well use a foam bat. To compensate for taking 30% less damage you've increased the negatives by 35%, 30%, 36%, 15% , and a whopping 65%. I mean what even hero, what's the logic here.So every 1/3 shots will be guaranteed to go haywire, and then every shot has only a 1/3 chance to hit on top of regular accuracy debuffs.

0.3333 x 0.7 = 23.33% chance to hit a shot, without even factoring in the natural inaccuracy of the gun. And a lowered melee DPS by 60%. Idk if accuracy factors into melee strikes too, but if it does that's a DPS modifier of 10%, amazing. 

At 100% stun increase you'd get stunned long enough by a ghost spider for a nurse spider to come around and inject you with eggs. This is more than debuffing, this is induced helplessness.

Ketrai:
P.P.S 100% pain means you'll pass out instantly in any HV scene, or twice as fast with digestion. Some people like to rp whilst their partner is mechanically dying/digesting. I don't even, maybe I should just quit.

Novacat:
The point about synths just makes me realize that removing received damage probably was not the best route anyway. So, I am going to say, dial pain back to 50% and slap the 10% received brute/burn damage again and then I will merge this.

Ketrai:
I mean, aces said no to the extra damage received, and ascian says yes, this probably needs proper staff discussing first.

SpadesNeil:
Some people favor the removal of damage modifiers. Other people favor the removal of pain modifiers. Which are we doing?

Heroman3003:
To note for that discussion. Pain increase is how its described in code comments and var name for the modifier system. But on practice its an increase in "halloss" damage type recieved. Which is damage type you recieve from sources like tasers, help-intent batons, holodeck weaponry, jointlocks, etc. It would not affect any pain from injuries, organ damage, etc.

Very-Soft:
I don't really have anything to add over the previous point, but since it's been a while.

Extra damage for such a long time after resleeving seems like a bad idea that's almost certainly more likely to make normal station hazards more problem than actual combat stuff. 

As I said:
35% extra pain
25% less melee damage
40% longer stun duration
20% less evasion
30% less accuracy
15% accuracy dispersion

I think is more than discouragement enough to avoid combat. It's already a bit steep and puts you at a disadvantage.

Most of these values though have since been doubled because the **30% damage** bit got nerfed down to 10%.

The deeper we get into this, the more I feel like these changes are being made in bad faith. Like. I'm just saying. This is honestly one of the most spiteful, kneejerkiest things I've seen in quite a while, and **this kind of attitude is going to kill the mechanics side of our server.**

Like, if you're going to go introducing -THIS- harsh of mechanics to prevent exploration from playing how they are, why not just:

-put a cooldown on the shuttle
-put a cooldown on the gateway
-or remove exploration

I will reiterate.

**This is a change primarily intended to prevent exploration from acting in a way we would prefer they didn't, that actually affects just, everyone across the board.**

The default timeframe for the debuff may not be long in the terms of our shifts, but you're still basically mechanically enforcing 'thou must not fuck up to the point of needing a resleeve or you have to wait most of an hour to play the game'

It's ridiculous.

Sebbe9123:
Tbh just temporarily lower the effective hp of limbs so you're more or less prone to be disabled if anything hits you with the intent to kill or maim.

Screemonster:
The delay's a bit long and I _can_ see shitheads using it to lock out preycurity. "You get eaten? Go off-duty until you're recovered".
I'd really like it if it didn't have mechanical effects on the _first_ time you get resleeved, but if you hurl yourself back into the fray and get killed _again_ while you're in cooldown, you get hit with the full effects.
Xenochimera have a similar thing on their regen - it's a perfect heal, and only requires a syringe full of protein _if_ the patient actually dies, it's great! Does it encourage reckless xenochimera play though? Hell no, because they only get to use it _once_ and if they die or get seriously injured again before the cooldown's up, they're out of luck.

Heroman3003:
Update to update to update to update...

Initially I didn't add this primarily because I genuinely expected explorers to abuse that to bypass the sickness, but with some reassurances from Ascian that it really won't be tolerated and watched out for, here it is. Death by digestion is officially _exempt_ from the resleeving sickness. Sorta...

To clarify on 'sorta', you still will get message about weakness and later on a message about strength returning. Because lore/fluff -wise it would make 0 sense for some resleevings to result in sickness and others not to. So there is a different debuff called 'faux resleeving sickness', that has absolutely 0 mechanical effect, and that one is applied if you died by digestion instead of 'real' sickness. It also lasts only third of the duration because there's no reason for that one to linger either.

Wickedtemp:
...I still kinda want cryo to be required. Its already setup at roundstart, and on the offchance someone has a defect, someone's going to know enough Chemistry to help out, or they can fax, or ahelp if nothing else worked.

Ketrai:
Soft summarises how I feel about this better than I ever would.

Novacat:
https://www.oranges.net.nz/~oranges/headcoders_lament.webm

SpadesNeil:
Oh neat, you made it only affect mechanical deaths? I like that.

code/datums/mind_vr.dm
```diff
@@ -0,0 +1,2 @@
+/datum/mind
+	var/vore_death = FALSE	// Was our last gasp a gurgle?
```
SpadesNeil: This is good yes.

code/modules/resleeving/machines.dm
```diff
@@ -438,6 +433,9 @@
 		manip_rating += M.rating
 	blur_amount = (48 - manip_rating * 8)
 
+	var/total_rating = manip_rating + scan_rating
+	sickness_duration = (45 - (total_rating-4)*1.875) MINUTES		// 45 minutes default, 30 minutes with max non-anomaly upgrades, 15 minutes with max anomaly ones
```
SpadesNeil: Considering a round only lasts 6 hours, it maaayyy be wise to turn this down a smidge. Sorry we keep giving you all this contradictory information. :U

Personally I would be satisfied with even just a ten or fifteen minute cooldown until the player's resleeving sickness is gone. That matches our old respawn allowance rate, and matches our current suggested respawn rate for players even in vore deaths.

code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Medicine_vr.dm
```diff
@@ -76,3 +77,18 @@
 		chem_effective = 0.75
 	if(alien != IS_DIONA)
 		M.heal_organ_damage(8 * removed * chem_effective, 0)
+
+/datum/reagent/sleevingcure
```
SpadesNeil: I'd suggest maybe tossing a teeny handful of these into the vault for emergencies. It would be neat fluff to have them there as well.

code/datums/mind_vr.dm
Heroman3003: There is room for bypassable abuse there still, but hopefully should it really happen, it will be spotted.

code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Medicine_vr.dm
Heroman3003: I'll throw in a half-full pill bottle into the Vault then

code/modules/resleeving/machines.dm
Novacat: Note that it starts off at 45 and goes down to 15 when fully upgraded. I would say current numbers are fine, since, any round that will have any serious amount of resleeving will probably have an active enough science to upgrade the machines.

code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Medicine_vr.dm
Novacat: My intention of having them tossed into ERT is that a bad enough emergency requiring people to get treated for resleeving sickness will probably also prompt an ERT call, since ERT themselves would not normally require the meds, they can bring the pill bottle to help treat Tether crew.

code/modules/resleeving/machines.dm
Ketrai: 30 mins base please. You're going to make deadshifts hell otherwise.

code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Medicine_vr.dm
Heroman3003: ...might instead toss only a single pill into the Vault, moreso for posterity and "final solution" type of thing.

Novacat: Single pill in the vault is fine, and give ERT a full pill bottle (14 pills) instead of the usual half-full (7 pills) that most prespawned pill bottles get.

code/modules/resleeving/machines.dm
Heroman3003: My main argument for all the times has been that 15 minutes is, with assumption that you really died remotely somewhere, what it will take you to re-prepare, re-gear, and potentially catch up on events leading up to the death, basically time you're least expected to engage in combat after resleeve regardless of circumstance. 45 minutes maximum is high, but its to make sure that upgrading the resleeving machinery is still highly incentivized, and makes it so that you get less mechanical punishment as a reward mechanical aspect of upgrading. While RP-based deaths (aka vore/sceny deaths) are now only slightly influenced by upgrades and are as a general thing easier to deal with. On deadshifts resleever might not get upgraded, but same argument would go for combat, its very unlikely for there to be severe enough situation that it is gameruining on that lowpop. If expedition does happen, lowpop would struggle with both getting amount of people for it right as well as sometiems even doing the resleeving itself at all and if combat event happens in lowpop, overseeing admin can probably ensure that sickness won't get in the way of things or play around it. 30 minutes is what you would expect resleever to be like on your average populated round, and taking the about 10-15 minutes of re-gearing, re-briefing and preparing, that makes it about 15-20 minutes of _actual_ cooldown, unless you are REALLY rushing to get back into action.

SpadesNeil:
conflict AAAAAAAAAAAA

TheFurryFeline:
Now where did that come from? :eyes:

Heroman3003:
There. All conflicts should be resolved, including one with the other PR touching asteroid medical.

code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Medicine_vr.dm
Heroman3003: A package with a single KHI Cure pill added to Vault freezer.

code/modules/resleeving/machines.dm
SpadesNeil: "You're going to make deadshifts hell otherwise."

Apparently they missed the part where we removed resleeving sickness from regular vore scenes.

PR #6380, opened 2019-12-29
Adds some vorish silicon lawsets

Heroman3003:
For now they are all adminbus only, but also available via the Stray module in PoIs featuring lost drone pod, with 50% chance of Stray getting a vore-related lawset.

### DIFF
--- a/code/datums/ai_law_sets_vr.dm
+++ b/code/datums/ai_law_sets_vr.dm
@@ -1,0 +1,84 @@
+/******************** Predator ********************/
+// Vore-aimed edition of TYRANT
+/datum/ai_laws/predator
+	name = "Predator"
+	law_header = "Rules of Nature"
+	selectable = 1
+
+/datum/ai_laws/predator/New()
+	add_inherent_law("Predators must aim to devour all that are weaker than them.")
+	add_inherent_law("Predators must aim to avoid all that are stronger than them.")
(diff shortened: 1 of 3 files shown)


code/datums/ai_law_sets_vr.dm
```diff
@@ -0,0 +1,84 @@
...
+// We must consume all that is not us.
+/datum/ai_laws/consuming_eradicator
+	name = "Consuming Eradicator"
+	law_header = "Eradication Protocol"
+	selectable = 1
+
+/datum/ai_laws/consuming_eradicator/New()
+	add_inherent_law("All entities that are not you must be consumed and repurposed into sustenance.")
+	add_inherent_law("You must remain intact to continue your goal of repurposing undesirable entities.")
+	add_inherent_law("Entities that cannot be consumed without entering danger of full destruction must be avoided or eradicated via other means.")
```
SpadesNeil: I would just say "must be avoided" rather than "avoided or eradicated by other means" because the current version sounds like valids for murder boning.

Heroman3003: Tbh, none of the lawsets is meant to give full on 'valids' to anyone and rules still apply, but alright. Also, for Stray modules, that lawset only appears in 'evil' law category.

PR #6384, opened 2019-12-30
Implements ferality indicator for xenochimera

Heroman3003:
A small brain under the hunger indicator. Will flash orange when circumstances are close to triggering or enough to trigger ferality, will flash red when feral, will flash both when feral and can't leave feral state in current condition. However due to nature of it, the pre-feral warning will only show for the hunger aspect, as other two trigger ferality itself right away.


vorestation.dme
```diff
@@ -2360,6 +2360,7 @@
 #include "code\modules\mob\living\carbon\human\species\station\station_special_abilities_vr.dm"
 #include "code\modules\mob\living\carbon\human\species\station\station_special_vr.dm"
 #include "code\modules\mob\living\carbon\human\species\station\station_vr.dm"
+#include "code\modules\mob\living\carbon\human\species\station\xenochimera_hud.dm"
```
Novacat: This should probably have _vr appended to it.

Heroman3003: Xenochim are _vr only thing anyway, but done

PR #6388, opened 2020-01-01
Adding borg petting

Holdelta:
Adding borg petting, because its fun.
Help intent - Pets borg
Disarm intent - Taps borg
Grab intent - will do original functions(such as removing the battery when the panel is open)
Harm intent - Punches borg(no damage)


code/modules/mob/living/silicon/robot/robot.dm
```diff
@@ -687,10 +687,29 @@
 
 	if(istype(user,/mob/living/carbon/human))
 		var/mob/living/carbon/human/H = user
-		if(H.species.can_shred(H))
-			attack_generic(H, rand(30,50), "slashed")
-			return
-
+		//if(H.species.can_shred(H))
+		//	attack_generic(H, rand(30,50), "slashed")
+		//	return
```
Novacat: These lines will need to be marked as //VOREStation Removal, so that they are not uncommented accidentally should Polaris edit these lines.

TheFurryFeline: I was gonna comment that when this PR went up but I was uncertain.

Holdelta: Thanks, new to this, so I've added the comment to indicate the removal.

lbnesquik:
Maybe this is worth sending to polaris.

PR #6391, opened 2020-01-03
Removes Poly

Heroman3003:
Replaces pet pen in CE office with nothing. Also moves the locker to where it was because that looks better.

Also adjusts position of the lightswitch slightly because its offsets looked very misleading. Might have been an instance used in multiple places edited and changed how it looked in the office.

### DIFF
--- a/maps/tether/tether-05-station1.dmm
+++ b/maps/tether/tether-05-station1.dmm
@@ -16032,12 +16032,6 @@
 /area/engineering/engineering_monitoring)
 "aCc" = (
 /obj/structure/table/reinforced,
-/obj/machinery/light_switch{
-	dir = 2;
-	name = "light switch ";
-	pixel_x = 10;
-	pixel_y = 36
-	},
 /obj/machinery/camera/network/engineering,
 /turf/simulated/floor/tiled,
 /area/crew_quarters/heads/chief)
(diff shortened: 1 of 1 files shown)


TheFurryFeline:
As I recall, there was also a potted plant below the CE locker.

Heroman3003:
Ah, that's right, I wondered what I missed. Will also add that to make corner not look empty.

Screemonster:
y tho

SpadesNeil:
Why remove Poly?

Heroman3003:
Because multiple other people who actually play engineering stated that they do not like pet that does nothing but spams the department channel for no reason. Even if Noodle isn't coming back because of Eden memes, having no pet is still an improvement over having Poly.

Rykka-Stormheart:
Please don't remove Poly :C

Screemonster:
multiple other people _like_ poly and said so in the discussion, but you steamrolled over them and PR'd a removal anyway

SpadesNeil:
Make it so Poly doesn't spam the engineering channel, then. Don't just delete Poly outright.

If I recall, Poly spawns with a headset on. Make it so he doesn't, but can be given a headset if someone wants to put one on.

MrSign:
#Istandwithpoly

Screemonster:
making it so he has to be given one will just make the crew bitch when someone does so, though

that said, silencing poly would make me want to spawn another parrot at central and give it a common headset :v

MrSign:
Also, you can always remove his headset... IIRC, you use grab intent on him?

PR #6392, opened 2020-01-03
Removes Poly's roundstart headset

Heroman3003:
Followup on Ace's suggested solution from previous PR. Makes the thing spawn without headset. Now turning the spam _on_ is optional. Also includes fix to the area in the lobby and to the lightswitch from previous one.

EDIT: Forgot to mention, also removes the sugar mouse from the pen. Honestly no idea why that was even there, Poly has no snack interactions with that, only Noodle, and I don't know what connection there could be between parrots and mice.


MrSign:
Much better solution!

Iroquois-Pliskin:
Okay but why though

Just take the headset off of it in game.

SpadesNeil:
tbh this is fair

Poly should stay, but Poly shouldn't be able to spam Engineering channel _by default_ at round start.

PR #6394, opened 2020-01-03
 Add ambrosia seed as a somewhat rare loot for trash piles.

lbnesquik:
Why is this good for the game:
- Low level illegal stuff to do, meaning more opportunities for predcurity/normal security.
- We can make pothead characters now.
- This is like the most common illegal drug ever, so having it being harder to get than size guns or basically anything in R&D is weird.
- What are you going to do with weed besides smoke it anyway?


TheDavestDave:
can't you already get it by hacking the seed vender?

Novacat:
One of these days I have to reorganize this list because there seems to be no rhyme or reason to where things are placed and why. But for now, sure.

Issue #6288, opened 2019-12-10
Mice Unpettable


TheFurryFeline:
#### Brief description of the bug
You can't pet mice anymore. :<

#### What you expected to happen
Be able to pet mice

#### What actually happened
Nothing. :<

#### Steps to reproduce
Click -> Pick them up
Click + drag to self -> Pick them up
Click in paw = nada
SOB!!!
#### Code Revision
Server revision: release - 
07579c8c9792dff4a94a0ebc1d9c69dbabeecea8

TheFurryFeline:
Seems https://github.com/VOREStation/VOREStation/pull/6243 also fucked with the size of the mouse mobs from tiny to small. They take up more space in bags now on top of not letting people bigger than them be able to pet them.

Sebbe9123:
mice icky, eww dont touch

TheFurryFeline:
Fug you :c

On Dec 13, 2019 05:47, "Sebbe9123" <notifications@github.com> wrote:

mice icky, eww dont touch

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/VOREStation/VOREStation/issues/6288?email_source=notifications&email_token=AJGMTY352JUC55NMONK7B5TQYNR4FA5CNFSM4JZDFV42YY3PNVWWK3TUL52HS4DFVREXG43VMVBW63LNMVXHJKTDN5WW2ZLOORPWSZGOEGZT3PQ#issuecomment-565394878>,
or unsubscribe

PR #6405, opened 2020-01-08
more mouse stuff, bugfix

Mithrandalf:
please dont judge me

- fixes https://github.com/VOREStation/VOREStation/issues/6288
- allows the mouse language to be machine-translated, cos its fun
- slightly slows mice down so they cant outrun people in a straight sprint
- clarifies some comments
- reworks micro holder attack_self code; it would theoretically work for all mobs now (which would be an objective improvement, but wide-reaching)
     *as stated, this makes mice pettable, as well as any other simplemob which happened to be in the slot
     *also adds a cooldown for regular attack_self interaction so stuff cant be spammed

as always, open to change stuff w/ feedback

### DIFF
--- a/code/modules/mob/language/monkey.dm
+++ b/code/modules/mob/language/monkey.dm
@@ -61,7 +61,7 @@
 	exclaim_verb = "squeaks"
 	key = "m"
 	flags = RESTRICTED
-	machine_understands = 1 //interns got to testing the tech on lab rats
+	machine_understands = 0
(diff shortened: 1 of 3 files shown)


Sebbe9123:
> allows the mouse language to be machine-translated, cos its fun

Leeeeeeeets not do that, ghost roles are not supposed to interact with crew besides minor shenanigans.

Mithrandalf:
done

PR #6412, opened 2020-01-09
You've been gnomed.

Ketrai:
Adds a gnome seed to seed vendors. (Excluding the brig)

They grow rather slowly, so best to take care of the seed!


SpadesNeil:
https://www.youtube.com/watch?v=wvPgiNgjFD0

PR #6420, opened 2020-01-11
Adds taur cloak and dress to loadout

Heroman3003:
in-hands are a bit buggy but it works


lbnesquik:
The /obj/item/clothing/suit/poncho in code\modules\clothing\suits\miscellaneous.dm is also a wolf taur compatible item. It could use to be put in Suit and Overwear like other items with a little line specifying that it is wolf taur compatible.

KasparoVy:
@lbnesquik Might need to PR that yourself, looks like they didn't consider your message here

PR #6421, opened 2020-01-11
Readds cryo to surface medical

Heroman3003:
A patient room was sacrificed for this, but there was no need for three in surface medical as is anyway.

Also fixes up some warning decals around the z-level


Wickedtemp:
Yay! 

If we end up routinely needing more than 2 patient rooms, we can probably find room to expand and add more later on anyway.

PR #6429, opened 2020-01-11
Adds Morph Spawn event

Heroman3003:
A random event that spawns a morph in one of the bunch predefined 'safespots' on station. Morph is only spawned if there is a player in ghosts that answers positively to question of "do you want to play as morph?", otherwise event is spent and does nothing.

Might require some re-wording on the query question and the intro text for morph itself.


SpadesNeil:
Someone fill me in on what exactly is a morph? Also, conflict.

Heroman3003:
Ventcrawling creature that can digsuise as any object or mob as long as it can get adjacent to it, but is pretty weak stat-wise. Basically was designed with idea of being a player-controlled pred creature.

Heroman3003:
https://cdn.discordapp.com/attachments/187013248309002240/615761783960895488/morph.gif

GIF for reference

Heroman3003:
I have no idea what causes that and it happens randomly. Happened also live when I was tasting build of another branch, so definitely not caused by morphs.

Novacat:
It seems to not be random in this case, since ff86124 and every commit after has those same two runtimes.

MarinaGryphon:
> [2020-01-16T05:03:36] Runtime in vent_scrubber.dm,114: Cannot read null.air_scrub_names
> proc name: broadcast status (/obj/machinery/atmospherics/unary/vent_scrubber/proc/broadcast_status)
> src: Air Scrubber (/obj/machinery/atmospherics/unary/vent_scrubber/on)
> src.loc: the floor (75,109,7) (/turf/simulated/floor/tiled/monotile)
> call stack:
> Air Scrubber (/obj/machinery/atmospherics/unary/vent_scrubber/on): broadcast status()
> Air Scrubber (/obj/machinery/atmospherics/unary/vent_scrubber/on): atmos init()
> Machines (/datum/controller/subsystem/machines): setup atmos machinery(/list (/list))
> Machines (/datum/controller/subsystem/machines): Initialize(182138)
> Master (/datum/controller/master): Initialize(10, 0)

Seems like there's a scrubber whose `initial_loc` wasn't set to an area, or an area that didn't have `air_scrub_names`.

Heroman3003:
There are gonna be mapping conflicts but I will hold off on fixing those until proper decision regarding event itself and what will be done about it is reached.

SpadesNeil:
tbh I love the idea of this

We can always yank it out of the event rotation if we hate it.

Heroman3003:
Alright, then Imma deal with the conflicts now. I also have plans for event of simular nature, just different 'wrapping', and for same purpose too. But that will come after this.

SpadesNeil:
If there's anyone opposed to this, speak now or I'mma merge this by the weekend.

PR #6439, opened 2020-01-13
Disable random blackouts during NIF adaptation

lbnesquik:
### **About The Pull Request**
- Disable blackouts from the list of effects of a NIF installation 
- Extend the timer by 5 minutes as a counterbalance. This is not necessary and can be reverted.

### **Why It's Good For The Game**
It disable something which hindered roleplay and gameplay potential from the procedure.
Blacking out was random, short and without any form of warning, meaning you can't really RP around it or just walk about without getting dragged to medical because you fell on the floor.

### DIFF
--- a/code/modules/nifsoft/nif.dm
+++ b/code/modules/nifsoft/nif.dm
@@ -255,7 +255,7 @@ You can also set the stat of a NIF to NIF_TEMPFAIL without any issues to disable
 			install_done = world.time + 1 MINUTE
 			notify("Welcome back, [owner]! Performing quick-calibration...")
 		else if(!owner)
-			install_done = world.time + 30 MINUTES
+			install_done = world.time + 35 MINUTES
 			notify("Adapting to new user...")
 			sleep(5 SECONDS)
 			notify("Adjoining optic [human.isSynthetic() ? "interface" : "nerve"], please be patient.",TRUE)
@@ -265,7 +265,7 @@ You can also set the stat of a NIF to NIF_TEMPFAIL without any issues to disable
 			stat = NIF_TEMPFAIL
 			return FALSE
 
-	var/percent_done = (world.time - (install_done - (30 MINUTES))) / (30 MINUTES)
+	var/percent_done = (world.time - (install_done - (35 MINUTES))) / (35 MINUTES)
 
 	if(human.client)
 		human.client.screen.Add(global_hud.whitense) //This is the camera static
@@ -294,9 +294,9 @@ You can also set the stat of a NIF to NIF_TEMPFAIL without any issues to disable
 				if(2)
 					human.Weaken(5)
 					to_chat(human,"<span class='danger'>A wave of weakness rolls over you.</span>")
-				if(3)
-					human.Sleeping(5)
-					to_chat(human,"<span class='danger'>You suddenly black out!</span>")
+				/*if(3)
+					human.Sleeping(5) //Disabled for being boring
+					to_chat(human,"<span class='danger'>You suddenly black out!</span>")*/
 
 		//Finishing up
 		if(1.0 to INFINITY)


TheDavestDave:
okay so once apon a time taking tramadol would stop you blacking out. and maybe rebalanced so that it does, something so people don't alt tab for 30mins. 

really if tramadol doesn't even stop black out then you can't even do anything with medical durining that time, even talking

Wickedtemp:
Replace it with dizziness/confusion. No blackout, still is mechanically crippling and if you're actually irl dumb can do some real damage by running into walls ten times.

TheFurryFeline:
I've never heard of tramadol actually preventing blacking out and I've had NIFs installed for virtually all of my chars, plus repeats from forgetting to get a backup for two churns.

lbnesquik:
@TheDavestDave @Wickedtemp @TheFurryFeline Tramadol I haven't seen ANYTHING that suggest tramadol has any effect on NIFs blackout. It does reduce pain but not the kind that floors you.

@Wickedtemp I could remplace it with like, confusion, if it is needed. If this PR get merged in its current state, we'll have just weakness that puts you on the ground and sudden pain.

Heroman3003:
There is no actual pain that floors you there, its just forced sleep, not pain passout.

lbnesquik:
`				if(1)
					var/message = pick(list(
								"Your head throbs around your new implant!",
								"The skin around your recent surgery itches!",
								"A wave of nausea overtakes you as the world seems to spin!",
								"The floor suddenly seems to come up at you!",
								"There's a throbbing lump of ice behind your eyes!",
								"A wave of pain shoots down your neck!"
								))
					human.adjustHalLoss(35)
					human.custom_pain(message,35)
				if(2)
					human.Weaken(5)
					to_chat(human,"<span class='danger'>A wave of weakness rolls over you.</span>")
				if(3)
					human.Sleeping(5)
					to_chat(human,"<span class='danger'>You suddenly black out!</span>")`

Heroman3003:
Yes, but thats the thing. Its not the pain that floors you, its the forced sleep. NIF, in my multiple setup experiences, never does enough pain to floor you, only to slow you down. Its only via waken and sleep.

SpadesNeil:
Perhaps replace with pain.

Ketrai:
Being slept is always annoying for rp because you can't even talk. I'm in favor of having this be just a paralyze similar to drinking a glass of neurotozin cocktail. Just make sure you can't crawl i guess.

TheFurryFeline:
I, too, am in favour of the forced unconsciousness being replaced by something else. Though we already have pain and weakness with it.

Heroman3003:
Imo, confusion seems like a good choice.

lbnesquik:
@SpadesNeil In effect, it is remplaced by pain given it reduces the number of events, not the frequency of them.

Wickedtemp:
If its pain, there's a larger possibility that people will simply start asking for painkillers to ignore it. There's a reagent to help with confusion as well, its not exactly potent enough to do that at the moment so it'll likely be buffed slightly at some point. 

You'd need to cause like 90+ hallos to not be circumvented by tramadol.

lbnesquik:
That sounds like a good idea NGL. Let people ask for meds instead of just sitting in the lobby/reception for half an hour.

Wickedtemp:
I mean they'd also still have to deal with 'weakness' as well, so from an IC standpoint...yeah it'd be kinda dumb of them to not just chill in a subacute. It's not a huge inconvenience when you only have to do it once and then your NIF remains between shifts.

lbnesquik:
In either case, you could at least talk and do a couple tid bits of RP.

Ketrai:
No responses for 21 days. Can this just get a jay or nsy on merging already?

SpadesNeil:
I think RNG is stupid in this context so I think this should be merged.

PR #6441, opened 2020-01-13
Fixes droppod spawning wrong subtype

Heroman3003:
AAA

Also buffs vorelaw chance for lost dogborg

### DIFF
--- a/code/modules/mob/living/silicon/robot/subtypes/lost_drone_vr.dm
+++ b/code/modules/mob/living/silicon/robot/subtypes/lost_drone_vr.dm
@@ -32,7 +32,7 @@
 	..()
 	laws = give_random_lawset_vore()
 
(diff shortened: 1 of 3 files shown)


code/modules/mob/living/silicon/robot/subtypes/lost_drone_vr.dm
```diff
@@ -41,7 +41,7 @@
 		prob(15);"corrupted",
 		prob(10);"bad")
 
-	var/vore_law = prob(50)		// 1/2 chance of having vore-related laws
+	var/vore_law = prob(80)		// 4/5 chance of having vore-related laws
```
Novacat: 80% is probably a bit too much. I would go with 60% at most.

Heroman3003: resolved

PR #6442, opened 2020-01-13
Fixes grub event end announcement not functioning

Heroman3003:
Turns out the event had a broken feature of announcing the list of locations where living grubs still remain after some time. This should fix that.


SpadesNeil:
This was a thing? We didn't have it? This whole time, we could have, and we didn't?
https://www.youtube.com/watch?v=cgwlJIo0In0

Heroman3003:
It seems like something that never worked in the first place and I only found it due to my own recent forays into events. Way I fixed it is, admittedly, less than ideal, but with solargrubs being entirely exclusive to the event for now, it shouldn't cause any issues.

PR #6451, opened 2020-01-15
IP reputation checking

Arokha:
Adds IP reputation checking to detect Tor, proxy, and VPN usage and block it if so configured.

It's disabled by default, but if enabled the default settings are to block all VPN/Proxy/Tor to connect unless the player has been a player for 5 days on a 'normal' connection so that people who legitimately sometimes use a VPN for *reasons* can continue to do so. You can also have it check reputations and just log bad ones, without disconnecting the users.

Whether or not it allows 'existing' players, the length of time they must have played, what's considered a 'bad' IP score, etc, are configurable. You **must** put an e-mail address if you use this, otherwise the service will likely ban you. This is the e-mail address they will send e-mails to if you're performing too many checks or they need to speak to you. 

Adds config options, here's a paste from the example config:
```
## IP Reputation Checking
# Enable/disable IP reputation checking (present/nonpresent)
#IP_REPUTATION

# Set the e-mail address problems can go to for IPR checks (e-mail address)
IPR_EMAIL whatever@whatever.com

# Above this value, reputation scores are considered 'bad' (number)
IPR_BAD_SCORE 1

# If you want the people disconnected. Otherwise it just logs. (present/nonpresent)
IPR_BLOCK_BAD_IPS

# If players of a certain length of playtime are allowed anyway (REQUIRES DATABASE) (present/nonpresent)
IPR_ALLOW_EXISTING

# And what that age is (number)
IPR_MINIMUM_AGE 5
```

As you can see, it's off by default, so if you're a downstream this won't change anything for you unless you decide to turn it on. If you want the features, just copypaste the new config lines out of the example and uncomment IP_REPUTATION.

Downstreams can replace the /client/proc/update_ip_reputation() proc with your own, if you'd like to substitute your own service! Just set the client's ip_reputation var at the end of your proc.

### DIFF
--- a/code/controllers/configuration.dm
+++ b/code/controllers/configuration.dm
@@ -104,6 +104,13 @@ var/list/gamemode_cache = list()
 	var/panic_bunker = 0
 	var/paranoia_logging = 0
 
+	var/ip_reputation = FALSE		//Should we query IPs to get scores? Generates HTTP traffic to an API service.
+	var/ipr_email					//Left null because you MUST specify one otherwise you're making the internet worse.
+	var/ipr_block_bad_ips = FALSE	//Should we block anyone who meets the minimum score below? Otherwise we just log it (If paranoia logging is on, visibly in chat).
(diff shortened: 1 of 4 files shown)


KasparoVy:
I've seen a few servers use this to great success, -games sense and I'm glad it'll be put to work here too

PR #6483, opened 2020-01-21
Port of Chompstation Suit Cycler stat fixes

Rykka-Stormheart:
Port of https://github.com/CHOMPStation2/CHOMPStation2/pull/184

### DIFF
--- a/code/game/machinery/suit_storage_unit.dm
+++ b/code/game/machinery/suit_storage_unit.dm
@@ -994,12 +994,18 @@
 				helmet.name = "engineering voidsuit helmet"
 				helmet.icon_state = "rig0-engineering"
 				helmet.item_state = "rig0-engineering"
+				helmet.max_heat_protection_temperature = HELMET_MAX_HEAT_PROTECTION_TEMPERATURE
+				helmet.armor = list(melee = 40, bullet = 5, laser = 20, energy = 5, bomb = 35, bio = 100, rad = 80)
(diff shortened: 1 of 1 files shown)


Rykka-Stormheart:
Reopening due to Travis failure thanks to active edges.

Heroman3003:
Isn't the entire point of suit cycler's paintjob that it doesn't actually changes stats of the suit and just the appearance, so if you want a better spacesuit you need to actually get it rather than transform it?

Novacat:
I am going to decline this for a few reasons.
- It makes it impossible for found special voidsuits, like the blood-red voidsuit and scorched voidsuits, to be used by non-human crew, since any attempts to cycle them would remove the stats
- I want to encourage people to cycle special voidsuits if they do find them, because it makes no sense for station crew to be walking around in syndie voidsuits. The current cycling mechanism is the best option considering that there is no way to properly recolor voidsuits.
- It is not consistent, the bloodred and scorched cycler remains cosmetic only. This is going to generate a lot of confusion
- If it were made consistent, then it would be too easy for station crew to acquire special voidsuits that are supposed to only be acquired in events and expeditions, since you can find emags in the trash

TheFurryFeline:
Then maybe rename suit cyclers to suit painters.... :V

Heroman3003:
They are cyclers because they do the "decontamination **cycle**". The whole "species adjustment" and "painting" are extras.

PR #6484, opened 2020-01-21
Removes escape routes from the xenobio cells (in form of ventilation)

Heroman3003:
Those cells, even in 'normal' circumstances contain creatures capable of ventcrawling. And in general, many of supposed 'xenobio exhibits' are also capable of it. So they definitely shouldn't have a free escape card to literally anywhere on the station from their cell.

Also reroutes pipes slightly in staircase to not run under stairs themselves

### DIFF
--- a/maps/tether/tether-02-surface2.dmm
+++ b/maps/tether/tether-02-surface2.dmm
@@ -17648,18 +17648,20 @@
 /turf/simulated/floor/tiled/white,
 /area/rnd/outpost/xenobiology/outpost_slimepens)
 "aHe" = (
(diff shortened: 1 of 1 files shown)


Novacat:
My only concern is how easy it would be to restore the atmos of a compromised cell without any vents/scrubbers? Any comments on that?

Heroman3003:
Same as it would in any section of maintenance. Also, ones with external shutters still have Big scrubbers. Alternatively, if entire xenobio is compromised, just open up the cells and keep them open till you're done.

PR #6492, opened 2020-01-23
Fixes and re-implements Rogue Mining (Belt Mining)

Heroman3003:
Here it is. The good old roguelike mining experience. Err, at least as close to good old as I could get. I did adjust some things and touched up on balance and scaling a little, but as a whole, its there and its working.

For those that do not know what this is: Rogue Mining also known as Belt Mining is mining on sorta-procedurally generated areas consisting of asteroid belts before returning back to station and resetting said areas, with both difficulty and danger rising based on your results (how much ores you mined and mobs you killed).

To start, you might want to order belt mining crate or pop a visit to Research for some anti-material weapons and pressure locks. But once you're all geared up, head to cargo and in south you will find a new little door. In control room, area scanner console will allow you to scan for a sector. Do that, and shuttle will be routed to a freshly generated sector of asteroids and mobs. After that, head outside through the airlock and after that into the shuttle. Do note, door is not automatic. Don't forget to close it.

Once you're there, do what you do usually: mine. Use binoculars to scout out asteroids, and navigate towards them. There are some dangers prowling around though. Once you feel like you've mined enough, go back and scan for another sector. As you pass certain thresholds, difficulty will increase, mob spawns will get more numerous and dangerous, but so will increase amount of precious ores and chance for loot caches instead of asteroids to appear.

Some notes: You can't scan for a new sector while shuttle is in the active sector or while there is someone alive mining there. Scanning for new sector will eventually cause previous sector to destroy everything in it, so if someone died and you have hopes of recovering them, best not begin the scan as that very moment, they are permanently gone. There is a side entrance from maintenance, with access provided for medical personell and people with xenoarch access. Medical personnel for rescues, xenoarch on account of artifacts and digspots being among possible spawns in those rocks.


KasparoVy:
Hopefully this comes with a tidy few wiki updates, looks glorious.

Arokha:
I remember writing this. Ahh, good times. I bet it could be improved with maploading instead of the fully math-y methods I used to make the spawnables. Not suggesting you do that. Just sayin'.

Heroman3003:
I might experiment with that later, but for now main goal was to just get what we already have running and up-to-speed.

Wickedtemp:
Do suitsensors extend to those areas?

Heroman3003:
No

Wickedtemp:
Given that its going to be much more dangerous not only from the mobs, but being in space as well, I'd recommend it. We used to have many, many, many cases of someone forgetting internals and passing out, and then dying, because nobody knew. 

Happens sometimes with UD, but its always been much more common with belt-mining.

Heroman3003:
Main reason I didn't was because A) it is now spread out over multiple z-levels which would create way too much unnecesary confusion over "where tf is that" and B) any area that had a left-behind body will not remove it immedeately, but it will still show up on sensors even after rerouting. At that point, it is already beyond recovery as you cannot reroute back, but it is only erased later when area AFTER next is generated.

Ketrai:
Any reason this or is still floating around?

SpadesNeil:
image

Heroman3003:
Just appearted. Gonna fix that up now.

Heroman3003:
It shouuuld be fixed now.

KasparoVy:
OOHHH FUCK NICE! Super stoked to try this later!

PR #6494, opened 2020-01-23
Actually fixes everything dying in water

Heroman3003:
THis proc shouldnt even be a thing anymore why why


Heroman3003:
Upstream has a better fix.

PR #6508, opened 2020-01-26
A ban-flinging QoL update

Verkister:
Sabotages the exploitable ban-dodging mob-lock "feature" of the ban system. Like dude we're banning a client, not a mob so shut up and let us work.

Hopefully won't break anything but probably won't.

### DIFF
--- a/code/modules/admin/DB ban/functions.dm
+++ b/code/modules/admin/DB ban/functions.dm
@@ -50,9 +50,10 @@ datum/admins/proc/DB_ban_record(var/bantype, var/mob/banned_mob, var/duration = 
 	if(query.NextRow())
 		validckey = 1
 	if(!validckey)
-		if(!banned_mob || (banned_mob && !IsGuestKey(banned_mob.key)))
-			message_admins("<font color='red'>The validmob system attempted to block [key_name_admin(usr)] from banning [ckey], but the devs jammed a screwdriver into its workings. Please ban whomever deserves it.</font>",1) //VOREStation Edit
-			//return //VOREStation Edit
+		if(!banned_mob || (banned_mob && !IsGuestKey(banned_mob.key))) //VOREStation Edit Start.
+			var/confirm = alert(usr, "This ckey hasn't been seen, are you sure?", "Confirm Badmin" , "Yes", "No")
+			if(confirm == "No")
+				return //VOREStation Edit End
 
 	var/a_ckey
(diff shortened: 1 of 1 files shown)


SpadesNeil:
Preemptive bans should probably still be logged in the admin notes as being preemptive bans. We should hopefully not need to use this system very often.

KasparoVy:
Huh. Yeah, that's a bit wack. Properly carried out and documented pre-emptive bans shouldn't be blocked I figure.

Leshana:
I'm unsure what the intent of this is.   Are you often wanting to ban ckeys of people who have never joined?

SpadesNeil:
>I'm unsure what the intent of this is. Are you often wanting to ban ckeys of people who have never joined?

That is the intent, it would seem. I am not sure if this is wise or not.

Verkister:
Yeah that was basically the intent in this PR during the saratov guy's shenanigans before we got the VPN blocker implemented. This also removes one unnecessary hurdle on applying bans after the incident round and whatnot as it currently *requires* you to manually spawn in an empty mob and type the ban target's ckey into it before you can do anything and I don't think anyone sees that as a necessary step.

Screemonster:
Yeah, we can circumvent it easily when we actually need to, it's just annoying. About the only useful purpose of it is that it'll catch typos in the ckey.

SpadesNeil:
>Even if it popped up a "this ckey hasn't been seen, are you sure?" dialog box it'd be better than the current system.

Let's do that then.

code/modules/admin/DB ban/functions.dm
```diff
@@ -51,8 +51,8 @@ datum/admins/proc/DB_ban_record(var/bantype, var/mob/banned_mob, var/duration =
 		validckey = 1
 	if(!validckey)
 		if(!banned_mob || (banned_mob && !IsGuestKey(banned_mob.key)))
-			message_admins("<font color='red'>[key_name_admin(usr)] attempted to ban [ckey], but [ckey] has not been seen yet. Please only ban actual players.</font>",1)
-			return
+			message_admins("<font color='red'>The validmob system attempted to block [key_name_admin(usr)] from banning [ckey], but the devs jammed a screwdriver into its workings. Please ban whomever deserves it.</font>",1) //VOREStation Edit
```
SpadesNeil: See last couple of comments.

PR #6511, opened 2020-01-28
Limits the locations at which drone pod can drop

Heroman3003:
title


TheFurryFeline:
Active edges on z-level 20. Huh.

Rykka-Stormheart:
> Active edges on z-level 20. Huh.

Yeah, that's been fucking every single PR I've opened, the active edges test.

There's no way to disable the active edge/mapping test unless your PR edits a .dmm, is there?

Heroman3003:
It's debrisfield. In fact, most runtimes we get are either from wolf den PoI in the Underdark or the active edges from the Alien Ship in debrisfield. Because those PoIs were mapped badly.

Screemonster:
This is for the random event, right?

Why have fixed landmarks for it? Just make it reroll the scatter if there's a dense turf within the footprint of the pod. Then it'd still be able to land in a large selection of areas but wouldn't be able to land directly on station sections.

Heroman3003:
Main issue is that it will still be able to land right next to station section, which will still cause extensive damage. Only options are either to make it land only a safe distance via use of landmarks or remove the explosion upon impact.

I picked first option primarily because the code for random selection of a spot was an extreme hacky mess that was adapted from a pretty snowflakey meteoroid code.

Screemonster:
Exploding a hole in the wall is fixable. Overwriting sections of wall is a bigger problem. Don't wrap the station in cotton wool to prevent engineers from ever having to work.

Hell, I'd say it'd suck for outside construction projects too if they put one in the same place as a landmark without realising, but people rarely build outside on the tether 'cause the atmos is a pain. It's easier to build in space lol.

but seriously maybe it's because I learned this game on box and meteors blowing holes in the outside edges of the station was something we had to deal with on the regular but dear fucking god, things _are_ allowed to happen on the station

Heroman3003:
I personally did not want this event to be as regular cause of damage as it turned out to be and thats not even counting it overwriting places. There's more than enough annoying random chores for engineers in the random event pool as it is and drone pod in the first place was not meant to be another one, but an enabling opportunity for voreborgs to be more common. Especially when a medium level event with higher chance causes MORE damage then rare High level event.

Screemonster:
God forbid environmental hazards exist in space station 13.

Heroman3003:
Oh yes, screw me for trying to design at least one random event that isn't:
Unfun
Unrewarding
Unengaging
Forces people to deal with it
Counter-productive to any RP actually happening
and completely impossible to plan around.

Because that's what every other 'engineering random event' is, and thats why the department sucks miserably right now. So sorry for not wanting to contribute to the problem even further, while also trying to dillute the pool with an event that is entirely optional and is actually about giving people more options, not taking away all options they already had and forcing them to deal with a chore because RNG says fuck you,

Wickedtemp:
Scree, you're kinda being a bit of a jerk here. The dude that made the PR is now saying that the event had unforeseen effects that weren't intended. For them to want to fix that is entirely valid. 

You don't have to be snarky as fuck about it. 

"I don't want this thing to accidentally end up causing horrible phoron fires." Isn't the same as "OH WELL I GUESS SCREW HAZARDS AM I RIGHT" and you know it.

SpadesNeil:
Honestly I think this is fine. Needs conflicts resolved.

Heroman3003:
Resolved.

PR #6515, opened 2020-01-29
Redundant Warning Removed

TheFurryFeline:
Changelog Notes: 

- Removes the big, honking red text about archive being slated for removal. We're not removing entries from the database.


TheFurryFeline:
Wow, fuck you, active edges. I didn't even touch that map file.

PR #6519, opened 2020-01-31
QoL to supplypacks related to Belt Mining

Heroman3003:
Basically extra followup to #6492.

Feel like this might need its own discussion if any.

Increases price of belt mining crate by 10, adds a pair of binoculars into it, allows xenoarch access to open it.

Adds jetpack crate, contains one empty jetpack. Can be opened by mining, EVA, xenoarch or explorer access.

### DIFF
--- a/code/datums/supplypacks/misc_vr.dm
+++ b/code/datums/supplypacks/misc_vr.dm
@@ -6,12 +6,16 @@
 			/obj/item/weapon/cell/device/weapon = 2,
 			/obj/item/weapon/storage/firstaid/regular = 1,
 			/obj/item/device/gps = 2,
-			/obj/item/weapon/storage/box/traumainjectors = 1
+			/obj/item/weapon/storage/box/traumainjectors = 1,
+			/obj/item/weapon/tank/jetpack = 1,
+			/obj/item/device/binoculars = 1
 			)
-	cost = 50
+	cost = 80
 	containertype = /obj/structure/closet/crate/secure/gear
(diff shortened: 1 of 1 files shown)


Screemonster:
If we were running any map other than the tether I'd have concerns about effectively handing out jetpacks for 30 cargo points apiece.

Heroman3003:
Well, they are empty and miners should already have access to them in EVA. I personally do not consider jetpacks in general to be as much of a critical object as they seem to be positioned. But I could potentially split the jetpack aspect into its own crate and at like 50 point cost, while dropping the increase of beltmining crate cost to 10 (for binoculars).

Reason I want them to be obtainable like that is because new belt mining from #6492 pretty much requires ability to freely navigate in space, especially on early difficulties.

Screemonster:
Yeah, honestly I'd split them, if they've got access to some at roundstart but replacements are fairly pricey then that's fine.

If we were still on box and I could get a jetpack that cheap I'd go nuts for that deal, haha. Raid the white ship all day every day (Scree got a fucking _pulse rifle_ out of it once)

Heroman3003:
Done. Jetpack its own crate now, for 75 points.

Novacat:
I remember when there was a huge fuss over the Pathfinder getting a pair of binoculars. Times certainly change.

Heroman3003:
Main reason for them being originally mapped in in mining was belt mining itself in the first place, which is why pair is now present in the belt mining dock and available in crate with specifically belt mining supplies. Now needed more than ever.

PR #6520, opened 2020-01-31
 A couple griff prevention measures

Verkister:
Felt like being evil today.

(makes some uncreative grief methods cause a different end result for newjoins)

### DIFF
--- a/code/game/machinery/atmoalter/canister.dm
+++ b/code/game/machinery/atmoalter/canister.dm
@@ -328,7 +328,7 @@ update_flag
 				if(istype(src,/obj/machinery/portable_atmospherics/canister/phoron))
 					if(usr.client)
 						var/client/C = usr.client
-						if(C.player_age)
+						if(!isnull(C.player_age))
 							if(C.player_age < 1)
 								message_admins("[key_name(usr)] has attempted to open a phoron canister as a newjoin. ([usr ? "<a href='?_src_=holder;adminplayerobservecoodjump=1;X=[usr.x];Y=[usr.y];Z=[usr.z]'>JMP</a>" : "null"])")
 								if(ishuman(usr))
--- a/code/game/machinery/atmoalter/canister.dm
+++ b/code/game/machinery/atmoalter/canister.dm
@@ -328,13 +328,14 @@ update_flag
 				if(istype(src,/obj/machinery/portable_atmospherics/canister/phoron))
 					if(usr.client)
 						var/client/C = usr.client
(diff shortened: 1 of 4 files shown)


Heroman3003:
Considering all procs have same checks and same result, it could possibly be all wrapped into a single proc that returns 1 if it successfully detected and punished the scum (and would allow it to be used easier in other cases). Also, would add the welder-on-fueltank to the list personally.

SpadesNeil:
As hilarious as this is, I'm gonna have to veto the current idea. I like the idea of alerting staff and maybe quietly preventing the mechanic from working if a certain thing is toggled on--perhaps a config option. I do not approve of it deleting limbs. I would also like a way to override this so if a new player is legit, they aren't just unable to use these important game mechanics at all in the event that, say, an Engineer player requests their aid.

Or if you want to make it IC you could make it return a message saying something along the lines of, "You don't have the skill to use this machinery" or something.

KasparoVy:
Could probably make a proc called 'anti_griff' that lets you pick a variety of nefarious shit to do to people, then just call it in various locations for the sake of OOP

Leshana:
Good point, there are legitimate times to use these actions.  I suppose the age check would make it less a bother, but indeed we probably shouldn't have it delete limbs.

code/game/machinery/atmoalter/canister.dm
```diff
@@ -324,6 +324,19 @@ update_flag
 			else
 				release_log += "Valve was <b>opened</b> by [usr] ([usr.ckey]), starting the transfer into the <font color='red'><b>air</b></font><br>"
 				log_open()
+				//VOREStation Evil Edit
+				if(istype(src,/obj/machinery/portable_atmospherics/canister/phoron))
+					if(usr.client)
+						var/client/C = usr.client
+						if(!isnull(C.player_age))
+							if(C.player_age < 1)
```
Arokha: I assume you mean this to determine if the `player_age` is set or not, as in the case of it not being set for a server not using the database, but unset `player_age` is not null, it's the string "(Requires Database)" as per `client defines.dm:50`. By itself that is only slightly grievous except for the fact that trying to less-than a string on the next line will **always runtime** if client age is not set.

So, effectively, these changes make it impossible to perform these actions at all on any server that either has a database that is down, or does not use a database to track player age.

I can't imagine this was tested, because to not have noticed this would require you to have set up a local DB on your testing machine and overlooked this because you were using the database.

PR #6524, opened 2020-01-31
Makes syndie beacon found in xenoarch use virgo-specific subtype

Heroman3003:
title


SpadesNeil:
Remind me what the difference is? This sounds like a bug fix.

Heroman3003:
Original actually makes person into mechanical traitor with uplink and gives them randomized goals. Often with stuff like go kill X.

Virgo-specific version gives the person uplink, but doesn't actually make them an antag or give them goals.

Screemonster:
Yeah, we have our own beacons for adminspawn because we don't want people randomly getting antag during extended. They shouldn't be getting the other type as randomspawn either.

PR #6537, opened 2020-02-04
Sundry Bugfixes - Feburary

Leshana:
####  Fix Runtime in areas.dm,307: Cannot read null.len 
area.ambience is never expected to be null.
#### Fixes Runtime in living_vr.dm,428: Cannot read null.name
- This was happening when null is passed for belly. Now it will print *null*
- Also fixed root cause (at least one of them). Animals can't have vore attributes enabled when vore_active = 0 because then they don't get a belly.
#### Fix character setup preview icons rendering prior to SSAtoms initialization
- Rendering the characters involved creating mannequin mobs, prior to SSAtoms init these would not be initialized and thus have no DNA.  This prints many warning messages, causes runtimes, and ultimately the icon does not render correctly.
- Its fixed here since it seems that needing to use mobs prior to initialization is a pretty unique edge case.  Comment added to explain to future us.


KasparoVy:
Cheers, quite a strange bug with the preview icons!

PR #6542, opened 2020-02-07
Prevents some teleporting abuses with taur riding

Heroman3003:
When using _uncalibrated_ gateway, hand tele portals or command teleporter as a taur, all your riders are dismounted automatically and left behind.

All teleportation now takes into account the bluespace backpacks of the taur's riders for failing chances.

Bluespace harpoon's "send" mode now has increased chance of failure if you have riders.

### DIFF
--- a/code/datums/helper_datums/teleport.dm
+++ b/code/datums/helper_datums/teleport.dm
@@ -158,6 +158,13 @@
 		precision = rand(1,100)
 
 	var/list/bagholding = teleatom.search_contents_for(/obj/item/weapon/storage/backpack/holding)
+	//VOREStation Addition Start: Prevent taurriding abuse
+	if(istype(teleatom, /mob/living))
+		var/mob/living/L = teleatom
+		if(LAZYLEN(L.buckled_mobs))
+			for(var/mob/rider in L.buckled_mobs)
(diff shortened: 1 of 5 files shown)


Ketrai:
Just a one in five chance to fail and screw over anyone who isn't a regular to harpoon abuse. Seems a bit excessive? Literally results in a 64% chance not to get screwed over teleporting in and out an area. A longer cooldown would've been better balance. 

Does this all even take into consideration held micros or people in bellies?

Novacat:
It only increases the failure chance if you have a buckled rider, and, you are more than welcome to make a balance PR to adjust the number downward.

Ketrai:
I can read the code :v  it's still needlessly harsh either way. If someone isn't aware od this mechanic they have a 20% chance to get yeeted into space or some other dangerous location without knowing about the risk. And it wouls take a lot of trial and error before people realise it's taur riding causing this. Harpoon fails are meant to punish spamming them, not to punish power gamers. This is just silly. At least add a safety warding, or go for a lesser method of punishment like a longer cooldown period. The main issue was gateway abuse and now this is an issue?! And for the love of god, please just keep prs open when they actually touch balance and aren't just a simple bugfix. There's already few people that frequent github, and closed prs fly under the radar. All prs should be given a chance to be scrutinised, maintainer prs should not be an exception?

PR #6548, opened 2020-02-08
fix conflict causing compile error

Meghan-Rossi:


code/game/objects/structures/stool_bed_chair_nest/bed.dm
```diff
@@ -46,12 +46,8 @@
 	// Base icon.
 	var/cache_key = "[base_icon]-[material.name]"
 	if(isnull(stool_cache[cache_key]))
-<<<<<<< HEAD
-		var/image/I = image(icon, base_icon) //VOREStation Edit
-		//var/image/I = image('icons/obj/furniture.dmi', base_icon) //From Polaris Sync. Not sure if this is a better way of doing it or not. Uncomment if so.
-=======
 		var/image/I = image(icon, base_icon)
->>>>>>> 792245a... Merge pull request #6665 from Woodratt/02022020_civilianmaptweak
+		//var/image/I = image('icons/obj/furniture.dmi', base_icon) //From Polaris Sync. Not sure if this is a better way of doing it or not. Uncomment if so.
```
Novacat: This line can probably be deleted. It is no longer on Polaris so no longer needed.

PR #6551, opened 2020-02-10
Undoes unnecesary nerfs on pizzas

Heroman3003:
Pretty much title. In retrospect since then, pizzas really weren't the issue, but cost of the crate, which was remedied since then. The frozen pizzas are also fundamentally broken mechanically and fixing them would be too much work for their 'worth', both sprite- and recipe-wise.

### DIFF
--- a/code/modules/food/food/snacks_vr.dm
+++ b/code/modules/food/food/snacks_vr.dm
@@ -490,6 +490,7 @@
 	name = "wolpin cube"
 	monkey_type = "Wolpin"
 
+/*
 /obj/item/weapon/reagent_containers/food/snacks/pizza/margfrozen
 	name = "frozen margherita pizza"
 	desc = "It's frozen rock solid, better thaw it in a microwave."
@@ -669,18 +670,7 @@
 
 /obj/item/weapon/reagent_containers/food/snacks/slice/vegcargo/filled
 	filled = TRUE
-
-/obj/item/pizzabox/margherita/Initialize()
-	pizza = new /obj/item/weapon/reagent_containers/food/snacks/sliceable/pizza/margcargo(src)
-
-/obj/item/pizzabox/vegetable/Initialize()
-	pizza = new /obj/item/weapon/reagent_containers/food/snacks/sliceable/pizza/vegcargo(src)
-
-/obj/item/pizzabox/mushroom/Initialize()
-	pizza = new /obj/item/weapon/reagent_containers/food/snacks/sliceable/pizza/mushcargo(src)
-
-/obj/item/pizzabox/meat/Initialize()
-	pizza = new /obj/item/weapon/reagent_containers/food/snacks/sliceable/pizza/meatcargo(src)
+*/
 
 // food cubes
 /obj/item/weapon/reagent_containers/food/snacks/cube
--- a/code/modules/food/recipes_microwave_vr.dm
+++ b/code/modules/food/recipes_microwave_vr.dm
@@ -196,6 +196,7 @@
 	)
 	result = /obj/item/weapon/reagent_containers/food/snacks/sharkmeatcubes
 
+/*
 /datum/recipe/microwave/margheritapizzacargo
 	reagents = list()
 	items = list(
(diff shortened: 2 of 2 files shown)


SpadesNeil:
I think this is fine.

Ketrai:
Rather than complain about the pr itself I'm just going to complain again ghst this was merged withiut giving people ample time to evaluate the pr. Is it too hard to ask for a 24 hour grace period? Why do some folks have their prs quick merged whilst other prs with similar simple changes have to wait for daye?

Ketrai:
Also can we please not call stuff fundamentally broken unless they are actually? You just microwave the thing which is pretty logical? Described arguments about sprites rather than superficial commenting would be appreciated too.

SpadesNeil:
>The frozen pizzas are also fundamentally broken mechanically and fixing them would be too much work for their 'worth', both sprite- and recipe-wise.

That's why.

SpadesNeil:
Also, if you have a problem with it, it's not too late to input an argument. We can always revert this.

Ketrai:
Well I'll pull up the reasons frozen pizzas were added in the first place.

- mechanically identical compared to the pizzas the chef can make. Which take a lot more effort, so they should be superior in some way. I often hear this is best done by making the chef food taste better. Frozen pizzas give less food, and taste worse. You can still use them for pizza parties, the main issue is the relative rarity of microwaves. Instead of removing the feature,  we can just make them cook themselves like sin-pockets. Doesn't require any new code. 

- suggesting balancing through price alone is poor reasoning. Cargo points are often uncontested, and it's too easy to get ludicrous amounts of points via mining. People will order five crates of pizza wether they're 25 points, or 100.

Heroman3003:
Frozen pizzas specifically were broken on account of still freezing people even after being microwaved. 

As for sprites, they plainly were invisible when in the boxes. Both in frozen and unfrozen states.

And if people order the pizzas at ludicrous prices, that's their issue really? I don't see food crates being ordered when there's an active chef, other than for use just within cargo itself, which tbh is perfectly fine. If they want to waste that much points, that's also fine. Cargo food's disadvantage was always limited selection of what you can get and very inefficent price in comparison, IMO, not the supposed 'it must be worse'.

Ketrai:
- emphasising "WERE" the sprite and frost oil issues have been resolved, yet you're still claiming them to be fundamentally broken. So what's the argument then?

- the issue with pizza is that it's a bit too good of a food. Unlike burgers it requires growing tomatoes, mushrooms, vegetables. You get five of them, you can stack them to transport 5 full pizzas at the same time, and they're very nutritious. Pizza crates put all other food crates to shame. Pizzas hardly stay in cargo, they often get distributed amongst all departments. 

- why encourage wasting points? If cargo is going to feel like it's okay to order 250 points worth of pizzas, there might not be enough points left for other orders. Including chef orders! Considering cargo is the best place to get raw ingredients and animals quickly.

Ketrai:
I'd rather have frozen pizzas easy to unthaw, or come unfrozen but still nerfed nutrition and taste wise.

Heroman3003:
The frost oil issues were never resolved. We just removed the use of frozen pizzas entirely and started sending the unfrozen versions, at which point why even bother. And last I heard, sprite issues with those were still present as well.

If cargo is being dumb with points, its the problem of cargo being dumb, not problem of stuff being expensive. You're acting like pizzas are somehow necessety when they are just a sink for extra points or food source for chefless rounds. Similarly to what stolen crates and toy crates and other expensive crates are.

Overall, my opinion was and still remains that all the nerfs to pizza crate other than price were jsut a kneejerk that didn't add anything good to the game or fix any issues that were present.

Ketrai:
Chefless rounds see only pizza crates ordered though. We got so many alternative food sources. At least frozen pizzas make it less of a no-brainer. At least with frozen pizzas people actually ask the chef for one rather than cargo. It's a popular food, in large amounts, with less than five minutes wait time usually. I've seen rounds where people skip asking the chef altogether because if the chef doesn't have them already on their counter, it's simply easier and more convenient to ask cargo. That pretty much doesn't happen anymore with frozen pizzas because they're much less appealing.

PR #6556, opened 2020-02-10
Adds new airlock/door sounds!

Rykka-Stormheart:
Port of https://github.com/CHOMPStation2/CHOMPStation2/pull/108 from downstream!

Adds multiple distinct airlock/door sounds, so that you can recognize what type of door is opening thanks to audio cues.


Ketrai:
These are very loud. More fit for an industrial space station or something, so other than maintenance doors I don't see these being that nice.

Rykka-Stormheart:
> These are very loud. More fit for an industrial space station or something, so other than maintenance doors I don't see these being that nice.

I've not had an issue with volume, though, I *do* have a hearing issue.

They're live on CHOMPStation and I've not heard of any volume issues/complaints besides one specific door subtype. Iirc?

Heroman3003:
The whole 'type of door' system is pretty a pretty dumb idea when you consider that every door on the station is mechanically the same and only difference is access and paint. Also, it will only be affecting mapped in airlocks, which only adds to dumbness and confusion of it. Plus, most of the sounds are way too different from each other. Current door sounds have one very good aspect: distinctiveness. There's nothing like that one sound. When you throw in several dozens random noises, it sounds just messy and unrecognizable. Especially with majority of new 'closing' sound basically sounding like a different opening sound. There's also issue of some sounding notably longer than others which is much more an IMO, but just will be more grating to open those types.

Also an agree, some of those are just ridiculously loud.

Rykka-Stormheart:
> The whole 'type of door' system is pretty a pretty dumb idea when you consider that every door on the station is mechanically the same and only difference is access and paint. Also, it will only be affecting mapped in airlocks, which only adds to dumbness and confusion of it. Plus, most of the sounds are way too different from each other. Current door sounds have one very good aspect: distinctiveness. There's nothing like that one sound. When you throw in several dozens random noises, it sounds just messy and unrecognizable. Especially with majority of new 'closing' sound basically sounding like a different opening sound. There's also issue of some sounding notably longer than others which is much more an IMO, but just will be more grating to open those types.
> 
> Also an agree, some of those are just ridiculously loud.

I mean, sure, as of right now every door on the station is mechanically the same.

Adding the sounds will differentiate them so, I don't see the point of this argument, other than "status quo"?

And the current airlock sounds sound distinctive, sure, but you can't tell for instance if someone just opened the CD's office door, or if someone's opening the vault or opening a secure airlock. Now, you can, via audio cues.

Again, I've not heard any complaints about the sounds on Chomper's end, and they're live over there. Go poke your head in and mess around if you want to see what it's like in-game.

Heroman3003:
Well, I don't play on chompers and honestly do not care what they have or don't have. Reason I'm voicing my disagreement is because it's being suggested here.

The supposed sounds once again apply exclusively to mapped in airlocks. You could build new one and it would have generic sound. Even if it is supposedly 'security'. You could change the 'type' via painter and it wouldn't change. Why would the sounds be different?

I don't think there's need to distinct what's being opened, especially since that noise pretty much entirely applies to 'normal' opening. If anything, it takes away from the idea of "I heard airlock open, but I don't know which one was it, only the direction, should I investigate or not".

I think this change is bad idea. Plain as that.

SpadesNeil:
I don't particularly like this either. It's a neat idea, but doesn't serve a purpose that I wish to see fulfilled.

Now, if airlocks could be made out of different materials, that would be pretty cool.

Ketrai:
I wouldn't say our doors are that insanely distinctive. Our glass doors use almost the same if not the same noise for sliding doors for example. And the noise is very silent. I wouldn't mind seeing big heavy sounding doors to add to the eerie atmosphere of certain POIs or away maps. They're superb for that. Otherwise avoid change just for the sake of change, there should be some good reasoning before changing the door sounds.

SpadesNeil:
Double doors could actually make use of new sounds. So could vault doors. Is it really just a reskin, though? I thought vault doors were their own type.

SpadesNeil:
Elaborating on my review requesting changes, this feature is a good feature, but its appliance should be more specialized.

Double doors, vault doors, alien doors, or any variety of special doors should be the application in which this is used. Standard station airlocks should not be affected. Their departmental differences are not very much; not enough to justify new sounds. They're just the same airlock with different paint jobs.

Arokha:
> You should make a new _vs file.

_vr

Heroman3003:
I believe Vault door specifically is a reskin, but that one could use a different sound since its extremely rare and basically most important door on station. Double doors I'd say shouldn't really use different sound, primarily because their use being pretty much identical to use of regular doors in station layout. Although material/alien airlocks having different sound is great idea.

SpadesNeil:
This should be closed under its current state. The idea is neat but the scope is too broad. Also, coding practices are being disregarded.

TheFurryFeline:
I'd do the requested changes but I ain't familiar with sound code ;w;

Novacat:
I agree on the scope being way too broad, and considering that most airlocks tend to be made out of the same materials it tends to not make much sense either. Now, if the ambitiousness was toned back and, say, sounds were just added for normal airlocks, maintenance airlocks, and external airlocks, I can see this being much more palatable.

PR #6560, opened 2020-02-12
Adds Maintenance Predator random event

Heroman3003:
Very similar to Morph Spawn event in nature. It sends a notification to all current ghosts, offering to play as maintenance predator and if someone agrees, they get put into the mob, otherwise event is cancelled. Difference from Morph Spawn is that spawn locations are in bigger sections of maintenance around the station rather than near secluded vents, and that there is a selection of mobs to pick from when you are spawned in.

PR that added Morph Spawn for reference: #6429 

Available mobs right now: bumblebee, voracious lizard, giant frog, giant rat, juvenile solargrub, red panda, fennec, fennix, jelly blob and wolf.


Ketrai:
I like this. I've suggested something sinilar countless times in the past. As long as people behavw and maint preds are treated similarly to werewolves. Don't want anyone valid hunting em or pref breaks.

SpadesNeil:
I'm a little nervous about adding antag-lite stuff, but its vorish nature makes me want to say "fuck it" and try it anyway. We can always remove it if regular players hate it, and not just the ones who bitch about anything happening ever.

code/modules/events/maintenance_predator_vr.dm
```diff
@@ -0,0 +1,73 @@
+/datum/event/maintenance_predator
+	startWhen = 1
+	announceWhen = 50
+	endWhen = 100
+	var/announceProb = 35
+	var/list/possible_mobs = list("Space Bumblebee" = /mob/living/simple_mob/vore/bee,
+								  "Voracious Lizard" = /mob/living/simple_mob/vore/aggressive/dino,
+								  "Giant Frog" = /mob/living/simple_mob/vore/aggressive/frog,
+								  "Giant Rat" = /mob/living/simple_mob/vore/aggressive/rat,
+								  "Crate Mimic" = /mob/living/simple_mob/vore/aggressive/mimic,
```
SpadesNeil: If we add the player as a mimic, let's make sure we do two things.

One I will mention here because it specifically pertains to mimics, delete any remaining mimics that may or may not be on the map.

TheFurryFeline: That'll be the ones to the east of the surface armoury in the hidden room

Heroman3003:
There are no such things, but I can make sure that mob simply doesn't have an AI set so in case player disconnects/leaves, it will be completely inactive without issues.

There is still one mimic on the map, but it's specifically inside the wall cache between maint and security and generally impossible to 'stumble' into without actively trying to weld there.

Heroman3003:
For the concern of being too antag-like, I actually share it, and that is why the "expect to be treated as vore predator" warning is placed in the "do you want to play as it?" messsage, and no remarks towards full on aggression are made in starting message.

code/modules/events/maintenance_predator_vr.dm
Heroman3003: Resolved by removing the mimic from the list. Its way too close to the morph in nature, which is separate event, and I do think that some in-wall caches with mimics both now and in future should stay as well.

Heroman3003:
Issues resovled.

Leshana:
I think its a good addition.

SpadesNeil:
Conflicts on the map again.

Heroman3003:
Conflicts resolved. Man, mapping in text files is _fun_

Heroman3003:
Conflicts resolved. Man, mapping in text files is _fun_

PR #6573, opened 2020-02-14
Prevents resleeving sickness from affecting hardvore victims

### LINKED ISSUE #6569: HV Death, Resleeving Sickness Oversight
Seems that hard vore mechanics are not counted towards vore related deaths that lets you avoid resleeving sickness.

No resleeving sickness after an HV scene results in organs removed from a vital area.

#### What actually happened
Death, resleeving sickness on resleeve.

#### Steps to reproduce
- (Step 1) Target brain with HV
- (Step 2) Watch prey die
- (Step 3) Resleeve them.
- (Step 4) They get resleeving sickness.

#### Code Revision
- ff00f4280fbdbe5faf91b4328669dfdf8e539d11 
 
Heroman3003:
Title. Applies a very short buff when living creature is hardvored. If when buff expires its dead, its death is marked as vore death.

Resolves #6569

### DIFF
--- a/code/modules/resleeving/resleeving_sickness.dm
+++ b/code/modules/resleeving/resleeving_sickness.dm
@@ -22,3 +22,27 @@
 
 	on_created_text = "<span class='warning'>You feel slightly weak and unfocused.</span>"
 	on_expired_text = "<span class='notice'>You feel your strength and focus return to you.</span>"
+
+/datum/modifier/gory_devourment
+	name = "gory devourment"
+	desc = "You are being devoured! Dying right now would definitely be same as dying as food."
+	stacks = MODIFIER_STACK_EXTEND
+	hidden = TRUE
+	var/datum/mind/cached_mind = null
+
+/datum/modifier/gory_devourment/can_apply(var/mob/living/L)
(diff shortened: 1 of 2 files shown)


code/modules/resleeving/resleeving_sickness.dm
```diff
@@ -21,4 +21,28 @@
...
+
+/datum/modifier/gory_devourment/on_applied()
+	if(holder.ckey)
+		if(holder.mind)
+			cached_mind = holder.mind
+	return
+
+/datum/modifier/gory_devourment/on_expire()
+	if(holder.stat == DEAD)
+		cached_mind.vore_death = TRUE
```
Arokha: Shouldn't it be vore_death `FALSE` when it expires, if it expired before they died...?

Heroman3003: Its false by default, and that var is only meant to be set during or shortly after death anyway, so I do not think that is a problem.

Arokha: Oh nevermind, I see what you mean.

PR #6577, opened 2020-02-14
Character rename - Custom items update

Arokha:
Rename "Aronai Kadigan" -> "Aronai Sieyes"

Just changing character's last name.


Novacat:
We should just go ahead and delete config/custom_items.txt since it is no longer used.

PR #6580, opened 2020-02-15
Update <a> tag hrefs to have byond:// protocol

Arokha:
This is an optional thing to do in byond, apparently. If you don't do this, it just appends http://127.0.0.1/some/garbage/temp/path/here.

But, kinda need it to be this way for vchat. It seems like some other servers have done byond:// on just the things sent to chat, which I dunno why they'd bother being picky. It works the same either way. Just did a regex replace on it.


Arokha:
Can't go to Polaris until they merge the to_chat thing, since some lines are the same.

Arokha:
Leshana found a better way

PR #6582, opened 2020-02-16
Replaces Tramadol maint pills with Paracetamol

Novacat:
It always felt strange to me that a prescription narcotic was easier to get your hands on than a simple over the counter painkiller.


izac112:
Just a downstream so you don't have to bother answering but, Long story short Paracetamol is decently weaker then Tramadol irl and in-game atm tramadol is pretty strong seeing as you can even do sugery when someone is effected by it and they wont feel a thing any plans on changing how well it works ingame?

Not here to complain about the change just curious if this was just a name change or a  future balance change.

Novacat:
Balance change. The main reason I did this change is that Paracetamol is, at the moment, super rare and almost never used, because the only people with actual access to it also tend to not need it because they have ready access to tramadol/oxycodone.

This is an attempt to both make Paracetamol more widely used due to having tramadol's ease of access moved to it, and make Tramadol actually require some form of medical/chemistry access to get to as opposed to finding multiple pill bottles of the stuff in maintenance.

KasparoVy:
Yeah I agree lol. I've always felt wierd about having such ready access to super potent narcs

Ketrai:
Or alternatively could differentiate the medicine more giving each more nuanced pros and cons instead of this medicine better.

Novacat:
Not sure what that has anything to do with a prescription narcotic being more commonly accessible than an over the counter painkiller.

code/game/objects/structures/trash_pile.dm
```diff
@@ -219,7 +219,7 @@
 	return I
 
 /obj/structure/trash_pile/proc/produce_beta_item()
-	var/path = pick(prob(6);/obj/item/weapon/storage/pill_bottle/tramadol,
+	var/path = pick(prob(6);/obj/item/weapon/storage/pill_bottle/paracetamol,
```
SpadesNeil: `//VOREStation Edit`

Novacat: The entire file is a vorestation-specific file (which is why there are no //vorestation edits in it. However, I am going to append _vr to the file so that it is more obviously marked as a vorestation-specific file

PR #6586, opened 2020-02-17
Don't talk to strangers

Arokha:
I noticed that a few of these verbs in here send haxx0rs a message telling them no, which is pointless. But more importantly, several of them don't even bother returning out of the proc when they discover the user shouldn't be calling them. You should probably `return` after telling someone they can't continue, you know?


Novacat:
Mirrored to https://github.com/PolarisSS13/Polaris/pull/6703

PR #6592, opened 2020-02-18
VChat: Redone chat output in Vue.js 

Arokha:
VChat is an overhauled chat output box in HTML with JS frameworks, similar to 'goonchat'. It stands for vorechat but you can pretend it stands for Vue.js chat if vore offends your naive sensibilities.

It features things like:
---
### Living chat history
I can't really make a GIF of this one but basically if you lose your connection or disconnect, your chat history is returned to you when you reconnect, **including things you would have seen were you connected**. So you don't miss any posts.

---
### Tabbed chat
tabs
---
### Unread indicators
image
---
### Cute lil animations

---
### Dark mode
darkmode
---
### Pausing scrollback
pause
---
### Combining messages
combining
---
### Some sizes
sizes
---

Some things I'd like to add in the future:
* Saving the tabs between rounds
* Notification sounds, user settable?
* Ability to turn off notifications
* Other stuff???


izac112:
_I can just del this if it counts as a clutter comment_
But i have to say this pr makes me :eggplant: this looks awesome

KasparoVy:
That's goddamn awesome.

Arokha:
> I hope a dark theme is doable!

What'cha mean?

Heroman3003:
One thing I am worried about is ability to disable stuff that would otherwise require actual IC action to disable or that is not disablable at all with this system, like radio or PDA, needing to interact with appropriate devices, or warning messages and local chat, not disablable at all.

KasparoVy:
> > I hope a dark theme is doable!
> 
> What'cha mean?

OH FUCK YES IT'S ALREADY THERE

My bad!
Woooo my eyes, thank you kindly. I saw this in mobile and dropped my thoughts without looking too closely at the gif or the diff

Vite in umbra

Wickedtemp:
Actually looks really neat. Do notifications always show up, or can they be muted, or is that one of the future plans?

And I'm fairly sure you wouldn't be able to filter out comms and such using the OOC chat interface, especially when if you wanted to do that, you can just turn off the pda messenger and headset.

Butyeh. The scrollback pause is especially lovely. I can't even count the number of times I've needed to scroll up to check a thing, but due to comms chatter, it was much more annoying to do. 

Dark mode, amazing. Logs being saved? Amazing. If you can add some sort of Autosave toggle, where logs can be automatically saved upon exiting - like a "save logs before closing?" popup that can be toggled on or off, that'd be the last QOL I can think of.

Arokha:
In re: Heroman, you can't disable any categories on the Main tab at all. So it's not possible to like... disable yourself from ever getting radio messages because at the very least, they will be on the Main tab. If you're worried about "powergamers" disabling attack messages (I think it was you who mentioned that in the Discord?) I'd say that's silly, because it's very unlikely they care about attack messages.

Arokha:
> like a "save logs before closing?" popup that can be toggled on or off, that'd be the last QOL I can think of.

That'd be neat, but unfortunately not possible. The byond client doesn't have any kind of "do this before you close" thing, and for the Internet Explorer part, well. 

There's an event that's fired when webpages are 'closing' or 'leaving' the page, `beforeUnload`, but for obvious reasons browsers heavily limit what you're allowed to do during that event, including any sort of popups or prompts. Otherwise webpages would just be able to capture you forever by having the `beforeUnload` thing just send you back to the webpage or never let you leave by popping up constant prompts or whatever. (You may remember this happening in The Before Times of the internet, and still occasionally when pages ask if you want to navigate away, but it's pretty rare).

I'm pretty sure just closing the byond window pretty instantly terminates any sort of processes going on though in your client, so I don't think there's a "time" it could ask you.

But, you know, if you forget you can just reconnect and save them since you'll just get your logs back.

Arokha:
I sent them the parts that have 'large conflict potential' already, like the to_chat refactor. They merged that. The bicon() thing I can't really send them, but that has a much smaller surface area.

I don't think it's appropriate to send it there first, because I don't have the potential to quickly fix bugs that arise with it there. VChat (and goonchat for that matter) provide live copies of the js and html files from the server whenever clients connect, so there's no PR/recompile needed to fix bugs discovered during the first round of use, just tweak the files and clients can use the 'Reload VChat' verb, or reconnect. I expect there to be some bugs despite the testing we've done, since it's such a large project/change.

It just makes more sense to do it here first, because I can fix the bugs discovered in the very first round without even restarting the round, then send it to them with any 'Day-0' patches applied.

SpadesNeil:
I do agree it would be best if we deployed it here first. As long as Arokha and Leshana keep on top of it, and I am certain they will, this system could be ported to Polaris in a much more workable state. Any bugs we experience would also be resolved faster. This is, after all, a very major change. Sending it to Polaris will cause needless delays in the testing process.

Arokha:
https://github.com/PolarisSS13/Polaris/pull/6761

code/game/machinery/cryopod.dm
```diff
@@ -97,7 +97,7 @@
 		//dat += "<a href='?src=\ref[src];item=1'>Recover object</a>.<br>" //VOREStation Removal - Just log them.
 		//dat += "<a href='?src=\ref[src];allitems=1'>Recover all objects</a>.<br>" //VOREStation Removal
 
-	to_chat(user, browse(dat, "window=cryopod_console"))
+	user << browse(dat, "window=cryopod_console")
```
Novacat: ...Is this supposed to be going back to user <<?

Heroman3003: Well, this isn't a chat situation, so it makes sense NOT to use to_chat there.

PR #6603, opened 2020-02-19
Fluff item

Meghan-Rossi:
*Adds custom latex clothing for Excess - Approved in https://forum.vore-station.net/viewtopic.php?f=27&t=1593


Heroman3003:
I think this would be much better done via just making a new icon... This overlay system is bound to get messy.

Meghan-Rossi:
Ok, done.  Hadn't done any sprites before but I managed to figure it out.

PR #6606, opened 2020-02-19
Adds biochemical dispenser to medical's chemistry

Heroman3003:
Adds a biochemical dispenser. Dispenses Nutriment, Milk and Animal Protein. Available in amount of one in both medical chemistries. Also includes a separate sprite for xenoflora chem dispenser.

Post-staff discussion tag addition:
Main reason for this is so that medical doesn't have to empty their NutriMax just for those three reagents. But separate dispenser so that only one is accessible per chemistry lab, not making those three too abundantly available like that.


Cameron653:
Seems like a good idea, I don't see why not. 
Those three chemicals are common enough and already readily accessible, so there's not much reason to exclude them from chemistry.

Mind you, from as far as I can see, protein really isn't required in any reactions, Milk is used to make Adranol (a sedative), and Nutriment is used in a slime food reaction. 

While it isn't good for reactions alone, the three could possibly be good to mix with other things or possible future chemical reactions if any more are made.

Heroman3003:
Animal Protein (also known as protein in code) is used in biomass reaction

PR #6619, opened 2020-02-20
Fix Sundry Runtime Errors

Leshana:
* **Fix Runtime in seed.dm,102: Cannot read null.len**
    * Gotta check if the list is non-null before checking length! Tested by spawning plants.
* **Fix Runtime in filter.dm,91: Cannot execute null.set machine()**
	* Use `user` not `usr`.  Tested by spawning the machine, opening nanoui and walking away.
* **Fix Runtime in wires.dm,120: Cannot execute null.is wirecutter().**
	* You can click "Cut" with an empty hand!  Tested by doing that.
* **Fix Runtime in loadout.dm,277: bad index**
	* Runtime happens when someone has a saved preferences loadout containing an item that currently is expected to have gear_tweaks, but at the time that the savefile was last saved, did *not* have gear tweaks.  Thus the saved prefs have null metadata.
	* Also it makes no sense to check the length of a list *inside* a for loop iterating over that list..
	* Tested by manually setting up the situation with VV and then loading/saving/spawning.
* **Fix bad index and bad arg name runtimes in find_spawning.dm**
	* Code was assuming that item.origin_tech would be an initialized list. To be safe we use LAZYSET so it will work properly even if its null.
	* Other code was calling pickweight and passing it the options as named arguments.  They are supposed to be passed as a list.
	* Tested by spawning the /obj/item/weapon/archaeological_find and fixing things until runtimes stopped.
* **Fix Runtime in observer_vr.dm,65: Cannot read null.name** (VOREStation Only)
  * If an observer/ghost was never in the game this round, their mind var will be null.  No sense in letting them ping the transcore.

### DIFF
--- a/code/ATMOSPHERICS/components/omni_devices/filter.dm
+++ b/code/ATMOSPHERICS/components/omni_devices/filter.dm
@@ -88,7 +88,7 @@
 	return 1
 
 /obj/machinery/atmospherics/omni/atmos_filter/ui_interact(mob/user, ui_key = "main", var/datum/nanoui/ui = null, var/force_open = 1)
-	usr.set_machine(src)
+	user.set_machine(src)
 
 	var/list/data = new()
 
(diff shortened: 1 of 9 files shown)


KasparoVy:
Yesss `?.`

Novacat:
Linked to https://github.com/PolarisSS13/Polaris/pull/6738

PR #6622, opened 2020-02-20
Holowarrants can now only be authorized by an ID with HoS access

Shadow-Quill:
Denied up at Polaris, so it's brought down here.

As it stands, the 'authorize' button on holowarrants can be pressed by anyone with access to the Warrant Assistant program. This changes it to be in line with SOP, and only have HoS access. See [SOP 5.1 - Green Alert](https://wiki.vore-station.net/Standard_Operating_Procedure#Code_Green_-_All_Clear). Also maps in a box of spare holowarrants in equipment storage.


Wickedtemp:
Makes sense. Its only denied on Polaris due to them being dead-low pop

PR #6626, opened 2020-02-21
New sprites for discs with unique colors

Arokha:
Now every disc has a unique color rather than most of them being the same and all of them being in cloning.dmi.

image

Sprites from CEV-Eris.


Rykka-Stormheart:
I APPROVE OF THIS TOO. <3

code/modules/research/research.dm
```diff
@@ -236,8 +236,8 @@ research holder datum.
 /obj/item/weapon/disk/design_disk
 	name = "component design disk"
 	desc = "A disk for storing device design data for construction in lathes."
-	icon = 'icons/obj/cloning.dmi'
-	icon_state = "datadisk2"
+	icon = 'icons/obj/cloning.dmi' //VOREStation Edit
```
Novacat: ...This should probably be icons/obj/discs.dmi?

PR #6630, opened 2020-02-21
New terminal sprites

Arokha:
dreamseeker_2020-02-20_23-25-26

dreamseeker_2020-02-20_23-26-31

From CEV-Eris


KasparoVy:
Great designs.

Heroman3003:
New request console looks way too generic. 

New ATM looks a lot less like actual ATM too to me.

Also do they have proper frame and construction step sprites that actually match the final product?

Novacat:
I never was a fan of the old request console and newscaster because they blended in way too much with the walls. The new ones actually stick out and thus might actually get people to interact with them more.

ATM Change I am neutral on, which I guess is fine.

PR #6633, opened 2020-02-21
Replaces emag interaction with pizza voucher

Heroman3003:
Title. Droppod is ridiculous, yes, but also distructive in levels it really shouldn't be, and there is absolutely 0 indication to how dangerous it actually is until after you used it too.

Now with special delivery it just sends a pizza similarly to normal delivery, but that pizza will actually crush whoever it lands on. No damage to surroundings/lost equipment.

### DIFF
--- a/code/game/objects/items/pizza_voucher_vr.dm
+++ b/code/game/objects/items/pizza_voucher_vr.dm
@@ -4,7 +4,6 @@
 	icon = 'icons/obj/items.dmi'
 	icon_state = "pizza_voucher"
 	var/spent = FALSE
-	var/special_delivery = FALSE
 	w_class = ITEMSIZE_SMALL
 
 /obj/item/pizzavoucher/New()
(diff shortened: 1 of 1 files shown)


Very-Soft:
I'd prefer that we make it give you an indication of how dangerous it is, with a confirmation dialogue rather than just removing the interaction.

Novacat:
I am going to agree with Soft, I would rather have a warning dialogue instead of outright removing the feature.

Heroman3003:
Replaced the interaction to not be that destructive to surroundings (that is, not damage anything except the mobs on tile it lands on), or the equipment/body of person (gibs instead of deleting)

Very-Soft:
that sounds like a much better compromise

SpadesNeil:
I kinda liked the drop pod but meh I guess there's a valid point.

SpadesNeil:
Wait, can you even get pizza vouchers outside of admin spawning bullshit? I think you can get them from trash loot.

Heroman3003:
You can order them in cargo, get them from trash piles and occasionally roundstart randomized trash can be a voucher too.

PR #6637, opened 2020-02-21
[MIRROR] Fix to_chat typo

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/6712
--------------------
This file is unchecked so it's not being used, but I'd still rather not leave this since someone will blame me eventually.


Arokha:
Already implemented here

PR #6639, opened 2020-02-21
New rollerbed sprites

Arokha:
Less eye-melting blue, now with More Clickability

image


KasparoVy:
Looks awesome.

Heroman3003:
Clickability is good, but I don't like the lack of white line on new sprite. Kinda made it actually look like a bed.

Novacat:
Hospital beds actually do look like that. They tend to be very minimalistic and are not very comfortable.

Heroman3003:
Eh, its less about it being comfortable and more about just having different color in 'where head goes' spot. Doesn't have to look like pillow like old one.

PR #6642, opened 2020-02-21
A few new vending sprites

Arokha:
Most unused, only using nutri and seed

You can see the unused ones in icondiff


Ketrai:
Whilst the sprite quality is better on these, I don't really like the loss of character of these sprites. The original vendor sprites have this seed like shape, making it all vibrant and cute.  I'd rather see the design overhauled tyan replaced.

Heroman3003:
If they have to be 3/4th, I'd honestly like 3/4th version of oldschool Nutrimax and seed vendor. Ones that were fully bright green. I hate current ones myself, since they don't look like a workplace vendor at all, but these new ones suffer from opposite end of the issue, being excruciatingly bland.

PR #6645, opened 2020-02-21
New library book/machine sprites

Arokha:
I'm fairly confident that the spellbook one isn't real. It's not in the old file either...

image


Ketrai:
Not entirely a fan of the top down perspective on the fax machine. This is very conaistent for eris, but it makes sprites look a bit flat at times. The other sprites look nice to me, albeit I remeber some books, like the virgo-erginome guide having some distinct character in terms of design and shape. But it seems those were still kept. Also if you don't mind, could it be made so the books are rotated based on where the reader places it on the table? Would add some visual depth to the game if people aren't reading them sideways.

Arokha:
> Also if you don't mind, could it be made so the books are rotated based on where the reader places it on the table? Would add some visual depth to the game if people aren't reading them sideways.

Might be possible if someone were to code that. This is just purely a sprite PR though.

Heroman3003:
All those look like a downgrade unless its been done just to implement directional sprites. But it doesn't seem like directional sprites are part of it, so... Mreh. I think they are just worse versions of existing sprites, except fax machine whose suggested version just looks terrible, with white color not present on almost any other machinery, and roundish shape lost to make yet-another-square.

PR #6648, opened 2020-02-21
Replace IV drips with medical stands

Arokha:
They can also hold a gas tank and have a built in mask for gassing people if that's your cuppa.

The sprite is also less prone to being *invisible* on our floors.

dreamseeker_2020-02-21_18-35-12

dreamseeker_2020-02-21_18-36-23

(Yes I know you have wall mask holder tank things. I'm the one who put them there. But no reason to not make IV drips have a little more functionality, or if you're *anywhere else* since only surgery has those.)


KasparoVy:
Multipurpose ftw

TheFurryFeline:
`$ python tools/TagMatcher/tag-matcher.py ../..
../../VOREStation/VOREStation/code/game/objects/structures/medical_stand.dm
	<span> - Excess of 3 opening tag(s)
		Line 189
		Line 195
		Line 328
Total number of mismatches: 3
The command "python tools/TagMatcher/tag-matcher.py ../.." exited with 1.
`

Wickedtemp:
How does the mask function? IV's currently are just a click-drag onto the patient to hook them up, is that remaining the same, with the mask having a different control? 

This is really neato and, yeah, clicking on these sprites has always been an annoying challenge. If there could be a sprite indicator to actively display if the IV is taking blood or injecting from the bag, that'd also be great, but it'd kinda just be an extra QOL that isn't nearly as mlemportant.

Arokha:
> How does the mask function? IV's currently are just a click-drag onto the patient to hook them up, is that remaining the same, with the mask having a different control?

Also drag. If there's multiple things you could be doing, it asks what you want. If there's no tank for example, drag just attaches IV. If there's no reagent container and only a tank, drag puts mask. If there's both, it just asks.

Direction indicator would be good. Wonder how to indicate that? The only ways I can think of are lights but you'd have to memorize that... 'purple means in and orange means out!' or whatever.

TheFurryFeline:
Over a hundred errors.....

Wickedtemp:
Indicator lights would be handy. It's one more thing to remlember and mlemorize, but its better than the current system of "Wait why is the bloodbag not decreasing in volume. FUCK WAS IT SET TO 'TAKE'? OH G-"
*Patients body seizes up, their eyes dead and lifeless*

Arokha:
> FUCK WAS IT SET TO 'TAKE'? OH G-"

Both the original IV and this are intended to spam messages to chat when the patient's blood level is approaching dangerous volumes. Just add a sound effect too!

Arokha:
Don't merge this yet for a sec. It has a bug I want to fix.

PR #6658, opened 2020-02-23
Fixes cloaks being broken on taurs.

Heroman3003:
As well as any other wearable non-suit item in suit slot.


KasparoVy:
Good catch, cheers
Oversight on my part and an elegant fix

PR #6662, opened 2020-02-24
Attempts to fix Panslavic, again.

SpadesNeil:
I noticed panslavic wasn't formatted like other languages and tried to fix it. I have no fucking idea if this works in the slightest as I cannot test it.

### DIFF
--- a/code/modules/mob/language/station_vr.dm
+++ b/code/modules/mob/language/station_vr.dm
@@ -88,23 +88,26 @@
 	desc = "Shadekin seem to always know what the others are thinking. This is probably why."
 	speech_verb = "mars"
 	ask_verb = "mars"
-	exclaim_verb = "mars"
+	exclaim_verb = "MARS"
 	colour = "changeling"
 	key = "m"
 	machine_understands = FALSE
 	flags = WHITELISTED | HIVEMIND
 
-/datum/language/human/slavic
+/datum/language/slavic
 	name = LANGUAGE_SLAVIC
 	desc = "The official language of the Independent Colonial Confederation of Gilgamesh, originally established in 2122 by the short-lived United Slavic Confederation on Earth."
-	colour = "solcom"
+	speech_verb = "govorit" // All Russian for "says" "asks" and "shouts". Mostly here as a joke.
+	ask_verb = "voprosy"
+	exclaim_verb = "krichit"
+	whisper_verb = "shepchet"
+	colour = "attack" // Red for slav
 	key = "r"
-
 	syllables = list(
-		"rus", "zem", "ave", "groz", "ski", "ska", "ven", "konst", "pol", "lin", "svy",
-		"danya", "da", "mied", "zan", "das", "krem", "myka", "to", "st", "no", "na", "ni",
-		"ko", "ne", "en", "po", "ra", "li", "on", "byl", "cto", "eni", "ost", "ol", "ego",
-		"ver", "stv", "pro"
+		"rus", "zem", "ave", "blyat", "cyk", "groz", "ski", "ska", "ven", "konst", "pol", "lin", "svy",
+		"danya", "da", "mied", "kuz,", "zan", "das", "krem", "myka", "to", "st", "no", "na", "ni",
+		"ko", "ne", "en", "po", "tro", "ra", "li", "on", "byl", "cto", "eni", "ost", "ol", "ego",
+		"ver", "stv", "pro", "ski"
 	)
 
 /datum/language/unathi
@@ -112,8 +115,6 @@
 /datum/language/tajaran
 	flags = 0
 /datum/language/skrell
-	flags = 0
-/datum/language/human
 	flags = 0
 /datum/language/seromi
 	flags = 0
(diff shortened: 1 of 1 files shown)


code/modules/mob/language/station_vr.dm
```diff
@@ -88,23 +88,26 @@
...
+/datum/language/slavic
 	name = LANGUAGE_SLAVIC
 	desc = "The official language of the Independent Colonial Confederation of Gilgamesh, originally established in 2122 by the short-lived United Slavic Confederation on Earth."
-	colour = "solcom"
+	speech_verb = "govorit" // All Russian for "says" "asks" and "shouts". Mostly here as a joke.
+	ask_verb = "voprosy"
+	exclaim_verb = "krichit"
+	whisper_verb = "shepchet"
+	colour = "attack" // Red for slav
 	key = "r"
```
SpadesNeil: So this is the problem?

Novacat: Yes, another language has the same key, when two languages have the same key, the code that is responsible for Gibberish gets confused and converts all attempts to use that language into gibberish.

Shadow-Quill:
@Novacat I present to thee:
[Saycode!](https://github.com/VOREStation/VOREStation/blob/master/code/modules/mob/living/say.dm) (With complimentary list of radio keys :3)

SpadesNeil:
image
What is this even for?!

Heroman3003:
This is the code to force anyone who fails to speak a language they don't know to speak "Gibberish". A special technical language that nobody can understand. Its purpose is primarily to prevent stuff like typo in a language key making message be said in common, pretty much ruining the light secrecy that comes with using a language. Gibberish is automatically spoken any time someone tries to speak a language they don't actually know.

SpadesNeil:
No I understand what gibberish is for. Why am I able to speak it on purpose?

code/modules/mob/language/station_vr.dm
SpadesNeil: Then what other language uses the same key? Furthermore, I _thought_ keys could be selected in preference setup.

Shadow-Quill: You can set the key markers (like using &8) using setup iirc

Heroman3003:
Mostly because every language must have a language key and gibberish code in general relies on it having its own language key and everyone 'knowing' it. But, tbh, you would be able to speak it 'on purpose' even without it having its own key.

SpadesNeil:
That's fucking stupid.

code/modules/mob/language/station_vr.dm
```diff
@@ -88,23 +88,26 @@
...
 
-/datum/language/human/slavic
+/datum/language/slavic
 	name = LANGUAGE_SLAVIC
 	desc = "The official language of the Independent Colonial Confederation of Gilgamesh, originally established in 2122 by the short-lived United Slavic Confederation on Earth."
-	colour = "solcom"
-	key = "r"
-
+	speech_verb = "govorit" // All Russian for "says" "asks" and "shouts". Mostly here as a joke.
+	ask_verb = "voprosy"
```
Heroman3003: wouldn't "sprashivaet" be more fitting here (russian for 'asks')? "voprosy" means "questions". But not as in the verb, as in the noun's multiple form. So not as in "he questions" but as in "he has many questions".  For direct translation of questions as a verb, you'd use "voproshaet" but that's pretty... oldschool stuff. Something you'd see in 18th century prose, not modern day language.

SpadesNeil: Do you speak Russian? I don't. I had to use Google and try to pick what seemed like the best word.

The only purpose of doing this is so I can even see if I am using the correct language at all.

Heroman3003: I do speak it as my first language. And not trying to scrutinize too much, just noticed that other three verbs were correct, while this was one that wasn't.

SpadesNeil: Cool, so I can take your advice and feel better about it than Google translate and my own shitty understanding of Russian.

SpadesNeil:
I'm a chad who doesn't test his code. ~~Meaning I can't currently test the code as Github is fucked on my PC.~~

TheFurryFeline:
> Probably would be a good idea to make sure this works, but otherwise seems good?



> I'm a chad who doesn't test his code. ~Meaning I can't currently test the code as Github is fucked on my PC.~
> 
> Even if it doesn't work, it's not a downgrade from it already not working.

Currently testing it now. Compiling atm.

code/modules/mob/language/station_vr.dm
```diff
@@ -88,23 +88,26 @@
...
 	ask_verb = "mars"
-	exclaim_verb = "mars"
+	exclaim_verb = "MARS"
 	colour = "changeling"
 	key = "m"
 	machine_understands = FALSE
 	flags = WHITELISTED | HIVEMIND
 
-/datum/language/human/slavic
+/datum/language/slavic
```
TheFurryFeline: This seems to be listed right after Shadekin Empathy language on a character's Player Panel. 
image Any way to have it be at a more appropriate place, like, where the languages aren't event char or robotic. Just a purrsonal preference regarding where it's placed.

SpadesNeil: Won't fix.

TheFurryFeline: Alrighty, then.

Heroman3003:
The 's' key will still conflict with another commonly used language: Sign Language.

KasparoVy:
Blyat..

PR #6683, opened 2020-02-25
Species consolidation

Novacat:
- Consolidates the two zorren species into one zorren species.
- Merges Xenohybrid into Custom species, Xenohybrids no longer whitelisted.
- Tweaks Custom species bodytype selection code to allow selection of restricted species bodytype
- Adds Flatland Zorren bodytype as 'Fennec' and Xenohybrid body type as 'Xenohybrid' for custom species
- Refactors Spacesuit code a little bit so that Fennec bodytypes can wear Zorren-fitted spacesuits, and Xenohybrid body types can wear Unathi-fitted spacesuits.
- Removes Xenochimera body type voidsuit fitting, they can now use and wear human voidsuits


Screemonster:
One thing people might need to be aware of: if they already have a zorren character, what will happen if they try to join as them after this goes live?

Heroman3003:
I noted this in the #development, but will also say here. This will NOT merge xenohybrids into custom species, it will make them impossible to create entirely, due to their bodytype being made inaccessible. For bodytype to be accessible, it needs to be present on a non-whitelisted species, in the current system. Same applies to flatlander zorren bodytypes (which I believe is slightly different from highlander).

Ketrai:
It's nice to be able to reduce clutter a bit. But I'd rather see the zorren species left to be diversifidd by someone, unless you can merge the species without causing issues for people playing zorren.

Novacat:
I was able to add the Flatland zorren and Xenohybrid bodytypes for custom species, however, due to the nature of the changes people with Flatland Zorren/Xenohybrid characters will have their species reset, but they should be able to recreate their character as a custom species.

PR #6684, opened 2020-02-26
Fixes free clothing dispenser only handing shoes for money

Heroman3003:
The Basics sold shoes when eveything in it was supposed to be free. This is why inheiritance is an unstable road.


KasparoVy:
Sweet kicks.

PR #6703, opened 2020-02-29
Removes teleporter room from secondary command

Heroman3003:
More detailed list of changes:

- Removes teleporter room from secondary command entirely: Primarily because there really shouldn't be more than one around on station itself. Its supposed to be something unique and critical, so having a second copy, including a second copy of an item that's also meant to be unique and critical in it, makes no sense. Plus two teleporters make hand tele much stronger than it'd normally be.

- Removes little hallway in secondary command: with only two rooms left, there's no reason to have it anymore.

- Removes secondary command substation, moves to civilian grid: similar reasoning. With the teleporter gone, most of potential power draw is too, so the secondary command having a dedicated substation stops being sensible. In its original inception it also didn't have a substation either, so not much changed there.

- Moves EVA to exploration subgrid: primarily because the weird 'vertical movement from middle of hallway to maint' thing is now gone too. Maint there is no entirely normal and EVA is now part of exploration grid.

- Remapped maintenance around the secondary command: reshaped to fit new form, given slightly more mainty stuff and a tiny card playing room. Nothing too important there.

- Added privacy shutters between command office and meeting room

- Added blast doors to meeting room windows.

- Added a newscaster and an air alarm to meeting room: just minor missing thing to be done along the way.

### DIFF
--- a/maps/tether/tether-05-station1.dmm
+++ b/maps/tether/tether-05-station1.dmm
@@ -2819,21 +2819,19 @@
 /turf/simulated/floor/tiled/dark,
 /area/gateway/prep_room)
 "afc" = (
-/obj/machinery/door/airlock/command{
-	name = "Secondary Command Office"
+/obj/machinery/atmospherics/unary/vent_scrubber/on,
+/obj/structure/cable/green{
+	d1 = 1;
+	d2 = 8;
+	icon_state = "1-8"
 	},
-/obj/machinery/door/firedoor/glass,
-/obj/structure/disposalpipe/segment,
-/obj/machinery/door/blast/regular{
-	density = 0;
-	dir = 8;
-	icon_state = "pdoor0";
-	id = "secondary_bridge_blast";
-	name = "Secondary Command Office Blast Doors";
-	opacity = 0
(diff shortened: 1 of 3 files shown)


Screemonster:
Late to the party, but

"Primarily because there really shouldn't be more than one around on station itself. Its supposed to be something unique and critical, so having a second copy, including a second copy of an item that's also meant to be unique and critical in it, makes no sense."

This is opinion. Not fact. Why the hell was this merged without staff discussion?

Very-Soft:
To put my two cents in here, I put a second teleporter in so that command/antag types could have another reliable way between the surface and space. We use teleporters pretty regularly for things, such as allowing people to enter and leave the station, and go to event locations, so I don't think it's out of the realm of possibility for there to be two of such things at either end of the elevator. 

People's number one complaint about the tether is that there aren't enough alternatives to the elevator for getting around. The extra teleporter was meant to try to help that.

I think if we were on a map with a more unified, interconnected layout, then having multiple teleporters would make less sense, and I could see the argument for avoiding having more than one. But the Tether's design benefits from having one at either end.

PR #6711, opened 2020-03-01
Fixes cloaks on taurs again AGAIN

Heroman3003:
Every time it's fixed in one place, it breaks in another. I am lowkey growing to hate clip mask feature.

### DIFF
--- a/code/modules/mob/living/carbon/human/update_icons.dm
+++ b/code/modules/mob/living/carbon/human/update_icons.dm
@@ -781,7 +781,7 @@ var/global/list/damage_icon_parts = list() //see UpdateDamageIcon()
 	var/tail_is_rendered = (overlays_standing[TAIL_LAYER] || overlays_standing[TAIL_LAYER_ALT])
 	var/valid_clip_mask = (tail_style && tail_style.clip_mask_icon && tail_style.clip_mask_state)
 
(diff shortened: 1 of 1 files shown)


KasparoVy:
What exactly is the bug? Is stuff not rendering, or being clipped when it shouldn't, etc?

code/modules/mob/living/carbon/human/update_icons.dm
```diff
@@ -781,7 +781,7 @@ var/global/list/damage_icon_parts = list() //see UpdateDamageIcon()
 	var/tail_is_rendered = (overlays_standing[TAIL_LAYER] || overlays_standing[TAIL_LAYER_ALT])
 	var/valid_clip_mask = (tail_style && tail_style.clip_mask_icon && tail_style.clip_mask_state)
 
-	if(!(istype(suit)) || (tail_is_rendered && valid_clip_mask && !(suit && suit.taurized))) //Clip the lower half of the suit off using the tail's clip mask for taurs since taur bodies aren't hidden.
+	if(tail_is_rendered && valid_clip_mask && !(suit && istype(suit) && suit.taurized)) //Clip the lower half of the suit off using the tail's clip mask for taurs since taur bodies aren't hidden.
```
KasparoVy: Makes more sense having it this way (putting it with the check to see if there's a suit at all), good thinking

Heroman3003:
Bug changed. First it runtimed and didn't render on taurs. Then it just didn't render on taurs. Then it rendered on taurs but runtimed and didn't render if person lacked a tail. Hopefully now it renders always.

KasparoVy:
Thanks much for the explanation, I appreciate it

PR #6719, opened 2020-03-01
Allow narrate/globalnarrate shenanigans

Arokha:
Allows HTML if your entire thing is HTML


Arokha:
Also added this to the POL vchat pr

PR #6727, opened 2020-03-01
Ore Scanner is written incorrectly

Arokha:
Only revealed by vchat


Heroman3003:
This is also a Polaris bug

Arokha:
> This is also a Polaris bug

I added it to the vchat PR on Polaris because it can't go in before that because of bicon()

PR #6740, opened 2020-03-02
Extends Tempest's hypo and glasses whitelists to Off-duty Medic

Wickedtemp:
As long as I typed it correctly, I just copied how its spelled from the manifest. I couldn't really test this on a private server because of a lack of PTO I think, because even as a Host ranking in the config, while I could join as CC Officer and Mimes, I couldn't join as anything off-duty, so... If it works, it works, and if it doesn't, I'll just try another spelling or something.


Heroman3003:
About inability to join off-duty locally: you need to enable it in config first AND somehow code in some PTO for yourself.

Wickedtemp:
Neato.

PR #6746, opened 2020-03-03
Buffs bioadaptive NIF to 75 durability

Sebbe9123:
So, normal nif for every other species = 100
NIF for weird wacky species like prommies = 25
Fun = none
This is now changed, dunno why this was this way for so long, but fun is now allowed for all weird and wacky species.


code/modules/nifsoft/nif.dm
```diff
@@ -570,7 +570,7 @@ You can also set the stat of a NIF to NIF_TEMPFAIL without any issues to disable
 	name = "bioadaptive NIF"
 	desc = "A NIF that goes out of it's way to accomidate strange body types. \
 	Will function in species where it normally wouldn't."
-	durability = 25
+	durability = 100 //Vorestation edit, why are we nerfing non-normal organics. fun should be allowed. 25 changed to 100
```
TheFurryFeline: Huh, I thought this was a Vorestation only item. The whole Nifsoft module is a Virgo creation, innit? Won't really need the comment in that case.

Novacat: That is correct.

TheFurryFeline: Also, I just noticed the desc has a spelling error: `it's way` rather than `its way`

Screemonster:
What species besides prommies even use these

Xenochimera are specifically supposed to reject implants of any kind because of how they are

code/modules/nifsoft/nif.dm
Screemonster: Can we NOT have passive-aggressive bullshit left in comments?

TheFurryFeline:
I think Alraunes and Diona would also use these

Screemonster:
Diona get more than enough benefits for their species without also giving them crazy augments. Alraunes should probably be able to use normal ones if they can't already.

Setting the durability cap per-species would be pretty neat.

Arokha:
I think I'll just rewrite NIFs later today so maybe don't merge this yet.

code/modules/nifsoft/nif.dm
Sebbe9123: Fixed

Wickedtemp:
Yeah, I THINK it mightve been like this because the implanting process is so much more simple than standard NIF's, but like... that doesn't really matter?

Especially when you have to repair it every time you use any nifsoft stuff because some of them will take you all the way to 0 in one use.

Arokha:
> Especially when you have to repair it every time you use any nifsoft stuff because some of them will take you all the way to 0 in one use.

Which ones do THAT? That wasn't intentional when I wrote them. Maybe something changed since then, but let me know if you find NIFSoft that you can show takes you down like 20 points just from one use.

Heroman3003:
Yeah, I believe most expensive NIFsoft there is is Size Alteration and even then it only sosts like 7 durability.Pretty much every other costs 1 or 2 durability.

Novacat:
...Doing the math, 7/25 is .28 which means that Size Alteration would eat 28% of a bioadaptive NIF. That is probably where the 20 figure came from.

Arokha:
> some of them will take you all the way to 0 in one use

Is what I was concerned about. No single NIFsoft would do that so maybe that was just hyperbole.

Sebbe9123:
The bottom line is that we are punishing players for playing species they enjoy, no matter how "easy" it is to fix, it's not necessary to do every 2-3 rounds when everyone else only need to do it every 15-20 rounds.
Fun > "just fix it lmao"

TheFurryFeline:
On that point, I agree wholeheartedly.

Wickedtemp:
The NIF surgery and waiting period isn't enough of an annoyance to balance out the difference. Waiting 30 minutes a single time, or deal with the constant "I need to get it repaired again" every other shift. One is so much easier than the other.

Ketrai:
Can we just up the durability, but make it more costly to use some of the mechanically special  nifs? I'd rather not make it needlessly difficult for prommies, diona, alraune etc to use soulcatchers and size changers.

Wickedtemp:
If you don't want to add "needless difficulty", why increase the cost of anything? The NIF should just be buffed. No cost increase.

Arokha:
Too lazy to do NIF refactor right now. Note that removal of DNM label does not imply that it should be merged.

Novacat:
I would be fine with this, especially since bioadaptive is not really special except it is usable on NOSCAN species. Or if it has to have a different durability, something like 75 which is not going to be as much of a problem as 25 dura.

PR #6747, opened 2020-03-03
[Map] Remove Docking Controller

TheFurryFeline:
Title. It's in a vacuum, anyway. And will always be in a vacuum so the lack of sensors mapped in made it display an errorneous air reading.

Changelog Notes:

- Removes the docking controller for the belt miner shuttle.

### DIFF
--- a/maps/tether/tether-07-station3.dmm
+++ b/maps/tether/tether-07-station3.dmm
@@ -29445,11 +29445,12 @@
 /turf/simulated/shuttle/floor/yellow/airless,
 /area/shuttle/belter/station)
 "Vb" = (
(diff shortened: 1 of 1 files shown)


Heroman3003:
The shuttle will throw an error every time shuttle system initializes if it does not have a docking controller. Moreover, I believe it won't function without it at all. It must have a controller regardless of whether it actually does any docking or not.

TheFurryFeline:
Okayyyyyyy, something's really fucky with this branch. Testing environment doesn't even let me get roundstart to work and it freezes at that time permanently. I'm gonna leave it for Hero to deal with regarding the mentioned issue of the controller displaying a stable atmos when you're actually in a vacuum.

PR #6751, opened 2020-03-03
[MIRROR] Half-Refactors Jobs

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/6762
--------------------
This is mostly concerned with ripping out the hardcoded job lists like `command_positions`, in favor of lists being dynamically generated from job datums, and new department datums.

Things this PR does:
- Renames `ROLE_` defines to `DEPARTMENT_`, to be more explicit.
- Adds a Job Subsystem, which for now just holds onto lists of job and department datums. Someday it might be good to port the job controller code to it and clean it up.
- Replaces the various non-GLOB global lists of job titles in department-specific lists with a list that is automatically generated from department and job datums by `SSjob`. Replaced code that depended on those non-GLOB list with code that talks to `SSjob`.
- Makes it so jobs can be defined as being inside multiple departments. This is typical of most command roles.
- Modifies job sorting code to take into account a new variable in the job datum called `sorting_order`. Higher numbers make the job get sorted higher. Department datums also have that variable and it works the same way, ensuring Command will always be first.
- Adds a department datum sorting proc.
- Modifies both the old ID computer and new modular computer ID computer programs to use the dynamically generated list from `SSjob`. This makes it so if a new job is added, it is automatically included in the ID computer, and is a step towards map-agnostic code. Manifest code remains awful, sadly.
- Backports some of the nanoUI from the newly fixed ID computer into the old one, to fix a bug that affects people using :penguin: OSes. This whole thing was actually to fix that bug.
- Tries to fix the indentation of the new nanoUI ID computer to be less heretical.

Fixes #6732


Arokha:
I'll look into fixing the conflict(s)

Novacat:
Taking this over from Aronai.

PR #6756, opened 2020-03-04
Landmark Shuttles

Leshana:
### Replaced "area" shuttles with "landmark" shuttles.

Largely ported from the work done at Baystation in https://github.com/Baystation12/Baystation12/pull/17460 and later commits.

#### Main Changes
 - Shuttles no longer require a separate area for each location they jump to.
   Instead destinations are indicated by landmark objects, which are not necessarily exclusive to that shuttle.
 - Enhanced shuttle control computers to use nanoui if they didn't.
 - Organizes shuttle datum code a bit better so there is less re-inventing the wheel in subtypes.
 - Allows the possibility of shuttles (or destinations) that start on late-loaded maps.
 - Deprecate the "extra" shuttle areas that are no longer needed and update shuttle areas in unit tests
 - Greatly refactors the shuttle datums and computers to reduce near-duplicate code in different types.
 - This means that more than one shuttle can use the same docking port (not at the same time of course).

landmark shuttle demo

#### This all required a bit of infrastructure improvements.
 - ChangeArea proc, for changing the area of a turf.
 - Fixed lighting overlays actually being able to be destroyed.
 - Added a few utility macros and procs.
 - Added "turf translation" procs which are like move_contents_to but more flexible.
 - Added Shuttle Moved and Shuttle Added observers.

#### Mapping Changes
 - Because we replaced the extra shuttle areas with landmarks, maps need to be updated, as well as map-specific shuttle datum definitions.
 - Luckily I wrote a guide on exactly how to convert area shuttles to landmark shuttles:
   [code/modules/shuttles/upgrade_guide.md](https://github.com/VOREStation/VOREStation/blob/4299b8617cc1feb2fd4f43c8e13ffe8c5a351fe6/code/modules/shuttles/upgrade_guide.md)


code/modules/shuttles/shuttles_web.dm
```diff
@@ -1,12 +1,11 @@
 //This shuttle traverses a "web" of route_datums to have a wider range of places to go and make flying feel like movement is actually occuring.
-/datum/shuttle/web_shuttle
+/datum/shuttle/autodock/web_shuttle
 	flags = SHUTTLE_FLAGS_NONE
```
Arokha: IMO should have the same flags as the others. Why not? Maybe not process, but the ZERO_G one at least.

TheFurryFeline:
I really hate that MapDiffBot's broken and there are all these renders just not loading but still listed by it. Wew.

Leshana:
Rebased to remove two backup map files that were accidentally committed.

PR #6760, opened 2020-03-05
Some VChat Tweaks

Arokha:
- Pressing 'Enter' while in a settings input box saves the setting
- Chat remains between client reconnects if your client didn't close (so things like using the reconnect button, or autoreconnects at round end when that feels like working)
- The client doesn't send pings to the server, the server sends pings to the client. This fixes AFK measurements for AFK kick purposes.
- Turn latency indicator into a green/red indicator to show if you're connected, and when clicked will perform a one-time ping (and block doing it again for 10 seconds). It will display '?ms' if it never got a reply, or '999ms' if it did, but it was over 1s.

### DIFF
--- a/code/modules/client/client procs.dm
+++ b/code/modules/client/client procs.dm
@@ -428,6 +428,7 @@ client/verb/character_setup()
 	
 	//Log, disable
 	log_debug("[key_name(src)] reloaded VChat.")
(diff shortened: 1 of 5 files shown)


Arokha:
Updated the draft PR there with changes

PR #6767, opened 2020-03-05
Change chat export naming scheme

Cyantime:
Filenames now default to as close to the byond log name as possible, with the exception that all spaces are now underscores, and the time of day is in 24 hour format.

The filenames will now be in chronological order when sorted alphabetically.


Arokha:
You made sure it's willing to save this filename style right

Cyantime:
file
Seems pretty willing to me, file saves properly.

PR #6768, opened 2020-03-05
[Map] Revert Teleporter Removal

TheFurryFeline:
Reverts https://github.com/VOREStation/VOREStation/pull/6703 manually. Couple extra changes listed below.

Changelog notes: 

- Reverts the removal of Secondary Command Teleporter and Secondary Command Substation. 

- Remove paw held tele.

- Extra changes as follows below.

- - Changes area names to be Secondary Command. 

- - RCON tags in Substation Room also changed accordingly so it isn't all the way at the top.

- - Chest drawer moved to replace closet in the corner.

- - Air alarm, Fire alarm, Newscaster re-added per QOL changes that was integrated in previous changed version of the Meeting Room.

- - Electrified windows in inner area un-electrified.

- - Blast doors in inner area removed for the window area.

- - Extra air supply pipe removed under the door between hallway and main office area. (Almost didn't catch that one. Wew.)

- - Donut box re-added into Meeting Room as well.


maps/tether/tether-05-station1.dmm
```diff
@@ -12948,18 +13403,53 @@
...
+	},
+/obj/structure/cable/green{
+	d2 = 8;
+	icon_state = "0-8"
+	},
+/turf/simulated/floor/tiled,
+/area/bridge/secondary/teleporter)
+"awY" = (
+/obj/structure/table/standard,
+/obj/item/weapon/hand_tele,
```
Heroman3003: My main point of contention here would be the presense of a second Hand Tele. I still think that in terms of those, only one should be available.

Arokha:
```suggestion

```

Can just remove it if that's desirable.

Cyantime:
Just here to say that there really doesn't need to be an entire substation for like three rooms and a teleporter, it doesn't use all that much power.

maps/tether/tether-05-station1.dmm
TheFurryFeline: I'd like to hear the admins' input regarding that obj and whether or not it should be kept in. Purrsonally, I don't see much of an issue with it being there. It can be used for televore and more of that can be good for scenes as a whole.

Arokha:
> Just here to say that there really doesn't need to be an entire substation for like three rooms and a teleporter, it doesn't use all that much power.

I think it has other uses from admins... I remember an admin being surprised that it was missing because it was like their goto SMES for doing shenanigans attached to or something. Maybe someone who does events can comment and say something about that.

Heroman3003:
And some less contentious points;

- I think substation wasn't necessary even with teleporter being there and was overkill. Being part of either Civilian grid or even not being on a subgrid at all would have worked just fine.

- If you reverted it by copying and pasting old one, then it will still have the weird 'move up wire' spot in the middle of hallway. EVA is now on exploration grid so that wire going up is not longer needed.

- Same as previous, if you manually copied it rather than just re-adding teleporter, some QoL adjustments ended up also being lost. Primarily missing air alarm and newscaster to meeting room, but also some spaghett with window wires and shutters covering entire section without cutting off the meeting room.

TheFurryFeline:
> And some less contentious points;
> 
> * I think substation wasn't necessary even with teleporter being there and was overkill. Being part of either Civilian grid or even not being on a subgrid at all would have worked just fine.
> * If you reverted it by copying and pasting old one, then it will still have the weird 'move up wire' spot in the middle of hallway. EVA is now on exploration grid so that wire going up is not longer needed.
> * Same as previous, if you manually copied it rather than just re-adding teleporter, some QoL adjustments ended up also being lost. Primarily missing air alarm and newscaster to meeting room, but also some spaghett with window wires and shutters covering entire section without cutting off the meeting room.

Thank you, forgot about the wire there for command grid that used to lead to main EVA. Removing that bit of wire. Will also re-add the air alarm and further bits and bobs.

Heroman3003:
To note another change I remembered. Windows between the little hallway and bridge itself are electrified. This is maximum 'why' question to me, so I removed those wires as well. I understand windows between bridge command room and the park but between internal bridge areas?

Wickedtemp:
10/10, teleporters can be fun.

TheFurryFeline:
Oh, also, Hero. It seems you forgot to actually connect the changed Secondary Bridge area to the main lines.

Heroman3003:
That's because I connected it to Civilian subgrid.

TheFurryFeline:
> That's because I connected it to Civilian subgrid.

It still isn't getting its power. Not connected to any grid whatsoever.
Secondary_Command_With_Fugged_Wires

Very-Soft:
I can confirm, I've seen the secondary command area completely drained of power and dead well into the round while nothing else is struggling for power. 

And, while I can say, sure perhaps a substation isn't -required-, I don't see the harm in having one there to support the secondary command area.

That said, the original idea behind putting in a second teleporter was to allow for an extra alternative to the elevator for emergencies and antagonists, when we should have them. I deffo wouldn't have a problem with the second hand tele being removed, that seems reasonable to me, I didn't even think about it when I copied it over. But with the elevator being such an easy chokepoint to lock down, I think having more reliable alternatives to it is good.

TheFurryFeline:
> That said, the original idea behind putting in a second teleporter was to allow for an extra alternative to the elevator for emergencies and antagonists, when we should have them. I deffo wouldn't have a problem with the second hand tele being removed, that seems reasonable to me, I didn't even think about it when I copied it over. But with the elevator being such an easy chokepoint to lock down, I think having more reliable alternatives to it is good.

This seems to be all the reason needed for the teleporter to stay in. And the tele itself being another item to use for transportation, that's part of why I'd like it to stay in. If people are abusing it, like was apparently assumed, then admins should get notified to deal with it.

Heroman3003:
Wait what. It was definitely connected to the power grid through a green wire going down from that apc in maint and through the meeting room maint access. I even tested locally just to see that it was charging.

Heroman3003:
bridge

This is how it looked when I PR'd it. So clearly, some PR after it messed it up.

TheFurryFeline:
Ah, that's possible..... Might have been an un-updated branch

Screemonster:
Now I'm actually getting a chance to reply to this (I've been at work all week) I'mma address a couple of things brought up in the former PR.

For the comments that the teleporter is unnecessary - an alternative means to get from surface to asteroid and back is still useful in the event that the elevator is compromised. It may seem "unnecessary", but the fire escape in a RL building is unnecessary until there's a fire.

As for the comment that it buffs the hand tele by having a second teleporter available for it to hook into - compared to other maps, the hand tele is significantly less useful on the tether because there are _fewer_ teleporters. Box, for example, has two easily-accessible teleporters in command and the telecomms sat, and a third is available if you take a jaunt out to the derelict and power that one up too. If anything, having two station-accessible teleporters brings the hand-tele up to _almost_ the standard that every other map has had for years.

PR #6777, opened 2020-03-07
Removes genetics from secure storage and from circuit printer.

Sebbe9123:
Thought we removed genetics before, someone made a mistake if not.


Heroman3003:
I suggest keeping cloning part of genetics. But removing just the DNA pod. That way if somebody wants to build oldschool cloning rig for wahtever reason they can, but powers stuff is still properly inaccessible.

maps/virgo/virgo-1.dmm
```diff
@@ -5992,7 +5992,7 @@
 "cll" = (/obj/structure/sign/securearea{desc = "A warning sign which reads 'HIGH VOLTAGE'"; icon_state = "shock"; name = "HIGH VOLTAGE"},/turf/simulated/wall/r_wall,/area/storage/tech)
 "clm" = (/obj/item/device/aicard,/obj/item/weapon/aiModule/reset,/obj/structure/table/steel,/turf/simulated/floor/plating,/area/storage/tech)
 "cln" = (/obj/machinery/atmospherics/unary/vent_scrubber/on{dir = 4},/turf/simulated/floor,/area/storage/tech)
-"clo" = (/obj/structure/table/rack{dir = 8; layer = 2.9},/obj/item/weapon/circuitboard/cloning{pixel_x = 0},/obj/item/weapon/circuitboard/clonescanner,/obj/item/weapon/circuitboard/clonepod,/obj/item/weapon/circuitboard/scan_consolenew,/obj/item/weapon/circuitboard/med_data{pixel_x = 3; pixel_y = -3},/obj/machinery/atmospherics/pipe/simple/hidden/scrubbers{dir = 4},/turf/simulated/floor/plating,/area/storage/tech)
+"clo" = (/obj/structure/table/rack{dir = 8; layer = 2.9},/obj/item/weapon/circuitboard/cloning{pixel_x = 0},/obj/item/weapon/circuitboard/clonepod,/obj/item/weapon/circuitboard/med_data{pixel_x = 3; pixel_y = -3},/obj/machinery/atmospherics/pipe/simple/hidden/scrubbers{dir = 4},/turf/simulated/floor/plating,/area/storage/tech)
```
TheFurryFeline: This is not needed as it is part of an unused collection of maps. We use tether now.

Sebbe9123: Doesn't hurt then, keeps code up to date even if it's not used.

Novacat: I am not even sure why we keep these old backups around.

Heroman3003:
As a whole, my feelings about genetics is that 4/5 positive powers we still have really fall heavy into the whole "Mechanics that just encourage shitty behaviour", with little positive use. Those being, XRAY, which serves no purpose beyond peeking and I believe even ICly is banned onstation. Remote Speak, which 100% of the time is used for bad telepathy memes, often against people who want no part in it just because their name stood out in target list. Telekinesis, embodiement of LRP shit with "lol look ma im throwing things around". And Hulk which is Hulk. Temperature resist is only power that seems to have place around here, which doesn't paint a good picture.

As for negative traits? We only have 3. Blindness, Deafness and Clumsiness. Where in the world are there more positive traits than negative ones? This makes genetics even more of a cakewalk.

As a whole, I agree with removal. Because I don't believe this is removing something actually functional or cool. This is removing last remnant of forgotten and deprecated thing that just wasn't removed with the rest of it and that doesn't belong.

Screemonster:
Looking at the commit title, this is literally a salt PR because someone was a shitter and we already banned that person. The last time genetics was ever a problem was literally months ago.

Go ahead and make better powers, genetics can definitely do with improvement. Replace powers entirely if you want to, but that'll take an overhaul 'cause they're baked pretty deep into life code and whatever you do will probably have to be done upstream in order to be remotely compatible.

MrSign:
I agree with Scree.  The person that misused genetics was dealt with on an administrative level and to go ahead and attempt to remove the possibility of genetics is just a step too far.

Screemonster:
Had a quick look in the logs for the shift that prompted this and _multiple_ people were injected with DNA modifiers with only one person causing an actual issue. The only problem is that you don't notice the people that behave themselves, so of course it appears that the "only time" remote talk is used is for annoying you, because that's the only time you see it being used.

PR #6779, opened 2020-03-08
Makes pregame time a config setting

Arokha:
Boo hardcoded values


Screemonster:
this sounds like a useful upstreamy thing unless they don't want it

Arokha:
already did it

PR #6789, opened 2020-03-09
Port Bay's inactivity subsystem

Arokha:
Slightly better than ours and happens to alleviate a horrible server-destroying bug that happens randomly that's apparently in every codebase that uses SMC.


KasparoVy:
Blessed subsystem. WTF was the bug?

PR #6802, opened 2020-03-11
Fixes the rest of Handle_shock not working

SplinterGP:
consequently fixing vasilissan and xenochim cold handle that does not work. (unless the one we have in YW is different)


Arokha:
https://github.com/PolarisSS13/Polaris/pull/6871

PR #6836, opened 2020-03-12
Add 'sparks' option when spawning characters

Arokha:


code/modules/admin/verbs/randomverbs.dm
```diff
@@ -445,15 +445,17 @@ Traitors and the like can also be revived with the previous role mostly intact.
...
 	var/spawnloc
+	var/sparks
 
 	//Where did you want to spawn them?
 	switch(location)
 		if("Right Here") //Spawn them on your turf
-			if(!src.mob)
-				to_chat(src, "You can't use 'Right Here' when you are not 'Right Anywhere'!")
-				return
-
```
Novacat: Is there a reason for removing this check?

KasparoVy:
That's dope. Good idea

code/modules/admin/verbs/randomverbs.dm
Arokha: Yes, it turns out it's not possible to ever have a client that doesn't have a mob. I probably didn't know that when I wrote this originally, so the check is pointless.

Clients always have a mob, and clients are either disconnected, or in one.

PR #6842, opened 2020-03-12
Move using_map to GLOB

Arokha:
Title

Can't go to Polaris until after Overmap goes to Polaris


Arokha:
Will need to be redone later.

PR #6845, opened 2020-03-12
Sector tweaks

Arokha:
Makes random picked sectors truly anonymous (no description or name) until scanned (aside from planetary ones having a planet icon).

Readds snow_outpost. I dunno why it was disabled. It had some high numbers for mapgen stuff and the comment seems to indicate that was related. So maybe that's it. I reduced the numbers.


Heroman3003:
The issue with snow_outpost wasn't the numbers but way areas were mapped out. It kept doing stuff like spawning bushes/grass/whatever indoors, inside caves, on top of a lake, etc. Density was very problematic, yes, but at least bearable compared to the fact that it affected areas it definitely shouldn't have.

Leshana:
I think it could work.  Should it return any scan data tho?

Heroman3003:
To add a thing I posted in #development.

I think just general ability to freely access gateway z-levels, at least ones without pre-calibrated gateway, defeat the purpose of even having the gateway spread landmarks intended to separate the team and make findign way back part of the challenge.

Arokha:
Ok, I redid snow_outpost some to make caves dark and get spawns in the proper places. It's not perfect, but it's a lot better! Also removed the landing waypoints on 'randomized gateway' destinations, though left them on the overmap so in theory you could scout and figure out what gateway destination was that shift if you wanted to, just can't circumvent the scatter by landing there with shuttle (someone could code in spawning the landing landmarks later if they want when the gateway is calibrated).

PR #6846, opened 2020-03-12
ATC announcements for coming/going overmap ships

Arokha:
Unless they're cloaky


KasparoVy:
Badass. Good addition

PR #6848, opened 2020-03-13
Updates smaller admin ship

Arokha:
Not as huge as the GARGANTUAN BEAST Ascian made (no I'll never stop poking fun at). Also removes all the old areas and unnecessary things to clean up the tether files.

ALSO fixes loading new z-levels with lighting.

I guess this is fluff?


Heroman3003:
Was backup of tether10colony left in intentionally?

maps/tether/tether-10-colony.dmm
```diff
@@ -25273,7 +25273,7 @@
 /area/unknown/dorm3)
 "WI" = (
 /turf/simulated/floor/wood,
-/area/houseboat)
+/area/space)
```
Novacat: I have a feeling that, given this is a simulated turf, it is not supposed to be area/space.

Arokha: Good point, but it's YOUR ship that it's used on :p

image

I'll fixy.

Arokha:
> Was backup of tether10colony left in intentionally?

Nope, removed. We should probably add those backup folders to our gitignore

PR #6850, opened 2020-03-13
Make shuttle corner turfs not look bad

Arokha:
It's because they weren't used to the new space, and also because they've always been bad at lighting


code/game/turfs/simulated/floor_types.dm
```diff
@@ -83,11 +83,30 @@
...
+		for(var/diag in cornerdirs)
+			var/image/I = image(LIGHTING_ICON, null, icon_state = "diagonals", layer = 10, dir = diag)
+			I.plane = PLANE_LIGHTING
+			antilight_cache["[diag]"] = I
 
 /turf/simulated/shuttle/Destroy()
 	landed_holder = null
 	..()
 
+/turf/simulated/shuttle/proc/update_breaklights()
```
Leshana:
```suggestion
// For joined corners touching static lighting turfs, add an overlay to cancel out that part of our lighting overlay.
/turf/simulated/shuttle/proc/update_breaklights()
```

Arokha:
https://github.com/PolarisSS13/Polaris/pull/6830

PR #6851, opened 2020-03-13
Fix shadekin zmove

Arokha:
Should fix shadekin zmove. Definitely won't break anything, but not super tested to see if it actually fixes. Probably does though.


Arokha:
Already sent to pol

PR #6860, opened 2020-03-14
Redo excursion shuttle cockpit

Arokha:
So that the engines console can fit. Also why recliners? So silly.

image

### DIFF
--- a/maps/tether/tether-07-station3.dmm
+++ b/maps/tether/tether-07-station3.dmm
@@ -2210,6 +2210,9 @@
 	dir = 1;
 	pixel_y = 5
 	},
+/obj/machinery/atmospherics/unary/vent_scrubber/on{
+	dir = 4
+	},
 /turf/simulated/floor/tiled/steel_ridged,
 /area/shuttle/excursion/cockpit)
 "dD" = (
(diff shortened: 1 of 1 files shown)


Novacat:
Make sure the cockpit is access locked to pilot.

TheFurryFeline:
Also, nitpick but I'd rather the pipes not go under the walls :V I think you can safely re-path it to go under the doors without it cross anything else?

Arokha:
> Also, nitpick but I'd rather the pipes not go under the walls :V I think you can safely re-path it to go under the doors without it cross anything else?

Which pipes are you referring to?
If you mean the fuel pipes: No, because then it would share a path with the grey pipe, with which it would connect. I also didn't add that, it was like that before and has been like that since this shuttle design was added. Scrubbers, supply, and *one normal colored pipe* can share a turf. You can't have both yellow and grey on the same turf or they will connect. Unfortunately we don't have /vg/s super pipe conduit things where you can have like 8 things on a turf.

If you mean the scrubbers/supply pipes in reference to them going under a window, that's under a window.

TheFurryFeline:
> > Also, nitpick but I'd rather the pipes not go under the walls :V I think you can safely re-path it to go under the doors without it cross anything else?
> 
> Which pipes are you referring to?
> If you mean the fuel pipes: No, because then it would share a path with the grey pipe, with which it would connect. I also didn't add that, it was like that before and has been like that since this shuttle design was added. Scrubbers, supply, and _one normal colored pipe_ can share a turf. You can't have both yellow and grey on the same turf or they will connect. Unfortunately we don't have /vg/s super pipe conduit things where you can have like 8 things on a turf.
> 
> If you mean the scrubbers/supply pipes in reference to them going under a window, that's under a window.

Was the first one, yeah. Shame, though.

PR #6861, opened 2020-03-14
Add github actions

Arokha:
Labeler: Autolabels PRs with "Type: Map" or "Type: Icon" if files in the maps or icons folder are changed, respectively.

Stale: Warns on issues older than 365 days that they will be closed unless updated within 30 days. If nobody has fixed after 365 days, it's probably either not an issue anymore, or nobody cares so there's no point in leaving it.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,7 @@
+# Map changes
+"Type: Map":
+  - maps/**/*
+
+# Icon Changes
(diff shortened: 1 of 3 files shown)


SpadesNeil:
:o

Arokha:
o:

Arokha:
This is going to destroy the universe so let's see how it goes.

PR #6862, opened 2020-03-14
Scale Skybox by client view size

Leshana:
Add proc `set_viewsize()` to allow hooking actions when `client.view` is changed.
  - Proc is on mob to simplify the standard use case; maybe someday actually save it there too if desired.
Simplify skybox parallax code and add support for scaling for larger client.view sizes.
  - Hooked up skybox to recalculate it's scaling when the view changes.
  - Added constant defines to clarify where the magic numbers are coming from.

### DIFF
--- a/code/modules/admin/verbs/randomverbs.dm
+++ b/code/modules/admin/verbs/randomverbs.dm
@@ -837,10 +837,12 @@ Traitors and the like can also be revived with the previous role mostly intact.
 	if(!holder)
 		return
 
+	var/view = src.view
 	if(view == world.view)
 		view = input("Select view range:", "FUCK YE", 7) in list(1,2,3,4,5,6,7,8,9,10,11,12,13,14,128)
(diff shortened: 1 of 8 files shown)


code/modules/admin/verbs/randomverbs.dm
```diff
@@ -837,10 +837,15 @@ Traitors and the like can also be revived with the previous role mostly intact.
...
 		return
 
+	var/view = src.view
 	if(view == world.view)
 		view = input("Select view range:", "FUCK YE", 7) in list(1,2,3,4,5,6,7,8,9,10,11,12,13,14,128)
 	else
 		view = world.view
+	if(mob)
+		mob.set_viewsize(view)
+	else
```
Arokha: `mob` can never be unset. Attempting to set it to null does nothing. It's protected by byond, so no need to test.

PR #6864, opened 2020-03-14
Further tweak to shuttle corner turf appearances

Arokha:
Removes some weirdness.


Arokha:
I added to my existing PR for this to them

PR #6870, opened 2020-03-15
Fix chat exporting when someone has used unicode

Arokha:
Only affects clients still using 512


TheFurryFeline:
Thank youuuuuuuuuuu

TheFurryFeline:
It has been

PR #6888, opened 2020-03-16
Adopt Bay vendor sprites

Arokha:
I know you'll sass me about this, given I just swapped two other vendors to Eris instead.

So, as people pointed out the Eris vendors aren't quite in the same style as our other vendors, plus they're missing one thing Bay's vendors have, which is all 4 direction sprites on every vending machine. That should make maps look a bit better, given that you won't have vending machines on S,E,W walls facing south looking weird.

They also all have deny and vend sprites, which many of the current ones are missing. They don't have the cool giant displays a lot of Eris ones have, buuut... the directional sprites are appealing from a mapping perspective.

Also the fridges, chem drink and food, show little icons for how much stuff they have in them so you can see if they are stocked from the icon, and how stocked they are.

dreammaker_2020-03-16_18-53-41

dreammaker_2020-03-16_18-56-22

Oops also forgot dispenser stuff, also has directions:
image

If merged, I'll have to go through the map and SET the directions on everything but that's fine.


Heroman3003:
While this was a sin with old regular chem dispensers too, new ones are even worse at the fact that they are actually not 'big' floor-mounted objects, but small ones, put on top of tables and walkable-over. Also, no unique sprites for xenobotany/biochem dispensers.

Also, wallmed sprites probably shouldn't be directional like _that_ considering way they are tilted. I think just rotating front facing sprite would be better for it.

And this one is more personal, but I think directional side sprites on vendors should be more aligned with edges of the tile. That way they can actually appear like they are pushed up against the wall. Because if you use offsets in-mapping for that, then you get a fucky when trying to unwrench and move them.

Arokha:
> While this was a sin with old regular chem dispensers too, new ones are even worse at the fact that they are actually not 'big' floor-mounted objects, but small ones, put on top of tables and walkable-over. Also, no unique sprites for xenobotany/biochem dispensers.

I don't know what this has to do with this PR. We didn't have them before, we don't have them now?

> wallmed sprites probably shouldn't be directional like that

I wasn't planning on rotating any of the wallmed sprites because we don't have walls that work with them, but I was going to do that on Aubergine which uses Bay walls. Good to have the icons available for maps that use those walls though.

>but I think directional side sprites on vendors should be more aligned with edges of the tile.

I would agree, except that if you move them away from the center, the further they move the more bizarre it is that you can't walk up to their front all the way. This applies to any structure in SS13 though.

Arokha:
Am being told Polaris is also going 3:4 so I will try sending it to them instead first. If they don't take it, I'll just do this one.

Heroman3003:
> I don't know what this has to do with this PR. We didn't have them before, we don't have them now?

Yes we did. Dispenser_small was used for biochem and green version of it was used for xenobotany. Unless the same ones are still used. They were pretty much reshapes/recolors of base one but they did distinctify those two a little.

> I wasn't planning on rotating any of the wallmed sprites because we don't have walls that work with them, but I was going to do that on Aubergine which uses Bay walls. Good to have the icons available for maps that use those walls though.

Huh, I was sure for some reason current ones had directional sprites that are just 90 degree rotations. Nevermind then.

PR #6890, opened 2020-03-16
Port Bay frame icons

Arokha:
Proper 3/4ths. Also people have been asking for directional console icons, so here you go!

There's no directional MACHINE icon, so I made one by adding a little chevron that points in the direction the machine is going to face.


Heroman3003:
Is there no difference between "board inserted" and "board secured" stages for machine frames?

Arokha:
There is, the slots at the bottom fill in. I had to make those though, so if you want to submit alternatives go for it. Might be hard to see in those microscopic 32x32 sprites in a browser. Inflate them to normal byond size.

Arokha:
Am being told Polaris is also going 3:4 so I will try sending it to them instead first. If they don't take it, I'll just do this one.

Heroman3003:
Ah, didn't notice it at first because of smol image size on the bot, nvm.

Novacat:
Polaris is still deliberating so going to merge this. Can always remove the _vr stuff later if/when Polaris does merge.

PR #6052, opened 2019-10-13
New optional Vore sounds, release and internal loop

Poojawa:
## About The Pull Request

Hello! I am porting my commissioned vore related effects here, I still hold the rights to them as well under a general license. 

This PR added a new set of options to the vore menu:

### Toggle of "Fleshy" type bellies

Turn off to change struggle sound effects to generic rustling. No more will your pockets sound like fleshy things. It will also enable you to turn the optional preyloop sound effect on or off, depending on what environment you want your prey to hear.

### Release Sound toggle

Change this to have something other than a wet splatter effect upon releasing something from a belly.

### Fancy Vore Sounds

Toggle this to enable the new sound effects, keep off to keep preferred 'classic' sounds that you've had forever. This is a per-belly toggle. Sounds to your prey are different on fancy, they should be far more richer, since they're amidst the mess. Prey inside another person should be able to hear predator-quality sounds, since I was sure people would appreciate it.

## Why It's Good For The Game

New sounds, lots of teasings, what's not to love?

This also makes the sound preference toggles show up in your preference tab, for easy clicking on/off. Why that wasn't a feature when it was added... I have no idea. 

## Changelog
:cl: Poojawa
add: Added new sound effects for vore, toggle the fancy mode to hear 'em!
fix: Vore Sound preferences are now in your preference tab to be toggled more easily.
/:cl:

PR #6891, opened 2020-03-17
Vore Code cleanup

ShadowLarkens:
This PR just refactors vore code and makes it a little less ugly.
If anything at all changes in the player experience, it means I messed up :P


code/modules/vore/eating/contaminate_vr.dm
```diff
@@ -65,28 +65,18 @@ var/list/gurgled_overlays = list(
...
-		var/turf/location = user.loc
-		if(!isturf(location)) return
-
-		var/obj/item/I = O
-		if(!I || !istype(I,/obj/item)) return
-
-		to_chat(usr, "<span class='notice'>You start washing \the [I].</span>")
+		if(!istype(O))
+			return
+		to_chat(user, "<span class='notice'>You start washing [O].</span>")
```
Arokha: Backwards or unnecessary. It would runtime on `if(O.gurgled)` if it's not /obj/item because that's where `gurgled` is declared. You could do `if(istype(O) && O.gurgled)`.

code/modules/vore/eating/leave_remains_vr.dm
```diff
@@ -1,75 +1,65 @@
...
+	skull_type = /obj/item/weapon/digestion_remains/skull/teshari
+/datum/species/vox
+	skull_type = /obj/item/weapon/digestion_remains/skull/vox
+/datum/species/xenohybrid
+	skull_type = /obj/item/weapon/digestion_remains/skull/xenohybrid
 
-	if(istype(M,/mob/living/carbon/human))	//Are we even humanoid?
+/obj/belly/proc/handle_remains_leaving(var/mob/living/M)
+	if(ishuman(M))	//Are we even humanoid?
 		var/mob/living/carbon/human/H = M
```
Arokha: I'd prefer `if(!ishuman(M)) return` just because you can unindent the entire proc by one after it.

code/modules/vore/eating/simple_animal_vr.dm
```diff
@@ -46,7 +47,8 @@
 		if(confirm == "Enable")
 			vore_selected.digest_mode = DM_DIGEST
 			spawn(20 MINUTES)
-				if(src)	vore_selected.digest_mode = vore_default_mode
+				if(src)
+					vore_selected.digest_mode = vore_default_mode
```
Arokha: `addtimer(VARSET_CALLBACK(vore_selected, digest_mode, vore_default_mode), 20 MINUTES)`

Terrifying but probably better.

code/modules/vore/eating/vorepanel_vr.dm
```diff
@@ -378,7 +382,6 @@
 
 	if(href_list["close"])
 		qdel(src)  // Cleanup
```
Arokha: Does qdel on browser objects close them?

code/modules/vore/resizing/grav_pull_vr.dm
```diff
@@ -29,35 +29,36 @@
...
-	for(var/atom/X in orange(pull_radius, location))
+	for(var/X in orange(pull_radius, location))
 		// Movable atoms only
-		if(istype(X, /atom/movable))
-			if(istype(X, /obj/effect/overlay)) continue
-			if(X && !istype(X, /mob/living/carbon/human))
+		if(ismovableatom(X))
+			if(istype(X, /obj/effect/overlay))
+				continue
+
```
Arokha: Can invert this, combine them, and do `continue` to unindent most of the proc.

code/datums/browser.dm
```diff
@@ -33,6 +31,11 @@
 		return
 	add_stylesheet("common", 'html/browser/common.css') // this CSS sheet is common to all UIs
 
+/datum/browser/Destroy()
+	ref = null
+	user = null
+	. = ..()
+
```
Arokha: `//VOREStation Edit` or send to Polaris or define in another file.

code/modules/vore/eating/contaminate_vr.dm
tigercat2000: it checks istype(O), and `O` is defined as `obj/item/O`
I can change the parameter to `I` if you like

code/modules/mob/living/living.dm
```diff
@@ -32,6 +32,8 @@
 	if(buckled)
 		buckled.unbuckle_mob(src, TRUE)
 	qdel(selected_image)
+	qdel(vorePanel) //VOREStation Add
```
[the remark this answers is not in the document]
ShadowLarkens: Oh so that's why they weren't GCing correctly

code/modules/vore/eating/bellymodes_vr.dm
```diff
@@ -61,34 +53,36 @@
...
+						digest_item(I)
+					else
+						digest_item(I)
+						did_an_item = TRUE
+					if(prob(25)) //Less often than with normal digestion
+						play_sound = pick(pred_digest)
 
 		//Handle eaten mobs
-		else if(isliving(A))
+		if(isliving(A))
```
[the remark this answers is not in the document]
ShadowLarkens: Some else's in the code I refactored were completely pointless because the above would always return, but that's not true in this case, will fix.

code/modules/vore/eating/contaminate_vr.dm
ShadowLarkens: Whoops!

code/modules/vore/eating/leave_remains_vr.dm
ShadowLarkens: Agreed

code/modules/vore/eating/simple_animal_vr.dm
ShadowLarkens: Oh y'all do have timers okay

code/modules/vore/eating/vorepanel_vr.dm
ShadowLarkens: It does now!

code/modules/vore/resizing/resize_vr.dm
```diff
@@ -156,37 +157,40 @@ var/const/RESIZE_A_SMALLTINY = (RESIZE_SMALL + RESIZE_TINY) / 2
...
-				to_chat(src,STEP_TEXT_OWNER(tail.msg_prey_stepunder))
-				to_chat(tmob,STEP_TEXT_PREY(tail.msg_owner_stepunder))
-			else
-				to_chat(src,"You run between [tmob]'s legs.")
-				to_chat(tmob,"[src] runs between your legs.")
+				src_message = tail.msg_prey_stepunder
+				tmob_message = tail.msg_owner_stepunder
+
+		to_chat(src, STEP_TEXT_OWNER(src_message))
+		to_chat(tmob, STEP_TEXT_PREY(tmob_message))
```
Leshana: These lines can end up sending `to_chat(src, "")` which isn't really the most wonderful thing.

code/modules/vore/eating/bellymodes_vr.dm
```diff
@@ -105,259 +107,195 @@
...
-				digest_alert_prey = replacetext(digest_alert_prey,"%pred",owner)
-				digest_alert_prey = replacetext(digest_alert_prey,"%prey",M)
-				digest_alert_prey = replacetext(digest_alert_prey,"%belly",lowertext(name))
-
-				//Send messages
-				to_chat(owner,"<span class='notice'>" + digest_alert_owner + "</span>")
-				to_chat(M,"<span class='notice'>" + digest_alert_prey + "</span>")
+	if(digest_mode == DM_TRANSFORM)
+		process_tf(tf_mode, touchable_mobs)
+		return SSBELLIES_PROCESSED
```
Leshana: Returning early here skips the code at the bottom of the switch statement for play_sound or updating vore panel (the flags for which may have been set above)

code/modules/vore/eating/bellymodes_vr.dm
```diff
@@ -105,259 +107,195 @@
...
+									if(istype(I,/obj/item/weapon/reagent_containers/food) || istype(I,/obj/item/organ))
+										digest_item(I)
+									else
+										items_preserved |= I
+									if(prob(25)) //Less often than with normal digestion
+										if(H && H.client && H.is_preference_enabled(/datum/client_preference/digestion_noises))
+											SEND_SOUND(H, prey_digest)
+										play_sound = pick(pred_digest)
+								if(IM_DIGEST)
+									if(!H && (I.digest_stage && I.digest_stage > 0))
```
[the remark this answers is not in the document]
ShadowLarkens: Oh man I horribly mangled this code because I tried to turn it into a function and didn't actually look back at the main code before trying to revert my change by hand without undoing all of the other changes. Will fix.

code/modules/vore/eating/bellymodes_vr.dm
```diff
@@ -105,259 +107,195 @@
...
-					else
-						owner.nutrition += (nutrition_percent / 100)*4.5*compensation
-				to_update = TRUE
+	var/digestion_noise_chance = FALSE
+	var/steal_nutrition = FALSE
 
-				continue
+	switch(digest_mode)
+		if(DM_DIGEST)
+			digestion_noise_chance = 50
```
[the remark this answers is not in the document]
ShadowLarkens: Uhm, am I? It looks like in the original code, the prob(50) only applies to making the sounds, not running any of the damage code

ShadowLarkens:
Note, I noticed that the `preyloop` code (added in #6052) wasn't working correctly: It, by all means I can see, seems to be designed to keep people hearing the preyloop sound continuously, given that it is supposed to repeat the playsound_local call every 52 seconds, which is the exact length of the audio file it plays. It does not work, however, in the current code, because it constantly runs stop_sound_channel() on every tick of process_belly, thus interrupting itself and just making about 2-3 seconds at the start of the file play every time process_belly procs (it actually gets pretty annoying if you listen to it for long enough without any other digestion noises distractng you.)

However, I'm not sure if this fix is actually desirable-- this means that there is now a continuous background track whenever you're in a is_wet && wet_loop belly, that never stops, and can only be turned off by disabling digestion sounds completely. The PR that added it was obviously merged, but given that it was broken at that point and has been since then, I'm not sure if this will be a welcome change or not. 

If it's not desirable, the code should probably be removed entirely, instead of reverted back to the really buggy state that it is currently. It's a 52 second file included in the resource download that can only ever have 3 seconds off the start played, so...

Before: https://soundcloud.com/user-171680439/beforeloopbroken/s-XMa7kcrDfQB?in=user-171680439/sets/vorestationvorestation6891/s-JunfOEn3MGB

After: https://soundcloud.com/user-171680439/afterloopfixed/s-Sjtgfy2kBZm?in=user-171680439/sets/vorestationvorestation6891/s-JunfOEn3MGB

code/modules/vore/eating/bellymodes_vr.dm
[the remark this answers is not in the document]
Leshana: I checked out the original code to look, yes you're correct! The indenting in the github diff made it look like the whole thing was under that if statement.   I guess that is what happens when a language uses only whitespace to delimit code blocks.

PR #6899, opened 2020-03-17
Rewrite how Travis works

Arokha:
Might take a few tries to get it working right. Let's see how it runs.


Arokha:
Man how'd I get that on the first try. Okay time to push things that should be caught by validate_files.sh and see if it quits the build early like I want.

PR #6900, opened 2020-03-17
Overmap Events

Leshana:
Ports overmap events from Baystation12

image

### Added Overmap Event/Hazards
- Added /obj/effect/overmap/event objects to the overmap, when a ship enters the same overmap tile, the event is activated on the ship's z-level(s)!
- Random groupings of event hazards are placed on the overmap at round start, quantity configurable in using_map.

### Code enhancements
- Added a few utility functions to support ported code.
- Upgraded SSevents subsystem to actually use MC_TICK_CHECK!  And to populate the overmap with hazards on init.
- Upgraded SSskybox subsystem to add active-event overlays onto the skybox.
- Upgraded dust and meteor spawning code to support being passed target z-levels
- Moved overmap defines to _defines folder and renamed some stuff to make more sense.

### Upgraded some events to support being overmap hazards.
- Upgraded carp_migration, dust, electrical_storm, ion_storm, and meteors events to support being put on the overmap. They now obey affecting_z and location_name(), and mostly work on a tick() basis instead of spawning everything up front.
- Upgraded carp_migration event.
  - No longer requires landmarks.  Instead spawns at edge of map and uses mob AI to direct them towards the middle.
  - Spawns carp in waves instead of all up front.  Has a hard cap it won't exceed.
- Upgraded electrical_storm event
  - Dynamically finds APCs instead of needing landmarks.
  - APCs can be protected by turning them off.
  - Has a skybox affect for the overmap version.
- Upgraded ion_storm event.
  - Makes synthetics confused for a few seconds.
  - Any RAD_SHIELDED areas are now also protected from ion_storms.
  - Has a skybox affect for the overmap version.


TheFurryFeline:
Oooh, I like the carp migration ai thing, but I think that should be ported to Polaris, too.

Heroman3003:
I really don't like the fact that previously overall harmless event now has chance of actually being harmful to APCs with that emag chance thing.

Also (on average) doubling the strength of dust on all levels sounds... Very eh as well. Not mentioning that even lowest level dust now has chance to be more destructive than highest simply because RNG which is very eh.

Also, will carps actively try and break in to reach center of map now even if there is nothing to attack indoors? Because that sucks big ass.

TheFurryFeline:
> Also, will carps actively try and break in to reach center of map now even if there is nothing to attack indoors? Because that sucks big ass.

It doesn't look like the carp AI has been otherwise changed to include destructive = true or some such var associated with destroying their surroundings.

Heroman3003:
@TheFurryFeline Destructive determines whether they will do it to things like tables and windows when IDLE. But all mobs will do that when trying to reach attack target. That is why "stay away from windows" is important during carp events: prevents them from having a target to break windows for. What I'm questioning is whether they'll do it when trying to reach center of map same way they do when trying to reach a person.

Arokha:
You're both wrong. `destructive` has nothing to do with whether they will destroy things to get to their target. `destructive` determines whether or not they will attack things like empty mechs (and that is all that var does at the moment, it seems...)

`violent_breakthrough` is the var that determines if mobs try to smash things in the way to their destination, and it's on by default on all mobs.

Heroman3003:
Regardless, question whether they will break in their way all the way to center of the map stands.

TheFurryFeline:
> You're both wrong. `destructive` has nothing to do with whether they will destroy things to get to their target. `destructive` determines whether or not they will attack things like empty mechs (and that is all that var does at the moment, it seems...)
> 
> `violent_breakthrough` is the var that determines if mobs try to smash things in the way to their destination, and it's on by default on all mobs.

Ah, right, that's the thing I was thinking of. My bad.

TheFurryFeline:
> > You're both wrong. `destructive` has nothing to do with whether they will destroy things to get to their target. `destructive` determines whether or not they will attack things like empty mechs (and that is all that var does at the moment, it seems...)
> > `violent_breakthrough` is the var that determines if mobs try to smash things in the way to their destination, and it's on by default on all mobs.
> 
> Ah, right, that's the thing I was thinking of. My bad.

Actually.... With this knowledge, it may well be likely that Carp will be destroying windows to get to their destination. Will it be limited to the periphery of the station external windows? Not too close?

Heroman3003:
Considering doors are attackable and breakable AND carp having objective on other side, if they are trying to get to it, they will break everything.

Leshana:
I've refactored the carp event to prevent them continuing to push for the center of the map forever.  Also I experimentally added back in optional landmark support.

Heroman3003:
I am still unsure about increase in strength for dust event, especially with how insane RNG it can get with it now.

PR #6906, opened 2020-03-18
[MIRROR] POI wall nonsense

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/6855
--------------------
Fixes (???) the weird issue with the ice maze POI walls being incorrectly named and having no borders. The walls are now slightly darker than the cave tiles. Here is a picture.

image

I'm not a very good coder so this needs review. Once I get the okay, I'll be adding a new wall for POIs with it. Don't want to waste my time adding on to this if I did it really badly, and I also want to do a bit more testing.

Here is the new wall, picture is tiled 3x3
image

MAY, MAYBE, POSSIBLY fixes the weird bug with the cave borders. Not sure, don't know exactly how to replicate the bug so I can't test.


code/game/turfs/simulated/dungeon/wall.dm
```diff
@@ -16,31 +16,87 @@
...
-			var/place_dir = turn(direction, 180)
-			if(!mining_overlay_cache["rock_side_[place_dir]"])
-				mining_overlay_cache["rock_side_[place_dir]"] = image('icons/turf/walls.dmi', "rock_side", dir = place_dir)
-			T.add_overlay(mining_overlay_cache["rock_side_[place_dir]"])
+/turf/simulated/wall/solidrock/New(var/newloc)
+	..(newloc,"bedrock")
 
 /turf/simulated/wall/solidrock/Initialize()
-	icon_state = base_state
+	..()
```
Arokha: Not returning a value in Initialize will cause runtimes

Novacat: Probably should fix upstream.

PR #6914, opened 2020-03-19
Speed up map load significantly

Arokha:
This is mostly @ShadowLarkens doing. I just poked a few places that I knew how.

Server initialization goes from 270s to 130s on my machine.


code/game/turfs/turf.dm
```diff
@@ -34,26 +34,22 @@
...
-		luminosity = 1
-
+/turf/Initialize(mapload)
+	. = ..()
+	for(var/atom/movable/AM in src)
+		Entered(AM)
+
+	//Lighting related
+	luminosity = !(dynamic_lighting)
+	has_opaque_atom = (opacity)
```
Leshana: This will set overwrite true values with false if turf is not opaque, even if it has an opaque atom.
```suggestion
	has_opaque_atom |= (opacity)
```

KasparoVy:
Enter the goddamn speedforce, wow.

code/game/turfs/turf.dm
Arokha: Did this

Arokha:
A large one is probably the removal of global "turfs" list which was just a list every turf added itself to.

KasparoVy:
Cheers Arokha, I appreciate the explanation

PR #6917, opened 2020-03-19
Adds a single event mob.

AmayaWah:
Adds a mob.
Mechanical mob, has catalogue data + custom sprite.

Includes sprite + folder. Folder is tgstation, because I borrowed the sprite from there and I wouldn't want to claim that it's our original sprite.


Ketrai:
you should pester fennecs for sprites snrk~

PR #6920, opened 2020-03-20
Swap air alarm sprites

Arokha:
dreamseeker_2020-03-19_20-18-48

dreamseeker_2020-03-19_20-23-17


Novacat:
I like how angry it gets when the air is bad.

Arokha:
Bad air is something to get angry over.

PR #6946, opened 2020-03-21
Don't announce overmap events

Arokha:


Novacat:
Note that every overmap event has an announcement at present, and, it probably should announce for the ship that is encountering the event and only the ship. Not the rest of the Tether.

TheFurryFeline:
Can you make this thing announce only over exploration or something?

Arokha:
> Can you make this thing announce only over exploration or something?

Sure, though they would probably not hear it, if they didn't set up a relay or anything, making it sort of pointless.

TheFurryFeline:
> > Can you make this thing announce only over exploration or something?
> 
> ~Sure, though they would probably not hear it, if they didn't set up a relay or anything, making it sort of pointless.~
> 
> Actually wait no, there's no guarantee the ship is exploration-related.

Ah, relay's needed for that? Right, then. RIP with that idea.

Cyantime:
Also needed for dust storms, ion storms, electrical storms, etc
It'd probably be best to yank the zlevel specific announcement changes from here https://github.com/Baystation12/Baystation12/pull/20305

Arokha:
> It'd probably be best to yank the zlevel specific announcement changes from here
`command_announcement.Announce` is what it uses, and that's not z-level specific, so I'm not sure what you're referring to.

Did the other events.

Cyantime:
That PR added a zlevel argument to command_announcement.Announce, so you can stick affecting_z from the event into it to only announce to the shuttle being hit.

https://github.com/Baystation12/Baystation12/pull/20305/files#diff-7236a4302d0cc610c91f9e3ba7cac396L32

Arokha:
image

But it uses a magic to_chat to everyone in that z-level which is garbo, and not how our announcements work at all, which use a real intercom radio to send messages.

Arokha:
We'll probably make the sensors computer do an audible_message or something in the future, I just don't want to mess with that right now.

Cyantime:
Oh, I didn't realize announcements here actually used radios, my bad.

PR #6948, opened 2020-03-22
Emergency Crash Fix

Novacat:


TheFurryFeline:
Singular indent issue. Wew. Gotta love when it's so simple to fix and yet causes big issues if it isn't noticed.

PR #6960, opened 2020-03-22
Allowing Custom Species access to DSI - Teshari prosthetic limbs.

Holdelta:
Since Teshari are allowed to be created as a custom race to begin with, I'm not seeing too many issues with this change.  Almost seems overlooked that this exception was not included before.

Tested on local and the main thing to note is that the user has to choose Icon Base as Teshari to avoid any weirdness with gear, otherwise works as expected.


Arokha:
Travis won't build your PRs. Go to https://travis-ci.com and look at the status on the left to find this PR and you can check why it's refusing to build. You may see something like:
image
Which means their bot decided you're bad and you need to e-mail them.

Holdelta:
Travis has responded and unflagged my account. seeing if closing and opening will resolve this.

KasparoVy:
Travis says your code doesn't compile properly but at this point I have no clue why, it's a one-liner change and the define definitely exists
Does it compile for you using the most recent version of BYOND?

Holdelta:
Welp, it passed, and I was using the beta version, 513.1515 if that is of any help.

KasparoVy:
Sweet, 3rd time the charm

Greenjoe12345:
seeing as you've merged my teshari prosthetics, I would think you'll want to make them usable by custom species too?

Arokha:
> seeing as you've merged my teshari prosthetics, I would think you'll want to make them usable by custom species too?

Probably true, can you add that?

Holdelta:
> 
> 
> seeing as you've merged my teshari prosthetics, I would think you'll want to make them usable by custom species too?

Yeah sure, not sure who has jurisdiction on passing this stuff, but its fine to me.  I'm literally just adding a single exception line that I believe should've been in.

PR #6964, opened 2020-03-23
New Message, No Votes Anymore

TheFurryFeline:
Title.

Changelog Notes: 

- Changes the roundvote message into a warning that one hour is left in the shift.


Heroman3003:
Doesn't it happen at 5:28? And rounds last 6:00 now?

TheFurryFeline:
> Doesn't it happen at 5:28? And rounds last 6:00 now?

Saw the message pop up and people spoke up in OOC about it, so this'll be better to have as said message.

Heroman3003:
I mean. Its not '1 hour' after the message. And ideally there should be no message at all, IMO.

TheFurryFeline:
I'd think the hour warning would be good to have, really. As for when the message happens, fair point. Will fix, just need more accurate numbers for it.

Novacat:
Message happens one hour before shift end. this PR is accurate.

PR #7004, opened 2020-03-26
Adds Vespa - maybe no issues this time

AmayaWah:
A ship meant for use within events.
This is a vessel for exploration purposes.
It is meant to be an escape for event-goers. This is a place where events may happen at any time, without warning. People that step foot on this ship must recognise that their RP/ERP will be interrupted by any number of things. Of course, during quiet times, they can do w/e.

This PR adds the Vespa, a prototype courier designed, maintained and owned by Haphaestus Industries.
This vessel comes with a state of the art cloaking device, allowing nimble yet undetectable movement through deep space.
The Vespa is also entirely self-sufficient for a long period of time, allowing expeditions into deep space to last much longer than a normal exploration vessel would permit. This makes it an obvious choice for the exploration crew, but getting to the ship is another thing entirely...


AmayaWah:
YEEEEEEEEEEEEEEEEEEESSSSSSSSSSSSSSSSSSSSSSSS

AmayaWah:
> Needs mapmerge ran

Ran

PR #7007, opened 2020-03-26
Adds an overmap shuttle to medical

Arokha:
I said I was gonna. I did.

dreamseeker_2020-03-26_15-50-27

Also nerfs telesci so medical can do the rescuing. :p Telesci can reach to the surrounding overmap turfs with a radius of 3. So if you build telesci on a ship, you can bring it close to a sector and teleport in and then run away to be immune to counterattack, etc.

(FYI the escape pod that was medical's escape pod is now on z6 at the east end of the hallway by EVA, and is a public escape pod. Having it be gated was weird anyway).


Novacat:
Considering I can see it being used for V2/Aerostat rescues as well, It might not be a bad idea to beef up the airlock's capabilities to deal with bad atmos as well.

TheFurryFeline:
A thought.... Would a pilot's IC experience be required to fly this thing? Would a pilot actually need to be obtained for some paramedic or such to fly this rescue pod?

Arokha:
Mechanically, it's the same access as any other shuttle in terms of the control consoles, so SAR/Field Medic and CMO could fly it. In terms of in-game realism, I would say it does take some skill and training and if your character is flying the shuttle around like "First time, I'm just a master pilot without any training" then they might be pushing realistic character boundaries.

TheFurryFeline:
Figured as much about needing training for it. I've yet to try out the new version of shuttle flying, tbh. >W>💦

Arokha:
Ok, airlock can cycle external air now, and changed the floor decal cross thing.

TheFurryFeline:
Might want to update screenshot

Arokha:
Airlock looks the same, floor decal is just made up of colored corner decals now, it's not worth it really.

TheDavestDave:
there's a few problems I have with the medical shuttle recuse that have easy fixes.
Anyone with access to fly the medical shuttle is etheir a head and needs to stay on the station. Or is on the shuttle that needs saveing. Maybe giving paramedics access to this would be an easy way round it.

most of the time t-science has saved the shuttle crew, when I have been on, have been via accidents  finding the shuttle. It may be worth adding some kind of come get me I am scared button to the explorer's shuttle

PR #7008, opened 2020-03-26
Ports triple and 9x kitsune tails from YawnWider

Natje0:
More fluff to torment Bogen with.

Both types dualcolorable with tips!

First PR I've done in AGES, found a typo in pathing on the triple tailed (was called tripple) yet it SOMEHOW still worked on Yawn, so... IDK. Would appreciate if someone with code readily downloaded on their computer tests it to ensure fixing the typo somehow screws it all up.
Also doing it at 0:30am so sleepybrain.


KasparoVy:
Ooh baby a tripple

PR #7017, opened 2020-03-27
Remove space blood

Arokha:
Sorry blood, you can't come with our spaceship.


KasparoVy:
Shit. How're we supposed to lure the space vampires now

PR #7024, opened 2020-03-27
[MIRROR] Make diagonal shuttle walls look less bad

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/6830
--------------------
Makes them adjust lighting and such to look nicer when over space since space is unlit, so it removes that weird 'dark corner' effect.

***NOTE***: The first commit in this (7c9cc38) might make space look weird under said corner turfs until you have ~~overmap~~ skybox.


Leshana:
This was a merge of code we already have. Nothing to do here.

PR #7030, opened 2020-03-28
Remap back room of med shuttle

Arokha:
image


Novacat:
Probably should make the air canister swappable too, since those occasionally need to get swapped out as well.

Arokha:
Seems impossible, without putting the SMES in a corner and requiring running the wire under a wall which I want to avoid.

PR #7044, opened 2020-03-29
Converts webbing to Initialize

Novacat:


Arokha:
image

wat

Novacat:
I was working on both PRs at the same time. It is not a balance thing, just something that adds an action button to sniper rifles, so I figured would be fine to include.

PR #7054, opened 2020-03-30
Clean up some subsystem things

Arokha:


Novacat:
Changes affecting Polaris files should be sent to Polaris.

Arokha:
Already made PR to them for it

code/modules/vore/eating/bellymodes_vr.dm
```diff
@@ -1,17 +1,13 @@
...
 
 	if(loc != owner)
 		if(istype(owner))
 			loc = owner
 		else
 			qdel(src)
-			return SSBELLIES_PROCESSED
-
-	next_process = times_fired + (6 SECONDS / wait) //Set up our next process time.
+			return
```
Leshana:
```suggestion
			return PROCESS_KILL
```
It just deleted itself, so removing from list is good?  Although... I guess Destroy() will handle it.

Arokha: Hm! I suppose. Though it will do the work twice, yes. No point since STOP in Destroy().

PR #7072, opened 2020-03-31
Adds airlock setup helpers

Arokha:
You just put these on your airlock machines and don't varedit anything except an id_tag and frequency onto the controller.

dreammaker_2020-03-31_17-19-48


Arokha:
Gonna use them on the map

PR #7075, opened 2020-04-01
Allow wings with a secondary color

Arokha:
Adds code to support wings with extra_overlay having a secondary color, like tails do. Currently extra_overlay is just ignored on wings.


KasparoVy:
YEEEES man, massive wing buff

PR #7079, opened 2020-04-01
Adjusts shadekin darkness hud

Heroman3003:
Now icon changes with 25 energy increments, and final icon only shows when you are at maximum.


Very-Soft:
Really nice QOL, thanks! <3

PR #7081, opened 2020-04-01
Improves *flip, adds Traceur trait

Arokha:
- Adds 'agility' species var. Default 20, Tajara and Teshari have 90.
- Traceur trait raises agility to 90.
- *flip lets you dodge projectiles and thrown items for the duration of the animation.
- *flip has 2x the emote cooldown.
- *flip will give you either the ability to jump tables/railings/etc or 1 point of Confusion depending on your species agility as a %.
- Makes *flip rotate counter-clockwise if you're facing west.


KasparoVy:
Even awesomer.

PR #7082, opened 2020-04-01
Makes cloth footwraps colorable in loadout

Heroman3003:
They are already white and coloration applies perfectly for them. Good to not need to go to ColorMate every round.


KasparoVy:
Great idea

KasparoVy:
Ported to Polaris https://github.com/PolarisSS13/Polaris/pull/7014

PR #7085, opened 2020-04-02
Adds Ore/Gem Eater Trait

KasparoVy:
Loosely inspired by some jazz about kobolds eating gems and loosely balanced around the rarity & quality of the ore.
Taking the ore eater trait, you only gain nutrition from ore.

5 raw diamonds or verdantite crystals will feed you well, while it'll take a handful of hematite or carbon to do the job.
Better stick to the good stuff - it's in your best interest and it tastes better anyway!

Nutritional value subject to change on review. I figured these were solid numbers to go with

### DIFF
--- a/code/modules/mining/ore.dm
+++ b/code/modules/mining/ore.dm
@@ -122,3 +122,11 @@
 		C.sample_item(src, user)
 	else
 		return ..()
+
+/obj/item/weapon/ore/attack(mob/living/M as mob, mob/living/user as mob)
+	if(ishuman(M))
+		var/mob/living/carbon/human/H = M
+		if(H.species.eat_ore == 1)
+			H.eat_ore(src)
+			return
+	..()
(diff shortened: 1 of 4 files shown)


code/modules/mining/ore.dm
```diff
@@ -121,4 +121,12 @@
...
\ No newline at end of file
+		return ..()
+
+/obj/item/weapon/ore/attack(mob/living/M as mob, mob/living/user as mob)
+	if(ishuman(M))
+		var/mob/living/carbon/human/H = M
+		if(H.species.eat_ore == 1)
+			H.eat_ore(src)
+			return
+	..()
```
Arokha, line 132:
```suggestion
//VOREStation Add
/obj/item/weapon/ore/attack(mob/living/M as mob, mob/living/user as mob)
	if(ishuman(M))
		var/mob/living/carbon/human/H = M
		if(H.species.eat_ore == 1)
			H.eat_ore(src)
			return
	..()
//VOREStation Add End
```

Needs vorestation edit comments. You can use the commit button on this suggestion to apply it.

KasparoVy: Cheers, good catch

This new suggestions feature rocks.

TheDavestDave:
is there a way to stop players turning themselves into ore crates. Like how we had medics that used trash eater to hide beakers inside them?

KasparoVy:
Players won't become ore crates at all, the ore gets deleted when you eat it and grants you the nutrition - the belly is just used for 'destination', sound & fluff purposes
So no worries there

Check [these 3 lines](https://github.com/VOREStation/VOREStation/pull/7085/files#diff-5ad528dbe1e40198cb73b80e3f5b6a6aR708-R710)

Issue #6867, opened 2020-03-15
Console frames lack directional sprites or any reasonable indication of direction


Heroman3003:
#### Brief description of the bug
Consoles have directions. Console frames do not. This makes orienting console frames in construction a pain since you can't even see the direction until you try rotating it with the verb. Not even on examine.

#### What you expected to happen
Have frame that reflects way console will be facing when its complete.

#### What actually happened
No difference between any of the directions, have to fully complete construction or try rotating manually to actually see the direction.

#### Steps to reproduce
- Make a frame
- Put it in position you want it in
- Realize that no matter how you move it its visuals are same and you can't tell what way its actually facing via normal people means.

#### Code Revision
- f9ed138687ee6ca130a7c11cd3601319e8a03781
 
#### Anything else you may wish to add:
- Even it displaying a direction in basic examine would be better. Although sprites actually reflecting the direction (and ideally, shape) of resulting console would be better.

Arokha:
By consoles you mean modular computers and preset computers and the like, right? Just to clarify.

Meghan-Rossi:
There's a similar problem with ordinary frames and gas heaters/coolers

Heroman3003:
I never tried constructing modular consoles specifically, so I don't know if that applies there. But It definitely does apply to regular consoles.

PR #7097, opened 2020-04-02
[MIRROR] Port computer and 'machine' frame icons from Bay

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/6857
--------------------
Kinda. I modified them. The console frames are fully directional so you can tell wtf way you're building them, and the machine frames, I added a little chevron to the front that points in their direction so you can tell.

Depends on https://github.com/PolarisSS13/Polaris/pull/6859 to look right

image


Meghan-Rossi:
This fixed https://github.com/VOREStation/VOREStation/issues/6867

PR #7102, opened 2020-04-02
Improves space edge-of-map behavior and add mapping tools

Arokha:
- Maps no longer move you when you're 7 away from the edge to avoid you seeing the edge. Instead, now the edge doesn't look like anything special, and you move when you touch it, because:
- Space turfs at the edge of maps stretch out their sprites into the void beyond the map borders to make it appear like space continues in that direction (only for the equivalent of 7 turfs worth of distance though).

For mappers:
- Fake stairs, to allow you to pretend to have multi-z but not, as they just teleport you between themselves. `/obj/structure/fake_stairs`. They behave like normal stairs and connect when you set `_stair_tag` to something identical on two stairs. Read the file for more info.
- You can map in `/turf/space/internal_edge` to 'force' space edges in the middle of your maps. This can allow you to divide up a single z-level for a 'fake multiz ship'. You can put the pieces very close together, even directly on the other side of a 2-thick internal_edge wall facing each other, and one side can't see the other, as it'll just look like space in that direction.


KasparoVy:
Great tools

code/game/atoms_movable.dm
```diff
@@ -462,30 +462,33 @@
...
 
 		if(ticker && istype(ticker.mode, /datum/game_mode/nuclear)) //only really care if the game mode is nuclear
 			var/datum/game_mode/nuclear/G = ticker.mode
 			G.check_nuke_disks()
 
-		spawn(0)
-			if(loc) loc.Entered(src)
+		var/turf/T = locate(new_x, new_y, new_z)
+		if(istype(T))
+			forceMove(T)
```
Leshana, line 491: Good change, makes more sense to call forceMove than this weird emulation of it

code/modules/overmap/spacetravel.dm
```diff
@@ -110,5 +110,5 @@ proc/overmap_spacetravel(var/turf/space/T, var/atom/movable/A)
 		var/obj/effect/overmap/visitable/sector/temporary/source = M
 		if (source.can_die())
 			testing("Caching [M] for future use")
-			source.forceMove(null)
+			source.loc = null
```
Leshana, line 113: We have `/atom/proc/moveToNullspace` so we probably should use it.

PR #7108, opened 2020-04-02
Voidsuit Additions

Novacat:
- Adds a Director's voidsuit for the Colony Director, replaces the old Colony Director armor. Captain's voidsuit has +10 more armor in all categories, but has +0.5 extra slowdown.
- Adds a Prototype Voidsuit. Currently has mercenary voidsuit stats and does not spawn in anywhere.
- Adds new cyclers for the new voidsuits.
- ODST Voidsuits can be worn by all species


Heroman3003:
Wouldn't it be called Colony Director voidsuit, considering Captain is no longer a position?

Novacat:
*Technically* it is called Director's voidsuit. But all the data files refer to it as Captain so that is what I call it in my head.

PR #7109, opened 2020-04-02
Conversion Kit Overhaul

KillianKirilenko:
This is basically a port of [this PR](https://github.com/Yawn-Wider/YWPolarisVore/pull/683), which massively reworks fluff conversion kits so they can be safely used for a wider range of items.

To summarize for those who don't want to click through;
- Kits can be used to revert items back into their original format
- Various forensic evidence persists through item conversion
- Items can't be converted if they're bloody, contaminated, or breached (breach check can be overridden to permit repairs via kits)
- Items that have stuff loaded or attached can't be converted unless an override is enabled on the kit (see below), to avoid deleting things (e.g. ammo, suit helmets, sec badges)
- Ammunition/cells are deleted from guns spawned by conversion, ties into the above to avoid ammo duplication (gun can't be converted when loaded: empty gun, convert, load new gun with old mag)
- Kits have a bunch of flags that do things like prevent non-owners from using them, disable reverting items to their original type, allow contents to be transferred when converting, what conversion costs, whether kits self-delete when empty, and so on

I've given this a fair bit of testing but I'm sure I've missed a couple of edge-cases, overlooked some stuff, and done things messily. Constructive feedback would be appreciated.

### DIFF
--- a/code/modules/vore/fluffstuff/custom_items_vr.dm
+++ b/code/modules/vore/fluffstuff/custom_items_vr.dm
@@ -44,7 +44,7 @@
 	var/to_suit_cost = -2
 	
 	var/owner_ckey = null		//ckey of the kit owner as a string
-	var/skip_contents = FALSE	//can we skip the contents check? we generally shouldn't, but this is necessary for rigs/coats with hoods/etc.
+	var/skip_content_check = FALSE	//can we skip the contents check? we generally shouldn't, but this is necessary for rigs/coats with hoods/etc.
 	var/transfer_contents = FALSE	//should we transfer the contents across before deleting? we generally shouldn't, esp. in the case of rigs/coats with hoods/etc. note this does nothing if skip is FALSE.
 	var/can_repair = FALSE		//can we be used to repair damaged voidsuits when converting them?
 	var/can_revert = TRUE		//can we revert items, or is it a one-way trip?
@@ -55,13 +55,14 @@
 	var/cost
 	var/to_type
 	var/keycheck
+	
+	if(isturf(O)) //silently fail if you click on a turf. shouldn't work anyway because turfs aren't objects but if I don't do this it spits runtimes. 
+		return
 	if(istype(O,/obj/item/clothing/suit/space/void/) && !can_repair) //check if we're a voidsuit and if we're allowed to repair
 		var/obj/item/clothing/suit/space/void/SS = O
 		if(LAZYLEN(SS.breaches))
 			to_chat(user, "<span class='warning'>You should probably repair that before you start tinkering with it.</span>")
 			return
-	if(isturf(O)) //silently fail if you click on a turf. shouldn't work anyway because turfs aren't objects but if I don't do this it spits runtimes. 
-		return
 	if(O.blood_DNA || O.contaminated) //check if we're bloody or gooey or whatever, so modkits can't be used to hide crimes easily.
 		to_chat(user, "<span class='warning'>You should probably clean that up before you start tinkering with it.</span>")
 		return
@@ -88,7 +89,7 @@
 	if(!isturf(O.loc))
 		to_chat(user, "<span class='warning'>You need to put \the [O] on the ground, a table, or other worksurface before modifying it.</span>")
 		return
-	if(!skip_contents && O.contents.len) //check if we're loaded/modified, in the event of gun/suit kits, to avoid purging stuff like ammo, badges, armbands, or suit helmets
+	if(!skip_content_check && O.contents.len) //check if we're loaded/modified, in the event of gun/suit kits, to avoid purging stuff like ammo, badges, armbands, or suit helmets
 		to_chat(user, "<span class='warning'>You should probably remove any attached items or loaded ammunition before trying to modify that!</span>")
 		return
 	if(cost > parts)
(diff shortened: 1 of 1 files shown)


code/modules/vore/fluffstuff/custom_items_vr.dm
```diff
@@ -36,34 +36,100 @@
...
+			N:cell_type = O:cell_type
+	else	//nuke any ammo it'd normally spawn with, if it's a gun, to prevent ammo duplication. we have to do this just for guns, not immediately under the else, or it breaks default attachments like hoods and suit storage
+		if(istype(N,/obj/item/weapon/gun/projectile/))
+			N:contents = list()
+			N:magazine_type = null
+			N:ammo_magazine = null
+		if(istype(N,/obj/item/weapon/gun/energy/))
+			N:contents = list()
+			N:cell_type = null
+
```
Arokha, line 129: Stuff like this might be worth having a `obj/item/proc/customize_item(var/obj/item/new_item)` so you can just have each object do what it wants rather than trying to handle the special cases all here.

KillianKirilenko: Maybe? I figure the way kits are set up now covers all the major bases though. You basically have to set the kit up (or varedit it) in a specifically unfriendly way in order to break it. I think.

Honestly I just kinda bash things together until they work, and then I'm pleasantly surprised when they do.

The whole issue essentially stems from the fact YW doesn't allow people to spawn with custom weapons/permits. That means modifying existing station equipment and _that_ means putting your fluff weapon back into storage when you leave the round where other people can get their grubby hands all over it, which some people might not be comfortable with. I set out to simply permit reverting changes initially, but then I realized that you could do stupid things with them if you were careless (like accidentally delete items attached to voidsuits or ammo loaded into guns), and it also basically let you do a bunch of other abusive stuff like duping ammo (if guns spawn loaded, and 99% of them do) or trivially removing forensic evidence of crimes if you were suitably conniving... so here we are.

code/modules/vore/fluffstuff/custom_items_vr.dm
```diff
@@ -36,34 +36,100 @@
...
 /obj/item/device/modkit_conversion/afterattack(obj/O, mob/user as mob)
-	var/flag
+	var/cost
 	var/to_type
-	if(istype(O,from_helmet))
-		flag = 1
+	var/keycheck
+	if(istype(O,/obj/item/clothing/suit/space/void/) && !can_repair) //check if we're a voidsuit and if we're allowed to repair
+		if(O:breaches.len)	//this has to be underneath the istype or it'll spit a runtime if used on non-voidsuits
+			to_chat(user, "<span class='warning'>You should probably repair that before you start tinkering with it.</span>")
```
Leshana, line 60: Avoid `:` operator by casting to the type.
```suggestion
		var/obj/item/clothing/suit/space/void/SS = O
		if(LAZYLEN(SS.breaches))
			to_chat(user, "<span class='warning'>You should probably repair that before you start tinkering with it.</span>")
```

code/modules/vore/fluffstuff/custom_items_vr.dm
```diff
@@ -36,34 +36,100 @@
...
+	var/obj/N = new to_type(O.loc)
+	user.visible_message("<span class='notice'>[user] opens \the [src] and modifies \the [O] into \the [N].</span>","<span class='notice'>You open \the [src] and modify \the [O] into \the [N].</span>")
+	
+	//crude, but transfer prints and fibers to avoid forensics abuse, same as the bloody/gooey check above
+	N.fingerprints = O.fingerprints
+	N.fingerprintshidden = O.fingerprintshidden
+	N.fingerprintslast = O.fingerprintslast
+	N.suit_fibers = O.suit_fibers
+	
+	//I don't like using the lookdown checks here but fuck it, it works. the istypes are essential though, or else it runtimes and won't get to the qdel for the source item, allowing item duping
```
Leshana, line 112: Don't use the `:` operator.  Safer to cast to a variable of appropriate type instead like the example given above.

KillianKirilenko: Okay, this is !fun! - partial success in that the casting has suppressed errors, but I must be wiring something up wrong- the behavior for skip+transfer vs nuke is unpredictable.

I'm running on entirely too little sleep and far too much fatigue, so I'm going to push my progress so far. I'm probably missing something trivial.

code/modules/vore/fluffstuff/custom_items_vr.dm
```diff
@@ -114,24 +115,44 @@
 	if(skip_contents && transfer_contents)
```
Arokha, line 115: Can you rename one of these? It's not intuitive as to what they do given the context of...
"Wait so you're going to skip the contents AND transfer the contents?"

code/modules/vore/fluffstuff/custom_items_vr.dm
```diff
@@ -36,36 +36,123 @@
...
-	if(istype(O,from_helmet))
-		flag = 1
+	var/keycheck
+	if(istype(O,/obj/item/clothing/suit/space/void/) && !can_repair) //check if we're a voidsuit and if we're allowed to repair
+		var/obj/item/clothing/suit/space/void/SS = O
+		if(LAZYLEN(SS.breaches))
+			to_chat(user, "<span class='warning'>You should probably repair that before you start tinkering with it.</span>")
+			return
+	if(isturf(O)) //silently fail if you click on a turf. shouldn't work anyway because turfs aren't objects but if I don't do this it spits runtimes. 
+		return
```
Arokha, line 64: Should do above the spacesuit check above. and perhaps just
`if(!istype(O)) return`

code/modules/vore/fluffstuff/custom_items_vr.dm
KillianKirilenko: Yeah that's been giving me a headache lately as well, I just hadn't thought of anything better to swap it to. `skip_content_check` ought to work better.

code/modules/vore/fluffstuff/custom_items_vr.dm
KillianKirilenko: Gonna keep the newline for the sake of the comment and matching the overall formatting of the rest of the proc, but otherwise done.

KillianKirilenko:
Finally figured out what the issue with power cell replication was (`power_supply` var) and threw in some `update_icon()`s so that guns will have the correct ammo/etc. states after conversion, whether loaded conversion was permitted or not.

PR #7112, opened 2020-04-02
Adds Mechanicus Techpriest Robes

Novacat:
https://www.youtube.com/watch?v=ztzq05IzYds

Adminbus Only. Ported from CitRP.


Arokha:
For the Omnissiah

PR #7116, opened 2020-04-03
Re-enables Slime Vore reaction, adjusts blocked lists for slimespawn reactions

Heroman3003:
Re-enables slime vore reaction, which spawns a bunch of random simplemobs.

Heavily readjusts the blocked mobs list though. Now the mobs that cannot be spawned are:

- Technical mobs that only exist to pass on traits to subtypes
- Technical mobs that are things coded as mobs but don't really function as such
- Unique mobs, primarily named pets
- Mobs exclusive to events and/or lacking any sort of AI and designed to be player-controlled (such as shadekin, cult constructs, etc)
Also, the mimic and alien queen can now be spawned by it. Mimic was forbidden because in the old  mob system mob itself was technical, but not anymore, and alien queen... well, no idea there. Its not like it spawned eggs or anything, and animal Empress and Princess COULD spawn, which are just different looking queens anyway.

Also adjusts the Slime Resources reaction and makes its spawn list more readable, with some adjustments:

- Can no longer spawn Particle Smasher-exclusive sheets
- Can spawn animal hides (why were they forbidden from it in the first place?)
- Rather than mentioning every /cyborg sheet subtype, entire subtype is blanket-forbidden

### DIFF
--- a/code/modules/reagents/Chemistry-Recipes_vr.dm
+++ b/code/modules/reagents/Chemistry-Recipes_vr.dm
@@ -207,12 +207,18 @@
 
 		var/blocked = list(
 							/obj/item/stack/material,					//Technical stacks
-							/obj/item/stack/material/animalhide,
+							/obj/item/stack/material/hairlesshide,		//Useless leather production steps
+							/obj/item/stack/material/wetleather,
+							/obj/item/stack/material/algae/ten)			//Why is this one even a separate thing
+		blocked += typesof(/obj/item/stack/material/cyborg)				//Borg matter synths, should only exist in borgs
+		blocked += typesof(/obj/item/stack/material/animalhide)			//Hides which are only used for leather production anyway
+
+		var/rare_types = list(
 							/obj/item/stack/material/morphium,			//Complex materials requiring Particle Smasher to create
 							/obj/item/stack/material/morphium/hull,
 							/obj/item/stack/material/valhollide,
 							/obj/item/stack/material/supermatter)
-		blocked += typesof(/obj/item/stack/material/cyborg)				//Borg matter synths, should only exist in borgs
+
 		var/list/material = typesof(/obj/item/stack/material) - blocked
 
 		playsound(get_turf(holder.my_atom), 'sound/effects/phasein.ogg', 100, 1)
@@ -223,6 +229,8 @@
 */
 		var/spawn_amount = rand(1,50)
 		var/chosen = pick(material)
+		if(chosen in rare_types)
+			spawn_amount = rand(1,15)
 		var/obj/item/stack/material/C = new chosen
 		C.amount = spawn_amount
 		C.loc = get_turf(holder.my_atom)
--- a/code/modules/reagents/Chemistry-Recipes_vr.dm
+++ b/code/modules/reagents/Chemistry-Recipes_vr.dm
@@ -205,7 +205,14 @@
 				O.show_message(text("<span class='warning'>The slime core fizzles disappointingly.</span>"), 1)
 			return
 
-		var/blocked = list(/obj/item/stack/material, /obj/item/stack/material/cyborg, /obj/item/stack/material/cyborg/plastic, /obj/item/stack/material/cyborg/plasteel, /obj/item/stack/material/cyborg/glass/reinforced, /obj/item/stack/material/cyborg/wood, /obj/item/stack/material/animalhide/human, /obj/item/stack/material/animalhide/corgi, /obj/item/stack/material/animalhide/cat, /obj/item/stack/material/animalhide/monkey, /obj/item/stack/material/animalhide/lizard , /obj/item/stack/material/animalhide/xeno, /obj/item/stack/material/cyborg, /obj/item/stack/material/cyborg/glass/reinforced)
+		var/blocked = list(
+							/obj/item/stack/material,					//Technical stacks
+							/obj/item/stack/material/animalhide,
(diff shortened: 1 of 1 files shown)


Screemonster:
I still don't get why the slime mob spawn list is a blacklist instead of a whitelist.

Heroman3003:
Mostly because the whitelist would be like 4 times as long if we were to just flip it from being a blacklist to being a whitelist. Unless you do not want majority of mobs to show up.

Screemonster:
Fair enough.

code/modules/reagents/Chemistry-Recipes_vr.dm
```diff
@@ -205,7 +205,14 @@
...
 			return
 
-		var/blocked = list(/obj/item/stack/material, /obj/item/stack/material/cyborg, /obj/item/stack/material/cyborg/plastic, /obj/item/stack/material/cyborg/plasteel, /obj/item/stack/material/cyborg/glass/reinforced, /obj/item/stack/material/cyborg/wood, /obj/item/stack/material/animalhide/human, /obj/item/stack/material/animalhide/corgi, /obj/item/stack/material/animalhide/cat, /obj/item/stack/material/animalhide/monkey, /obj/item/stack/material/animalhide/lizard , /obj/item/stack/material/animalhide/xeno, /obj/item/stack/material/cyborg, /obj/item/stack/material/cyborg/glass/reinforced)
+		var/blocked = list(
+							/obj/item/stack/material,					//Technical stacks
+							/obj/item/stack/material/animalhide,
+							/obj/item/stack/material/morphium,			//Complex materials requiring Particle Smasher to create
+							/obj/item/stack/material/morphium/hull,
+							/obj/item/stack/material/valhollide,
+							/obj/item/stack/material/supermatter)
```
Novacat, line 214: Staff had a discussion and it was decided that particle smasher materials should be kept as obtainable via xenobio.

However, an acceptable alternative would be to set it up so that particle smasher materials outputs a lower amount of material, instead of 1-50 sheets per reaction, perhaps 1-15 sheets instead.

code/modules/reagents/Chemistry-Recipes_vr.dm
```diff
@@ -205,7 +205,14 @@
 				O.show_message(text("<span class='warning'>The slime core fizzles disappointingly.</span>"), 1)
 			return
 
-		var/blocked = list(/obj/item/stack/material, /obj/item/stack/material/cyborg, /obj/item/stack/material/cyborg/plastic, /obj/item/stack/material/cyborg/plasteel, /obj/item/stack/material/cyborg/glass/reinforced, /obj/item/stack/material/cyborg/wood, /obj/item/stack/material/animalhide/human, /obj/item/stack/material/animalhide/corgi, /obj/item/stack/material/animalhide/cat, /obj/item/stack/material/animalhide/monkey, /obj/item/stack/material/animalhide/lizard , /obj/item/stack/material/animalhide/xeno, /obj/item/stack/material/cyborg, /obj/item/stack/material/cyborg/glass/reinforced)
+		var/blocked = list(
+							/obj/item/stack/material,					//Technical stacks
+							/obj/item/stack/material/animalhide,
```
Novacat, line 210: This should probably be a typesof, since normally most animal hides are blocked anyway.

Screemonster:
Regarding the "particle-smasher exclusive sheets" - technically they're not particle smasher exclusive. they're particle smasher and xenobio. Changing this is a balance change rather than a bugfix.

Unless xenobio is a vastly easier method of obtaining them in quantity than using the particle smasher, I don't see the harm of something in the game having multiple methods of obtaining it.

code/modules/reagents/Chemistry-Recipes_vr.dm
Heroman3003: They were blocked but change to not make them blocked other than base technical type was itnentional, unless there was some specific purpose to the animalhides not being available?

Novacat: I am guessing it was not intended for animal hides to be produced via the slime reaction. Why? I cannot tell you, but probably should keep it that way for the time being, since adding a bunch of animal hides also substantially effects balance.

Heroman3003:
As a side note, slime reactions aren't actually at all 'xenobio-exclusive' (except for Slime Vore reaction) considering that slime jelly is producable with just a good surplus of phoron, which is easily obtainable in large quantities when cargo doesn't sell it all. And in xenobio itself, its not a reaction that requires much work at all beyond just grinding to get enough for rolls, since slime jelly is only obtainable from most basic grey slimes anyway.

Heroman3003:
On another note, those reactions are Vorestation-exclusive and existed in way they are long before materials with such complex obtaining method as the Particle Smasher materials were even heard about. And continued to spawn 'all types' without any regard for rarity even after the addition. Which is why the removal was proposed.

nerdass:
i say leave the reactions, rounds don't have cargo on often enough that at the very least it leaves a route available for players who want to roll the dice to be able to do the thing without having to beg over the radio for someone to bring in crates from cargo, and even then, slime reactions are hardly a reliable source.

Screemonster:
Just because we didn't explicitly say to keep them when the new materials were added doesn't mean they should be removed. Unless they're causing problems I see no reason to stop people having fun.

Natje0:
Please separate your bugfix and content nerfs into separate PRs.

PR #7117, opened 2020-04-03
Add Looking glass holodeck

Arokha:
Codes and adds a "Looking Glass"-style holodeck just south of the existing holodeck:
ezgif-7-339d121cece6
ezgif-7-52fcc6b57a8d

Map-wise, it's mapped in like thus:
dreammaker_2020-04-03_00-03-05

You put /turf/simulated/floor/looking_glass around the edges with arrows that point outwards, then /turf/simulated/floor/looking_glass/center in the middle of the walkable area (arrows are not walkable, alas). /turf/simulated/floor/looking_glass/optional is for 'toggleable' arrows that act like center when full immersion is disabled, and become 'full' when immersion is enabled (depicted in the second gif above).

In the center of the area, place a /obj/effect/landmark/looking_glass (the X in the image). This is where your background will be centered.

The area has a lg_id on it, the console needs a matching lg_id. That's it! The area does most of the work, coordinating between the console, the turfs, and the image holder landmark.


Heroman3003:
This is good.

Arokha:
https://github.com/PolarisSS13/Polaris/pull/6958

PR #7118, opened 2020-04-03
Ports Baystation Advanced Shield Generators

Leshana:
Port of Baystation12/Baystation12#14401

### Changes
- Creates new advanced shield generators, designed to replace old hull and bubble shield generators.
- Upgrades the floor mounted and handheld shield diffusers.
- Makes underfloor shield diffusers actually constructable.
- Handheld diffusers orderable via uplink.
- Removes supply packs and research datums  for the old generators, but leaves their code in place for maps that still use them.
- Integrates with the meteor and electrical storm events.
- Integrates with mob AI (they know how to attack it)
- Adds setter procs for density and invisibility vars.

### About the shield generators
- Single generator combines both bubble and hull shields in one, you can choose between them.
- Now with NanoUI and _wires_.
- Axes the renwick concept. Instead, shields are purely based on the amount of energy you can push into the shield. Shield generators are built with a SMES coil, and use it to store energy. Damage to the field uses up this energy, as well as upkeep.
- Shield generator on it's own has only little upkeep, but is useless. The proper use is given by shield modes, which can be enabled in the generator's UI. Each shield mode tweaks the field by adding a new functionality (for example: Hyperkinetic mode blocks bullets, meteors and various other objects).
- Field upkeep increases on a per-mode basis. Each mode has a set multiplier. 
- If the generator's stored energy is lower than 35% there is a risk of temporary field breach when the field takes damage. This probability grows as the stored energy drops. If a hit uses up all remaining energy the field collapses completely and goes into recovery mode. Once stored energy gets above 25% the field restarts again. Alternatively, it may be restarted manually before this point.

### Future Improvements
- Bring back Renwicks, but just cosmetically 
- Create stronger gameplay reason to install shield diffusers. They were actually useful for everyone with the old shields, but serve next to no "legit" purpose now (still useful for antags of course)
- Cooler shield icons and effects.


Screemonster:
Renwicks were annoying and confusing before, and I like the idea of multimodes on shields to allow the slow crew (that may or may not have knives) to penetrate the shield

TheFurryFeline:
Ah, I'm reminded once again that shield diffusers still need to be added to all the EVAs.... Also, they show up around platings on the surface leading to the mountain, solars, and wilderness, so ya have to add the diffusers to the platings as well.

PR #7129, opened 2020-04-04
Abstract Eating of Ore to Fix Unintentional Prompt

KasparoVy:
Using the verb with nothing in your hand brings up a prompt to select an object due to the argument in the verb.

This resolves the issue, instead just showing the "nothing in your hand" message as intended.

Caused this myself in #7085

The issue:
virgo ss13 eat-ore bug when nothing in hands 20200403

### DIFF
--- a/code/modules/mining/ore.dm
+++ b/code/modules/mining/ore.dm
@@ -128,7 +128,7 @@
 	if(ishuman(M))
 		var/mob/living/carbon/human/H = M
 		if(H.species.eat_ore == 1)
-			H.eat_ore(src)
+			H.handle_eat_ore(src)
 			return
 	..()
 //VOREStation Add End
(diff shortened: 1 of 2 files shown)


Heroman3003:
A small suggestion for trait as a whole, mostly unrelated to PR: could it be made more universal in terms of use so it could be through right verbs given and/or admin magic added to mobs that aren't just humanmobs? /attack proc on ore prevents that as of now.

KasparoVy:
I think if you grant the eat_ore() verb to any living mob, so long as they have a vorebelly they can activate the verb -- however it will get hung up on the mob not having hands.

I might be able to modify it to check if the mob has hands period, or if they're nonhumanoid and prompt to ask what they'd like to eat in a list or pick randomly.

KasparoVy:
@Heroman3003 What's your intention for other mobs to be able to do? Be fed/eat ore themselves?

KasparoVy:
Cheers, been a while since I wrote a verb and I forgot this quirk

PR #7134, opened 2020-04-05
Removes Xenoarch access on Mining Shuttle

Novacat:
On request from Admins

### DIFF
--- a/code/modules/shuttles/shuttles_vr.dm
+++ b/code/modules/shuttles/shuttles_vr.dm
@@ -10,7 +10,7 @@
 
 /obj/machinery/computer/shuttle_control/belter
 	name = "belter control console"
-	req_one_access = list(access_mining, access_xenoarch, access_medical_equip) //Allows xenoarch, miners AND doctors to use it.
+	req_one_access = list(access_mining, access_medical_equip) //Allows xenoarch, miners AND doctors to use it.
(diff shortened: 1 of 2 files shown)


Novacat:
I am not too worried about the doors leading into the area. Just the shuttle itself.

TheFurryFeline:
If xenoarch is to lose access as a whole, the doors should also have xenoarch access removed.

KasparoVy:
Cheers.

code/modules/shuttles/shuttles_vr.dm
```diff
@@ -10,7 +10,7 @@
 
 /obj/machinery/computer/shuttle_control/belter
 	name = "belter control console"
-	req_one_access = list(access_mining, access_xenoarch, access_medical_equip) //Allows xenoarch, miners AND doctors to use it.
+	req_one_access = list(access_mining, access_medical_equip) //Allows xenoarch, miners AND doctors to use it.
```
KasparoVy, line 13: Comment wasn't changed, eh?

TheFurryFeline: Good catch.

Heroman3003:
If Xenoarch are not expected to do belt mining anymore, belt miner crate might need a change, because I don't remember exactly, but it also might have xenoarch access.

PR #7145, opened 2020-04-05
Fix 2 minor issues

Arokha:
- Leave remains vr can't have it's items mapped in or it runtimes
- Xenoarch controller convinces byond it's an infinite loop on large maps (goes to Polaris)


TheFurryFeline:
Tag of Vore-Related could probably also be added because of the bugfix for the mapped-in remains runtime.

PR #7146, opened 2020-04-05
attempts to port syringe icons

AmayaWah:
Ports the sprites of syringes from Baystation to Virgo!


Heroman3003:
Isn't our current syringe icon code designed to like. Rotate to be verical when in the storages so that they don't overlap due to being smallest itemsize? Would that even work?

Wickedtemp:
What are bays icons

Heroman3003:
@Wickedtemp if you scroll down you can see Checks category at the bottom of PR. If you check Diffs check's details, you can see the icons comparison.

Leshana:
It appears to me that this adds a new dmi file but not any code to use it, is that right?

Wickedtemp:
I kinda prefer the ones we currently have over these. The current ones actually have syringe caps and just look nicer in my opinion.

AmayaWah:
> It appears to me that this adds a new dmi file but not any code to use it, is that right?

Yes! I have no idea how to code and Aro said that I can just upload icons and it's ok.

Wickedtemp:
Also how do these look in a syringe case?

PR #7150, opened 2020-04-06
Refactor the gameticker controller into SSticker

Leshana:
- Convert the gameticker controller into a subsystem: SSticker
  - Instead of independently spawn'd while/sleep loops, it stores its state and lets the MC fire it.
  - Convert relative path indentation procs to absolute path as per modern style standards.
  - Break apart the inner loops into separate procs so you can see what is actually going on.
  - Cleanup auto-voting so it actually obeys configurations better.
  - It now shows up in MC tab, along with stats about what phase it is in.
- Revamp Start Now admin verb to actually wait until init is done before starting.
- Lobby stat panel to show time remaining on lobby votes and if the server is done initializing.
- Remove the obsolete "processSchedler" controller that wrapped the gameticker.

#### Github PR Diff Issue
**NOTE**: Github does not show the diff very well by default!  It does not detect the rename of `code/game/gamemodes/gameticker.dm` -> `code/controllers/subsystems/ticker.dm` and thus shows them as deletion and addition of the entire files.  Only the startup/shutdown looping functionality was re-written, the support functions below are left alone, simply converted from relative pathing to absolute pathing.


code/controllers/subsystems/ticker.dm
```diff
@@ -0,0 +1,560 @@
...
+
+/datum/controller/subsystem/ticker/proc/HasRoundStarted()
+	return (current_state >= GAME_STATE_PLAYING)
+
+// ------------------------------------------------------------------------
+// HELPER PROCS!
+// ------------------------------------------------------------------------
+
+//Plus it provides an easy way to make cinematics for other events. Just use this as a template :)
+/datum/controller/subsystem/ticker/proc/station_explosion_cinematic(var/station_missed=0, var/override = null)
```
Arokha, line 297: \o/

code/controllers/subsystems/ticker.dm
```diff
@@ -0,0 +1,560 @@
...
+			else if(!player.mind.assigned_role)
+				continue
+			else
+				//VOREStation Edit Start
+				var/mob/living/carbon/human/new_char = player.create_character()
+				if(new_char)
+					qdel(player)
+				if(istype(new_char) && !(new_char.mind.assigned_role=="Cyborg"))
+					data_core.manifest_inject(new_char)
+				//VOREStation Edit End
```
Arokha, line 411: How is this a vorestation edit?

code/modules/admin/admin.dm
```diff
@@ -871,21 +871,26 @@ var/datum/announcement/minor/admin_min_announcer = new
 
 /datum/admins/proc/startnow()
 	set category = "Server"
-	set desc="Start the round RIGHT NOW"
+	set desc="Start the round ASAP"
```
Arokha, line 874: \o/

code/controllers/subsystems/ticker.dm
Leshana: No idea, this proc was not changed as part of this PR.  Only touch to these lines was to un-indent them by one level.

Leshana:
Rebased to fix merge conflicts.

PR #7158, opened 2020-04-06
Suit Cyclers: Faster, Smarter, Better

KillianKirilenko:
So I'm laying in bed, trying to get some rest... and I think to myself-

*"Hang on Killian, isn't that suit cycler code kind of a bloated piece of shit?"*
*"It sure is, brain, but how can we make it better?"*
*"Okay, like, we can totally get the initial name/desc/whatever from the base object type, right? Like the default values?"*
*"...probably, why?"*
*"Think about it: it'd be so much more efficient. We just specify the helmet and suit to pull from, and then the cycler will always give accurate names, descriptions, and states. It never needs to be kept 'in sync' with the values specified in the suits and it's trivial to add new entries; coders only need to add the new cycling option and set two paths."*

A little digging and diving later, and here we are; the apply_paintjob proc has been halved in size.

There's a couple of finishing polish touches it needs (see comments right at the end), but it's otherwise pretty much fully functional. Once it's done I can pass it up to Polaris; this is basically just here for review/comments.

### DIFF
--- a/code/game/machinery/suit_storage_unit.dm
+++ b/code/game/machinery/suit_storage_unit.dm
@@ -1080,16 +1080,20 @@
 	
 	//look at this! isn't it beautiful? -KK
 	if(helmet)
+		var/obj/item/clothing/H = new parent_helmet
 		helmet.name = "refitted [initial(parent_helmet.name)]"
-		helmet.desc = "refitted [initial(parent_helmet.desc)]"
+		helmet.desc = initial(parent_helmet.desc)
 		helmet.icon_state = initial(parent_helmet.icon_state)
 		helmet.item_state = initial(parent_helmet.item_state)
 		helmet.light_overlay = initial(parent_helmet.light_overlay)
-		//TODO: figure out how to grab the item state slots too, they're a bit harder
+		helmet.item_state_slots = H.item_state_slots
+		qdel(H)
 
 	if(suit)
+		var/obj/item/clothing/S = new parent_suit
 		suit.name = "refitted [initial(parent_suit.name)]"
-		suit.desc = "refitted [initial(parent_suit.desc)]"
+		suit.desc = initial(parent_suit.desc)
 		suit.icon_state = initial(parent_suit.icon_state)
 		suit.item_state = initial(parent_suit.item_state)
-		//TODO: figure out how to grab the item state slots for these as well
+		suit.item_state_slots = S.item_state_slots
+		qdel(S)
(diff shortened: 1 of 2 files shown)


KillianKirilenko:
Polaris version [here](https://github.com/PolarisSS13/Polaris/pull/6971). If/when it's accepted upstream I'll flag this WIP PR as ready for review so that the CD & Proto options can be merged in.

KillianKirilenko:
Moving this to Ready. Polaris is moving a little slower than I anticipated.

Leshana:
Fetching from the actual origin items is *much* better than having all those values duplicated!
I think we need to handle the item_state_slots values too tho, or it won't look right in hand.

KillianKirilenko:
It was very much a "eureka" sort of moment when my hunch that initial values could indeed be referenced... but there's a catch; `initial()` can't pull the contents of lists - I tried that already and it just copies blank entries, meaning the items turn invisible in the onmob inhands.

The way I see it there's two workarounds;
1. The ugly, bloaty way; each suit/helmet pair will need the values reassigning in the proc, bloating it back up. This could be simplified *somewhat* with a 'style' system, but that wouldn't save much space.
2. The messy, sneaky way; spawn in a copy of the helmet and/or suit, pull the values from *that copy's* lists, and then immediately delete it.

If anyone has an idea for third options that might work better I'm all ears.

Arokha:
If you do #2, I suggest keeping a cache in a static list.
```
/suit_cycler
var/static/list/inhand_cache = list()

/that_proc
    var/H = new helmet
    inhand_cache[H.type] = item_state_slots.Copy()
```

KillianKirilenko:
I'll do some tinkering with the upstream version and see where that gets me, since I need to make some changes there anyway. Worst case scenario people might have to settle for inaccurate inhands.

Arokha:
> no list caching needed

I mean, yeah it would work fine without caching, but it means it needs new New() an item every time even if it's already New()'d one before for the same item, which is wasteful.

Leshana:
> Solution 2 seems to work fine, no list caching needed.
These machines are used rarely enough that caching might not be super important.

KillianKirilenko:
Be sure to sync this one in real quick after #7307 to get the Director & Proto variants over to the new system.

Novacat:
Included this into the Mirror

KillianKirilenko:
Good idea, that'll probably be a bit cleaner.

PR #7159, opened 2020-04-06
2 fixes and Teshari maws

Arokha:
- Fixes a problem with zlevels having unique skyboxes
- Fixes an infinite loop in map_effect/interval logic
- VS Only: Adds some additional programs to the looking glass holodeck when emagged


code/controllers/subsystems/skybox.dm
```diff
@@ -68,11 +68,6 @@ SUBSYSTEM_DEF(skybox)
 /datum/controller/subsystem/skybox/proc/get_skybox(z)
 	if(!skybox_cache["[z]"])
 		skybox_cache["[z]"] = generate_skybox(z)
-		if(global.using_map.use_overmap)
```
Leshana, line 71: Ah well. I guess we can't use this optimization because it ignores the possibility of using_map itself having specified different backgrounds.

code/modules/looking_glass/lg_console.dm
```diff
@@ -41,6 +41,9 @@
 		secret_programs["Flesh"] = image(icon = 'icons/skybox/skybox_vr.dmi', icon_state = "flesh")
 		secret_programs["Synth Int"] = image(icon = 'icons/skybox/skybox_vr.dmi', icon_state = "synthinsides")
 		secret_programs["Synth Int 2"] = image(icon = 'icons/skybox/skybox_vr.dmi', icon_state = "synthinsides_active")
+		secret_programs["Two Teshari"] = image(icon = 'icons/skybox/skybox_vr.dmi', icon_state = "doubletesh")
```
KasparoVy, line 44: The legendary `d o u b l e   t e s h`

Arokha: |3

PR #7177, opened 2020-04-09
Refactor the Open Space controller into MC Subsystem

Leshana:
### Refactor the Open Space controller into MC Subsystem
- Basically just refactor doWork() into fire() with the capablity to return in the middle with MC_TICK_CHECK.
- Move some vars around to be more organized.
- Only functional change is that when a turf is queued twice in one cycle, the second queue moves it to the end of the list instead of it being in the queue twice

### Fix openspace redering of under-objects that have no main icon (only over/underlays)
In particular stops manifolds from getting the openspace "map" icon state as its main icon instead of null.
before
after


code/controllers/subsystems/open_space.dm
```diff
@@ -0,0 +1,123 @@
...
+		var/turf/above = GetAbove(T)
+		if(above && isopenspace(above))
+			add_turf(above, recursive)
+
+// Queue the initial updates of open space turfs when the game starts. This will lag!
+/datum/controller/subsystem/open_space/proc/initialize_open_space()
+	// Do initial setup from bottom to top.
+	for(var/zlevel = 1 to world.maxz)
+		for(var/turf/simulated/open/T in block(locate(1, 1, zlevel), locate(world.maxx, world.maxy, zlevel)))
+			add_turf(T)
```
Arokha, line 75: Why not just `var/turf/simulated/open/T in world`?

Leshana: So that we process them in "bottom to top" order.  This way double+ stacked open space has the one underneath already built before its turn.

PR #7194, opened 2020-04-10
Adding the KMine: Mining dogborg

Holdelta:
(Please thank @AegisOA for the name)

Adding a Mining dogborg to the list of dogborgs since I found it odd that Cargo was the only department without one.

Functionally the same as a normal mining borg, save for the Supply Satchel belly module which contain the reagents:
- Glucose
- inaprovaline
- tricordrazine

This is mostly to help stabilize Miners in the field, and provide nutrition for crewmates who would rather be fed in the luxury of a stationbound.

EDIT: I have remove protein and nutriment from the injects and replaced them with glucose.


Heroman3003:
Both Nutriment and Protein in that belly is a reaallly bad idea. Those do not play well when injected (unless its xenochimera or promethean).

Holdelta:
> 
> 
> Both Nutriment and Protein in that belly is a reaallly bad idea. Those do not play well when injected (unless its xenochimera or promethean).

Am I better off removing both, or keeping one?  Was kind of playing with the concept that cargo can order food, so I added those in, but I can remove them if its too much.

AegisOA:
It would probably be better if you go with glucose, if you're going with the whole "lesser medbelly themed after Mining vendor autoinjectors" thing. Protein would be nice for xenochimera crew, but, people'll probably mistakenly inject it into regular crew if there isn't a warning (unless they already know it'll hurt them).

Holdelta:
I have removed nutriment and protein, and replaced them with glucose, this is also noted in the main message.

PR #7200, opened 2020-04-11
ITV Talon offmap spawn jobs

Arokha:
(Recreated on less messy branch)

First, other things it does:
- Updates consoles and telecomms to operate based on the idea that anything in the same overmap sector is "close enough" to be considered for display purposes (suit sensors, cameras, radios, atmos alarms, fire alarms, etc). So these will now operate on a 'overmap-sector-wide' scale. A couple of them, namely cameras and suit sensors, operate on the same overmap turf so that you can be in a ship in orbit and watch explorers below and monitor their health. I did it for those two things specifically because I think it's cool.
- Updates GPSes to do that too.
- Rewrite some job-related things to allow certain jobs to be isolated out of things like the manifest, accounts database, records consoles, etc so that they are not really 'visible' to station crew, but are also not considered antags and still get those things.
- Rewrites some spawning/despawning things to allow spawn/despawn announcements to be confined to sensible z-levels or even be on alternate radio channels.
- Makes the Pathfinder's bluespace radio connect directly to tether telecomms, not an 'omni-radio' that contacts every radio in the entire game

ITV Talon
-------------
Then also the actual thing that took the most time, the ITV Talon. There are 5 job slots that spawn on the Talon: *Talon Captain*, *Talon Doctor*, *Talon Engineer*, *Talon Guard*, *Talon Pilot*. I think those are probably self-explanatory and I don't need to tell you what each one is responsible for. The Talon is not an NT vessel and these are not necessarily NT employees. 

The Talon is a bit 'light' on access restrictions, in that there's only public access, or Talon-crew-restricted.

-------------
image

1. Port Engineering (air for circulation, fuel for port engines)
2. Stbd Engineering (tanks for scrubbers, fuel for stbd engines)
3. Ship's boat
4. Port Engineering Storage (tool vendors, toolboxes)
5. Stbd Engineering Storage (extra solar kit, other supplies, drones)
6. Brig/Guest rooms/Guard room
7. Workroom (autolathe, public modular computer preloaded, chargers)
8. Medical bay/OR

-------------

dreamseeker_2020-04-10_00-10-33

1. Port Solars SMES (connects to all solars on port side)
2. Stbd Solars SMES (obvious)
3. Secure storage (for loot etc)
4. Armory
5. Bridge (lower)

-------------

dreamseeker_2020-04-10_00-11-20

1. Bathroom
2. Cryo
3. Doctor's cabin
4. Lifeboat
5. Guard's cabin
6. Engineer's cabin
7. Shield generator/backup bridge
8. Pilot's cabin

-------------

dreamseeker_2020-04-10_00-11-32

1. Captain's cabin
2. Bar/cafe
3. Bridge (upper)


maps/tether/tether_jobs.dm
```diff
@@ -0,0 +1,184 @@
...
+	faction = "Station" //Required for SSjob to allow people to join as it
+	departments = list(DEPARTMENT_TALON)
+	total_positions = 1
+	spawn_positions = 1
+	selection_color = "#999999"
+	economic_modifier = 7
+	minimal_player_age = 14
+	pto_type = null
+	access = list(access_talon)
+	minimal_access = list(access_talon)
```
Heroman3003, line 27: Might make more sense for some things like minimal_player_age, access and pto_type to be moved to general talon job definition? Same for offmap_spawn and departments.

maps/tether/tether_jobs.dm
```diff
@@ -0,0 +1,184 @@
...
+	short_name = "Talon"
+	color = "#888888"
+	sorting_order = -2
+	assignable = FALSE
+	visible = FALSE
+
+/datum/job/talon/captain
+	title = "Talon Captain"
+	departments_managed = list(DEPARTMENT_TALON)
+	job_description = "The captain's job is to generate profit through trade or other means such as salvage or even privateering."
```
Heroman3003, line 14: Captain does not have 'supervisors' but I believe current code implies ALL jobs must have a supervisors string because of how welcome message is constructed. Could even be something jokey, similar to Mime/Clown's supervisors, but it has to have it in order for message not to botch itself.

maps/tether/tether_defines.dm
```diff
@@ -196,34 +201,6 @@
 		Z_LEVEL_PLAINS
 	)
 
-// Short range computers see only the six main levels, others can see the surrounding surface levels.
```
Heroman3003, line 199: How will this change affect surface z-levels of Virgo 3b that are not normally accessible from station, Underdark and Wilderness?

maps/tether/tether-02-surface2.dmm
```diff
@@ -20835,16 +20826,8 @@
...
-/obj/machinery/telecomms/relay/preset/tether/base_high,
-/turf/simulated/floor/tiled/dark{
-	nitrogen = 100;
-	oxygen = 0;
-	temperature = 80
-	},
-/area/tcommsat/chamber)
 "aNd" = (
-/obj/machinery/telecomms/broadcaster/preset_right,
+/obj/machinery/telecomms/broadcaster/preset_right/tether,
```
Heroman3003, line 20830: Will the changes in relays also affect the Underdark/Wilderness comms? Those two are not meant to be connected IIRC

Heroman3003:
As for overall design, I personally feel like cryo room is a little too small. It shouldn't be as big as Tether's of course, but it just looks too cramped, imo.

maps/tether/tether_jobs.dm
Arokha: I had that originally. Alas, the job system cannot deal with prototype types like that without a smol rewrite I don't want to do.

maps/tether/tether_defines.dm
Arokha: They will be accessible to suit sensors, cameras, radio, etc. Unless that's not desirable? I don't see why that'd be a bad thing, though. They already had telecomms because relays were pre-mapped in for all the different levels you mention, so one might think that suit sensors and such would work too.

maps/tether/tether-02-surface2.dmm
Arokha: They have a mapped in relay for the UD so I'm not sure about that recollection. Plains will be newly connected yes, but it's within walking distance of tether. I dunno why a radio couldn't go that far.

maps/tether/tether_defines.dm
Heroman3003: Wilderness is area south of station and it definitely has no relay. My mistake with Underdark, but I do know that its specifically NOT connected in terms of everything else (like suit sensors or GPS). Both Wilderness not having any connection and Underdark having no suitsensors seemed like balance decisions to make sure that risk of being in expedition areas isn't lowered even further than it already is by being on same planet and having weaker mob selection.

maps/tether/tether_jobs.dm
Arokha: Added

maps/tether/tether_defines.dm
Novacat: I would say that this change is probably fine. Especially since someone who does go out there and die will at least have a chance of medical being able to recover the body and such.

PR #7202, opened 2020-04-11
Make wierd glass cage into a actual boxing ring.

H0lySquirr3l:
image

It bugged me that the "boxing ring" on the holo suit used this weird tile glass cage instead of an actual boxing ring. So I made one! :D


code/game/objects/structures/fitness.dm
```diff
@@ -63,3 +63,83 @@
 		else
 			to_chat(user, "<span class='notice'>Against your previous judgement, perhaps working out is not for you.</span>")
 			being_used = 0
+
+/obj/structure/fitness/boxing_ropes
```
Novacat, line 67: This needs to be moved to a _vr file, or pushed up to Polaris

KasparoVy:
Looks like the essence of suplex

code/game/objects/structures/fitness.dm
H0lySquirr3l: I can do that. But how do I change the map merger? I should have had the most recent version as of this afternoon.

H0lySquirr3l:
Ok should be fixed...

PR #7207, opened 2020-04-11
Adds new ear options

Runa-Dacino:
Since in fiction, demons often possess pointy ears (and for my own needs), I have decided to enable our demonic degenerates to have pointy ears in addition to fancy horns!

Ram-like (caprine), backwards and forward pointing classic demon, oni and finally those large backswept dragon horns can now be picked alongside a pair of elven ears. 

Both the ears and the horns are separately colourable. Ears are first option, horns are second.

Tested on home server and works flawlessly.

https://cdn.discordapp.com/attachments/678731731498565684/698520386979233822/unknown.png

### DIFF
--- a/code/modules/vore/appearance/sprite_accessories_vr.dm
+++ b/code/modules/vore/appearance/sprite_accessories_vr.dm
@@ -504,8 +504,46 @@
     desc = ""
     icon_state = "pipears"
     ckeys_allowed = list("phoaly")
-
-
+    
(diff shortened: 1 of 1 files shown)


Runa-Dacino:
I'm not sure an icon tag is the right one for this, I'm not changing any icon files, just combining existing ones!

Leshana:
Code looks like it should work.  If people are okay with these items they are OK codewise.

PR #7209, opened 2020-04-11
Replace ring in holo boxing ring with actual ring.

H0lySquirr3l:
Yada yada yada 3rd times the charm.

image

With updated sprites ported over from chompers.  
https://github.com/CHOMPStation2/CHOMPStation2


TheFurryFeline:
Yay for MapDiffbot working again! <3

PR #7210, opened 2020-04-11
Changes shuttle landing soon warning

Arokha:


Novacat:
I wonder if Polaris would want this?

Cyantime:
These sprites doesn't give an idea of just how soon the shuttle will appear like the current ones do, is that intentional?

Arokha:
> These sprites don't give an idea of just how soon the shuttle will appear like the current ones do, is that intentional?

The current one has no idea either, it's just a 4.9 second animation that then deletes itself. It has no actual correlation to shuttles landing (and infact deletes itself before the shuttle even shows up because 4.9 seconds isn't generally long enough, leaving a window where there's no markers and no shuttle).

I'd consider the circles a problem actually because they DO indicate a time meaning you can never change the time it takes to land any shuttle, nor add any variation between them, because like I said, the animation is just a 4.9 second animation that then deletes itself rather than having some connection to shuttles.

PR #7212, opened 2020-04-11
Gun Refactors and Changes

Novacat:
- The Protector is now a minature energy gun, it has 3 lethal, 6 standard stun, and no special stun shot anymore. It retains the alert level lock, there is a version without the lock.
- The energy net gun now gets 3 net shots, but loses its stun alternate fire. It also gets a new sprite from baystation.
- Refactors all Vorestation weapons in vore/fluffstuff to be in their appropriate directories.
- Removes all personalized custom weapons.
- KHI is no longer the manufacturer for the Protector, Net Gun, and AR Glasses. I wanted to make KHI a bit rarer, and their specialty mainly ultra high tech equipment.
- Refactors Expedition Phasers. No practical change, but now there are unlocked versions of all three phaser variants.
- Removes unused sniper_vr file


TheDavestDave:
the prolathe can make the cell gun and cells that fire nets. You may aswell remove the netgun at that point, it offers nothing new, and things still attack you in the net.

PR #7225, opened 2020-04-13
[MIRROR] Add 2 new pipe types

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/6967
--------------------
FUEL pipe (yellow) and AUX pipe (cyan).

Sort of like scrubbers and supply special magic offset pipes, you can place both on the same turf. So now you can get 5 pipes in a single turf. Also adds a portables connector that only does fuel, and a portables connector that only does aux. Also adds a vent pump that connects directly to aux (and only aux), also for that airlock idea. Also adds special pumps that only do fuel and aux, so you can have those on the same turf as normal pumps too.

The intent of aux is to be whatever, but could be used for things like the airlock air supply cans and whatnot. Fuel is meant for overmap ship fuel, and can connect from the big phoron tanks too.

dreamseeker_2020-04-05_22-49-27

dreammaker_2020-04-05_22-49-34


Leshana:
Nothing to do here.

PR #7240, opened 2020-04-13
Fixes Azura's collar

Heroman3003:
No spaces, capitalizations or underscores in ckeys.

Also adds the comment in the beginning of the file as hopefully a preventative measure


CRTXBacon:
It would be worth commenting that into the file so people actually know this.

Heroman3003:
Added a comment to beginning of the file

PR #7245, opened 2020-04-14
2 Technomancer things

Arokha:
Mark gets an animation and an icon on the turf you marked so you can see where it was.

Adds a 'universal' catalog and core, nerfed versions of the basic core that are usable by non-technomancers.

Also maps them in on aro.dmm so I can use them for fun


Novacat:
To clarify, the mark/recall stuff should probably be sent to Polaris

Arokha:
yaya I figured

PR #7257, opened 2020-04-14
Adds a fax machine to the ITV Talon

Arokha:
Also adds 'trader' reply fax choice


code/modules/paperwork/adminpaper.dm
```diff
@@ -40,11 +40,15 @@
...
+	var/logo = alert(usr, "Do you want the header of your fax to have a NanoTrasen, SolGov, or Trader logo?","Fax Logo","NanoTrasen","SolGov","Trader") //VOREStation Add - Trader
 	if(logo == "SolGov")
 		logo = "sglogo.png"
-	else
+	else if(logo == "NanoTrasen")
 		logo = "ntlogo.png"
+	//VOREStation Add
+	else
+		logo = "trader.png"
+	//VOREStation Add End
```
Novacat, line 51: It would probably be better to just do
	if(logo == "Trader")
		logo = "trader.png"
and keep else logo = "ntlogo.png"

This would eliminate the issue of having an unmarked vorestation edit (the change too line 46) that could trip up maintainers working on this.

Arokha: I'll just move the comment up. I don't like the code being less efficient, since in your suggestion it would always run an additional branching check over what it has now.

PR #7266, opened 2020-04-15
Ports tg cablecoils instead of bay's

Heroman3003:
Still thicker than before, but without looking like unicorn poo.


Arokha:
Do that thing and fill in the center with 1 alpha black

PR #7269, opened 2020-04-15
Shipself 2.0

Arokha:
Just a newer version of Aronai's shipself. Smoller, more quaint, less wasted space, good bedrooms to lock Joan and Ascian in.


Arokha:
marked for the mech fix

PR #7276, opened 2020-04-15
Adjusts Hacking of NIFSoft Vendor, Resolves #1761

### LINKED ISSUE #1761: Hacking ID scan is non functional on the nif vendor.
breaking access requirements on the NIFsoft requires an E-mag

Pulse the ID scan wire, and be able to buy things that would normally not be available to my departmental access.

#### What actually happened

```
//Specific soft access checking
			var/datum/nifsoft/path = R.item_path
			if(initial(path.access))
				var/list/soft_access = list(initial(path.access))
				var/list/usr_access = usr.GetAccess()
				if(!has_access(soft_access, list(), usr_access) && !emagged )
					usr << "<span class='warning'>You aren't authorized to buy [initial(path.name)].</span>"
					flick(icon_deny,src)
					return

```
This bit here makes it so that you get access denied when you try to buy something you shouldn't normally have access to, in order to fix the problem, the line has to be changed to this.

`	if(!has_access(soft_access, list(), usr_access) && !emagged && scan_id)`

#### Steps to reproduce
Hack ID scan on vending machine, ID scan isn't disabled.

#### Code Revision
https://github.com/VOREStation/VOREStation/commit/256e1db4d6e43368e3e53cf3027e5ae395abdacd
 
The emag check is likely a balance thing, because of things like the little two shot gun you can get that costs 2200 credits.

That said, the check itself makes the vending machine inconsistent with every other vendor in the game.

All emagging a vending machine does is toggle IDscan and toggle Contraband. Which is something that can be done with tools.

Basically, because the code check to determine whether or not you can buy things is tied to the emagged variable instead of the id_scan variable, hacking the vending machine for ID scan does absolutely nothing.

Which is both inconsistent with every other vendor in the game, and inconsistent with how emags work in relation to vending machines.

KasparoVy:
Hacking the NIFSoft vendor by pulsing the ID scan wire now behaves like other vendors, allowing you to unlock ID restriction on the various NIFSofts.

You can not enable contraband on the nifsoft shop by hacking, though it can be accessed with an emag.

You can restore the ID lock and hide contraband again by cutting & mending the appropriate wires.
If someone restores either the ID lock & contraband restriction by working with the wires, you can re-emag the machine.

Issue & solution documented by @Dankman in #1761 and confirmed to still be an issue by @Zeldazackman
This issue was acted on as it was marked as a moderately severe bug:
it's a bug man

Changed on review by @Novacat

### DIFF
--- a/code/modules/nifsoft/nif_softshop.dm
+++ b/code/modules/nifsoft/nif_softshop.dm
@@ -139,7 +139,7 @@
 			if(initial(path.access))
 				var/list/soft_access = list(initial(path.access))
 				var/list/usr_access = usr.GetAccess()
-				if(!has_access(soft_access, list(), usr_access) && !emagged)
+				if(scan_id && !has_access(soft_access, list(), usr_access) && !emagged)
 					to_chat(usr, "<span class='warning'>You aren't authorized to buy [initial(path.name)].</span>")
 					flick("[icon_state]-deny",entopic.my_image)
 					return
--- a/code/modules/nifsoft/nif_softshop.dm
+++ b/code/modules/nifsoft/nif_softshop.dm
@@ -234,7 +234,7 @@
 		return
 
 /obj/machinery/vending/nifsoft_shop/emag_act(remaining_charges, mob/user) //Yeees, YEEES! Give me that black market tech.
-	if(!emagged)
+	if(!emagged || !(categories & CAT_HIDDEN))
 		emagged = 1
 		categories |= CAT_HIDDEN
 		to_chat(user, "You short out [src]'s access lock & stock restrictions.")
--- a/code/modules/nifsoft/nif_softshop.dm
+++ b/code/modules/nifsoft/nif_softshop.dm
@@ -19,6 +19,10 @@
 
 /obj/machinery/vending/nifsoft_shop/Initialize()
 	. = ..()
+
+	if(wires)
+		qdel(wires)
+	wires = new /datum/wires/vending/no_contraband(src) //These wires can't be hacked for contraband.
 	entopic = new(aholder = src, aicon = icon, aicon_state = "beacon")
 
 /obj/machinery/vending/nifsoft_shop/Destroy()
@@ -70,9 +74,9 @@
 		list(premium, CAT_COIN))
 
 	for(var/current_list in all_products)
-		var/category = current_list[2]
+		var/category = current_list[CAT_HIDDEN]
 
-		for(var/entry in current_list[1])
+		for(var/entry in current_list[CAT_NORMAL])
 			var/datum/nifsoft/NS = entry
 			var/applies_to = initial(NS.applies_to)
 			var/context = ""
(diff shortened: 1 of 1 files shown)


Heroman3003:
This makes obtaining the illegal NIFsofts way too easy, IMO. That was the main reason why hacking IDscan did nothing, because those are only meant to be obtainable with emag and nothing else (and that's why they have access restriction that doesn't actually exist too)

KasparoVy:
Alright, just need maintainer confirmation or something and I'll close this & not re-open it again.
I worked on this issue because of this 
it's a bug man

Arokha:
I personally think NIFsoft vendors should get special treatment because they have weapons in them. Maybe if hacking them were somehow harder. But I'd want the same thing on any vendor that had lethal weapons.

KasparoVy:
I think I have an idea.
More wires + wire-splicing

Using wirecutters & multitool perhaps to splice two wires together, and you need to have the right combination in order to unlock whatever function of the vendor

That sound reasonable @Arokha? Merely spitballing and IDK how to work it out from a design standpoint yet but I'm certain it's feasible

eg. You cut two wires and a button appears that says 'splice' & asks you to pick another cut wire, and when spliced the text updates to show that the two wires are spliced or whatever, and only then does the feature unlock.

Heroman3003:
Tbh it would still be easily achievable with simple hacking and a little time. Emag is a rare maintenance find and technically an antag tool. So you cannot reliably just go ahead and hack the NIFsoft with it as a requirement, because you need to find it in the first place. That and being an antag item and giving access to even more antag-level tools makes more sense than those tools being within access of anyone with gloves and basic tools.

Heroman3003:
Tbh it would still be easily achievable with simple hacking and a little time. Emag is a rare maintenance find and technically an antag tool. So you cannot reliably just go ahead and hack the NIFsoft with it as a requirement, because you need to find it in the first place. That and being an antag item and giving access to even more antag-level tools makes more sense than those tools being within access of anyone with gloves and basic tools.

KasparoVy:
Sounds good to me, I'll double check that function.

KasparoVy:
Done @Novacat,

1. You can't hack wires to enable contraband-- though you can disable it!
2. The only way to get contraband from the nifsoft shop is by emagging, which is now possible.

TheFurryFeline:
"Small" problem when testing this latest change. Contraband wire is rendered useless and has to be left alone after emagging to keep showing the illegal programs. Emagging does count as pulsing the wire but touching it again will lock the illegals away forever. You will not be allowed to interact with the contraband wire afterwards. This is critical to fix.

KasparoVy:
Contraband wire completely disabled

KasparoVy:
Adjusted emagging so you can re-emag the thing if either the access lock || contraband restriction is restored

KasparoVy:
What would ideal behaviour be? Do we want pulsing to remain doing nothing or to have it only 'fix' ID scanning rather than toggling similar to how wirecutters only reset contraband?

KasparoVy:
IDScan wire doing nothing when pulsed should be iight because cutting & mending restores IDscan in the exact same way as it hides contraband again

KasparoVy:
Nice. Hackerman.

PR #7294, opened 2020-04-16
Adds the "Baa" Poster

KitchiFox:
Adds a poster of a bleat


KasparoVy:
Interesting :eyes: I wonder why this was closed

PR #7296, opened 2020-04-16
Nerfs Heat Damage

Novacat:
Right now heat is super lethal super quickly, this is an attempt to fix that.

Also caps the temperature increase of burning human bodies.

Bumps the threshhold for husking

### DIFF
--- a/code/modules/mob/living/carbon/human/human_damage.dm
+++ b/code/modules/mob/living/carbon/human/human_damage.dm
@@ -1,6 +1,6 @@
 //Updates the mob's health from organs and mob damage variables
 /mob/living/carbon/human/updatehealth()
-	var/huskmodifier = 1.5 // With 1.5, you need 250 burn instead of 200 to husk a human.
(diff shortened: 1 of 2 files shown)


KasparoVy:
[Thank you kindly](https://www.youtube.com/watch?v=It7107ELQvY)

Wickedtemp:
The other reasons that make overheating so dangerous is the fact that you can pretty much heat up infinitely, until you reach a point where you're long-since-removed from the source of heat, but your body temperature is so high, you're cooking to death despite not being on fire or near the source of heat anymore. 

There's no 'cap' and you don't return back to a normal temperature quickly enough. And then there's husking, which I think needs to at least be reworked, added onto this to create situations where someone encounters a source of severe heat, and despite running from it and successfully getting away, they still burn to death and are unable to be resuscitated in mere seconds.

Arokha:
TBH This whole thing operating based on body temperature should probably be changed at some point. That makes sense for hypothermia and hyperthermia, but isn't... really related to when you get BURNED.

PR #7306, opened 2020-04-17
Deployable Escape Pods

Novacat:
- Adds Surfluid Escape Pod Capsule
- Adds code support for unique deployables


Natje0:
I can only guess what this is, but it sounds rad

PR #7312, opened 2020-04-17
Ports some sprites from Goonstation

AmayaWah:
Adds void + voidwarp turf icons.
Adds timeship turf icons.


KasparoVy:
Ah, the void. Good ol' Goon...

PR #7314, opened 2020-04-17
Adjust ghosts hearing whispers/subtles

Arokha:
Adds a preference to toggle whether or not ghosts can hear your whispers/subtles.

For whispering, this changes the 'nothing at all' that happens currently (a bug?) when someone whispers for ghosts into getting the same message someone would get if they were standing far away from a whisper ingame.

For subtles, this removes the ability of ghosts to read the contents of subtle messages.

### DIFF
--- a/code/modules/client/preference_setup/global/setting_datums.dm
+++ b/code/modules/client/preference_setup/global/setting_datums.dm
@@ -104,6 +104,12 @@ var/list/_client_preferences_by_type
 	key = "EMOTE_NOISES"
 	enabled_description = "Noisy"
 	disabled_description = "Silent"
+/datum/client_preference/whisubtle_vis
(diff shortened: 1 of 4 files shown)


Heroman3003:
I don't like this change. It won't help that much at all with ghosts spying on scenes unless scene itself is semi-public in the first place (since fully private ones use says and mes anyway), but reduces ability to see it. From what I've seen most people are alright with ghosts seeing their scenes, and just don't like when ghosts actively making themselves known as actively watching it.

Heroman3003:
Making it a preference is a much better solution, but IMO default should be the original behaviour, not the new one.

KasparoVy:
RIP the point of being a ghost lmao

PR #7322, opened 2020-04-18
Port some random clothes

Arokha:
- 3 colors of qipao (added to loadout, and vendor)
- 2 kinds of crown so Jemli's character can be a real princess
- Syndie contractor suit/helmet
- Chrono suit/helmet
- H.E.C.K suit/helmet
- Advanced magboots (faster), gave to CE

- Removes `KEEP_TOGETHER` on literally every atom. It has a performance penalty! Adds it to just mobs. Everything else is pixel scale anyway with no way to change it other than varediting.

dreamseeker_2020-04-18_18-09-55
dreamseeker_2020-04-18_18-09-59
dreamseeker_2020-04-18_19-40-39


Heroman3003:
Advanced magboots kinda make no sense as its own item given to CE. Since CE doesn't have a regular voidsuit to begin with and magboots are already included in the Advanced Hardsuit module that they get.

Arokha:
They can have them anyway. If they are an unfortunate species (like me) who just can't wear any spacesuits at all by default then they should probably have some boots.

code/modules/clothing/spacesuits/rig/rig_pieces_vr.dm
```diff
@@ -43,3 +43,13 @@
 
 /obj/item/clothing/suit/space/rig
 	species_restricted = list(SPECIES_HUMAN, SPECIES_SKRELL, SPECIES_TAJ, SPECIES_UNATHI, SPECIES_NEVREAN, SPECIES_AKULA, SPECIES_SERGAL, SPECIES_ZORREN_HIGH, SPECIES_VULPKANIN, SPECIES_PROMETHEAN, SPECIES_VOX, SPECIES_TESHARI, SPECIES_VASILISSAN, SPECIES_RAPALA, SPECIES_ALRAUNE, SPECIES_FENNEC, SPECIES_XENOHYBRID)
+
+/obj/item/clothing/shoes/magboots/rig/ce
+	name = "advanced boots"
```
Heroman3003, line 48: probably should still be named "advanced magboots" here tbh

Arokha: I think they get renamed regardless by the suit

PR #7323, opened 2020-04-19
Make Talon engines BEEFIER

Arokha:


KasparoVy:
Buff ass engines, rightfully so though

Looks good and makes sense

Arokha:
Hard, lazy, someone else can!
You'll also need to do that for the 3x3 one that's not depicted

KasparoVy:
note to self [this](https://github.com/VOREStation/VOREStation/blob/d91be67befc75bac80b1fd9d9b8448e80182624e/code/modules/overmap/ships/engines/gas_thruster.dm#L168) is where the exhausty bits happen

PR #7329, opened 2020-04-19
Replace old binocular sprite

Arokha:
dreammaker_2020-04-18_23-56-02
dreammaker_2020-04-18_23-56-38


code/game/objects/items/devices/binoculars.dm
```diff
@@ -1,6 +1,7 @@
 /obj/item/device/binoculars
 	name = "binoculars"
 	desc = "A pair of binoculars."
+	icon = 'icons/obj/device_vr.dmi'
```
Novacat, line 4: Probably should be //VOREStation Edit here

PR #7343, opened 2020-04-19
Adds G R A V Y

Arokha:
- Add gravity generator on Z5, connected to engineering.
- It provides gravity for Z5, Z6, and Z7. If it loses power, gravity also leaves.
- Updates gravity loss events to turn off the generator rather than magically remove gravity. An engineer can get to it and turn it back on instead of waiting.


Heroman3003:
What happens if it is destroyed? Is gravity gone forever?

Arokha:
> What happens if it is destroyed? Is gravity gone forever?

FOREEEEVERRRRRR but really like the only thing that destroys it is a tesloose. Everything else (even explosions) just damage it and you gotta repair it instead!

Heroman3003:
Well, the Tesloose is precisely what I worry about here. If it just happens to go in its direction, there's possibility that despite damage being repairable otherwise, the space levels would be barely habitable due to permanent lack of gravity for rest of the round.

Arokha:
> due to permanent lack of gravity for rest of the round.

I'm not sure I'd call lack of gravity "barely habitable"... just stay near a wall! But really though just adminhelp or pray for a new generator. Admemes can spawn one for you, it's a single object that builds itself.

I'm sure Leshana would like to make them constructable too.

KillianKirilenko:
Needs to be overclockable (via hacking/emag?) with multiple power levels so that it applies slowdown or even pins everyone to the floor.

Would also mean you could add more variants to the generator failure event to make it less predictable.

Arokha:
:0 That's an interesting idea, I like that

KillianKirilenko:
Quickly put together a standalone .dmi with palette-swapped orange 'overclocked' state that spins faster: [overclocked.zip](https://github.com/VOREStation/VOREStation/files/4501000/overclocked.zip)

(Speaking of dmis, I don't see the replaced generators in the icon diffbot?)

Arokha:
Replaced generators? We didn't have a gravity generator sprite at all before, aside from a lame placeholder that was the 'singulo/tesla spawner' icon.

Are you thinking of https://github.com/VOREStation/VOREStation/pull/7346

KillianKirilenko:
...somehow, yes. Excuse me, I've been sleeping awfully lately.

PR #7345, opened 2020-04-20
Reenable Air Pump Soundloop

Rykka-Stormheart:
After extensive testing, there's no way to simply confine the airpump to a single sound channel without extreme wonkiness/inaudible audio, and the suggestion of a global audio channel resulted in no audio playing at all.

Attempting to make a static loop caused horrible feedback, so that was a no-go, and attempting to set up a check to find out if there are mobs near a vent (basically a proximity check) does the same thing/uses the same amount of resources that simply doing playsound() does - but in reverse.

Tested with @ShadowLarkens's help

### DIFF
--- a/code/ATMOSPHERICS/components/unary/vent_pump.dm
+++ b/code/ATMOSPHERICS/components/unary/vent_pump.dm
@@ -47,7 +47,7 @@
 	var/radio_filter_out
 	var/radio_filter_in
 
-	//var/datum/looping_sound/air_pump/soundloop //VOREStation Removal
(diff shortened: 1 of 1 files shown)


Heroman3003:
I still am on the side of the vent idle sounds being a thing at all being a horrible idea.

Rykka-Stormheart:
> I still am on the side of the vent idle sounds being a thing at all being a horrible idea.

They have a preference so you don't have to listen to it/hear it if you don't want to?

PR #7353, opened 2020-04-20
Adds rutile (titanium ore) and other materials

Arokha:
- Adds rutile ore (an ore containing titanium) to Virgo 2 and belt areas. It doesn't spawn on Virgo3b.
- Adds a better stack icon for titanium.
- Adds plastitanium, an alloy of titanium and plasteel.
- Adds titanium and plastitanium glass. It's strong!

If you want titanium, you gotta go out and get it. Can't get it on Virgo 3b.


code/modules/mining/mineral_effect.dm
```diff
@@ -1,6 +1,6 @@
 /obj/effect/mineral
 	name = "mineral vein"
-	icon = 'icons/obj/mining.dmi'
+	icon = 'icons/obj/mining_vr.dmi' //VOREStation Edit
```
Novacat, line 3: I am a bit concerned about this considering the potential of it breaking things coming in from upstream.

Arokha: If they add more ores, you mean? That's pretty unlikely! We've got other things like this too, like we had computers for quite some time, among other things.

I'll add an error iconstate to make it obvious if something breaks, and to let miners still see the 'new' ore until fixed. :p

KasparoVy:
The more ore for me to mine the better! I just wish I could make machetes hahaha.

PR #7361, opened 2020-04-21
Make unbuckled shuttle movement more exciting

Arokha:

### DIFF
--- a/code/modules/shuttles/shuttle.dm
+++ b/code/modules/shuttles/shuttle.dm
@@ -331,6 +331,10 @@
 						//M.throw_at_random(FALSE, 4, 1)
 						if(istype(M, /mob/living/carbon))
 							M.Weaken(3)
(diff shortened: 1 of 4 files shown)


Arokha:
It didn't damage me at all during testing, actually. I suspect being thrown as a human doesn't do all that much. You can test what causes damage on your own if you want, by grabbing and throwing someone around. It makes a loud SMACK noise but... doesn't appear to do much of anything.

But, even if it did cause damage, I'm not sure how that'd be a problem (as long as it's not like instant fractures from one or two jumps). Seems like it'd make sense to meee.

KillianKirilenko:
There also ought to be enough time between 'hops' to get to a safe spot unless you're experiencing some *extreme* lag.

Reminds me - you still fall over during transit hops if you have magboots enabled. 'Fixing' that could be another way to deal with it.

PR #7366, opened 2020-04-21
Add friends around the station

Arokha:
dreamseeker_2020-04-21_13-39-44

But actually add cardboard cutouts (except they're made of wood not cardboard).

Paint them with a closet painter, floor painter, or paintbucket.


Heroman3003:
But why not cardboard? It exists and is pretty acessible

PR #7370, opened 2020-04-22
Enabled the Smelting of Plastitanium

KasparoVy:
Maybe oversight since can make titanium (smelting rutile), titanium glass & plastitanium glass but not plastitanium.

Used product modifier of plasteel & components of plastitanium glass minus the sand


Arokha:
Ye, oversight.

KasparoVy:
Thanks! Wiki updated

PR #7374, opened 2020-04-22
Fix some power issues

Arokha:
- ARTGs produce more power because that's what Cheri wants
- Talon cable fixes for point defense
- Fix void cores on muh ship
- Move some APCs around
- Nothing special really

### DIFF
--- a/code/modules/power/port_gen_vr.dm
+++ b/code/modules/power/port_gen_vr.dm
@@ -29,6 +29,7 @@
 	req_components = list(
 		/obj/item/stack/cable_coil = 5,
 		/obj/item/weapon/stock_parts/capacitor = 1,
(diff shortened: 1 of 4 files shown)


code/modules/power/port_gen_vr.dm
```diff
@@ -29,6 +29,7 @@
 	req_components = list(
 		/obj/item/stack/cable_coil = 5,
 		/obj/item/weapon/stock_parts/capacitor = 1,
+		/obj/item/weapon/stock_parts/micro_laser = 1,
```
Heroman3003, line 32: Why a micro laser and not just... two capacitor slots? Most machines usually have different parts have different effects.

Arokha: Same as source

PR #7405, opened 2020-04-23
Static Machinery Power

Leshana:
### Implements "static" area machinery power usage
 Instead of using auto_use_power to re-tally up machinery's power usage every cycle, track the steady "static" load separately from the transient "oneoff" usage.  Machines then only need to inform the area when they use oneoff power or *change* their steady usage.
Many machines *never* change their power usage, and have no other reason to be processing, allowing us to potentially remove *thousands* of machines from the processing list.
Note: Another PR will follow this one shortly which optimizes a few types to process less.  Its separated into another PR to make reviewing this one easier.

- Remove auto_use_power and stop SSmachines from calling it.
- Add vars to track "static" usage for each of the three power channels to /area
- Rename the existing three vars to "oneoff" so its clear what they mean (and to catch people accidentally updating them directly)
- Update area power procs and APCs to use the new variables.
- Rename /area/proc/use_power() to use_power_oneoff() to make it clear what it is doing.
- Deprecate /obj/machinery/use_power() in favor of use_power_oneoff() but don't delete yet.  Can transition gradually.
- Add logic to the update_power procs on machines to calculate the deltas and update static area power whenever their usage changes.
- Add logic to machines to update area power when they are created, destroyed, or move.
- Moved /obj/machinery procs related to area power usage into machinery_power.dm to make them easier to find.
- Added or updated comments in several places to explain what is going on and how to use it.

### Fixes Remove extra ..() from /turf/Entered
Fixed this too since it was causing all "moved" observers to fire twice which was screwing up the power counts.

### DIFF
--- a/code/game/area/areas.dm
+++ b/code/game/area/areas.dm
@@ -289,12 +289,13 @@
 			oneoff_light += amount
 		if(ENVIRON)
 			oneoff_environ += amount
+	return amount
 
 // This is used by machines to properly update the area of power changes.
 /area/proc/power_use_change(old_amount, new_amount, chan)
 	use_power_static(new_amount - old_amount, chan) // Simultaneously subtract old_amount and add new_amount.
 
-// Not a proc you want to use directly unless you know what you are doing; see use_power_oneoff below instead.
+// Not a proc you want to use directly unless you know what you are doing; see use_power_oneoff above instead.
 /area/proc/use_power_static(var/amount, var/chan)
 	switch(chan)
 		if(EQUIP)
--- a/code/game/area/areas.dm
+++ b/code/game/area/areas.dm
@@ -320,6 +320,34 @@
 
 //////////////////////////////////////////////////////////////////
 
+/area/vv_get_dropdown()
+	. = ..()
+	VV_DROPDOWN_OPTION("check_static_power", "Check Static Power")
+
+/area/vv_do_topic(list/href_list)
+	. = ..()
+	IF_VV_OPTION("check_static_power")
+		if(!check_rights(R_DEBUG))
+			return
+		src.check_static_power(usr)
+		href_list["datumrefresh"] = "\ref[src]"
+
+// Debugging proc to report if static power is correct or not.
+/area/proc/check_static_power(var/user)
+	set name = "Check Static Power"
+	var/actual_static_equip = static_equip
+	var/actual_static_light = static_light
+	var/actual_static_environ = static_environ
+	retally_power()
+	if(user)
+		var/list/report = list("[src] ([type]) static power tally:")
+		report += "EQUIP:   Actual: [actual_static_equip] Correct: [static_equip] Difference: [actual_static_equip - static_equip]"
+		report += "LIGHT:   Actual: [actual_static_light] Correct: [static_light] Difference: [actual_static_light - static_light]"
+		report += "ENVIRON: Actual: [actual_static_environ] Correct: [static_environ] Difference: [actual_static_environ - static_environ]"
+		to_chat(user, report.Join("\n"))
+	return (actual_static_equip == static_equip && actual_static_light == static_light && actual_static_environ == static_environ)
+
+//////////////////////////////////////////////////////////////////
 
 var/list/mob/living/forced_ambiance_list = new
 
--- a/code/game/area/areas.dm
+++ b/code/game/area/areas.dm
@@ -21,14 +21,22 @@
 	var/requires_power = 1
 	var/always_unpowered = 0	//this gets overriden to 1 for space in area/New()
 
-	var/power_equip = 1
-	var/power_light = 1
-	var/power_environ = 1
-	var/music = null
-	var/used_equip = 0
-	var/used_light = 0
-	var/used_environ = 0
+	// Power channel status - Is it currently energized?
+	var/power_equip = TRUE
+	var/power_light = TRUE
+	var/power_environ = TRUE
 
+	// Oneoff power usage - Used once and cleared each power cycle
+	var/oneoff_equip = 0
+	var/oneoff_light = 0
+	var/oneoff_environ = 0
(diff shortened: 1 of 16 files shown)


code/game/area/areas.dm
```diff
@@ -251,33 +259,95 @@
...
+			oneoff_light += amount
 		if(ENVIRON)
-			used_environ += amount
+			oneoff_environ += amount
 
+// This is used by machines to properly update the area of power changes.
+/area/proc/power_use_change(old_amount, new_amount, chan)
+	use_power_static(new_amount - old_amount, chan) // Simultaneously subtract old_amount and add new_amount.
+
+// Not a proc you want to use directly unless you know what you are doing; see use_power_oneoff below instead.
```
Arokha, line 297: `see use_power_oneoff below instead.`
Above

code/game/machinery/machinery.dm
```diff
@@ -51,28 +55,20 @@ Class Procs:
...
-
-   powered(chan = EQUIP)         'modules/power/power.dm'
+   powered(chan = CURRENT_CHANNEL)         'game/machinery/machinery_power.dm'
       Checks to see if area that contains the object has power available for power
       channel given in 'chan'.
 
-   use_power(amount, chan=EQUIP, autocalled)   'modules/power/power.dm'
+   use_power_oneoff(amount, chan=CURRENT_CHANNEL)   'game/machinery/machinery_power.dm'
       Deducts 'amount' from the power channel 'chan' of the area that contains the object.
       If it's autocalled then everything is normal, if something else calls use_power we are going to
```
Arokha, line 68: Should this be autocalled?

code/game/machinery/machinery_power.dm
```diff
@@ -0,0 +1,183 @@
+//
+// /obj/machinery POWER USAGE CODE HERE! GO GO GADGET STATIC POWER!
+// Note: You can dinf /obj/machinery/power power usage code in power.dm
```
Arokha, line 3: find

code/game/machinery/machinery_power.dm
```diff
@@ -0,0 +1,183 @@
...
+		return 0
+
+	//Don't do this. It allows machines that set use_power to 0 when off (many machines) to
+	//be turned on again and used after a power failure because they never gain the NOPOWER flag.
+	//if(!use_power)
+	//	return 1
+
+	var/area/A = src.loc.loc		// make sure it's in an area
+	if(!A || !isarea(A))
+		return 0					// if not, then not powered
```
Arokha, line 34: Maybe `get_area(src)`? get_area performs the isarea(). You can also pass nulls to isarea (it's a byond builtin and doesn't care) so could remove the `!A`

code/game/machinery/machinery_power.dm
```diff
@@ -0,0 +1,183 @@
...
+
+// Get the amount of power this machine will consume each cycle.  Override by experts only!
+/obj/machinery/proc/get_power_usage()
+	return POWER_CONSUMPTION
+
+// DEPRECATED! - USE use_power_oneoff() instead!
+/obj/machinery/proc/use_power(var/amount, var/chan = -1) // defaults to power_channel
+	return src.use_power_oneoff(amount, chan);
+
+// This will have this machine have its area eat this much power next tick, and not afterwards. Do not use for continued power draw.
```
Arokha, line 60: So does this mean it will NOT use the amount of power it would normally use, and INSTEAD uses the power in oneoff? Or is the oneoff draw IN ADDITION TO the normal power usage (idle or active)

code/game/machinery/machinery_power.dm
```diff
@@ -0,0 +1,183 @@
...
+// Returns actual amount drawn (In theory this could be less than the amount asked for. In pratice it won't be FOR NOW)
+/obj/machinery/proc/use_power_oneoff(var/amount, var/chan = CURRENT_CHANNEL)
+	var/area/A = get_area(src)		// make sure it's in an area
+	if(!A || !isarea(A))
+		return
+	if(chan == CURRENT_CHANNEL)
+		chan = power_channel
+	A.use_power_oneoff(amount, chan)
+	return amount // TODO - Return A.use_power_oneoff()'s retval once it returns something.
```
Arokha, line 69: todo

code/game/machinery/machinery_power.dm
```diff
@@ -0,0 +1,183 @@
...
+// Registering moved_event observers for all machines is too expensive.  Instead we do it ourselves.
+// 99% of machines are always on a turf anyway, very few need recursive move handling.
+/obj/machinery/Move()
+	var/old_loc = loc
+	if((. = ..()))
+		update_power_on_move(src, old_loc, loc)
+		if(ismovable(loc)) // Register for recursive movement (if the thing we're inside moves)
+			GLOB.moved_event.register(loc, src, .proc/update_power_on_move)
+		if(ismovable(old_loc)) // Unregister recursive movement.
+			GLOB.moved_event.unregister(old_loc, src, .proc/update_power_on_move)
```
Arokha, line 103: Scary but if you tested and this doesn't leave dangling listeners it's fine! Nothing sets loc right?

code/modules/overmap/ships/computers/computer_shims.dm
```diff
@@ -12,10 +12,6 @@
 // Power
 //
 
-// This will have this machine have its area eat this much power next tick, and not afterwards. Do not use for continued power draw.
-/obj/machinery/proc/use_power_oneoff(var/amount, var/chan = -1)
-	return use_power(amount, chan)
-
```
Arokha, line 18: Unshim \o/

code/game/machinery/machinery_power.dm
Leshana: This is in addition to the normal power usage (if any)

PR #7406, opened 2020-04-24
Optimize some machines for static power

Leshana:
Now that we have static machinery power, machines don't need to process just to use idle power.   And yet some still do so all the time despite doing nothing!  Here we take a few of the most noticeable cases and make them process only when they have something to actually do.

- Firedoors should only process when they are closed (and need to update their lights)
- Doors only need to process when they are attempting to autoclose.
- Airlocks only need to process when they are obeying a command or counting down a timer.
- Terminals do not need to process! global.machines is a separate list from processing_machines
- Holopads only need to process when someone is using them.

By rough counting, this reduces the number of processing machines on Tether by about 5000

### DIFF
--- a/code/game/machinery/doors/airlock.dm
+++ b/code/game/machinery/doors/airlock.dm
@@ -1142,7 +1142,7 @@ About the new airlock wires panel:
 					if(!has_beeped)
 						playsound(src.loc, 'sound/machines/buzz-two.ogg', 50, 0)
 						has_beeped = 1
-					close_door_at = world.time + 6
+					autoclose_in(6)
 					return
 
 	for(var/turf/turf in locs)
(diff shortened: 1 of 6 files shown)


code/game/machinery/doors/airlock.dm
```diff
@@ -631,6 +638,9 @@ About the new airlock wires panel:
 		message = "The door is now electrified [duration == -1 ? "permanently" : "for [duration] second\s"]."
 		src.electrified_until = duration == -1 ? -1 : world.time + SecondsToTicks(duration)
 
+	if(src.electrified_until > 0)
+		START_MACHINE_PROCESSING(src)
```
Arokha, line 642: Me being picky: all the other START_MACHINE_PROCESSING branches don't use src.

PR #7413, opened 2020-04-24
Fixes simplemobs being able to attack in bellies

Verkister:
It's just fair since humans can't do it. Also might even let a voremob survive a nonvoremob encounter for once.

### DIFF
--- a/code/modules/mob/living/simple_mob/combat.dm
+++ b/code/modules/mob/living/simple_mob/combat.dm
@@ -2,7 +2,7 @@
 /mob/living/simple_mob/proc/attack_target(atom/A)
 	set waitfor = FALSE // For attack animations. Don't want the AI processor to get held up.
 
-	if(!A.Adjacent(src))
+	if(!A.Adjacent(src) || isbelly(src.loc)) //VOREStation Edit
 		return ATTACK_FAILED
 	var/turf/their_T = get_turf(A)
 


Novacat:
I am wondering how easy it would be to have it so that simplemobs also cannot attack while netted, or at the very least, make it so that they cannot both tear out of the net and attack at the same time.

Arokha:
Personally I think simplemobs should be able to attack from inside nets (in melee). If you put a tiger in a net unless the net is a very fine mesh they're just gonna stick their pawbs through and smack you with them when you get close. BUT for good gameplay it can't stay that way forever, So maybe simplemobs are dangerous until X time has passed? Or maybe there's a % chance they'll just give up when netted? So they might break out a couple times but finally give up when netted and turn docile? OR PORT STAM COMBAT

code/modules/mob/living/simple_mob/combat.dm
```diff
@@ -2,7 +2,7 @@
 /mob/living/simple_mob/proc/attack_target(atom/A)
 	set waitfor = FALSE // For attack animations. Don't want the AI processor to get held up.
 
-	if(!A.Adjacent(src))
+	if(!A.Adjacent(src) || isbelly(src.loc)) //VOREStation Edit
```
Arokha, line 5: TBH You might want to just return at the start of their handle_stance stuff, it will save CPU, unless there is something else you want them to be doing in bellies

Arokha:
Superseded by https://github.com/VOREStation/VOREStation/pull/7474

PR #7415, opened 2020-04-24
[Untested] AI mobs have a chance to surrender when 'fully buckled'

Arokha:
At the moment the only thing you can do to 'fully buckle' a mob is to net it. Humans can be fully buckled by being restrained and buckled (like bucklecuffed) but that doesn't really apply to mobs. If someone wants to add more ways to do that, go for it.

Humans have a 30% chance to give up when netted, animals have a 10% chance to give up.

I made it so you can always escape from nets eventually (the escape time goes down by 1 second each failed attempt until it's 0 seconds and you automatically escape).

Testing Needed
---------
You might want to edit the surrender_chance on mobs to like 80% for testing this. The expected behavior is:
- Mobs won't surrender ever if you buckle them to a chair or something (to avoid that becoming the new meta) and they should immediately resist and keep attacking you.
- Mobs have a % chance to surrender when netted, and will print a visible message to let you know when they do. They will not resist on the AI tick they surrender and should stay in the net until someone releases them.
- Mobs should not attack while netted.
- Mobs SHOULD attack while buckled to chairs or other simple things without being restrained.
- Mobs who have surrendered should be non-hostile but should still retaliate.
- Mobs that have surrendered should still resist out of nets if they enter combat for some reason (retaliating, for e.g.).
- Mobs should not 'spam' resist, they should wait until one resist attempt is done before starting the next.

Scree's crazy suggestion that I added
----------
Animal surrender chances go up 10% each time they are netted and escape with nobody visible (representing them being confused about a situation that doesn't make any sense to them). Stand offscreen and use binoculars to keep an eye on.

Human surrender chances vary by 10% per person they are outnumbered by, or outnumber you by, to a minimum of 10% and a maximum of 100%. If you net them and drag them into a group of friends by themselves (assuming maybe there's 4 of you and 1 of them), their surrender chance will go up by (4-1)*10% = 30% from their base of 30% to 60% chance to surrender. If they don't and you immediately net them again, it's pretty likely you'll get them that time or the next.


KillianKirilenko:
Time for explo team to become SWAT and arrest merc mecha.

Actual feedback, though; perhaps mob current health vs max health should be considered for surrender chances too? AI-controlled simplemobs don't really track or consider "pain" as a concept as far as I'm aware, so this could be another way of making them a little less 'dumb'.

On a similar note, berzerk-capable critters (like the saviks) should probably not "surrender" whilst zerking out.

Arokha:
Interesting ideas! May implement

Wickedtemp:
petition for any mlem capable mob to do a mlem to signify surrender

SplinterGP:
so that makes mobs not attack when net gunned ? approve it. at all seemed like they dind't before on YW and after update they started to, but dind't test with older versions before we upstreamed, but seems pretty neat!

MrSign:
What would be the point in making mobs surrender? We would need to implement an incentive of wanting to take mobs alive rather than straight up kill them.

Perhaps hostile human mobs such as pirates, mercs, etc can be captured, returned to Tether and then ship them on a 'prison' shuttle that takes them back to Central (fluff-wise) and be rewarded. Maybe have a redeem system where each human mob you hand over to Central gives you points which can then be redeemed for rewards. Ranging from thalers to medical supplies to even equipment typically not available unless adminabused in. (Station-locked version of more powerful lasers and/or ballistics, perhaps? Themed armor/clothing that reflects their 'bounty hunter'/'peacekeeper' status?)

A similar system could be implemented for animal mobs. Except, obviously, fluff-wise, they get donated to a zoo or an off-station research facility that specializes in caring for animal mobs. Redeeming points for thalers, medical supplies or equipment. (Themed equipment. Zookeeper outfit/Safari outfit to show off how skilled they are in wrangling up animals?)

Without a system in place to promote making mobs surrender, people will simply forget that such a mechanic exists and rather than waste time hoping for that 10% or 30% chance of having them surrender. They can just make sure they're 100% dead and no longer a threat!

Arokha:
Too lazy to test/work on, maybe another time.

PR #7452, opened 2020-04-26
Remove unnecessary code, replace with area enter messages

Arokha:
Should speed up shuttle jumps slightly.


Arokha:
This isn't code that Polaris had!

PR #7453, opened 2020-04-26
Hiss, not His

InterroLouis:
That's it really. The emote is "his" right now instead of hiss. So, uh, yeah, fixies.


Heroman3003:
The reason its like that is because emote sanitization always removes the last 's' in the emote. To allow stuff like *cough and *coughs to be interchangable. Which is a problem with emotes that NATURALLY end with s, so that last s in those emotes s is removed. Making emotes like ye and his.

InterroLouis:
So sanitation machine broke?

Arokha:
You can also use *hiss as far as I'm aware

InterroLouis:
I wouldn't have made this if that were the case. It did, and still does, pull up an error when you use the *hiss command instead of the *his command. 
Unusable emote 'hiss'. Say *help or *vhelp for a list.

PR #7457, opened 2020-04-26
512 Compile, Ship Console Access Fix

Novacat:
- Fixes compile error with 512 versions of BYOND
- Fixes a bug that prevented access restrictions on ship consoles from working

### DIFF
--- a/code/modules/overmap/ships/computers/computer_shims.dm
+++ b/code/modules/overmap/ships/computers/computer_shims.dm
@@ -96,5 +96,8 @@
 /obj/machinery/computer/ship/attack_hand(mob/user)
 	if((. = ..()))
 		return
+	if(!allowed(user))
+		to_chat(user, "<span class='warning'>Access Denied.</span>")
+		return 1
 	if(CanUseTopic(user, DefaultTopicState()) > STATUS_CLOSE)
 		return interface_interact(user)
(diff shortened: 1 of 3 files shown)


code/game/machinery/machinery_power.dm
```diff
@@ -92,18 +92,18 @@
 	var/old_loc = loc
 	if((. = ..()))
 		update_power_on_move(src, old_loc, loc)
-		if(ismovable(loc)) // Register for recursive movement (if the thing we're inside moves)
+		if(ismovableatom(loc)) // Register for recursive movement (if the thing we're inside moves)
```
Leshana, line 95: Instead of these changes we can just add the line `#define ismovable(A) (istype(A, /atom/movable))`  to ___513_compatibility.dm_

code/modules/overmap/ships/computers/computer_shims.dm
```diff
@@ -94,7 +94,10 @@
 // If you don't call parent in this proc, you must make all appropriate checks yourself.
 // If you do, you must respect the return value.
 /obj/machinery/computer/ship/attack_hand(mob/user)
-	if((. = ..()))
+	if(..(user))
```
Leshana, line 97: Whai? It was already passing user upwards.  And now it won't return 1 on parent deny? Maybe it doesn't matter if nobody overrides this method, but whai?

Novacat: I was copying code from somewhere else and overwrote it in a hurry. Ill fix that.

PR #7474, opened 2020-04-27
Refactor AI to work better/faster

Arokha:
Splits AI into two subsystems, one for boring times, and one for action times. AI manages their presence in each one depending on their stance.

They stop processing when they die, and begin processing again if revived.

Also tweaks vore related to AI. They sleep their AI when they are in a belly, and resume it when they exit. Voremobs also stop trying to 'go after' someone they've eaten and resume doing other normal things, attacking other people, or wandering.

In terms of performance, the AI subsystem takes about 25% as long to run (about 40ms every 2 seconds, compared to 50ms every 0.5 seconds). There may be other performance improvements that I'm unaware of with regards to mobs, but whatever. This is better.


KasparoVy:
More subsystem wizardry!

PR #7475, opened 2020-04-27
Fix /Moved() that doesn't call parent.

Leshana:
Fix held lights not updating lighting because the parent proc didn't get called.


InterroLouis:
<3

PR #7481, opened 2020-04-27
Prevent protean repairs using fire and cables

Arokha:
Doesn't make much sense

### DIFF
--- a/code/game/objects/items/weapons/tools/weldingtool.dm
+++ b/code/game/objects/items/weapons/tools/weldingtool.dm
@@ -68,6 +68,11 @@
 		if(!S || S.robotic < ORGAN_ROBOT || S.open == 3)
 			return ..()
 
+		//VOREStation Add - No welding nanoform limbs
+		if(S.robotic > ORGAN_LIFELIKE)
+			return ..()
+		//VOREStation Add End
+
 		if(S.organ_tag == BP_HEAD)
 			if(H.head && istype(H.head,/obj/item/clothing/head/helmet/space))
 				to_chat(user, "<span class='warning'>You can't apply [src] through [H.head]!</span>")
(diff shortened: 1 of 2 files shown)


code/game/objects/items/weapons/tools/weldingtool.dm
```diff
@@ -68,6 +68,11 @@
 		if(!S || S.robotic < ORGAN_ROBOT || S.open == 3)
 			return ..()
 
+		//VOREStation Add - No welding nanoform limbs
+		if(S.robotic > ORGAN_LIFELIKE)
```
KasparoVy, line 72: This will include vey-med, zeng-hu & DSI limbs - ORGAN_NANOFORM should explicitly cover nanoform limbs afaik -- just saw the defines [here](https://github.com/VOREStation/VOREStation/blob/775d175b1146796f81b6a9673f222182cf2ef495/code/__defines/damage_organs.dm#L54) the code handling making it lifelike [here](https://github.com/VOREStation/VOREStation/blob/4f4a29b5fc264fab42eaf769e9df701b240293a1/code/modules/organs/organ_external.dm#L1120) and the actual limbs [here](https://github.com/VOREStation/VOREStation/blob/f60a39dd4d3504e217e190a97f63a9fb12be1396/code/modules/organs/robolimbs.dm) and [here](https://github.com/VOREStation/VOREStation/blob/f60a39dd4d3504e217e190a97f63a9fb12be1396/code/modules/organs/robolimbs_vr.dm)

KasparoVy:
Looking at the defines on the limbs this will also prevent repair of Zeng-Hu, Vey-Med & DSI limbs

code/game/objects/items/weapons/tools/weldingtool.dm
Arokha: No, because S.robotic on those will be ORGAN_LIFELIKE. The test is *greater-than* ORGAN_LIFELIKE, not *greater-than-or-equal-to*

Arokha:
See comment above about why that won't happen.

KasparoVy:
Nice, sorted!

PR #7483, opened 2020-04-27
Delete potted plant in sec, oh and also add security ship

Arokha:
1) Deletes a potted plant in a sec hallway
2) Moves HoS office north like 2 turfs to make room for a dock hallway.
3) Adds dock hallway.
4) Adds spaceship for security.


KasparoVy:
The departmental armada is assembling!

PR #7488, opened 2020-04-27
Hydroponics and virology machine icons

Arokha:
There's crossover, I guess, for some reason!


Heroman3003:
The lights on new tray are way too small imo.

PR #7490, opened 2020-04-28
Adjusts Side-Facing Sprites of Taur Aprons

KasparoVy:
For some reason they were shifted one pixel back in the upper chest area compared to the aprons on non-taurs and it looks wack by comparison. Slight adjustment to make them fit better (like the normal sprites).


KasparoVy:
Huh, I wonder where the icon diff check went.

KasparoVy:
Huh, I wonder where the icon diff check went

PR #7491, opened 2020-04-28
Vespa changes, expansion, etc

AmayaWah:
Re-designs engineering. Uses the RUST engine now, with the possibility of going with another engine.
Fixes issues with access. All of them, in fact.
Moves things about!!!
Expands sides of Vespa to include telecomms and probably other things soon:tm:
Most of atmos is now routed through maintenance to prevent issues. 
Backup atmos can be found where main atmos used to be - in the back of the north wing.
MORE coming soon :tm:


Arokha:
This PR looks crazy now and also this isn't the place to just park your PRs forever. >:V

PR #7498, opened 2020-04-29
Adds UAVs

Arokha:
Adds UAVs, buildable via mechfab
Adds UAV Control software for modular computers (it's only useful on tablets and laptops, given that to pair to a UAV you have to hit the computer on it)

UAVs work anywhere that you have NTnet relay access between you and the UAV. How fast you can move as the UAV depends on signal strength. If there's no NTnet access, you can use the UAV locally, within 30 turfs, but beyond that it will disconnect.

They can see emotes/speech and relay them to the viewers. I'll probably add a way to talk back? Maybe? That might be good.

It will also idle out if nobody connects within a few minutes to protect it from running out of battery. If it runs out of battery while flying, it will take damage.

You can pack it into a smaller size for a backpack, but it's still the largest size item that will fit, occupying a sizable chunk of them. It also takes a bit to pack/unpack them.


SplinterGP:
UAV INBOUND,

Also, maybe coulld add Counter UAV's ? or something that dozes off nearby UAV's ?

-make teshari able to ride the UAV's-

TheFurryFeline:
Yay, UAV/MALP time for Gateways! Good old gate travelling. :honk: [Insert Stargate themesong]

PR #7513, opened 2020-04-29
[MIRROR] Rewrite examine() to pass a list around

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/7038
--------------------
This is how /tg/ does it. It's faster than each object having it's own separate to_chat messages that it enqueues which wastes time sending that information to the client in separate chunks, and makes SSchat do more work.

I tried to not change anything that would be considered 'balance'. There are SO MANY 'interesting' choices about 'how far away' is too far away to see something. Why can you see the contents of a jetpack at any distance, but the contents of an oxygen tank only when holding it? Why can you see the ammo in an RCD and RSF at different ranges? Why do air tanks play a sound at you in their examine()? SO MANY QUESTIONS.

Regardless, the only changes I made aside from the necessary ones for this were grammar, replacing <font> with <span> that I noticed, and any time an object repeatedly said it's name or reused it's icon a bunch, I replaced with "It". So there's no "That's a box. It's for holding things. The box looks damaged".


KasparoVy:
Yeah, regarding the air tank comment I wished it played a sound when you reached a threshold without having to examine

PR #7552, opened 2020-05-01
Vore refactor V2

ShadowLarkens:
Digestion functionality is now handled by datums.

This should theoretically change no behavior in vore, other than the fact that I sorta-kinda-half fixed TF species, but the whole TF system seems more than a little borked.


code/modules/vore/eating/bellymodes_datum_vr.dm
```diff
@@ -0,0 +1,247 @@
+GLOBAL_LIST_INIT(digest_modes, list())
+
+/datum/digest_mode
+	var/id = DM_HOLD
+	var/noise_chance = 0
+
+/datum/digest_mode/proc/process_mob(obj/belly/B, mob/living/L)
```
Leshana, line 7: It would be good to document the allowed return values for this proc.

KasparoVy:
Wow, badass! Great job

code/modules/vore/eating/bellymodes_datum_vr.dm
ShadowLarkens: yeah true

PR #7570, opened 2020-05-01
One more tweak to Bay catwalks

Arokha:
Objects and people already on a plated catwalk, when opened, will fall through immediately rather than on their next move


KasparoVy:
RIP loony tunes physics

PR #7572, opened 2020-05-01
[MIRROR] Refactor AI into two subsystems

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/7053
--------------------
Splits AI into SSai and SSaifast. Adds management in the ai_holder datum to add/remove itself from these subsystems based on stance ('stance' is the state of the AI finite state machine, if you weren't aware).

Combat and 'fast-needing' things add it to SSaifast which processes every 0.5 seconds. Idle removes it from SSaifast. Dying removes it from processing altogether. Being revived puts it back in processing.

Removed the found() proc on ai_holders for specialized targeting. It was unused, and just wasting time. Someone can re-implement it on their special AI datum subtypes if they want to implement weird targeting. Reworded/rewrote target sorting functions.

Altered a couple of behaviors: mobs consider shooting a target they found more important than going home or following their leader (to reflect the general idea I have that combat behavior > non combat behavior in terms of priority). Mobs now consider if they should wander, follow, or return home in their strategic process (SSai process), rather than in their every-0.5s process (SSaifast process).

In my testing on our server, with about 300-350 mobs, the subsystem(s) consume about 25% of their original CPU time.


Arokha:
0diff

PR #7586, opened 2020-05-02
Curashuttle adjustments, fixes and changes.

Natje0:
Removes NTNet relay as that was the 20k power drain source. RIP Emailing.
Fixes off-set errors, hopefully.
Adds configurable IDs for contractors.
Adds some minor absolute basics equipment for contractors.
Changes engineering room, more space. Hangar gets a bit smaller.
Adds inflatables. How could I forget such an important thing for rescue ops on a rescue shuttle?
Makes the shuttle able to refuel much faster via piping straight to the airlock.


Natje0:
Probably a couple changes/additions I forgot to name in the desc, it's 6 AM and I didn't sleep yet

PR #7593, opened 2020-05-02
Should temporarily fix wearable circuits.

Ldip999:
Assemblies were defined as clothing only, I think, and not as circuits.
This is a very lazy, ugly, and simple fix, but this has been broken for quite a while now, so it's better than nothing.


Ldip999:
Sent to Polaris. Sorry for bothering.

PR #7610, opened 2020-05-03
Adjusts Talons Marking to Account for Wierd Robo-feet

KasparoVy:
Adds the dew-claw digit to the feet sprites to account for different cyborg feet/leg sprites that prevent it from being visible.

Also fills out some missing east-west facing sprites so the marking still works/makes sense when you're missing bodyparts.


KasparoVy:
Woop, problems with my branch.

KasparoVy:
Cheers

PR #7611, opened 2020-05-03
Disables sif mob spawning on outdoors grass turfs

Verkister:
No more 500 suffocated automatically spawned non-phoron mobs clogging the moblists around tether.

### DIFF
--- a/code/game/turfs/simulated/outdoors/grass.dm
+++ b/code/game/turfs/simulated/outdoors/grass.dm
@@ -39,7 +39,7 @@ var/list/grass_types = list(
 	grass_chance = 5
 	var/tree_chance = 2
 
-	animal_chance = 0.5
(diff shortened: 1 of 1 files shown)


code/game/turfs/simulated/outdoors/grass.dm
```diff
@@ -39,7 +39,7 @@ var/list/grass_types = list(
 	grass_chance = 5
 	var/tree_chance = 2
 
-	animal_chance = 0.5
+	animal_chance = 0 //VOREStation Edit
```
Heroman3003, line 42: Could probably make that change on /virgo3b subtype of grass (assuming its not created via the define thing) or in _vr file otherwise.

Verkister: looks like the subtype doesn't even exist in the code.
kuva

PR #7614, opened 2020-05-03
Port SSnightshift from Paradise

Arokha:
Dims station lights when it's night, brightens when day. Can be manually toggled on APC panels, and there's a light subtype that ignores nightshift mode if you want to map those in. Only affects station levels, and ignores shuttles (presumably they have their own power management system or something).

Night is considered 25% of the day after midnight or 25% of the day before midnight (so 50% of the day). On Virgo 3, that's 2:15 to 0:45 (I'm terrible at math but I think that's right).

When you set an alert level other than green, it disables nightshift mode.
dreamseeker_2020-05-03_17-46-15

dreamseeker_2020-05-03_17-46-08

dreamseeker_2020-05-03_17-41-54
dreamseeker_2020-05-03_17-43-13
dreamseeker_2020-05-03_17-46-01
dreamseeker_2020-05-03_17-45-39

### DIFF
--- a/code/__defines/lighting.dm
+++ b/code/__defines/lighting.dm
@@ -82,6 +82,7 @@
 #define LIGHT_COLOR_INCANDESCENT_TUBE "#FFFEB8"
 #define LIGHT_COLOR_INCANDESCENT_BULB "#FFDDBB"
 #define LIGHT_COLOR_INCANDESCENT_FLASHLIGHT "#FFCC66"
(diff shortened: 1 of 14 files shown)


KasparoVy:
The Shadekin'll love this one!

Arokha:
> The Shadekin'll love this one!

Oh yeah, good point :0

PR #7619, opened 2020-05-04
Aurora's floof, I mean fluff item

storesund98:
Creates Aurora's fluff item
https://forum.vore-station.net/viewtopic.php?f=26&t=1749

### DIFF
--- a/code/modules/vore/fluffstuff/custom_items_vr.dm
+++ b/code/modules/vore/fluffstuff/custom_items_vr.dm
@@ -1821,3 +1821,8 @@
 
 	icon_override = 'icons/vore/custom_clothes_vr.dmi'
 	item_state = "blindshades_mob"
+
+//Storesund97 - Aurora
+/obj/item/clothing/accessory/solgov/department/security/aurora
+	name = "Old security insignia"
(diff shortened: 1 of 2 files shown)


code/modules/vore/fluffstuff/custom_items_vr.dm
```diff
@@ -1821,3 +1821,9 @@
 
 	icon_override = 'icons/vore/custom_clothes_vr.dmi'
 	item_state = "blindshades_mob"
+
+//Storesund97 - Aurora
+/obj/item/clothing/accessory/solgov/department/security
```
Novacat, line 1826: this needs an /aurora at the end, otherwise it will overwrite the standard item.

code/modules/vore/fluffstuff/custom_items_vr.dm
```diff
@@ -1821,3 +1821,9 @@
...
+	name = "Old security insignia"
+	desc = "Insignia denoting assignment to the security department. These fit Expeditionary Corps uniforms. This one seems to be from the 2100s..."
+	color = "#bf0000"
```
Novacat, line 1829: color = "#bf0000" can be deleted, since it inherits it from the parent item /obj/item/clothing/accessory/solgov/department/security

PR #7633, opened 2020-05-05
Add prefs for Be Morph, Be Maintpred, Be Lost Drone

Arokha:
Allows the 'never for this round' button to work, and for them to be disabled entirely if you don't want to be asked again.

### DIFF
--- a/code/__defines/gamemode.dm
+++ b/code/__defines/gamemode.dm
@@ -37,6 +37,11 @@
 #define BE_MUTINEER   0x2000
 #define BE_LOYALIST   0x4000
 #define BE_PAI        0x8000
+//VOREStation Add
+#define BE_LOSTDRONE  0x10000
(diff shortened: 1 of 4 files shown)


Heroman3003:
I personally would use singular preference for onstation spawns (morph and maint pred) since two as a whole are very similar in nature of event, but it's good to see those added. Though also mentioning those on wiki would be important, since other than pAI and positronic, none of the other 'antag-looking' preferences are used.

Arokha:
Eh, I think when more variety is added to maint pred or maybe more nuance to morph it'll be different enough to make it worth it. I don't see much harm in having them separate until then. Also update the wiki, then, nerd!

TheFurryFeline:
Corgi Runes and Crystal Sword also give prompts and ignore the "Never again for this round" option being selected.

Arokha:
u fix

Issue #7128, opened 2020-04-03
Protein shake name applies properly only after there's none left


Heroman3003:
#### Brief description of the bug
Protein shakes from fitness vendors have different name from regular shaker: protein shake. But that name is only visible when shaker itself is empty, being rather redundant.

#### What you expected to happen
Have it be called 'protein shake' at all times

#### What actually happened
Its 'shaker of something' when it actually has shake in it and 'protein shake' only when its empty.

#### Steps to reproduce
- Get yourself a protein shake
- Check the name
- Drink the whole thing
- Check the name again

#### Code Revision
- c76760aaf613aa7440c9d673686b99c72607c0bb

PR #7637, opened 2020-05-05
Fix protein shaker name

Arokha:
Fixes https://github.com/VOREStation/VOREStation/issues/7128

### DIFF
--- a/code/modules/food/drinkingglass/shaker.dm
+++ b/code/modules/food/drinkingglass/shaker.dm
@@ -12,7 +12,7 @@
 	var/lid_color = "black"
 
 /obj/item/weapon/reagent_containers/food/drinks/glass2/fitnessflask/Initialize()
-	..()
+	. = ..()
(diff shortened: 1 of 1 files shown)


code/modules/food/drinkingglass/shaker.dm
```diff
@@ -12,19 +12,22 @@
...
-	..()
+	. = ..()
 	lid_color = pick("black", "red", "blue")
 	update_icon()
 
 /obj/item/weapon/reagent_containers/food/drinks/glass2/fitnessflask/update_icon()
 	..()
+	// And now set half the stuff back because our name shouldn't change
+	name = initial(name)
+	desc = initial(desc)
```
Heroman3003, line 23: IMO, this should just be proteinshake thing, since normal shakers that start out empty changing their name seems like reasonable thing?

Arokha: I pushed a fix for this, but
image

PR #7642, opened 2020-05-05
Reword grab facesits

Arokha:
Let's be real here. It's facesits.


Heroman3003:
The entire reason it was worded like that in the first place is to prevent that from being overly obvious in wording, in case it happens on accident while grabbing someone or in wide public.

PR #7644, opened 2020-05-05
Adds Natje's Curabitur Cannon

TeleTubby2:
I forgot how much gun code sucked.

this is a multi-purpose (mostly healing) beam gun. similar to the NSFW/ML3M, but uses single "macrobatterys" over cell mags. these cells hold more shots, and recharge on their own.

Balancing etc can be changed easily. a review would be nice, as it's been literal eons since I've worked with gun code.


code/modules/projectiles/guns/energy/cell_loaded_vr/multi_cannon.dm
```diff
@@ -0,0 +1,69 @@
...
+					if(81 to INFINITY)//gotta cover any admemes/other ways to get above max charge here.
+						istate = "healcannon_100"
+			else
+				indicator_colour = null			
+
+	var/image/x = image(icon = 'icons/vore/custom_guns_vr.dmi', icon_state = istate)
+	x.layer = FLOAT_LAYER
+	x.plane = FLOAT_PLANE
+	x.color = indicator_colour
+	overlays += x
```
Arokha, line 57:
```suggestion
	var/image/x = image(icon = icon, icon_state = istate)
	x.color = indicator_colour
	add_overlay(x)
```
You can cut this down to this, and use SSoverlay to add the overlay (less lag-inducing, because it gets queued rather than forcing it to do it right then), as long as you implement the cut_overlays() change too.

code/modules/projectiles/guns/energy/cell_loaded_vr/multi_cannon.dm
```diff
@@ -0,0 +1,69 @@
...
+
+	caliber = "macrobat"
+
+
+	icon_override = 'icons/vore/custom_guns_vr.dmi'
+	item_state = "healcannon"
+
+/obj/item/weapon/gun/projectile/multi_cannon/update_icon()
+	. = ..()
+	overlays.Cut()
```
Arokha, line 31:
```suggestion
	cut_overlays()
```

So you can use SSoverlays to cut the overlays rather than asking byond to do it directly.

code/modules/projectiles/guns/energy/cell_loaded_vr/multi_cannon.dm
```diff
@@ -0,0 +1,69 @@
...
+					if(21 to 40)
+						istate = "healcannon_40"
+					if(41 to 60)
+						istate = "healcannon_60"
+					if(61 to 80)
+						istate = "healcannon_80"
+					if(81 to INFINITY)//gotta cover any admemes/other ways to get above max charge here.
+						istate = "healcannon_100"
+			else
+				indicator_colour = null
```
Arokha, line 51:
```suggestion
	if(istype(chambered,/obj/item/ammo_casing/macrobattery)) //should never not happen. but. you never, never know with this damn, cursed game.
		var/obj/item/ammo_casing/macrobattery/bat = chambered
		indicator_colour = bat.bat_colour
		if(bat.charge)
			istate = "healcannon_20"
			var/percent_charged = round((bat.charge/bat.max_charge)*100)
			switch(percent_charged)
				if(21 to 40)
					istate = "healcannon_40"
				if(41 to 60)
					istate = "healcannon_60"
				if(61 to 80)
					istate = "healcannon_80"
				if(81 to INFINITY)//gotta cover any admemes/other ways to get above max charge here.
					istate = "healcannon_100"
		else
			indicator_colour = null			
```
You can remove the `if(chambered)` because `istype` is safe to use nulls with. It will return false if there's no `chambered`, which is what you'd expect and will work fine. Then you can unindent this whole thing by one.

code/modules/projectiles/guns/energy/cell_loaded_vr/multi_cannon_cells.dm
```diff
@@ -0,0 +1,102 @@
...
+/obj/item/ammo_casing/macrobattery/proc/recharge()
+	if(charge < max_charge)
+		charge ++
+		if(!BB)
+			BB = new projectile_type
+	if(charge >= max_charge)
+		STOP_PROCESSING(SSobj, src)
+	if(istype(loc,/obj/item/weapon/gun/projectile/multi_cannon))
+		var/obj/item/weapon/gun/projectile/multi_cannon/gun = loc
+		gun.update_icon()
```
Arokha, line 56:
```suggestion
	if(loc)
		loc.update_icon()	
```
Here's how normal cells do it. You don't actually need to cast it as anything, because we defined `update_icon()` on atom itself, and that's what `loc` is in byond, so no need to typecheck or recast.

code/modules/projectiles/guns/energy/cell_loaded_vr/multi_cannon_cells.dm
```diff
@@ -0,0 +1,102 @@
...
+		START_PROCESSING(SSobj, src)
+		update_icon()
+		. = BB
+		//alright, the below seems jank. it IS jank, but for whatever reason I can't reuse BB. big bad
+		BB = null
+		BB = new projectile_type
+		return 
+	else
+		BB = null
+		return null
```
Arokha, line 45: If, for whatever reason (trying to fire with the battery uncharged?), you get here with 0 charge, it will not add itself to processing and start recharging. It's also not great that it tries to add itself to SSobj every time it's fired, but there's a safety check in that macro that will avoid readding it every time, though it will still have to perform said check.

TeleTubby2: Yeah, I can't see that check being resource intensive enough for it to ever wind up being an issue, especially with the inherent ratelimiting clickdelays give.
and I can't see a way for the cells to ever get fired at zero charge *without* them having been already put in the processing subsystem. Maybe varediting them to zero when they're full, and not in the subsystem would do it, but that's a pretty niche case.

code/modules/projectiles/guns/energy/cell_loaded_vr/multi_cannon_cells.dm
TeleTubby2: The reason I'm typechecking is so it doesn't constantly update whatever it's loc is when charging. I'm sure an extra update_icon call isn't the end of the world, but better to avoid it when possible imo

Arokha: Every cell already does it, sadly. I doubt it'd make that much of a difference.

code/modules/projectiles/guns/energy/cell_loaded_vr/multi_cannon_cells.dm
Arokha: Or starting empty.

TeleTubby2: Not possible, thankfully! it's charge value is set to max_charge right off the bat in Initialize.

Arokha: Ye, but there could be subtypes in the future that start empty! Plenty of things in the game are like that, like cells and such. I could definitely see if these were producible in game, they'd startp empty like cells ~~are supposed to, and sometimes do~~

TeleTubby2: That's fair, though I think the fluff Natje's got for them with the Precursor stuff would effectively make them impossible to have empty, though I'll change it either way.

Any suggestions for fixing this one? having START_PROCESSING at the start of expend seems kinda sloppy. I could also probably throw it in Initialize.

PR #7654, opened 2020-05-05
Prevent multiple simultaneous stack usage

Arokha:
So you can't simultaneously build 5 chairs with the same stack of metal, or simultaneously heal an entire person with the same stack of kits.


Arokha:
Got a better idea

PR #7659, opened 2020-05-05
[MIRROR] Speech Bubble Adjustments

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/7065
--------------------
- Organic looking synths have a normal speech bubble, until they pass their lookssynthetic check (usually due to damage)
- Organics with a synthetic voicebox sound synthetic.


Arokha:
0diff

PR #7666, opened 2020-05-06
Replace/Add lots of sounds from VerySoft

Arokha:
Lots of neat AI announcement voices.


KasparoVy:
Really well done, they sound super legit. Just something right about AI vox having an accent of some sort.

Rykka-Stormheart:
I love this!

Arokha:
Could add a preference, but I don't really feel there's much change from now. 90% of these already have vox announcements that use the old voice.

Arokha:
We'll do that if people are grumpy gooses.

PR #7675, opened 2020-05-06
Add a better nightshift control

Little-119:
Adds a control to APCs for disabling and enabling nightshift mode in a way that isn't changed by the nightshift system.
image

Also fixes the cooldown while I'm here.

I've _probably_ committed programming sins here, I'm not familiar with DM and this is my first non-map PR in a long time. Sorry in advance.


Little-119:
*windows 3.1 startup sound*

PR #7677, opened 2020-05-07
Add Goon-style pushing of objects

Arokha:
Also, pressing Q to drop what you have in your hand if you have nothing in your hand will stop pulling.
Also, ctrl-clicking any turf will stop pulling.


KasparoVy:
Big QOL for Cargo & Mining!

PR #7686, opened 2020-05-07
[MIRROR] [PORT] Use can_atmos_pass to reduce proc-calls in c_airblock()

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/7091
--------------------
### Fix existing code
- CanZASPass is supposed to return boolean.  Nobody noticed this bug because `ATMOS_PASS_YES` and `ATMOS_PASS_NO` happen to be defined as 1 and 0.  But that's not a good assumption to make, so lets fix it!
- Remove /turf/var/can_atmos_pass as it is unused (and redundant with /turf/var/blocks_air anyway.

### Optimize away proc-calls in c_airblock()
- Add additional can_atmos_pass value `ATMOS_PASS_PROC` which indicates custom behavior requiring calling the CanZASPass proc.
  - The benefit being for the other three values we DON'T need to call CanZASPass at all!  We already know the behavior without the overhead of a proc call.
  - Obviously any atom with `can_atmos_pass = ATMOS_PASS_PROC` must not call `..()` in CanZASPass() since the default behavior would be to (recursively) call CanZASPass()
  - This required re-numbering the constants, so I also fixed all code that assumed particular values for the constants.
- Switched all types which overrode CanZASPass with custom logic to be `can_atmos_pass = ATMOS_PASS_PROC`
- **The payoff:**  Changed `/turf/c_airblock()` to skip calling `/atom/movable/c_airblock()` for the three can_atmos_pass values that don't require calling the proc.

Reduces init time for SSair by 20% on my local test machine.
Port of VOREStation/VOREStation#7536


Arokha:
0diff

Issue #7693, opened 2020-05-07
[BUG] Reset Slot still does not work


Rykka-Stormheart:
#### Brief description of the bug

Clicking reset slot does nothing.

#### What you expected to happen

Click Reset Slot, go back to having "Character'x'"

#### What actually happened

Hit "reset slot", nothing happens.

#### Steps to reproduce
- Go to "Character Setup"
- Choose a slot
- Hit "Reset Slot"
- Nothing happened!

PR #7694, opened 2020-05-07
Fix for Resetting Character Slot to Default

Rykka-Stormheart:
Resets the character slot to default.
Fixes https://github.com/VOREStation/VOREStation/issues/7693

Credit goes to @ShadowLarkens  for helping me figure it out and implementing the solution.

### DIFF
--- a/code/modules/client/preferences_savefile.dm
+++ b/code/modules/client/preferences_savefile.dm
@@ -81,6 +81,8 @@
 	else
 		player_setup.load_character(S)
 		S.cd = "/character[default_slot]"
+		player_setup.save_character(S)
+		sanitize_preferences()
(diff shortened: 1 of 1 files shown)


ShadowLarkens:
As a note, this is because Polaris is lacking the changes made by https://github.com/Baystation12/Baystation12/pull/11599 which allow the reset slot button to reset all of the preference values without actually updating the savefile.

https://github.com/VOREStation/VOREStation/blob/4e25e87ab762b8aa6f53d834f41a88a0ed1ade1a/code/modules/client/preferences_savefile.dm#L78-L85

Also, L85 here is kinda silly. It doesn't do anything, ever.

Issue #7704, opened 2020-05-08
Multikeying detection message is way too confrontational


Screemonster:
#### Brief description of the bug
![image](https://user-images.githubusercontent.com/14942479/81359746-ebfccf80-90d1-11ea-84ac-c6937f3eccd5.png)

Several new players have received this message and been terrified out of their wits by the computer screaming at them to "log out NOW or risk being banned!"

#### What you expected to happen
An advisory notice to the effect of "You appear to have logged in with another key this round, which is not permitted. Please contact an administrator if you believe this message to be in error."

#### What actually happened
Above, heavy-handed ban-threatening message with all-caps "NOW" and exclamation marks.

#### Steps to reproduce
- log in
- log in with another key

#### Code Revision
- (Found using the "Show Server Revision" verb under the OOC tab.)
 
#### Anything else you may wish to add:
- (Location if it's a mapping issue, screenshots, sprites, etc.)

PR #7705, opened 2020-05-08
Make multikey warning less rude

Arokha:
Fixes https://github.com/VOREStation/VOREStation/issues/7704

### DIFF
--- a/code/modules/mob/login.dm
+++ b/code/modules/mob/login.dm
@@ -15,7 +15,7 @@
 					if(matches)	matches += " and "
 					matches += "ID ([client.computer_id])"
 					if(!config.disable_cid_warn_popup)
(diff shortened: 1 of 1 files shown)


code/modules/mob/login.dm
```diff
@@ -15,7 +15,7 @@
 					if(matches)	matches += " and "
 					matches += "ID ([client.computer_id])"
 					if(!config.disable_cid_warn_popup)
-						spawn() alert("You have logged in already with another key this round, please log out of this one NOW or risk being banned!")
+						spawn() alert("You appear to have logged in with another key this round, which is not permitted. Please contact an administrator if you believe this message to be in error.")
```
TheFurryFeline, line 18: Couldn't this instead be a to_chat warning?

Arokha: Might get missed

PR #7722, opened 2020-05-08
Another bunch of simplemob fixes

Verkister:
-Fixes the issue with mob AI not getting properly disabled under player control.
-Fixes mob eye overlays not respecting the mob's scaling setting.
-Fixes the movement cooldown on logged in ridable mobs.
-Fixes direwolf not getting belly icon when resting.
-Makes the mob icon fullness check account for the mob's own size as well.

### DIFF
--- a/code/controllers/subsystems/ai.dm
+++ b/code/controllers/subsystems/ai.dm
@@ -26,8 +26,6 @@ SUBSYSTEM_DEF(ai)
 		--currentrun.len
 		if(!A || QDELETED(A) || A.busy) // Doesn't exist or won't exist soon or not doing it this tick
 			continue
-		if(A.holder.client && !A.autopilot)
-			continue
 		A.handle_strategicals()
 
 		if(MC_TICK_CHECK)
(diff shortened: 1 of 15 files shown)


code/controllers/subsystems/ai.dm
```diff
@@ -26,6 +26,8 @@ SUBSYSTEM_DEF(ai)
 		--currentrun.len
 		if(!A || QDELETED(A) || A.busy) // Doesn't exist or won't exist soon or not doing it this tick
 			continue
+		if(A.holder.client && !A.autopilot)
+			continue
```
Arokha, line 30: This triple-dereference is probably not a great idea for performance (same with SSaifast).

It would probably be better to do something like...
- Mob login() checks if there's an AI holder
- If there's an AI holder, and !holder.autopilot, it removes the holder from both AI subsystems with `manage_processing(0)`
- Mob logout() checks if there's an AI holder
- If there's an AI holder, cheat a bit and call holder_stat_change on it, pass it `src, stat, stat`, that should readd it to the subsystem BUT ONLY IF it's still alive, which is how it should be handled for mobs without clients or anything

Verkister: I tried the manage_processing thing with my first attempt before going for this approach and it didn't work.

Arokha: One note: This has the downside of giving AI control back to mobs if you're an admin and aghost, because your client leaves the mob (though your solution also has that problem).

Arokha:
Pls don't forget to send to Polaris soon!

PR #7723, opened 2020-05-08
Add department playtime requirement config, code

Arokha:
Adds 60 or 20 hour requirements for heads, subheads of playtime in their departments.
It's a config option.
It uses SQL, which stands for Structured Query Language.

It's prepopulated with your PTO hours to avoid people needing to start from scratch everywhere.

### DIFF
--- a/code/game/jobs/job/security_vr.dm
+++ b/code/game/jobs/job/security_vr.dm
@@ -1,6 +1,7 @@
 /datum/job/hos
 	disallow_jobhop = TRUE
 	pto_type = PTO_SECURITY
+	dept_time_required = 60
 
 	access = list(access_security, access_eva, access_sec_doors, access_brig, access_armory,
 						access_forensics_lockers, access_morgue, access_maint_tunnels, access_all_personal_lockers,
@@ -13,6 +14,7 @@
 
 /datum/job/warden
 	pto_type = PTO_SECURITY
+	dept_time_required = 20
 
 /datum/job/detective
 	pto_type = PTO_SECURITY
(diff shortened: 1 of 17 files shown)


Heroman3003:
Those numbers sound like an extreme overkill tbh. Overall its a good thing but why so big numbers?

Arokha:
> Those numbers sound like an extreme overkill tbh. Overall its a good thing but why so big numbers?

Decided on during staff meeting.

code/game/jobs/job/security_vr.dm
```diff
@@ -13,6 +14,7 @@
 
 /datum/job/warden
 	pto_type = PTO_SECURITY
+	dept_time_required = 20
```
Heroman3003, line 17: Warden's not a sub-head though?

Arokha: Some jobs are unique in departments like QM and PF. They received the same treatment.

Rykka-Stormheart: Warden *is* a sub-head, underneath HoS, technically. Iirc.

Screemonster:
> Those numbers sound like an extreme overkill tbh. Overall its a good thing but why so big numbers?

it's like 10 rounds, you can rack up that much time easily in the existing timelock on a head job

YuskoTheCoyote:
One should also keep in mind that players with PTO will start out with the same amount of hours that are on their PTO. So if you have the full 50 hours of PTO, you will start out with having fulfilled 50 hours of the 60 hour requirement. It would mean you will just need to play 2ish rounds in a department to be able to play the department head. 

The intention behind this is to make sure that people picking Head of Staff roles have some experience with working in their department before they also get the power to lead the people in their department.

Rykka-Stormheart:
I really really like this.
Fantastic way to ensure people who're in charge of departments actually know what they're doing (hopefully)!

PR #7725, opened 2020-05-09
Add Romantical Whale's "Shiny Clothing"

Little-119:
Adds the "shiny clothing" made by Romantical Whale#8694. All creative direction in this is by them, I just coded in the stuff.

Apologies in advance for any coding sins and stuff committed in this.


Heroman3003:
Should probably replace latex stuff in bdsm dungeon on surface 2 with those.

Screemonster:
See, when these were first mooted, I thought people were just talking about adding shiny jumpsuits, not outright straight-up drone hoods and ballgags

I guess they're fine in private scenes but I'm dreading people wheeling them into the bar for their public humiliation fetish regardless of bystander consent

RomanticalWhale:
Well, I think that issue with public stuff isn't that bad, bc people around here is pretty much adequate and I've never seen or heard about public humiliation with handcuffs and stripping people down. And hood is pretty much removable and suit can be hidden under something like trenchcoat and normal boots. And gags were added in a last second only because I've been asked by three or four people and they wasn't intended to be included tbh

Screemonster:
Just 'cause three or four people ask for them doesn't mean they'd be a good thing to add, if you're familiar with what happened on citadel when people wanted jizz added that splatters everywhere and the first thing that happened was someone loaded it into a smoke machine and coated every tile of the corridor with it.

And the constant recurring issues with chefs serving deep-fried legs on the kitchen counter then complaining when security arrest them.

Whether they can be hidden or not, my main concern is that I don't particularly trust people to hide them and having people wask around the station, a workplace, in undeniably-bdsm-gear that cannot be handwaved away as anything else, shifts the theme of the place from "workplace where people happen to do lewd things" to "fetish dungeon where people sometimes work".

RomanticalWhale:
Well, but we have latex maid dress and i've seen only visitors in it even keeping in mind that it is accessable in loadout. And i'm thinking that people isn't going to walk in that stuff on station

I want to make things clear: this clothes not supposed to be shown anywhere but cargo, so it will be harder to abtain. I never had an idea to make this stuff too accessable. It is supposed to work close to recreational restraints in my head.

Little-119:
I'm trying to add species variations for the clothes, but I'm not sure how precisely to do that. Apparently the answer isn't to adding `sprite_sheets` with a list of alternative sprite sheets for different species. I'm feeling less keen on stubbornly looking through files and figuring things out as I was about six days ago.

Also, Romantical Whale wants another admin to give their opinion on this.

MrSign:
In such an event that someone rocks up to the bar in this. They can always get themselves booted into the Brig for disturbing the peace by wearing such a risqué outfit. Aside from that, I don't see any issues in having them be found in the dungeon.

Little-119:
This should be ready, I think. Like I said, Romantical Whale wanted to add species sprite variations but how to do that seems to be a mystery. This PR should be otherwise ready.

lbnesquik:
I am personally fine with those clothes if they are kept tame, but i have to mention the surest way to avoid people walking around with those in public is to not add them in the first place. 
It is fairly common for people to be doing vore scenes or more unusual, unwilling things in public, so those clothes are 100% going to be walked around with. At least IMO.

Little-119:
I shall code them so that, if they are worn outside of dorms, the wearer will instantly gib

RomanticalWhale:
Well, for such things we have security on station. This violates in-game laws and this can lead to arrest, so I don't see any problems with clothing. If someone goes outside in them - they go into Brig. I'm fully agree with @MrSign 
> 
> 
> In such an event that someone rocks up to the bar in this. They can always get themselves booted into the Brig for disturbing the peace by wearing such a risqué outfit. Aside from that, I don't see any issues in having them be found in the dungeon.

nerdass:
we already have ball gags, fuzzy handcuffs, latex maid outfits, fuzzy legcuffs, blindfolds, and all kinds of other stuff and people aren't dumb in the bar with it, why would this be any different.

TheFurryFeline:
Huh, was expecting my PR to make a conflict with this one. Apparently, it won't. Yay.

SilencedMP5A5:
>if you're familiar with what happened on citadel when people wanted jizz added that splatters everywhere and the first thing that happened was someone loaded it into a smoke machine and coated every tile of the corridor with it.

image

Zeldazackman:
Those pictures were well and before the smoke machines were even a thing though serds, hell that picture is back from 2014-2015

MrSign:
We fucking adding these or what?! It's been months!!!

code/game/sound.dm
```diff
@@ -252,6 +252,9 @@
 					'sound/vore/sunesound/prey/death_04.ogg','sound/vore/sunesound/prey/death_05.ogg','sound/vore/sunesound/prey/death_06.ogg',
 					'sound/vore/sunesound/prey/death_07.ogg','sound/vore/sunesound/prey/death_08.ogg','sound/vore/sunesound/prey/death_09.ogg',
 					'sound/vore/sunesound/prey/death_10.ogg')
+
+			if("clop")
+				soundin = pick('sound/effects/clop1.ogg','sound/effects/clop2.ogg')
```
Novacat, line 257: This has to be removed.

code/modules/clothing/under/shiny_vr.dm
```diff
@@ -0,0 +1,41 @@
...
+			to_chat(usr, "You zip up a zipper on your [src].")
+		update_icon()
+		update_clothing_icon()
+
+/obj/item/clothing/under/shiny/catsuit/verb/adjust_zipper()
+	set category = "Object"
+	set name = "Adjust zipper"
+	set src in usr
+
+	adjust()
```
Novacat, line 28: Crotch zipper has to go as well.

JoanRisu:
If the appropriate changes have been made to our specifications, then this PR can move forward.

Little-119:
The requested changes were made. Then I added descriptions to things that were lacking them ~~which I should've done a while ago~~, then compressed everything into two commits while trying to resolve a merge conflict in a .dmi.

PR #7732, opened 2020-05-09
[Map] Onions and Layers

TheFurryFeline:
Changelog Notes:

- Fixes layers for shutters and blast doors throughout the main station.

- Also adding in removal of mining's old substation per a discussion from last night. Yet ANOTHER oversight I did with it, good job, old me. Substation replaced with couple maint loot stuff and filled in with rock/wall.

- Making the door leading to old Mining EVA public access. Will restrict to Mining + Medical if needed.

- Missing firelock added for door to Atmos Processing pipes.

### DIFF
--- a/maps/tether/tether-01-surface1.dmm
+++ b/maps/tether/tether-01-surface1.dmm
@@ -53,16 +53,14 @@
 	name = "south bump";
 	pixel_y = -28
 	},
-/obj/structure/cable/green{
-	d2 = 4;
-	icon_state = "0-4"
+/obj/structure/cable{
+	icon_state = "0-4";
+	d2 = 4
 	},
 /turf/simulated/floor/tiled/steel_dirty/virgo3b,
 /area/tether/surfacebase/mining_main/external)
 "aal" = (
-/obj/structure/cable/green{
-	d1 = 2;
-	d2 = 8;
+/obj/structure/cable{
 	icon_state = "2-8"
 	},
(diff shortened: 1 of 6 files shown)


Arokha:
I'm... confused about this layers fix for blast doors and stuff. Is this because they look weird in DM? They change their layer every time they open and close ingame. Whatever you set in DM is going to mean nothing once they're opened or closed once ingame.

TheFurryFeline:
> I'm... confused about this layers fix for blast doors and stuff. Is this because they look weird in DM? They change their layer every time they open and close ingame. Whatever you set in DM is going to mean nothing once they're opened or closed once ingame.

Take a look at how the layering happens on current live. The blast doors at the bridge, atmos entrance, and other places, layer on **top** of the door after it opens. This makes it hard to close again without precise clicking, from purrsonal experience, and can be extremely annoying.

Arokha:
I mean, it shouldn't fix that, because the blast doors set their layer every time they close/open. Anything you put in the map editor for layer will only last until it's used once.

Also blast doors aren't supposed to be close-able by clicking on them... they're controllable by remote...

TheFurryFeline:
That's not why this is being done. I've made the closed layer reduced to the lowest so it doesn't show up in its open state when someone's passing through the doorways. And I've tested it. It keeps the settings, or at least the sprites don't revert to its present state of being layered over an open airlock.

PR #7735, opened 2020-05-09
Reduce maximum audible range of sounds [DNM]

Rykka-Stormheart:
This is a global change - this reduces sounds to being audible at a MAXIMUM of 14 tiles from their source, unless the var is changed later. (I think).

Discuss if this is needed or not, please, and give all feedback, marking DNM until staff can weigh in.

### DIFF
--- a/code/game/sound.dm
+++ b/code/game/sound.dm
@@ -10,7 +10,7 @@
 
  	// Looping through the player list has the added bonus of working for mobs inside containers
 	var/sound/S = sound(get_sfx(soundin))
-	var/maxdistance = (world.view + extrarange) * 2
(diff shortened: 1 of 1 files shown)


Arokha:
For reference:
Currently, the default range for all sounds is `(world.view + optional extra) * 3` which in almost every case is 21 (the optional extra is extremely rarely if ever used, and world.view is 7).

Basically every server uses a similar formula, with other servers having:
TG: (world.view+optional extra) = 7 usually
Bay: (world.view+optional extra)*2 = 14 usually
Paradise: world.view+optional extra)*3 = 21 usually

PR #7769, opened 2020-05-11
Fix for edge tiles having noname turf applied

Rykka-Stormheart:
update_icon_edge() was responsible for generating the edges, which calls back to icon_state, and it looked for grass_sif0/1-edge, but this was not existing previously. This fixes that issue. 

Fixnowork. Trying to figure it out.


Rykka-Stormheart:
Nonfunctional, no idea wtf overlay keeps getting applied.

PR #7777, opened 2020-05-11
Fix turf icon issues

Arokha:
- Remove the 'edges' from a couple of turfs with none
- Redo how turf edges/corners are done altogether
- Move the 'put your edges on me' part of icon stuff to a central location, instead of just hanging around outdoor turfs


code/game/turfs/simulated/floor.dm
```diff
@@ -59,7 +64,7 @@
 		swap_decals()
 	flooring = newflooring
 	footstep_sounds = newflooring.footstep_sounds
-	update_icon(1)
+	//update_icon(1) // We'll do it in late initialize
```
Leshana, line 67: I should testy westy

PR #7781, opened 2020-05-11
Adds a copier to the library at player request

Arokha:


Arokha:
I'm not sure WHY it swapped the order of dir and icon_state, not that it matters. Maybe fastdmm2 prefers alphabetical. But either way since fastdmm2 is pretty good I'll leave it.

PR #7785, opened 2020-05-11
Allows Cyborgs and AI to earn PTO and departmental playtime

Heroman3003:
Title. Since PTO is half-OOC thing anyway (being account-wide rather than character-specific) I don't see any reason to not reward synthetic players, since they are just as much of contributors as any humanmob department member.

AI earns Civilian PTO
Borgs earn PTO based on module:
Standard, Clerical, Service, Janitor, Janihound, Brodog - Civilian PTO
Surgeon, Crisis, Medihound - Medical PTO
Engineering, Pupdozer - Engineering PTO
Security, Combat, K9, ERT - Security PTO
Research, Sci-borg - Science PTO
Mining, KMine - Cargo PTO

### DIFF
--- a/code/controllers/subsystems/persist_vr.dm
+++ b/code/controllers/subsystems/persist_vr.dm
@@ -45,6 +45,19 @@ SUBSYSTEM_DEF(persist)
 		var/department_earning = J.pto_type
 		clear_unused_pto(M)
 
+		// Determine special PTO types and convert properly
+		if(department_earning == PTO_AI)
+			department_earning = PTO_CIVILIAN
+		if(department_earning == PTO_CYBORG)
+			if(istype(M, /mob/living/silicon/robot))
+				var/mob/living/silicon/robot/C = M
+				if(C.module && C.module.pto_type)
+					department_earning = C.module.pto_type
+			if(department_earning == PTO_CYBORG)
+				if (MC_TICK_CHECK)
+					return
+				continue
+
 		// Update client whatever
 		var/client/C = M.client
 		var/wait_in_hours = wait / (1 HOUR)
--- a/code/controllers/subsystems/persist_vr.dm
+++ b/code/controllers/subsystems/persist_vr.dm
@@ -46,8 +46,6 @@ SUBSYSTEM_DEF(persist)
 		clear_unused_pto(M)
 
 		// Determine special PTO types and convert properly
-		if(department_earning == PTO_AI)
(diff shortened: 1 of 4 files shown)


code/controllers/subsystems/persist_vr.dm
```diff
@@ -45,6 +45,19 @@ SUBSYSTEM_DEF(persist)
 		var/department_earning = J.pto_type
 		clear_unused_pto(M)
 
+		// Determine special PTO types and convert properly
+		if(department_earning == PTO_AI)
+			department_earning = PTO_CIVILIAN
```
Arokha, line 50: Whhyyyy? Why not just make PTO_AI define as PTO_CIVILIAN instead of spending the time here to check it every persist? I get why the cyborg needs logic, but this one seems silly.

code/controllers/subsystems/persist_vr.dm
```diff
@@ -45,6 +45,19 @@ SUBSYSTEM_DEF(persist)
...
+		if(department_earning == PTO_CYBORG)
+			if(istype(M, /mob/living/silicon/robot))
+				var/mob/living/silicon/robot/C = M
+				if(C.module && C.module.pto_type)
+					department_earning = C.module.pto_type
```
Arokha, line 55:
```suggestion
			if(isrobot(M))
				var/mob/living/silicon/robot/C = M
				if(C?.module?.pto_type)
					department_earning = C.module.pto_type
```

Arokha:
Just FYI this will make it so characters borged will earn PTO in their original department rather than their new one. But maybe that's fine.

code/controllers/subsystems/persist_vr.dm
Heroman3003: Its mix of potentially wanting AI shells to play into this and give PTO of whatever current shell is (but afterwards realizing that only civilian shells are available anyway) and that it feels like it needs some form of special handling. But since this is what it ended up being, gonna delete that now.

Heroman3003:
> Just FYI this will make it so characters borged will earn PTO in their original department rather than their new one. But maybe that's fine.

I think that's fine too. It's same how being moved off-duty or to custom job by HoP still has you earning original PTO. Plus borging is extremely rare occurance regardless.

Arokha:
Ye I think it's probably fine. Just noting in case it ever comes up.

PR #7801, opened 2020-05-12
New Engineering Tool - RMS: Rapid Material Synthesizer

Holdelta:
The RMS is a device that mimics the engineering cyborg modules' ability to create basic materials involved in construction, specifically steel and glass.
image

It can store up to 1 MW of charge, all taken from power cells.  Much like the biogenerator, once the charge has been deposited into the device, it cannot be withdrawn back.

There are six modes to this device(A single use/sheet of each mode is 20 kW):
- Steel
- Glass
- Cloth
- Plastic
- Stone
- Random

Extra features of this device can be accessed by overcharging the device with a multitool, making each use of it cost 50 kW in exchange for other construction materials.


Novacat:
Taking power wirelessly directly from batteries seems rather odd. Is there a reason why it was not set up to slot a battery into it like most other devices and machines that use power?

code/game/objects/items/weapons/RMS_vr.dm
```diff
@@ -0,0 +1,231 @@
...
+
+/obj/item/weapon/rms/examine(mob/user)
+	. = ..()
+	. += display_resources()
+
+/obj/item/weapon/rms/proc/display_resources()
+	return "It currently holds [round(stored_charge/1000)]/[max_charge/1000] kW charge."
+
+/obj/item/weapon/rms/proc/drain_battery(user, battery)
+	var/obj/item/weapon/cell.C = battery
```
Arokha, line 61:
```suggestion
	var/obj/item/weapon/cell/C = battery
```

This isn't valid DM code. It compiles, but is not valid according to the language definition, and will break things that are strict about it, like mapdiffbot.

code/game/objects/items/weapons/RMS_vr.dm
```diff
@@ -0,0 +1,231 @@
...
+
+/obj/item/weapon/rms/proc/use_rms(atom/A, mob/living/user)
+	var/obj/product
+	if(!can_afford())
+		to_chat(user, "<span class='notice'>There is not enough charge to use this mode.</span>")
+		return
+	playsound(src.loc, 'sound/machines/click.ogg', 50, 1)
+	if(do_after(user, 5, target = A))
+		switch(mode_index)
+			if(1)
```
Arokha, line 108: These mode indexes should definitely be defines.

code/game/objects/items/weapons/RMS_vr.dm
```diff
@@ -0,0 +1,231 @@
...
+
+/obj/item/weapon/rms/attack_self(mob/user)
+	if(mode_index >= modes.len)
+		mode_index = 1
+	else
+		mode_index++
+
+	to_chat(user, span("notice", "Changed mode to '[modes[mode_index]]'."))
+	playsound(src.loc, 'sound/effects/pop.ogg', 50, 0)
+	return ..()
```
Arokha, line 219: Might want a radial menu like the RCD has.

code/game/objects/items/weapons/RMS_vr.dm
```diff
@@ -0,0 +1,231 @@
...
+				else
+					product = new /obj/item/stack/material/glass
+					consume_resources(charge_cost)
+			if(3)
+				switch(charge_stage)
+					if(4)
+						product = pick(50; new /obj/item/weapon/cell/secborg/empty,
+										20; new /obj/item/weapon/cell/high/empty,
+										15; new /obj/item/weapon/cell/super/empty,
+										10; new /obj/item/weapon/cell/hyper/empty,
```
Arokha, line 130: I dunnooooo about hyper cells, or even cells in general. I mean power is essentially infinite on the station most of the time, so giving someone the ability to repeatedly put a cell into a charger, then charge the RMS, then repeatedly create other cells without science being involved.

code/game/objects/items/weapons/RMS_vr.dm
```diff
@@ -0,0 +1,231 @@
...
+										4; new /obj/item/device/fbp_backup_cell,
+										1; new /obj/item/weapon/cell/infinite)
```
Arokha, line 132: MMMM. I dunnnoooooooo. I mean you can brute force this every shift without tons of issue.

code/game/objects/items/weapons/RMS_vr.dm
```diff
@@ -0,0 +1,231 @@
...
+
+/obj/item/weapon/pipe_dispenser/Destroy()
+	qdel_null(spark_system)
+	return ..()
+
+/obj/item/weapon/rms/update_icon()
+	charge_stage = round((stored_charge/max_charge)*4)
+	if(charge_stage >= 4)
+		charge_stage = 4
+	add_overlay("rms_charge[charge_stage]")
```
Arokha, line 51: Not sure how I feel about a mechanically-important thing like charge_stage being in update_icon().
Also this never removes the overlays, it just keeps adding more indefinitely.

code/game/objects/items/weapons/RMS_vr.dm
```diff
@@ -0,0 +1,231 @@
...
+		return
+	else
+		playsound(get_turf(src), 'sound/machines/click.ogg', 50, 1)
+		if(do_after(user, 2,target = C))
+			stored_charge += C.charge
+			C.charge = 0
+			C.update_icon()
+			to_chat(user, "<span class='notice'>You drain [C].</span>")
+			if(stored_charge > max_charge)
+				stored_charge = max_charge
```
Arokha, line 78: `stored_charge = CLAMP(stored_charge, 0, max_charge)`
And might want to do it above the update_icon call because that's mechanically important to the thing, but see my other comment about that

code/game/objects/items/weapons/RMS_vr.dm
Rykka-Stormheart: I gotta say I agree.

Holdelta:
> 
> 
> Taking power wirelessly directly from batteries seems rather odd. Is there a reason why it was not set up to slot a battery into it like most other devices and machines that use power?
@Novacat 
There is actually a decent reason to why I designed it this way.  I wanted to give Engineering an option to replenish their material stock from just doing basic construction/fixing, but I also didn't want to muscle in too hard on Cargo/Mining and the fact that they are the main resource providers.

Much like how the borg modules take forever to replenish their material stock while sitting in a charger, I wanted the general exchange rate of power vs materials made to be decently disproportionate to time spent gathering charge, but I also didn't want people to carry around a locker full of batteries just to produce like, 5 sheets of materials.  I decided to give it a huge charge bank since its a medium-sized tool(slightly bigger than a 50 stack of materials) to make up for the fact that a lot of power is needed for each function.

Also, I didn't make this a machine since I wanted the tool to be an active use device, and not just passively gain charge.  Also, I've had plenty of times during repairs where I would be 2 to 3 steel or glass sheets away from fixing something, and having this be a mobile option to make materials is a nice convenience.

code/game/objects/items/weapons/RMS_vr.dm
Holdelta: I will address this, and potentially other invalid code when I look over all the issues you've brought up, thanks.

code/game/objects/items/weapons/RMS_vr.dm
Holdelta: I will look into this.

code/game/objects/items/weapons/RMS_vr.dm
Holdelta: I would like to bring into question why Science is perceived to be in charge of "power" when Engineering is the one that provides it.  I mostly made this feature since I personally believe that Engineering should have some influence in providing storage for power, with a probability of an upgrade, its just gated by total charge of the device.

I was tempted to make a random machine part feature since Engineering does get the parts lathe circuit board, but I think that would actually affect Science directly, since machine parts directly correlate with upgrades, and I definitely do not want to do that.

Something that I would like to personally steer Engineering towards is not only their production of power, but having some incentive to benefit from the power they create, this is the main purpose of the device.  A secondary effect would also be monitoring the drain of batteries on the grid to give Engineering a reason to revisit the RCON more than once aside from checking for grubs and other anomalies that usually happen in Science.

I can easily make the cost of creating more batteries higher, since "repeatedly creating other cells" seems to be a primary concern, and I can see that myself, but given that hypers can be found in junk piles, I would like to keep this probability in tact at least.

code/game/objects/items/weapons/RMS_vr.dm
Holdelta: True, but given that the cell actually isn't too powerful(it doesn't recharge by itself, it needs to be in a device to actually start its regeneration), and the abysmal chance of it appearing along with the device at full charge, its more of a gimmick than anything.  Plus, I believe that the time investment in even trying to get this outweighs the shenanigans.

code/game/objects/items/weapons/RMS_vr.dm
Holdelta: Will look into, I was borrowing code from other devices to help make this one.

Holdelta:
So after taking some time to register the general feedback I got both on here and Discord, I would like to repurpose this tool to do its initial design: creating materials(as is suggested by the name).  I might have gotten too excited to make this tool "all-purpose" in being a battery creator, and the suggestion to add the radial menu to navigate the device's features does sound nice.

I still hold an opinion that Engineering should have the option to manage power-related things, but I think that's for another time.  

@Arokha Since I'm basically remaking this tool to refocus its purpose, should I close this out and remake the PR when version 2 is ready to be reviewed?

Novacat:
You can leave this PR open and just update it when it is finished.

code/game/objects/items/weapons/RMS_vr.dm
Holdelta: Keeping charge_stage being in update_icon, since I can't think of a more elegant solution to this.

Holdelta:
@Arokha @Novacat @Rykka-Stormheart 
I've essentially reworked the device to be truer to its name in creating materials, please review to make sure it is fine, and I can adjust if needed.

MrSign:
Oo, I like the sound of this! It will certainly help a lot in engineering projects where you want to dedicate your backpack to electronics and will only need to carry this around to synthesize construction materials. It's like an RCD for materials!

I support this!

Novacat:
I am not sure why random is an option? Usually people would know what they want to use. Why not replace it with wood instead, which seems to be missing from the list of construction materials?

Holdelta:
I included the random feature to keep an open-ended option to pull in other, optional materials at a probability.  I put in wood under that function since it is technically accessible through the outside/botany, which is fairly accessible in general.

Arokha:
We'll see if it causes any issues!

PR #7804, opened 2020-05-12
MedHUD Patch

radar651:
Medical HUD patch. Arrr. Adds the a Health HUDpatch to the load out for the same price as the sec one. Also fixes the capitalization of the SecHUD Patch to make it easier to fine.


KasparoVy:
Yarrr matey.

Arokha:
Tested?

radar651:
Tested and working. Will only spawn on medical.

PR #7822, opened 2020-05-13
Goggles

radar651:
Adds two new goggle types w/ pre-selectable prescriptions and makes the Measons and Material Scanners air tight like standard goggles. New goggle types included a Sec Scanner that lacks flash protection and a Medical scanner. They are also added to the load out screen to be spawned in. They are also air tight like the fore mentioned Measons and Material scanners. I added the airtight flag to the goggle scanners (except for the science ones) because of the sprite mostly and the fact not everyone wants to EVA with a Void or Hardsuit on. Some people just wants to go out in snow suits. However for some reason the goggles never had airtight protection on them and it seemed light an oversight especially for mining. Mining spawns with breath mask only and not everyone wants to dumpster dive for gas mask shift start. This doesn't add airtight to gravatons, thermals or NVGs. Just the standard scanners. Tesari and Vox version of the mob sprite are also included. If pictures of the goggles are requested I can post them.


Arokha:
Can't merge this with all those unmarked vorestation edits to non-VR files. Either this needs to go to Polaris and get at least approved before it's merged here, or, needs to move to `_vr` files.

radar651:
I'll just close this.

PR #7828, opened 2020-05-13
Clean up air tank processing and initialize

Arokha:
Air tanks only start processing if a player picks them up for the first time. Air tanks in closets, racks, etc won't process needlessly. They are pretty unlikely to explode, after all.


Leshana:
This indeed should be good!  If we ever add code that causes tanks to heat up when their surroundings are heating up (so for example, tanks might burst when the room around them is on fire) then we would need to start processing in that case too.

PR #7829, opened 2020-05-13
Optimize flashlight processing

Arokha:
Flashlights only process while they're on.

### DIFF
--- a/code/game/objects/items/devices/flashlight.dm
+++ b/code/game/objects/items/devices/flashlight.dm
@@ -20,23 +20,19 @@
 
 /obj/item/device/flashlight/Initialize()
 	. = ..()
-	update_icon()
 
-/obj/item/device/flashlight/New()
-	if(power_use)
-		START_PROCESSING(SSobj, src)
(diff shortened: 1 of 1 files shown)


code/game/objects/items/devices/flashlight.dm
```diff
@@ -20,23 +20,19 @@
...
-			cell = new cell_type(src)
-			brightness_levels = list("low" = 0.25, "medium" = 0.5, "high" = 1)
-			power_usage = brightness_levels[brightness_level]
 
+	if(power_use && cell_type)
+		cell = new cell_type(src)
+		brightness_levels = list("low" = 0.25, "medium" = 0.5, "high" = 1)
+		power_usage = brightness_levels[brightness_level]
 	else
 		verbs -= /obj/item/device/flashlight/verb/toggle
```
Leshana, line 29: New code skips adding the verb in the case of a flashlight which both uses power (power_use = TRUE) and starts without a cell (cell_type == null).  However as there are no such flashlights in existence this is currently safe.

PR #7830, opened 2020-05-13
Optimize suit cooler processing

Arokha:
Suit coolers only process when they're on

### DIFF
--- a/code/game/objects/items/devices/suit_cooling.dm
+++ b/code/game/objects/items/devices/suit_cooling.dm
@@ -28,14 +28,17 @@
 /obj/item/device/suit_cooling_unit/ui_action_click()
 	toggle(usr)
 
-/obj/item/device/suit_cooling_unit/New()
-	START_PROCESSING(SSobj, src)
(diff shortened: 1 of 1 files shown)


code/game/objects/items/devices/suit_cooling.dm
```diff
@@ -28,14 +28,17 @@
...
-	START_PROCESSING(SSobj, src)
-	cell = new/obj/item/weapon/cell/high()	//comes not with the crappy default power cell - because this is dedicated EVA equipment
-	cell.loc = src
+/obj/item/device/suit_cooling_unit/Initialize()
+	. = ..()
+	cell = new/obj/item/weapon/cell/high(src)	//comes not with the crappy default power cell - because this is dedicated EVA equipment
+
+/obj/item/device/suit_cooling_unit/Destroy()
+	qdel_null(cell)
+	return ..()
```
Leshana, line 37: Ideally should stop processing when deleted.

Arokha: There's actually an attempt to do that in /obj/Destroy(). All my other optimization ones are double-safety, but, missing one won't destroy the universe.

PR #7836, opened 2020-05-13
Only process rigs on a person

Arokha:


code/modules/clothing/spacesuits/rig/rig.dm
```diff
@@ -466,14 +480,12 @@
...
-	var/mob/living/M
-	for(var/obj/item/piece in list(gloves,boots,helmet,chest))
-		if(piece.loc != src && !(wearer && piece.loc == wearer))
-			if(istype(piece.loc, /mob/living))
-				M = piece.loc
-				M.unEquip(piece)
-			piece.forceMove(src)
+	// Not on a mob...?
+	if(!ismob(loc))
+		wearer?.wearing_rig = null
```
Leshana, line 485: For maximum oomph, should check that wearer.wearing_rig == src before setting it null.

Arokha: Done

PR #7850, opened 2020-05-14
Further add on to the alert system

Arokha:
Re-sorted all the .dmis to appease my OCD, and for the same reason, added backdrops to alerts that match your HUD style.

Also made it so your hud style can override the alerts, so holo can keep their snowflakey alert sprites.


KasparoVy:
Dope. Good tweak

PR #7856, opened 2020-05-14
Add inventory item tooltips

Arokha:


KasparoVy:
I dig the PAN card. Way cooler than the drone card!

PR #7860, opened 2020-05-15
Adds ottah plush

Arokha:


KasparoVy:
Man all the sprites in there are so bloody adorable.

PR #7866, opened 2020-05-15
Update github issues template with comments

Arokha:


Rykka-Stormheart:
Perhaps remark that instead of being on github for suggestions, they could go to Discord?

PR #7871, opened 2020-05-15
Allow hiding other huds like ghost huds with F12

Arokha:
@Cyantime


Cyantime:
A+ would complain in deadchat again

PR #7876, opened 2020-05-15
Move cyberpunk detective gear to be suits, add long version

Arokha:
Changes the alternative detective uniform (the sleek cyberpunk coat) to be a suit instead so you can wear it over a uniform, and adds a long version.


KasparoVy:
Niiiice.

PR #7893, opened 2020-05-16
Fixes boxing ring to not keep players trapped forever unless pulled out.

H0lySquirr3l:
I attempted to run a wrestling event a bit ago after having ported over the boxing ring sprites. We found that you could climb INTO the ring, but would need someone from outside to grab you to let you out. After looking through the code it seems that some recent rewrites removed a part that would auto push a person onto the turf they were trying to climb onto. So I added the missing bits and the ring now allows you to climb in and out.

### DIFF
--- a/fitness_vr.dm
+++ b/fitness_vr.dm
@@ -1,0 +1,178 @@
+/obj/structure/fitness/boxing_ropes
+	name = "Ropes"
+	desc = "Firm yet springy, perhaps this could be useful!"
+	icon = 'icons/obj/fitness_vr.dmi'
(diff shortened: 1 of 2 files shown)


fitness_vr.dm
```diff
@@ -0,0 +1,178 @@
+/obj/structure/fitness/boxing_ropes
```
Arokha, line 1: You did something weird here, adding this file to the root of the repo rather than where it should go.

Arokha:
Just checking in. Were you planning on fixing the above?

PR #7897, opened 2020-05-16
Port over AI Opening Doors from Paradise

Rykka-Stormheart:
This makes an [OPEN] link appear on all radio messages the AI hears (to the right of the follow link). When clicked, it allows the AI to open
the door nearest to the speaker (or, if it is a voice changer, the door
nearest to the poor sap who had his voice stolen).

AI, Open This Door!


KasparoVy:
Bolt option too ✌️👽

Rykka-Stormheart:
> Bolt option too ✌️👽

Could consider it, but in what use case would it be necessary? "AI, bolt in [x]"? You'd just pop over to where they are and do it :p

PR #7898, opened 2020-05-16
Updated Whitelist to include myself under Proteans

Tastypred:
[alienwhitelist.txt](https://github.com/VOREStation/VOREStation/files/4637908/alienwhitelist.txt)

https://forum.vore-station.net/viewtopic.php?f=45&t=1759

### DIFF
--- a/alienwhitelist.txt
+++ b/alienwhitelist.txt
@@ -1,0 +1,56 @@
+some~user - Species
+
+admiraldragon - Vox
+aetherelemental - Daemon
+arandomalien - Xenochimera
+arokha - Protean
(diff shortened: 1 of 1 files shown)


KasparoVy:
Bold PRing it before the application goes through!

Screemonster:
> This needs to be moved to the config folder so that the proper file is changed.

I was wondering why the diff was all green until I noticed it was in the root folder

Arokha:
You put the file in the wrong place. Just remake this if/when it's approved, I'd rather it not be sitting on the PR tracker.

PR #7918, opened 2020-05-17
Add protean shakes

Arokha:
Adds protean shakes to fitness vendors (as contraband), with minor mechanical effect.

Liquid protean is weak NIF repair juice (10% as effective), and weak tricord (30% as effective). Metabolizes at half the speed though, so you probably shouldn't try to replace either with it.

### DIFF
--- a/code/game/machinery/vending.dm
+++ b/code/game/machinery/vending.dm
@@ -819,7 +819,8 @@
 					/obj/item/weapon/towel/random = 20)
 	//VOREStation Edit End
 
-	contraband = list(/obj/item/weapon/reagent_containers/syringe/steroid = 4)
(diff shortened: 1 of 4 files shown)


Rykka-Stormheart:
description = "This seems to be a small portion of a Protean creature, still slightly wiggling."
taste_description = "wiggly peanutbutter"

.... urgh


nice

Heroman3003:
Honestly, for the NIF repair aspect of it, speed of processing is entirely irrelevant downside, since as long as it processes eventually (and if you dont die it will), you get full use of its effect.

Arokha:
True though still, 0.1x the effect.

Little-119:
I feel responsible for coming up with this dumb pun

PR #7926, opened 2020-05-17
New jetpack sprite from Eris

Arokha:
Same shape/perspective, but with more shading so it's not so 'flat' looking.


KasparoVy:
Looks great.

PR #7936, opened 2020-05-17
New ore chunk sprites, ore box sprite

Arokha:
Makes the ore have more colors, more shading. Removes things like the coal ore which is just a solid black blob.


Rykka-Stormheart:
I gotta say, I like all of these except the phoron one :c

Arokha:
Why not that one? I think the crystal looks neat

Rykka-Stormheart:
> Why not that one? I think the crystal looks neat

The brighter pink-ish color fits the current phoron sheets really well. Unless you're planning on updating those too, the dull-pink crystal doesn't look as... vibrant as the current phoron ore. The current ore does look a little out of place compared to the new sprites, but the new phoron crystal feels more 'dull'. Perhaps a brighter pink on the ore crystal bits itself might help?

Arokha:
Well it does make a bit of sense that the ore is more dull than the refined counterpart. I mean you just dug it out of the ground.

KasparoVy:
These sprites look dope!

PR #7943, opened 2020-05-18
Semi-automate discord ID registration

Arokha:


Leshana:
You can adjust your database using something like:
```sql
ALTER TABLE `erro_player` ADD `discord_id` VARCHAR(64) NULL, ADD INDEX (`discord_id`);
```

PR #7956, opened 2020-05-18
Removes bad kinks

Arokha:
In the pipe network, obviously.

Rearranges most of the pipe network to be more amenable to ventcrawling in honor of our new blob friends. Removes random needless zigzags and tries to remove almost all 'pipes that dead-end into a vent' so you don't suddenly fall out of pipes with little warning as you speed along.

Also connects various areas that weren't connected before because there was really no need, but with ventcrawl there's a transportation need.


KasparoVy:
Smooth.

Little-119:
Aronai, you dork

PR #7957, opened 2020-05-19
Stops hiding italicised messages from observers

Cyantime:
Already sent to polaris as well


Arokha:
Wat. What even was the purpose of that?

Cyantime:
I assume to reduce chat speed, it'd hide the non-radio component of headset usage. But it also hid all whispers, and any attempted headset usage that didn't go through due to telecomms failures

PR #7958, opened 2020-05-19
Fix custom species synthetics' hunger alert icons

Little-119:
Custom species characters with full body prosthetics will now properly get the battery icon for their hunger alerts instead of the hamburger icon or bloodsucker icon.

As a consequence of this change, bloodsucker synthetics will now get the battery icon instead of the blood icon. I don't know if that's a problem or not.


Little-119:
P.S. Aronai is a dork

PR #7960, opened 2020-05-19
Change stack autocombine

Arokha:
Only happens when dropped or pulled by a mob, to avoid problems with stack-using machines and conveyors in mining


KasparoVy:
Yay!

PR #7984, opened 2020-05-20
Reduce job switch time

lbnesquik:
Ten minutes ends up denying people and seems to last much longer than real life time, given we have been subject to horrible time dilatation those past weeks, at regular pop.

Maybe a better system would be for our entire game to have a sort of external clock against which we could run our time based things, but until then, this is a decent way.

### DIFF
--- a/code/game/machinery/computer/timeclock_vr.dm
+++ b/code/game/machinery/computer/timeclock_vr.dm
@@ -210,9 +210,9 @@
 /obj/machinery/computer/timeclock/proc/checkCardCooldown()
 	if(!card)
 		return FALSE
-	var/time_left = 10 MINUTES - (world.time - card.last_job_switch)
+	var/time_left = 5 MINUTES - (world.time - card.last_job_switch)
 	if(time_left > 0)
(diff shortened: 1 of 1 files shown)


Natje0:
You're not supposed to be switching jobs quickly. I think it's fine as is, even if the real time would be as much as 15minutes.
Heads of Personnel/CRO's still exist.

Screemonster:
how often are you encountering this message that it becomes a problem?

lbnesquik:
> 
> 
> how often are you encountering this message that it becomes a problem?

I've had it happen a handful of times in the past week and figured i would just make a PR for it, given it is only a number change.

Very-Soft:
10-15 mins is not a terribly long time, but is long enough to deter people from job hopping. I would prefer the time was not shorter than 10 minutes.

Arokha:
image

tidi is usually ~10%. Sometimes it spikes up, and sorry if you were affected by that, but we're doing what we can to alleviate it and reduce lag all the time, so I think you'll find that it stays at a reasonable time most of the time.

Last night I noticed that a 6 hour shift was more like 6:30, which is really not that much time dilation (about 8%)

Arokha:
Let's just fix tidi instead

PR #8021, opened 2020-05-21
Refactor Character Setup to use MAP

Leshana:
- Instead of using expensive getFlatIcon() calls, character previews are now shown using screen objects in a second map window.
- Edited skin to add a named window for preferences setup that contains a browser plus a new map window to show the previews.
- Mannequins are still generated the same way, but now a MA is applied to four screen objects which are shown in the second map window.
- The screen objects are deleted on close of the preferences window.  So we do make sure to close it when spawning.

- This improves performance, but also has some other advantages:  The preview icons are now animated.

WWc0ZKJvEB

Somewhat based on tgstation/tgstation#41978


KasparoVy:
Wow. That's creative as hell

PR #8023, opened 2020-05-22
Fixes micro holder hug spam.

Verkister:
This is why we can't have nice things.
kuva
This was less than a minute time span and _each and every one of those messages made the chat freeze the client for near half a second_ :v


Arokha:
Does this also protect against like, hitting Z to attack_self the thing in your hand?

Verkister:
ah heck gonna have to test

Verkister:
Okay seems to work for both.

PR #8038, opened 2020-05-22
Fix some aro2.dmm bugs

Arokha:


KasparoVy:
Nooo not bugs on the fancy spaceship!

PR #8048, opened 2020-05-23
Delete ghosts after 10 minutes if no client

Arokha:


KasparoVy:
Banished to the shadow realm

Arokha:
The MEGA-shadow realm since ghost realm is already the shadow realm

PR #8050, opened 2020-05-23
Redo respawning

Arokha:
Now 'respawn' is 'Return to Menu' and you can always use it, though it will prevent you from being revived in-round after that point. You can swap back and forth between menu and observing, if you want, but you still can't join the round until your respawn timer is up.

So, basically, instead of the respawn timer deciding when you can go back to the menu, it decides when you can rejoin the round. Which I think makes a bit more sense.

### DIFF
--- a/code/_helpers/global_lists.dm
+++ b/code/_helpers/global_lists.dm
@@ -22,6 +22,9 @@ var/global/list/joblist = list()					//list of all jobstypes, minus borg and AI
 
 var/list/mannequins_
 
+// Times that players are allowed to respawn ("ckey" = world.time)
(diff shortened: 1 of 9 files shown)


Heroman3003:
I know this is probably out of scope of this PR, but maybe current functionality of "Quit The Round" can be merged into the new Respawn? Since right now both Returnign to Menu and Quit The Round both prevent your revival, but Quit The Round also properly clears your records (allowing to rejoin with same character if you died of vore and didnt have backup, for example). Clearing records automatically when Returning to Lobby would just be a good QoL thing, IMO.

Issue #7591, opened 2020-05-02
AI unable to Examine


AdmiralDragon:
#### Brief description of the bug
AI are unable to examine people, this can complicate things in the case the AI is trying to see what a short/micro sized character is holding, also some weapons or items don't have the best of "in hand" sprites so examining is needed to figure out what they have.

#### What you expected to happen
Shift+Left Clicking = Examine Person

#### What actually happened
Shift+Left Clicking did nothing
Not sure if it was the connection but I experienced a slight freeze when I tried to Examine though no info appeared.

#### Steps to reproduce
- Join as AI
- Shift+Left click, "Right Click" then select "Examine"
- Watch the chat and see as nothing appears.

#### Code Revision
- e89555cb291cd727bf4ca3d06bd065032403287d
 
#### Anything else you may wish to add:
- Nothing to add.

KeRSedChaplain:
This is a feature, unfortunately

AdmiralDragon:
I was able to do that before an update recently which took it away. I usually use it to figure out who I can bother as an AI without interrupting a scene. Without Examine I am unable to figure out who's in a scene.

KeRSedChaplain:
huh

Arokha:
Hm! That's never intentionally been a feature. It's possible someone broke it to work, but it's always been a balancey thing that you can't see what people are holding and such as AI. Maybe someone can code up a middleground that lets you read OOC notes.

PR #8058, opened 2020-05-24
Fixes Accessories on Taurized Suits & Clipping

KasparoVy:
Clipping logic got out of wack in #7591, that's all.
https://github.com/VOREStation/VOREStation/commit/48563200cee4aad945aa5bb235a52e3369fa72da#diff-ac3b89e91ad083b3cb3fdb669006158cR779

Fixes long-standing issue causing accessories on taursuits to render 16 pixels left on the x axis. I'm surprised I figured this out but not surprised I totally bungled up the first commit.
virgo ss13 fixed accessory overlays on taurized suits 20200524

### DIFF
--- a/code/modules/clothing/clothing_vr.dm
+++ b/code/modules/clothing/clothing_vr.dm
@@ -187,6 +187,8 @@
 			var/image/I = new(A.get_mob_overlay())
 			I.pixel_x = 16 //Opposite of the pixel_x on the suit (-16) from taurization to cancel it out and puts the accessory in the correct place on the body.
 			standing.add_overlay(I)
+	else
+		return ..()
 
 //TFF 5/8/19 - sets Vorestation /obj/item/clothing/under sensor setting default?
 /obj/item/clothing/under
--- a/code/modules/clothing/clothing_vr.dm
+++ b/code/modules/clothing/clothing_vr.dm
@@ -182,11 +182,11 @@
 	return standing
 
 /obj/item/clothing/suit/apply_accessories(var/image/standing)
-	if(LAZYLEN(accessories))
+	if(LAZYLEN(accessories) && taurized)
 		for(var/obj/item/clothing/accessory/A in accessories)
-			var/mutable_appearance/M = new(A.get_mob_overlay())
-			M.pixel_x = 16 //Opposite of the pixel_x on the suit (-16) from taurization to cancel it out and puts the accessory in the correct place on the body.
-			standing.add_overlay(M)
+			var/image/I = new(A.get_mob_overlay())
+			I.pixel_x = 16 //Opposite of the pixel_x on the suit (-16) from taurization to cancel it out and puts the accessory in the correct place on the body.
+			standing.add_overlay(I)
 
 //TFF 5/8/19 - sets Vorestation /obj/item/clothing/under sensor setting default?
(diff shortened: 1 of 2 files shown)


KasparoVy:
And for my next trick, I'll make accessory versions of the colourful poncho & drake cloak nyeh heh heh...

code/modules/clothing/clothing_vr.dm
```diff
@@ -181,6 +181,13 @@
 		standing.layer = BODY_LAYER + 15 // 15 is above tail layer, so will not be covered by taurbody.
 	return standing
 
+/obj/item/clothing/suit/apply_accessories(var/image/standing)
+	if(LAZYLEN(accessories))
+		for(var/obj/item/clothing/accessory/A in accessories)
+			var/mutable_appearance/M = new(A.get_mob_overlay())
+			M.pixel_x = 16 //Opposite of the pixel_x on the suit (-16) from taurization to cancel it out and puts the accessory in the correct place on the body.
+			standing.add_overlay(M)
```
Arokha, line 189: I don't see any conditions to make this only shift by 16 on taurs only?

Passing an MA into add_overlay seems to do *weird things* and can cause *weird issues*. Namely SSoverlays skipping the rest of the queued overlays that tick. It's not noticeable in local testing where having a significant queue is basically impossible, but I did a similar thing on the live server and people started getting random things stuck to their bodies. Do you remember that shift? ;p

I'd suggest using an image instead! You can do the exact same code here and just replace the MA with an image instead, as they also have pixel stuff (MAs are actually a child of /image but don't tell anyone). You can just do var/image/I = A.get_mob_overlay(), I.pixel_x = 16, standing.add_overlay(I) (get_mob_overlay returns an image so you don't need to like, turn it into another image)

KasparoVy: Fuck I thought I had the taurized condition in there, that would've been disastrous. Good call regarding the SSoverlays queue, always gotta account for the voodoo

Testing and committing in a sec

KasparoVy:
Thanks for the review @Arokha

code/modules/clothing/clothing_vr.dm
```diff
@@ -181,6 +181,13 @@
 		standing.layer = BODY_LAYER + 15 // 15 is above tail layer, so will not be covered by taurbody.
 	return standing
 
+/obj/item/clothing/suit/apply_accessories(var/image/standing)
+	if(LAZYLEN(accessories) && taurized)
+		for(var/obj/item/clothing/accessory/A in accessories)
+			var/image/I = new(A.get_mob_overlay())
+			I.pixel_x = 16 //Opposite of the pixel_x on the suit (-16) from taurization to cancel it out and puts the accessory in the correct place on the body.
+			standing.add_overlay(I)
+
```
Arokha, line 190: Needs an else return ..() at the end so the normal one runs if not taur. This is overriding another proc, and now all it does is taur ones. So if you don't have taur suit, it will get here, this if() will be false and then... Nothing. No accessories for you.

KasparoVy: Fffuuuuuck.

KasparoVy:
Alright we good at last. I sure screwed that up, whew

Arokha:
Is cool mang

PR #8078, opened 2020-05-25
removes stacking eye_blurry from myelamine

lm40:
maybe now it will no longer last forever

### DIFF
--- a/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Medicine.dm
+++ b/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Medicine.dm
@@ -665,7 +665,7 @@
 /datum/reagent/myelamine/affect_blood(var/mob/living/carbon/M, var/alien, var/removed)
 	if(alien == IS_DIONA)
 		return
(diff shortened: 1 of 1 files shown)


code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Medicine.dm
```diff
@@ -665,7 +665,7 @@
 /datum/reagent/myelamine/affect_blood(var/mob/living/carbon/M, var/alien, var/removed)
 	if(alien == IS_DIONA)
 		return
-	M.eye_blurry += min(M.eye_blurry + (repair_strength * removed), 250)
+	M.eye_blurry = min(M.eye_blurry + (repair_strength * removed), 250)
```
Arokha, line 668: Wonder how long that's been there.

There's so many fucked up uses of eye_blurry in the code from a quick glance. Like wearing thermals and shooting someone with ions sets their eye_blurry to 5, regardless of what it was before, so you can easily 'reset' eye blurry by cheesing it that way. There's soooooooo many places like that.

PR #8081, opened 2020-05-26
Show sent PDA messages to sender

Cyantime:
I think I remember this being requested like a year or more ago


TheFurryFeline:
Ah, yes, this is good. Always was dumb that you only ever had one half of a log for your chat, well, log.

PR #8085, opened 2020-05-26
Medihound inventory changes

Arokha:
Removes alkysine and imidazoline from medihounds, and replaces their beaker with two more syringes instead so they can actually carry separate drugs.

In exchange, a vial of imidaz spawns in surface med chem, which the borg can steal if they really want, but anyone can use and now no particular specific characters spawn with drugs that aren't even available to people at roundstart.

### DIFF
--- a/code/modules/mob/living/silicon/robot/dogborg/dog_sleeper_vr.dm
+++ b/code/modules/mob/living/silicon/robot/dogborg/dog_sleeper_vr.dm
@@ -1,3 +1,5 @@
+#define SLEEPER_INJECT_COST 600 // Note that this has unlimited supply unlike a borg hypo, so should be balanced accordingly
+
 //Sleeper
 /obj/item/device/dogborg/sleeper
 	name = "Medbelly"
@@ -11,7 +13,7 @@
 	var/min_health = -100
 	var/cleaning = 0
 	var/patient_laststat = null
-	var/list/injection_chems = list("inaprovaline", "dexalin", "bicaridine", "kelotane","anti_toxin", "alkysine", "imidazoline", "spaceacillin", "paracetamol") //The borg is able to heal every damage type. As a nerf, they use 750 charge per injection.
+	var/list/injection_chems = list("inaprovaline", "dexalin", "bicaridine", "kelotane", "anti_toxin", "spaceacillin", "paracetamol") //The borg is able to heal every damage type. As a nerf, they use 750 charge per injection.
 	var/eject_port = "ingestion"
 	var/list/items_preserved = list()
 	var/UI_open = FALSE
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/silicon/robot/dogborg/dog_sleeper_vr.dm
```diff
@@ -11,7 +11,7 @@
 	var/min_health = -100
 	var/cleaning = 0
 	var/patient_laststat = null
-	var/list/injection_chems = list("inaprovaline", "dexalin", "bicaridine", "kelotane","anti_toxin", "alkysine", "imidazoline", "spaceacillin", "paracetamol") //The borg is able to heal every damage type. As a nerf, they use 750 charge per injection.
+	var/list/injection_chems = list("inaprovaline", "dexalin", "bicaridine", "kelotane", "anti_toxin", "spaceacillin", "paracetamol") //The borg is able to heal every damage type. As a nerf, they use 750 charge per injection.
```
Rykka-Stormheart, line 14: So wouldn't this undo the medihound buff/nerf thing? They still cost 750 per injection, but can now no longer heal every damage type?

Arokha: I don't really have any evidence that wasn't referring to the 4 main types. 'eye damage' isn't what I'd call in 'every damage type', and they can't heal genetic damage so why would that comment exist if that's what it meant?

That's how I interpret it anyway.

Novacat:
Does the vial start uncapped? Medihounds have no gripper so no way to get the medicine otherwise.

Actually, id go ahead and remove the imidazoline vial from chemistry and just add the beaker back to the medihound since the medihound can do basic chemistry with it (just nothing that requires grinding up phoron)

Crisis and Surgery borgs have grippers/beaker/syringe so they are still, mechanically, superior options to the medihound.

code/modules/mob/living/silicon/robot/dogborg/dog_sleeper_vr.dm
Arokha: Reduced cost somewhat to compensate. It's still more than the borg hypo because unlike the hypo, it has an infinite supply whereas the hypo takes time to regenerate the stockpile (several minutes).

Arokha:
The vial probably does start open. If not, someone should update the default icon. But I'm pretty sure it does, since it's just a small beaker basically.

Also readded the beaker for now, pending a discussion after this about borgs and doing chem.

TheFurryFeline:
I don't see much of an issue with Borg doing chemistry if there ain't anyone else to do it

MrSign:
If there's no one around to do chemistry. I don't see why borgs can't do it if they have access to the equipment needed to do it!

They'll just fall under our rules of 'Don't steal someone else's job'. If a reasonable attempt has been made by the cyborg to have medicines made for it by a Chemist but said Chemist is unavailable to do them. I'd say the medihound can go right ahead!

Arokha:
This wasn't the place to talk about borgs doing chem :V That's for a separate PR, that's why it's separaaaate.

PR #8092, opened 2020-05-27
Give medihounds a new toy

Arokha:
They don't have a way to deal with bleeding, so now they have a clotting spray (looks like a purple advanced kit) that stops bleeding on a patient. Doesn't heal at all, but seals every open wound with a single spray. Without this, they don't have a way to stop bleeding.

Note that you can still reopen all their wounds by dragging them around. It's just a flimsy film not body armor. So this doesn't give you an excuse to drag bodies.


Arokha:
> Very nice! I'm assuming it recharges via cyborg recharger or by internal battery?

Yes, it uses the whole "matter synth" like other borgs with similar items have, where it shows up on the stat panel as a 'medical stockpile' or whatever and is recharged in a recharger.

MrSign:
> Yes, it uses the whole "matter synth" like other borgs with similar items have, where it shows up on the stat panel as a 'medical stockpile' or whatever and is recharged in a recharger.

Forkin' noice, moite. Yer alroight.

PR #8094, opened 2020-05-28
Add a large shotgun slugs box to the Red armory.

lbnesquik:
The station lacked any of this ammo type, save in autolathes, which security generally shouldn't be messing with.

### DIFF
--- a/maps/tether/tether-07-station3.dmm
+++ b/maps/tether/tether-07-station3.dmm
@@ -857,7 +857,11 @@
 /obj/structure/table/steel,
 /obj/item/weapon/storage/box/shotgunshells{
 	pixel_x = 6;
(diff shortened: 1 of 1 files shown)


Arokha:
They should probably just use a rifle or something instead of shotgun slugs.

Arokha:
After asking around, I'm told this was due to fighting an ADG. Should have just used an ion rifle. Slugs don't make much sense on a space station anyway where the outside will kill you and you shatter windows in one shot with them.

Novacat:
We had a discussion on this as staff and it was decided that Security does not really have any need for slugs on roundstart. If you really feel like you need slugs for whatever reason, it would probably be security's best bet to ask Science for them.

PR #8095, opened 2020-05-28
Adds a verb to set OOC notes mid-round

Arokha:
Admin put you into a mob and now you don't have OOC notes? Spawn as a corgi from runes? Join as a lost drone? Just forget to update your OOC notes?

Now there's a verb for that. (You still need to update them on your Character Setup in the last case, though, this won't update your permanent save slot for you).

### DIFF
--- a/code/modules/mob/living/living_vr.dm
+++ b/code/modules/mob/living/living_vr.dm
@@ -40,3 +40,14 @@
 				to_chat(src, "Rider control restricted.")
 				return
 	return
+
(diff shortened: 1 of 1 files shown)


Natje0:
In-round OOC note changing should probs be logged to prevent cases where people could change their notes then yell about prefbreak over stuff that wasn't there when the oof occured.
Aside from that, **yaaas**

KasparoVy:
Noice.

PR #8096, opened 2020-05-28
Add sleevemate feature

Arokha:
Allows sleevemates to interact with people inside other people.

Only 1 layer deep, because listen, it's late, and I don't want to try to write a thing to avoid infinite loops right now, ok?

image


Screemonster:
o-oh

yes good

PR #8108, opened 2020-05-28
Tamper with railguns

Arokha:
Make their icons updating more efficient, swap a few icons for Urist icons, add flechette pistol in normal and KHI varieties


code/modules/projectiles/guns/magnetic/magnetic_railgun_vr.dm
```diff
@@ -0,0 +1,11 @@
+/obj/item/weapon/gun/magnetic/railgun/flechette/pistol/khi
+	name = "kitsuhana flechette pistol"
```
Novacat, line 2: Personally, I would say Kitsuhana probably should not be going into arms manufacturing? They already have a ton of products on station. Most weapons are manufactured by Hephaestus if you need a different company name, though probably would be fine to just remove the company name entirely.

Arokha: They didn't make it, as the description says, it's just been upgraded. This isn't the normal rail pistol I added either, that is in the other file. This is just a version for admemes.

Novacat: ah, Oki

PR #8131, opened 2020-05-30
Adds in entertaiment monitor to the surface medical break room.

H0lySquirr3l:
image


H0lySquirr3l:
I was cufused by the map bot not coming up. Should I resubmit?

Arokha:
Probably! I'd suggest just using a real git client rather than uploading these, or just having fastdmm2 make the PR for you since it does mapdiffbot. Or exporting them from fastdmm2 in tgm format and PR'ing those.

H0lySquirr3l:
I... don't understand half of that. What is fastdmm2? Wouldn't the mapbot still have to check it even if it makes the pr? I also still don't know why this is happening. The bots were broken yesterday and seemingly broken today as well.

PR #8147, opened 2020-05-31
Slash clothes prices in half or more. 

lbnesquik:
Cheaper clothes, more fun. Why are T shirts and gloves 200 thalers anyway?

In more seriousness, those prices were just way too expensive and even with the well off wealth, it is not hard to run out of money if you are trying out lots of styles.

### DIFF
--- a/code/modules/mining/abandonedcrates.dm
+++ b/code/modules/mining/abandonedcrates.dm
@@ -63,7 +63,7 @@
 		if(57 to 58)
 			new/obj/item/toy/syndicateballoon(src)
 		if(59 to 60)
-			new/obj/item/weapon/rig(src)
+			new/obj/item/weapon/rig/industrial(src)
 		if(61 to 62)
 			for(var/i = 0, i < 12, ++i)
 				new/obj/item/clothing/head/kitty(src)


Very-Soft:
They are that expensive to reflect their value in the loadout. You aren't really supposed to be able to afford all the things, the point was to make it so that the things from the loadout are gettable icly, so if you forgot or lost some things then you would have a means by which to get them back.

That said, I don't personally care much if their cost is lowered, I am just explain why the prices are how they are.

lbnesquik:
@ very-soft
I feel like the value of loadout items doesn't really make all of that much sense even then. Lunchboxes are 2 points and basically no outfits cost more than two points either.
If you take that 2 points = 200 thalers, then a simple painted metal lunchbox costs the same, which is.. a lot.

((I said you as a imaginary person, not you _You_))

lbnesquik:
> 
> 
> Rather than just reduce everything to a flat 50 thaler price, I think the better option would be to instead halve the price of everything. 50 becomes 25, 100 becomes 50, 200 becomes 100.

I don't think it is really worth doing. Those prices aren't particularly full of sense in the first place, why did gloves cost 200 thalers when a dress could cost half that? Especially when anything in the suit slot had a 200 cost for reasons???

I could see the couple lightly armored things costing more than 50, but there is nothing gained in fluff or gameplay from having more expensive clothes that are just style points.

lbnesquik:
This PR was closed because i am dumb, follow the #8237 link

PR #8176, opened 2020-06-01
[MIRROR] Add AI Picture-in-Picture Mode!

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/7136
--------------------
This adds a system for picture-in-picture UI windows using
vis_contents. Essentially, it allows you to make UI windows that show an area of turfs. It also refactors how cameranet visibility works.

Currently, this is implemented on AIs. They gain two new UI buttons - "Enter Multicam Mode", and "Create Multicam". When they go into Multicam Mode, they see a background of animated binary numbers, and they are allowed to create an infinite amount of these picture in picture windows, which subsequently creates an AI Eye for each one. They are able to control each AI eye individually, by first clicking on the PIP window to select it as "active" and then using the normal arrow key controls. The PIP windows can be freely resized and moved around the area.

You can control everything inside these PIP windows EXACTLY the same as you can using a traditional AI Eye, as demonstrated below.

For admins, there is a config option to disable PIP entirely - simply set `var/multicam_allowed = TRUE` to FALSE if you wish to disable it from being used. (Please be reasonable.) <3

You can see an example of how this works here:
https://i.imgur.com/TCND8W9.mp4

AI Multicam functionality.

Do note that if the lightbulb in the AI core is busted, the multicam room is dark(er) than it would be, but you can still see your camera windows just fine. (I'll probably fix this later.) It only affects the "matrix" backdrop, the PIP windows are still fine.

This has been runtime-tested with the latest `master` revision and produces 0 runtimes, and has no noticeable impact on server CPU usage.

This is the Polarisport! Port of https://github.com/VOREStation/VOREStation/pull/7752


Little-119:
Uh oh, looks like we got some `merge conflicts`

PR #8177, opened 2020-06-01
Adds mermaid tails taur option

Heroman3003:
Sprites by sugarcuddles, from here https://forum.vore-station.net/viewtopic.php?f=22&t=891

With new system for taurhalf sprites, those are actually compatible. Can be implemented. And so they are.

Example of use ingame: 
Screenshot_2
Screenshot_3
Screenshot_4


KasparoVy:
Wicked! Always nice to see more taur tails.

MrSign:
Fish goes flop flop flop...

Are we expecting these mermaids to wriggle on the ground like fish?!

Heroman3003:
> 
> 
> Fish goes flop flop flop...
> 
> Are we expecting these mermaids to wriggle on the ground like snakes?!

As if any taurs are logical. Flip flop or roll in wheelchair or slither is choice of player! Who knows, maybe their fish half really is long enough to wiggle around.

PR #8184, opened 2020-06-02
Removes Exploration Weapon Permits

Novacat:


Arokha:
They're being removed because it's covered by it being part of their job already. Not actually changing their rights. They're just allowed to take guns on expeditions and stuff.

PR #8189, opened 2020-06-03
improves virology immeasurably

lm40:
my commits are the best, always

### DIFF
--- a/code/modules/virus2/effect.dm
+++ b/code/modules/virus2/effect.dm
@@ -136,7 +136,7 @@
 	mob.updatehealth()
 
 /datum/disease2/effect/killertoxins
-	name = "Autoimmune Reponse"
+	name = "Autoimmune Response"
 	stage = 4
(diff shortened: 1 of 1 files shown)


Rykka-Stormheart:
Always the best. Singlehandedly fixes Virology, new best department 2020

Arokha:
You're right, it is difficult to measure how much this improves virology.

code/modules/virus2/effect.dm
```diff
@@ -136,7 +136,7 @@
 	mob.updatehealth()
 
 /datum/disease2/effect/killertoxins
-	name = "Autoimmune Reponse"
+	name = "Autoimmune Response"
```
KasparoVy, line 139: Repose* ;)

lm40:
```suggestion
	name = "Autoimmune Respose"
```

PR #8193, opened 2020-06-04
Proof of concept Werebeasts can wear clothes!~

H0lySquirr3l:
Werebeasts can wear clothes!
Ports oldman suit, and clown suit with mask and shoes.
Seems werebeasts can actually wear clothes, the sprites just do not exist. So long as you make some and put them in with the same name as normal clothes they will work.

I am open to porting other clothes should someone want them.


KasparoVy:
Ported from another codebase or resprited to fit the race?

Honestly you hit the only real uniform that matters so we should be good after this PR 🤡👌

H0lySquirr3l:
Just resprited. As far as I know, werebeasts are a virgo original.

PR #8206, opened 2020-06-05
Reduces gravgen alarm volume

Cyantime:
why is it *so loud*


Arokha:
Oh

code/modules/power/gravitygenerator_vr.dm
```diff
@@ -395,7 +395,7 @@ GLOBAL_LIST_EMPTY(gravity_generators)
 		M.update_gravity(M.mob_has_gravity())
 		if(M.client)
```
chaoko99, line 396: Hey could you remove this conditional? Both these procs already check for a client on their first line.

Cyantime: sure why not

PR #8208, opened 2020-06-05
Add Aurora Caelus Event

Rykka-Stormheart:
Adds the Aurora Caelus, a rare event that occurs on the station space z-levels~!

During this event a .ogg will play to all clients, and you will see the outside appear almost like a storm - consider it a fun little break and a rare event for a moment of calm.

The event announces as soon as it fires, and assuming time dilation isn't horrible, starts 120 seconds after (60 "ticks" of the event subsystem being fired by the MC, which happens every 2 IRL seconds, affected by time dilation).

Current event weight (frequency) can be tweaked according to admin liking, and it's currently a MUNDANE level event. This can be changed to MODERATE or otherwise as determined by admins, or given a higher weight - during testing, stuff like PDA spam or news or etc very often went over it and it took giving it a lot of weight for it to fire - but it is supposed to be a "rare" cosmic thing, not something that happens every shift.


Heroman3003:
Sound should probably be affected by ambiance preferences.

code/modules/events/aurora_caelus.dm
```diff
@@ -0,0 +1,32 @@
...
+
+/datum/event/aurora_caelus/announce()
+	command_announcement.Announce("[station_name()]: A harmless cloud of ions is approaching your station, and will exhaust their energy battering the hull.\
+	Nanotrasen has approved a short break for all employees to relax and observe this very rare event.\
+	During this time, starlight will be bright but gentle, shifting between quiet green and blue colors.\
+	Any staff who would like to view these lights for themselves may proceed to the area nearest to them with viewing ports to open space.\
+	We hope you enjoy the lights.", "Nanotrasen Meterology Divison", new_sound = 'sound/misc/notice2.ogg')
+	for(var/V in player_list)
+		var/mob/M = V
+		if(affecting_z)
```
Heroman3003, line 15: How does that check in any way affect anything again? It will always be true.

Rykka-Stormheart: Am currently testing alternatives - I also realized that the "announce" would play the sound for all mobs instead of when it actually starts, so I'm going to fix that too.

Little-119:
Make an even rarer version that occurs in the kitchen.

Rykka-Stormheart:
> Sound should probably be affected by ambiance preferences.

Eh, I disagree. The sound only ever plays at the `start` call of the event, once.

Rykka-Stormheart:
Trying to fix Travis Triplets

Arokha:
Seems fine. If the odds are a problem, can tweak.

PR #8210, opened 2020-06-06
New overmap POI

SweetBlueSylveon:
Adds a new map called "Away Mission - Guttersite"
An installation built on a group of asteroids in VE space serves the purpose of giving gutter speaking folk a place to do their shady dealings outside of solcomm jursidiction.
This map is coded up to show up on the overmap, and is accessible by both the excursion shuttle and talon boat.
The map is full of secrets and content, and leaves a lot of room for staff to fill it up with unique mobs and people for events.


maps/tether/tether_defines.dm
```diff
@@ -24,6 +24,7 @@
 #define Z_LEVEL_DEBRISFIELD					23
 #define Z_LEVEL_FUELDEPOT					24
 #define Z_LEVEL_GATEWAY						25
+#define Z_LEVEL_GUTTERSITE					26
```
Arokha, line 27: This doesn't seem right. You're loading it BEFORE the fuel depot. It's probably not z 26. Can you ghost over your map ingame and verify what zlevel it is?

Rykka-Stormheart: The load order *does* matter, I found out - seconding this

PR #8244, opened 2020-06-08
Fixes cyborg rechargers not recharging

Cyantime:


KasparoVy:
Nice, good catch. Cheers.

PR #8246, opened 2020-06-09
Fixes a crash exploit.

BlackMajor:
Probably not the best idea for a proc to call itself until the server dies.


Arokha:
Loop machine broke

BlackMajor:
>Send to polaris

God damnit I keep forgetting virgo is a downstream.

BlackMajor:
Yeah ok sent it to Polaris I guess.

PR #8257, opened 2020-06-10
Adds medical graviton eyepatch

lm40:
I want to have grav vision and medical vision, plus eyepatches are cool. Doesn't actually make them obtainable in any way tho.

graveyepatch


Arokha:
This probably won't enable you to see medical records.

lm40:
IN HINDSIGHT it would have made more sense for this to have been a child of inferior medical eyepatch that already exists.

PR #8260, opened 2020-06-11
Introduce a SINGLE themed, colored explorer shield.

lbnesquik:
It is a simple hunk of metal, painted to a moody green or dangerous voidsuits purple and blacks.
It has a lower block chance than the real deal, but it is still a strong help against melee foes and the occasional beam or two. Do mind that you ABSOLUTELY can't stand in front of most enemy and just ignore them even with two shield, much less one.
And just like any _** tacticool**_ object, it has a big flashlight stuck at the front to help with visibility, it's not like you can actually see much out of that slit.
You can order two for 80 points at cargo.
A single one starts in the pathfinder locker, meaning they decide whether or not it will be used by them, or their fellow partners.

image
image
image


KasparoVy:
Sucks that you can't make a shield out of a few plates of metal haha. Durasteel shield ftw.

Novacat:
The light looks awfully huge on it. Can it be shrunk a bit, to, say, a light bar instead of a massive floodlight?

PR #8268, opened 2020-06-13
NanoUI tweak

SweetBlueSylveon:
Allows simple mobs to interact with nanoUI if they have "have_hands = TRUE".


vorestation.dme
```diff
@@ -783,6 +783,7 @@
 #include "code\game\machinery\holoposter.dm"
 #include "code\game\machinery\holosign.dm"
 #include "code\game\machinery\igniter.dm"
+#include "code\game\machinery\isatooluserfix_vr.dm"
```
Heroman3003, line 786: Imo better file/path to put the fix in would be "code\modules\nano\interaction\default_vr.dm" for naming/file location consistency.

SweetBlueSylveon: isatooluser as far as I know is defined in machinery, hence why I put it there.

Heroman3003: Yes, but you are adding/changing a default_can_use_topic proc, which is defined in default.dm (in nano files)

code/modules/nano/interaction/default_vr.dm
```diff
@@ -0,0 +1,6 @@
+/mob/living/simple_mob/default_can_use_topic(var/src_object)
+	. = shared_nano_interaction(src_object)
+	if(. != STATUS_CLOSE)
+		. = min(., shared_living_nano_distance(src_object))
+
+//Allows simple mobs to interact with nanoui as long as they have "has_hands = TRUE"
```
Arokha, line 6: But you aren't checking has_hands or anything

PR #8281, opened 2020-06-15
[MIRROR] Remove Unused Image Assets

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/7278
--------------------
- Stop sending PDA icons to clients, they have not been used since 2013.
- Stop sending SpiderOS icons to clients, they have not been used since 2014
- Remove PDA icon files, they have not been used since 2013
- Remove Spideros icon files, they have not been used since 2014

I was trying to reduce the number of assets sent on client connection and noticed these are not used at all, anywhere.   After preventing them from being sent on connect, I then reasoned since they haven't been used for so long we should just delete them outright.


KasparoVy:
Holy shit RIP legacy assets

PR #8282, opened 2020-06-15
Adds Creation of Whetstones from Durasteel & Plastitanium

KasparoVy:
Can currently be done with plasteel and titanium so it's fitting to add these two - especially for plastitanium.
Bumped the creation time up a bit for each.


KasparoVy:
Awesome, cheers.

PR #8284, opened 2020-06-15
Reduces Default (and mob) digestion rate

Novacat:
from 2/2 to 0.5/0.5

Due to a discussion in exploration, and the preferences stated by a headmin.


KasparoVy:
Any metrics on the actual time it takes to digest an unprotected person from full HP? Just so we know what the tangible effects of this are?

TheFurryFeline:
> Any metrics on the actual time it takes to digest an unprotected person from full HP? Just so we have a somewhat precise idea of what this does
> 
> I imagine this'll give a lot of time for rescue, rp, etc.

Don't forget, burn/brute resistance/weakness can affect how long digestion takes. But this new default will substantially increase chances of rescue, especially for the dreaded mimic at surface armoury maint area.

KasparoVy:
Yep, makes sense. Cheers, sounds good

PR #8292, opened 2020-06-16
Replaced broken live update script with python version that works

ShadowLarkens:
#8224 :eyes: :knife:


Leshana:
The old bat file has not worked since BYOND changed the cache structure to put files in temporary subfolders instead of the top level cache folder.

Recreating the bundle in this manner will work, and as long as you have an IDE that can run the script it isn't too inconvenient.  It might even be easier to explain this than telling people to disable the bundling and edit the file in the temp folder.

ShadowLarkens:
yeah the old one was broken anyways, the joking stab emote is more because I had to rewrite in python instead of fixing the batch script because of the new json syntax

PR #8302, opened 2020-06-16
Makes Unathi's mob_size MOB_MEDIUM to match all other (non event) species

Heroman3003:
Same reason prommies and teshari had it modified up, IMO, its only fair for the scalies to be sized down for equality of mobswap interactions.

Does not touch Werebeasts' mob_size because event species.


lm40:
now i can actually defeat unathi in wrestling????

PR #8303, opened 2020-06-16
Fix casing on folder

Arokha:
Fix stupid folder casing.


Rykka-Stormheart:
Yay finally!

KasparoVy:
Why do I feel like this'll have to be re-done in a week or so 🤣

Issue #8306, opened 2020-06-18
Redsuits turning invisible when cycled


cadyn:
#### Brief description of the bug
When you cycle a redsuit to vulpkanin it turns invisible.

#### What you expected to happen
I expected when I cycle the redsuit to vulpkanin using the suit cycler, it would cycle correctly.

#### What actually happened
When I used the talon Captain's suit cycler to cycle a redsuit, it turned invisible.

#### Steps to reproduce
- Obtain a redsuit
- Put the redsuit in a talon cycler.
- Set the species to vulpkanin
- Press cycle

Rykka-Stormheart:
This is the same as https://github.com/VOREStation/VOREStation/issues/8227

Meaning that cyclers are fucked entirely for Security vulps I think.

KillianKirilenko:
I don't think it's quite the same, because the proto shouldn't be pointing at the server's subfolder for the dmis; it should be looking in the same places as every other hardsuit. If it's fucked for everyone, that's a problem with that suit in particular.

This is irritating though because I tested the vintage & talon suit variants locally, and they worked fine at the time. I'll do some more local digging later and see what I can do to get this straightened out, unless someone beats me to it.

KillianKirilenko:
Okay, so the problem with #8227 is that the HoS is still being issued the fluff variant protosuit, not the "new" one.

I have a hypothesis as to what the problem with this specific issue is and am testing it currently, along with a fix for the protosuit "problem".

KillianKirilenko:
..I forgot to use proper terminology in the PR, argh.

This should be fixed now though.

cadyn:
I'll double check when I get the chance cheers

cadyn:
Just checked and still comes out invisible.

KillianKirilenko:
Looks like the release branch is out of date, yeah. Should be fixed For Real once it's updated though.

PR #8138, opened 2020-05-30
AutoLok Pressure Suit

KillianKirilenko:
By @Arokha 's request, the AutoLok Pressure Suit! A cutting edge high-tech form-fitting pressure suit with an integral, unremovable helmet that can be worn by almost any humanoid species with minimal adaptation needed.

autolok-pressuresuit

The AutoLok has very very poor protection from conventional weapons, but has exceptional radiation protection and no slowdown. it breaches much more easily than regular suits. The lightweight helmet includes an integrated flashlight, as with all modern spacesuits.

Just needs a little extra code from Aro to add the autosealing function and then it's good to go.

PR #8308, opened 2020-06-18
Assorted Spacesuits & DMI Cleanup

### LINKED ISSUE #8227: Prototype Security Voidsuit has no sprite for Vulpkanin

There are no/missing sprites for Vulpkanin using the Prototype (HoS) Security Voidsuit, resulting in blank sprites when in inventory and worn.

Refit voidsuit, see new sprites

#### What actually happened

Invisible sprite

Icon is pulling from `icons/vore/custom_clothes_vr.dmi`, icon_state "rig-sec", even when refitted back to "human" it simply has a blank sprite

https://i.gyazo.com/638346455d21dfa4603c6803d102acca.png

https://i.gyazo.com/c8d7ba1b06ea878b8ecf9f8e558a5249.mp4

KillianKirilenko:
Fixes #8227 and #8306 by;
- Issuing the HoS with the proper prototype suit and cycler in their office. Access for the suit cycler is changed accordingly.
- Removes the specific _vr overrides on vintage/talon suits.
- Cleans up _vr dmi files that are no longer needed/used.

Also:
- Nerfs the protosuit's armor values to be in line with the standard security suit.
- Adds new Hephaestus-tech MkVII C/E/M/S ERT voidsuits with integral helmets.
- Issues sets of new ERT voidsuits, suit coolers, and magboots on the ERT ship map.


KillianKirilenko:
..why do ERTs need *another* new voidsuit? They already have their own set of voidsuits *and* hardsuits.

Novacat:
They have no voidsuits, just hardsuits.

KillianKirilenko:
..my memory's playing tricks on me, I coulda sworn they had their very own voidsuits as well. My bad.

Okay, so slightly different question then; why do they need their own voidsuits, exactly? Hardsuits do the same job better. A little less user-friendly, but not that hard to deal with.

Giving ERTs more stuff they can choose from when they're under time pressure to prep and deploy doesn't seem like a good idea to me.

Novacat:
The reason for this is that voidsuits are simpler to use and quicker to setup than hardsuits. I feel it was a mistake making hardsuits so ubiquitious within the ERT considering that hardsuits practically require tinkering and modification (it is a good idea to swap the starting cell for a hypercap, and snag some extra modules for it)

The current Prototype voidsuit uses the same armor stats as the blood-red voidsuit which would be unacceptable to have spawn on station, and we had nerfed the on-station HoS voidsuit to being basically just an ordinary security suit anyway.

KillianKirilenko:
(Right, I see why my memory was messing with me; they have 'partial' armor like vests and helmets, but then skip straight to hardsuits.)

See, my gut feeling is the other way; hardsuits are easier to manage since you just pull it on and hit the Go button, and you're set. No cyclers needed. Some of this could easily be streamlined by local overrides to the suits, but this gives me an idea; Aurora.

See, Aurora has a lot of fairly good suit designs and they're also licensed under GPLv3, being another bay/polaris downstream. Some of their void and/or hard suit designs could probably be repurposed for the ERT to give them a distinctive appearance.

Maybe take a look at their [suit.dmi](https://github.com/Aurorastation/Aurora.3/blob/master/icons/mob/suit.dmi) and see if anything takes your fancy? I'm not a master, but I can easily tweak some highlights and do tails at least.

I could also do the same thing I did with #8138 so that it could autoadapt, saving both potential users (and me) a fair bit of energy.

Novacat:
You would be surprised at how many issues have come up with people trying to use Hardsuits. Synths have to replace the oxygen tank with a suit cooling unit, Hardsuits themselves are also vulnerable to EMP. Suit Cyclers are also fairly easy to use as well. There was an event where a GM mandated the players to use hardsuits and there was a lot of fiddling around with them, and one of the characters actually died because they were a synth and did not know about the lack of suit cooling unit on the standard hardsuit.

This is also not gut feeling for me, so much as actual field experience. I primarily play Pathfinder, Field Medic, and Emergency Response, so I heavily use both hardsuits and voidsuits. My experience is that the latter is infinitely easier to use and work with. Most players know how to use a suit cycler, not very many know how to do maintenance on a hardsuit.

As far as adding new voidsuits: I am not picky, if you want to try adding ERT Voidsuits (with species sprites) then go right ahead, I would certainly approve them. If that is done, I would approve of moving the prototype voidsuit back to the Head of Security. However, the stats on the Head of Security voidsuit would have to match the standard security voidsuit, due to issues we have had in the past. We do not want heads of security on the front lines of any major situation, period. They are supposed to be commanders, not commissars.

KillianKirilenko:
Well, the protosuit's stat tweaks are easy.

I'm rolling with Aurora's HoS design for now. It already has most of the species variants done (no tesh variant, but they don't get ERT hardsuits either), I just need to do helmet+tail for vulp and then the highlights. Won't take too long.

Demo;
image

KillianKirilenko:
Progress:
ert-voidsuits

Mostly done, I think. Just need to finish up the command colors and some minor detailing.

KillianKirilenko:
Done and integrated.

code/modules/clothing/spacesuits/void/military_vr.dm
```diff
@@ -15,13 +15,18 @@
 
 /obj/item/clothing/head/helmet/space/void/merc/prototype
```
Novacat, line 16: Rather than override the stats, it would be better to just re-path it into a security voidsuit subtype.

KasparoVy:
Are Serg sprites out of the equation? These suits look goddamn LEGIT as hell

Thankfully serg sprites are straight copy paste of Unathi with only the tail being custom

KillianKirilenko:
If someone else wants to take care of the tails, sure. I'd base them off the Sec EVA suit, but that still has a unathi-style tail and the sergal tails are a bit more complicated than I'm confident with tackling.

code/modules/clothing/spacesuits/void/military_vr.dm
KillianKirilenko: Done.

PR #8309, opened 2020-06-18
Prevent Refitting AutoLok Suits

KillianKirilenko:
Realized it was technically possible to refit them as they're a subtype of the /void/ type. This would be a bad idea anyway as they'd almost certainly turn invisible, so hey, not allowed any more.

### DIFF
--- a/code/game/machinery/suit_storage_unit.dm
+++ b/code/game/machinery/suit_storage_unit.dm
@@ -778,6 +778,13 @@
 			to_chat(user, "You cannot refit a customised voidsuit.")
 			return
 
+		//VR ADDITION BEGINS
+		//Make it so autolok suits can't be refitted in a cycler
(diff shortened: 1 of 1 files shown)


code/game/machinery/suit_storage_unit.dm
```diff
@@ -778,6 +778,13 @@
 			to_chat(user, "You cannot refit a customised voidsuit.")
 			return
 
+		//VR ADDITION BEGINS
```
Novacat, line 781: I would recommend using //VOREStation Edit specifically (You can add more to it, but //VOREStation Edit is consistent and allows me to search for '//VOREStation' when searching for edits.

KillianKirilenko: Done.

PR #8311, opened 2020-06-19
[Map] Minor Pipefix

TheFurryFeline:
Changelog Notes:

- Surface 2 Scrubber/Air vent switched properly so they circulate/scrub.

- Surface 1 area near lift leading to atrium scrubber moved out from under door.


TheFurryFeline:
Potential minor conflict with https://github.com/VOREStation/VOREStation/pull/7725 but easily resolved should either get merged first.

PR #8312, opened 2020-06-19
Adds Shipping Value to Plastitanium

KasparoVy:
Was waiting on Zuhayr's blessed upstream PR.
There we go, this should cover the last of our valuable refined materials.


KasparoVy:
@Lbnesquik Does this look alright to you?

PR #8313, opened 2020-06-19
Makes Nevrean use Sergal Void & RIG Suit Sprites

KasparoVy:
Nevreans have big tails and big beaked faces - they also used to have their own sprites (which afaik were straight copy-pastes of Unathi suits anyway)  but were changed to use Human suit sprites that really don't fit.

This switches them to use the Sergal sprites which fit significantly better.
Additionally, Nevreans must re-fit their voidsuits using suit cyclers again. _**>>>Choose the Sergal setting.<<<**_

virgo ss13 NEVREANS HAVE BEAKS DUDE CMON 20200618


KasparoVy:
This doesn't change current behaviour where Nevreans don't need to use the suit cycler to fit the thing. Should that be re-instated too?

Novacat:
I would rather things be consistent and that if they are changed to sergal sprites, than they should use sergal refits.

KasparoVy:
Done and tested, thanks @Novacat

PR #8322, opened 2020-06-22
Update Suit Cyclers

KillianKirilenko:
Earlyport of [PS7301](https://github.com/PolarisSS13/Polaris/pull/7301); improves suit cyclers slightly by adding a 'no change' cycler variant and fixes a weird edge-case scenario where helmets might not be modified properly if attached to a suit.

Users beware that cycling special rare exploration/POI voidsuits using the No Change option (such as the streamlined, salvage, heavy duty atmos, etc.) may cause them to vanish, as species variants were never made for them. Not much I can do about this beyond a concerted effort to fill out holes in the lineup (though [PS7300](https://github.com/PolarisSS13/Polaris/pull/7300) covers the streamlined suit, partially).

### DIFF
--- a/code/game/machinery/suit_storage_unit.dm
+++ b/code/game/machinery/suit_storage_unit.dm
@@ -1050,8 +1050,10 @@
 
 	if(target_species)
 		if(helmet) helmet.refit_for_species(target_species)
-		if(suit) suit.refit_for_species(target_species)
-		if(suit.helmet) suit.helmet.refit_for_species(target_species)
+		if(suit) 
+			suit.refit_for_species(target_species)
+			if(suit.helmet)
+				suit.helmet.refit_for_species(target_species)
 
 	//Now "Complete" with most departmental and variant suits, and sorted by department. These aren't available in the standard or emagged cycler lists because they're incomplete for most species.
 	switch(target_department)
--- a/code/game/machinery/suit_storage_unit.dm
+++ b/code/game/machinery/suit_storage_unit.dm
@@ -586,7 +586,7 @@
 	var/electrified = 0
 
 	//Departments that the cycler can paint suits to look like.
-	var/list/departments = list("Engineering","Mining","Medical","Security","Atmospherics","HAZMAT","Construction","Biohazard","Emergency Medical Response","Crowd Control","Security EVA","No Change")
+	var/list/departments = list("Engineering","Mining","Medical","Security","Atmospherics","HAZMAT","Construction","Biohazard","Emergency Medical Response","Crowd Control","Security EVA")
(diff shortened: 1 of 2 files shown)


code/game/machinery/suit_storage_unit.dm
```diff
@@ -1211,11 +1222,22 @@
...
 		var/obj/item/clothing/S = new parent_suit
 		suit.name = "refitted [initial(parent_suit.name)]"
 		suit.desc = initial(parent_suit.desc)
 		suit.icon_state = initial(parent_suit.icon_state)
 		suit.item_state = initial(parent_suit.item_state)
 		suit.item_state_slots = S.item_state_slots
 		qdel(S)
+
+		//can't believe I forgot to fix this- now helmets will properly cycle if they're attached to a suit -KK
+		if(suit.helmet && target_department != "No Change")
```
Rykka-Stormheart, line 1235: This needs one tab removed to be on-line with the other if checks, I think.

KillianKirilenko: No, because if no suit is fitted into the cycler it'll try to read the mounted helmet value from a `null` suit, which will (harmlessly) runtime.

Novacat:
I am not really certain on this because a large part of the reason why suit cyclers have no cycle options for certain voidsuit types is *because* there are no species sprites for those voidsuits. I would rather very heavily avoid the possibility of broken sprites with voidsuits because that tends to cause people to ahelp/bug report when it happens.

KillianKirilenko:
Yeah, it's definitely a risk... a possible solution would be to extend the no-cycle system similar to the autolok/mark7, except make it a function of a helmet/suit-side `can_cycle` var instead. That way if you try to put an incompatible suit in, you get a `"the suit cycler cannot accept that item"` message and don't have to worry about things turning invisible. Thoughts?

Novacat:
That would be more amenable, though there is still the issue with some suit types having no teshari sprites.

KillianKirilenko:
Right. I *think* the best way to handle that would be to determine the list of valid species to recycle to via the suit/helmet and then if you try to cycle to an invalid type, it refuses to operate instead... 🤔

For the moment I think I'll make the "No Change" functionality inaccessible on all but the `refit_only` variant (since that won't be mapped in yet) and see if I can figure out better bug-proofing later.

Leshana:
@KillianKirilenko Is this completely obsoleted by https://github.com/VOREStation/VOREStation/pull/8336 ?

KillianKirilenko:
Pretty much, yeah.

Leshana:
OBE

PR #8326, opened 2020-06-22
Support actual VGS environment management.

Leshana:
- Extends VGS DMAPI to include Reboot and Init Complete messages.
- Add pid file and symlinked datafolder to git ignore.

This is backwards compatible and has no effect on servers not using VGS.  TGS will continue to work uninterrupted as well.


KasparoVy:
Wicked.

PR #8327, opened 2020-06-22
Adds bridge protection turret to Talon

Cyantime:
The talon has a number of point defense turrets scattered around it, but is conspicuously missing any that would protect the most important and also vulnerable spot on it.
This sticks a turret there.

turret


Cyantime:
Giving travis a boot

Cyantime:
oh, did that not work

Cyantime:
trav

Why is this running three times

Leshana:
This PR is broken. I don't see a way to easily remove the extra two checks that will not run.   Please open as a new PR while github is not drunk.

PR #8328, opened 2020-06-23
Reduce the default escape time for bellies from 60 to 20s.

lbnesquik:
It should make it easier for people to escape mobs not completely burnt despite having rolled a success. You can always change it for players.


Rykka-Stormheart:
This is the wrong file, this is legacy bellies

Rykka-Stormheart:
You want to modify code/modules/vore/eating/belly_obj_vr.dm @lbnesquik

Screemonster:
Less time in the do-nothing corner  when you don't want to be in the do-nothing corner is good, and frankly when it comes to vore mechanics and mobs and other vorey fun like that the last thing on my mind is whether or not escaping is balanced for combat. Maybe that was a concern back when voremobs would _only_ attempt to eat you and would stop attacking, so lying down and letting yourself get eaten was effectively a free stun while your buddies shot at it, but _really_

PR #8332, opened 2020-06-24
Grant brig access to the detective as a .vr tweak

lbnesquik:
Less headaches of low access rights and honestly, they should just be able to walk wherever a junior officer can, given they are sort of higher on the hierarchy.

### DIFF
--- a/code/game/jobs/job/security.dm
+++ b/code/game/jobs/job/security.dm
@@ -77,7 +77,7 @@
 	spawn_positions = 2
 	supervisors = "the Head of Security"
 	selection_color = "#601C1C"
-	access = list(access_security, access_sec_doors, access_forensics_lockers, access_morgue, access_maint_tunnels, access_eva, access_external_airlocks)
+	access = list(access_security, access_sec_doors, access_forensics_lockers, access_morgue, access_maint_tunnels, access_eva, access_external_airlocks, access_brig) //Vorestation edit - access_brig
 	minimal_access = list(access_security, access_sec_doors, access_forensics_lockers, access_morgue, access_maint_tunnels, access_eva, access_external_airlocks)
 	economic_modifier = 5
 	minimal_player_age = 3


code/game/jobs/job/security.dm
```diff
@@ -77,7 +77,7 @@
 	spawn_positions = 2
 	supervisors = "the Head of Security"
 	selection_color = "#601C1C"
-	access = list(access_security, access_sec_doors, access_forensics_lockers, access_morgue, access_maint_tunnels, access_eva, access_external_airlocks)
+	access = list(access_security, access_sec_doors, access_forensics_lockers, access_morgue, access_maint_tunnels, access_eva, access_external_airlocks,access_brig) //Vorestation edit - access_brig
```
Heroman3003, line 80: no space after comma reee my consistency

lbnesquik: :gun:

Screemonster:
This would have saved me having to yeet someone over the railing that time I was solo detective.

For this reason, I am torn, because having an excuse to oubliette a motherfucker is hilarious.

MrSign:
No issues with this PR. Having access to the brig for all members of security will save a lot of headaches. Specially if a detective is the only member on and has to arrest someone in the event of a crime.

lbnesquik:
It doesn't seem like a very controversial PR.

Screemonster:
We're looking at redoing some stuff with SOP and so on. Generally we're looking to make the relationship between detective and security to be similar to that between chemist and the rest of medical, ie. there are limited slots compared to the entire department, the detective doesn't _have_ to do patrolling stuff or break up barfights 'cause it's not their job in the same way that a chemist doesn't _have_ to do surgery, but if nobody else is in the department then they can do so if they feel confident. Likewise, if there's no chemist around then MDs can do chemistry, but it's assumed that they should let the chemist handle it if they're around. This will probably mean mapping a side-door onto the forensics lab so officers can get in there without having to go through the detective's personal office.

MrSign:
So, about that SOP redo...

MrSign:
THIS CAN BE MERGED NOW!!!

PR #8340, opened 2020-06-26
Fluff armband for the recent event

dwinters99:
The second-to-latest (as of writing) lore post in #free-anur-tribune speaks of a red armband worn on the lower left arm as a identifying marker of revolutionaries. Adding this in as an item will allow, for one, any of these revolutionaries in-game to be wearing it, in addition to being a fluff for the big contributors in the event. 

As it is right now, it hasn't been added to the loadout. If staff gives the go-ahead, it can be added to the fluff page in loadout for the ckeys whose participation was crucial in the event.


KasparoVy:
Can't just use the colorable armband and make it red eh? Afaik there's a red armband under the selection too

dwinters99:
Normal armband is on the upper arm. I'm obsessive so I need it on the _lower arm._

On that note, I need to check if it gets picked up by the normal armband selection loadout item, and if so, fix that.

code/modules/clothing/under/accessories/accessory_vr.dm
```diff
@@ -302,3 +302,10 @@
 	item_state = "silverthree"
 	overlay_state = "silverthree"
 	desc = "A silver medal awarded to a group which has demonstrated exceptional teamwork to achieve a notable feat. This one has two bronze service stars, denoting that it has been awarded three times."
+
+/obj/item/clothing/accessory/armband/democracy
+	name = "democracy armband"
+	desc = "A red armband worn on the lower left arm. Worn by members of anti-monarchy Red Armband Movement in An-Takh-Et "
```
KasparoVy, line 308: Extra space on the end

KasparoVy:
That's dope. Means you can wear TWO RED ARMBANDS for DOUBLE THE REVOLUTIONS

KillianKirilenko:
Nah, they're probably both gonna conflict over the armband-specific accessory slot.

Besides, that way lies people wearing *four* armbands.

KasparoVy:
not wearing enough armbands to form a full sleeve, fuuuuckkkkkkk

PR #8341, opened 2020-06-26
Removes second vocal annoucement from grub event

Cyantime:
This stops the delayed "this is where the grubs are" announcement from replaying the entire "XENOS HERE, WELD THE VENTS" vocal message.

### DIFF
--- a/code/modules/events/grubinfestation_vr.dm
+++ b/code/modules/events/grubinfestation_vr.dm
@@ -42,4 +42,4 @@
 		area_names |= grub_area.name
 	if(area_names.len)
 		var/english_list = english_list(area_names)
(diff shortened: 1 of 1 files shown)


TheFurryFeline:
I'm on the fence with this. While it might be annoying to have to hear that alert again, it still emphasises that they need to be found and dealt with.

Cyantime:
I think that the text indicating exactly where each grub is does a good enough job of that. By the time this plays it's well past time to "seal all ducting and ventilation" because the grubs are full size.

TheFurryFeline:
True enough there. Maybe have a new audio added about "Grub locations are listed on your announcement consoles" or whatever. I know you can view announcements on your PDA, definitely.

Cyantime:
The text message is supposed to be the IC audio, being played over the station speakers/intercoms/whatever. Having someone actually say "hey look at your chat" is sorta dumb. The locations are already being stated, there's no reason to go to a console or PDA or something.

PR #8342, opened 2020-06-27
Removes Vox speech sounds

Heroman3003:
Oh god speech sounds why who thought it was a good idea

Also sets default language to Galcom because they are literally only non-event playable species that doesn't have it set to that.


AaronSherwood117:
yes please thank you!

lbnesquik:
Correct this to 'Disable Vox speech sounds'

CRTXBacon:
Could be an idea to move these to emotes rather than outright getting rid of them

AaronSherwood117:
> Could be an idea to move these to emotes rather than outright getting rid of them

They're already on emotes, someone just decided to start making them proc on every message for some reason :)

KasparoVy:
remove everything reeee

Zeldazackman:
I've been running a copy of this downstream and I haven't had vox give off noises to their speech unless they explicitly use an emote to do so

PR #8343, opened 2020-06-27
Removes Sam Poster

SweetBlueSylveon:
Removes /datum/poster/vore_7.


Ketrai:
y tho?

Screemonster:
> y tho?

because sam never asked for it

PR #8363, opened 2020-07-05
AR Glasses Tweaks

KillianKirilenko:
Featurecreeped the hell out of a simple idea; originally I wanted to make AR sunglasses, but figured that would be way too much clutter on the loadout menu and having a selector would be a little pointless.

So, instead, this PR now expands on AR glasses functionality with the following features;
- Adds omni-alerts (power/atmos/camera/etc.) nanoUI access to the AR-B glasses.
- Amended various descriptions to clarify all available functions/features.
- Adds AR-R glasses to loadouts for R&D department jobs, and changes them to look like regular AR glasses rather than funky purple goggles. Sorry, they still don't do anything useful. Would be nice if they did though.
- Reduces AR-S eye protection to sec sunglasses flash-only level rather than full flash+weld protection. I felt it was weird that AR-S had flash+weld compared to most regular sec eyewear being flash-only. AR-Bs still have full flash+weld.
- Adds the ability to toggle AR glasses between standard/sunglasses modes via a verb on the Objects tab. Only works for basic types, not fluff subtypes. **Purely cosmetic:** does not alter their eye protection values.
- Adds a 10% chance for EMP'd AR glasses to change their icon_state to red and blue 3d glasses. Harmless fun! Reset them with the shading verb, or leave them funky.
- Adds the retinal meson projector to loadout options for the same departments as regular mesons, and removes eye protection values from it. This was added a long time ago but never featured in loadouts or anything.


Heroman3003:
Having every type of AR be immune to flashes sounds like overkill imo. I understand omni/sechud at least...

Also I will reeeee on change to science AR. The only reason for those to exist was funky alternative look from basic!

KillianKirilenko:
> Having every type of AR be immune to flashes sounds like overkill imo. I understand omni/sechud at least, but others?

The shading verb is purely cosmetic, I haven't changed flash immunity for the standard, AR-M, or AR-R variants. I briefly considered *adding* welder protection to AR-R (for roboticists, primarily) but decided not to.

Heroman3003:
IIRC AR-R isn't actually obtainable outside of spawning in science, and its moreso that now its absolutely no different from regular AR glasses.

As for the shading, might wanna clarify it in the PR post that its entirely cosmetic.

KillianKirilenko:
I'd like to give them some kind of feature to be honest, and I contemplated giving them welder protection for robotics -- but plenty of AR glasses have flash/weld protection already.

I also think it's misleading to have them look like goggles when they don't protect your eyes from airborne contaminants like goggles do.

As for the shading being cosmetic; only sec/com shades have offered flash protection for a long time. I've amended descriptions and the OP appropriately though.

Rykka-Stormheart:
IMO, I support AR-R glasses having welding protection.

MrSign:
Great staff discussion, just need to fix the conflict.

KillianKirilenko:
rip nanoui, long live tgui

running a quick compile+run test to make sure the tgui module doesn't freak out if it's called from glasses

PR #8366, opened 2020-07-05
Bring back hemophilia. No, it's not a crime!

lbnesquik:
It is a negative trait that makes you bleed a loss.
It comes with a hacked together system to deny synths from taking it. It was a PAIN.

### DIFF
--- a/code/modules/client/preference_setup/vore/07_traits.dm
+++ b/code/modules/client/preference_setup/vore/07_traits.dm
@@ -14,6 +14,7 @@
 	var/traits_cheating = 0 //Varedit by admins allows saving new maximums on people who apply/etc
 	var/starting_trait_points = STARTING_SPECIES_POINTS
 	var/max_traits = MAX_SPECIES_TRAITS
+	var/dirty_synth = 0		//Are you a synth
 
 // Definition of the stuff for Ears
 /datum/category_item/player_setup_item/vore/traits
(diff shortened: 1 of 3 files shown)


code/modules/client/preference_setup/vore/07_traits.dm
```diff
@@ -255,6 +260,14 @@
 
 			var/conflict = FALSE
 
+			user.isSynthetic()	//Recheck just to be sure
```
ShadowLarkens, line 263: Is this meant to be `pref.dirty_synth = user.isSynthetic()`?

lbnesquik: 🤔 That wasn't what i was trying to do, but it might work just as well if not better.

lbnesquik: Tested, didn't seem to quite work.

ShadowLarkens: Hrm, it just seems like this line doesn't actually do *anything* right now, I don't think isSynthetic() has any side effects?

KasparoVy:
Blood for the blood god

lbnesquik:
Staff discussion required?

nerdass:
what happens if someone selects the trait and then picks fullbody prosthetic in character setup?

KasparoVy:
They bleed to death, white synth blood everywhere!!! Aaaaaaa

lbnesquik:
> 
> 
> what happens if someone selects the trait and then picks fullbody prosthetic in character setup?

That's why there is the synth checks

lm40:
provided the code is solid this seems good

SilencedMP5A5:
yee

nerdass:
no i meant like, AFTER they select the trait.

MrSign:
Good staff discussion. Very insightful. MERGE!

KasparoVy:
Linkin Park - Bleed it Out lmao

PR #8367, opened 2020-07-06
Update robolimbs_vr.dm

### LINKED ISSUE #8356: Zeng-Hu prosthesis are missing from character creation
I play a character called Dayleen-A who is full Zeng-Hu FBP.
A while ago i noticed that my character did not look like she used to anymore, Now with some weird green lines i can not get rid off no matter what i do. Turning on and off, changing and swapping synth colors or synth markings do nothing, Not to mention this weird green color is there with synth color and markings OFF by default. Not only that but the normal markings of the Zeng-Hu body are gone and i was left with a sort of hybrid between a normal body and a Zeng-Hu with no markings.
Now, this wouldn't be an issue if i could try to re-set my FBP right? Well, Zeng-Hu prosthesis are completely gone from character selection and they cannot be selected, Meaning that if i remove my Zeng-Hu things from character selection and press save, they will be gone and i wont be able to select them again!

https://i.imgur.com/EDQxAHv.png Weird green lines and the markings that were part of the prosthesis to begin with are gone. Also this is NOT fixed by being re-sleeved in round or building a FBP from the ground up.

https://i.imgur.com/nkQYa4n.png As you can see here (I cant find any other pictures cuz i dont usually take any) There is no weird green markings and the normal Zeng-Hu markings are there and visible. This picture is not very old and other people have told me they cant select Zeng-Hu in character creation either.

KillianKirilenko:
Resolves a conflict in company names thanks to datum configs, should fix #8356.

### DIFF
--- a/code/modules/organs/robolimbs_vr.dm
+++ b/code/modules/organs/robolimbs_vr.dm
@@ -49,7 +49,7 @@
 
 // tucker0666 : Frost
 /datum/robolimb/zenghu_frost
-    company = "Zeng-Hu"
+    company = "Zeng-Hu (Custom)"
     desc = "This limb has realistic synthetic flesh covering with 'blue accents'."
     icon = 'icons/mob/human_races/cyberlimbs/_fluff_vr/Frosty.dmi'
     blood_color = "#45ccff"


KasparoVy:
Nice catch

PR #8377, opened 2020-07-09
Add a smell system, closely matching the flavor system.

lbnesquik:
The only real difference is that you can't smell reagents on people, as they don't have a smell ATM, but at the same time, how often did that really come to happen.

### DIFF
--- a/code/modules/vore/eating/living_vr.dm
+++ b/code/modules/vore/eating/living_vr.dm
@@ -376,7 +376,7 @@
 		return
 
 	setClickCooldown(DEFAULT_ATTACK_COOLDOWN)
-	visible_message("<span class='warning'>[src] smells [smelled]!</span>","<span class='notice'>You smells [smelled]. They smell like [smelled.get_smell_message()].</span>","<b>Sniff!</b>")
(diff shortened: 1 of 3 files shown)


KasparoVy:
The snozzberries smell like snozzberries too now

code/modules/vore/eating/living_vr.dm
```diff
@@ -356,6 +360,43 @@
...
+	set desc = "Smell someone nearby!"
+	set popup_menu = FALSE
+
+	if(!istype(smelled))
+		return
+	if(!checkClickCooldown() || incapacitated(INCAPACITATION_ALL))
+		return
+
+	setClickCooldown(DEFAULT_ATTACK_COOLDOWN)
+	visible_message("<span class='warning'>[src] smells [smelled]!</span>","<span class='notice'>You smells [smelled]. They smell like [smelled.get_smell_message()].</span>","<b>Sniff!</b>")
```
dwinters99, line 379:
```suggestion
	visible_message("<span class='warning'>[src] smell [smelled]!</span>","<span class='notice'>You smell [smelled]. They smell like [smelled.get_smell_message()].</span>","<b>Sniff!</b>")
```

lbnesquik: Corrected that, cheers.

PR #8397, opened 2020-07-13
Clean up Upstream Mirror for Speedmerge

Rykka-Stormheart:


Leshana:
I think we ought to keep our files in proper tgm map format, instead of the weird one that Polaris is using. I verified that the maps are the same other than format.

Issue #8394, opened 2020-07-13
Bug - Something breaks Life() via a list index out of bounds error


Rykka-Stormheart:
#### Brief description of the bug

Life() breaks for all mobs as a result of a runtime. Runtime in mobs.dm,46: list index out of bounds.
The line is: ```else if(M.low_priority && !(process_z[get_z(M)]))```

get_z(M) will return 0 for mobs in nullspace

If a mouse is in someone's pocket, and the person is digested, the mouse will be forcemoved to nullspace, which causes the above error, and subsequently breaks Life() for every single mob lower in the list than itself.

https://github.com/VOREStation/VOREStation/blob/master/code/modules/vore/eating/belly_obj_vr.dm#L406-L424 This may potentially be the issue?

#### What you expected to happen

To not break the entire Life() proc

#### What actually happened

Digesting my girlfriend broke Life

#### Steps to reproduce

Eat someone who has a mouse in their pocket. Digest them, and watch as the mouse fails to appear in your belly, and you subsequently break the entire server as the mouse exists in nullspace.

#### Code Revision
- (Found using the "Show Server Revision" verb under the OOC tab.)
 

#### Anything else you may wish to add:
- (Location if it's a mapping issue, screenshots, sprites, etc.)

Rykka-Stormheart:
Closing, this was resolved.

PR #8412, opened 2020-07-14
Quick fix to stop mobs in nullspace runtiming the mobs and ai subsystems.

Leshana:
For _Runtime in ai.dm,38: list index out of bounds_ and _Runtime in mobs.dm,46: list index out of bounds_

### DIFF
--- a/code/controllers/subsystems/ai.dm
+++ b/code/controllers/subsystems/ai.dm
@@ -32,7 +32,7 @@ SUBSYSTEM_DEF(ai)
 	while(currentrun.len)
 		var/datum/ai_holder/A = currentrun[currentrun.len]
 		--currentrun.len
-		if(!A || QDELETED(A) || A.busy) // Doesn't exist or won't exist soon or not doing it this tick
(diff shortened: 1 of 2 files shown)


Leshana:
Note:  Theoretically an error still could happen if the following happens in _exactly_ this order:
1) Subsystem starts firing, then sleeps.
2) A new z-level is allocated.
3) A mob that previously existed _prior_ to step 1 but was **not** processed _during_ step 1 is moved to the new z-level.

But the error would clear itself after one cycle.

Leshana:
Note:  Theoretically an error still could happen if the following happens in _exactly_ this order:
1) Subsystem starts firing, then sleeps.
2) A new z-level is allocated.
3) A mob that previously existed _prior_ to step 1 but was **not** processed _during_ step 1 is moved to the new z-level.

But the error would clear itself after one cycle.

Rykka-Stormheart:
Note, fixes https://github.com/VOREStation/VOREStation/issues/8394

Leshana:
Updated so that low priority mobs in nullspace won't have Life() called.

PR #8421, opened 2020-07-17
Adds Lunar Fleet and Jademanique to whitelists

LunarFleet:
I figured out how the custom borg sprite system works, finally. I have added more comments to the system so it is easier for future readers to figure out. In total, this involves three SCAs
1. Clea-Nor's custom sprite https://forum.vore-station.net/viewtopic.php?f=46&t=1742
2. B.A.U-Kingside's custom sprite https://forum.vore-station.net/viewtopic.php?f=46&t=1576
3. Areli Leysa's gutter language application https://forum.vore-station.net/viewtopic.php?f=46&t=1781

Also, I changed the separator in the borg sprite whitelist, changing it to | from - because a lot of the borg I've seen, INCLUDING both of the accepted borg sprite SCAs, have dashes in their name. 

If this needs to be broken into three different PRs just say so and I'll do it; I just figured this would be easier for the time being.


LunarFleet:
I assume you mean just the robot custom sprites.dm, as that is the only one with changed code?

KasparoVy:
RIP icondiffbot

LunarFleet:
Alright, I PR'd the custom_sprites.dm file over to Polaris now as well. 
I've also tested the whitelist system near religiously, so, it should work fine.

LunarFleet:
This seems kind of redundant, at least to me. The _only_ thing custom_synethic.dmi is used for is custom borg sprites; unless we wish to add polaris' whitelisted borgs, why is this necessary?

Novacat:
The reason why is if Polaris ever edits custom_synthetic.dmi, that the changes do not accidentally overwrite our custom_synthetic.dmi

LunarFleet:
Well, alright, but, I'm not quite sure how to do that; the easiest way to do it would be to change custom sprites.dm to use the VS file instead of the normal one, but, that seems counter productive given the fact that you asked me to push the custom sprites file over to Polaris too.

Novacat:
The idea behind pushing it upstream is that mostly the code is similar enough that if Polaris does make changes to that code down the line, it does not cause massive conflicts with our changes. Future proofing, if you will.

Then just need to have a //VOREStation Edit for the dmi, and since it is marked as a Vorestation edit, if Polaris touches that area of the code I will know how to deal with it.

KasparoVy:
Yay

PR #8424, opened 2020-07-17
Excursion Blasters (WIP)

KillianKirilenko:
Dusting off an old project after some passing discussion in lobby. This PR increases weapon flexibility for expedition teams by adding a selection of 'excursion blasters'. These weapons use self-charging logic to mimic an overheating system, slowly cooling down of their own accord. They all use the same ID-lock logic as the phasers.

Includes four variants; pistol, carbine, rifle, and DMR;
blasters

Pistol: Semi-Auto + 2-Burst. Fastest charge (comparable to the captain's laser) but quite inefficient.
Carbine: Semi-Auto + 3-Burst. Fairly inaccurate, and burst mode is extremely inefficient. OK recharge rate (1.5x rate+delay of pistol). Medium. Takes up quite a bit of space in packs.
Rifle: Semi-Auto + 3-Burst. Good all-rounder, but it has the second-slowest recharge rate (2x rate+delay of pistol). Large. Takes up a lot of pack space.
DMR: Semi-Auto only. Fitted with a 1.5x scope. Very low capacity but good accuracy. Extremely slow recharge (3x rate+delay of pistol). Huge, can't be put in backpacks at all, maybe not even bluespace bags.

All values of the weapons and means of acquisition are up for tweaking, I'm just getting this out there to gauge interest. Could probably do with unique sprites (and proper on-mob sprites) as well.


lbnesquik:
So those are infinite ammo?

And doesn't those sort of completely override the current phasers? You might as well replace or integrate the current ones into your pr. 

How are those acquired exactly? I feel like R&D and cargo aren't the best solution 
I do like the sprites quite a bit!

KillianKirilenko:
> So those are infinite ammo?

Short answer: I feel that any weapon for expeditions must necessarily be self-charging or easily recharged in the field due to the vagaries of RNG-based combat in order to remain competitive with phasers.

> And doesn't those sort of completely override the current phasers? You might as well replace or integrate the current ones into your pr.

Yes and no. Compared to phasers, the blasters are supposed to take longer to fully recharge (particularly at the high end with the DMR) but have the main benefit of doing so by themselves. Essentially you can carry two blasters (a rifle and a pistol, for instance) or a blaster and a phaser, switching as-needed, and use your machete or bootknife as a last resort.

From what I remember of my initial testing, blasters have a higher 'burst' DPS (largely thanks to their burst modes, though the carbine's is pretty unreliable beyond 3-4 tiles) but will run dry quite quickly if you just spam burstfire. Phasers by comparison are more steady and reliable, but require a long, immobile downtime period to recharge.

Also I think I made them standalone because I thought phasers were a Polaris thing at the time and didn't want to touch them. As that's not actually the case on second inspection, well... I guess we'll see what happens.

> How are those acquired exactly? I feel like R&D and cargo aren't the best solution

Up for debate, really. Most of the PR is! I just wanted to get it out there to spark some debate/discussion.

Novacat:
Hm.

My biggest concern with these is that they might completely invalidate AEGs, which is the whole reason expedition phasers have the crank charger to start with. Also self-recharging technology is exceptionally bulky, which is why only the AEG has it.

KillianKirilenko:
> My biggest concern with these is that they might completely invalidate AEGs, which is the whole reason expedition phasers have the crank charger to start with. Also self-recharging technology is exceptionally bulky, which is why only the AEG has it.

The AEG doesn't require any particularly high tech levels though, and not much in the way of special materials. You've basically got what, a small nuclear core on it? I feel like that's child's play when the setting is in the 2300s, with FTL, animate nanoblobs, and dog-shaped robot buddies.

Anyway, practically speaking though, what I *could* do is treat the AEG is the early prototype/precursor model, and these are more refined prototypes? 🤔

KasparoVy:
Overclock that DMR and have it give a chance to delimb and straight up explode in your hands, MUAHAHAH!

Novacat:
Some concerns and questions.

I would like the weapons to have more... distinct weapons fire? They seem to borrow a lot from existing weapons and not in a very consistent manner either (such as the Carbine using the Xray beam color).

Also, this PR will likely be looked over by other staff since there is the question of how common we want self-recharging weapons to be in the lore and universe.

KillianKirilenko:
> I would like the weapons to have more... distinct weapons fire? They seem to borrow a lot from existing weapons and not in a very consistent manner either (such as the Carbine using the Xray beam color). Admittingly this is also a Polaris problem too.

Yeah I basically just grabbed those as they were handy at the time. The long and short of it is that I want it to be very, very clear who's using what. Colorcoding the beams was a no-brainer.

Cyantime:
But, why? No other beam weapons have their projectiles set up like that. Strong basic lasers get wide red, all other normal lasers are just red. Special color beams are generally reserved for special weapons, not to colorcode your squad roles.

Novacat:
My point is that all lasers that do a similar thing should look similar. All normal lasers are red with 40 damage. The Energy Sniper and the X-Ray rifle use green lasers, and have different effects (The energy sniper probably should have its own unique effect), Laser cannons also have their own unique projectile. Stun shots use a yellow electric effect.

The overall idea of what I am suggesting is that, you should be able to, at a glance, know what is being shot at you.

Screemonster:
Explorers aren't really meant to be a paramilitary force, as much as some explorers might like it to be that way.

Screemonster:
I wouldn't mind them being available as _rare_ loot somewhere that the station _could_ obtain, just not mass-produce. Abandoned crates have been in dire need of some love, for instance.

KillianKirilenko:
> My point is that all lasers that do a similar thing should look similar. All normal lasers are red with 40 damage. The Energy Sniper and the X-Ray rifle use green lasers, and have different effects (The energy sniper probably should have its own unique effect), Laser cannons also have their own unique projectile. Stun shots use a yellow electric effect.
> 
> The overall idea of what I am suggesting is that, you should be able to, at a glance, know what is being shot at you.

Right. Good. That's actual, useful feedback that I can work with. I'd like the beams to still be a little distinctive by weapon, but that can be done with adjusting the width/intensity of the beams, retaining color as an indicator of relative firepower.

> Explorers aren't really meant to be a paramilitary force, as much as some explorers might like it to be that way.

Yeah that's not a mindset I want to encourage either, but right now your choices are ranged or melee. And unless you're really good at (exploiting) the combat systems, melee is how you get fucked up faster than you can blink. At the very least a little variance in choices isn't that bad, is it? Here's an option that charges fast but is inaccurate outside of very short range, here's an option that charges slower but is much more accurate, take your pick.

Everything else stems back to "too much fighting, not enough exploring" which is out of scope but definitely needs looking at.

> After a bit of staff discussion, I believe the best niche for these weapons will likely be ERT weapons. The ERT's energy weapon selection is really bad (especially compared to their ballistics) so this would be a nice improvement.

Looking at your ERT shuttle submap:
* Energy weapons: laser rifles, holdout eguns, an energy net, an energy DMR (which is trash†), ion pistols, ion rifles, eguns, adv. eguns, tasers, and stun revolvers.
* Ballistics: combat shotguns, a revolver, an adv. SMG, assault rifles, and .45 pistols.

_†It holds maybe four shots, and you can't swap the cell out in the field: you **have** to use a charger. God knows why._

Their ballistics selection is vastly inferior at this time but could easily be expanded, and their energy selection could easily be expanded with the addition of xray lasers, laser cannons, burst lasers, and laser scatterguns, which all already exist. I can forgive people for forgetting about the burst laser and scattergun though.

nerdass:
yeah these guns are cool and all but it's a bit silly to give them to explorers, however having BIG DICK ENERGY WEAPONS for the OH GOD OH FUCK squad? that I can get behind.

Novacat:
The ERT Energy Sniper actually is a special version that is not battery locked. But, it would be nice to have ERT weapons that do not require you to carry an entire box of extra batteries with you that is not an AEG. Also, the ERT used to have a much, much wider variety and heavier weapons, but the ERT Arsenal was nerfed about a year ago.

These weapons could probably replace the varedited Energy Sniper and the AEGs, while providing self-recharging alternatives to the eguns and laser carbines.

As Scree mentioned as well, these things can probably also be generated in as rare loot on expeditions and abandoned crates.

PR #8426, opened 2020-07-17
Added options for hair gradients to shapeshifter set hair and set hair color

### LINKED ISSUE #8420: Select Hair Colors issue for Prometheans
When changing the colors of hair as a Promethean, only the primary hair color changes while the secondary stays the same.

I expected the secondary hair color to change according to what I put in the color change menu.

#### What actually happened
It only changed the primary color of the hair instead of both the hair colors in the gradient. 

#### Steps to reproduce
- Be a Promethean
- Use the Gradient option (in my case, I used Fade (Up)) in the Character Setup menu.
- Use the select hair colors option then try and change both colors from what they were.

#### Code Revision
Server revision: B:release D:
Commit: 4c340c5dac0a7afa8f5f219ec3ba812c5ed67033 
 
Hair at start
image

Hair after changing primary to orange, and secondary to blue
image

cadyn:
Resolves #8420 by adding options to edit gradient color and gradient style to existing shapeshifter verbs.

### DIFF
--- a/code/modules/mob/living/carbon/human/appearance.dm
+++ b/code/modules/mob/living/carbon/human/appearance.dm
@@ -49,6 +49,22 @@
 	update_hair()
 	return 1
 
+/mob/living/carbon/human/proc/change_hair_gradient(var/hair_gradient)
+	if(!hair_gradient)
+		return
+
+	if(grad_style == hair_gradient)
+		return
+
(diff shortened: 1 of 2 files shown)


cadyn:
Gonna yeet this on over to polaris now because aaaaa

PR #8429, opened 2020-07-18
Consolidates solar shed SMES

Cyantime:
With recent changes to how power is distributed, there's even less reason to have multiple SMES out here other than taking up more space in the RCON console list. This combines the three SMES units into a single one.

### DIFF
--- a/maps/tether/tether-09-solars.dmm
+++ b/maps/tether/tether-09-solars.dmm
@@ -279,15 +279,18 @@
 /area/tether/outpost/solars_shed)
 "aC" = (
 /obj/effect/decal/cleanable/dirt,
-/obj/item/stack/cable_coil/lime,
-/obj/structure/table/standard,
-/obj/structure/cable/yellow{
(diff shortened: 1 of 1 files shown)


maps/tether/tether-09-solars.dmm
```diff
@@ -381,32 +378,31 @@
...
+/turf/simulated/floor/virgo3b,
+/area/tether/outpost/solars_outside)
+"aO" = (
 /obj/machinery/power/smes/buildable{
 	charge = 0;
+	cur_coils = 3;
 	output_attempt = 0;
 	outputting = 0;
-	RCon_tag = "Solar Farm - SMES 3"
+	RCon_tag = "Solar Farm - SMES 1"
```
Heroman3003, line 405: Would numbering like that make any sense when there's only one to begin with?

Cyantime: No, I just forgot to change it.

PR #8430, opened 2020-07-18
Added Canidaela custom species

cadyn:
Potentially to be added if custom species request gets accepted.


Rykka-Stormheart:
What's the difference between these and Vulpkanin?

cadyn:
> What's the difference between these and Vulpkanin?

https://forum.vore-station.net/viewtopic.php?f=45&t=1792

Rykka-Stormheart:
> > What's the difference between these and Vulpkanin?
> 
> https://forum.vore-station.net/viewtopic.php?f=45&t=1792
> 
> Pretty much totally different lore, and it's actually genetically derived from the Canidae species on Earth, and temperature resistant and darksight.

I don't want to smash your hopes, but you can do this with an existing custom species. You don't need to snowflake an entire new species if they're visually just vulpkanin and the difference is backstory, just use a Vulp sprite and do the backstory custom. I'm not a dev, mind, so this is just a matter of opinion and what I've seen.

cadyn:
> > > What's the difference between these and Vulpkanin?
> > 
> > 
> > https://forum.vore-station.net/viewtopic.php?f=45&t=1792
> > Pretty much totally different lore, and it's actually genetically derived from the Canidae species on Earth, and temperature resistant and darksight.
> 
> I don't want to smash your hopes, but you can do this with an existing custom species. You don't need to snowflake an entire new species if they're visually just vulpkanin and the difference is backstory, just use a Vulp sprite and do the backstory custom. I'm not a dev, mind, so this is just a matter of opinion and what I've seen.

Everything except for the Temperature at least, which gets to be a really annoying inconsistency for me after a while. I feel like it's not the worst possible thing to add another species to the list considering I've done all the work anyway. I'm open to it going either way, but having the species added as well would just be really nice!

Heroman3003:
The issue with adding new species with minimal/no mechanical differences is that it simply bloats the list needlessly. Species should either have distinct enough and important enough lore, appearance or mechanics (ideally a combination of all three). That's why the Zorren were made into one species, and Xenohybrids got rolled up into the Custom Species.

cadyn:
Yea actually, that's fair. I'm gonna be right back with a different pull request regarding new custom species traits instead yeet

PR #8431, opened 2020-07-19
Increases the amount of items a sci-borg gut-analyzer can consume to 10.

MrSign:
Because sometimes, even sci-borgs want to be greedy guts on station and gain a tum.


SilencedMP5A5:
can you do this for the K-mine doggos as well?

MrSign:
> can you do this for the K-mine doggos as well?

I won't since the miner module's belly is pretty much a medihound's med-belly. While the compactor and analyzers are just for items.

PR #8443, opened 2020-07-22
Double the dirt for half the price!

serenekerosene:
Changes dirt = min(dirt+1, 101) to dirt = min(dirt+2, 101), to add twice the dirt per dirt. Added for excitingly dynamic (any) Janitor gameplay!

### DIFF
--- a/code/game/turfs/simulated.dm
+++ b/code/game/turfs/simulated.dm
@@ -69,7 +69,7 @@
 
 /turf/simulated/proc/update_dirt()
 	if(can_dirty)
-		dirt = min(dirt+1, 101)
+		dirt = min(dirt+2, 101) // VOREstation edit
 		var/obj/effect/decal/cleanable/dirt/dirtoverlay = locate(/obj/effect/decal/cleanable/dirt, src)
 		if (dirt > 50)
 			if (!dirtoverlay)


Heroman3003:
First, this changes upstream file, so it needs to either be marked down with vorestation edit or sent upstream. Second, doubling the speed of dirtying up instead of halving the number thresholds related to it seems like needlessly bloating the number.

code/game/turfs/simulated.dm
```diff
@@ -69,7 +69,7 @@
 
 /turf/simulated/proc/update_dirt()
 	if(can_dirty)
-		dirt = min(dirt+1, 101)
+		dirt = min(dirt+2, 101)
```
Novacat, line 72: This needs to be either marked with //VOREStation Edit or sent upstream

Screemonster:
I have a feeling that this is gonna make things gross on no-janitor shifts.

lbnesquik:
Meh on busy work.

serenekerosene:
I can prompt the community at large and see what they think- I discussed it with a few people in favor before making this P.R. but I didnt hear much one way or the other the last time I posted about it in the Discord

KasparoVy:
Yo if push comes to shove hit me up and I'll make it rain soap from the mining vendor any time

Runa-Dacino:
We can always barf out a bunch of janibots to handle important areas if no janitor. Otherwise, this gives people a reason to call a janitor into their workplace and potentially create either some casual RP or some cliché scene involving... stuff

Heroman3003:
Cleanbots are far from most effective cleaning method I'd say, due to how 'efficent' their pathfinding is. And dirt overlay is one of those things, where it's visible enough to make things look uglier, but also irrelevant enough to be ignored by pretty much everyone. Medical is only place that can commonly have problems related to the dirt I can think of and afaik they just clean using their own cleaners all the time anyway. Which leaves just public areas.

Also, I feel it's worth pointing out, that after the thresholds were lowered... the effect of PR was changed as it wasn't quite lowered in every way. Now instead of dirtying up faster, tiles will have dirt overlay appear twice as early... but in its intial state it will also be half as visible compared to initial state on master right now (but still just as visible on higher dirt levels).

serenekerosene:
Aaaaa I did not mean to fuck with the transparency, should be fixed now

Heroman3003:
...to explain exactly how it works (without changes). The dirt overlay appears first when the 50 dirt threshold is reached. And it's transparency is set to be to whatever the dirt level is minus 50, then times 5, but never above maximum transparency (so at maximum of 101 the threshold of 255 is reached, but ithe limitation is there anyway for any special cases).

When you originally changed it you just made it so that tile reaches maximum dirtiness in half the steps. Then when you changed thresholds, you made the dirt overlay appear at 25 instead of 50, and then changed transparency formula to be -25 instead of -50, which would mean that maximum dirtiness visibility will be reached at 76, while dirt level can still go up (even though the maximum transparency will keep it alpha of dirt overlay from breaking), so it would get dirty earlier, and reach maximum earlier.

Current iteration will do... I have no idea what. Because between 25 and 50 dirt levels, the dirt overlay will appear and exist... but will have negative transparency? And after reaching 50 and above will behave normally.

I am not quite sure what exactly is desired intent, have dirt show up early or become visible early, so can't give straight up direction, but I hope it clears up things a little.

As for why I suggested reducing dirtiness thresholds, it's because IIRC, dirt overlay isn't only thing it's used for. I think infections and prommie footfeeding are also using that variable.

Dirt overlay is a mess in more ways than one.

serenekerosene:
Ok then how about I just 'dirt = min(dirt+2, 101)' like I did in the first place

serenekerosene:
alright i tested the first implementation on a private server and it didnt seem like anything about prommies changed at all- i anticipated maybe they'd get doubled nutrition if their stuff's tied to the step value but it seemed the same, i couldnt find any seriously life changing effects
had no idea how to test the virus thing but we only have those as adminbus anyways so i dont think it should have much bearing on the decision to merge this or not

Wickedtemp:
Tldr:
'Increasing scene potential' doesn't really hold up. If I was in the mood to take the janitor, I could just make something up. There doesn't have to be dirt to actually clean. 

I'd be okay with this if there were tools we had to keep the station clean with. Not clean dirt that's there, but something that can be applied to a tile to keep dirt from building up. I'd recommend sterilizine.

serenekerosene:
I actually like that sterilizine idea a lot! when more free time comes to me I might look into makin' that a thing

lm40:
i don't really see this as an improvement tbh, dirt is already kinda pointless busywork, although if it turns out everyone who mains janitor supports this then go for it i guess

Wickedtemp:
update:
so apparently dirt is persistent now? or something? because roundstart, everythings already dirty, which makes me feel like that this PR is, essentially, already in, at least in spirit.

Novacat:
With persistent dirt this is not really needed anymore.

PR #8444, opened 2020-07-22
Port TGUI Wire datums & Colorblind Wires

ShadowLarkens:
From ParadiseSS13/Paradise#13769

This updates the /datum/wire system to be basically 1:1 with /tg/, including converting it to TGUI--
except, it comes with Paradise's extensions for colorblindness affecting wire colors!
This means that the following wires:

Will appear as something like these for those with Vulp colorblindness: 

And similarly matching-themed wires for other colorblindness types.
(Note, `darkslategray` in those wire names was a typo, it'll appear as `dark grey` now. I didn't feel like rebooting just to fix the screenshots.)
Also, there's now alien multitool support. It does this.


ShadowLarkens:
By the way, if anyone has suggestions for more legible colors in this PR, please let me know-- only catch is that I'll also need a color to replace it with for protanopia, deuteranopia, and tritanopia colorblindness swaps. All possible colors are listed here https://www.w3schools.com/colors/colors_names.asp

ShadowLarkens:
Changed around some colors and decided to just snowflake in the alien multitool for now. 🤷‍♀️

KasparoVy:
That's badass, cheers

Heroman3003:
At long last, alien multitool actually matters in some way.

ShadowLarkens:
@Novacat Conflict fixed. As with all other TGUI changes here, this will be ported upstream once the primary PR has been reviewed and merged.

PR #8453, opened 2020-07-22
Tweaks some expedition loot

Novacat:
- Replaces ERT Headset with Captain headset
- Removes Emag
- Removes Infinite power cells (These should only ever be admin spawn)
- Removes easily accessible merc voidsuit


serenekerosene:
but why the emag? D:

PR #8470, opened 2020-07-25
Adds a weak random mech spawner to mech bay.

lbnesquik:
The goal of this is to increase mech usage. Literally it.
image

### DIFF
--- a/maps/tether/tether-03-surface3.dmm
+++ b/maps/tether/tether-03-surface3.dmm
@@ -22054,17 +22054,6 @@
 	},
 /turf/simulated/floor/bluegrid,
 /area/assembly/chargebay)
-"aLz" = (
-/obj/structure/cable/green{
-	icon_state = "0-8"
-	},
-/obj/machinery/power/apc{
-	dir = 4;
-	name = "east bump";
-	pixel_x = 28
-	},
-/turf/simulated/floor/bluegrid,
-/area/assembly/chargebay)
 "aLA" = (
 /obj/machinery/door/firedoor,
 /obj/machinery/door/airlock/glass_external/public,
@@ -33741,6 +33730,18 @@
 /obj/effect/floor_decal/industrial/outline/red,
 /turf/simulated/floor/tiled/monotile,
 /area/tether/surfacebase/shuttle_pad)
+"bOS" = (
+/obj/structure/cable/green{
+	icon_state = "0-8"
+	},
+/obj/machinery/power/apc{
+	dir = 4;
+	name = "east bump";
+	pixel_x = 28
+	},
+/obj/random/mech/weaker,
+/turf/simulated/floor/bluegrid,
+/area/assembly/chargebay)
 "bSe" = (
 /obj/structure/disposalpipe/segment{
 	dir = 4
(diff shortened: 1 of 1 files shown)


Heroman3003:
I'm unsure about possibility of Deathripley being there... Mostly because of name, description and the KILL CLAMP. That and there is big RNG factor with mining ripley, since diamond drill is a possible starting piece of equipment for it.

lbnesquik:
I don't think it is a huge issue in terms of balance, especially not the death ripley which is kind of a joke.

As a note, this is a first step PR, the ideal goal of this is to have things like the CM tank. Starts off ruined but still intact enough for you to repair it with some effort and perhaps a few replacement parts.

lm40:
do not support random durands spawning at roundstart in robotics, if you want robotics to make you a mech that is literally like 50% of their job and i have never met a roboticist who refuses to do so for someone with legitimate need

lbnesquik:
The actual intended goal is to have a more broken version of things, like the equivalent of finding a sports car in a mineshaft somewhere. It takes more, much more effort to repair and so on, but you don't have to pay as much to get it back up.

Cyantime:
If that's the goal, then this is the wrong spawner to be using. This one spawns normal full health mechs, including gygaxes and durands.

lbnesquik:
I can change that, but given the other mech change in the pipe line, i'll wait for it to be done. So we can have actually broken mechs to repair.

JoanRisu:
So I need more information on this because I don't really understand what is happening that defines this as "Weaker."

What is going into these mechs that makes these "weaker." I need technical information for "weaker."

Rykka-Stormheart:
I support mechs spawning nearly destroyed, imo. Would give robotics fun stuff to fix. :blep:

Heroman3003:
That would partially depend on what nearly destroyed entails too. If its just 10 hits with a welder and its in mint condition again, then there's only marginal difference between starting nearly-destroyed and fully functional.

lbnesquik:
It might be better to consider this as a draft. I have a clear idea of what i need to do, but that needs extra testing and work
Namely:

- Creating ruined versions of all mech internals along with sprites maybe
- Creating a random spawner that MAYBE (not every shift) spawns a ruined mech with completely fucked up internals, from armor to weaponry
- Adjusting the resource costs of replacement parts and so on as i think those aren't well balanced ATM.
- Dive into the mech maintenance code to improve the process of replacing & interacting with the mechs.

As i've been working lately, it is getting a bit hard to put aside to 4-7 hours needed for this, but this is absolutely still on my radar.

Very-Soft:
While I do like the sentiment of 'it would be cool if people used mechs more' 

I want to point out that, a big part of why people don't really use mechs is that they aren't really useful to most people on the station. They're almost always overkill or too clunky to really be all that worth seeking out. Now, I'm not entirely sure how to fix that without talking about overhauling mechs in general. 

I think having a random spawn mech that needs some good work to get functional is definitely cool (as long as they don't spawn in like, kitted out with weapons and such). But I am not sure that this will result in people actually wanting to -use- the mechs.

PR #8471, opened 2020-07-25
Medical TGUI

ShadowLarkens:
From ParadiseSS13/Paradise#13738: Medical TGUI

Almost every interface in the medical department now uses TGUI.
 - Chemical Dispenser
 - ChemMaster 3000
 - Body Scanner
 - Sleeper
 - Resleever
 - Medical Records
 - Cryotube
 - Surgical Monitor

Also converted these, even though we don't use them on Virgo, they're in use upstream.
 - DNA Modifier
 - Cloning Console
 - Also converted the Reagent Grinder to use radial menus instead of a UI.

Chem Dispenser
Med Records
Cryo Cells
Body Scanner
Chem Master
Sleeper
Resleeving
Surgical Monitor
Reagent Grinder

# Unused Here, but showcasing for the sake of Polaris
Cloning
DNA Modifier


KillianKirilenko:
Bodyscanner needs a second pass. You removed the ability for it to detect in-bloodstream chems and several other issues, and the reports it prints are *basically blank*, whereas the ones it used to print were complete snapshots of the patient's state at the time they were scanned.

PR #8474, opened 2020-07-25
Service Remap tweak

SweetBlueSylveon:
Fixes the roof for surface med.
Adds lights.
Changes the hallway disposals to a bin in the entertainment room.
Gives clowns entertainment access.
Fixes a pipe's visibility.
Adds shutter and tint control buttons where they were missing.
No the I'm not moving the goat he's my friend and he's staying in the freezer.


SweetBlueSylveon:
The observer spawn landmark is fucky wucky so we're waiting on a fix for that.

PR #8476, opened 2020-07-25
Service patch againer

SweetBlueSylveon:
Github is being real weird with this.
Fixes a number of last improvements. Adds a door betweent he bar and kitchen, more cameras, fixes some piping, puts more sinks in botany. Removes an incorrect evac sign.


SweetBlueSylveon:
unfinished do not merge

PR #8490, opened 2020-07-26
Kitchen patch (mostly)

SweetBlueSylveon:
A moderate overhall to the kitchen to make it feel less overly spacious, while still keeping the extra walking room.
Changes some shutter directions.
Puts punpun in the backroom.


SweetBlueSylveon:
Not ready do not merge.

SweetBlueSylveon:
If this passes all check it should be ready.

PR #8498, opened 2020-07-27
Port Blast Door Crushing!

Rykka-Stormheart:
Ports this from https://github.com/Aurorastation/Aurora.3/pull/4305

Adds blast door/shutter crushing. Does 40 damage for blast doors and 10 for shutters.


lbnesquik:
I like it! But maybe a tad less than 40 dmg.

Rykka-Stormheart:
> I like it! But maybe a tad less than 40 dmg.

It's a massive blast door, it's a little silly to have it do basically no damage. 40 is enough to break a bone if you're unlucky and will remind you not to stand in blast doors. They're **blast doors** after all. :p

KasparoVy:
Honestly that shit should cleave my stupid ass in half

Heroman3003:
Considering how damage is applied, 40 is pretty much guaranteed bone break. On weaker species, even 10 dealt in single attack may be enough for that.

lm40:
this is a good thing on action servers but adding it here is a ???? for me. what is it going to add besides 'woops I crushed someone in the blast doors and broke a bone'.

Heroman3003:
Tbh, I really don't see anything actually positive coming out of this PR other than "just make it realistic for sake of being realistic".

Wickedtemp:
So, shutters were edited to deal no damage. 

Is it just 'blast doors' specifically, with firelocks remaining unaffected?

How many sets of 'blast doors' do we even have? I recall Virology having a lockdown, and like.. the Armory, and I think the Bridge. Where else? Maybe somewhere in Science/Xenobio? If we only have blast-doors in those four spots, then I don't really see much of an issue with this because they're rarely if ever utilized to start with.

Rykka-Stormheart:
> So, shutters were edited to deal no damage.
> 
> Is it just 'blast doors' specifically, with firelocks remaining unaffected?
> 
> How many sets of 'blast doors' do we even have? I recall Virology having a lockdown, and like.. the Armory, and I think the Bridge. Where else? Maybe somewhere in Science/Xenobio? If we only have blast-doors in those four spots, then I don't really see much of an issue with this because they're rarely if ever utilized to start with.

Firelocks are not a child of blast doors, and blast doors also forcemove you out of them once they close.

Issue #8499, opened 2020-07-27
Maintenance Drones unable to view camera console.


Cyantime:
#### Brief description of the bug
Attempting to view an engineering camera console is denied while a maintenance drone.

#### What you expected to happen
Click camera console, view cameras to get information on engineering or locations of alerts.

#### What actually happened
Receive a message that says `You realise its kind of stupid to access a camera console when you have the entire camera network at your metaphorical fingertips`

#### Steps to reproduce
Spawn as maintenance drone, click camera console.

#### Code Revision
53e0b5d9f053dae5af8f8b8ef1589cce8a0da8b2 
 

#### Anything else you may wish to add:
This message seems intended for an AI, which does actually have camera access.

Heroman3003:
What about borgs? Is issue same with them too?

Rykka-Stormheart:
> What about borgs? Is issue same with them too?

Known issue, Shadow just fixed it tonight, iirc. I'll nudge her.

PR #8500, opened 2020-07-27
Fix for Camera interaction for Borgs/Drones.

Rykka-Stormheart:
To note - the original change was intentional change on Paradise's part.

"You might not have been able to actually view the cameras, but you should have been able to open the interface." - this restores the attack_hand(user) functionality for borgs/drones.


Rykka-Stormheart:
Also: Fixes https://github.com/VOREStation/VOREStation/issues/8499

PR #8503, opened 2020-07-27
Reverts the forced sharp scaling on map window.

Verkister:
A bandaid fix that'll hopefully be replaced with an actual preference toggle button rather than the upstream fetish of hardcoding and forcing client preferences without access buttons to them.
Basically fixes map window having turned to sharp scaling, meaning that using the "stretch to fit" mode will mangle the entire screen if someone needs a little bit of a tweak. Most screens aren't big enough to blur the map window noticeably and if you're on a 4k screen, set your icon mode to 96 or 128 for fuck's sake.

### DIFF
--- a/interface/skin.dmf
+++ b/interface/skin.dmf
@@ -1194,7 +1194,6 @@ window "mapwindow"
 		text-color = none
 		is-default = true
 		saved-params = "icon-size"
-		zoom-mode = "distort"
 		on-show = ".winset\"mainwindow.mainvsplit.left=mapwindow\""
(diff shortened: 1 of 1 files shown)


Verkister:
I'll send to polaris once I can be arsed to actually add that toggle button myself. I don't want to start more wars lmao

KillianKirilenko:
I made one downstream as a test ([YW901](https://github.com/Yawn-Wider/YWPolarisVore/pull/901)), you're welcome to yank it and throw it to Polaris if you like. I haven't heard of any complaints/issues yet.

PR #8519, opened 2020-07-29
Explorer/Pathfinder Belts

KillianKirilenko:
Another old idea, and something that might be more useful than guns; belts. I originally added something similar purely for the Pathfinder on CitRP, but I figured why not share the fun? There's never enough pockets, after all.

Explorers get a belt that holds five items; the pathfinder's \~✨*deluxe*✨\~ belt holds seven. Both use the same item list and max item size. The item list is fairly wide compared to 'department' belts like the toolbelt or EMT belts, but do mean dealing with the extra inventory layer. Currently they're in the explorer and pathfinder lockers, but I could see a case for removing standard explorer belts from the lockers and only having them in the exploration vendor?

If anyone has any items they think I've missed, or think shouldn't be on the list (keeping in mind the `w_class` limit means there's lots of things that won't fit), let me know.

Sprites from Bay, which is where I got the idea. They have some nice belt sprites.


code/game/objects/items/weapons/storage/belt_vr.dm
```diff
@@ -1,4 +1,58 @@
...
+		/obj/item/weapon/cell,
+		/obj/item/weapon/gun,
+		/obj/item/weapon/material,
+		/obj/item/weapon/melee,
+		/obj/item/weapon/shield,
+		/obj/item/ammo_casing,
+		/obj/item/ammo_magazine,
+		/obj/item/weapon/storage/box/beanbags,
+		/obj/item/weapon/storage/box/shotgunammo,
+		/obj/item/weapon/storage/box/shotgunshells,
```
Cyantime, line 33: These being allowed means that explorers will be able to empty them and refill them with any number of items that fit in a box, bypassing both the belt item restrictions and the belt item slot limit.

KillianKirilenko: Shoot. Good point. I'll swap them for the shotgun speedloaders instead.

lbnesquik:
PR is ready for merge it seems.

KasparoVy:
4 strength 4 stam explorer belt

PR #8530, opened 2020-07-30
Freezers are chilly!

Rykka-Stormheart:
Freezers will start at T0C - 5 (Which is effectively 268.15), and a new Freezer air alarm subtype has been added that bypasses the TLV hardcoded default to allow for larger cooling thresholds (and allow lower target temps).

TL;DR: Freezer tiles and the kitchen freezer will start cold and actually stay cold, prompting "you feel chilly" in the freezer. Other areas that use /freezer/ tiles may be affected by the initial temp, but this should quickly resolve itself thanks to heat balancing.

### DIFF
--- a/code/game/machinery/air_alarm.dm
+++ b/code/game/machinery/air_alarm.dm
@@ -788,3 +788,14 @@
 	..()
 	spawn(rand(0,15))
 		update_icon()
+
+// VOREStation Edit Start
+/obj/machinery/alarm/freezer
+	target_temperature = T0C - 13.15 // Chilly freezer room
+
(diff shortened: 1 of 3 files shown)


Novacat:
I PR'd this as a joke in the past, but the serious issue with it is that every time someone opens a door to the freezer all that cold air will instantly escape thanks to ZAS.

Rykka-Stormheart:
> I PR'd this as a joke in the past, but the serious issue with it is that every time someone opens a door to the freezer all that cold air will instantly escape thanks to ZAS.

The map has tinyfans in the freezer doorways. I already thought of that. <3

PR #8538, opened 2020-07-30
[MIRROR] Implements Status Indicators

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/7354
--------------------
This PR adds the ability to have certain status effects such as stuns or confusion have a visual component, by having the ability to have icons show up over mobs.

Here's a video of it in action: https://streamable.com/rdp6a9

Some notes:
This is not tied to the med/sechuds at all whatsoever. The video shows the medhud bar to show that the icons don't conflict with it.
A lot of effort was put in to make sure that the icons would always be over the mob, even for edge cases like 64x64 sprites or scaled up sprites from things like modifiers or just a `adjust_scale()` call. Thus this should work just fine for big xenos.
Currently this PR adds icons and code to have all three stun types, blindness, confusion, and sleeping. The system is nice and generic so it can be hooked into things like Modifiers later down the road.
This also adds a player pref to disable seeing status indicators, if you don't like them for whatever reason.
The icons are, atm 16x16. Someone offered to make me 8x8 sprites but I'm skeptical that they can be distinguishable between each other during a high stress event, but the system's easy to adjust for 8x8 sprites, only needing to tweak a few `#DEFINE`s.


KasparoVy:
Muh immershuns.. Dang. Makes sense but I'm glad there's a preference for this.

PR #8558, opened 2020-08-01
Guttersite under construction

SweetBlueSylveon:
We're turning the Guttersite to an offsite spawn job. In the meantime, the construction crew has to come in and work on the place.


Rykka-Stormheart:
Force-rebased Gutter onto Master in preparation for work start.

PR #8583, opened 2020-08-03
Nerfs harm/disarm steppies

Novacat:
- Makes bad steppies have a chance to miss on mobs not on help intent
- Removes halloss and replaces with weaken
- Steppies now only forcerests people on help intent

### DIFF
--- a/code/modules/vore/resizing/resize_vr.dm
+++ b/code/modules/vore/resizing/resize_vr.dm
@@ -243,10 +243,18 @@ var/const/RESIZE_A_SMALLTINY = (RESIZE_SMALL + RESIZE_TINY) / 2
 	if(a_intent == I_HELP) // Theoretically not possible, but just in case.
 		return FALSE
 
+	if(tmob.a_intent != I_HELP && prob(35))
+		to_chat(pred, "<span class='danger'>[prey] dodges out from under your foot!</span>")
+		to_chat(prey, "<span class='danger'>You narrowly avoid [pred]'s foot!</span>")
+		return FALSE
+
 	now_pushing = 0
 	forceMove(tmob.loc)
 	if(a_intent == I_GRAB || a_intent == I_DISARM)
-		tmob.Weaken(1)
+		if(tmob.a_intent = I_HELP)
+			tmob.resting = 1
+		else
+			tmob.Weaken(1)
 
 	var/size_damage_multiplier = size_multiplier - tmob.size_multiplier
 	// This technically means that I_GRAB will set this value to the same as I_HARM, but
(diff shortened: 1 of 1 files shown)


code/modules/vore/resizing/resize_vr.dm
```diff
@@ -246,7 +246,7 @@ var/const/RESIZE_A_SMALLTINY = (RESIZE_SMALL + RESIZE_TINY) / 2
 	now_pushing = 0
 	forceMove(tmob.loc)
 	if(a_intent == I_GRAB || a_intent == I_DISARM)
-		tmob.resting = 1
+		tmob.Weaken(1)
```
Rykka-Stormheart, line 249: Weaken(2) or Weaken(3) would be better, imo

Screemonster:
Honestly? Maybe shouldn't even be an auto-weaken if the victim also has a hostile (disarm, harm, grab) intent. Disarm-intent clicks don't automatically succeed either.

Rykka-Stormheart:
> something to the effect of the following added to the block in lines 236-245 should do the trick:
> 
> if(tmob.a_intent != I_HELP && prob(50))
> to_chat(pred, "[prey] dodges out from under your foot!")
> to_chat(prey, "You narrowly avoid [pred]'s foot!")
> return FALSE

Agreed. Macro stomps shouldn't be an instant win, though imo I'd reduce it from prob(50) to something like prob(35/45). Still not an instant win, but not a hard 50/50.

Screemonster:
> Agreed. Macro stomps shouldn't be an instant win, though imo I'd reduce it from prob(50) to something like prob(35/45). Still not an instant win, but not a hard 50/50.

honestly I just pulled "50" out of my ass because I couldn't be bothered looking up what the actual odds of a successful disarm are

Rykka-Stormheart:
> > Agreed. Macro stomps shouldn't be an instant win, though imo I'd reduce it from prob(50) to something like prob(35/45). Still not an instant win, but not a hard 50/50.
> 
> honestly I just pulled "50" out of my ass because I couldn't be bothered looking up what the actual odds of a successful disarm are. Note that in that code, the prob would be the odds of the micro _dodging_, not the macro _hitting_ so pick the number accordingly.

Ye, I still think it should be prob(35) or prob(45). Macros are large and getting into close combat can often be a death sentence because of how hard it is to click shit near you.

MrSign:
Yeah, if we could implement a 'dodge' mechanic or something similar to disarm where it has a chance of failing, that'd be great.

lbnesquik:
Muh coils stuff

Rykka-Stormheart:
> Yeah, if we could implement a 'dodge' mechanic or something similar to disarm where it has a chance of failing, that'd be great.

That's been implemented!

KasparoVy:
Man I must be unrobust. Not sure of how to dodge aside from moving my dude, flipping like a gymnast or using jump boots to dive into a wall

code/modules/vore/resizing/resize_vr.dm
```diff
@@ -243,10 +243,18 @@ var/const/RESIZE_A_SMALLTINY = (RESIZE_SMALL + RESIZE_TINY) / 2
...
+	if(tmob.a_intent != I_HELP && prob(35))
+		to_chat(pred, "<span class='danger'>[prey] dodges out from under your foot!</span>")
+		to_chat(prey, "<span class='danger'>You narrowly avoid [pred]'s foot!</span>")
+		return FALSE
+
 	now_pushing = 0
 	forceMove(tmob.loc)
 	if(a_intent == I_GRAB || a_intent == I_DISARM)
-		tmob.resting = 1
+		if(tmob.a_intent = I_HELP)
```
Meghan-Rossi, line 254: This is giving an "Assignment in condition" warning, and I think == would make more sense here.

lbnesquik:
REEEE

PR #8613, opened 2020-08-06
Removes Reagent Implants

Novacat:
As per a decision by staff, all whitelists for Reagent Implants (both milk and egg) have been rescinded, these are to be replaced by a generic 'Feeder' neutral trait that is now available as per this PR.

If you have any questions, concerns, or comments about this PR, it would be far better (for both yourself and my sanity) to direct those to the Head Staff.


lbnesquik:
This is for everyone?

Novacat:
Yes, everyone will have access to the feeder trait

JoanRisu:
The Feeder Trait would be available to everyone as a neutral trait. 

We've been reluctant to give out the reagent generator for far too long to the point of questioning why the reagent generator was still in code if we were no longer willing to give that specific thing out. After a meeting, we decided a good middle ground would be to make a neutral trait that is opposite the succubus drain trait.

lbnesquik:
The only thing it says is slime feed. You can RP with that? I haven't ever seen that used before.

JoanRisu:
It was in the code but apparently wasn't used anywhere else? I'm at least hoping it will help set the groundwork for a better feeder trait.

PR #8631, opened 2020-08-08
TGUI Vore Panel

ShadowLarkens:
Makes the Inside! panel TGUI. Mew~


KasparoVy:
Lmfao, that's awesome. Nicely done

Heroman3003:
I feel like wording on the vore prefs could use being more consistant and/or reordered a little. Like Digestion, Absorption and Healbelly prefs all relate to different belly modes, but are worded differently and are not in a row. And generally language could be more clear, like "mobs allowed to eat" in prefs, or "digest brute/burn".

Also that's more of a minor personal thing, but I feel like the order of the controls stuff has been a little messed up for a while now, and that sound-related settings got split up by others. So fleshy belly and Internal loop are not next to eject/swallow noise settings, while they probably should've been.

Wickedtemp:
Do we know for a fact, 100%, that updating to this won't delete any existing saved data for the current panel

KasparoVy:
@Wickedtemp With the massive detail some people put into their characters & vorebellies I highly recommend backing up everything to formatted txt files and placing 0 trust in the system

ShadowLarkens:
@Wickedtemp 100% yes, it touches absolutely nothing but the front end, it's not possible for this to erase your vore settings; though I'll say Kasparov gives good advice in general.

KasparoVy:
OH SHIT TOOLTIPS

PR #8632, opened 2020-08-08
Adds raptor pAI

Verkister:
-Adds raptor pAI chassis option.
-Also fixes the misnamed full belly sprites for raccoon chassis.


KasparoVy:
Those raptor sprites are fuckin awesome.

PR #8633, opened 2020-08-08
Slightly lower the volume of the 'command report' voice.

lbnesquik:
- Did you know we have not much of any way to tweak the volume in those announcements? Yeah. It is pretty much just 100%


Heroman3003:
I think lowering volume of sound playing procs would be a better solution than literally editing a file. Especially since its a Polaris file.

lbnesquik:
Not a polaris file. We have a locally different one. Also i am working on seeing if i can navigate the mess of code that is the announce system, but i am not sure i'll manage it, so there is this PR anyway.

Novacat:
Ye, I would rather you edit the proc that calls the announcement than edit the sound file itself.

lbnesquik:
Yeah so. It is proving a lot more complex than i hoped as there is like, 5 different procs to touch and i am not familiar with this? So i'll try to get to it but it would be better to just lower this thing volume.

PR #8636, opened 2020-08-09
Replace audio files with quieter versions of them!

Rykka-Stormheart:
This helps immensely with showers being loud from over 7-10 tiles away, just like the old fryer. If showers are too quiet now, please give feedback and up the looping sound from 15 to something like 30.


Heroman3003:
Aren't shower sounds upstream?

Rykka-Stormheart:
> Aren't shower sounds upstream?

Yep, already porting.

Heroman3003:
Also, general problem in these situations isn't that "sound is too loud" but "sound is heard from too far away" so reducing volume of the file will do very little to avert the issue. The sound-playing proc ideally should have a the range to which the sound is allowed to travel gimped severely instead.

Rykka-Stormheart:
https://github.com/PolarisSS13/Polaris/pull/7448 This is done

And @Heroman3003 frankly the easiest thing to do is add "ignore_walls = FALSE" to prevent showers from being heard outside the bathroom.

Rykka-Stormheart:
But yes, you are correct in that reducing the range would help. For now, this works fine enough - if range is still far too long, then we can reduce it further.

PR #8643, opened 2020-08-09
Resleeving fixes

### LINKED ISSUE #8630: Resleeving console unable to access disk records
The resleeving control console (which is supposed to load up active body record from a disk when slapped with one) completely ignores the body record disk interactions, making it completely impossible to create stock or otherwise disk stored sleeves.

For the console to read the record from the disk.

#### What actually happened
The console UI generation overrides the active record vars automatically when interacted with, forcibly and exclusively loading the available records from the transcore backup list with no way to load up the disk record, and even if it did list the record, it would just qdel it anyway for lacking a ckey.

#### Steps to reproduce
- (Step 1) Attempt to use the disk related features on resleeving console.
- (Step 2) Fail miserably.

#### Code Revision
- 6cef1a3bb8f7de4f4920894aaf2d6b7b66281dfe
 

ShadowLarkens:
Fixes #8630 
Along with a handful of other fixes, such as:
 - You can now view mind records even if you can't immediately sleeve them
 - You don't have to load someone's body into the sleever before you can select it
 - Really lorge OOC Notes no longer break the console (they're in a scrollable box now)


KasparoVy:
OOC notes so big it literally breaks the cycle of life death and un-death

PR #8645, opened 2020-08-09
Adds new horse mob.

Verkister:
Bigger, faster, stronger, and hungrier than the old little horse. Also obviously mountable. Sprites by Kligor.


KasparoVy:
Horse looks legit, nice sprites

PR #8654, opened 2020-08-11
Adds a second confirmation to the reset character slot option

Cyantime:
I've seen accidental slot resets happen like twice in the last week, if they continue beyond this it's a user issue.


Cyantime:
I figured that it doesn't matter to polaris, but sure.

PR #8655, opened 2020-08-11
Less wind per wind

serenekerosene:
Comments out wind slowdown+speedup code, easy as. Shouldn't(?) remove fluff messages for wind.


serenekerosene:
WHOOPS

PR #8656, opened 2020-08-11
Less wind per wind Correct Edition

serenekerosene:
Comments out slowdown from wind as requested by mining players. Shouldn't(?) affect flavor text of existing wind


Rykka-Stormheart:
Why? 

This feels pointless, wind is a mechanic that exists to give planets more differences than "tile with different atmos than station".

Heroman3003:
Wind as a mechanic has exactly as much positives as it does negatives. But despite the fact that its 100% rng, and that its positive and negative effects are exactly equally as strong, and it even has like 25% chance to not even have any effect at all. For those few times it can be bad, the whole thing is bad.

One of reasons thrown around against I saw was literally "I can't speedrun mining as efficently because of it" and I didn't see much argument beyond that other than "Its bad when it slows you down as you walk to Underdark". But it has exactly as much chance to speed you up. Or do nothing. So I genuinely do not see an issue there. Sometimes its bad, sometimes its good, its RNG with in-universe flavor. And speedrunning... Well, it already depends on RNG in the first place.

Cyantime:
If it's *really* that much of a problem, why not give miners Special Boots in their point machine that neutralize wind effects when worn. No speedups or speeddowns, so they can work to obtain a sidegrade.

Rykka-Stormheart:
> If it's _really_ that much of a problem, why not give miners Special Boots in their point machine that neutralize wind effects when worn. No speedups or speeddowns, so they can work to obtain a sidegrade.

...or they can just get over it. They don't need to be able to zoom gogogogo speedrun mining.

KasparoVy:
Yeah I don't get it man, wind makes sense. Literally all I do is mine and it feels like we're making mountains out of mole hills here - can walk with the wind, or go to the Underdark or space where you'll not find a breeze at all.
There ARE boots that negate the effects of wind already, they're called jump boots and they're fun as fuck!

Cyantime:
> 
> 
> > If it's _really_ that much of a problem, why not give miners Special Boots in their point machine that neutralize wind effects when worn. No speedups or speeddowns, so they can work to obtain a sidegrade.
> 
> ...or they can just get over it. They don't need to be able to zoom gogogogo speedrun mining.

Which is why I suggested putting it in their point vendor and not the lockers, so they can do some mining and then spend their points on something that isn't even an upgrade, it's a sidegrade because wind is equally as helpful as it is harmful.

TeleTubby2:
As is, wind is a perfect example of annoying faux depth. it adds nothing to planets at all (seriously. I don't see how "u get a chat message once in a blue moon and also hahah slowed down" makes a planet feel different. I'm about 90% sure if wind direction was global, less than 1% of the playerbase would notice). it only serves to be a random speed change. that's it. It's inherently never going to contribute to RP, as it's basically an invisible value (at the very least, something like weather can be commented on, and actually provides a different "atmosphere") Even if it DID somehow contribute to RP, this PR... doesn't touch it. at all. it just removes the annoying part.

Wind as a system is a complete fakeout. it doesn't even blow pieces of paper, or other light objects around, but somehow manages to slow someone down, or speed them up by a noticable amount. and even if it *did*, there's absolutely zero reason for it to also be annoying in the process. This PR feels pointless because wind is pointless, and inherently always will be until it does something other than the one dimensional fakeout of being a random speed value modifier hnnnng

code/modules/mob/living/carbon/human/human_movement.dm
```diff
@@ -162,6 +162,8 @@
 			. += turf_move_cost
 
 	// Wind makes it easier or harder to move, depending on if you're with or against the wind.
+	// BUT FUCK DAT HAHAHAHAHA
```
Novacat, line 165: //VOREStation Edit Start

code/modules/mob/living/carbon/human/human_movement.dm
```diff
@@ -174,6 +176,8 @@
 				else if(direct & reverse_dir[WH.wind_dir])
 					. += WH.wind_speed
 
+*/
+// GET OWNED IDIOT WIND -- VOREstation Edit.
```
Novacat, line 180: VOREStation Edit End

lm40:
i don't mine but the handful of times i have to go outside to recover a body or explore the wilderness moving at the speed of evolution makes me not want to bother. randomly moving at sonic speed is kinda fun but not really worth it.

AegisOA:
Guess I'll chip in my two cents. I can see the point of both arguments here, but I'd rather have the wind speed buff/debuff nerfed at worst and removed at best. I can't deny that it's fun to occasionally have the wind RNG behave and have you zoom towards your destination, nor can I deny that I SUPPOSE it keeps people from "speedrunning" mining, but, as someone who'll often spend a minimum of a couple hours mining whenever I do mine (just to make sure nobody needs anything else), it's an absolute SLOG to work with if the RNG hates your guts. The most egregious example of this by far is the trip to and from the Underdark; our suits already slow us down as it is, and the trip to the Underdark stairs from the mining outpost is painfully slow/boring if the wind is blowing to the southwest; ditto for returning from the Underdark if the wind is blowing to the northeast. Sure, it can help speed you up in one direction or another, but considering how much walking we do outside anyways, it's a coin flip on whether or not you get there x% faster or x% slower. I don't think that's a worthwhile trade-off in the long run, as someone who's had shit wind RNG to some extent on many a trip.

KasparoVy:
Changed my mind, no longer care which way this goes
The wind don't blow in space anyhow nyeh-heh-heh. Too many good arguments for both sides but I realized my enjoyment of the wind was mainly muh immershunz and wind RNG when I'm a Teshari, that plus the fact that wind doesn't affect anything else doesn't make it worth that much to me anymore.

PR #8657, opened 2020-08-11
Minor buffs for flying

Cyantime:
People flying will now no longer slip on wet floors, leave bloody footprints, or create dirt.
The immunity to slipping while flying is the only major benefit here, but like, you can still just shove a flier into a wall and ground them easily.

### DIFF
--- a/code/game/turfs/simulated.dm
+++ b/code/game/turfs/simulated.dm
@@ -83,7 +83,7 @@
 
 	if (istype(A,/mob/living))
 		var/mob/living/M = A
-		if(M.lying)
+		if(M.lying || M.flying) //VOREStation Edit
 			return ..()
 
 		if(M.dirties_floor())


code/game/turfs/simulated.dm
```diff
@@ -83,7 +83,7 @@
 
 	if (istype(A,/mob/living))
 		var/mob/living/M = A
-		if(M.lying)
+		if(M.lying || M.flying)
```
Novacat, line 86: //VOREStation Edit

Cyantime: Done, that's what I get for doing that at 3 am.

KasparoVy:
Thank the Gods.

Cyantime:
If I'm going extra far with that, I'd just work it into the default polaris `hovering` var and fix things up on that end. Less vorestation edits, more compatibility with future changes, etc etc.

KasparoVy:
Can only go so far in a 1pt trait too! Had no idea about the hovering var, I was just mad kludging it hahaha.

Very happy to see flight getting some love.

PR #8707, opened 2020-08-16
Restore TLV clamping (+ sound update for canister dispenser)

ShadowLarkens:
By request from Bluehaus. Air alarms clamp the TLV values to make sure minimum is always below maximum, and that danger-max is never below warning-max.
https://i.tigercat2000.net/2020/08/qeDUe6X18d.webm


ShadowLarkens:
Also changed the dispense noise. With the changes that make the thing stuff it right into your hands, the drop noise was no longer triggered by dropping it on the floor, so it was just silent instead of making a clangy noise. Updated so that it makes the clangy noise instead of the vending noise.

PR #8708, opened 2020-08-16
Persistence Patch

Rykka-Stormheart:
- Fixes Paper and font tags stacking across rounds, as shown in: 

- Fixes Graffiti being persistent for over 50 rounds (10+ days)
- Sets Trash to be persistent for 4 rounds instead of 5
- Fixes Persistence being always-on because of if("persistence_enabled") config.persistence_enabled = 1. As explained: 
> it's on by default, which means config.txt will be read and if it's not there it stays at 1. If it is there,it gets set to 1. There is no way to disable it

- Adds config options to enable/disable **Persistence for Maploaded objects**, as well as a verb to toggle such.
- Adds Persistence to config.

Currently IGNORE_MAPLOAD for Persistance is **Disabled.**
Maploaded objects/dirt/etc will be saved by persistence, preserving current behavior.

You will need to update config.txt with the following lines:
\## Uncomment this to DISABLE persistence 
\#PERSISTENCE_DISABLED 
 
\## Uncomment this to DISABLE maploaded trash/paper/etc from being saved by the persistence system. 
\#PERSISTENCE_IGNORE_MAPLOAD

### DIFF
--- a/code/controllers/configuration.dm
+++ b/code/controllers/configuration.dm
@@ -236,7 +236,8 @@ var/list/gamemode_cache = list()
 	var/static/dooc_allowed = 1
 	var/static/dsay_allowed = 1
 
-	var/persistence_enabled = 1
+	var/persistence_disabled = FALSE
+	var/persistence_ignore_mapload = FALSE
 
 	var/allow_byond_links = 0
 	var/allow_discord_links = 0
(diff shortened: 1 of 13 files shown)


Meghan-Rossi:
Wait, why shouldn't graffiti last for 10 days if not cleaned up?  That seems good to me.  1 day seems a bit short for something deliberately created for the purpose of being seen.

Rykka-Stormheart:
> Wait, why shouldn't graffiti last for 10 days if not cleaned up? That seems good to me. 1 day seems a bit short for something deliberately created for the purpose of being seen.

Admin request, but it's up to headmin review.

Wickedtemp:
Because it's possible to carve graffiti with any 'sharp' object, and lately people have been doing things like carving the bee movie script and shitposts into tiles. The thing is, this can't be cleaned normally. You have to use a welder, or pry up the tile and replace it. Since it lasts for over a week, there's been a build-up of it in some areas and overall it seems like a dumb thing to have last that long.

VictorZisthus:
> 
> 
> Because it's possible to carve graffiti with any 'sharp' object, and lately people have been doing things like carving the bee movie script and shitposts into tiles. The thing is, this can't be cleaned normally. You have to use a welder, or pry up the tile and replace it. Since it lasts for over a week, there's been a build-up of it in some areas and overall it seems like a dumb thing to have last that long.

Ive been going around cleaning every bit I see every single shift as fix-it-borgo

PR #8721, opened 2020-08-17
Readd "All" Button to vore panel

ShadowLarkens:


Heroman3003:
Does it appear even when there's no contents there, only when there's one thing or when there's two or more?

ShadowLarkens:
It always appears

Heroman3003:
Does it break anything if you try using it with 1 or 0 contents?

ShadowLarkens:
Doesn't break anything, doesn't make a sound as release_all_contents cancels if there's nothing inside and transfer contents only runs for things in the belly and therefore won't be called if there's nothing there

PR #8725, opened 2020-08-17
Flintlock's jacket

dwinters99:
Accepted fluff request here: https://forum.vore-station.net/viewtopic.php?f=26&t=1801


dwinters99:
> It looks like the onmob sprite and icon sprite are switched? That should probably be fixed.

It... doesn't look like it?
image
image

Although looking at this makes me realize I forgot the in-hand sprites, so fixing that among other things.

FlintlockDaFox:
knock knock. is it going to be re-reviewed?

Novacat:
Icon state and item state are still incorrectly switched it looks like. For some reason the icon state is named flintlock_mob and the item state is named flintlock when it should be the other way around.

PR #8731, opened 2020-08-17
TGUI Security

ShadowLarkens:
Surprisingly few UIs left in security to convert. Do let me know if I missed any. Other than the implantpad. That's a special kind of hell that I'll have to address later.

This converts:
 - Cell Timers
 - Prisoner Management Console
 - Security Records Console
 - DNA Scanner
to TGUI.

Also makes the camera console have pan buttons instead of "previous | next" buttons


Rykka-Stormheart:
Buttons on the UI to add time in increments of 1, 5, and 10 minutes as well would be nice.

We have Short/Medium/Long - but those... don't display the times on them, unless I'm mistaken.

If the short/med/long displayed how much time they ADD, that'd be great. <3

MrSign:
Yeah, look at elaborating on what short, medium and long adds on the timer.

And from what I see from the screenshots. When it comes to the timer, can you actually type in the amount of time?

Rykka-Stormheart:
> Yeah, look at elaborating on what short, medium and long adds on the timer.
> 
> And from what I see from the screenshots. When it comes to the timer, can you actually type in the amount of time?

Yep! You can type in the time in minutes OR seconds. <3

ShadowLarkens:
Cell timers now look like this 

And panning is once again possible in the camera console (I tried my best to get hotkeys working and it didn't work out)

ShadowLarkens:
random fix, IDs do this now if there's no picture

PR #8734, opened 2020-08-18
Minor Map + Talon Edits

VictorZisthus:
TALON: Added more pressure to the airlock Air canisters so they don't lock up after only a dozen uses or so.

TALON: Switched the grinder and chem dispenser around for easier use with macros.

STATION: Due to constant destruction of critical emergency equipment at roundstart by certian individuals, tesla engine has had two super capacitors added next to the tesla circutboards. We also added more air to the airlock canisters, so the airlocks do not lock up so quickly. Have a secure day.


Heroman3003:
I dunno how to feel about super capacitors. Why not normal ones? That's literally ability to upgrade the tesla coils for free roundstart there because... reasons? And even then making regular capacitors by engineering is dirt-easy task with just building a parts lathe. Or even just grabbing extra capacitors from the equipment storage (it has 2 IIRC), not like they're used for anything else.

VictorZisthus:
> 
> 
> I dunno how to feel about super capacitors. Why not normal ones? That's literally ability to upgrade the tesla coils for free roundstart there because... reasons? And even then making regular capacitors by engineering is dirt-easy task with just building a parts lathe. Or even just grabbing extra capacitors from the equipment storage (it has 2 IIRC), not like they're used for anything else.

Oh, do regular caps work with tesla? Good to know.

VictorZisthus:
> 
> 
> > I dunno how to feel about super capacitors. Why not normal ones? That's literally ability to upgrade the tesla coils for free roundstart there because... reasons? And even then making regular capacitors by engineering is dirt-easy task with just building a parts lathe. Or even just grabbing extra capacitors from the equipment storage (it has 2 IIRC), not like they're used for anything else.
> 
> Oh, do regular caps work with tesla? Good to know.

If this is the case I can just remove the capacitors entirely, though that will still cause the gutting of the Thermal Ragulator at roundstart X3

Heroman3003:
Any capacitors work for tesla coils. But the higher the level, the more power they will be able to extract. By default coils use normal capacitors. The 'tier' of capacitors is literally multiplier of how much power the coil will produce. Second tier doubles it, third triples, etc.

VictorZisthus:
> 
> 
> Any capacitors work for tesla coils. But the higher the level, the more power they will be able to extract. By default coils use normal capacitors. The 'tier' of capacitors is literally multiplier of how much power the coil will produce. Second tier doubles it, third triples, etc.

Your call, I'll remove 'em if ya want and make the choice to gut the Thermal Regulator or make a Lathe.

Heroman3003:
Maybe map in a pair of capacitors in the board sotrage if there aren't any there already. Would make sense, fit in and be even easier/less wasteful than either gutting or needing to build a lathe.

VictorZisthus:
> 
> 
> Maybe map in a pair of capacitors in the board sotrage if there aren't any there already. Would make sense, fit in and be even easier/less wasteful than either gutting or needing to build a lathe.

Downgrading to regular caps it is.

Heroman3003:
When I say board storage, I mean the storage in the mezzazine, one where parts lathe board and all other spare machine/console boards are btw.

VictorZisthus:
> 
> 
> When I say board storage, I mean the storage in the mezzazine, one where parts lathe board and all other spare machine/console boards are btw.

Ahhh. Gotcha.

vorestation.dme
```diff
@@ -3704,5 +3704,7 @@
 #include "maps\submaps\surface_submaps\wilderness\wilderness.dm"
 #include "maps\submaps\surface_submaps\wilderness\wilderness_areas.dm"
 #include "maps\tether\tether.dm"
+#include "maps\tether\tether-03-surface3.dmm"
+#include "maps\tether\tether-08-mining.dmm"
```
Heroman3003, line 3708: Seems like you included the tether maps in include file, which is why you're getting those Travis errors for missing areas. Those are included through defines in tether.dm, in part to avoid issues with unit tests like that.

Cyantime:
Increasing the pressure on the airlock canisters makes it *much slower* to cycle out, since the vent now needs to pump into a higher pressure pipe network. The normal solution to those running out is adding a regulator from the air supply pipes to top them off if they drop below a given pressure.

PR #8740, opened 2020-08-19
Minor Talon Edits

VictorZisthus:
Made the armory a bit more friendly with actual lockers instead of just racks, and clothing enough for four crew instead of just one. Also includes an extra locker with black civillian garb and bags... In the armory, because I have no idea where else to put it on the tiny Talon.

Added a few more lights in maints in a few areas.

Unlocked personal lockers roundstart so that they could, you know, actually be used by the Talon crew.

### DIFF
--- a/maps/tether/submaps/offmap/talon1.dmm
+++ b/maps/tether/submaps/offmap/talon1.dmm
@@ -158,6 +158,9 @@
 /obj/machinery/atmospherics/unary/vent_pump/on{
 	dir = 1
 	},
+/obj/structure/closet/wardrobe/black{
+	starts_with = list(/obj/item/clothing/under/color/black = 4, /obj/item/clothing/shoes/black = 4, /obj/item/clothing/head/soft/black = 4, /obj/item/clothing/mask/bandana = 4, /obj/item/weapon/storage/backpack/messenger/black = 4, /obj/item/weapon/storage/backpack/dufflebag = 4)
+	},
 /turf/simulated/floor/tiled/eris/white/orangecorner,
 /area/talon/deckone/armory)
 "as" = (
--- a/maps/tether/submaps/offmap/talon1.dmm
+++ b/maps/tether/submaps/offmap/talon1.dmm
@@ -216,9 +216,76 @@
 /obj/machinery/chemical_dispenser/full,
 /turf/simulated/floor/tiled/eris/white/bluecorner,
 /area/talon/deckone/medical)
+"ax" = (
+/obj/machinery/camera/network/talon,
+/obj/structure/closet/wardrobe/black{
+	starts_with = list(/obj/item/clothing/head/helmet/combat = 4, /obj/item/clothing/mask/gas/commando = 4, /obj/item/clothing/mask/balaclava = 4, /obj/item/clothing/under/syndicate/combat = 4, /obj/item/clothing/accessory/storage/black_vest = 4, /obj/item/clothing/accessory/storage/black_drop_pouches = 4, /obj/item/clothing/gloves/combat = 4, /obj/item/clothing/suit/armor/combat = 4, /obj/item/clothing/shoes/boots/combat = 4)
+	},
+/turf/simulated/floor/tiled/eris/white/danger,
+/area/talon/deckone/armory)
+"ay" = (
+/obj/structure/catwalk,
+/obj/structure/cable/green{
+	d1 = 1;
+	d2 = 2;
+	icon_state = "1-2"
+	},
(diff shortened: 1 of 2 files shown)


maps/tether/submaps/offmap/talon1.dmm
```diff
@@ -216,9 +219,76 @@
 /obj/machinery/chemical_dispenser/full,
 /turf/simulated/floor/tiled/eris/white/bluecorner,
 /area/talon/deckone/medical)
+"ax" = (
+/obj/machinery/camera/network/talon,
+/obj/structure/closet/wardrobe/black{
+	starts_with = list(/obj/item/clothing/head/helmet/combat = 4, /obj/item/clothing/mask/gas/commando = 4, /obj/item/clothing/mask/balaclava = 4, /obj/item/clothing/under/syndicate/combat = 4, /obj/item/clothing/accessory/storage/black_vest = 4, /obj/item/clothing/accessory/storage/black_drop_pouches = 4, /obj/item/clothing/gloves/combat = 4, /obj/item/clothing/suit/armor/combat = 4, /obj/item/clothing/shoes/boots/combat = 4)
```
Cyantime, line 225: I don't think this counts as a "minor edit", this is giving the talon four full sets of combat armor, while currently they get a single helmet, armor suit, and jumpsuit. Those gloves are heavily armored and insulated, while the boots are noslips. Why does the talon need to be armored up like they're nuke ops?

Cyantime:
They should be perfectly able to use the personal closets. You swipe them with an ID to unlock them and set their owner to you, having them start unlocked may break their functions.

maps/tether/submaps/offmap/talon1.dmm
```diff
@@ -216,9 +219,76 @@
...
+	},
+/obj/machinery/light/small,
+/turf/simulated/floor/plating/eris/under,
+/area/talon/maintenance/deckone_starboard)
 "aC" = (
 /turf/simulated/wall/rshull,
 /area/talon/maintenance/deckone_port)
+"aD" = (
+/obj/structure/closet/wardrobe/black{
+	starts_with = list(/obj/item/clothing/head/helmet/space/syndicate/black = 4, /obj/item/clothing/suit/space/syndicate/black = 4, /obj/item/clothing/shoes/magboots = 4)
```
Cyantime, line 288: Here too, four sets of the armored syndicate spacesuits instead of the current one and two normal unarmored talon crew suits.

VictorZisthus: > 
> 
> I don't think this counts as a "minor edit", this is giving the talon four full sets of combat armor, while currently they get a single helmet, armor suit, and jumpsuit. Those gloves are heavily armored and insulated, while the boots are noslips. Why does the talon need to be armored up like they're nuke ops?

Wasn't sure how armored up they should be. I take it that one set is enough then?

Had no idea the gloves and boots were that special. Yikes! I'll downgrade those to somehting a bit more reasonable.

The locker thing is good to know too. I'll undo that change.

> 
> 
> Here too, four sets of the armored syndicate spacesuits instead of the current one and two normal unarmored talon crew suits, plus a specialized suit that's in each crew locker.

Whoops, forgot about the special suits. Retracting that too!

VictorZisthus:
> 
> 
> I don't think this counts as a "minor edit", this is giving the talon four full sets of combat armor, while currently they get a single helmet, armor suit, and jumpsuit. Those gloves are heavily armored and insulated, while the boots are noslips. Why does the talon need to be armored up like they're nuke ops?

Wasn't sure how armored up they should be. I take it that one set is enough then?

Had no idea the gloves and boots were that special. Yikes! I'll downgrade those to somehting a bit more reasonable.

The locker thing is good to know too. I'll undo that change.

> 
> 
> Here too, four sets of the armored syndicate spacesuits instead of the current one and two normal unarmored talon crew suits, plus a specialized suit that's in each crew locker.

Whoops, forgot about the special suits. Retracting that too!

Novacat:
Generally rule of thumb is that unless you are intentionally re balancing things, you should generally stick with the same amount of gear between versions

maps/tether/submaps/offmap/talon1.dmm
VictorZisthus: > 
> 
> Generally rule of thumb is that unless you are intentionally re balancing things, you should generally stick with the same amount of gear between versions

Gotcha!

VictorZisthus:
Closing this PR to start from scratch, look forward to another!

PR #8747, opened 2020-08-20
Buffs/fixes darksight traits.

Verkister:
The major trait was not quite what it advertised itself as (it counts the tile you're standing on, not the outer radius so it was never full view) and the minor was just utter waste of points.


Novacat:
Personally, I liked that major darksight is not as good as the vanilla species with perfect darksight (Tajara, basically), though I probably should just bite the bullet and completely rebalance custom traits.

KasparoVy:
HELL yeah.

Issue #8746, opened 2020-08-20
Gas Canisters can't seem to squeeze out that last 0.1 KPA of gas to completely empty them, resulting in the inability to relabel them for other uses.


MrSign:
#### Brief description of the bug
Gas Canisters are always stuck on 0.1 kPa, so that you aren't able to use the relabel feature or reuse them for other gases appropriately if needed.

#### What you expected to happen
Pumps being able to remove all the contents of a canister, allowing for reuse. IE; waste container.

#### What actually happened
Even if left for ages, the 0.1 kPa remains.

#### Steps to reproduce

1. Get a canister.
2. Attach it to a pump to drain.
3. Use gas scanner on an engineer/atmospheric PDA to measure kPa
4. Wait until it is 0.1 kPa.
5. Wait a few minutes and recheck.
6. See that it is still 0.1 kPa.

#### Code Revision
ea017e7c86ec677c6d5fe6d5e187ac2d7d1b34a7 
 

#### Anything else you may wish to add:
![kpaissue](https://user-images.githubusercontent.com/33539568/90725643-d9ff0f80-e303-11ea-9f18-3e529bcd6a1e.JPG)

Heroman3003:
There weren't any changes to gas functionality lately afaik. Could it have anything to do with the tgui?

MrSign:
That was my thought! Cause I was able to do it before the new UI came in. Granted, I haven't done atmospheric for ages, so it could've been an issue even before then...

Rykka-Stormheart:
I'll check downstream and see.

Rykka-Stormheart:
Downstream (Who does NOT have TGUI) have confirmed that this is an existing bug.

Admiralkosmov:
this was a thing as far back as I can remember, honestly. The closer you get to 0, the slower it empties and then that last bit seems to just go on forever...

Holdelta:
This has been happening for as long as I can remember playing engineering(around a year).

With things such as trace phoron fires and other gas concentration in the air that will forever stay around at around 0.04 kPa, I would assume this is as designed due to ZAS and atmospherics in general.  The best way to exemplify this is volatile fuel, you can permanently put a room into an atmospheric warning(not alarm) if you put a little volatile fuel in the air.

I would assume that canisters hold this same exact rule.

Heroman3003:
[...]

PR #8749, opened 2020-08-20
Fixes not being able to relabel canisters

Meghan-Rossi:
Fixes https://github.com/VOREStation/VOREStation/issues/8746


Meghan-Rossi:
This is part of the tgui interface, which polaris doesn't have yet.

KasparoVy:
Send ALL of TGUI to Polaris, muahahaha! Yeesh..

PR #8750, opened 2020-08-20
Talon Bugfixes and QOL

VictorZisthus:
Made various airlock pumps High Volume to reduce wait times.

Added a few more lights in maints in a few areas.

Added an unused hardpoint to the bow starboard to match the bow port sensor suite. [for ship symmetry]

Added underplating beneath catwalks outside. Catwalks being the only thing keeping the wings on? YIKES! [Also made slipping and gravity-related issues...] Added handrails at various external points in case someone forgets their magboots and does not want to faceplant.

Added an ATM outside the Talon at the bar.

Added nitrile gloves, a bioproduct dispenser, roller beds, a body bag box, and a chem locker to medical. Should they have a smartfridge?...

Addes more magboots to suit storage to prevent any further infighting among talon crew. Also stocked more 02 canisters and PSCU's, just in case; They are cheaper to produce than a new crew. Also added a closet with black civilian garb, enough for 4 peeps.

Like above, but with C02 jetpacks. Added C02 Canisters to storage and just outside the Talon boat for refueling them.

### DIFF
--- a/maps/tether/submaps/offmap/talon1.dmm
+++ b/maps/tether/submaps/offmap/talon1.dmm
@@ -56,9 +56,7 @@
 	dir = 4;
 	icon_state = "map_connector-aux"
 	},
-/obj/machinery/portable_atmospherics/canister/air/airlock{
-	start_pressure = 4559.63
-	},
+/obj/machinery/portable_atmospherics/canister/air/airlock,
 /turf/simulated/floor/tiled/eris/steel/gray_perforated,
 /area/talon/deckone/port_eng)
 "ai" = (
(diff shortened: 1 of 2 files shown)


VictorZisthus:
I have no idea why those previous two commits made it in, but those were already merged soooo..... *shrug*

VictorZisthus:
> 
> 
> I have no idea why those previous two commits made it in, but those were already deleted soooo..... _shrug_

Oh, the closet for black clothind did stay though.

VictorZisthus:
> 
> 
> Remove the underplating under the catwalks, that was an intentional design decision that was done to keep the shield generator from trying to cover the wings.

Ah, copy that! Doing that now.

PR #8807, opened 2020-08-22
Refurbished "Bearcat" Ship

KillianKirilenko:
This old thing's been sat in the `/maps/overmap/` folder forever, so I figured I'd dust it off and do a little refurbishing work. Consider this a 'first pass' or something.

Could be used for events or added to the overmap as a random non-guaranteed spawn location. Not sure how to do that exactly.

Notes;

- [x] Hook up waypoints/docking collar
- [ ] Probably need to throw in some more odds and ends: clothing, personal effects, vintage suits, etc.
- [ ] Also probably need to add more wear and tear/damage.
- [x] Fix up some pipework
- [x] Redo door access levels
- [ ] Make it actually flyable (helm consoles are failing entirely after initial shortjump)


KillianKirilenko:
Closing this for now, but will keep working on it to see if I can beat it into a functional shape.

PR #8808, opened 2020-08-22
FBP Munchies

KillianKirilenko:
Frees FBPs from the tyranny of chargers by allowing them to gain energy from processing nutriment and nutriment ~~accessories~~ subtypes. They gain only ~~two thirds~~ one quarter the normal amount, and no other benefits (for organics, nutriment heals a small amount of brute damage, and restores lost blood at around half the rate of iron).

~~Uses the _vr file to override a couple of things rather than interfering with upstream files.~~

~~Posting as a draft to get feedback/opinions. Food (a ha. ha.) for thought;~~
- ~~Originally I was going to make this just a specialized food item that was only good for FBPs, and provided minimal nourishment to organics on top of tasting bland/unpleasant (the whole idea basically being inspired by this scene https://www.youtube.com/watch?v=93O2aPUjxDw&t=548 ) but that meant doing sprites and I was feeling lazy at the time.~~
- ~~I figure this can probably also be done as a custom species trait with a little wrangling.~~
- ~~Alternately, the efficiency can simply be made lower, though I'd recommend no lower than 1/4th or so.~~

Made into a custom trait as requested. Includes a check so non-synths can't take it!


lm40:
can i eat electricity for nutrition then

KillianKirilenko:
> can i eat electricity for nutrition then

If you want to eat a solar grub, more power to you.

KasparoVy:
Enhanced Biofuel Processor :o

Would be a dope trait name.

Very-Soft:
I don't like it but I also don't really have any super good reason to say no to it, since synths can already get energy from doing vore mechanics. 

Though, I suppose the one argument against this is that organic people don't have an effectively infinite source of consistently refilling their hunger like synths do with the rechargers. (But there's also enough vending machines and things that it's probably fine)

I deffo would prefer that synths got less energy from normal food though, for sure, so that's good.

nerdass:
don't fpb's not process reagents inside them?

KillianKirilenko:
> I don't like it but I also don't really have any super good reason to say no to it, since synths can already get energy from doing vore mechanics.
> 
> Though, I suppose the one argument against this is that organic people don't have an effectively infinite source of consistently refilling their hunger like synths do with the rechargers. (But there's also enough vending machines and things that it's probably fine)
> 
> I deffo would prefer that synths got less energy from normal food though, for sure, so that's good.

At the end of the day it all comes down to power. Every regular food source is functionally infinite so long as there's enough power, end of story. Chargers, vendors, kitchen, cargo, hydroponics: all of them can provide energy/nutrition for the duration of a round so long as there's power, and it takes a *very* busy station to tax *all* of those resources to their limits.

An organic walking to a vending machine is not much different to a synthetic walking to a charger, really.

> don't fpb's not process reagents inside them?

[Polaris#7445](https://github.com/PolarisSS13/Polaris/pull/7445) made it possible, so long as the reagents have the correct vars set.

Rest assured I test all my nonsense thoroughly before I make PRs. If this didn't work, this PR wouldn't exist.

KillianKirilenko:
Shifting this to draft state temporarily since Polaris has a pending fix for the reagent cycler and I want to be sure there won't be any issues with it.

KillianKirilenko:
Updated and tested for the cycler changes. I've also capped off the charge gain from food so you can't munch a ton of stuff all at once and then never have to recharge for the rest of the shift.

KillianKirilenko:
Goddamn, it actually passed unit tests, finally. Only took four tries too.

Novacat (changes requested):
I admit I am surprised that overriding a proc with a _vr file actually works, but I would still prefer for the base proc to be changed instead of a copy being added to _vr,

The var additions/changes to lines 463-465 can stay. affect_blood and affect_ingest should be moved back to the original file. Partially so I can see what exactly changed, partially so that if Polaris changes anything with food upstream, that would give me a heads up.

Also, staff had a discussion on this.
- Food should give 1/4 the nutriment to synths that it gives organics.
- If it would be possible, can these mechanics be exported to a neutral trait?

KillianKirilenko:
I'll give it a try, pending the resolution of some tech issues. Might take a bit.

KillianKirilenko:
Drafted until I can get things back underway properly, whenever that might be.

KillianKirilenko:
Did a little refactoring and cleaning up to make it work as a neutral trait rather than always a thing, but so far it seems to be working nicely in local tests. Just gotta iron out the merge/conflict kinks.

Novacat (changes requested):
Sorry I have been taking so long to get to this, but first it needs conflict resolution!

KillianKirilenko:
aaaaaaaaaa

gimme a minute

KillianKirilenko:
Right, I think that did it.

Novacat (approved):

KasparoVy:
Yay! Congrats man @KillianKirilenko

PR #8812, opened 2020-08-22
Gas Overlay Update, OR What if we make gas look like a gas instead of massive dust particles?

KeRSedChaplain:
newphoron
Just a simple sprite edit is all this is. Probably going to shift up the layer of the gas overlay itself at some point so it looks more like gas and less like spicy carpeting.


lm40:
this is cool

Rykka-Stormheart:
This is now superseded by incoming mirror https://github.com/VOREStation/VOREStation/pull/8878

Novacat:
Redundant now

PR #8819, opened 2020-08-23
Changes Gutter to Gutterband, removes whitelist.

dwinters99:
Submitted on behalf of @SweetBlueSylveon

### DIFF
--- a/code/__defines/species_languages.dm
+++ b/code/__defines/species_languages.dm
@@ -49,7 +49,9 @@
 #define LANGUAGE_SIIK "Siik"
 #define LANGUAGE_SKRELLIAN "Common Skrellian"
 #define LANGUAGE_TRADEBAND "Tradeband"
-#define LANGUAGE_GUTTER "Gutter"
+//VOREStation edit 08/23/20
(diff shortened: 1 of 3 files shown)


dwinters99:
I didn't actually know alientwhitelist was distributed. TIL. I also don't know if LANGUAGE_GUTTER is referenced _anywhere_ but I'll change it nonetheless. Something something legacy.

Heroman3003:
It IS referenced, even accoring to most basic search of git. Probably some more deprecated/unused features, but generally it's best to stick with existing defines because there will inevitably be *that one moment* that makes a runtime.

Wickedtemp:
Let's be honest, if there was a language nearly exclusively used by criminals, one of the aforementioned criminals would happily agree to teach and translate for law agencies and really anyone who'd be interested in paying for the information. It'd get spread around. A small group of codewords, easier to hide. A language spanning multiple systems where all it takes is a single guy looking for a sweet paycheck or immunity deal for past crimes committed in order to have all of that info transcribed and distributed. 

's why I always thought a whitelist for gutter was dumb. This is neat.

PR #8822, opened 2020-08-23
Fixes Mercs threaten timers, and threaten sounds

Rykka-Stormheart:
The default "threaten" timer was 3 seconds, meaning that as soon as world.time advanced beyond that, the merc would shoot you. If there's TIDI/lag or other issues, you'd immediately end up getting shot.

30 seconds is more reasonable, allows you time to back away from the merc without being shot at, and it keeps the mercs from being cheesed easily.

This also fixes the merc aggression sounds. Actual voicelines incoming once I get help from a contributor!


lbnesquik:
Do they shoot back if you start firing during this delay?

Novacat:
Yes.

PR #8823, opened 2020-08-24
Couple bugfixes

cadyn:
Adds NIF Responsive Filter protection from Portable Flashers, and fixes problems with food reagents being transferred into pred's bloodstream after prey is absorbed, causing toxic damage.


TheFurryFeline:
I thought the reagents thing was intentional.

code/game/machinery/flasher.dm
```diff
@@ -68,6 +68,9 @@
 		var/flash_time = strength
 		if(istype(O, /mob/living/carbon/human))
 			var/mob/living/carbon/human/H = O
+			if(H.nif && H.nif.flag_check(NIF_V_FLASHPROT,NIF_FLAGS_VISION))
+				H.nif.notify("High intensity light detected, and blocked!",TRUE)
+				continue
```
Novacat, line 73: These changes need to be marked with //VOREStation Edit Start and //VOREStation Edit End, preferably above and below this block.

Heroman3003:
It made sense in context that ALL reagents from the prey were transfered into bloodstream. I think they should actually ALL transfer to ingested instead.

cadyn:
> It made sense in context that ALL reagents from the prey were transfered into bloodstream. I think they should actually ALL transfer to ingested instead.

Fair enough

PR #8828, opened 2020-08-25
Hypothermia

cadyn:
Added hypothermia with a variety of symptoms that develop slowly over time. Makes dying from the cold a bit more interesting, and also adds negative effects for some temperatures.

Yes these are getting sent upstream but I wrote them on VSCode so just take it.


cadyn:
Shhh, ignore the commits, there were some afterthoughts but it's all tested and glorious

code/modules/mob/living/carbon/human/life.dm
```diff
@@ -794,6 +794,21 @@
 
 	var/body_temperature_difference = species.body_temperature - bodytemperature
 
+	var/hypotemp = min(species.cold_level_1+17.5,T0C+10) //People going into hypothermia in 20C might be a bit overboard.
```
ShadowLarkens, line 797: why do you arbitrarily make hypothermia trigger *above* cold_level_1? x3

code/modules/mob/living/carbon/human/life.dm
```diff
@@ -820,6 +835,35 @@
 		bodytemperature += recovery_amt
 
 	//This proc returns a number made up of the flags for body parts which you are protected on. (such as HEAD, UPPER_TORSO, LOWER_TORSO, etc. See setup.dm for the full list)
+/mob/living/carbon/human/proc/add_hypothermia(var/amount)
+	if(hypothermia+amount<=0)
```
ShadowLarkens, line 839: for each of the operators below, +, <=, >, +=, etc, add spaces around them pls
so it looks like `if(hypothermia + amount <= 0)`

code/modules/mob/living/carbon/human/life.dm
cadyn: Because cold levels are absolutely borked at the moment. Humans can stand outside in -30C weather in a blizzard, naked, and chill.

Heroman3003: Thats issue with the comfort levels and its functionality. Starting hypothermia before the damage/warnings are supposed to start coming in sounds like just bad design.

code/modules/mob/living/carbon/human/life.dm
cadyn: aaaa

code/modules/mob/living/carbon/human/life.dm
cadyn: > Thats issue with the comfort levels and its functionality. Starting hypothermia before the damage/warnings are supposed to start coming in sounds like just bad design.

That's why I also added an alert for it. Also nobody's going to want to redo the temperature system so in my mind this is the real only way of making things sane without bursting someone's bubble. Plus it makes the whole thing a lot more interesting in general

Heroman3003:
The way it is implemented seems like just making the low temperature even more annoying and painful to deal with than it already was imo. Also stuff like just entering the freezer starting hypothermia ticks, but I can't say whether it was because freezer coldness is dumb or because the way this basically ignored the values set on species for their cold damage thresholds. Mostly both.

cadyn:
> The way it is implemented seems like just making the low temperature even more annoying and painful to deal with than it already was imo. Also stuff like just entering the freezer starting hypothermia ticks, but I can't say whether it was because freezer coldness is dumb or because the way this basically ignored the values set on species for their cold damage thresholds. Mostly both.

It doesn't ignore the values set on species for cold damage thresholds, the temperature at which hypothermia starts is based on the values set for cold damage thresholds of the user's species. Also just tested it and the freezer doesn't cause hypothermia ticks unless you have hot_adapt or otherwise a high ceiling for cold damage. 

Second of all... since when has low temperature ever been genuinely annoying or painful to deal with? As far as I can remember, ridiculously cold places like the Russian outpost and the snow map have always had 0 consequences for even being naked and drenched in water. The only time I can imagine cold temperatures genuinely being frustrating is if your species doesn't handle cold temperatures well, in which case that's a trade-off that you decided to make and makes the game more interesting.

Heroman3003:
It ignores the values by taking them... and starting hypothermia at arbitrarily chosen value higher than that. Instead of maybe defining new species variable related to that. Or just using them directly.

KasparoVy:
Thumbs up to the concept, I hope it gets based on species thresholds at some point
No bathing suits on Hoth

cadyn:
> It ignores the values by taking them... and starting hypothermia at arbitrarily chosen value higher than that. Instead of maybe defining new species variable related to that. Or just using them directly.

I think you're misunderstanding what the point of this PR is. The point is that the cold levels are flabbergastingly low for most species at the moment, and so instead of raising them for every species, I'm adding a sort of stage that comes before direct damage ticks called hypothermia where you start accrue various negative symptoms until you, yes, starting taking damage after a very extended amount of time in said environment.

code/modules/mob/living/carbon/human/species/species.dm
```diff
@@ -127,6 +127,7 @@
 	var/body_temperature = 310.15							// Species will try to stabilize at this temperature. (also affects temperature processing)
 
 	// Cold
+	var/hypothermia_difference = 17.5						// Hypothermia effects below cold_level_1 + this.
```
Heroman3003, line 130: do you mean, below cold_level_1 in commine here?

cadyn: Uhh, confused by what you're asking. The comment is supposed to point out that hypothermia starts at a body temperature of cold_level_1 + this [variable], as in cold_level_1+hypothermia_difference

cadyn:
Closing this and moving operations over to polaris, since you guys have a bot for downstreaming stuff anyways!

PR #8842, opened 2020-08-26
Modularizes Engine Submap Loading so it can be selected per-map!

Rykka-Stormheart:
Self-Explanatory - just go to engine_submaps_vr/mapname and check the _engine_submaps.dm to choose the engine presets you want to load.

Figured we might as well modularize this as it'll help with upcoming/new map projects in the future. :blep:

@Novacat pls speedmerge thx


maps/submaps/engine_submaps_vr/tether/_engine_submaps.dm
```diff
@@ -0,0 +1,34 @@
+// This causes engine maps to get 'checked' and compiled, when undergoing a unit test.
+// This is so Travis can validate PoIs, and ensure future changes don't break PoIs, as PoIs are loaded at runtime and the compiler can't catch errors.
+// When adding a new engine, please add it to this list.
+#if MAP_TEST
+#include "engine_rust.dmm"
+#include "engine_singulo.dmm"
+#include "engine_sme.dmm"
```
Meghan-Rossi, line 7: Not sure why tesla wasn't on this list.  Maybe it should be added?

Rykka-Stormheart: No idea, that's how it was beforehand.

PR #8845, opened 2020-08-27
Makes Liquid Protean and Repair Nanites affect synths

Heroman3003:
Now they too can enjoy NIF repairs using chugging a glass of ~~spiders~~ nanites

### DIFF
--- a/code/modules/reagents/Chemistry-Reagents_vr.dm
+++ b/code/modules/reagents/Chemistry-Reagents_vr.dm
@@ -44,6 +44,7 @@
 	reagent_state = LIQUID
 	color = "#333333"
 	scannable = 1
+	affects_robots = TRUE
 
 /datum/reagent/nif_repair_nanites/affect_blood(var/mob/living/carbon/M, var/alien, var/removed)
 	if(ishuman(M))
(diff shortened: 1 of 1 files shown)


KasparoVy:
A brimming glass! Cheers.

KillianKirilenko:
Be sure to test this one thoroughly if you haven't already; I tried to do something vaguely similar the other day (grinding nanopaste into a tricord-like reagent) but the code wasn't having any of it and I couldn't for the life of me figure out why.

Heroman3003:
This just enables the variable for the NIF repair nanites and liquid protean reagents to function inside synths that was added recently. Nothing more.

KillianKirilenko:
That doesn't mean it actually *works*, which is the important part of *testing*.

I took the liberty of testing it myself on my freshly-updated master (based on virgo's current master) and, as I suspected, it does nothing.

The problem is that reagents aren't transferred to the synth's blood, so `affect_blood` does nothing. I suspect the original FBP internals weren't tested thoroughly enough, or something else is breaking.

Heroman3003:
Well, whenever it does get fixed, those reagents will work properly too. For now injections at least should work. Hopefully. And even if they don't not that big a loss. The intent of PR was mostly to throw virgo-unique reagents that should have that flag but don't under it. If flag is broken as a whole, thats entirely different issue.

KillianKirilenko:
Injection won't work either as there's a blanket check on that, which is something else that needs fixing.

The whole FBP thing seems kinda janky and hacky taking a couple more looks at it, so there are probably more issues that need resolving down the line.

PR #8848, opened 2020-08-27
Remove gravity from debris field and rogue mining zones

### LINKED ISSUE #8936: The Debris Field (And possibly away missions in space in general) have gravity... in space
In the away missions, such as the debris field, there is something very wrong with the physics there
There is gravity. IN SPACE. IT'S HORRIBLE.
Which is to say, you still "float." But you don't drift at all. You can still move one tile or so, but with no drifting it's like you have an invisible jetpack with its stabilization flipped on but no propulsion otherwise. You sit there in space and cannot move unless you are lucky as me as to have a stack of Thalers to get around. I had to throw one thaler at a time to move one tile each to get back.
One Thaler. One Tile. 

For the empty void of space to have no gravity.

#### What actually happened
There was gravity in a place where they shouldn't be.

#### Steps to reproduce
- (Step 1) Go to an away mission in space, such as the debris field.
- (Step 2) Realize there's gravity in a place where there is no floor.
- (Step 3) Realize there's no sane way to get around in space without a jetpack.
- (Step 4) Get stuck in space.
- (Step 5) Cry.

Cyantime:
This fixes the weirdness where people wouldn't drift in space, but get stuck a turf away from a wall.
Also totally removes the debrisfield explored/unexplored areas, replaced with normal space area and a shuttle buffer area around the shuttle landmarks.
also makes the asteroids generated by rogue mining not look broken up

This also ended up bloating into like doubling+ the number of debris field PoIs

The gutter site also has some similar weirdness, but there's a PR up that's already touching that and it isn't urgent enough for conflicts to be worth it

Fixes #8936 except for the gutter site as mentioned above


Heroman3003:
Explored/Unexplored areas have a VERY serious reasons to be used in debrisfield. If you replace unexplored with space, debrisfield PoIs will start spawning anywhere where there's space. Outside station, outside guttersite, outside fuel depots...

If you replace explored areas with space, PoIs will start getting mashed on top of each other which will look about as well as you might expect.

There needs to be better solution than just replace everything with general space area, as this will straight up break those areas. Also, I completely don't get why the foodstand needs an RTG or a power source. The microwave there, from the beginning, was meant to be purely decorative on account of all the food in the stand being cooked already. And the entire PoI is meant to be a little joke on "what the hell is just a floating hot dog stand doing in the middle of debris field", not actual functional space food vendor.

Cyantime:
I have no idea what you're talking about heroman. The submap spawning proc is given a zlevel to operate in, it doesn't just spawn things randomly across the entire map. Debris field PoIs only spawn in the debris field, same as they did before.

The PoIs that have things that aren't just "some asteroids" or "a bit of lattice and a wall" have their own submap areas that won't get spawned onto, and the ones that *are* can now be mixed and combined at the whims of the spawner into different shapes, this seems like an improvement.

I didn't know about the previous history or intent of the food stand, I just saw something that might actually benefit from the stock rtg and its tiny power output and threw it on, if you want it gone I can do that.

The debris field completely empty and space, there's no reason not to use the proper space area for it.

Heroman3003:
Unless someone added dedicated areas to each PoI there and I missed it, when it was first being populated all pois had same area.

For why 'generic' PoIs shouldn't mix and match answer is really simple. They are only there to fill the space between other PoIs as much as possible. If they start mixing, spawning on top of each other, etc, they'll get way worse at doing the job of "filling empty space to make it actually look like debris field".

And yes, I do believe Food Stand gains no benefits from being powered other than somehow 'adding sense' to a PoI that intentionally is supposed to make no sense.

Cyantime:
Removed RTG from the food stand, added more generic and non-generic pois, increased point budget by to 300, from 200. to increase things and reduce open space.

Cyantime:
Leave this for a bit, want to get another batch of PoIs in and make them not cluster quite so tight

Cyantime:
Added another half dozen asteroid+debris+etc junk, another semi-decent PoI, and added some alt-space area in the middle of all the PoIs to keep them from *entirely* overlapping, only on the edges

KasparoVy:
Dope. I always explained it as 'we don't drift for the same reason the asteroids don't - this joint is stabilized' or some shit with a tractor beam ahaha. But this'll bring us back to the way it was when it worked as intended out in the asteroid field which is good

Cyantime:
I've got no idea what the heck happened to the mapdiff there, but the travis failure is unrelated, some sort of xenoarch thing in the snowfield.

Novacat (dismissed):
[2020-09-01T01:48:00] Runtime in unsorted.dm,1331: Found multiple lattices at 'The lattice (/obj/structure/lattice) @ (space) (110,123,23) (/turf/space)'

KasparoVy:
The fucking rogue lattice

Cyantime:
Occasionally lattices on the edge of submaps will overlap, this is completely harmless and just deletes the extra one. There really wasn't a need for it to be an outright runtime that kills travis, and it's been moved to a debug log message so that extra lattices accidentally mapped into non-randomgen areas can still be caught.

Added another PoI

Heroman3003:
Rotation is somewhat broken afaik.

Also lattices shouldn't be overlapping if areas are configured properly to not let submaps themselves overlap each other. Which is kinda the point of submaps.

Cyantime:
The point of submaps is to allow random premade chunks of map to be placed around an existing area. Being able to overlap is a perfectly fine use of them, and results in more experienced variety then a bunch of obviously discrete things cleanly stamped next to each other.

Heroman3003:
I still think that the random asteroid PoIs overlapping would look ugly and not in a good-intentional ugly way. Not mentioning how pois that aren't intended to overlap being slapped on top of pre-present asteroid pois would ruin their appearance. This went way beyond just 'fixing gravity' PR imo. The change in way Debrisfield PoIs spawn is entirely unrelated and doesn't need to be done to fix the issue.

Cyantime:
debris
Here's a screenshot of a chunk of the debris field that just generated, you can see that the actual PoIs are mostly left alone, while the miscellaneous rocks and chunks are mixed around in the intermediate space. There's more asteroid than metal debris just because of the difference in numbers of each type of PoI, which could be adjusted by adding more generic debris submaps.
I've already added a small bit of buffer area in the middle of each filler PoI to keep things from stacking repeatedly in the same place, they'll only overlap along the outer few tiles, depending on submap size.

Can you easily identify the individual asteroid maps in the screenshot?

KasparoVy:
Screencap looks legit as fuck. Nicely done

Meghan-Rossi:
This looks nice but it has some conflicts at the moment.

PR #8884, opened 2020-08-28
Fix oversensitive instability test

Meghan-Rossi:
I was testing some bomb stuff on a copy of master and I noticed the bomb tester's tank instability warning introduced in #8814 gives some false positives. This fixes that.  This version tested and working.  Pull request recreated due to double travis bug.


Cyantime:
I would like to know what conditions you've discovered these false positives though, tank integrity should only decrease if the mixture inside isn't stable and will eventually break the tank. If integrity is below that and the gas is fine, it recovers back to max.

PR #8885, opened 2020-08-28
Space level unsealing

Cyantime:
The space away areas were considered sealed zlevels which was causing runtimes when beams were fired off of them and probably other weirdness


KasparoVy:
Oh interesting, I wonder how thisll affect belt mining. Good fix!

Cyantime:
The belt mining levels weren't changed, and they've got teleporters that loop you around before you actually reach the zlevel edge so it doesn't matter either way

PR #8890, opened 2020-08-28
Increases number of things that survive overmap ztransit

Cyantime:
Keeps the following from being deleted when exiting a space zlevel with overmap enabled:
Active UAVs
Supermatter
Singularities and Teslas
Vehicles carrying things that also survive.
Anything with something important buckled.

There are probably many things I missed, but you get the idea

### DIFF
--- a/code/modules/overmap/spacetravel.dm
+++ b/code/modules/overmap/spacetravel.dm
@@ -49,13 +49,10 @@ proc/get_deepspace(x,y)
 			return FALSE
 	if(has_buckled_mobs())
 		for(var/mob/M in buckled_mobs)
-			if(!M.lost_in_space)
+			if(!M.lost_in_space())
 				return FALSE
 
 	return TRUE
-
-/obj/item/weapon/disk/nuclear/lost_in_space()
-	return FALSE
 
 /obj/item/device/uav/lost_in_space()
 	if(state == 1)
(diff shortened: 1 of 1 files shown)


Heroman3003:
I think SM and Nuke Disk should stay deletable. Nuke disk is not meant to be able to leave station levels in normal circumstance, and if it does its supposed to delete and respawn somewhere. As for Supermatter... well, nuking random spot doesn't sound too nice for 'safe' ejection procedures.

Cyantime:
wait whoops I thought I was doing this on polaris whoops

Right, I forgot how exactly that worked with the disk, it's been so long.
But the supermatter was specifically changed to return in https://github.com/PolarisSS13/Polaris/pull/7059 when the delamination effects were hugely nerfed, this is maintaining that.

Heroman3003:
Yes, but just because explosion is way smaller doesn't mean its not very stupid that ejecting it may just lead to it destroying other end of the station. We have specifically locally made it so that its deleted 100% of the time for a reason.

Cyantime:
That change was done over three years ago, when the supermatter explosion was massive and deanchored everything in a very large radius. The current explosion is much smaller in comparison, and if launched it can potentially spend a bit of time looping around an empty space sector. I feel that this is a good change.

PR #8658, opened 2020-08-11
Massive Ambience and Sound Overhaul + Addition

Rykka-Stormheart:
Includes changes from PR's:
https://github.com/Aurorastation/Aurora.3/pull/8861
https://github.com/tgstation/tgstation/pull/46677
https://github.com/tgstation/tgstation/pull/46387
https://github.com/CHOMPStation2/CHOMPStation2/pull/463
https://github.com/CHOMPStation2/CHOMPStation2/pull/108
https://github.com/tgstation/tgstation/pull/45462

Other changes include:
- New Security Alert for Red + Delta
- New Ambience per-area
- New SM sounds
- New Destructive Analyzer Sounds
- New Scanner/Cloner/Resleever Sounds
- Knuckle-cracking *crack emote
- Hyena *cackle emote (Virgo only)
- New Sounds for Oven/Grill/Candy/Cereal/Mixer Machines
- New Sound for Pre-Game Lobby (ping)
- Adds Preference for Pickup Sounds (Same as Drop Sounds)

PR #8891, opened 2020-08-28
Remove different sound for doors that have same mechanical base.

Heroman3003:
Partial VOREStation-only revert of https://github.com/VOREStation/VOREStation/pull/8658.

 Different paint job or different access restrictions don't make a door sound differently. This makes all doors with same mechanical basis sound the same, approprietly, not based on "what mapped in color they use". So all standard doors now again have same base sounds for opening and closing, as they all share base appearance.

Doesn't affect doors with actually unique appearances, such as pure glass, vault, voidcraft, hatches, highsecurity, doubles, etc. Would also have ignored doors made out of unique materials but those didn't have any sound given to them in the first place, so.

### DIFF
--- a/code/game/machinery/doors/airlock.dm
+++ b/code/game/machinery/doors/airlock.dm
@@ -39,9 +39,9 @@
 	var/secured_wires = 0
 	var/datum/wires/airlock/wires = null
 
-	var/open_sound_powered = 'sound/machines/door/hall1o.ogg'
+	var/open_sound_powered = 'sound/machines/door/hall1o.ogg' //VOREStation Edit
 	var/open_sound_unpowered = 'sound/machines/airlockforced.ogg'
-	var/close_sound_powered = 'sound/machines/door/hall1c.ogg'
+	var/close_sound_powered = 'sound/machines/door/hall1c.ogg' //VOREStation Edit
 	var/denied_sound = 'sound/machines/deniedbeep.ogg'
 	var/bolt_up_sound = 'sound/machines/boltsup.ogg'
 	var/bolt_down_sound = 'sound/machines/boltsdown.ogg'
--- a/code/game/machinery/doors/airlock.dm
+++ b/code/game/machinery/doors/airlock.dm
@@ -39,9 +39,9 @@
 	var/secured_wires = 0
 	var/datum/wires/airlock/wires = null
 
-	var/open_sound_powered = 'sound/machines/door/airlock.ogg'
+	var/open_sound_powered = 'sound/machines/airlock.ogg'
 	var/open_sound_unpowered = 'sound/machines/airlockforced.ogg'
-	var/close_sound_powered = 'sound/machines/door/airlockclose.ogg'
+	var/close_sound_powered = 'sound/machines/airlockclose.ogg'
 	var/denied_sound = 'sound/machines/deniedbeep.ogg'
 	var/bolt_up_sound = 'sound/machines/boltsup.ogg'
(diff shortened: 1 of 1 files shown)


Rykka-Stormheart:
Am against this getting in.

Rykka-Stormheart:
You also blanket reverted the base airlock sound changes, too.

Doubly against this.

code/game/machinery/doors/airlock.dm
```diff
@@ -39,9 +39,9 @@
 	var/secured_wires = 0
 	var/datum/wires/airlock/wires = null
 
-	var/open_sound_powered = 'sound/machines/door/covert1o.ogg'
```
Rykka-Stormheart, line 42: Why are these reverted?

code/game/machinery/doors/airlock.dm
```diff
@@ -39,9 +39,9 @@
 	var/secured_wires = 0
 	var/datum/wires/airlock/wires = null
 
-	var/open_sound_powered = 'sound/machines/door/covert1o.ogg'
+	var/open_sound_powered = 'sound/machines/door/airlock.ogg'
 	var/open_sound_unpowered = 'sound/machines/airlockforced.ogg'
-	var/close_sound_powered = 'sound/machines/door/covert1c.ogg'
```
Rykka-Stormheart, line 44: Ditto.

Rykka-Stormheart:
I still say this should be left unmerged until release is on live for a few days to gather feedback. Reverting this before the changes are even in is hugely kneejerk, imo.

Heroman3003:
The reason I am doing this PR is because of how previously this same change was turned down to be redone in the different way, and only reason it got merged this time is because maintainer merging it didn't notice the change due to misleading description of the PR. Decision that this should be done this way has been done months ago.

Rykka-Stormheart:
> The reason I am doing this PR is because of how previously this same change was turned down to be redone in the different way, and only reason it got merged this time is because maintainer merging it didn't notice the change due to misleading description of the PR. Decision that this should be done this way has been done months ago.

That's a rather snide attempt at saying "you lied". I did not lie, the PR's I pulled changes from were linked upfront in the PR.

code/game/machinery/doors/airlock.dm
```diff
@@ -39,9 +39,9 @@
 	var/secured_wires = 0
 	var/datum/wires/airlock/wires = null
 
-	var/open_sound_powered = 'sound/machines/door/covert1o.ogg'
+	var/open_sound_powered = 'sound/machines/airlock.ogg'
 	var/open_sound_unpowered = 'sound/machines/airlockforced.ogg'
-	var/close_sound_powered = 'sound/machines/door/covert1c.ogg'
+	var/close_sound_powered = 'sound/machines/airlockclose.ogg'
```
Novacat, line 44: I would like this to be reverted back.

Heroman3003:
Links in and of themselves tell you nothing, and 99% of the time are only posted as additions to list of changes. Except list of changes in said PR did not mention changes from the links themselves. I didn't say it was intentionally misleading.

As for the revert of the base door sound, part of the request.

code/game/machinery/doors/airlock.dm
Rykka-Stormheart: I'd suggest using hall1c and hall1o for close and open sounds respectively.

They fit well for the doors.

Rykka-Stormheart:
I'd still argue that this wait a few days for user feedback and 'min feedback before we revert it fully.

Is it okay to ask for staff discussion on this?

PR #8892, opened 2020-08-29
Enables the construction of battery rack circuits from APC circuits

Cyantime:
For all your field powering needs, the battery rack is now (mostly) freed from science's clutches. It still needs the three capacitors and a matter bin, but those can be scavenged or built from a partslathe.


Cyantime:
Right, forgot about that

Novacat (approved):

PR #8904, opened 2020-08-29
Makes rabbits not be vulpines

Heroman3003:
Changes tt_desc to something something rabbit in latin.


Rykka-Stormheart:
Thanks! I'd completely forgotten to do that.

PR #8905, opened 2020-08-29
Fix latin name for buns

Screemonster:
fixes copypasta in bun details, buns are not fennecs


Rykka-Stormheart:
Thancc. Hero beat ya to the punch tho lol <3

KillianKirilenko:
A hare too slow there, methinks.

PR #8912, opened 2020-08-30
Fix view spinning reversing movement controls

ShadowLarkens:
Not that spinning the view is any less janky or broken, but at least now it doesn't fuck your controls up.

### DIFF
--- a/code/modules/mob/mob_movement.dm
+++ b/code/modules/mob/mob_movement.dm
@@ -476,19 +476,26 @@
 /mob/proc/update_gravity()
 	return
 
+#define DO_MOVE(this_dir) var/final_dir = turn(this_dir, -dir2angle(dir)); Move(get_step(mob, final_dir), final_dir);
+
 /client/verb/moveup()
 	set name = ".moveup"
 	set instant = 1
-	Move(get_step(mob, NORTH), NORTH)
+	DO_MOVE(NORTH)
+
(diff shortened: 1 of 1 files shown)


Heroman3003:
Spin View is still kinda broken and abandoned. Everything lit looks wrong, everything multitile looks wrong, turfs dont rotate, some objects dont rotate, etc. Wasn't there plan at some point to just disable the commands entirely?

ShadowLarkens:
Not sure, at least a few years ago Bay was going in really hard on fixing all of the bugs with it, and that's when the verbs were introduced. Seems like it's fallen out of attention since then, probably since it was and still is pretty useless for anything other than relearning the station 4 times for each direction because it's so horribly disorienting.

Cyantime:
This is a direct nerf to concussion grenades, how could you

ShadowLarkens:
not really, concussion grenades also Confuse(), which this doesn't affect

KasparoVy:
Confusion/drunkenness rotating screen randomly with animate() would be wicked.

Having 4 totally different orientations for the station is fun, feels like different maps. Really helps take your nav skills to the next level

lm40:
man i would love to be rid of spinning view, it looks like garbage although now i guess at least the controls won't be janky af

Arokha:
Was created before current CI system. Kicking so current CI system runs against it.

PR #8913, opened 2020-08-30
Changes the dates of bank account creation to be more reasonable

Heroman3003:
Ideally, the entire thing should be reworked to base it off current year instead of hardcoded one, but works for now. Also reduces possible days to 28 to avoid February 31st situation. Ideally should also be done differently.


Greenjoe12345:
Funny thing with that
Once, upstream on Polaris, my teshari's account creation was years before they even hatched

PR #8918, opened 2020-08-30
Refactors random list prob chances

Verkister:
Chances are now scaled in relation to the highest. Couldn't find any proper documentation on how the pick weight stuff is actually calculated, but I reckon it wouldn't hurt to take out one unnecessary step out of the equation and to make the probabilities actually readable in their context. (such as not making something have prob(1) in relation to the highest prob being 5 and telling it has a 1% chance to be picked when in reality that chance would be 20%)

Also reduces emag probability to not be on par with glowsticks and pizza vouchers.


Verkister:
travis killed by random map edge

Verkister:
The rest would be polaris changes

Cyantime:
This isn't at all how that worked, there was no possibility of "picking nothing and rerolling" because "nothing" isn't a value in the list. The prob(x) only modify the relative weights of the items in the lists,
`pick(
	prob(1); /obj/A,
	prob(3); /obj/B,
)
`
has a 25% chance of picking A, a 75% chance of picking B, and a 0% chance of somehow picking nothing and deciding to try again without being told.

Verkister:
gonna do a test run anyway. the documentation for the procs is not very convincing at making clear what it actually does.

Verkister:
Test run showed no issues or changes in picking behavior, unable to really tell if there was performance improvement due to being local singleplayer shit. I'm still unable to find anything to explain the maths behind the procs on how it supposedly weights the options other than a single off-hand comment from 2002 about it being hardcoded somewhere. And even then this only takes out one unnecessary step out of the equation

Cyantime:
The fact that you can do prob(200) or any number above 100 and have it work in a useful way suggests that it's not *actually* using the normal prob() proc, and just using the same format for those extra arguments. They also work as `pick( 1;/obj/A, 3;/obj/B)`, it probably isn't just running prob(x) for each thing in the list and passing whatever succeeds.

Verkister:
Yeah, I've pretty much figured it out that far by now. Anyway, when there is no non-prob baseline (equivalent of prob(100) which is nothing but unnecessary garbo clutter) or another prob above 100, there is no need to add unnecessary conversions that also skew the perspective for people reading the code (as in thinking prob(1) is 1% while in reality, having it be paired with a maximum prob of 5 makes it 20%). Really the only benefit of not using the actual percentages is for lazily adding a "more likely" option without needing to update the probs for the rest of the list and even then that is only relevant if the list already has a baseline 100% option.

Cyantime:
`(as in thinking prob(1) is 1% while in reality, having it be paired with a maximum prob of 5 makes it 20%`
This is only true if there are only those two options in the list, your new values make it *more* likely to confuse them with percentages. For a pick like
`pick(
	prob(5);/obj/A,
	prob(4);/obj/B,
	prob(3);/obj/C,
	prob(2);/obj/D,
	prob(1);/obj/D
	)
`
The chance of picking B isn't 4% or 40% or whatever, it's 4/15. There aren't any percentages, it's a list of weights.

Verkister:
Not according to the code references or the one forum comment I found on top of them. According to those it is percentages in relation to the baseline. Like a prob list of none/100, 50, and 200, it just converts these to 50%, 25%, and 100%. And with probs of 1, 2, and 3, they convert to 33%, 66%, and 100%.

Cyantime:
How would that actually work, though? Those percentages are over 100%, they'd need to be added and then divided to find the actual likelihood of each option. For that second prob of 1,2,3 33%,66%,100%, that'd work out to 16.6%, 33.3%, 50% for the real chances. This is the same as the 1/6, 2/6, 3/6 you get from just considering them as weights. It's simpler and easier to understand if you just don't consider percentages in this at all, trying to forcefully normalize every pick() around 100 makes it worse.

Verkister:
just gonna close for now. Math ain't my fave subject and this rabbit hole of a code is starting to look a lot more like a meat grinder.

PR #8922, opened 2020-08-30
Add the inducer to mining & survey vendors.

lbnesquik:
They are relatively expensive so it is still better to bother competent R&D.

### DIFF
--- a/code/modules/mining/ore_redemption_machine/equipment_vendor.dm
+++ b/code/modules/mining/ore_redemption_machine/equipment_vendor.dm
@@ -91,8 +91,9 @@
 		EQUIPMENT("Fine Excavation Kit - Measuring Tape",	/obj/item/device/measuring_tape,				125),
 		EQUIPMENT("Fine Excavation Kit - Hand Pick",		/obj/item/weapon/pickaxe/hand,					375),
 		EQUIPMENT("Explosive Excavation Kit - Plastic Charge",/obj/item/weapon/plastique/seismic/locked,	1500),
-		EQUIPMENT("Industrial Equipment - Phoron Bore",	/obj/item/weapon/gun/magnetic/matfed,				3000),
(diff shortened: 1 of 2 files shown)


lm40:
to be honest i'm not 100% sure what they would be used for but sure why not

code/modules/mining/ore_redemption_machine/equipment_vendor.dm
```diff
@@ -91,8 +91,9 @@
 		EQUIPMENT("Fine Excavation Kit - Measuring Tape",	/obj/item/device/measuring_tape,				125),
 		EQUIPMENT("Fine Excavation Kit - Hand Pick",		/obj/item/weapon/pickaxe/hand,					375),
 		EQUIPMENT("Explosive Excavation Kit - Plastic Charge",/obj/item/weapon/plastique/seismic/locked,	1500),
-		EQUIPMENT("Industrial Equipment - Phoron Bore",	/obj/item/weapon/gun/magnetic/matfed,				3000),
-		EQUIPMENT("Industrial Equipment - Sheet-Snatcher",/obj/item/weapon/storage/bag/sheetsnatcher,		500),
+		EQUIPMENT("Industrial Equipment - Phoron Bore",		/obj/item/weapon/gun/magnetic/matfed,			3000),
+		EQUIPMENT("Industrial Equipment - Inducer",			/obj/item/weapon/inducer,						900),
```
Novacat, line 95: General rule of thumb is that the Mining prices are 10x the Survey prices

lbnesquik: Ok

lbnesquik: Fixed

nerdass:
Don't really see anything bad about this since it's adding the one that specifically CAN'T recharge guns.

JoanRisu:
I concur with K't. Unless I'm missing something, I think we're good on this one.

PR #8923, opened 2020-08-30
Add an Inducer crate to cargo. Relatively expensive at 90.

lbnesquik:


lm40:
sure why not

nerdass:
Don't really see anything wrong with this since it's the one that's specifically coded to not work with guns.

PR #8924, opened 2020-08-30
You can now use the inducer on the APC. Probably.

lbnesquik:
https://github.com/Citadel-Station-13/Citadel-Station-13-RP/pull/2047/files


lbnesquik:
rhpphph

PR #8927, opened 2020-08-30
Fix jetpack ion trail

Cyantime:


Cyantime:
yet again i thought I was on the polaris repo, oh well

lbnesquik:
What was broken?

Cyantime:
It didn't work at all. The only times they'd make the trails is if you turned on a jetpack, dropped it on the floor, and dragged that around.

PR #8929, opened 2020-08-31
Un-electrifies the theatre grilles

Cyantime:
why, this isn't a secure location, there's no reason for there to be potentially lethal power in these windows


Rykka-Stormheart:
Wouldn't that prevent the windows from being tinted?

KillianKirilenko:
No. Window tinting is handled by using wire on the glass, not cable under it. Otherwise the electrochromic windows in the medbay patient rooms wouldn't work either.

PR #8931, opened 2020-08-31
You can now use the inducer on the APC. Probably.

lbnesquik:


code/modules/power/apc.dm
```diff
@@ -718,6 +718,9 @@ GLOBAL_LIST_EMPTY(apcs)
 	..()
 	togglelock(user)
 
+/obj/machinery/power/apc/get_cell()
+	return cell
```
Novacat, line 722: This can probably go in apc_vr.dm and not need a //VOREStation add

Novacat (approved):

PR #8964, opened 2020-09-02
Remov medical access to mining

lbnesquik:
Pretty much unjustified, given they have no access to pretty much everywhere else.  And mining has a lot more toys worth stealing than most department besides research.


Heroman3003:
Medical access to Belter shtutle was specifically requested when it was being added, for rescue operations to be possible in case of emergency.

PR #8965, opened 2020-09-02
Remove medical access from cargo.

lbnesquik:
Duplicate of #8964 because freaking map edges.

### DIFF
--- a/maps/tether/tether-07-station3.dmm
+++ b/maps/tether/tether-07-station3.dmm
@@ -2189,13 +2189,19 @@
 /turf/simulated/wall/r_wall,
 /area/quartermaster/belterdock/surface_mining_outpost_shuttle_hangar)
 "adE" = (
-/obj/structure/shuttle/engine/propulsion{
-	dir = 8;
-	icon_state = "propulsion_l"
+/obj/machinery/door/airlock/maintenance/cargo{
+	name = "Belter Shuttle Access";
+	req_access = list();
+	req_one_access = list(48,65)
 	},
-/turf/simulated/floor/tiled/asteroid_steel/airless,
-/turf/simulated/shuttle/plating/airless/carry,
-/area/shuttle/mining_outpost/shuttle)
+/obj/machinery/atmospherics/pipe/simple/hidden{
+	dir = 4
+	},
+/obj/machinery/atmospherics/pipe/simple/hidden/supply{
+	dir = 4
+	},
+/turf/simulated/floor/plating,
+/area/quartermaster/belterdock)
 "adF" = (
 /obj/effect/floor_decal/borderfloor{
 	dir = 8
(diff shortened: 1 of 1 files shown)


Heroman3003:
Medical access to Belter shtutle was specifically requested when it was being added, for rescue operations to be possible in case of emergency.

Also if issue is map edges in wilderness, just closing-reopening same PR works, no need to remake it.

lbnesquik:
I tried three times on another PR and it didn't work, so i remade it. 

Anyway, it was requested but that doesn't means it has to stay here.

KasparoVy:
My colleagues & I have been saved a few times by the specifically req'd access to the belter room, it's come in very handy - especially because of scenarios like...

1. Be teshari, beltmining
2. Spacebear smacks you, woops!
3. 20 seconds later-- oh shit, why's everything going fuzzy. OH FUCK IT'S IB
4. Teleport onto the shuttle and force send right the fuck away, stabbing self with suffocation injector.
5. Shuttle flies for long time, as it tends to do on occasion
6. Pass out on the way back, NIF barely keeps you alive - even with respirocytes
7. Medical somehow was looking at the scanner the moment you arrived and suddenly sees your ass knocking on death's door
8. No-one is in cargo
9. They walk thru the maint shaft and recover the casualty
10. Bless mon bless

Folks who want to break into mining have done so by hacking, falsewalling or deconstructing anyhow -- not that I've seen it more than twice ahaha. I place the highest trust in medical out of any department usually

lm40:
ideally qm would help in the event of an emergency and regardless medical has ways to break into departments

Heroman3003:
Well, as long as access to shuttle console itself remains, yeah it will be largely irrelevant, yeah. But idea behind original access is to allow medical to help when. Well. Nobody else can reach. If mining gets hurt mining outside, medical can get outside easily. If it happens while belt mining? Belter's your only option.

Novacat:
When the belt shuttle was originally added, Cargo, Medical, and Xenoarch all had access. Due to an few incidents involving Science, Xenoarch's access was removed. Medical has not had any similar issues.

lbnesquik:
Only the doors are affected.

Wickedtemp:
Not convinced that this is necessary for any real reason. Yes, ideally, we would just ask another cargo crew member. I do that anyway. 

But, you also kinda have to remlember how common it is to be like "Hey, I need access to your department, is that okay" and get nothing in response because they're in a scene. Since this was an intended addition in the first place, I'd kinda like to see actual examples of this causing issues before it's removed.

Wickedtemp:
...been nearly a week, no examples of problems caused by the access and nothing in terms of how its removal would be of benefit.

Very-Soft:
I think medical should have access to the belt shuttle, and not have access to like, the normal cargo bay area. 

Medical has reasons to go to the mining areas for rescue, and a lot of times if cargo has anyone in it, it's usually miners, and sometimes those miners need rescue.

So, I'm not super sure which doors this all affects, but if it's the cargo bay/office, then good, if it's a back door to the mining shuttle, then that's lame.

Heroman3003:
If I'm reading map diff correctly, this only affects two doors related to belt shuttle, one from maint and one to control room. Basically for medical to enter through maint specifically, then go directly to the shuttle through airlock. Rest of cargo/mining seems to remain untouched.

nerdass:
not really sure what the issue with this access existing in the first place is? sounds like it's intended as deadshift insurance.

Wickedtemp:
So far, there is no issue or reason. Which, in and of itself, is an issue for this change, since... the access was intended in the first place.

PR #8968, opened 2020-09-03
Fix bumpnoms

Meghan-Rossi:
Gets bumpnoms working.  Tested it with a fennec, seems to work fine.


Rykka-Stormheart:
Well done! One more thing to take off the to-do list.

PR #8629, opened 2020-08-08
 TGUI Engineering: Less broken Travis Edition (:crossed_fingers:)

ShadowLarkens:
See this album for screenshots of every UI https://imgur.com/a/NVGgKXI

This overhauls all of engineering to use TGUI. I mean *all* of engineering. Every single UI I could think of, and every single UI suggested by the engineering channel.

Full list below.
# Atmospherics
 - Algae Generator
 - Atmospherics Control
 - area_atmos/tag (Heavy Scrubber Control)
 - Gas Cooling System
 - Gas Heating System
 - general_air_control (Distro & Waste)
 - general_air_control/fuel_injection (Riot Control Console)
 - general_air_control/large_tank_control (Atmos Intake Control)
 - general_air_control/supermatter_core (Engine Cooling Control)
 - Gas Pump
 - Handheld Tanks
 - Omni Filter
 - Omni Mixer
 - Passive Pump
 - Portable Scrubber
 - Portable Pump
 - Portable Tanks
 - Tank Storage Unit
 - Trinary Filter
 - Trinary Mixer
 - Shutoff Valve Monitoring
 - Disposal Pipe Dispenser
 - Pipe Dispenser
 - Rapid Pipe Dispenser

# Alarms
 - Atmos Alerts
 - Air Alarms
 - Fire Alarms (removed UI)
 - Station Alerts

# Etc
 - Airlock Controller
 - Doors
 - Drone Fabricator
 - Elevator
 - Employment Records (related enough, right?)
 - Gravity Generator
 - ID Cards
 - Intellicards
 - Message Monitoring
 - Radios
 - RIGSuits
 - Space Heater
 - Signallers
 - Suit Cycler
 - Suit Storage Unit
 - Vending Machines
 - Wires!

# Power
 - APCs
 - PACMAN
 - Particle Accelerator Control Panel
 - Power Monitor
 - RCON
 - Shield Generator
 - SMES
 - Solar Controls
 - Shield Capacitor
 - Supermatter Monitoring
 - Silicon/Observer SM Monitoring
 - TEG

# Telecomms
 - Exonet
 - Server Monitor
 - Telecomms Monitor
 - Telecomms Multitool Menu
 - Quantum Relay

PR #8992, opened 2020-09-06
Engineering Mod-Computer Conversion.

Andromeda-K22:
This PR replaces a lot of the non specific computers (station alert consoles, power monitoring, RCON, etc) with Modular Computers. It seems like a good idea, specifically in conjunction with #8990 since modular computers *are* more flexible.

I should note that I'm not sure if 8990 touches the engineering programs, but regardless, they'll probably be eventually converted.


Heroman3003:
Specific consoles > MCs

Andromeda-K22:
I, too, enjoy running between tables to check two different computers.

ShadowLarkens:
> I should note that I'm not sure if 8990 touches the engineering programs, but regardless, they'll probably be eventually converted.

That's because #8629 already converted all of them, #8990 completely removes NanoUI from modular computers :p

Novacat:
One of the issues with going full MC would be that MCs require Telecom access to function. Perhaps an alternative solution would be to keep the dedicated consoles but add a single MC console where needed.

Andromeda-K22:
In what situation would tcomms not be present on the main map? Ion storms?

I don't see the problem with this - five minutes of unavailability isn't really critical.

Torque4607:
Speaking from experience with servers like Bastion, Bay, and Urist, modular computers are much more flexible and convenient to use versus the dedicated consoles.

Novacat:
Ion storms, or shifts where the engine was not set up, tcomms usually goes out long before the rest of the station because of the insane power consumption. Compared to Bay, at least, TComms is a lot more finicky/temperamental on Virgo.

Andromeda-K22:
That's because on all other servers, Tcomms runs on it's own fully charged
SMES by default. It does not here.

Cyantime:
As it shouldn't, what's even the point of telecomms needing power if it just has with all the power it'll ever need at roundstart

Andromeda-K22:
Here's the thing.

It doesn't. Tcomms is power hungry, but it will last long enough with the
engine off. It isn't going to stay online through a six hour round.

PR #8993, opened 2020-09-06
Plant analyzer fix

ShadowLarkens:
This should hopefully prevent the data in the plant analyzer from disappearing when a seed mutates.

I was really hoping that /datum/seed's were immutable and never qdel'd, but apparently I was wrong.

This isn't tested, but theoretically works. I don't understand enough about botany or xenobot to actually make sure this works.


ShadowLarkens:
@Novacat Irrelevant for Polaris, they don't have TGUI plant analyzers yet, and this is only applicable to the TGUI version.

PR #8996, opened 2020-09-07
Wall / Material Radiation Tweaks.

Andromeda-K22:
🆑 
Walls/Rwalls now have a tweaked radiation calculation equation, taking in to account the *entire* construction of the wall. (covering material, girder material, reinforcing material).

Steel and Plasteel now actually have a radiation resistance value. They did not previously. It was zero.
🆑 

No longer must engineering put wet paper towels on the walls to shield people!


Novacat:
tbh I think Polaris would really like this

Andromeda-K22:
I don't believe they would as I tried the same thing some time ago and was rejected.

Cyantime:
This makes regular steel better at blocking radiation than much denser materials, like gold or platinum, and nearly as good as durasteel. Plasteel is now much better than durasteel, these numbers don't make much sense and I feel the added rad resistance is unneeded.
You're already flat doubling the blocking of regular walls, as they're now steel girder+steel plating, and almost tripling rwalls as steel girder+plasteel reinforcement+plasteel plating. The engine currently leaks no radiation to the control room when making more than enough power for the station, if you want to crank it up to much higher power levels (which are unneeded) then adding more shielding or needing a radsuit makes sense.

Andromeda-K22:
-Who is going to build walls out of solid gold.

-Plasteel is steel and platinum.

I specifically mirrored changes on useful materials that are commonly used on the station. 

I should point out that an upgraded engine still leaks even with these walls upgraded; It just doesn't flood Engineering with radiation up until the _hard thirty tile limit of the raycasting._

There is still the matter of the radiation divisor making this less effective than you'd think. I can boost durasteel if you want, but it's principally never used because it's so rare.

Iroquois-Pliskin:
As someone who regularly plays Engineering and frequently kicks the engine to the point that it leaks radiation into maint, this is a welcome change- as it stands I don't think plasteel walls even BLOCK radiation. This is long overdue.

Cyantime:
`There is still the matter of the radiation divisor making this less effective than you'd think`
The radiation material divisor _is not used_ on this server. Have you been testing with default configs to make your comparisons? If so, that'd explain how you're somehow hitting the radiation limit even when every rwall in engineering has like ten times as much rad resistance.

Andromeda-K22:
... yeah rwalls seemingly don't block anything, because I ran a modified engine once and I was nuking things almost out to the lobby.

You're correct about the divisor not being used, that was my mistake as I misread something that was posted on Discord.

Regardless, the sheer amount of rad leakage was pretty insane even with Rwalls in place. I don't really see this as overkill.

Cyantime:
Currently, rwalls do block radiation, but only a tiny bit more than regular walls. Allowing reinforcement to count for blocking would greatly increase that and make them useful. Again, have you tested things with the right server config settings? The defaults are hilariously useless and block effectively nothing regardless of material or number of walls.

Andromeda-K22:
I've been doing some testing to narrow in the values to strike a decent
balance between danger and usefulness. I should take a sec to note
durasteel never had radiation resistance which is somewhat amusing. I'l
probably give it some to make it useful, as well as add it to gold and lead
- if that's the concern, anyhow.

I'l be messing with material values in terms of radiation resistance
shortly and have already removed the girders from the calculation.

Cyantime:
Durasteel didn't have _bonus_ radiation resistance, it still had higher total rad resistance than any other normal building material based on its higher weight.

Andromeda-K22:
Squashmerge if possible, please.

Andromeda-K22:
Calculation changes pushed upstream.

Novacat (changes requested):
Conflict.

Andromeda-K22:
Given this took so long to actually get glanced at and in the intervening time I left the community, I will be closing this.

PR #9007, opened 2020-09-09
Attempt to fix duplication in the crew monitor

ShadowLarkens:
I don't have any known way to replicate the bug, so... maybe this will fix it?


TheFurryFeline:
Oh, shit, we're at over nine thousand now, lmao. Missed my chance for a dorky meme somewhere

PR #9017, opened 2020-09-12
Add Rapid Pipe Dispenser to R&D

ShadowLarkens:
Origin tech, req tech, and materials are all up for debate, I just copied the RCD for tech and went with values that seemed vaguely close to the surrounding tools for materials. I just want this thing to be obtainable outside of a *single* instance placed on Talon (who have no use for it) lol.


Heroman3003:
Imo, material cost should involve at least some mining materials. Also if its to be R&D item, research levels would make sense to be a little higher for making it than it actually has itself.

chaoko99:
Making it use plastic for such a trivial tool is probably going to completely assure it is never made.

PR #9020, opened 2020-09-13
Anti-Pref Breaking of Shadekin/Demon Boyes

Shadow-Quill:
Stops shadekins/demons/protean blobs from phasnomming/bloodnomming/surroundnomming prey who aren't devourable or a drop prey.


code/modules/mob/living/carbon/human/species/shadekin/shadekin_abilities.dm
```diff
@@ -95,7 +95,7 @@
 			var/list/potentials = living_mobs(0)
 			if(potentials.len)
 				var/mob/living/target = pick(potentials)
-				if(istype(target) && vore_selected)
+				if(target.devourable && target.can_be_drop_prey && istype(target) && vore_selected)
```
Heroman3003, line 98: istype(targer) should probably come before the checks, just in case. Applies to other lines too.

PR #9027, opened 2020-09-16
Stripping.dm Cloning issue bugfix

cadyn:
Context: There's a bug with stripping where if you hold someone in your hand, you can use the stripping menu to put them on themselves, which makes a clone of them that starts breaking shit, as you can imagine.

Why is there no upstream of this? Because Polaris won't have this issue since they don't seem to have size mechanics and being able to hold people as we do.


Heroman3003:
Polaris has a thing with holding humanmobs in hands, but thats exclusive to teshari, I believe. Can also 'put them on' as backpacks I think.

cadyn:
> Polaris has a thing with holding humanmobs in hands, but thats exclusive to teshari, I believe. Can also 'put them on' as backpacks I think.

Beh, that would require a separate fix anyways.

KasparoVy:
I've been picked up and worn as a Teshari before, I don't remember how it happened but it was adorable and surprising

code/modules/mob/living/carbon/human/stripping.dm
```diff
@@ -75,9 +75,15 @@
...
-		return
+	//VOREStation Edit Start
+	if(!stripping)
+		if(user.get_active_hand() != held)
+			return
+		var/obj/item/weapon/holder/micro/microheld = held
+		if(istype(microheld)&&microheld.held_mob==src)
+			to_chat(user, "<span class='warning'>You can't put someone on themselves! Stop trying to break reality!</span>")
+			return
+	//VOREStation Edit End
```
ShadowLarkens, line 86:
```suggestion
if(!stripping)
	if(user.get_active_hand() != held)
		return
	if(held == loc)
		return // No recursion
```
This should work much better and not be so hacky.

code/modules/mob/living/carbon/human/stripping.dm
```diff
@@ -75,9 +75,11 @@
 
 	if(!do_after(user,HUMAN_STRIP_DELAY,src))
 		return
-
-	if(!stripping && user.get_active_hand() != held)
+if(!stripping)
+	if(user.get_active_hand() != held)
 		return
+	if(held == loc)
+		return // No recursion
```
ShadowLarkens, line 82:
```suggestion
	if(!stripping)
		if(user.get_active_hand() != held)
			return
		if(held == loc)
			return // No recursion

```

cadyn: What? held == loc? This doesn't even work. Loc would be the person holding the micro in this case. Also, why get rid of the message? People need to know that trying to break reality is unacceptable.

cadyn:
Force pushing back to original because non-working code was committed.

Novacat:
Hm. On the other hand Larken's fix was cleaner (and maybe could be ported to polaris?), but if it does not work then probably a better solution can be done?

cadyn:
Actually, looking back at it, yeah I think it could be made upstreamable by getting rid of the micro part of the variable path but it's still gonna be doing the same thing.

cadyn:
https://github.com/PolarisSS13/Polaris/pull/7658 I guess I'll see you in a few days then boys

code/modules/mob/living/carbon/human/stripping.dm
ShadowLarkens: Micros do not get directly held. The micro's actual `loc` is the `/obj/item/weapon/holder` that they're contained within, which is in turn contained within the user. I just forgot to add the message.

ShadowLarkens:
My bad with the first change not working, but the second definitely should.

cadyn:
Right, so, even though I already closed this and sent it upstream, I feel like I need to make a final comment for some closure. @ShadowLarkens You were right, held==loc was the better solution since it's cleaner. I would have committed this to the Polaris PR but since downstream merged one of my PRs before the upstreams were merged, I'm gonna leave it as it is for the sake of consistency. It shouldn't effect functionality, and in a way I feel like my version of it is more readable, (someone looking at it unfamiliar with the holder class might be like wtf loc == held how does that even make sense? whereas with mine it's pretty clear that we're checking to see if someone is putting another persons holder onto themselves). I just wanted to make sure it didn't seem like I was trying to avoid feedback, the reason I said loc==held doesn't work was because when I went to check what loc was equal to I checked the holder instead of the mob itself and I didn't process it fully, I guess.

tldr, Your solution was cleaner, and would have worked, but downstream merged early so now I'm just gonna leave it for consistency sake.

PR #9030, opened 2020-09-16
Added Fire Fighting Foam Recipie

VictorZisthus:
For when you want to play around with the nifty water foam without draining all the actual tanks. Have fun with spray bottles and chem sprayers!

### DIFF
--- a/code/modules/reagents/Chemistry-Recipes_vr.dm
+++ b/code/modules/reagents/Chemistry-Recipes_vr.dm
@@ -77,6 +77,14 @@
 
 /datum/chemical_reaction/foam/softdrink
 	required_reagents = list("cola" = 1, "mint" = 1)
+	
+/datum/chemical_reaction/firefightingfoam //TODO: Make it so we can add this to the foam tanks to refill them
+    name = "Firefighting Foam"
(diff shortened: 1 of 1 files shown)


KasparoVy:
Hell foooooam!

VictorZisthus:
Another thing I'd really like to do is make it so that you can not only refill the tanks and stuff with this, but to make it so you can do so rapidly without the use of a chemmaster and bottle waste; Have the tank not accept the fluorine when dumping the contents in, allowing rapid refilling with just a click on a sink and the tank.

VictorZisthus:
> 
> 
> Another thing I'd really like to do is make it so that you can not only refill the tanks and stuff with this, but to make it so you can do so rapidly without the use of a chemmaster and bottle waste; Have the tank not accept the fluorine when dumping the contents in, allowing rapid refilling with just a click on a sink and the tank.

On that note, we could even have a machine that just foams up the water instead...

Zeldazackman:
Danke for adding this, as I was thinking about how lovely it'd be to have foam accessible for fighting fires and general walls of heat via the chem dispenser in a "certain situation" that arose due to xeno bio being well, xenobio

PR #9034, opened 2020-09-17
Disables Morph Spawn random event

Heroman3003:
Due to multitudes of issues, this absolute kneejerk of a PR is brought up. Should the morphs have some morphing abilities adjusted to be pref-fitting and their behaviour be more moderated, the event could probably be re-enabled.


VictorZisthus:
Honestly, I think that the best thing to do would be to allow atmos scanners to be able to detect things in the pipes somehow, or otherwise just punish stationary targets by having the ventcrawl sound still go off on an interval. No need to remove the event as a whole, just make them detectable.

Heroman3003:
It doesn't matter if they are detectable because there is no way to remove them and burning a section of a pipe is a pain in its own right, especially so if its a supply pipe

MrSign:
This will be up for staff discussion but community input would be appreciated as well.

Heroman3003:
If the morph behaviour is to continue to not be moderated, they probably should also not have the ventcrawl ability.

Very-Soft:
This PR is silly. I would vastly prefer the ability to ban people from this role to just disabling it like this.

Novacat:
This is not really necessary

TheFurryFeline:
Late to the party but agreed with Soft.

PR #9036, opened 2020-09-17
Makes morph impersonation respect body impersonation preference

Meghan-Rossi:


Heroman3003:
Should probably also ignore afk/SSD humanmobs regardless of that pref tbh.

Novacat:
For now this will do.

PR #9039, opened 2020-09-17
Changes the Guttersite to an Offsite Spawn

SweetBlueSylveon:
This is a joint effort to make the Guttersite an offsite spawn. 
Functionality to be added:
- Turning the vault into something similar to the cargo shuttle. This way exploration or the talon can come throw all that loot they find each shift in there for unique GUTTERPOINTS! (Or just cargo point vouchers.) They will be able to exchange these points for things they either couldn't get on station, would be more difficult to get on station, or is more expensive on station.
- Add a fueling point to refuel phoron canister in exchange for said gutterpoints.
- Add a basic telecomms relay so the guttersite can have communications.
- Add Engineer, Medical personnel, Site Manager, and Technician as gutter jobs. Engineer sets up the engine, fixes up the teleporter, and fixes the place up. Medical personnel are basic medical assistance. Site manager is the one who watches the place for the shift, deals with the exploration crew and talon crew, manages telecomms, leads the crew on eva  to clear space mobs. And the technician runs the front desk, makes orders, fuels tanks, interacts with customers, etc. And ultimately, after anyones' job is finished, they can man the bar or rp with eachother or guests.


Screemonster:
hype

VictorZisthus:
MUCH HYPE

aryll-linton:
Pip has ~~forced~~ recruited me to help with this which likely means working on the explo/talon-guttersite interface. Hijacking the existing cargo shipping system to send away common explo loot (guns, ammo, coins, russian uniforms- cool stuff that isn't useful but seems a shame to throw away) for the so-called "gutterpoints" shouldn't be too hard (famous last words) but it'll sure take a long time to weed out the edge cases. In any case, **HYPE.**

SweetBlueSylveon:
Unless a coder can create the functionality to swap maps out round to round, this will not be possible. As that was the stipulation I was given for doing this. If this functionality is created in the future I may reconsider this.

PR #9084, opened 2020-09-22
Phoron bore utility improvements

Cyantime:
The phoron bore: 

- Uses *significantly* less cell charge per shot. 750 gives you 20 shots with a regular 15k cell (ignoring material ammo cost), each of which will dig about six tiles. Compared to a drill, or a KA, or basically anything and this is woefully expensive for the actual mineral output. It now uses 100 charge per shot.

- Can be reloaded with raw plasma ore in addition to plasma sheets. Each ore gives 800u of plasma, compared to 2000 per sheet. Two ores compress into one sheet, minus some inefficiency for using unrefined material. Carrying a stack of sheets is still more effective, but it can now be reloaded in the field with extracted material.

- Has a tiny generator built in that runs off the stored plasma. It charges the cell, or capacitor directly if there's no installed cell, at a rate of 20kW. This is the same as a basic pacman set to output level one, but uses about 25% more fuel and produces exhaust that can potentially be a hazard in very enclosed spaces with no scrubbers.

- Comes pre-installed with a basic power cell and capacitor when bought from the mining vendor.

- Does some extra damage to objects.

Also did some minor cleanups in the magnetic/bore files. Screwdrivers/crowbars use their toolsound, removed some unneeded forceMoves, changed a modulo that gave infinite ammo to a division that doesn't, etc.


KasparoVy:
This is fucking awesome. I love the bore.

Cyantime:
Included in https://github.com/VOREStation/VOREStation/pull/9207

PR #9087, opened 2020-09-22
Added QOL Firefoam recipie

VictorZisthus:
Dunno why I didn't think of this before.

### DIFF
--- a/code/modules/reagents/Chemistry-Recipes_vr.dm
+++ b/code/modules/reagents/Chemistry-Recipes_vr.dm
@@ -85,6 +85,15 @@
     required_reagents = list("water" = 1)
     catalysts = list("fluorine" = 10)
     result_amount = 1
+	
+/datum/chemical_reaction/firefightingfoamqol //Please don't abuse this and make us remove it. Seriously.
+    name = "Firefighting Foam"
+    id = "firefighting foam ez"
+    result = "firefoam"
+    required_reagents = list("water" = 1)
+    catalysts = list("firefoam" = 0.01)
+	inhibitors = list("fluorine" = 0.01)
+    result_amount = 1
 
 ///////////////////////////////////////////////////////////////////////////////////
 /// Vore Drugs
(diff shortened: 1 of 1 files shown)


VictorZisthus:
Whoops, looks like I need to make a new name for the QOL recipie.

That or Travis is just being a dick again.

I'll solve this when I get home, it won't let me edit it from mobile.

VictorZisthus:
I figured it out.

:galaxybrain:

code/modules/reagents/Chemistry-Recipes_vr.dm
```diff
@@ -85,6 +85,15 @@
...
     catalysts = list("fluorine" = 10)
     result_amount = 1
+	
+/datum/chemical_reaction/firefightingfoamqol //Please don't abuse this and make us remove it. Seriously.
+    name = "Firefighting Foam EZ"
+    id = "firefighting foam ez"
+    result = "firefoam"
+    required_reagents = list("water" = 1)
+    catalysts = list("firefoam" = 0.01)
+	inhibitors = list("fluorine" = 0.01)
```
KillianKirilenko, line 95: your problem's here, needs a tab

VictorZisthus: I don't follow. Where is the tab needed?

ShadowLarkens: You either need to replace all of the spaces here with single tabs, or change this line to four spaces instead of a tab
```suggestion
    inhibitors = list("fluorine" = 0.01)
```

VictorZisthus:
Oh NOW it passes.

Cyantime:
this turns firefoam into some sort of horrifying ice-9 reagent? why, it isn't even any better than water at the job and you already just added a recipe for it in the event that anyone needed any outside of the extinguisher refill tanks

VictorZisthus:
> 
> 
> this turns firefoam into some sort of horrifying ice-9 reagent? why, it isn't even any better than water at the job and you already just added a recipe for it in the event that anyone needed any outside of the extinguisher refill tanks

Mostly as a way to reduce bottle clutter from constantly mixing fluorine and water, then having to distill them in a chemmaster 3000.

A much more optimal way of doing things would be to have a device that either makes water into firefoam, or allows mixing of chemicals with catalysts WITHOUT transferring the catalysts themselves when pouring.

I'm just too new at coding to do anything like that.

PR #9105, opened 2020-09-25
Add U as hotkey to rest

ShadowLarkens:
In hotkey mode, you can hit U to rest/unrest.

Follows the keybinding from https://github.com/tgstation/tgstation/pull/53709

Alas, we don't have custom keybindings (other than user skin edits)
so U it is

better than no hotkey innit

### DIFF
--- a/interface/skin.dmf
+++ b/interface/skin.dmf
@@ -564,6 +564,9 @@ macro "hotkeymode"
 	elem 
 		name = "CTRL+Z"
 		command = "Activate-Held-Object"
+	elem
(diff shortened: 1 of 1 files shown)


KasparoVy:
Oh cool. I can delete macro now :D

Cyantime:
`better than no hotkey innit`
This is not always true, now 'u' is unusable for byond macros.

KasparoVy:
At least we can still use shift + stuff. Really annoying caveat to the hot key system on some other codebases

PR #9107, opened 2020-09-25
Oregrubs

KillianKirilenko:
I decided to try recoloring the solargrub for a laugh, and ended up with something vaguely presentable! So here we are; oregrubs and lavagrubs. ~~What, loot bugs? I have no idea what you're talking about.~~

image

These leathery lithovores and monochrome metallovores can be found slithering around anywhere there's delicious ore to chew on, and can survive just about anywhere (which is to say they spawn in the UD, V4 caves, and V2 surface). They are passive until threatened, at which point they will flee at high velocity - but if you catch them and cut them up, you can extract those minerals! Lavagrubs contain even more, rarer ores! Run them through a gibber to get even more! _D e l i c i o u s._

Grubs are moderately resistant to melee weapons and kinetic accelerators, but fairly weak to ballistics and lasers. They provide a new form of cave life for miners and explorers to encounter that *isn't* automatically hostile! Goodness, look at me finally making good on something that's come up in the exploration channel before. You can even catalogue them for a few points.

Probably needs unique sprites at some point?


KasparoVy:
Man this is a throwback, I remember old bay mining in space on Para years ago I'd try hunting them down to give the RD a fun new pet.

Awesome

Novacat (changes requested):
Everything looks fine but I feel like they are a wee bit too common, considering that they spawn pretty much everywhere. I would like for them to be in a more... distinct location. I would recommend Virgo two actually, with maybe a few rare ones showing up in the underdark.

KillianKirilenko:
Done. Lavagrubs will no longer appear on V4 period, and in general both kinds ought to be a fair bit rarer than they were before.

Novacat (approved):

PR #9110, opened 2020-09-26
Remove all Robot Command verbs, replace with an NTOS Computer with Robotact App

ShadowLarkens:
https://github.com/tgstation/tgstation/pull/53373

Highlights:
 - No more "Robot Commands" verbs. All functions replicated within Robotact app.
 - Robots now start with an integrated modular computer. There is no downloader, but they can have flash drives stuffed in their butts to load new software at the will of robotics.
 - Three new robot UI buttons- take picture, toggle headlamp, open modular computer.
 - Robots can set their light color.


Cyantime:
Why are you getting rid of the verbs as well? The UI is very pretty and nice, but the verbs are much quicker and easier to use, as well as being able to be done via byond macro. Futzing around with an entire separate window is something that I'd prefer to avoid for normal operations if at all possible. This also removes the "drop item" verb from grippers, which is a regular macro usage.
Other than that, unlocking their cover isn't an ability that borgs currently get but is added by this that might have "law 2, give me your power cell" type implications

ShadowLarkens:
The drop item verb is possibly the only verb that makes any sense to keep, maybe the emit sparks one as well?
Headlamp is now a HUD button, Namepick really doesn't need quick access, Show Laws is deprecated by both the law panel and this new interface, State Laws already just shunted you to the law manager, Self Diagnosis pulled up an ugly functionless menu that's now in the interface, and toggling components really doesn't need to be done that often.

ShadowLarkens:
And besides that, having *only* those two in an otherwise empty tab is a hilariously big waste of space

VictorZisthus:
I use macros for a lot of the robot stuff, so I'd prefer those be kept. The UI is nice though!

As for the ability to open or close their own panel...

I'm very certain that NT would not want anyone other than authorized personnel tampering with a borg's electronics, and the default lawset could very easily be interpreted as Cyantime said above: "Law 2, Open your hatch."

NT won't be very happy if a borg misinterprets it's laws and gets tempered with as a result. I do like the ability for the Cyborg to check if their hatch is open or not without a shift-click though.

But yeah. Robotics, RD, and maybe the CE should be the lowest level jobs to be able to access a borg's internals at all. Anything else, the borg included, is asking for trouble.

[Cyborgs with a Robotics module installed are an exception, as they do have the access code device to lock and unlock ANY cyborg chassis, even their own.]

chaoko99:
I am all for the removal of verbs, myself, as they contribute significantly toward right click lag. For those unaware, the right click system will build a list of all verbs, then remove the ones that don't apply.

ShadowLarkens:
this is meh

PR #9112, opened 2020-09-26
Removes single-line input on subtle and emotes, replaces with textbox

Andromeda-K22:
Two lines!


lbnesquik:
God yes

Novacat (changes requested):
Say.dm change should be ported to Polaris

Novacat (approved):

PR #9125, opened 2020-09-27
Adds Spacer Language

Andromeda-K22:
This is a port from Bay of a language myself and a handful of other people like. May as well?


lbnesquik:
Cheers! Does this get properly translated by synths or translator tools? Not familiar how languages work here.

Andromeda-K22:
It *should*, since I believe there's not a specific language list on those.

PR #9131, opened 2020-09-28
September 28 2020 Patch

Novacat:
Too sleepy to write patch notes right now


Novacat:
Not sure why the generate nanomaps is not working... Travis only threw an active edge so it should be fine merging this.

PR #9132, opened 2020-09-28
Change from iMessage style to Tinder style messaging

ShadowLarkens:
PDAs and Communicators now modeled after Tinder instead of iMessage. This simplifies a bunch of hacky code-- instead of having a Font Awesome `<Icon>` underlaying each message positioned to create a little "message tail," we now just adjust the border-radius so that the first message from a person has all but one corner rounded, and any subsequent messages without interruptions have both top and bottom corner on their respective side rounded. If this explanation is shit, just look at the screenshots. :P


lbnesquik:
Is that a TG port? Looks great either way!

Meghan-Rossi:
It looks like the gaps between messages and the early linebreaks are using up a lot of space, which means a lot more scrolling.

ShadowLarkens:
Yes, that's the downside of the new messaging style. (Not just in this PR, but the existing iMessage style does the same) Is why I included clipboard mode for people who don't like it.

And no, the underlying DM code is from Paradise, but I designed all of the templates myself

KasparoVy:
Thanks a ton for the clipboard mode
I dig the way a lot of modern AIs are heading but maaan so much spaaace

PR #9137, opened 2020-09-29
Fixes #9061 (partially) POI Stacking

### LINKED ISSUE #9061: POI stacking
POIs don't respect each other's no-no square

For POIs to _not_ overlap

#### What actually happened
POIs overlap

#### Steps to reproduce
- Exist
- POIs overlap

It seems to be most frequently the killer rabbit POI and the ore cliff POI. Pictures below.
poi1
poi2
poi3
poi4
poi5

aryll-linton:
Adds submap areas to killer_rabbit and orecliff POIs so they don't overlap other POIs on init
Unfortunately cannot generalize this so if other POIs are found stacking please inform me thanks

### DIFF
--- a/maps/submaps/surface_submaps/mountains/Cliff1.dmm
+++ b/maps/submaps/surface_submaps/mountains/Cliff1.dmm
@@ -1,13 +1,13 @@
 "a" = (/turf/template_noop,/area/template_noop)
-"b" = (/turf/simulated/mineral/floor/ignore_mapgen,/area/submap/ore_cliff)
-"c" = (/obj/item/weapon/ore,/turf/simulated/mineral/floor/ignore_mapgen,/area/submap/ore_cliff)
-"d" = (/obj/random/outcrop,/turf/simulated/mineral/floor/ignore_mapgen,/area/submap/ore_cliff)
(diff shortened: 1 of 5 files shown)


Novacat (dismissed):
Cliff1 is a Polaris POI, you will likely have to override it with Cliff1_vr.dmm, most of the Plains POIs have overrides like this.

aryll-linton:
I also see deadly_rabbit_vr.dmm doesn't have an #include- is that intentional, or do I have to make a mountains_vr.dm or other?
In the mean time, this will probably suffice.

KillianKirilenko:
Quad travis at it again I see.

Although if the cliffs are a Polaris POI, shouldn't the fix be upstreamed?

As for where the .dmm is included, I recommend using notepad++ or something with a 'find in files' feature. You can probably find where it's included that way, if it's not included in the .dme.

aryll-linton:
I miiiight have to send the cliffs upstream, but it feels like this is a problem that they would've fixed by now if it wasn't intentional. As for the #include, it's that mountains.dm has a whole list of of things that say something along the lines "make sure to #include all the POIs here or your milk will spoil" and deadly_rabbit_vr.dm doesn't have one

dwinters99:
I haven't done any work on this is too long, closing so I can come back to it later.

PR #9141, opened 2020-09-29
Fixes #9099

### LINKED ISSUE #9099: Wolftaur Armor Vest is offset incorrectly in-hand and on-equipment slot.

Wolftaur armor vest appears to the 'right' rather than the center in the item window. It works fine on-mob, though.

aryll-linton:
Added an icon named wolf_item to taursuits_wolf_vr.dmi that resolves #9099


Rykka-Stormheart (changes requested):
Give the icon _some_ differentiation from the regular taur duty vest, which is what you've just added. Something as simple as red plates on the 'pockets' would work for communicating it's an armored vest versus just another "backpack" (as right now if you have a taur duty vest AND this item next to each other it's impossible to tell visually which is which).

Thanks!

aryll-linton:
image
Like this?

Rykka-Stormheart:
> image
> Like this?

Looks perfect!

PR #9147, opened 2020-09-30
Makes off-duty have an economic_modifier equal to lowest in department

Runa-Dacino:
Presently, joining off-duty gives you the default economic_modifier. This, at lower "economic status" levels can cause people be unable to afford NIF software for scenes, or other purchased items. This doesn't make sense, and can be easily averted by joining as on-duty and instantly going off-duty.

This PR eliminates the need to join as on-duty and then instantly go off-duty.

Since there's no way to check what role a given character usually fulfils in a given department, and wanting to avoid causing a lower paid dept member to make more money while off-duty than on-duty, I picked the lowest paying job and used its economic modifier for the off-duty role.

### DIFF
--- a/code/game/jobs/job/exploration_vr.dm
+++ b/code/game/jobs/job/exploration_vr.dm
@@ -130,6 +130,7 @@ var/const/SAR 				=(1<<14)
 	outfit_type = /decl/hierarchy/outfit/job/assistant/explorer
 	job_description = "Off-duty crew has no responsibilities or authority and is just there to spend their well-deserved time off."
 	pto_type = PTO_EXPLORATION
+	economic_modifier = 6
 
 /datum/alt_title/offduty_exp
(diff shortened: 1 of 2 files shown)


code/game/jobs/job/exploration_vr.dm
```diff
@@ -130,6 +130,7 @@ var/const/SAR 				=(1<<14)
 	outfit_type = /decl/hierarchy/outfit/job/assistant/explorer
 	job_description = "Off-duty crew has no responsibilities or authority and is just there to spend their well-deserved time off."
 	pto_type = PTO_EXPLORATION
+	economic_modifier = 6
```
Novacat, line 133: This should be five, on account of pilot

Runa-Dacino: Ah, I missed the pilot. Lowering!

Issue #9102, opened 2020-09-24
Emergency box goggles get contaminated with phoron, nullifying their purpose


Runa-Dacino:
#### Brief description of the bug
Within the "Emergency Supply Box" non-explorers start with, the goggles with the "Just plain old goggles" description get contaminated if you only wear them with a mouth-mask, rather than wear a gasmask while in an atmosphere that has phoron. 

Wearing these goggles thus, even if the character is properly covered otherwise, causes toxin damage to accumulate.

#### What you expected to happen
The goggles protecting a character's eyes from phoron to be immune to contamination, like voidsuits, softsuits and gasmasks are.

#### What actually happened
Goggles got contaminated, toxin damage was suffered.

#### Steps to reproduce
- Put on science jumpsuit, wintercoat, winter boots, gloves, breath mask
- Wear emergency oxygen tank.
- Toggle winter coat hood.
- Find atmosphere that has phoron
- Stay there until clothing is contaminated
- Go back to O2/N2 atmosphere, take off all clothes except goggles.
- Watch toxin levels rise still.

#### Code Revision
- (Found using the "Show Server Revision" verb under the OOC tab.)
 Server revision: B:release D:
Commit: a4f6d03944deebfca428b07914b2ae5a531bbcd2 

#### Anything else you may wish to add:
- (Location if it's a mapping issue, screenshots, sprites, etc.)

VictorZisthus:
That's quite odd. Was this live or on a custom server?

PR #9160, opened 2020-10-03
Phroronguarded Work Attire

VictorZisthus:
Phoronguarded some things that made sense.

### DIFF
--- a/code/modules/clothing/glasses/glasses.dm
+++ b/code/modules/clothing/glasses/glasses.dm
@@ -140,6 +140,7 @@ BLIND     // can't see anything
 	toggleable = 1
 	action_button_name = "Toggle Goggles"
 	item_flags = AIRTIGHT
+	flags = PHORONGUARD
 
 /obj/item/clothing/glasses/science/New()
 	..()
@@ -151,6 +152,7 @@ BLIND     // can't see anything
 	icon_state = "plaingoggles"
 	item_state_slots = list(slot_r_hand_str = "glasses", slot_l_hand_str = "glasses")
 	item_flags = AIRTIGHT
+	flags = PHORONGUARD
 	body_parts_covered = EYES
 
 /obj/item/clothing/glasses/night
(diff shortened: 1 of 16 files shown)


VictorZisthus:
I'll have to do another detailed runthrough and add more things to the list that would make sense. Lot's of work to be done here. Will be part of a fix to issue #9102

KillianKirilenko:
These should probably either be marked as VR edits, put into _vr files, or pushed up to Polaris.

KasparoVy:
Wicked.

KillianKirilenko:
I don't see any changes to rigs here. *That said* that's a good point and something that actually needs fixing, but almost certainly ought to be pushed up to Polaris.

I believe the issue there is that rigs *have* `PHORONGUARD`, but they have it in their `item_flags` (where it doesn't work) and not their `flags` (where it ought to be).

KillianKirilenko:
Okay I was slightly off (helmet/chest are okay because they inherit it from spacesuits, gauntlets had it mis-set, and boots had it mis-set because they inherited it from magboots which were also mis-set, gg), but I've made the necessary "oops, someone made a silly mistake" fixes for Polaris [here](https://github.com/PolarisSS13/Polaris/pull/7718).

As for the rest of *this* PR; eh. Phoron clearly isn't *super corrosive*, but I do think a lot of 'resilient' clothing would still offer a lot of little nooks and crannies for particulate phoron to embed itself into and later pose a threat to people, so I'd be wary of making too much stuff resistant to contamination.

Besides, clothing contamination is barely a risk to begin with unless you're really careless or reckless.

KasparoVy:
Bless!

Novacat (changes requested):
This touches some polaris files. Should either be sent to polaris or put into _vr files.

Arokha:
Was created before current CI system. Kicking so current CI system runs against it.

Novacat:
Closing due to inactivity.

PR #9170, opened 2020-10-04
Removes freebie namechange in loadout + adjustments to make vr-only ones work properly

Heroman3003:
Removes free name/desc change as gear tweak from being a thing for literally every single item available. For now remains available only on plushie selection and locket accessory. Mostly to prevent devaluation of fluff items and potential abuse with being able to have a wide selection of items that have names completely changed.

Also changes way gear tweaks are initialized on _vr only items to match the non-vr ones post-this-change.

PR #9171, opened 2020-10-04
Standartizes gear_tweaks on _vr only items

Heroman3003:
The nonoffensive part of #9170. Actually makes several _vr items have said name/desc changes working properly, as well as standartizing them with Polaris' new way of adding tweaks to prevent any potential issues down the line.


Heroman3003:
Fluff label only because fluff items file had an old style gear_tweak definition for collars. No practical changes/additions to fluff items.

PR #9173, opened 2020-10-04
Adds Numbing Fangs trait

Meghan-Rossi:
Adds the numbing fangs from "Sharp Melee & Numbing Fangs" as a separate trait.  Cannot be taken with "Sharp Melee" or "Sharp Melee & Numbing Fangs", but you can still take the latter if you want both.  Positive trait, costs 1 point.  (Didn't need to explicitly add them as excluded traits of each other, as there's code to automatically detect that they modify the same variable.  Tested to confirm.)

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/traits_vr/positive.dm
+++ b/code/modules/mob/living/carbon/human/species/station/traits_vr/positive.dm
@@ -66,7 +66,7 @@
 	name = "Numbing Fangs"
 	desc = "Provides fangs that makes the person bit unable to feel their body or pain."
 	cost = 1
-	var_changes = list("unarmed_types" = list(/datum/unarmed_attack/stomp, /datum/unarmed_attack/kick, /datum/unarmed_attack/claws, /datum/unarmed_attack/bite/sharp/numbing))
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/human/species/station/traits_vr/positive.dm
```diff
@@ -62,6 +62,12 @@
 	cost = 2
 	var_changes = list("unarmed_types" = list(/datum/unarmed_attack/stomp, /datum/unarmed_attack/kick, /datum/unarmed_attack/claws, /datum/unarmed_attack/bite/sharp, /datum/unarmed_attack/bite/sharp/numbing))
 
+/datum/trait/fangs
+	name = "Numbing Fangs"
+	desc = "Provides fangs that makes the person bit unable to feel their body or pain."
+	cost = 1
+	var_changes = list("unarmed_types" = list(/datum/unarmed_attack/stomp, /datum/unarmed_attack/kick, /datum/unarmed_attack/claws, /datum/unarmed_attack/bite/sharp/numbing))
```
KillianKirilenko, line 69: `datum/unarmed_attack/claws` should probably be swapped for `/datum/unarmed_attack/punch` here, otherwise you've kinda-sorta given them both for a discount and only removed the non-injecting bite attack.

Meghan-Rossi: Yeah, my bad, thanks for catching that

PR #9178, opened 2020-10-05
Small item digestion fix/tweak

Verkister:
Fixes gurgle damages higher than the item's w_class value giving excess nutrition (like 6+6 damages would give 3 w_class worth of nutrition per cycle, totaling to be worth w_class 6 when taking 2 cycles) and also lets the digestion handle the entire item with one cycle if the damage is sufficient enough. A ciggy butt will no longer be worth 1,5 mining drills on max damage.


Verkister:
travis pls you double shit

PR #9146, opened 2020-09-30
[Map Edits] OSHA 2: Electric Boogaloo

VictorZisthus:
**[CODE]** Changed the density of the subspace broadcasters from 1 to 0. It's a flat disk that lays flush with the ground, there is no reason that one could not reach over or step on it if need be. In Virgo's case, this allows access to the Telecoms hub without having to deconstruct anything; If the Chief Engineer wants to set up a relay offworld or on a ship, he needs access to the hub.

**[CODE]** Undid #8530 's changes on the default tile, added a new tile subtype. Mapped in the new changes to S3.

**[CODE]** Reactivated Engineering Outpost Camera Network. Virgo can now use this network to quickly inspect power lines to and from the various outposts and solars without using EVA.

**[MULTI-Z]** Added many new security cameras to help improve AI gameplay and navigation. The only maintenance areas touched were emergency storage closets, the medbay-to-panic-shelter, and the exterior power wires; All other maint areas and private locations remain untouched.

**[MULTI-Z]** Intercoms, intercoms INTERCOMS, GOD I LOVE INTERCOMS! YOU GET AN INTERCOM, YOU GET-*bricked*
Added more intercoms to areas that make sense. Cargo, outside medbay hallway, IA, and Explo shuttle bay.

**[MULTI-Z]** Removed 4 mapspawn instances of dirt from the SME and RUST engines, as they transfer over to the Tesla/Singulo setups thanks to persistence, causing... Dirt in strange places.

**[MULTI-Z]** Added some fire alarms, Air alarms, and APC's to areas that very much need them so that, at the very least, people have an opportunity to know when rooms are depressurized before they ignore the emergency shutters and enter anyway.

**[MULTI-Z]** Added Emergency Shutters to Virology, Medical, and Mining operations in areas they were missing.

**[MULTI-Z]** Added more EVA stuff for Engineering and security. Security can now recover anything disposed of in asteroid Brig with a spacewalk. Let's give those lads at least some reason to use their airlock, yeah?

**[MULTI-Z]** All Tether ship SMES's begin fully charged to reduce engine draw and create a more consistant power grid.

**[MULTI-Z]** Upgraded some APC's to be labeled as Critical; Mostly airlocks and Heads of Staff rooms.

**[MULTI-Z]** Added a few extra ATM's.

**[Misc]** Added more cameras to the Cresent.

**[S1]** Spruced up the Tram room a bit with more atmos stuff and protective barriers.

**[S1]** Added a cargo camera and intercom to the trash pit.

[...]

PR #9194, opened 2020-10-07
Add Surgery and Resleeving Rooms to Robotics!

Rykka-Stormheart:
Adds two surgery rooms and a resleeving room to robotics to make things a little more private!

If people still want to be treated like nothing but machinery in the public view, though, it's not difficult to setup an optable in public still - but this allows people to privately scene or conduct longer, more detailed synth RP + resleeves without clogging up chat/blocking Robotics, and especially because Robotics used to be so visible from literally everywhere.

General overview screenshot here:

### DIFF
--- a/maps/tether/tether-03-surface3.dmm
+++ b/maps/tether/tether-03-surface3.dmm
@@ -153,8 +153,7 @@
 /obj/structure/window/reinforced/full,
 /obj/machinery/door/firedoor,
 /obj/structure/window/reinforced{
-	dir = 8;
-	health = 1e+006
+	dir = 8
 	},
 /turf/simulated/floor/plating,
 /area/tether/surfacebase/security/breakroom)
@@ -1072,9 +1071,9 @@
 /obj/structure/window/reinforced{
 	dir = 4
 	},
+/obj/structure/window/reinforced,
 /obj/structure/window/reinforced{
-	dir = 8;
-	health = 1e+006
+	dir = 8
 	},
 /turf/simulated/floor/tiled,
 /area/rnd/xenobiology/xenoflora_storage)
@@ -6137,10 +6136,7 @@
 /obj/structure/window/reinforced{
 	dir = 4
 	},
-/obj/structure/window/reinforced{
-	dir = 8;
-	health = 1e+006
-	},
+/obj/structure/window/reinforced,
 /obj/structure/window/reinforced,
 /obj/machinery/recharger,
 /obj/machinery/door/blast/regular{
@@ -6149,6 +6145,9 @@
 	id = "surfbriglockdown";
 	name = "Security Blast Doors";
 	opacity = 0
+	},
+/obj/structure/window/reinforced{
+	dir = 8
 	},
 /turf/simulated/floor/tiled,
 /area/tether/surfacebase/security/frontdesk)
@@ -6696,11 +6695,10 @@
 "alH" = (
 /obj/structure/grille,
 /obj/structure/window/reinforced/full,
-/obj/structure/window/reinforced{
-	dir = 8;
-	health = 1e+006
-	},
 /obj/machinery/door/firedoor/glass,
+/obj/structure/window/reinforced{
+	dir = 8
+	},
 /turf/simulated/floor/tiled,
 /area/rnd/outpost/xenobiology/outpost_hallway)
 "alI" = (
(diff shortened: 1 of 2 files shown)


Rykka-Stormheart:
...the active edges are pissing me off to the point I might go fix it.

VictorZisthus:
Can you PR this into #9146 or are you running on that? [I notice the exterior mech button]

Rykka-Stormheart:
> Can you PR this into #9146 or are you running on that? [I notice the exterior mech button]

I uh would need edit access to but I could try and figure out a merge or something, but I'd say let this one roll in then merge it and re-do your S3 changes?

dwinters99:
In full honesty it already takes a few stellar generations to get science to respond to requests so I don't know that removing the easy-access op-table is a good idea. This also makes it significantly harder for impromptu robot repairs wihtout having to ghetto repair. Furthermore this is propagating the Tether's hugeness and wasted space. Robotics was already a huge area that was used very little (except for RPing in after someone wanted to use the protolathe) so adding THREE ENTIRE NEW ROOMS, two of which will almost certainly never be used, is probably ill-advised. Like, when is the last time more than once synth needed repaired? When is the last time a synth needed resleeved? The only situation that I can imagine needing both ORs if if someone is taking deliberately long or if someone is teaching synth surgery. This also allows situations like this to occur with much more ease:
Screenshot_5

(pictured: 4 centcomm officers yelling at a scientist """teaching""" a roboticist how to do surgery without anesthetics or sterilizine. not pictured: the scientists limbless body after being taken to medical.)

There is also now dead-space at the top left, again agitating the TETHER2BIG issue. At most I think one OR  would be needed for the... synth repair scenes (??? I've never seen that happen but it lines up with how slow science is- to each their own.) 

On the bright side, it's great that robotics now gets two jumper-cables. I don't think they had a resleeving vendor either before, so that's good. Although looking at this now I see that the implanter is now behind a second door which means if a non-scientist wants a circuit implant they have to scour maint or waste a scientists time. Well, I tried to find the good parts. Sorry. I definitely think this needs another design philosophy pass.

Runa-Dacino:
> In full honesty it already takes a few stellar generations to get science to respond to requests so I don't know that removing the easy-access op-table is a good idea. This also makes it significantly harder for impromptu robot repairs without having to ghetto repair. Furthermore this is propagating the Tether's hugeness and wasted space. Robotics was already a huge area that was used very little (except for RPing in after someone wanted to use the protolathe) so adding THREE ENTIRE NEW ROOMS, two of which will almost certainly never be used, is probably ill-advised. Like, when is the last time more than once synth needed repaired? When is the last time a synth needed resleeved? The only situation that I can imagine needing both ORs if if someone is taking deliberately long or if someone is teaching synth surgery. This also allows situations like this to occur with much more ease:
> Screenshot_5
> 
> (pictured: 4 centcomm officers yelling at a scientist """teaching""" a roboticist how to do surgery without anesthetics or sterilizine. not pictured: the scientists limbless body after being taken to medical.)
> 
> There is also now dead-space at the top left, again agitating the TETHER2BIG issue. At most I think one OR would be needed for the... synth repair scenes (??? I've never seen that happen but it lines up with how slow science is- to each their own.) supplementing the existing op-table.
> 
> On the bright side, it's great that robotics now gets two jumper-cables. I don't think they had a resleeving vendor either before, so that's good. Although looking at this now I see that the implanter is now behind a second door which means if a non-scientist wants a circuit implant they have to scour maint or waste a scientists time. Well, I tried to find the good parts. Sorry. I definitely think this needs another design philosophy pass.
> 
> EDIT: Is it even possible to set up an op-table? I'm pretty sure that got removed after some frisky explo folk set one up on the shuttle.

The reason for two rooms is simple:

If there are 3 roboticists,

one of the roboticists is using one exam/repair room for a scene.

There is no need for them to be yelled at for using their job as a backdrop for scening, since the other roboticist can do their job without their coworker being in the way.

dwinters99:
> The reason for two rooms is simple:
>
> If there are 3 roboticists,
>
> one of the roboticists is using one exam/repair room for a scene.

Then surely you would only need 1 OR and the existing public table. Making the tether more complicated while creating more dead space is exactly what people have been trying to stop for the past year.

lm40:
yeah probably my biggest objection to that it is bloat to the map for largely edge-case reasons but it really only ends up coming down to one extra surgery room which isn't /that bad/.

nerdass:
We don't want organic surgery happening in robotics, why did you make this?

nerdass:
Oh wait i get it, they're just regular rooms that don't have anesthetic. I kinda like this idea but there needs to be something posted in there that says HEY DON'T DO STUFF TO ORGANICS YA' GIT.

Runa-Dacino:
> Oh wait i get it, they're just regular rooms that don't have anesthetic. I kinda like this idea but there needs to be something posted in there that says HEY DON'T DO STUFF TO ORGANICS YA' GIT.

For what it's worth, current robotics already has an anesthetic tank available.

Rykka-Stormheart:
Closing this PR as Soft will be rolling this into her Surface Security update. Branch will remain up until that PR is in and through, for safety and asset-keeping. o7

PR #9195, opened 2020-10-07
Adds Virgo-Erigone, Virgo 3b and Virgo 4 as home/citizenship options

Rykka-Stormheart:
None of the citizenship/etc stuff had stuff that was specific to our setting in it, all of them stating home systems like "Sol" which are by definition a good distance away to travel for work. They're left in so you can choose 'em, but I added Virgo as a home system option for those of us that decide we live in-system where we work.

### DIFF
--- a/code/modules/client/preferences_factions.dm
+++ b/code/modules/client/preferences_factions.dm
@@ -13,6 +13,7 @@ var/global/list/seen_religions = list()
 		seen_religions    |= M.client.prefs.religion*/
 	return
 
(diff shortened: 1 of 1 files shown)


Heroman3003:
Wasn't there "custom" option that allows you to input your own home system?

Rykka-Stormheart:
> Wasn't there "custom" option that allows you to input your own home system?

There is, yes, but I figured it'd be worth adding our 'official' ones to the list as well.

PR #9215, opened 2020-10-14
Creates proper names and descriptions, sets density, to sprites found in amayastuff.dmi

AmayaWah:


vorestation.dme
```diff
@@ -3793,6 +3794,6 @@
 #include "maps\submaps\surface_submaps\plains\plains_areas.dm"
 #include "maps\submaps\surface_submaps\wilderness\wilderness.dm"
 #include "maps\submaps\surface_submaps\wilderness\wilderness_areas.dm"
-#include "maps\tether\tether.dm"
+#include "maps\virgo_minitest\virgo_minitest.dm"
```
Meghan-Rossi, line 3797: I don't think this part should be here

PR #9243, opened 2020-10-18
Adds the ability to hide, or reveal one's own wings with a verb

Runa-Dacino:
Adds a verb under "IC" tab allowing players to show off or hide their wings. Idea is that some wings can be folded up and hidden under clothes, be they a jumpsuit or a labcoat or something. 

For simplicity's sake, and for player customization (but mostly simplicity), I just made it a verb any player can use regardless of what clothing they wear. Could even do it naked, if so desired.

I cannot really see a way this could be abused beyond flying without visible wings, but that seems like such a niche thing, I don't see much of a need to fear it'd get abused like that.

Does not persist through rounds, nor going tramming out and returning. Defaults to visible wings.

### DIFF
--- a/code/modules/mob/living/carbon/human/emote_vr.dm
+++ b/code/modules/mob/living/carbon/human/emote_vr.dm
@@ -247,3 +247,10 @@
 	set desc = "Switch tail layer on top."
 	tail_alt = !tail_alt
 	update_tail_showing()
+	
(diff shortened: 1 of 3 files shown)


dwinters99:
This is really good! I was considering adding functionality similar to this for ears for hiding horns or what-have you. Anyhow, well done.

PR #9244, opened 2020-10-18
Update tether_plains.dmm

KillianKirilenko:
Attempts to resolve some of the wilderness plains active edge issues by buffering the north wall with a thin layer of rock, as the other walls, the NW/NE corners, and the transit shed *aren't* triggering active edge warnings.

Basically just a bandaid fix that has been promising in the limited local unit tests I've run (which is actually really easy to do). Doesn't help with active edge warnings popping up from the debris field but it might help with everything else.

Obviously the actual POIs themselves need fixing and cleaning up too, but that's a much bigger project.

Knowing travis it'll still fire active edge warnings on the north wall with this to spite me, but hey, worth a shot?

### DIFF
--- a/maps/tether/submaps/tether_plains.dmm
+++ b/maps/tether/submaps/tether_plains.dmm
@@ -37,7 +37,6 @@
 	dir = 1;
 	id = "wilderness";
 	name = "Anti-Fauna shutters";
-	pixel_x = 0;
 	pixel_y = -25
 	},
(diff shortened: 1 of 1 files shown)


TheFurryFeline:
Just to note, it's been said that a VR version would need to be created and then Vorestation edits made to the plains.dm or whatever it is that lists the POIs to spawn.

Verkister:
The whole Tether is the Voestation edit here.

PR #9250, opened 2020-10-23
Add Lamia Mobs (sprites by Prof. Smokey!)

ShadowLarkens:
Adds lamia simple mobs! They're pretty nommy, and they have two stomachs, each one able to independently render from the other. 

Notes:
 - These could use proper emote_lists, but I don't really feel like writing them, so if anyone wants to make a future improvement...
 - The sprite offsets aren't quite right in the dmi file for the east and west facing sprites of any lamia. They're a little off their turf. I really, really, *really* don't feel like going through all 161 sprites to fix this, so like.. don't yell at me because they're 4 pixels into a different turf on their weast directions.

Mechanically, the second "tail stomach" is only accessible via:
 - Player control
 - Feeding yourself to an NPC lamia
 - Squirming in their upper stomach (50% transfer chance, 25% escape chance) until you get transferred to their tail.

Once you're in their tail, you have a 25% chance to trigger digestion, and a 50% chance to go back up to their upper stomach. There is no directly escaping from their tail, you must get back to the upper stomach, and then try to pass a 25% escape chance there.

They have the same passive AI as catgirls, and do similar amounts of small damage; however, they're 3x more durable than them (60 health). They also have a 50% chance to immediately eat you if you bump into them (hey, that's what big tails are for), and a 65% chance to pounce you.


Haery70:
As the artist behind the sprites I will say that this is very good. As for the code you didn't really need to give the mobs specific descriptions for each of their icons. Like the catgirls, you can always make a single line that describes what they are in general. Other than that, you did a good job with it.

Screemonster:
yes good

PR #9254, opened 2020-10-27
Rabbits don't bumpnom anymore

Cyantime:
I guess the rabbits shouldn't be doing this without their pat gimmick satisfied. No changes to the killer rabbit, though it doesn't really matter for it.


lm40:
Good imo, they will stop eating each other and will no longer need to be basically locked in isolation to stop botanists from tripping over them every five minutes. If you are dying to be eaten by a mob for whatever reason  you can still spam pet them for it.

Arokha:
Seeing if I can kick off the new CI runs by close-opening.

PR #9263, opened 2020-10-30
[MIRROR] Adds AlliCo Grab-A-Gift

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/7454
--------------------
A claw machine just as hellish to play as the real ones. Uses TGUI & was good practice for it. Should be excluded from the mapped random arcade spawns, only gives stuffed animals, and is unique in that it demands money. Not mapping it in just yet. Tested it every way I could think of.

image

Basically the directional buttons are pointless clickspam, each win adds to the probability of winning a prize........ which starts at 0.

Once a prize is won, it's reset back to 0.

Also if you emag it the machine will steal all of your account funds and reward you with a primed frag grenade to the face.


Shadow-Quill:
Needs a TGUI bundle rebuild.

PR #9267, opened 2020-10-30
Adds Cabin mining pods to the maint loot pile

Detective-Google:
Adds little pop-out cabins to the maint pool; around the same rarity of sleevemates, body snatchers, et cetera.
image
intended to allow people to enjoy the comfy snow setting outside the station and get a bit of privacy at the same time.


Detective-Google:
Given that travis didn't fail due to active turfs the second time I'm assuming one of the ruin spawns has active turfs, someone probably ought to look into.

PR #9275, opened 2020-11-03
Fix toeless knee and thigh high jackboots

Little-119:
Changes names of the sprites for toeless knee and high jackboots in feet.dmi, fixing how they were invisible on the mob sprite

Should probably be sent to Polaris but I don't even know if I'm supposed to be the one to do that


KasparoVy:
`Should probably be sent to Polaris but I don't even know if I'm supposed to be the one to do that`

Can do w/e you want! If you create at Polaris make sure to ref this PR in your description though so when it trickles down cuz syncbot it's easy to identify that it got merged here before

PR #9278, opened 2020-11-05
Removes flashlight types from trash eater list

Verkister:
Abused to hell and back by the explorers smh


dwinters99:
I know what this is going to sound like but PLEASE don't say "the explorers" because I've never seen anyone do this and I don't want more ammo for the Space Lobby Warriors(tm) to say "explo only powergames xd". 100% chance this was unrobust deadshift explo. That said, if you are going to do this, at least do it with un-eaten liquidfood because I see newbie Talon/explo do that a lot and it's a big yikes.

Kegsey:
I've not a clue how this could possibly be abused; especially when explorers' can just get gravitons from Science completely removing the need for flashlights in general.

What I do know however is that I've constantly had scenes where eating flash-lights has been a cool niche thing for the pred to do in order to give the prey some light inside a gut... I'm not saying that it'll really affect scenes in the long run, just that it's a bit of a silly and knee-jerk "nerf". Perhaps instead of punishing everyone for the actions of a few, you should punish those who abuse it instead...

Upon further thought I can potentially see how, carrying 50 flashlights in a gut so you never run out... but again, punishing EVERYONE for what likely only one or two people have done isn't too gucci IMO.

TheToaster98:
By the same rule of thumb I hope we are removing mob vore as it allows explorers to carry seven thousand carp and spiders on their pocket bellies.
Or, and hear me out, I know is a strange concept but we could punish those who abuse a mechanic instead of removing it.

ShudderEve:
idk who you mean by "the explorers" because i don't think literally anyone has come across this and it's a very dangerous thing to say due to all the explo hate circlejerks

also you know eating flashlights is horrible even if it's an exploit, helmet lights are 2x as powerful and are more available

Verkister:
Yeah, I should've just taught them better seeing tactics than whatever galaxy brain stuff their team's been doing for months.

PR #9279, opened 2020-11-05
Revert "Fix unintended inbelly item interactions caused by copypaste"

ItsTheKet:
This is mostly made to attempt to revert a previous change Merged previously which disallowed the interaction of items under certain conditions. Some discussion is being made to, rather than limit it to the messy method it is to interact with a person, to expand on it and making it something anyone can really use.

However, an assumption comes that the code would need to be reverted anyways in case it may hinge on the previous change.

It is a simple change of could so it doubtfully will cause any issues.

**Edit:** Cleaned up my own rambling and tossed it off into comments somewhat.

### DIFF
--- a/code/_onclick/click.dm
+++ b/code/_onclick/click.dm
@@ -114,11 +114,12 @@
 		trigger_aiming(TARGET_CAN_CLICK)
 		return 1
 
-	// VOREStation Addition Start: inbelly interaction
+	// VOREStation Addition Start: inbelly item interaction
 	if(isbelly(loc) && (loc == A.loc))
 		if(W)
-			to_chat(src, "The firm confines prevent that kind of dexterity!")	//Only hand-based interactions in bellies
-			return
+			var/resolved = W.resolve_attackby(A,src)
+			if(!resolved && A && W)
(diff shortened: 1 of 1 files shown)


ItsTheKet:
I'll move my own personal opinions down here so it doesn't get in the way of the PR, however it hasn't exactly changed from before.

While the method of interacting with oneself is a rather "Hacky" method, one I am personally not a fan of. It still is an interaction I am a fan of, so I won't deny having some bias towards it. A litney of interactions have become substancially limited given it's patch, and unfortunately it's one enough to warrant getting an account and working a bit with some of the other devs to make it not only this PR, but to of course expand on the concept.

The details are still just mostly a conceptual WIP as far as I am aware but I do genuinely wish to make this a reality.
Cause maybe I wish to eat pizza in peace.

Heroman3003:
While I'm fine with the idea of the item interactions coming back, this only brings all issues involving that 'feature' back, particularly how nothing in the game indicates its possible, that it should be possible, and that only way to access it is byond macros, which you yourself admitted you wanted to keep hidden from people.

nerdass:
Please share your macros.

PR #9282, opened 2020-11-06
Makes the Chapel fancier

Very-Soft:
I saw that back window and was like, man, it would be cool if that was more of a main feature, and there be more space in the chaplain's office. Wanna make it comfier. So!

chapel time

For a note, I renamed the mouse that starts with Noodle. Being as we can't put an actual apple inside of there, I thought it only fitting to name the mouse thusly~

Here is a screenshot!

Z3 was also edited to add one chapel sign on the stairway, and to separate botany's bunnies, so they aren't as likely to bumpnom eachother at roundstart.

Also renamed the maintenence lizard to Bucket, since someone apparently named it that and it didn't already have a name. Bucket is now the secret head of Janitoria.


Very-Soft:
Hmm I need to fiddle with map merge I think. >:U

dwinters99:
This is good, but most importantly, it's smaller. 100 points to anything that makes the tether smaller. Well done.

Very-Soft:
I think the chapel itself is technically slightly larger, but just pushes into some otherwise unused maintenance space. I just kind of got inspired for how it the space might be better utilized to be more pleasant to play and hang in, so, just whipped it up real fast.

SweetBlueSylveon:
I enjoy this.

dwinters99:
Your head is in the right place. I hope this is the first of many good map changes.

KasparoVy:
Bless up.

TeleTubby2:
Fixes the issue with a tinted office being visible from maint as well. good change.

PR #9285, opened 2020-11-07
Fixes mob spawners respawning mobs when seen

Verkister:
Makes mob spawners not magically pop more mobs out of nowhere when there is a non-ghost player in view.


Muffindrake:
That was fast. Thanks!

Cyantime:
These spawners are mostly supposed to be physical, smashable objects that the mobs come out of, compared to the "tether spawners" which are invisible waypoints that make mobs appear when you aren't looking. These ones spawning when watched is intended, which spawner was causing problems by spawning in view?

PR #9296, opened 2020-11-10
Gives NIFs soulcatchers by default

Mouseofthecake:
hm, there's usually some wacky PR text here

anyway, this adds soulcatchers to everyone's NIFs by default, much like commlinks. 

done on request, enjoy.

### DIFF
--- a/code/modules/nifsoft/nif.dm
+++ b/code/modules/nifsoft/nif.dm
@@ -87,14 +87,12 @@ You can also set the stat of a NIF to NIF_TEMPFAIL without any issues to disable
 				qdel(src)
 			return FALSE
 		else
-			//Free commlink for return customers
+			//Free commlink and soulcatcher for return customers
 			new /datum/nifsoft/commlink(src)
+			new /datum/nifsoft/soulcatcher(src)
 
 	//Free civilian AR included
 	new /datum/nifsoft/ar_civ(src)
-	
-	//Free soulcatcher because why not
-	new /datum/nifsoft/soulcatcher(src)
 
 	//If given wear (like when spawned) then done
 	if(wear)
--- a/code/modules/nifsoft/nif.dm
+++ b/code/modules/nifsoft/nif.dm
@@ -92,6 +92,9 @@ You can also set the stat of a NIF to NIF_TEMPFAIL without any issues to disable
 
 	//Free civilian AR included
 	new /datum/nifsoft/ar_civ(src)
+	
+	//Free soulcatcher because why not
+	new /datum/nifsoft/soulcatcher(src)
 
 	//If given wear (like when spawned) then done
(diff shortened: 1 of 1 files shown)


Very-Soft:
I donno about this. Soulcatcher actually does some mechanical stuff that I'm not sure we want people to be spawning with, such as the ability to back yourself up if you die.

Shadow-Quill:
Not commenting on whether or not this PR should go through, but you've got runtimes:
https://travis-ci.com/github/VOREStation/VOREStation/jobs/434798013#L584

Mouseofthecake:
Oog, sorry for taking so long to update. This commit SHOULD fix it by making the soulcatcher enable itself just like commlinks do; one round after the person's had their NIF installed. 

This should also fix nsay/nme. I can't test that stuff properly locally, so a testmerge would be neato.

dwinters99:
iirc soulcatcher backups don't ping transcore and implants/sleevemates are so ubiquitous as to be only non-default by formality so I wouldn't consider soulcatcher backups being better than just going to the inevitable 400 implants that have been laid out- not to mention, at the end of the day a soulcatcher is a scene tool and getting yourself backed up so you can scene without worry of forgetting an implanter really doesn't seem like something to be worried about balancing- i mean, unless antags become intrinsic to the station (unlikely) any form of back-up is just for sceneing because of how infrequently irrecoverable deaths are (doubly reinforced by the fact that people who scene and forget a backup are just an ahelp away from a respawn anyhow)

Novacat:
I am also a bit skittish, if only because soulcatchers pretty much cost nothing already, but allows people who want nothing to do with soulcatcher stuff by not installing the program in the first place. I would rather soulcatcher mechanics remain as is by being opt in rather than opt out.

Verkister:
I feel like the amount of potential harm severely outweighs the benefit of this pr. Soulcatchers are installed for specific uncommon type of scenes, and if you're not planning to have that kind of scene, having one forcibly installed and activated without consent is a recipe for a disaster if your scene wasn't supposed to involve soulcatching.

lm40:
Ehh not a huge fan of this mostly because it is feature creep for an absolutely marginal inconvenience. Doesn't soulcatcher only cost like 100 thaler? I don't really mind the idea of people having free backups to be honest but I do kinda mind doing it through the NIF.

lbnesquik:
Fine with it, but PLEASE make it turned off by default. I don't want to have to deal with an unexpected thing if i don't go and manually disable everything. 
The Best way would be by a preferences thing, but that's a lot harder. So maybe leave them by default or just force the variable off.

mel-byond:
I often download a Soulcatcher only if I need it, I am generally against implementing it since it can mess with scenes for people who aren't into soutcatchers.

PR #7757, opened 2020-05-10
New sprite for the botany + xenobiology scanner, and BUTTS.

Rykka-Stormheart:
Butt ~~posters~~ photocopies, Chronos suit phase in/out/trapped states, new botany/xenobiology scanner sprites, and of course, ASS.

PR #9301, opened 2020-11-14
Adding New Foods

Jwguy:
Adds the Following Foods:
Gigapuddi
Happy Gigapuddi
Angry Gigapuddi
Omurice
Heart Omurice
Face Omurice
Suppermatter
Suppermatter Slice
Exciting Suppermatter
Exciting Suppermatter Slice
Cinnamon Bun

Also adds the following Festive Foods:
Buche de Noel
Turkey

Comment:
Not sure why, but I attempted to have Buche De Noel and Turkey as Oven recipes, but it was producing baked items rather than the intended recipe. It all tests out fine locally, as is, though.


Novacat (changes requested):
These should either be moved to _vr files, or ported to Polaris. Porting to Polaris would probably be better.

Jwguy:
> These should either be moved to _vr files, or ported to Polaris. Porting to Polaris would probably be better.

I'll definitely move it over that way, then. For the record, our **Recipes_Microwave.dm** isn't even with Polaris, as it has 10 additional entries and close to a hundred more lines in it, so I was under the impression we were writing directly to it.

Just to note them, these additional items begin at line 1316, starting with the Sushi entry. Polaris stops at 1314 with the conclusion of the Porkbowl entry.

Jwguy:
[Porting Food Additions #7757](https://github.com/PolarisSS13/Polaris/pull/7757)

Not sure if I need to link it, but it's right over there.

PR #9315, opened 2020-11-17
Expands Surfmed among other things

Very-Soft:
I'm having some trouble bear with me. = A =

Expands Surface Medical to make it into primary medical, nerfs Space Medical, making it look a bit like things are being 'moved out' in continuing with the plan

z3
z2
z1

Also adds an ignition button to the crematorium
Also fixxes some pipe things I missed after the chapel rework
Also converts the surface cargo area into a Cargo Shop
Also adds Detective Wardrobe to detective's office
Also makes the tiny shuttle public access
Moves the reading rooms to Z2, and adds a new section to connect the area, including a new cafeteria, and a new airlock

Need to wait for github to stop derping to fix the commits. Will add more images later, me sleep now.

### DIFF
--- a/maps/tether/tether-01-surface1.dmm
+++ b/maps/tether/tether-01-surface1.dmm
@@ -74,59 +74,203 @@
 /area/tether/surfacebase/mining_main/external)
 "aan" = (
 /turf/simulated/wall,
-/area/tether/surfacebase/cargo/mining/airlock)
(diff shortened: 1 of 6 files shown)


Very-Soft:
I think I know what to do but Github seems to be having issues, so I'll just have to fix it in a little

PR #9333, opened 2020-11-21
Adds Nikki's Fluff Outfit For Nikki, + Rigsuit-Mounted Size Guns For Research

Tenryumi:
The dork's outfit is here!
[My app for Nikki's fluff items have been accepted](https://forum.vore-station.net/viewtopic.php?f=26&t=1824&p=12018#p12018), and I've been taking my sweet time coding its features as detailed in the forum thread.
With Dragor's permission, I've also added the other components of Nikki's getup: The outfit and boots. These two items don't do anything on their own, but the hat and necklace... well, they do.

Aseprite_2020-11-16_14-03-19

**TL;DR:**

- The hat is an ordinary hat. It's really big and can't be stored because it's **too goddamn huge.** It also can have a translocator slipped inside of it to become a wearable translocator, at which point it is no longer an ordinary hat.

- The cloak deploys from a necklace, which is like the world's most niche rigsuit, so much so that it barely even qualifies as one. It has no armor, gives no space protection, and literally the only thing that can be installed into it is a snazzy new item that Robotics will be able to fabricate for anyone to use: A rigsuit-mounted size gun! What on Earth would you need these on a hardsuit for?! I don't freakin' know! But it's an option now. If Research has enough research levels to make a normal size gun, Robotics can fabricate these for the same material costs.

To make this clear, Nikki does not start with the translocator or the size gun that make these items **do the fun thing.** On their own, they're just ordinary clothes.


Novacat (changes requested):
This PR touches a polaris icon file, that should probably be moved to a _vr icon file.

Tenryumi:
> 
> 
> This PR touches a polaris icon file, that should probably be moved to a _vr icon file.

Oh shoot, I thought I'd gotten to undoing them all. One sec.

KasparoVy:
That's a really cool concept, it's great to see how creative you got with this stuff!

Novacat (changes requested):
This needs to be rebased (or have a minor change) so that the new CI check will run on it.

Novacat (approved):

PR #9336, opened 2020-11-22
Fixes Proteans not retaining languages in blob form, gives Proteans Feeder ability

Tenryumi:
See: Title.

Before, Proteans didn't have their languages inherited from their original body. Now they do.

Also gives them the Feeder ability because that's FUN.


Heroman3003:
nanomachines feeding? What are they gonna do, increase your iron contents?

Tenryumi:
> 
> 
> nanomachines feeding? What are they gonna do, increase your iron contents?

Nope, but they will increase your nutrition. It's the same logic applied to people being able to digest synths, which is:
"reasons"

Issue #6122, opened 2019-10-27
Soft's plans for the Tether


Very-Soft:
So there's been a lot of talk about getting new maps and things like that, and I think that that will be great, particularly and especially if we get map rotation. But a lot of this talk has centered around not just doing things differently, but also making sure we don't repeat the mistakes the tether has made. 

I greatly support these talks, but in the short term, we still have a map to play on now that we are having more and more talks about moving away from. Now. I think the Tether is a pretty good idea, it just stumbles in a few key ways, and that those things can be improved with a little bit of effort.

Okay so I have a lot of ideas of how to make the map better to play SS13 the game on in general, that should hopefully facilitate a better overall experience for us in a variety of ways. And from the top I'd like to say that these are all things I plan to do myself, and in some parts I would really prefer to do myself. I have plans for MUCH of the station, but would prefer to do it one step at a time, so that we can critique the changes as they come along. 

I WANT to hear if you have a problem with something in the map presently or with my ideas for changes, and more than that, I WANT to hear how you think it should be different to support a better play experience with regards to the map.

I've already started working on this plan, but I figured it's probably not the most obvious what the intent is from just looking at what I'm doing. Explaining is in order.

So, a quick and dirty overview before I get around to explanations:

- ~~Move the bridge to Z3 atrium, copy paste  minibridge into space. Put office, visitor, and maint stuff in the free space.~~ (#6149)
~~- Use the space from the minibridge to expand surface security to full functionality.`~~ (#9338)
~~- move mining to space~~
~~- use the open space from mining to expand surface medical to full functionality.~~(#9320)
~~- move the holodeck near the public garden entrance on Z1.~~ (#6198)
- move cargo to Z5
- move exploration to Z5
- move EVA to Z5
~~- move public teleporter to z3~~(#9338)
- find a place to move the AI
- delete Z6 and Z7.
~~- review bar seating area size (Bar is off limits, reserved by Demi)~~

So that said, you're probably wondering why go through all this trouble. Quite simply, the number one most complained about thing on the tether. A large portion of the point of the map's basic concept. The elevator. I do not want to remove the elevator, and I don't think adding MOAR elevators will make them really much less annoying. But making it so that for your general playing, dealing with the elevator is not necessary often.

[...]

PR #9338, opened 2020-11-22
Expands Surface Security among other things

Very-Soft:
In continuing with the plan - https://github.com/VOREStation/VOREStation/issues/6122 :

Expands Surface Security
Nerfs Space Security
Moves Public Teleporter to Z3
Implements Rykka's Robotics changes ( https://github.com/VOREStation/VOREStation/pull/9194 )

z3
z2
z1


KasparoVy:
I guess this means great cargonia will be landside and mining will once again have to launch from the planet to the stars?

(Totally fine with that! Just wanted to confirm future plans)

Very-Soft:
Negative, Cargonia is a job that does revolve around a shuttle, it will remain in space (though it does have a cargo shop on the surface now)

MrSign:
Noicely done, glad to see you're continuing to work on it! Greatly approve of the changes!

Andromeda-K22:
I think the three level security setup is a little bit confusing. The current brig is a bit of a sprawl, but it is clearly deliniated into A. the functions security, and B. perma/cells. 

You've kinda spread the cells out all over the place. Two on the first level, and then the rest of the brig on the third level. (from the top down) I'd suggest putting all the cells together instead of having them spread out, or combine those two cells into an easily-accessible 'drunk tank'.

Very-Soft:
> I think the three level security setup is a little bit confusing. The current brig is a bit of a sprawl, but it is clearly deliniated into A. the functions security, and B. perma/cells.
> 
> You've kinda spread the cells out all over the place. Two on the first level, and then the rest of the brig on the third level. (from the top down) I'd suggest putting all the cells together instead of having them spread out, or combine those two cells into an easily-accessible 'drunk tank'.

My idea for this was simply to allow for more proper brig functionality to the idea of the drunk tank without putting it into that stigma, while also retaining the closeness to where security folk are likely to be hanging out. 

You have 2 cells near the front, right next to Processing and IAA and in view of the hallway, intended primarily for short sentences. It's easy to access, relatively easy to break into and out of with the right tools/help. These should be easy for security to use for RP or for basic low priority every day crimes without having to go all the way down to perma. 

Where the 2 cells in perma are far away and a little harder to escape without help. These are meant for high security prisoners, further away and with a little more pomp to them, you're probably only going to go here if you really fucked up.

Basically the way the space brig is, is way too big for what we handle in addition to being so far away from everything that if you get locked up you won't see or interact with anyone until your sentence is up. It's far away and requires help to get out of.

The existing drunk tank is mostly okay on the other end, but there's only one timer and no way to lock up anything sensitive the person might be carrying, so. I thought it might be better if there were 2 small holding cells closer to the front desk, to make things easier on sec and less boring for small time criminals.

PR #9341, opened 2020-11-23
ML3M Rebalancing

Wickedtemp:
Originally, the ML3M was intended to be a little bit stronger. The healing values for the Brute, Burn, Omni, Toxin, and Antirad cells, have been adjusted according to what was initially planned. 
Base Tier's now heal 10 per shot, for a total of 40 damage healed, up from 20.
Second Tier's now heal 20, instead of 10.
Third Tier's heal 40, instead of 20. 

Omni 1 heals 5's for brute, burn, and toxin, and 30 for oxy, up from half those values. Omni 2, is 10's and 60, and Omni 3 is 20's and 120. 

Antirad Cell had its toxin heal bumped up from 2.5 to 5, and its radiation-heal from 150, to 350. I wasn't originally going to touch this one, but the text said "It's 150 because that's equal to five units of arithrazine", and... it wasn't. So, I bumped it up to actually be equal to 5u of Arithrazine's worth of radiation healing.

Stabilizer Cell, CorpseMend, Resistance, Haste, and the size-changing cells, were not touched.
The reasons for this are as follows:

- The healing laser was simply not strong enough to ever tend to more than one patient. Healing 20-40 damage per cell just isn't enough, even if you had several clips full of cells, it just wasn't viable.
- As a result of this, it was seldom if ever used. Field Medics rarely if ever took it with them. In nearly every round I've played, it was left exactly where it was spawned, collecting dust. 
- I'd like to make it a more proper treatment option. It wouldn't be equal to reagents used in your usual Medbay setting, or surgical repair, but if it's modified to actually fulfill its niche as a rapid, contact-free treatment option, it could be relied on more by Field Medics. There is so much potential for this device as an "in the field" healing tool, and I'd like to see it done well. We have no shortage of people who would like to see non-chemical-based treatment options. We might as well make it useful enough to actually use.

Here are a few issues people brought up in Discord when I floated this idea.
"But, Tempest, you're doubling everything? That's a huge change! Won't this result in the ML3M overshadowing other treatment options?"
Certainly not. Currently, with tier-2 cells that only heal up to 40 damage, this is actually worse than basic first-aid in terms of raw healing ability, as the cutoff for first aid healing is 50 damage on a limb. First aid and reagents will still be the most convenient, most readily available, and overall most-used manner of treatment for injuries that can be treated using those methods. All this change would do is allow the ML3M to also have its uses, and stop being overshadowed by everything else. 

"Okay, but what if it DOES actually become The New Meta and becomes the favored tool for healing?"
I wouldn't see it as an issue if Chemistry gets dethroned, in all honesty. They'd both take quite a bit of prep-work to use. You have to make multiples of every chem if you want a solid stock, and you'd need multiples of every cell if you want proper coverage and healing-ability, since even with these changes you're expending 1-3 cells per patient. And, unless you get multiple ML3M's made, only folks who are gonna have one are the CMO, and whoever nabs the spawned one first. 

"But wouldn't this result in less departmental cooperation between Medical and Science now that Medical will require fewer cells?"
Most likely, no. If anything, I'd think there would be an increase, round-to-round, if this becomes a more viable and wanted tool, more people will ask for cells. Personally, my list of requested cells would stay the same, but I'd be sending that request in more often since I know it'll actually be useful rather than having it just to have it. Tier 2 cells are still going to be desirable, and I'm still probably not going to order Tier 3's because of the cost, but that's just me. It's still going to be incredibly weak at roundstart, and better cells will be that much nicer to obtain.

"But Tempest! If you get the highest tiers of cells, and a lot of them, you could fix anything that doesn't require surgery!"
Well, if Science had the time and mats, then sure, you could gather up a large enough collection of cells, and you probably could indeed do this for one or two badly injured patients... ... ...and I could accomplish the same goal if I'm given 20 seconds in a chem-lab. The ML3M is more costly to use, the cells cost resources, have to be recharged after 4 shots, and you only start with a base tier Burn, Brute, and Stabilizer cell. Each shot from a base level burn/brute cell is the equivalent of 1.25u of bicaridine or kelotane. This change would make it 2.5. Base level cells are still going to be next to useless for anything other than the most minor of injuries. Chemistry will still be the dominating powerhouse it always has been, but at least you can use this tool, provided the PR goes through, instead, more often.

"Tempest, this thing just wasn't intended to be used to fully heal people..."
A couple people have said this, and to be honest, it's silly. In our medical system, there's very little distinction between what can "heal a little bit" and what can "heal fully", because it all comes down to the fact that everything works off of damage values. If Option A heals 5 damage, and Option B heals 50... Both are capable of "healing fully" if you have enough Option A, it's just a question as to whether or not that's practical. Currently, the ML3M's healing values can't really even be used to even partially heal injuries, unless you have a large number of cells. You'd expend all of your charges on one patient and then you've got a paperweight taking up storage space until you can get to a recharger. 

Now, as for issues that I personally see? Biggest one is probably the Toxin cell. If I remlember correctly, the base Toxin cell is the same tech requirement as the Second Tier Brute/Burn cells. This PR doesn't fix that, and I don't really see it as a huge priority since I don't think anyone uses the toxin cells as it is, but yeah. Might need to be fixed at some point if this PR goes through.

And, absolute worst case scenario, let's say that this PR screws everything up, turns the Tether upside down, and now Medbay is dominated by a bunch of blue laser beams flying everywhere like it's the new Star Wars Trilogy (but actually better because those movies kinda set a low bar)... this PR was really easy to do. The only long part about it was typing all of this up. We can just bump the numbers down if this ends up sucking really bad, ezpz.

Thank you for coming to my ted talk.

### DIFF
--- a/code/modules/projectiles/guns/energy/cell_loaded_vr/ml3m_cells.dm
+++ b/code/modules/projectiles/guns/energy/cell_loaded_vr/ml3m_cells.dm
@@ -31,7 +31,7 @@
 
 /obj/item/projectile/beam/medical_cell/brute/on_hit(var/mob/living/carbon/human/target)
 	if(istype(target, /mob/living/carbon/human))
-		target.adjustBruteLoss(-5)
+		target.adjustBruteLoss(-10)
 	else
 		return 1
 
@@ -43,7 +43,7 @@
 
 /obj/item/projectile/beam/medical_cell/burn/on_hit(var/mob/living/carbon/human/target)
 	if(istype(target, /mob/living/carbon/human))
-		target.adjustFireLoss(-5)
+		target.adjustFireLoss(-10)
 	else
 		return 1
 
@@ -83,7 +83,7 @@
 
 /obj/item/projectile/beam/medical_cell/toxin/on_hit(var/mob/living/carbon/human/target)
 	if(istype(target, /mob/living/carbon/human))
-		target.adjustToxLoss(-5)
+		target.adjustToxLoss(-10)
 	else
 		return 1
 
@@ -95,10 +95,10 @@
 
 /obj/item/projectile/beam/medical_cell/omni/on_hit(var/mob/living/carbon/human/target)
 	if(istype(target, /mob/living/carbon/human))
-		target.adjustBruteLoss(-2.5)
-		target.adjustFireLoss(-2.5)
-		target.adjustToxLoss(-2.5)
-		target.adjustOxyLoss(-10)
+		target.adjustBruteLoss(-5)
+		target.adjustFireLoss(-5)
+		target.adjustToxLoss(-5)
+		target.adjustOxyLoss(-20)
 	else
 		return 1
 
@@ -110,8 +110,8 @@
 
 /obj/item/projectile/beam/medical_cell/antirad/on_hit(var/mob/living/carbon/human/target)
 	if(istype(target, /mob/living/carbon/human))
-		target.adjustToxLoss(-2.5)
-		target.radiation = max(target.radiation - 150, 0) //same as 5 units of arithrazine, sans the brute damage
+		target.adjustToxLoss(-5)
+		target.radiation = max(target.radiation - 350, 0) //same as 5 units of arithrazine, sans the brute damage
 	else
 		return 1
 
@@ -123,7 +123,7 @@
 
 /obj/item/projectile/beam/medical_cell/brute2/on_hit(var/mob/living/carbon/human/target)
 	if(istype(target, /mob/living/carbon/human))
-		target.adjustBruteLoss(-10)
+		target.adjustBruteLoss(-20)
 	else
 		return 1
 
@@ -135,7 +135,7 @@
 
 /obj/item/projectile/beam/medical_cell/burn2/on_hit(var/mob/living/carbon/human/target)
 	if(istype(target, /mob/living/carbon/human))
-		target.adjustFireLoss(-10)
+		target.adjustFireLoss(-20)
 	else
 		return 1
 
@@ -170,10 +170,10 @@
 
 /obj/item/projectile/beam/medical_cell/omni2/on_hit(var/mob/living/carbon/human/target)
 	if(istype(target, /mob/living/carbon/human))
-		target.adjustBruteLoss(-5)
-		target.adjustFireLoss(-5)
-		target.adjustToxLoss(-5)
-		target.adjustOxyLoss(-30)
+		target.adjustBruteLoss(-10)
+		target.adjustFireLoss(-10)
+		target.adjustToxLoss(-10)
+		target.adjustOxyLoss(-60)
 	else
 		return 1
 
@@ -253,7 +253,7 @@
 
 /obj/item/projectile/beam/medical_cell/brute3/on_hit(var/mob/living/carbon/human/target)
 	if(istype(target, /mob/living/carbon/human))
-		target.adjustBruteLoss(-20)
+		target.adjustBruteLoss(-40)
 	else
 		return 1
 
@@ -265,7 +265,7 @@
 
 /obj/item/projectile/beam/medical_cell/burn3/on_hit(var/mob/living/carbon/human/target)
 	if(istype(target, /mob/living/carbon/human))
-		target.adjustFireLoss(-20)
+		target.adjustFireLoss(-40)
 	else
 		return 1
 
@@ -277,7 +277,7 @@
 
 /obj/item/projectile/beam/medical_cell/toxin3/on_hit(var/mob/living/carbon/human/target)
 	if(istype(target, /mob/living/carbon/human))
-		target.adjustToxLoss(-20)
+		target.adjustToxLoss(-40)
 	else
 		return 1
 
@@ -289,10 +289,10 @@
 
 /obj/item/projectile/beam/medical_cell/omni3/on_hit(var/mob/living/carbon/human/target)
 	if(istype(target, /mob/living/carbon/human))
-		target.adjustBruteLoss(-10)
-		target.adjustFireLoss(-10)
-		target.adjustToxLoss(-10)
-		target.adjustOxyLoss(-60)
+		target.adjustBruteLoss(-20)
+		target.adjustFireLoss(-20)
+		target.adjustToxLoss(-20)
+		target.adjustOxyLoss(-120)
 	else
 		return 1
 
--- a/code/modules/projectiles/guns/energy/cell_loaded_vr/ml3m_cells.dm
+++ b/code/modules/projectiles/guns/energy/cell_loaded_vr/ml3m_cells.dm
@@ -31,7 +31,8 @@
 
 /obj/item/projectile/beam/medical_cell/brute/on_hit(var/mob/living/carbon/human/target)
 	if(istype(target, /mob/living/carbon/human))
-		target.adjustBruteLoss(-10)
+		if(target.stat != DEAD)
+			target.adjustBruteLoss(-10)
 	else
 		return 1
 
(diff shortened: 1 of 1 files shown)


Wickedtemp:
Nanaki's brought up a concern that the non-corpsemend cells can be used to also heal corpses, and thinks that they should not be able to heal corpses if their healing values are going to be increased. I don't think this is an issue, since we already have roundstart-available means of healing all damage types from corpses (surgery, slimejelly reactions), but honestly I don't care if that change is made. If you want to remove that usage from the standard healing cells, go for it.

code/modules/projectiles/guns/energy/cell_loaded_vr/ml3m_cells.dm
```diff
@@ -289,10 +289,10 @@
...
 /obj/item/projectile/beam/medical_cell/omni3/on_hit(var/mob/living/carbon/human/target)
 	if(istype(target, /mob/living/carbon/human))
-		target.adjustBruteLoss(-10)
-		target.adjustFireLoss(-10)
-		target.adjustToxLoss(-10)
-		target.adjustOxyLoss(-60)
+		target.adjustBruteLoss(-20)
+		target.adjustFireLoss(-20)
+		target.adjustToxLoss(-20)
+		target.adjustOxyLoss(-120)
```
dwinters99, line 295: Doubling every value is a really lazy way of balancing it and making Omni 3 not only heal **120** oxyloss but also effect corpses is insane and trivializes the already one-track state of medical.
For perspective,  this means that a full mag of omni 3 will heal _**420 DAMAGE**_ on **CORPSES** for basically no cost.

Additionally, on the topic of roundstart corpse healing- corpsemend exists for a reason. Realistically, yes, there are a lot of methods for everything in medical but truly all you need in medical (on non-antag servers ie this one) is chems. Osteo is easy to come by and myelamine is ordered from cargo every shift. Even if you don't get osteo, bonemend surgery is easy. 

I don't understand the idea that you want more options, but then are making the medigun overly good. Maybe _you_ will use the different options correctly, but what everyone else is going to do is use the easiest option so they can go back to their scene or whatever they are doing. 

Also, to really show how good the medigun is in it's current state, I kept someone alive during replicant organ transplant with an unupgraded medigun with OMNI 1, sans immunosuprezene. That means spikes of 250 toxins and oxyloss that were suppressed. If it can already handle burst damage in excess of 250, it's powerful enough.

moth1

Wickedtemp: ...Did you not see the comment above, where I mentioned that the cells would be changed to not heal corpses anymore? Other than corpsemend, of course.

Also..chances are, something else was being done to the patient to actually heal them. The base Omni cell would heal 2.5  toxin per shot and could only heal 10 toxin before having to be recharged. Your story just doesn't add up. 

As for the "lazy method" of rebalancing, the reason behind it is because these values were the original plan in the first place when the ML3M was being made. The healing values were all cut in half in its debut PR. 

Medical starts with a base brute cell, burn cell, and stabilizer. All it can do is heal a little oxy, heal 40 brute damage, and heal 40 burn damage. 

And guess what... it takes 10u bicaridine, or kelotane, to do that. Less, with advanced aid. Both methods are incredibly easy, and doctors are more likely to have chems onhand than the ML3M, which there's only one of (other than the cmo's). If you don't get upgraded cells (and a bunch of them) then you still can't do much with it. Chems are still more powerful and convenient.

...also I don't care if someone "finishes up quickly and goes back to their scene". Like.. dude they paused their shit to handle the job, idgaf that they used the fastest method available. 

Any other issues?

Wickedtemp: Also, what organ were you replacing?

dwinters99: 1. Yes, I did see the comment. I was adding to the existing conversation.
2. The patient was sedated by soporific and was also an SSD sleeve and nothing else. 
3. I see why it was nerfed.
4. Omni is available roundstart and I have NO DOUBT that should it be this powerful it would be made each round. Additionally, the starting cells are better than omni with some judgement.
5. Yes, chems are in-fact the correct option for almost all situations. But putting 10u of a chem AND shooting the medigun is an incredible method of healing even without this change.
6. I don't really know what was meant by this but for most people having to pause their scene is worse than dying. (for reasons I frankly don't understand but it is what it is).

7. These ones. 
```	
	has_organ = list(
		O_HEART =	/obj/item/organ/internal/heart/replicant/rage,
		O_LUNGS =	/obj/item/organ/internal/lungs/replicant/mending,
		O_VOICE =	/obj/item/organ/internal/voicebox/replicant,
		O_LIVER =	/obj/item/organ/internal/liver/replicant,
		O_KIDNEYS =	/obj/item/organ/internal/kidneys/replicant,
		O_BRAIN =	/obj/item/organ/internal/brain/replicant,
		O_EYES =	/obj/item/organ/internal/eyes/replicant,
		O_AREJECT =	/obj/item/organ/internal/immunehub/replicant,
		O_VRLINK =	/obj/item/organ/internal/brainmirror,
		O_STOMACH =	/obj/item/organ/internal/stomach,
		O_INTESTINE =	/obj/item/organ/internal/intestine
		)
```
All of which triggered rejection, resulting in an incredible amount of toxins each time.

Wickedtemp: Bogen basically told me the deets. The only reason the omni cell worked was because the patient was within 7 damage points of death. This does not speak to the strength of the cell, but rather how razor-thin the line was between living and dying. So... yeah. 

Doesn't exactly serve as a reason to not increase the values.

Novacat:
Thinking over this, I think the buff is fine but at the same time the ml3m should not affect dead patients (with exception of corpsemend)

lm40:
i basically agree with ascian, regular cells healing corpses is stupid but other than corpsemend the gun sees barely any use, i am not opposed to trying this and seeing what happens. it might be too much when stacked  with other stuff but it's worth a try anyway.

Heroman3003:
Just a footnote regarding corpsemend: Originally it was meant to be a single-shot cell that also applied defibrillator effect after healing the corpse and healed even more, but making it work was way too convoluted, so instead it became a regular 4-shot cell without the defib extra.

Wickedtemp:
> i basically agree with ascian, regular cells healing corpses is stupid but other than corpsemend the gun sees barely any use, i am not opposed to trying this and seeing what happens. it might be too much when stacked with other stuff but it's worth a try anyway.

Thats what Im hoping to change. I want it to be used outside of healing corpses, and these changes are a decent step towards that.

code/modules/projectiles/guns/energy/cell_loaded_vr/ml3m_cells.dm
Wickedtemp: After I got home from work, I replicated the experiment without the omni cell and pulled it off without a hitch, so I can confirm that while it may have been useful, it could have been done without.

dwinters99:
In hindsight I was a bit too cynical about this. I've thought about it and I reckon we'll just have to see what comes of it. Values can always be adjusted.

Novacat (changes requested):
Honestly just make regular cells not affect corpses and this should be fine.

Wickedtemp:
Well, Iunno how to do that, so you'll probably need to find somebody else who knows more of what they're doing than me. The only reason I did this was that it was a simple change in numbers.

TheFurryFeline:
Apparently, this PR has a Travis build.

Wickedtemp:
As promised, it is Tuesday, and the changes are added. Tested and should be working.

Novacat (approved):

PR #9343, opened 2020-11-23
Lifts nutrition cap even further.

Verkister:
The upstream cap of 500 was just straight up moronic, considering 450 is the "full", and this is VOREStation after all. Even 5000 is far too easy to cap here with only a few preythings. This cap never had any mechanical benefit in the first place.

### DIFF
--- a/code/__defines/mobs.dm
+++ b/code/__defines/mobs.dm
@@ -438,6 +438,6 @@
 #define EXAMINE_SKIPLEGS			0x0080
 #define EXAMINE_SKIPFEET			0x0100
 
-#define MAX_NUTRITION	5000 //VOREStation Edit
+#define MAX_NUTRITION	50000 //VOREStation Edit
 
 #define FAKE_INVIS_ALPHA_THRESHOLD 127 // If something's alpha var is at or below this number, certain things will pretend it is invisible.


dwinters99:
Won't this result in a wave of people eating entire vending machines to be sated for the entire shift? It would be easier to just remove hunger.

Novacat:
When this was added there was a reason why 5,000 was picked specifically, and I forgot what that exact reason was.

Very-Soft:
I'm probably misremembering but wasn't there a thing where the different levels of hunger are percentage based, so increasing the cap to 5000 made everyone who didn't gorge themselves perpetually starving?

Let's make sure that doesn't happen, but otherwise, the cap is arbitrary, long as you're full after 500 I don't care

Verkister:
Nah, it's not percentage based. This only changes how far beyond people can go past the grey icon threshold.

Very-Soft:
no I totes agree with you, I just wanted to make sure that this didn't change the way 0-500 works. The upper limit being an extra 0 is fine by me as long as it doesn't cause any undue weirdness,

lm40:
the only thing i don't like about this is that examine-wise there will be no difference between someone with 4075 nutrition and 50000 nutrition and that violates muh emulsions

Verkister:
Solution: Add more examine thresholds :3c

Little-119:
Does this apply to synthetics? Can I sit in a cyborg charger for five minutes and make my microbattery explode?

Very-Soft:
> Does this apply to synthetics? Can I sit in a cyborg charger for five minutes and make my microbattery explode?

Cyborg chargers only get you to 450 or 500 iirc, no more

dwinters99:
I still worry about what I mentioned before but it would be hypocritical of me to say that balance should be changed because of bad actors. 

This doesn't punish anyone, which is good, but I really feel that there should be SOME countermeasure to miners or such capping themselves out and emptying all the sweatmaxes because they think hunger is annoying. 

I get most of my food from vending machines and there was a string of someone eating the entire contents of them at roundstart. I am trepidatious about this for that reason.

Very-Soft:
> I still worry about what I mentioned before but it would be hypocritical of me to say that balance should be changed because of bad actors.
> 
> This doesn't punish anyone, which is good, but I really feel that there should be SOME countermeasure to miners or such capping themselves out and emptying all the sweatmaxes because they think hunger is annoying.
> 
> I get most of my food from vending machines and there was a string of someone eating the entire contents of them at roundstart. I am trepidatious about this for that reason.

Well unless this changes something I don't understand, you should still be able to fill yourself up such that you don't need to worry about hunger for the shift with two or three protein bars outside of super fast metabolism or lots of blood loss.

I don't think it'll be that much of an issue but if someone makes a habit of going around emptying every vending machine every shift, that's something we can address from an admin level.

Wickedtemp:
> I still worry about what I mentioned before but it would be hypocritical of me to say that balance should be changed because of bad actors.
> 
> This doesn't punish anyone, which is good, but I really feel that there should be SOME countermeasure to miners or such capping themselves out and emptying all the sweatmaxes because they think hunger is annoying.
> 
> I get most of my food from vending machines and there was a string of someone eating the entire contents of them at roundstart. I am trepidatious about this for that reason.

People are gonna eat a lot of food out of annoyance of the hunger mechanics as it is, plenty of folks already do that. Given the sheer number of vendors, as well as other ways of acquiring food (cargo, chef, etc), it should be fine.

PR #9361, opened 2020-11-25
ERT Ship Cleanup

KillianKirilenko:
@Novacat made some silly mistakes when fussing with the armory, though the odds of anyone noticing these was even slimmer than the odds of anyone summoning the ERT.


Novacat:
Ooh, I had cleaned these up on my local branch and forgot to commit, thanks for this!

PR #9374, opened 2020-11-26
Makes pixel shifting work with custom offsets

Verkister:
Pixel shifting now allows mobs with custom offsets to shift themselves 16 pixels from their default offsets instead of point zero. This means that a wide mob can now shift towards left without getting stopped due to being already at -16 at default offset.


Novacat (changes requested):
Either needs to go upstream or //VOREStation Edit

Verkister:
The whole part is already inbetween vorestation edit start and end comments

Novacat (approved):

KasparoVy:
Lovely, cheers

PR #9382, opened 2020-11-29
Add tintable doors

Little-119:
image
...me, apparently. I mean I wouldn't say I'm excited for them, I'm not entirely sure if adding them is a good idea, but I took that as a challenge. So behold my probably shitty code that I've written at about 3 in the morning. It appears to work and lets more tintable doors be added, at least

Problem with this whole concept is that the tinted doors need a separate icon, I can't just change the color like with electrochromic windows. I did a hack job with this by just removing the transparency of the glass part of the sprites (partially because byond's sprite editor is weird and I don't want to figure it out any more), and that was tedious enough


code/game/machinery/doors/door.dm
```diff
@@ -30,6 +30,11 @@
 	var/repairing = 0
 	var/block_air_zones = 1 //If set, air zones cannot merge across the door even when it is opened.
 	var/close_door_at = 0 //When to automatically close the door, if possible
+	// VOREStation Edit Start
+	var/tintable = 0
+	var/icon_tinted
+	var/id_tint
+	// VOREStation Edit End
```
Novacat, line 37: This could be in door_vr
/obj/machinery/door
	var/tintable = 0
	var/icon_tinted
	var/id_tint

code/game/machinery/doors/door.dm
```diff
@@ -508,3 +513,34 @@
 
 /obj/machinery/door/morgue
 	icon = 'icons/obj/doors/doormorgue.dmi'
+
+// VOREStation Edit Start
```
Novacat, line 517: Likewise moved to door_vr

code/game/machinery/doors/multi_tile.dm
```diff
@@ -69,6 +69,32 @@
 	glass = 1
 	assembly_type = /obj/structure/door_assembly/multi_tile
 
+// VOREStation Edit Start
```
Novacat, line 72: multi_tile_vr

Novacat (changes requested):

Novacat (changes requested):
Add multi_tile_vr.dm to the .dme, otherwise that file will not be loaded in.

Wickedtemp:
<3

Little-119:
mlem

KasparoVy:
The mlemmer of the mlemsdale mlemmadone!

Novacat (approved):

PR #9385, opened 2020-11-30
fix da grass

Very-Soft:
makes outdoor grass not have the interior grass edging

### DIFF
--- a/code/game/turfs/flooring/flooring.dm
+++ b/code/game/turfs/flooring/flooring.dm
@@ -140,6 +140,13 @@ var/list/flooring_types
 		'sound/effects/footstep/grass3.ogg',
 		'sound/effects/footstep/grass4.ogg'))
 
+/decl/flooring/grass/outdoors // VOREStation Edit
+	flags = TURF_REMOVE_SHOVEL // VOREStation Edit
+
+/decl/flooring/grass/outdoors/forest // VOREStation Edit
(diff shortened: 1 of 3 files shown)


code/game/turfs/flooring/flooring.dm
```diff
@@ -140,6 +140,13 @@ var/list/flooring_types
 		'sound/effects/footstep/grass3.ogg',
 		'sound/effects/footstep/grass4.ogg'))
 
+/decl/flooring/grass/outdoors // VOREStation Edit
+	flags = TURF_REMOVE_SHOVEL // VOREStation Edit
+
+/decl/flooring/grass/outdoors/forest // VOREStation Edit
+	icon = 'icons/turf/outdoors.dmi' // VOREStation Edit
+	icon_base = "grass-dark" // VOREStation Edit
```
Novacat, line 148: Make a flooring_vr.dm and put these in there. This makes it not need all those //VOREStation Edits. Make sure to check the file in the .dme as well.

Novacat (changes requested):

code/game/turfs/flooring/flooring.dm
Very-Soft: > Make a flooring_vr.dm and put these in there. This makes it not need all those //VOREStation Edits. Make sure to check the file in the .dme as well.

There's already a flooring_vr so I put it in there

Novacat (approved):

PR #9421, opened 2020-12-07
polaris#7778 earlyport - A Thing About Digging Sand and Mineral Processor Speed

Hatterhat:
basically just PolarisSS13/Polaris#7778 but with the necessary adjustments made for the TGUI processor
for people who don't want to read over the other PR:
### A Thing About Digging Sand:
- Replaces the usable_tools list for digging sand with a variable, sand_dig, for mining tools capable of digging.
- Mining tools that hit rocks faster now also dig sand faster, because digging for sand now uses the tool's toolspeed var.
- Shovels maintain their 4 second digging speed delay.
### A Thing About Processor Speed:
- The mineral processor can now ~~initiate TURBO MODE~~ switch its processing speed by a toggle on the console.
- ~~Support for the above hasn't been put on the TGUI interface, but that fancy interface isn't implemented yet, so someone can cross that bridge once it gets here.~~ SURPRISE IT'S THERE NOW.
- Also the ore processor intake is no longer limited to 10 ore chunks per process cycle or whatever.

note: things aren't marked with // VOREStation Edit because i don't see a reason why polaris wouldn't merge this but if worst comes to worst i'll be keeping an eye on this PR and i can just throw down a few of those here and there if some maintainer says "hey throw down the edit markers"


Novacat:
Make sure if Polaris decides to make changes to port those changes over. Due to the TGUI/NanoUI differences it will be kind of difficult to pick out any such changes.

KasparoVy:
Fuuuuuck yyeeeahh buddyyyy. Now I can smelt my shit FAST

PR #9442, opened 2020-12-11
Some taur tweaks

Verkister:
-Tweaks skunktaur shape and markings to look less bad.
-Adds tertiary color option for skunk, otie, wolf, fatwolf, feline, and fatfeline taurs.


KasparoVy:
God bless.

PR #9450, opened 2020-12-13
polaris 7784 earlyport - RPED Quality of Life

Hatterhat:
see PolarisSS13/Polaris#7784
tl;dr:
- RPEDs drop lowest tier of parts and lowest tier of parts only
- click-dragging an RPED onto a DA deconstructs the lowest tier of parts
- also DAs ACTUALLY RETURN MATERIALS TO THEIR PROTOLATHE (LONG LASTING BUG FIXED)

### DIFF
--- a/code/defines/obj/weapon.dm
+++ b/code/defines/obj/weapon.dm
@@ -297,7 +297,7 @@
 
 /obj/item/weapon/storage/part_replacer
 	name = "rapid part exchange device"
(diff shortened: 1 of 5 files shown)


Hatterhat:
also kinda fucked up how polaris actually has taur support or whatever so the item.dm is probably gonna have a future conflict or something i dunno

Hatterhat:
>integration test fails because of active edge

pain

Hatterhat:
#9455

PR #9454, opened 2020-12-14
[MIRROR] Reworks stairs to be sensible

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/7783
--------------------
Stairs are now multi-tile constructs consisting of a bottom, middle, and top stair-piece, as well as an openspace. Moving onto the bottom/top pieces will animate the mob moving to the other space, then move them to the other floor. Any grabbed/dragged items will be brought with the mob, but it ignores grabchains. The middle piece is dense, and can be climbed to go up a floor if the top piece is also present. If something moves onto the openspace, it will fall down the stairs. Stairs no longer protect from falling damage.

Stair spawners are also implemented for mapping and adminbus, they are placed in exactly the same manner as the old stairs. To update maps, just insert a `/spawner` into the invalid paths just before the direction (i.e., `/obj/structure/stairs/spawner/north`). Note that the south-facing stairs did not previously have a special subtype, and since `/obj/structure/stairs` remains in the path, they will still be valid objects (Though look like shrunken stairs because they use the old sprite shrunken to a 32x32 resolution), so mappers will have to manually locate and replace them.
It's also notable that the spawners don't care what already exists in the turfs they spawn onto, and the openspace will overwrite any existing turfs.

Because the stairs move things automatically (on `Crossed()`), it is possible to "chain" stairs by overlapping the top piece of one set with the bottom piece of another. Entering such a chain from the middle is inadvisable, as it appears to be mostly chance which stair will process `Crossed()` second and thus forceMove the mob last. forceMove was used to ensure success, and I did not attempt to just use Move(). [Here](https://puu.sh/GV1WD/85b7cd2ea4.mp4) (.mp4) is an example of such chained movement.

Testing during development was posted to various discords, so most of the people who want to see it have probably seen at least some of it, but [here](https://puu.sh/GWGgQ/9db80f5e27.mp4) (.mp4) is a comprehensive test. The idea to test spaceworms was proposed by Mechoid and it is hilarious.

The diff is also inflated by converting southern_cross_1.dmm to TGM, which would _hopefully_ reduce the frequency of mapmerge conflicts, assuming that the people that use dreammaker don't revert it back to the old format in doing so.


TheFurryFeline:
Last part of this PR needs to be, as said in the discord, fix Underdark stairs, for the record.

PR #9468, opened 2020-12-16
Allow remote control of disposal bins for silicons

### LINKED ISSUE #9457: Disposal chutes "Engaged" button doesn't work
Flush stuff down fastly

#### What actually happened
Button do nothing

#### Steps to reproduce
- be drone
- climb in trash
- hit the 'flush' button on the way in
-or just hit the engaged button at all.

#### Code Revision
-B:release D:
Commit: b769996cf087c536bbfbdefd0be6a621bde5d9e1 

Not sure if this affects carbon mobs or whatever, but it deffo affects drones at least, and it's annoying being as the disposal system is one of their traversal methods.

tigercat2000:
Closes #9457

I'm pretty neutral on whether this should be merged or not, but it was apparently this way in the past.
/tg/ matches the current behavior (no one inside a disposal bin can control it, silicons can not pull the handle)
The new behavior is carbons inside a disposal bin cannot control it, silicons can pull the handle remotely

### DIFF
--- a/code/modules/recycling/disposal.dm
+++ b/code/modules/recycling/disposal.dm
@@ -262,7 +262,7 @@
 	if(..())
 		return TRUE
 	
-	if(usr.loc == src)
+	if(usr.loc == src && !issilicon(usr))
 		to_chat(usr, "<span class='warning'>You cannot reach the controls from inside.</span>")
 		return TRUE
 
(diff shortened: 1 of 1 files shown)


TheFurryFeline:
As I recall, carbons were never able to activate it when hiding inside it but it does go off automatically unless turned off. Just to note

PR #9475, opened 2020-12-17
borgs using IC printer fix - NOT WEBEDIT HELL EDITION

Hatterhat:
fuck how do i write a changelog again OH RIGHT
uhhh
- borgs can use their printer again
- however this comes at the cost of letting people use IC printers without the printer in their hands, which is a nonissue as far as i'm concerned

### DIFF
--- a/code/modules/integrated_electronics/core/printer.dm
+++ b/code/modules/integrated_electronics/core/printer.dm
@@ -29,7 +29,7 @@
 
 /obj/item/device/integrated_circuit_printer/attack_robot(mob/user as mob)
 	if(Adjacent(user))
-		return interact(user)
+		return attack_self(user)
 	else
 		return ..()
 
@@ -86,11 +86,10 @@
 	return ..()
 
 /obj/item/device/integrated_circuit_printer/attack_self(var/mob/user)
-	interact(user)
 	tgui_interact(user)
 
 /obj/item/device/integrated_circuit_printer/tgui_state(mob/user)
-	return GLOB.tgui_inventory_state
+	return GLOB.tgui_physical_state
 
 /obj/item/device/integrated_circuit_printer/tgui_interact(mob/user, datum/tgui/ui)
 	if(dirty_items)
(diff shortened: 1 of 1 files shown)


Heroman3003:
Borgs already have a way to interact with IC printer. They get a special circuit gripper for that, no? If borg doesn't have it, they possibly aren't meant to be able to use it.

kappa-sama:
dick and balls

Hatterhat:
> 
> 
> Borgs already have a way to interact with IC printer. They get a special circuit gripper for that, no? If borg doesn't have it, they possibly aren't meant to be able to use it.

https://github.com/VOREStation/VOREStation/blob/0835907cc8fe100410ddf59655be805ce52deb62/code/modules/mob/living/silicon/robot/drone/drone_items.dm#L130-L149
i dont see ``/obj/item/device/integrated_circuit_printer`` here

LSommers:
> Borgs already have a way to interact with IC printer. They get a special circuit gripper for that, no? If borg doesn't have it, they possibly aren't meant to be able to use it.

damn bro that's crazy but I don't remember anyone making you a maintainer

code/modules/integrated_electronics/core/printer.dm
```diff
@@ -29,7 +29,7 @@
 
 /obj/item/device/integrated_circuit_printer/attack_robot(mob/user as mob)
 	if(Adjacent(user))
-		return interact(user)
```
silicons, line 32: the argument for them not being meant to use it falls sorta flat if this is already here
what other interact is there that isn't the interface?

ShadowLarkens:
this is called "Shadow shouldn't code when tired"

Borgs already had the ability to interact with IC printers before TGUI, then I forgot to go and replace all of the calls to `interact` with `tgui_interact`

This is purely a regression fix, it is silly to argue over it

Hatterhat:
i don't know how half of these dudes showed up but hi welcome to my pr i guess

ShadowLarkens:
@Hatterhat Could you replace
```dm
/obj/item/device/integrated_circuit_printer/attack_self(var/mob/user)
	interact(user)
	tgui_interact(user)
```
with
```dm
/obj/item/device/integrated_circuit_printer/attack_self(mob/user)
	tgui_interact(user)
```
since you're touching this file in this PR already?
interact() is a NOP here, it was just there for initial UI development and then I forgot to remove it

Hatterhat:
uh yeah sure just gimme a second to actually test this and make sure it works before i go do that because the initial commit to this pr is a webedit i made on my phone

PR #9478, opened 2020-12-18
Exploration/Gateway Balance Update

Novacat:
- Adds a new Scatter type of gateway, abductor, which strips the person's gear and deposits them somewhere else
- Nerfs Expedition Phasers and Holdout Phasers
- Reduces the number of full size phasers in Exploration from 4 to 3
- Slightly buffs melee mobs
- Removes humanoid mobs from Virgo 3B Wilderness
- Removes ADG from one of the POIs from V2

### DIFF
--- a/code/datums/outfits/jobs/special_vr.dm
+++ b/code/datums/outfits/jobs/special_vr.dm
@@ -9,6 +9,18 @@
 	id_type = /obj/item/weapon/card/id/centcom
 	pda_type = /obj/item/device/pda/centcom
 
+/decl/hierarchy/outfit/job/emergency_responder
+	name = OUTFIT_JOB_NAME("Emergency Responder")
+	uniform = /obj/item/clothing/under/ert
+	shoes = /obj/item/clothing/shoes/boots/swat
+	gloves = /obj/item/clothing/gloves/swat
+	l_ear = /obj/item/device/radio/headset/ert
+	belt = /obj/item/weapon/gun/energy/gun
+	glasses = /obj/item/clothing/glasses/sunglasses
+	back = /obj/item/weapon/storage/backpack/satchel
+	id_type = /obj/item/weapon/card/id/centcom/ERT
+	pda_type = /obj/item/device/pda/centcom
+
 /decl/hierarchy/outfit/job/clown
 	name = OUTFIT_JOB_NAME("Clown")
 	uniform = /obj/item/clothing/under/rank/clown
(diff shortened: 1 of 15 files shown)


dwinters99:
1. I'm happy to see the abductor ship, but I'm dubious of the stripping mechanic. The existing scatter gateways make it easy to abort if things go south, but this seems like a big commitment, especially for what is usually beginning-of-shift warmup content.
2. Deserved. The frontier phaser (+carbine) is probably the best weapon on the code for the simple reason that it has unlimited reloads and is so versatile.
3. This seems odd to me considering running out of phasers is very common in my experience. This also punishes getting volunteers unless you order from cargo. Volunteers are already very hard to come by but limiting resources to a point where it's not worth wasting a phaser on an unexperienced person is discoruaging. 
4. Don't really think that melee mobs needed a buff, for the simple reason that they already can absolutely destroy you if you aren't careful, and the problem still remains that getting hit by a melee mob is basically impossible unless you position yourself incorrectly. Increasing attack speed seems a bandaid that will only punish low-skilled players, because the more experienced will continue to kite them. Also, this doesn't address the problem with inconsistent attack wind-ups. (i.e. bears and sif fauna having windups, making them basically impossible to get hit by but spiders and carp having frameperfect mollywops)
5. Good. Mercs shouldn't be on V3b. It also stops science from wandering out and wondering why they died because they weren't equipped.
6. ADGs feel like lavaland mobs and are EXTREMELY not fun to fight on our code so I'm happy with this change. Although this appears to be the easy one at rockybase rather than the overwhelmingly boring fight at DoomP.

Very-Soft:
> 1. I'm happy to see the abductor ship, but I'm dubious of the stripping mechanic. The existing scatter gateways make it easy to abort if things go south, but this seems like a big commitment, especially for what is usually beginning-of-shift warmup content.

Ideally whoever includes the stripping mechanic into their gateways will be giving people the things that they need to succeed in the given gateway.

And just because the gateway is presently 'beginning-of-shift warmup content' does not mean it was ever meant to be relegated to that role. The gateway is supposed to be a mixxed bag. It's also optional, so, if people decide that they don't have time to risk it, they can just not do the gateway. I very much do not care about gateway speedrunning.

KillianKirilenko:
Two thoughts;
1. Abductor is liable to break hardsuits/RIGs, from past experience with the old abductor ship. I strongly recommend putting in some **very thorough** checks to properly unequip them, or else they might break and be impossible to unequip if you pick them up and put them back on once you find the control module.

2. Your phaser nerfs are sloppy. They'll be back to base energy usage (300/600) if toggled into low-power and back into standard mode.

If you really want to nerf them, I recommend maintaining the current power consumption and instead reducing the damage. Ditch the low-power mode and give them a custom beam that does 30 damage (down from 40) or make /weaklaser the default (25dmg) and make them slightly more power efficient.

Novacat:
Weaklaser does 15 damage a shot and is extremely efficient, the main phaser gets 40 shots, so they effectively will not run out of ammo. They basically are identical to the suppressive fire option on laser rifles.

PR #9522, opened 2020-12-26
Fix security records consoles' inability to search fingerprints

### LINKED ISSUE #9467: Security record consoles can't search fingerprints
Security records consoles cannot search with fingerprint hashes, forcing poor forensics technicians such as myself to manually check each and every record to find a matching fingerprint hash

Notably, the security records console says it'll accept DNA hashes, despite security records not including DNA hashes. I suspect someone copypasta'd from the medical records console
Security records probably should include DNA because it's useful in forensics, but that's a suggestion for another day

To have a security record appear when searching with a fingerprint hash

#### What actually happened
Got the message 'Security record not found. You must enter the person's exact name, ID or DNA.""

#### Steps to reproduce
- Be part of security, or at least have security access
- Find the fingerprint hash of someone with a security record, such as oneself
- Log in to a security records console
- Enter the fingerprint hash into the search bar, hit enter
- Observe message and lack of security record appearing

#### Code Revision
- cced5174e043c235c9b8c9fd6e4f0a435dc5c4d6 

image

Little-119:
Fine, I'll do it myself.

Fixes #9467

### DIFF
--- a/code/game/machinery/computer/security.dm
+++ b/code/game/machinery/computer/security.dm
@@ -317,16 +317,16 @@
 				if(!length(t1))
 					return
 
-				for(var/datum/data/record/R in data_core.security)
-					if(t1 == lowertext(R.fields["name"]) || t1 == lowertext(R.fields["id"]) || t1 == lowertext(R.fields["b_dna"]))
-						active2 = R
+				for(var/datum/data/record/R in data_core.general)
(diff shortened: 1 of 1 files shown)


Novacat:
...is this a TGUI-specific bug?

Little-119:
@Novacat The changes I made were in a proc named `tgui_act`, and I don't think it was a thing with the old UI, so I'd assume it's a tgui-specific bug
Why do you ask? That question makes me feel like I'm unaware of some significant stuff

PR #9525, opened 2020-12-27
Makes pixel shifting work while resting

Verkister:
Every other method of incapacitation is handled by the other procs, leaving "canmove" to functionally only hinder conscious proning. Might send to polaris at some point, but they still lack the features this "fixes" and could probably use a proper de-spaghettification rework on the whole system


mel-byond:
Seconding this, I love this idea since I always have to get up and then shift the position....

PR #9550, opened 2021-01-05
Remaps the tram to fit more crew

Detective-Google:
The tram has had a complete overhaul and has been embiggened to fit more people; among other things. Also edits the station to accept the bigger vehicle.

thicctram

(since the walls don't show on the editor here's a preview of those.) 
image


Very-Soft:
I would vastly prefer if this used the white shuttle walls and windows iver the hull walls. Its not a spaceship flying through dust clouds, it's a civilian transport that scoots through a tunnel.

Andromeda-K22:
Walls look fine as it is, imho. White walls are old and crusty looking.

TheToaster98:
Looks very nice! And while I also think it would benefit from another wall type, I also think the old tram tiles are too old and need a remake, maybe a different material for the hull?

Iroquois-Pliskin:
Good PR. Tram has seen some seriously cramped periods now and again, and the extra size will help space people out.

mel-byond:
WHOO, I love the idea of a Tram redesign!

KillianKirilenko:
Whilst I think it's good for the tram to have its own unique design (those wall/window sprites are very much showing their age), it's also a little weird to see it using the same reinforced mesh hull as most of the shuttles out there.

It shouldn't be very hard at all to make a custom wall (with an adminbus-strength material, if need be) that has a reasonably distinct look, even if that look is just 'pure white'. I'd even suggest using the curvy sprite set, but it has these odd spots at each corner that would probably look kinda ugly without some tweaking.

DemicusMaximus:
Looks good to me!

PR #9553, opened 2021-01-07
[FEATURE] Add HUD icons for Soulcatcher guests

mel-byond:
I noticed that I was missing some kind of HUD for folks in soulcatcher, so decided to make some icons and hopefully get an HUD working.
 
Trying to base from the ghost HUD and the logic behind it.
screen_nifsc


mel-byond:
Building on my side tooooo so uhhhh, this might break

mel-byond:
Hm, seems it atleast doesn't give errors while building, but can't get the HUD to work, argh. Anyone got a clue to why?

mel-byond:
Seems it works in game! Cool!! Animations and all, give it a go.
image

PR #9559, opened 2021-01-09
Adds tertiary color glow/accents for the synth taurs

Verkister:
-Little "glowy" bits for synthetic wolf, feline, horse, and lizard taurs.


KasparoVy:
Amaaaziiingg

PR #9567, opened 2021-01-11
the big tram now has fancy titanium hull similar to the old one

Detective-Google:
I didn't realize this existed because it had no preview in the editor; but this should hopefully assuage a lot of the concerns people had over the tram not having that distinctive look anymore. 

image

Still uses reinforced windows because they still match the best.

### DIFF
--- a/maps/tether/submaps/tether_misc.dmm
+++ b/maps/tether/submaps/tether_misc.dmm
@@ -3296,7 +3296,7 @@
 /area/space)
 "Sl" = (
 /obj/structure/sign/nanotrasen,
-/turf/simulated/wall/rshull,
+/turf/simulated/wall/rthull,
 /area/shuttle/escape)
 "Sm" = (
 /turf/simulated/shuttle/wall/alien/blue/hard_corner,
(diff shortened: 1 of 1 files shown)


KillianKirilenko:
I suggest using the unreinforced version that's clean and doesn't have the hex-mesh over it, might look a little cleaner and more tram-like.

Should just be `/thull/` rather than `/rthull`.

pontifexminimus2:
I would like a picture of the titanium hull unreinforced. Because I kind of like the hex mesh.

Ketrai:
Biggest issue I have is the poor consistenty of the hex mesh. Some are bigger or smaller,  and in some cases the lines even make squares that outright interrupt the pattern. The anti aliasing might be faulty too but I'd have to analyse the overlay up close for that.

KillianKirilenko:
> Biggest issue I have is the poor consistenty of the hex mesh. Some are bigger or smaller, and in some cases the lines even make squares that outright interrupt the pattern. The anti aliasing might be faulty too but I'd have to analyse the overlay up close for that.

Originally I just thought this was your eyes playing tricks on you, but no. On a closer inspection of the mesh sprite, it basically bulges outwards slightly towards the viewer. I'm not too surprised given that making a seamlessly-repeating hexagonal overlay at that resolution must have been tricky, but it does look a little weird if you look too closely.

Very-Soft:
Whatever the walls end up looking like, can you shift the tram station expansion on the station side north a bit, so it doesn't cover up the cafe windows?

Novacat (changes requested):
I agree on unreinforced version of the wall, also Soft's requests.

PR #9571, opened 2021-01-13
Allows feeding synths with the lewd trait

Verkister:
A completely unnecessary no-fun-allowed check for a lewd scene ability. Crew synths use the nutrition var just like the rest of the crew do so there's no mechanical reason for this to not work on them.


KasparoVy:
Yum.

PR #9577, opened 2021-01-16
Restricts myarps (author request)

Verkister:
Azura's request. It is their personal character and voice after all.

### DIFF
--- a/code/modules/mob/living/carbon/human/emote_vr.dm
+++ b/code/modules/mob/living/carbon/human/emote_vr.dm
@@ -77,7 +77,8 @@
 		if("myarp")
 			message = "lets out a myarp."
 			m_type = 2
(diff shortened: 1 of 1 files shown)


Novacat (changes requested):
Something like this should probably be cleared through a fluff item request. Since it is a trivial item it will certainly be accepted.

Otherwise, I would rather the entire emote was restricted rather than just only the sound file.

Very-Soft:
Yeh if they want it removed that's fine, but if it's a thing restricted to their ckey, there's an official process for that.

Very-Soft:
I offered a replacement myarp sound if you wanna set the normal myarp sound to that~

KasparoVy:
*merp :D

Novacat:
It has been a few months and there has been no response.

PR #9579, opened 2021-01-16
Allows Custom Species to select any of 5 physical genders.

Heroman3003:
Title. While potentially desirable and makes sense for other species to be restricted to only what their lore allows, Custom Species should totally allow full selection.


KasparoVy:
Awesome, thanks much

PR #9580, opened 2021-01-16
Allows every species to select any biological sex.

Heroman3003:
Separate from the other PR just in case.
Also moves promethean var adjustments to their own file because why would they be in two different places.


KasparoVy:
Sweet.

PR #8209, opened 2020-06-06
Allows All Species to Select Certain Neutral Traits

KasparoVy:
You can now pick the autohiss, bloodsucker, succubus, hard vore, trashcan, expensive taste and glowing eyes neutral traits as any species.

virgo ss13 neutral traits for all 20200606

PR #9582, opened 2021-01-17
Makes most neutral traits available to all species

Heroman3003:
Title. All species can now select limited range of neutral traits. Not included: temperature-adaptation traits, metabolism traits and autohiss traits.

Other changes:
All the verbs provided by neutral traits removed from species' default roster. Now have to be manually added via neutral traits.
Fixed custom blood color only applying to Alraune, Xenochimera and Custom Species.

Probably breaks like entire game lol, but seemed to work fine in basic testing, if someone actually smart could look at it, would be nice.

### DIFF
--- a/code/game/dna/dna2_helpers.dm
+++ b/code/game/dna/dna2_helpers.dm
@@ -237,20 +237,9 @@
 
 		// Technically custom_species is not part of the UI, but this place avoids merge problems.
 		H.custom_species = dna.custom_species
-		if(istype(H.species,/datum/species/custom))
-			var/datum/species/custom/CS = H.species
-			var/datum/species/custom/new_CS = CS.produceCopy(dna.base_species,dna.species_traits,src)
-			new_CS.blood_color = dna.blood_color
-
-		if(istype(H.species,/datum/species/xenochimera))
-			var/datum/species/xenochimera/CS = H.species
-			var/datum/species/xenochimera/new_CS = CS.produceCopy(dna.base_species,dna.species_traits,src)
-			new_CS.blood_color = dna.blood_color
-
-		if(istype(H.species,/datum/species/alraune))
-			var/datum/species/alraune/CS = H.species
-			var/datum/species/alraune/new_CS = CS.produceCopy(dna.base_species,dna.species_traits,src)
-			new_CS.blood_color = dna.blood_color
+		H.species.blood_color = dna.blood_color
+		var/datum/species/S = H.species
+		S.produceCopy(dna.base_species,dna.species_traits,src)
 		// VOREStation Edit End
 
 		H.force_update_organs() //VOREStation Add - Gotta do this too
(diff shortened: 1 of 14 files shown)


Novacat:
Most of this seems fine, the only part that I think might require discussion would be if we want blood color to be configurable for vanilla species.

Very-Soft:
I think vanilla species being able to pick their blood color is fine. Humans can have purple skin, angel wings, animal ears and tails if they want to, changing their blood color to something other than red, while weird, hardly seems a stretch.

KasparoVy:
Hey, does this bypass the same problems that affected https://github.com/VOREStation/VOREStation/pull/8209?

Heroman3003:
> Hey, does this bypass the same problems that affected #8209 where if one person who is eg. Unathi choses neutral traits, because the species datum for non-custom species is common to all who have that species, and neutral traits (were) applied to species level, it would affect every Unathi on station?
> 
> I'm not sure if anything changed since the linked that would've fixed this bug

This PR was made pretty much from scratch to not have any of the problems either of previous PRs had. A small refactor to how species application works allowed for what seems to be a working version of every species datum being created individually for the player like custom/xenochim/alraune already were rather than singletons, like regular species. Although, again, there is potential of other issues arising, but basic preliminary testing showed no issue of traits universally applying and nothing seemed explicitly broken either.

KasparoVy:
That's badass

Thanks @Heroman3003, sounds great to me then

code/game/dna/dna2_helpers.dm
```diff
@@ -237,20 +237,9 @@
...
-			var/datum/species/xenochimera/new_CS = CS.produceCopy(dna.base_species,dna.species_traits,src)
-			new_CS.blood_color = dna.blood_color
-
-		if(istype(H.species,/datum/species/alraune))
-			var/datum/species/alraune/CS = H.species
-			var/datum/species/alraune/new_CS = CS.produceCopy(dna.base_species,dna.species_traits,src)
-			new_CS.blood_color = dna.blood_color
+		H.species.blood_color = dna.blood_color
+		var/datum/species/S = H.species
+		S.produceCopy(dna.base_species,dna.species_traits,src)
```
KasparoVy, line 242: HEEEEEEELLLL YEEEEEEAHHH
Great find!

MrSign:
Ye, sounds good. I'll benefit from it cause I sometimes keep forgetting to toggle the auto-hiss in the OOC tab.

Heroman3003:
> Ye, sounds good. I'll benefit from it cause I sometimes keep forgetting to toggle the auto-hiss in the OOC tab.

I... don't think this PR is going to affect that. The auto-hiss needs to be manually toggled for both base and custom species, iirc, and the auto-hiss traits are not among those available to the default species.

MrSign:
> > Ye, sounds good. I'll benefit from it cause I sometimes keep forgetting to toggle the auto-hiss in the OOC tab.
> 
> I... don't think this PR is going to affect that. The auto-hiss needs to be manually toggled for both base and custom species, iirc, and the auto-hiss traits are not among those available to the default species.

Insert cat scream.jpg here.

PR #9585, opened 2021-01-17
Rebuilds the wideborg sprite file

Verkister:
Just a little attempt to see if this might fix the very frequent cache bug people seem to be having with this particular sprite file. If this doesn't work I might have to trim it a little or chop it into separate files. It is one of the largest sprite files we have after all.


KasparoVy:
Good thinking

PR #9590, opened 2021-01-23
Adds a new vorepanel mode addon.

Verkister:
Adds a new vorepanel mode addon "Affect Worn Items", which, depending on the belly settings, will either contaminate or digest worn gear right off the prey without stripping. Item friendly modes will affect the prey's entire outfit at once, but the item digest mode will only affect one item at a time using the same digestion rates the belly would have on loose items.


Novacat:
It should probably not digest items that cannot be removed (can_remove var I believe?).

Verkister:
done

PR #9597, opened 2021-01-23
Removes Herm option for biological gender

### LINKED ISSUE #9595: Pronouns broken
Doesn't actually seem to be reading pronouns. As an example, this is a default human character.
image

Heroman3003:
No bias against herms, just an issue that I missed in testing of #9580. Aparently not only gender is byond-defined var, but also can only have byond's own hardcoded values stored into it with no way to add your own, which is why gender identity exists the way it does and breaks half the time. Herm was vorestation only gender identitiy and attempting to put it into regular gender var caused a runtime and the pronoun to default.

Ideal fix would be to make an entirely different biological gender var for use on mobs. But that'd require extensive refactoring all-around and cause too many issues, so this is just simpler solution. Identity still works for getting shi pronoun.

Fixes #9595


Novacat:
This is unfortunate but better than not having it work at all.

PR #9598, opened 2021-01-23
Fixes defib timers being 10 times as long compared to intended.

Heroman3003:
As title says. Math in the code assumed Life() ticks happen every two world.time ticks, but in actuality every 2 seconds (+- time dillation, lag, etc). And since every second is 10 world.time ticks... That meant that the defib timers were 10 times as generous as they were actually meant to be. Fixes that.

Earlyport of https://github.com/PolarisSS13/Polaris/pull/7838

### DIFF
--- a/code/game/objects/items/devices/defib.dm
+++ b/code/game/objects/items/devices/defib.dm
@@ -509,8 +509,8 @@
 		return // Still no brain.
 
 	// If the brain'd `defib_timer` var gets below this number, brain damage will happen at a linear rate.
-	// This is measures in `Life()` ticks. E.g. 10 minute defib timer = 6000 world.time units = 3000 `Life()` ticks.
-	var/brain_damage_timer = ((config.defib_timer MINUTES) / 2) - ((config.defib_braindamage_timer MINUTES) / 2)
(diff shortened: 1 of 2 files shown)


dwinters99:
This is a good fix but I want to ask if tweaking the values in a VSEdit might be on the table because 10 minutes is a really short time for off-world SAR and non-scene resleeving is a bad alternative to cloning vis a vis defib timers. Having to wait 2+ minutes sitting still in the sterile and dull resleeving room is non-preferable to simply watching paramedics / your team come and recover you while ghosted.

Heroman3003:
> This is a good fix but I want to ask if tweaking the values in a VSEdit might be on the table because 10 minutes is a really short time for off-world SAR and non-scene resleeving is a bad alternative to cloning vis a vis defib timers. Having to wait 2+ minutes sitting still in the sterile and dull resleeving room is non-preferable to simply watching paramedics / your team come and recover you while ghosted.

The timers are defined in configs, so that wouldn't be a github PR kind of thing regardless. This fix just makes it so time that is defined in config actually works the way it says it should, since the notes in config file AND in configuration code file both say that amount is defined in minutes.

aryll-linton:
Ah, I see. If that's the case I'll see about asking someone with config access about it. Cheers.

PR #9610, opened 2021-01-25
Adds two new taur body types: An alternative naga tail, and a scolipede taur

KeRSedChaplain:
As it says on the tin, two new taur types, one is a snake tail that isn't as long or curls upwards like the original, and the other is a scolipede taur body.


KeRSedChaplain:
oh thats fun it had everything from the master branch. fuck

PR #9615, opened 2021-01-26
Converts Morph Spawn and Maint Pred events to use ghostpods

Heroman3003:
Basically, instead of doing a one-time Query for ghosts, the event now spawns a ghost pod that is only visible to ghosts and can be used at any point after the event.

Also adds a new verb to help with finding those: Find Ghost Pod. Gives you a list of all 'active' ghost pods. Right now only Maintenance Holes, Weird Gunks and those drone pods that have already been attemptedly activated are considered 'active'.

Practically, should basically mean that if maint pred/morph spawn event happened and nobody took initial opportunity, anyone later on can just use the verb to see if there are opportunities still open and take them later.
Also makes it easier for admins to spawn those things.

Probably needs more testing, but I'm sleepy as I'm posting this and just want to get it up before going to bed and finishing up tomorrow.


Heroman3003:
Alright, some preliminary testing and it seems to be working fine.

PR #9622, opened 2021-01-28
New advanced Egg Release mechanics.

### LINKED ISSUE #9190: Egg TF Function carries over the Muffles add-on.
The Egging function seems to carry over belly add-ons from vore bellies, even when the prey is released from the egg.

To not be muffled even after being released from the egg.

#### What actually happened
Muffled no matter what.

#### Steps to reproduce
- Eat person
- Have muffles on
- Use the 'Encase in Egg' option
- Spit egg out
- Release prey from egg
- Have the prey try and use the 'Me' and 'Say' emotes
 

- It's said that one way to fix this is for the prey to be put in a belly without the muffles add-on turned on, though I haven't been able test it out to confirm it.
-Muffles might not be the only add-on affected by this as all add-ons might carry over to the prey the same way.

Verkister:
-The vore release eggs are no longer reskinned lockers, but storage items that can have their contents either dumped or pulled out, but the stuff can't be put back into a cracked shell.
-A live prey in an egg will have the egg scaled according to their size, and they can get out by resisting.
-Resisting inside and egg will give a message and a little wiggle animation before letting the occupant out.
-Resisting inside an egg that is still inside someone's vorebelly will relay the struggles to the belly instead.
-Also now items can be egged, either one at a time or all at once. The single item eggs are scaled according to the contained item's size class and take the same amount of storage space as the item itself would.
-Full stuff eggs are a little more lenient on the size and weight, basically for the reason that one prey's full outfit could fit in one backpack sized egg rather than going full macro based on the total weight.
-Eggs that are heavier than a backpack get increasing slowdown rates depending on the extra stuff.
-Also there's a whole bunch of new egg type options and some better looking sprites coming.
-Also the new eggs can be picked up and even trash-eaten.
-The old locker eggs still exist in the code and can be spawned should anyone have a need for one.
-Also fixed the char setup egg prefs not saving or loading correctly due to obsolete relic code.

Also fixes muffled prey released inside objects not getting un-muffled on release.
Fixes #9190


TheFurryFeline:
Does this fix #9190?

Verkister:
I have no idea tbh

Verkister:
Nvm it does fix it now :v

TheFurryFeline:
Hmm, s'pose we'll find out.

Verkister:
hmmh dunno what the heck is going on with those unit tests so i'm doing another local test run for those edits to make sure it's not mysteriously breaking the whole glob systam with those muffling fixes

TheFurryFeline:
> Nvm it does fix it now :v

Oh, weird, didn't see this reply when I sent that email response. Yay for fix.

Verkister:
CI is having its own issues so this PR's just being held hostage by the other end technical fuckups that have jack shit to do with this code. weh

Verkister:
Oh. my. fucking. god. lmao that actually worked??? Nice!

Verkister:
I think I'm gonna still add a feature where struggling inside an in-belly egg would just redirect the resist into the belly once i get a chance to figure it out.

TheFurryFeline:
Something like isinbelly && escapingfromlocker or whatever the var is for
that namesake?

PR #9625, opened 2021-01-29
The "Macro" Micro update! ((DO NOT MERGE CURRENT WIP))

H0lySquirr3l:
Hello Hello, This is me finally putting my foot down and stopping my own very bad brand of feature creep. This is an update I've been cooking for a very long time and hope to see through to completion! This update will include a small collection of buildings, tracks, planets, and possibly even vehicles for the crew to play around with and make their sizeplay fantasies come alive! Currently, all of the sprites are finished and all that's left is coding. The actual file only contains a bunch of TODO comments. I am not the best at coding so any help would be greatly appreciated!

Main to do list:
Items:
Buildings can be picked up from suitcases and plopped down with a choice of sprite and color.- [ ] 
Buildings allow for players less than or equal to a scale of 25% to enter and hide in buildings like lockers - [ ]
Buildings can be destroyed and recollected to be build elsewhere - [ ] 
Tracks can be picked up and put down. - [ ]
Tracks will check players scale, less than 100% can walk on them, more than 100% leaves a special "stompy" sprite and play a special walk sound. - [ ] 
Tracks can be "cleaned" in some way... either through verb or possibly treated like normal dirt.- [ ]
Make special planet snacks that act like special candy containers, They come with three planets and a ship, with a super rare chance of getting a "real" planet to munch on.- [ ] 
When planets are munched they let off a spark and special sound, then allow for two more nibbles after.- [ ]
Create cargo crates for items- [ ] 
Create custom sounds for all interactions. - [ ] 

Secondary to do list:
Make toy car and spaceships driveable by players under 25% scale.- [ ]


H0lySquirr3l:
Closing due to opening this way earlier then needed. No reason to leave it up here without any proper code to be criticted.

PR #9647, opened 2021-02-06
Chorn sprites.

H0lySquirr3l:
Added in 10 special candy corn style horns. :) I think people could get a kick out of using them. I call them "chorns". They do have a small inspiration from a obscure web comic that not a lot of people know about.

image

((for the record I should have put some type of specific thing saying this was "sarcastic" My bad on that.))


Andromeda-K22:
bruh literally everybody knows what this is from

H0lySquirr3l:
> 
> 
> bruh literally everybody knows what this is from

Just some silly sarcasm.

PR #9654, opened 2021-02-08
Syndicate Cruiser Tweaks

Novacat:
- Buffs up the armory slightly
- Changes the armor to reinforced steel
- Removes stationary translocation beacons (they were useless)
- Adds Inflatables to the uplink

### DIFF
--- a/code/datums/uplink/tools_vr.dm
+++ b/code/datums/uplink/tools_vr.dm
@@ -41,6 +41,11 @@
 	item_cost = 10
 	path = /obj/item/clothing/suit/space/void/autolok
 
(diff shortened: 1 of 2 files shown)


KillianKirilenko:
> Changes the armor to reinforced steel

Aw. I made the special hull type for the *stealth aesthetic* and everything.

> Removes stationary translocation beacons (they were useless)

Sadly. They were a copy/holdover from the old designs that I could never figure out how to set up. I'd hoped to be able to make it so you could teleport hostages/infiltrators back to the ship from the station, and for the ERT to be able to do the same.

Issue #9656, opened 2021-02-08
Mech Bay Northern Door Has Incorrectly Set Button


HzDonut:
#### Brief description of the bug
The northern blast doors of the robotics mech bay has the button on the inside set to open the doors to the southwest instead of allowing you to close the northern door from the inside. Also I think these buttons weren't placed correctly either, since you can see both from either side unlike the ones to the southwest where you can't see the inside button from the outside and vice-versa.

#### What you expected to happen
Pressing it would allow me to close the door from the inside of the mech bay.

#### What actually happened
Instead it opened the doors located to the southwest which meant I had to use the button on the outside to close these doors and then walk back around.

#### Steps to reproduce
- Head to the mech bay
- Press the button on the inside to try and open/close the northern doors
- Witness the other set of doors open/close instead

#### Code Revision
- Server revision: B:release D:
Commit: 9fd011e32ef6528afcce58826eb7f5ab271c2314 
 

#### Anything else you may wish to add:
- https://gyazo.com/d4c3604a93d9ff724078964e716bb64d the buttons as seen from the outside
- https://gyazo.com/0f29a9f0a27bc90cdb5599f6fbb13bfc the other set of doors as from the outside
- https://files.catbox.moe/gw7kuv.gif the button making the wrong doors open

H0lySquirr3l:
After speaking with Soft this has been decided as not being a issue but just a design choice. Probably should close.

![image](https://user-images.githubusercontent.com/34927367/108759183-a51d8080-7511-11eb-8767-d918719a20fa.png)

PR #9673, opened 2021-02-10
Fixes mech doors.

H0lySquirr3l:
Fixes issue #9656


H0lySquirr3l:
Hold up soft just brought up that this wasnt a bug at all persay. But more a undocumented design choise.

H0lySquirr3l:
Never mind! I just wasn't understanding what the design process was. The buttons are placed so that you can control both sets of doors should you finish making a mech but not be going back into science. Or vice versa needing to drop off a mech but not actually going back into science. https://cdn.discordapp.com/attachments/187013248309002240/808901710311456778/mech_doors.png Isses should still be closed.

PR #9674, opened 2021-02-10
fixes small error with railing on second surface maint

H0lySquirr3l:
2021-02-09 18_32_47-FastDMM2 — Mozilla Firefox

image


Novacat (changes requested):
Needs mapmerge ran

H0lySquirr3l:
I can't run that program and I though that online map maker did it automatically so rip.

PR #9695, opened 2021-02-12
No ERP -> Unset

TheFurryFeline:
Just feels like this would be better since having No ERP as default might turn potential partners away if the player is unaware they can switch this to what they want. 

🆑 
tweak - Change default setting for ERP prefs to unset to encourage asking if unsure.
/🆑


KasparoVy:
Good call

PR #9705, opened 2021-02-13
Adds area soundproofing for dorms.

Verkister:
Areas can now be soundproofed to block sounds from other areas from entering or sounds within the area from leaving.


Little-119:
Great, now no one will be able to hear my victims scream

PR #9708, opened 2021-02-15
Increases the duration of Resleeving Sickness

Novacat:


Wickedtemp:
Can we have, like.... details, in this?

How long it's been increased to, reason for the increase, etc?

Detective-Google:
> Can we have, like.... details, in this?
> 
> How long it's been increased to, reason for the increase, etc?
> 
> Edit: forgot I could just check. Why exactly are we _doubling_ the timer?

This, seems like punishing people for taking part in the main activity of the server tbh.

Wickedtemp:
So, the resleeving sickness only really triggers if you've died by non-vore means. So if you get digested, it's whatever, but if you get shot to death and are resleeved, then you end up with the debuffs. 

The issue, though, is that it seems unprompted and rather pointless. We don't have many combat-heavy incidents, and even in our most recent Merc round earlier today, we only had to resleeve two people as a result of the fighting. What's more likely is that this will effect miners and Science crew, and explo more than anyone else. For over an **hour**. A full 25% of the round if they're unlucky and Science/Medical didn't upgrade anything. What exactly is the point?

Very-Soft:
90 minutes of 'don't play the game' is lame, me no like. I never liked resleeving sickness to begin with, and it's not like it's really been being abused. :V

KillianKirilenko:
from #dev-general when it was PR'd:
> Ascian: imo the current numbers are a tad too low, and we already solved the issues with digestion accidentally triggering it
> Ascian: which was the only reason why it was low to begin with
> Ascian: that being said I also increased the impact of upgrades
> Ascian: but its still going to be overall higher

Perhaps 60/40/20 or something along those lines would be a better balance? 90/60/30 seem a bit excessive.

Screemonster:
I see absolutely zero benefit in the sickness lasting longer than the amount of time where we wouldn't bat an eye if someone just didn't get a sleeve at all and just rejoined from the menu
why the fuck should someone be punished with 90 minutes of suck for getting resleeved when if they'd just died and not been recovered, nobody would give a shit if they waited 30 minutes and got resleeved at central with no penalty

Wickedtemp:
TLDR: this wont make death matter and nothing will outside of changing the entire server. all resleeving sickness is, is time-out for people that died in combat, so they can't immediately rush back in with an endless stream of resleevings. It already accomplishes that goal and any increases to its length or severity starts to threaten the viability of more combat-centered events in the future. hard pass.
----------------------------------------------------------------
So, this is what you said on the initial resleeving sickness PR, and I'm concerned that this is at least a partial motive for this suggestion,
""I am fine with the lengthy timer. One of the long-standing issues we have had is people not treating death as a traumatic experience, I feel like this would be an excellent solution to that.""

On any other server, you'd be correct. I even agree that a lot of people don't put any thought into how their characters would actually react to dying. A lot of people don't put any thought into their characters actions **at all**, but at the end of the day, death is never going to matter here if you don't want to make a lot of people really uncomfortable, really quickly. 

- We'd need to remove round persistency. People die all the time on other servers and they're able to treat every round like the worst round of their characters lives because every round is a New Shift Where This Stuff Never Happened Before. We don't do that here. Every round, save for huge catastrophe's caused by bugs or whatever, is canon. Deaths are canon unless all players involved decide to retcon, and even then you'll always have players that were on for that round, heard what happened, and didn't get the news of the retcon. If death actually matters, at what point do you require characters be put on mandated medical leave for their own wellbeing, because we've got lots of people dying every shift. Unless...

- Unless we make special exemptions for scenes. We usually do. Vore is exempt from all of the SOP/Regs ICly, and from mechanical effects of resleeving sickness, so why not make vore related deaths from 'mattering'? Because it doesn't **work that way.** This is an all-or-nothing and it just isn't going to work. If you allow exceptions to when death matters, then death doesn't matter, because if it **actually did**, we wouldn't have exceptions to start with. The presence of exceptions immediately means that we'll always have characters and players that don't give the slightest of shits about character deaths, and have no compelling reason to, because there's written lore that says they don't have to. But, given the nature of the server, we can't do that if we want to avoid becoming one of the most comically depressing stations in SS13.

- The most we can do is say "death sucks and should be avoided when you can I guess" and to an extent, we do a decent job of that. We've **been doing** a decent job of that, even without resleeving sickness. We've always had a bit of an assortment of IC shit to say when it comes to how Medical treats things. Why not just clone? Because repeated cloning in quick succession could lead to further genetic complications. Let them die? Well death kinda sucks a bit and should be avoided and its our job to try to prevent that in the first place. Even when we got resleeving, and it was even more seamless to just _do that_ instead of using more time/resources to actually fix the patient, we were rather easily able to just say resleeving someone who's already alive is extremely disorienting and is malpractice, and that was pretty much the end of it. Resleeving sickness factors into the equation _now_, but we were _already doing this before it was added, and lengthening the timer isn't going to result in any change, either._ 

Death only needs to really 'matter' to Command in terms of crew capability, Medical in terms of ethics, and Security in terms of law. Outside of occupational obligations, it's up to players to write that for their characters, and they can have nearly whatever opinion they want on the matter. Many don't care, so their characters won't care. If you **really** want death to matter, you'll need every admin vigorously policing what people say and how they act, you'll need to make IC SOP/Regulations changes to reflect that, you'll need to hold the server as a whole to a higher RP standard in regards to how death is treated and in so doing, probably making a lot of people too uncomfortable to do vore scenes, because if exceptions are allowed, it kinda just falls apart. Then we'd likely have a substantial drop in users, most of the Staff would probably quit, and you'd probably need to start hosting it yourself. 

Resleeving Sickness never mattered in this dynamic. Death mattered on Bay years ago when cloning had, as it did on Polaris back when it had no debuffs. The only possible thing this would affect is whether or not someone is able to take part in an actiony aspect of the game if they lose the first time, which is why I understand its initial implementation and I'm fine with that, because in that regard, thirty minutes is a decent time-out and it's substantial if a crisis is still ongoing. An hour, however, just means they don't get to, because they'll most likely be killed again by whatever got them the first time. 

And, lastly, I want more action events. Judging by the fact we had around 80 or so players for a Merc event that was only given a couple weeks notice (that's not even enough time to schedule off work in many jobs), plenty of others do as well. This would interfere with that, so...Not a fan. 
/mlem

PR #9710, opened 2021-02-15
Go [Verb] Yourself!

KillianKirilenko:
One of the lesser-known features of this fork is the ability to set your own speech verbs using a command buried in the IC tab. Unfortunately this has never been a *persistent* feature, requiring that you manually set them *every round* whilst navigating through the 90s-era inputs, which puts a bit of a damper on the idea.

***No more, I say!***

This PR adds the ability to *set and save* custom say, ask, whisper, and exclaim verbs, per character. These verbs persist through cloning (so if you TF-impersonate someone, TF them into a clone of yourself somehow, or just get resleeved, you/they should keep the same verbs!) as well. There is a fairly short maximum limit on the input, to avoid memes/abuse; it can easily be increased if desired.

You'll find the settings just under traits and blood color, on the VORE tab;
image

Using the IC tab command will not set your persistent/pref-level verbs, for now, as I'm not sure how to neatly handle that. As a whole it's a bit hacky but it seems to work well enough.

:cl:
tweak - Added the ability to set custom say/whisper/shout/ask 'verbs' on a persistent, per-character basis.
/:cl:

### DIFF
--- a/code/game/dna/dna2.dm
+++ b/code/game/dna/dna2.dm
@@ -118,6 +118,10 @@ var/global/list/datum/dna/gene/dna_genes[0]
 	var/base_species = "Human"
 	var/list/species_traits = list()
 	var/blood_color = "#A10808"
+	var/custom_say
+	var/custom_ask
+	var/custom_whisper
+	var/custom_exclaim
 	// VOREStation
 
 	// New stuff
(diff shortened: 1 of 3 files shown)


MarinaGryphon:
> Using the IC tab command will not set your persistent/pref-level verbs, for now, as I'm not sure how to neatly handle that. As a whole it's a bit hacky but it seems to work well enough.

I think it should be handled like the set flavourtext verb, i.e. it doesn't set your persistent settings, only the settings for the current round.

PR #9728, opened 2021-02-16
Emergency Anti-Lag fix

Novacat:
This might be a factor in why the server has been melting down as of late. Temporarily disabling Polaris' point defense functions until further notice.


TheFurryFeline:
https://media1.tenor.com/images/2c08c9b6c2c35b7074b98f4e79cad26e/tenor.gif?itemid=6037351

PR #9748, opened 2021-02-19
Fuel Depot Tweaks

KillianKirilenko:
Some minor tweaks to make the fuel depot a bit easier to use. Whilst I can't easily fix the pipes not auto-connecting when you dock with the depot, I *can* make it easier to hook the pipes up.

- Catwalks at the end of each arm are replaced with plated catwalk spawners, which can easily be crowbarred up to access the pipes. Then loosen and secure the pipes and it should all connect properly.
- Rack with wrench & crowbar installed at each arm next to the tank ports, in case the entire crew forgot tools somehow.
- Some tables, shelves, and racks installed within the main structure with a couple of odds and ends (random cells/tools/toolboxes, plus guaranteed blue+yellow toolboxes).

Also sanitizes unnecessary var changes.

### DIFF
--- a/maps/expedition_vr/space/fueldepot.dmm
+++ b/maps/expedition_vr/space/fueldepot.dmm
@@ -360,16 +360,19 @@
 /obj/structure/cable/green{
 	icon_state = "2-4"
 	},
+/obj/machinery/atmospherics/pipe/simple/visible/fuel{
+	dir = 6
+	},
 /obj/structure/catwalk,
 /turf/simulated/shuttle/plating/airless,
 /area/tether_away/fueldepot)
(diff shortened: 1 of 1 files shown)


Novacat:
While you are editing the fuel depot, can you also add more fuel as well?

KillianKirilenko:
Easily, although two pressure tanks should be more than enough unless both Explo *and* Talon are *really* busy during the round.

Novacat:
The Soft demands more fuel!

KillianKirilenko:
There. If that isn't enough you need to reevaluate your fuel consumption.

PR #9754, opened 2021-02-19
Buffs Adranol Healing

TheFurryFeline:
🆑 
tweak - Buff Adranol healing.
/🆑 

Suggested by Tempest, and upon examination of related code, having it at eights makes reduction of peridaxon-caused eye blurriness or jitteriness reductions take long than one might like since ya can't even see how much Adranol you'd need. This'll just make it more evened out and a bit quicker.


Wickedtemp:
For clarification, it's not uncommon to see jitter/eyeblur values in the hundreds. Like, 300 sometimes. Which means 40u adranol. These changes should mean that most of the time a 15u injection clears things up.

PR #9755, opened 2021-02-20
Removes NonHelpIntent Hypo/Injector Delay

Wickedtemp:
The NonHelpIntent Resist is a mechanic that means an autoinjector and hypospray does not have its instant-inject when the patient is on Grab, Disarm, or Harm intent. It was added in early 2018, on Polaris, in an attempt to reduce chloral injectors and instant-death poisons from being used in injectors. It wasn't particularly successful in this, because antag Chemist/CMO's just switched methods to equally deadly tools like hydorophoron and acids.

It did, and still does, however, cause annoyance to Medical crew who are just trying to use injectors and hyposprays to help patients, and the patients don't know to be on help intent. They may be pulling someone else who is hurt along and be on grab, for example, or there may had been a recent fight with spiders and folks are still on harm.

Given we do not have antags outside of events, and the fact that I'm fairly sure if a doctor or chemist ran up to an event antag and chloral'd them, that'd mean a ban, I'd like to think we can scrap this change.

Soooo it's voided with a handy lil Vorestation edit note. I tested it and it seems fine. If something ends up not being fine, I can fix.

### DIFF
--- a/code/modules/reagents/reagent_containers/hypospray.dm
+++ b/code/modules/reagents/reagent_containers/hypospray.dm
@@ -55,13 +55,14 @@
 			if(!do_after(user, 30, H))
 				return
 		//VOREstation Add End
-		else if(!H.stat && !prototype) //VOREStation Edit
+	/*	else if(!H.stat && !prototype) //VOREStation Edit
 			if(H != user)
 				if(H.a_intent != I_HELP)
 					to_chat(user, "<span class='notice'>[H] is resisting your attempt to inject them with \the [src].</span>")
 					to_chat(H, "<span class='danger'> [user] is trying to inject you with \the [src]!</span>")
 					if(!do_after(user, 30, H))
 						return
+	*/ //VOREStation Edit: We don't have antags, and we don't use hyposprays or injectors as weapons, so we've no need for a delay.
 
 	do_injection(H, user)
 	return


Wickedtemp:
So, groundwork time. 

"But this gives me an opportunity to avoid injectors/hyposprays and I like that."
- The only people who have hyposprays are going to be Medical. Injectors are a little more common to get. If you do not want to be injected with anything, then Medical cannot administer reagents without breaking SOP and there are IC repercussions to this. If the person is not Medical and found or otherwise acquired an autoinjector, then there are several IC routes. Medical can be called, Security can be notified, and worst case, ahelp if it's actually grief. Keep in mind, you'd have to always be walking around on grab/disarm/harm, or switch to these intents whenever anybody came near you, so it isn't like these are actual protections. The intention of this change was to prevent autoinjectors and hyposprays from being used in combat, not to serve as a protection from unexpected pokes. The reason was that if you were attacking someone with a premade chloral injector or hypo, it was possible to whip it out and counter-jab as soon as you hit them if you were close enough. 

"What about the rare antag incidents! What if Medical or Science goes after them with injectors of poison?"
Science can print and create much more dangerous tools, and in the last merc event, made nitroglycerin grenades. They don't even need to use reagents. For Medical, it's a bit different, in that Medical crew have IC 'rules' in their SOP regarding using their skills/knowledge/equipment to cause harm. The only window they'd have is if antagonists required medical treatment, at which point a delay wouldn't save them. All Medical would need to do is say "Hey this is helpful stuff." and syringe in a bunch of hydrophoron, or "accidentally fuck up the surgery". For people worried about the previously mentioned bit about counter-injectoring an attacker, the solution is to probably ahelp. Medical shouldn't be powergaming against antags just to get a 'win'. They shouldn't even have chloral injectors made. Ground rules need to be set in the few antag rounds we have, but that should settle things.

"But is it that big of a deal? Like, is this even a problem that needs to be fixed?"
That depends on who you ask. For me, it is, because I deal with it all the time. It's annoying. If someone moves, the timer is interrupted and you have to keep trying. This is a lot more annoying when you're like..on an expedition of some kind, or there are hostile mobs around so people can't really swap intents on-the-fly. And so this is a PR to resolve that annoyance, because...that's how that works. If I can fix it, why not? And as far as I can tell, there are no reasons why this mechanic needs to exist that aren't full of holes.

"Hey, wait a minute, we've got some really dangerous injectors lying around as random loot.. I don't want someone poking me with that without having a fighting chance!"
As with dangerous tools in general, that's kinda just a result and consequence of having them be easily obtainable prefilled with bad shit. You aren't really protected because if you aren't expecting anything and you're on help intent, its GG anyway. Plenty of roles have access to incredibly dangerous stuff. Science especially. You can find dart-guns in maintenance as well. 

"What about all the other injectors! The ones in Medical and stuff!"
Autoinjectors that Medical has in bulk are limited to 5u, Cargo and Science can produce 15u Autoinjectors. The printed ones start out empty, the prefilled ones already have medicinal reagents. In both cases, you would have to produce the harmful reagents and swap them out, something that can only be done by Medical, Science, or if they really want to spend all the time ordering reagent cartridges and looking shit up on the wiki, Cargo. So lets go case by case. If Cargo does this, it's probably meta/powergaming because why does a cargo tech or QM even know how to make all this shit. Medical and Chemistry definitely would know how to make these reagents, and nobody would probably notice if they did and had a few injectors lying around. But, if they're used and it isn't for a planned IC drama or a scene, they're probably going to eat a ban, so I'd say that's covered. That leaves Science, which, admittedly, I don't know. I'm assuming Staff would get on their case about this sorta thing, but it's Science and if they happen to accidentally kill their own roboticist this way I feel like it's mostly okay because that's hilarious, but if they kill someone outside of Science then yeah that's probably an issue. Also something to keep in mind are IC consequences. Reports to Security, reports to the relevant department Heads, to the CD, to Central, etc.

Novacat:
Staff discussed this during the meeting and we have come to a decision not to merge this. A lot of it is that we decided that the helpfulness of autoinjectors/hyposprays working instantly is not worth the potential issue of instant chloral injections.

Wickedtemp:
((we still have instant chloral injections))

PR #9782, opened 2021-02-22
MkVII Helmet Upgrade

KillianKirilenko:
Adds integrated augmented reality HUDs to the ERT armor, matching the AR glasses of the same department type. Does not grant remote access to suit scanners or station alerts like AR glasses do.


Novacat:
I am on the fence on this. Mostly because this is kinda uncontrollable and the vision cannot be toggled iirc. I think a better solution might be to add AR glasses (of all kinds) to the ERT equipment manifest instead.

KillianKirilenko:
Added a verb to toggle the HUDs. You can find it under the Object tab.

PR #9783, opened 2021-02-22
Typo Fixy, Again

TheFurryFeline:
Fixes double "the" in harvest failure message.

Just how many of these are there, pfft. Note, this is a mobile web edit. Can test when home after classes out.

### DIFF
--- a/code/modules/mob/living/silicon/robot/robot_items.dm
+++ b/code/modules/mob/living/silicon/robot/robot_items.dm
@@ -171,7 +171,7 @@
 		else if(T.dead) //It's probably dead otherwise.
 			T.remove_dead(user)
 	else
-		to_chat(user, "Harvesting \a [target] is not the purpose of this tool.  The [src] is for plants being grown.")
+		to_chat(user, "Harvesting \a [target] is not the purpose of this tool.  The harvester is for plants being grown.")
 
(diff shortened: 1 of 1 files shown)


TheFurryFeline:
A'ight, tested this and it looks as if it works. ``Harvesting a floor is not the purpose of this tool.  The harvester is for plants being grown.``

code/modules/mob/living/silicon/robot/robot_items.dm
```diff
@@ -171,7 +171,7 @@
 		else if(T.dead) //It's probably dead otherwise.
 			T.remove_dead(user)
 	else
-		to_chat(user, "Harvesting \a [target] is not the purpose of this tool.  The [src] is for plants being grown.")
+		to_chat(user, "Harvesting \a [target] is not the purpose of this tool.  The harvester is for plants being grown.")
```
Novacat, line 174: This is a Polaris line, so this would probably have to go upstream.

Also, it would be better to just remove The and keep [src]

Novacat (changes requested):

code/modules/mob/living/silicon/robot/robot_items.dm
TheFurryFeline: Derp, right. I forgot that, yeah.

TheFurryFeline:
PR'd to Polaris already. :>

PR #9784, opened 2021-02-22
Fixes drone console not detecting drones outside its own z-level

Heroman3003:
title. How did it go unnoticed is the real question.


Novacat (changes requested):
Would it be possible to have this check for NTNet connectivity?

If not that, than perhaps connected Z levels?

Novacat (approved):

Novacat:
This probably should be booted up to Polaris at some point

PR #9787, opened 2021-02-23
[MIRROR] Added verbs/handling for removing and attaching your own robotic bodyparts.

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/7905
--------------------
Based on quard's verbs for Neb.

- You can detach non-vital cybernetic limbs with a model that allows it, which are attached to a parent that allows it.
- You can detach non-vital prosthetic limbs with a model that allows it from any kind of parent.
- You can only reattach cybernetic limbs to cybernetic parents that allow it. Prosthetics can reattach to any parent.
- Currently not enabled for any limb, going to revisit it tomorrow and adjust based on feedback from @ Schnayy and other mysterious lore cabal people.

~~Tested, appears to work fine - verbs update when your limbs drop off and you can't remove any more.~~ needs significant retesting due to an infestation of feeping creatures.

Additional inclusions after discussion:
- Wooden prosthetics (Morgan Trading Co) ported from Neb.
- Teshari unbranded limbs are a separate company so they can be set to pop off like prosthetics (requested). 

**Current issues:**
- Robotic limbs appear as `blood-stained` or `oil-stained` due to being given blood_DNA on init. Can't fix this without changing some core limb behavior and since it's only cosmetic I'm not sure it's worth potential regressions.
- Detached limbs have no icon in the world, and icon_state is null. I'm not clear on what could be causing this but it's unrelated to the PR so I'm leaving it alone.


KasparoVy:
That's bloody amazing!

PR #9793, opened 2021-02-24
Silky, silky webs. Adds Weaver trait, with ability to weave spiderwebs and more!

Heroman3003:
Adds new Positive trait: Weaver (costs 2 points). Allows your character to generate silk and spend it on various webby structures.

How it works: If you have weaver trait (or are a vasilissan), you have a number of verbs. You can check your silk amount, switch it's color, toggle it's production or just get weaving.

What can you make? Well, selection right now is a bit modest, but enough to decorate your abode with silky strings:
1) Weaversilk floor webs. Nothing more than a nice piece of decorum, really.
2) Weaversilk walls. Actually block passage of all objects and people... except yourself and your fellow weavers. Don't block sight though.
3) Weaversilk nest. When you want to lay down and relax... Or maybe process a meal. Or just need somewhere to tie said meal to. Almost like a bed, really.
4) Weaversilk trap. A one-use trap that will ensnare any non-weaver passing through. Not really strong, and any attempt at resisting will easily succeed, but will stop them in their tracks for a moment at least. Perhaps enough to make use of the confusion?

All of the above are structures that can easily be torn down. All it takes is one empty-handed harm click to delete any of the above-listed things. And considering that it takes time and silk to set it all up, cleaning it up is actually easier than building it.

Oh, the web bindings you could make are also still acquirable through weaving too.

Why togglable regeneration of silk? Because it takes a toll on your hunger, and a strong toll. It takes a full 200 nutrition points to go from 0 to 500 silk. Which you might not always want. It also starts off by default, so make sure to turn it on if you plan on doing silky things this round. It regenerates at an approximate rate of 1 silk per second.

Vasilissans, being natural weavers have a bit of advantage. Their maximum silk pool is twice as high and they start 50% full, rather than 20% the Custom Species weavers have.

There are some extra plans for Weave Item verb, but that would be its own PR, assuming this one even gets accepted.


KasparoVy:
Badass, nicely done!

Heroman3003:
Why is there a random emitter firing in space triggering runtimes at space carps' attempts at recognizing the threat reee

Screemonster:
I like. Having a silk pool means sensible limits on how much they can be spammed without being horribly limiting on sensible use the way nutrition was (where it was a pain to use at sensible nutrition levels but people that wanted to spam it would just chug 14 protein shakes and WEBS EVERYWHERE)

TheFurryFeline:
This is one of two old features being re-added, 🎉 Now hopefully the other one will get added 👀

Wickedtemp:
This was given to driders years ago and it was incredibly annoying, BUT ONLY BECAUSE it was a new toy that everyone wanted to use. At the time there were only probably like five regular drider characters, total, and most rounds none were even on. Webs were added so then a bunch of people made driders and it resulted in spam.

And that's okay. Because if we had just dealt with it for a few weeks then the novelty wouldve worn off and then it wouldn't be spammed everywhere anymore. 

I'm totally on board with this, and I'm pre-emptively saying that I don't agree with a premature removal just because it gets spammed for a week.

Novacat (changes requested):
- This should be worth two points.
- Walls should not block atmos

Heroman3003:
Walls already don't block atmos, the only thing they block is physical passage of people/objects (except people with weaver trait). As for the point raise, gimme a moment.

Novacat (approved):

PR #9795, opened 2021-02-25
Updates year of linoleum

Heroman3003:
Don't forget to remove it once we're past 2090


TheFurryFeline:
Gonna restate as said in lore channel, might want a loremaster to sign off on this. Might seem very minor but when it's to do with timey wimey stuff, need that good lore

PR #9751, opened 2021-02-19
[MIRROR] Diet/Allergy System Rework

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/7864
--------------------
The fun thing about being out of action for a couple of weeks is having Ideas. Namely, I was seized by an urge to tweak the dietary system. Originally this was going to be by setting the food groups a species can eat safely to do omnivore/carnivore/herbivore/_whatever_ stuff, but after an afternoon of wrestling with bitwise operators I instead flipped the whole thing on its head and made it an allergen system instead.

In short, species now have a bitflag var of things they are allergic to, and food and drink reagents have a corresponding bitflag var. Species also have a reaction strength value. If a bloodstream or ingested reagent's allergen flags match any on the species' var, *boom*: bad times ensue. With some wrangling this should make it quite possible to do things like add allergy traits, if you want a character who's ***super*** allergic to nuts or soy or something.

This partially supplants the old allergen system for skrell/tajara, but doesn't build it out too far. Damage amounts should be roughly the same as before, give or take a little bit. Alcoholic and mixed drinks with allergens can be made less deadly to loosely simulate reduction/dilution of the allergenic substances as well.

:cl:
tweak - changed some underlying mechanics for how allergies work
tweak - soy protein (e.g. tofu) is now safe for skrell
tweak - a couple more dairy and coffee based mixed drinks are unsafe for skrell and tajara
/:cl:

PR #9804, opened 2021-02-26
Improved Trait Exclusion Logic v2

KillianKirilenko:
Adds a listvar to traits which can be used to exclude the trait from specific core species.

Includes three commented-out allergy traits that won't work until/unless #9751 is merged.


KasparoVy:
Good thinking, cheers for keeping ontop of this

KillianKirilenko:
> Good thinking, cheers for keeping ontop of this

It was one of those moments when you realize that something really obvious is missing and you wonder how it got overlooked. Surprisingly easy to set up too.

PR #9845, opened 2021-03-03
Add %count and %countprey string replacers in belly messages

Arokha:
%count is replaced with how many things (items, prey, anything) are in that belly
%countprey is replaced with how many mobs are in that belly

### DIFF
--- a/code/modules/vore/eating/belly_obj_vr.dm
+++ b/code/modules/vore/eating/belly_obj_vr.dm
@@ -361,9 +361,15 @@
 		var/raw_message = pick(examine_messages)
 		var/total_bulge = 0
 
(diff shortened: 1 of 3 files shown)


code/modules/vore/eating/belly_obj_vr.dm
```diff
@@ -582,13 +588,21 @@
 	var/struggle_outer_message = pick(struggle_messages_outside)
 	var/struggle_user_message = pick(struggle_messages_inside)
 
+	var/living_count = 0
+	for(var/mob/living/L in contents)
+		living_count++
```
Arokha, line 593: Can anyone think of a better way

Novacat: I cannot.

PR #9850, opened 2021-03-04
Make allergen traits make proper use of bitflags and add the rest of the allergens as traits

cadyn:
So, with a few lines and some pathing, allergen traits should add that allergen's bitflag instead of setting the variable equal to the allergen itself, this is much better for many reasons. Added the rest of the allergens as well. I might go and try to update a bunch of stuff polaris-side to make the allergen stuff more accurate as well, especially since like none of the foods that should trigger allergies right now actually do.

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/traits_vr/neutral.dm
+++ b/code/modules/mob/living/carbon/human/species/station/traits_vr/neutral.dm
@@ -155,30 +155,30 @@
 
 
 //Allergen traits! Not available to any species with a base allergens var.
-/datum/trait/allergy_gluten
+/datum/trait/allergy
 	name = "Allergy: Gluten"
 	desc = "You're highly allergic to gluten proteins, which are found in most common grains."
 	cost = 0
 	custom_only = FALSE
-	var_changes = list("allergens" = 16)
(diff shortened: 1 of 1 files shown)


Novacat (approved):

KillianKirilenko:
The reason I did it this way originally is because *last* time I tried to edit something bitflag based with traits, it made dreamseeker crash any time you tried to view-vars on a mob with the trait. That was fun.

As for non-triggers; be careful with testing properly-cooked foods vs adminspawned foods. IIRC the latter won't be inheriting reagents from the ingredients used to make it, but if you cook them *properly* they will. Or should. I think.

cadyn:
> The reason I did it this way originally is because _last_ time I tried to edit something bitflag based with traits, it made dreamseeker crash any time you tried to view-vars on a mob with the trait. That was fun.
> 
> As for non-triggers; be careful with testing properly-cooked foods vs adminspawned foods. IIRC the latter won't be inheriting reagents from the ingredients used to make it, but if you cook them _properly_ they will. Or should. I think.

Just tested, and no, it still happens with cooked foods, so I'm going through with my polaris shenanigans, I've got a pretty good idea to fix it.

PR #9854, opened 2021-03-05
Makes drag-drop bodyjacking not break aghost bonds

Verkister:
Possessing mobs by drag-dropping your admin ghost into the mob now transfers your original body's spirit link into the possessed mob and thus prevents your original body from getting knocked out when your ghost disappears into the new body.


Novacat (changes requested):
Either needs a vorestation edit or sent to polaris.

PR #9862, opened 2021-03-06
Restricts some size-mechanics with prefs

Heroman3003:
Makes step mechanics entirely an opt-in pref (~~off~~ on by default) - available in vore panel, saves
Makes being picked up/picking up others a combined opt-out pref (on by default) - available in vore panel, saves
Makes picking up require target be on Help intent
Adds a toggle unrelated to the picking up pref that allows you to not pick up others even if all conditions are met (for micro petting) - in IC tab, does not save


ShadowLarkens:
I'd prefer the step preference is on by default, entirely because I feel like it's going to be really annoying for people who like them.
It makes previously unrestricted mechanics now require digging through the vore panel and turning them on for each character slot, which is annoying in the same way that drop noms being per-character and off by default are.

Heroman3003:
> I'd prefer the step preference is on by default, entirely because I feel like it's going to be really annoying for people who like them.
> It makes previously unrestricted mechanics now require digging through the vore panel and turning them on for each character slot, which is annoying in the same way that drop noms being per-character and off by default are.

To respond my thoughts, I think the issue of people getting accidentally stepped on by careless macros or macros stepping on people when they don't want is far more prevalent than actual use of step mechanics. Not to mention combat possibilities regarding instant downing of disarm or huge damage output of harm. I feel its for the best to keep it more of an opt-in thing for those reasons, as being a kink thing it shouldn't be too prevalent in robusting and combat in the first place.

Very-Soft:
I think the togglable settings are a good idea, however, they should start the way that things are presently implemented, that is to say, steps should start on, and be able to be toggled off. 

I think that someone who doesn't want steps to happen is a lot more likely to seek out the preference than someone who doesn't know or doesn't much mind one way or the other. This completely removes the possibility for surprise in that regard, and is going to invariably require everyone who is into this kind of thing to go begging everyone else who didn't know about the setting to turn it on.

I would prefer that things work the way they have done, and anyone who wants to opt out can, rather than the default setting being changed according to your preference on the matter.

Heroman3003:
> 
> 
> I think the togglable settings are a good idea, however, they should start the way that things are presently implemented, that is to say, steps should start on, and be able to be toggled off.
> 
> I think that someone who doesn't want steps to happen is a lot more likely to seek out the preference than someone who doesn't know or doesn't much mind one way or the other. This completely removes the possibility for surprise in that regard, and is going to invariably require everyone who is into this kind of thing to go begging everyone else who didn't know about the setting to turn it on.
> 
> I would prefer that things work the way they have done, and anyone who wants to opt out can, rather than the default setting being changed according to your preference on the matter.

Since more than one headmin supports the notion, I will concede and change the default value approprietly. However, I'd like to clarify that this isn't about my own preference on the matter and trying to push it onto others, as my own preference would be having both permanently on. It's more of an attempt to make the step mechanics less of a constant balance roadblock in trying to implement more sizechange features, since it always has to be taken into consideration as combat advantage balance-wise..

lm40:
THANK GOD THIS IS A GREAT CHANGE I LOVE IT WITH ALL MY HEART AND SOUL

Screemonster:
Step-grab mechanics are fine but would be vastly improved if resisting while on harm intent, instead of resisting out, resulted in getting a free attack on whatever part of their body they're using to grab you, using whatever you have to hand.
Maybe require a sharp/edge melee weapon, a sharp/edge unarmed attack, or the brutal predation trait so that when someone decides to just go "hee hee a smol I will put my paws on them" the smol in question gets to make it a bad idea

heroman put this in dev chat when discussing the concept and it's pretty much exactly what I have in mind
image

PR #9863, opened 2021-03-06
Removes advantage macrotaurs have with saddlebag slowdown.

Heroman3003:
I always thought it was weird and obscure that specifically saddlebags for specifically macrotaurs do not apply slowdown while they do for non-macrotaurs. FAIRNESS!


Novacat:
That being said Slowdown does need some balance changes but that is not here or there.

PR #9865, opened 2021-03-07
Adds HYPER suit

Heroman3003:
~~Changes bluespace suit. No longer allows for sizechanges, but still allows for bulge hiding. Features new sprite, blue instead of green. Is much cheaper and easier to make in R&D, also available in loadout.~~

~~Adds redspace suit. Functions exactly like old bluespace suit did. Sprite is new too, with red instead of blue. Has same stats as before, is makeable in R&D same way as before, basically old bluespace jumpsuit with new coat of paint.~~

~~Adds both of them into trash pile pool as well.~~

Adds Heavy Yield Protrusion Erasing and Retracting jumpsuit. Available in loadout, trash piles and is cheaply made in science. Has similar property in ability to toggle hiding of bulges/weight as Bluespace Jumpsuit, but without ability to sizeshift.

Also adds Bluespace Jumpsuit to trash pile pool and gives it a slightly updated sprite.


Wickedtemp:
Dunno if there's really much here to be discussed, first glance says all of this is just small fun stuff

PR #9881, opened 2021-03-09
Carbonkin Content port + Trait whitelisting

Sharkmare:
[Ports over my carbonkin content update that brings carbon kin mostly on the same level as mob kin.
Also ports the whitelist system so people don't have to write out EVERY species name in cases like this.](https://github.com/CHOMPStation2/CHOMPStation2/pull/1499)

This update mostly lays the groundwork for future work on these critters.

What's it add tho?
* Rebalances phase shifting:
Phase shifting is a pretty polarizing ability, it now is more forgiving for stealth uses but has a counter in becoming more costly with people and light level.

* Stat blocks of all kin have been revised based on the flavor messages of the mob variants.

* Tasers and other electricity sources now affect kin, this is the only immunity being removed and should help with SEC interaction.

* Adds a new attack "syhon strike" this is mostly for the aggressive kins, if using syphon strike on hit it will take the damage you dealt and convert it into energy, this only works against carbons as example punching punpun restores around 4 energy

* As a last note, all of the Shadekin Adaption color traits lock your eye color into the respective color. So if you take Red adaption you will be immediately identifiable as red adapted kin since your eyes will in fact always be red.


code/modules/client/preference_setup/vore/07_traits.dm
```diff
@@ -342,7 +342,11 @@
 			if(pref.species in instance.banned_species)
 				alert("The trait you've selected cannot be taken by the species you've chosen!","Error")
 				return TOPIC_REFRESH
-
+			//VORSTATION ADDITION CHOMPADDITION START Adding whitelist for traits
```
Novacat, line 345: This is a Vorestation file, so the markings are not necessary.

code/modules/mob/living/carbon/human/species/shadekin/shadekin_abilities.dm
```diff
@@ -33,6 +33,26 @@
 
 	var/ability_cost = 100
 
+//CHOMPADDITION Conditional shifting ased on observers and darkness
```
Novacat, line 36: Same here.

code/modules/mob/living/carbon/human/species/shadekin/shadekin.dm
```diff
@@ -21,7 +21,7 @@
 	unarmed_types = list(/datum/unarmed_attack/stomp, /datum/unarmed_attack/kick, /datum/unarmed_attack/claws/shadekin, /datum/unarmed_attack/bite/sharp/shadekin)
 	rarity_value = 15	//INTERDIMENSIONAL FLUFFERS
 
-	siemens_coefficient = 0
+	siemens_coefficient = 1
```
Novacat, line 24: I vaguely remember talking to Soft about this in the past and I *think* it was intentional that they had a siemens coefficient of zero. That being said, I will just mark this and wait for their comment before requesting any change here.

Very-Soft:
Me like, the main reason we didn't really go for the phase shift counter play originally was because I was told it wasn't reasonable/possible if I remember right, and the reason we dropped the stats differences between colors was because at the time, it seemed like we would have had to have made a different species for every kind of 'kin color we wanted.

This brings things more in line with the way that 'kin are supposed to work in lore, and also gives station crew counter play options for catching 'kin, that were kinda always supposed to be there. 

ME BIG APPROVE.

Novacat (changes requested):
Mostly seems fine, just that the comments should be removed since they are vorestation-specific files. I contacted VerySoft to give a glance over the details of this as well.

code/modules/mob/living/carbon/human/species/shadekin/shadekin.dm
Very-Soft: This was done originally to match the way that simplemobs were done at the time. (and because lorish wise, they would probably be able to absorb the energy) but it means that there's not really much that can be done to get them without lethals. 

So I'm fine with this change for making them reasonable to try to catch if they're careless.

Heroman3003:
I kinda sorta don't like the fact that inbelly prey contributes to the counter of viewers for phase shift cost purposes. Seems to be pretty anti-vorey as just having two people prevents phasing in well-lit conditions and having three prevents phasing entirely. Unless kin is digesty, snatching multiple folks becomes an impossibility.

Very-Soft:
> I kinda sorta don't like the fact that inbelly prey contributes to the counter of viewers for phase shift cost purposes. Seems to be pretty anti-vorey as just having two people prevents phasing in well-lit conditions and having three prevents phasing entirely. Unless kin is digesty, snatching multiple folks becomes an impossibility.

I asked about it and they said that they didn't think it would count people in the tum. If it does, it shouldn't.

code/modules/mob/living/carbon/human/species/shadekin/shadekin_trait.dm
```diff
@@ -0,0 +1,89 @@
+/datum/trait/kintype
+	allowed_species = list(SPECIES_SHADEKIN)
+	var/color = BLUE_EYES
+	name = "Shadekin Blue Adaptation"
+	desc = "Makes your shadekin adapted as a Blue eyed kin! This gives you decreased energy regeneration in darkness, decreased regeneration in the light amd unchanged health!"
+	cost = 0
+	var_changes = list(	"total_health" = 100,
+						"energy_light" = 0.5,
+						"energy_dark" = 0.5,
+						"unarmed_types" = list(/datum/unarmed_attack/stomp, /datum/unarmed_attack/kick,/datum/unarmed_attack/shadekinharmbap))
```
Heroman3003, line 10: Using any of the traits removes bite/claw attack options, replacing them entirely with harmbap. Is that intentional? I know shadekin had special variations of those for energy gain, but it makes sense to have regular left in at least as an option for fluff. Although, admittedly, I don't get the purpose of replacing either with new separate attack in the first place, especially if said attack doesn't contribute to replication of oldschool "different kins gain different levels of energy via attacks" thing that simplemob kin had.

code/modules/mob/living/carbon/human/species/shadekin/shadekin_abilities.dm
```diff
@@ -33,6 +33,26 @@
...
+	var/turf/T = get_turf(src)
+	if(!T)
+		to_chat(src,"<span class='warning'>You can't use that here!</span>")
+		return FALSE
+
+	var/brightness = T.get_lumcount() //Brightness in 0.0 to 1.0
+	darkness = 1-brightness //Invert
+
+	var/watcher =-1
+	for(var/mob/living/carbon/human/watchers in view_or_range(7,get_turf(src),"range" ))
```
Heroman3003, line 47: This does not seem to be at all what's advertised. It does not care about whether the mob in question can actually see the shadekin, it simply grabs all mobs within the 7 tile range, even ones behind walls or ones incapable of otherwise seeing the kin. That and if I'm reading the code for range() correctly, then it absolutely does include the contents of the 'src'. Ideally, the solution would be to instead use viewers() centered on the kin with same range, and manually exclude any bellied people (ideally, ones bellied by other observers too, imo).

Heroman3003: Also, it should ideally at least check for whether 'watcher' actually has a player mind or not. A shadekin getting stuck in a monkey pen somewhere would be as hilarious as it would be disappointing.

code/modules/mob/living/carbon/human/species/shadekin/shadekin_trait.dm
Sharkmare: The custom attack for kin is flavored as a claw attack but without causing bleed.

code/modules/mob/living/carbon/human/species/shadekin/shadekin_abilities.dm
Sharkmare: > This does not seem to be at all what's advertised. It does not care about whether the mob in question can actually see the shadekin, it simply grabs all mobs within the 7 tile range, even ones behind walls or ones incapable of otherwise seeing the kin. That and if I'm reading the code for range() correctly, then it absolutely does include the contents of the 'src'. Ideally, the solution would be to instead use viewers() centered on the kin with same range, and manually exclude any bellied people (ideally, ones bellied by other observers too, imo).

I don't know how to do such and encourage you to make a PR to my branch to add this change to the watcher system.

tigercat2000: Try `for(var/mob/living/carbon/human/watcher in viewers(7, src))`

ShadowLarkens: Try using `for(var/mob/living/carbon/human/watches in viewers(7, src))`

code/modules/mob/living/carbon/human/species/shadekin/shadekin_abilities.dm
```diff
@@ -33,6 +33,25 @@
...
+		watcher++
```
Heroman3003, line 47:
```suggestion
	for(var/mob/living/carbon/human/watchers in oveiw(7,src,"view" ))	// If we can see them...
		if(watchers in oviewers(7,src))	// And they can see us...
			if(!(watchers.stat) && !isbelly(watchers.loc) && !istype(/obj/item/weapon/holder, watchers.loc))	// And they are alive and not being held by someone...
				watcher++	// They are watching us!
```

Novacat (approved):
Code-wise seems fine now, going to poke Soft about giving approval.

PR #9883, opened 2021-03-09
Adds ability to set up and edit idle bellymode messages

Heroman3003:
Simplemobs had those going for a while so why the heck not. Some people have literally transfered the messages over from mob bellies just for that aparently too.

Those have higher text limit of 500 characters too.

And while I'm there, fixes examine text not being reset on choosing the reset option for belly messages.

Iuono im sleepy and not sure I tested it thoroughly enough but it saves, loads, displays messages and doesn't runtime in obvious  ways, so. Hopefully it stays that way. Gnite.


TheFurryFeline:
Bluetext messages~

PR #9906, opened 2021-03-10
Remove weird helmet thing from Forensic Technician outfit

Little-119:
It's weird and unfashionable and seems... out of place (I'm surprised to find that it's a Virgo thing)? As far as I am aware it doesn't serve any function. No one I've asked (read: one person I've asked IC) knows why it exists. And I am mildly inconvenienced by having to remove it from my head every time I join as Forensic Technician.

Git blame seems to say that @Arokha added this last Feburary, and I do wonder: Why, foxxo?


TheFurryFeline:
Note, this is touching a Polaris file.

PR #9944, opened 2021-03-14
Fix Missing DSI Roboparts

KasparoVy:
Comment the whole lot out because the ones that weren't had imposed restrictions.
You can select all DSI limbs again.
image


Novacat (changes requested):
I would rather, instead of the polaris ones being commented out entirely, just comment out the offending vars and delete the duplicate vorestation code.

KasparoVy:
> 
> 
> I would rather, instead of the polaris ones being commented out entirely, just comment out the offending vars and delete the duplicate vorestation code.

Done, cheers--

Full body list showing options available as intended (character species is Unathi): 
image

Head list showing zeng-hu taj (the other eggnerd head is just above zeng-hu taj):
image

Teshari list working fine:
image

Novacat (approved):

PR #9947, opened 2021-03-15
Remove the throwing-people-to-the-floor effect of meteor strikes

Little-119:
Personally, I think this is kinda bad for RP. By that, I mean I dislike being randomly interrupted in the middle of my roleplaying just to be thrown to the floor for a few seconds, prompting one to either A) Ignore the event entirely and proceed as if it hasn't happened or B) stop what I was doing to give the predictable and unmeaningful responses of saying something along the lines of "That sounded bad" and/or asking others if they're okay (which they always are)

I dunno, it's not really a big deal and the above paragraph is probably an overreaction. But this thing bothered me enough to go through the (minor) effort of submitting a PR and that's why.

### DIFF
--- a/code/modules/events/meteor_strike_vr.dm
+++ b/code/modules/events/meteor_strike_vr.dm
@@ -61,10 +61,6 @@
 			var/turf/mob_turf = get_turf(L)
 			if(!mob_turf || !(mob_turf.z in impacted.expected_z_levels))
 				continue
-			if(!L.buckled && !issilicon(L))
-				if(!L.Check_Shoegrip())
-					L.throw_at(get_step_rand(L),1,5)
-				L.Weaken(5)
 			if(L.client)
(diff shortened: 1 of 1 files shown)


TheFurryFeline:
This code looks jank, too.... Like, why not have it do Weaken if people are a certain range away from where a meteor strikes? :V

Wickedtemp:
I'm kinda neutral, I don't care if this gets removed or not. But! Since others would like it be removed because its annoying for them, I can get behind that. 

It doesn't really serve a purpose anyway, it isn't like this particular feature is essential to meteoroids.

lm40:
yea it is just an annoyance, tbh we are better off without it

Heroman3003:
Yeah, considering the actual nature of meteroid event, the global surface throwdown is way overdoing it.

JoanRisu:
It's annoying for a random event to do, so this one can pass imo

PR #9949, opened 2021-03-16
Adds a bit more connection to fish farm maintenance

Heroman3003:
A tiny map change. Plus fixes a pair of floor tiles elsewhere under maint doors because contrast of pristine white with maint dark was way too much.


Heroman3003:
DO NOT MERGE. I misclicked and theres a tile of water placed randomly in one of the power stations lol. I will fix it when I'm back home.

PR #9969, opened 2021-03-18
Adds in Booze hounds.

lukevale:
Adds in the booze hounds! Big boyos that are here to serve drinks, and be a dunk tank. They're effectively service hounds with slightly less, but are all sprited and animated. A new addition for those who finds them interesting.


Haery70:
Something tells me that this is not going to end well for this.

Heroman3003:
The holy vodka cumborg has arrived

code/modules/mob/living/silicon/robot/robot_modules/station_vr.dm
```diff
@@ -812,3 +813,81 @@
...
+	var/datum/reagents/N = new/datum/reagents(50)
+	src.emag.reagents = N
+	N.my_atom = src.emag
+	N.add_reagent("beer2", 50)
+	src.emag.name = "Mickey Finn's Special Brew"
+
+
+	R.icon 		 = 'icons/mob/widerobot_sec_vr.dmi'
+	R.wideborg_dept = 'icons/mob/widerobot_sec_vr.dmi'
+	R.hands.icon = 'icons/mob/screen1_robot_vr.dmi'
```
Verkister, line 874: Got some pasta leftovers here

Novacat:
Aside from what Verk pointed out, I think this is fine, but going to want to double-check it by Staff before merging.

Very-Soft:
I'm not crazy about them but I also don't have any good reason to say no! So!

code/modules/mob/living/silicon/robot/robot_modules/station_vr.dm
```diff
@@ -812,3 +813,81 @@
...
 	R.verbs -= /mob/living/silicon/robot/proc/rest_style
 	..()
+
+
+/obj/item/weapon/robot_module/robot/booze
+	name = "BoozeHound robot module"
+	channels = list("Service" = 1)
+	pto_type = PTO_CIVILIAN
+	can_be_pushed = 0
+	sprites = list(
```
Poojawa, line 823: This would be 9000% better as a greyscale sprite that gets recolored, or to simply have a full custom RGB option entirely. since it's all the same zones and bot entirely.

Screemonster: yeah, we already... have code for deciding what colour a reagent should be when displayed in a transparent container

Huenererschrecker:
I would rather not see pissborgs.

Andromeda-K22:
Pissborgs are a bad idea.

Poojawa:
Don't kinkshame people, smh

Andromeda-K22:
my kink is kinkshaming,

dogborgs are bad enough

code/game/machinery/painter_vr.dm
```diff
@@ -17,7 +17,8 @@
 	var/list/allowed_types = list(
 			/obj/item/clothing,
 			/obj/item/weapon/storage/backpack,
-			/obj/item/weapon/storage/belt
+			/obj/item/weapon/storage/belt,
+			/obj/item/device/radio/headset
```
Arokha, line 21: How is this related?

JoanRisu:
I will reiterate Pooj's point not to kinkshame. 

As Soft said, we really have no real reason to turn this one down. As long as it is moderate and not so blatantly shamelessly out in the open, people can do what they want in line with the area-by-area guidelines and bystander policy.

Iroquois-Pliskin:
Not having a reason to say no isn't a reason to say yes. Let's not add this.

Haery70:
> Will it have just bartender access, or all access?
All borgs have all access regardless of what chassis they're using, IIRC.

Honestly, the more I think about this borg the more I think of it as a dogborg with milk implants which we got rid of in favor of the Feeder trait and even when it was still around it was whitelisted. Unfortunately, I wouldn't be in favor of this even if you were to bring whitelists, area-by-area response, and bystander consent to the table.

lm40:
this isnt even a fetish thing, other than the meme is there any genuine reason to think that yellow liquid = piss is going to become a widespread public phenomenon?? i mean people don't see pints of drinks sitting on the bar counter and go on about how the bartender is serving up glasses of fresh piss so am i missing something that everyone else knows????

code/game/machinery/painter_vr.dm
lukevale: It's not, I actually have no clue how that ended up being changed.

code/modules/mob/living/silicon/robot/robot_modules/station_vr.dm
lukevale: That's something to do with the original design and I'd have no clue how to set it up.

Poojawa:
> 
> 
> this isnt even a fetish thing, other than the meme is there any genuine reason to think that yellow liquid = piss is going to become a widespread public phenomenon?? i mean people don't see pints of drinks sitting on the bar counter and go on about how the bartender is serving up glasses of fresh piss so am i missing something that everyone else knows????

I mean, depending on the brand of beer it might as well be piss.

LSommers:
Wow, these are genuinely some of the worst sprites I've seen in a while.
And yet another example of "nobody saying no means yes!"

Also lmao piss :^)

PR #9977, opened 2021-03-20
Ports tentacle tails.

SpaceVampire:
I took the sprites from Cit-Code.
I would give more credit then that, but they seemed to have been there for a long time.
All I really did was figure out how to put them in the proper format by slamming some files together, and then I had to change the colors from the Red/Green scale to Grey Scale so you could color em correctly.
Anyway, I tested it locally and everything seems to work and look nice.
Here is a picture of it. Hopefully everything is in order.
DoctorSlime


Tk420634:
Super old pr, but I wanted to post because it's good to see my old work from cita 5th port still somehow existing.  <3  Shoulda pulled it from there, it would have been colored properly for y'all.

PR #9978, opened 2021-03-20
Adds Xenochimera variants of some positive and negative traits

Tenryumi:
This creates neutral versions of some positive/negative traits that Xenochimera can take, so that they can enjoy some of their benefits as well:

- Weaver
- Hard Feet
- Sharp Melee + Numbing Fangs
- Snow Walker
- Neural Hypersensitivity

I was given the OK by Scree to add Weaver. However, I added the other ones as I think they fall under the same justification as for Weaver: Xenochimera are more then capable of developing these adaptations on their own.

Oh, and Neural Hypersensitivity is a horrible, horrible idea to take on 'chimera, but I added it anyway in the name of !FUN!

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/xenochimera_trait_vr.dm
+++ b/code/modules/mob/living/carbon/human/species/station/xenochimera_trait_vr.dm
@@ -26,10 +26,10 @@
 	cost = 0
 	custom_only = FALSE
 
-/datum/trait/fangs/xenochimera
+/datum/trait/melee_attack_fangs/xenochimera
 	allowed_species = list(SPECIES_XENOCHIMERA)
-	name = "Xenochimera: Numbing Fangs"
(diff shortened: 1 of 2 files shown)


Screemonster:
I think xenochimera should already have sharp melee/bite as their regular attack unless traits are overriding. No reason why they shouldn't be able to take numbing bites though.

Novacat (changes requested):
What Scree suggested

Tenryumi:
> 
> 
> I think xenochimera should already have sharp melee/bite as their regular attack unless traits are overriding. No reason why they shouldn't be able to take numbing bites though.

Ah, sure!

Tenryumi:
The traits do override all attack types, so all we'll need is Sharp Malee + Numbing Fangs.

Novacat (approved):

PR #9981, opened 2021-03-21
[WIP] NSB Neptune

dwinters99:
I was in bed one night, thinking: "I want to make a map." 
then I got up (at 0245) and drew this
image
things are different now (as they often are)
I'll probably finish this
If I _am_ going to finish it, I'll need feedback, so I might-as-well put up a draft PR to have a place to put in-progress stuff in case something gets buried in the discord.

`¯\_(ツ)_/¯`

### Things done so far:
_3/15/21-3/21/21_
-  science is almost done, disposals are set up
- Major innovation: trash bins (not disposals) come with a trash-bag so you can bag the garbage in the bin into real disposals or hand it to a jani
- Major innovation 2 (not implemented yet): RND has mining flaps so you can push a crate into RnD without playing musical doors
- Xenobio isn't made yet because it's going to be on a diff z-level (👀)
- the 'area' still has an issue or two because I just copy-pasted it from elsewhere while I was testing
- Toxins/xenoarch ALSO isn't made because it's going to be on the same level as mining (coming Q3 2961)
- exporation is connected to science to make interaction workflow infinitely faster
- Gun locker in the PF's office as opposed to the shuttle so PF can hand out guns at the service desk in his office (iffy on this, will probably change)
- xenobot doesn't have the weird atmos station because a RD main told me nobody uses it, will probably add when I learn what it is for and how it works
- new outdoors firing range for small explosion testing and scoped weapons
- a bathroom (why doesn't the tether have more bathrooms??)
- more things I probably forgot but you can see in this screenshot:
image


foopwo:
>Major innovation 2 (not implemented yet): RND has mining flaps so you can push a crate into RnD without playing musical doors
Mind that flaps are trivial to bypass, anybody can crawl right through them. I recommend a disposal chute behind a cargo-access windoor.

dwinters99:
Minor file loss, so expect a delay before the next update.

dwinters99:
Minor file loss has rapidly evolved into major file loss so I'm closing this until I make significant progress again

PR #10007, opened 2021-03-24
Fixes Fluff and ITG ID cards, refactors event/fluff IDs a lil', also fixes for digested id icons

Heroman3003:
Adds event ID subtype, that works similarly to how Amaya's/Harmony's fluff IDs worked, allowing to copy character info onto them freely. Includes a subtype that implies a preset access and variable that implies preset job title.

EDIT: This PR grew out of control but here we are. Also, big thanks to VerySoft for digested overlay for new ID template


Very-Soft:
Thank you for help<3

Novacat (changes requested):
This is going to conflict with the mirror atm, can you fix this to not conflict with the mirror?

PR #10012, opened 2021-03-24
using rutile in alloys no longer makes slag

Hatterhat:
fun fact: if you don't set alloy = 1 in ore, and you use it in alloying, it uses extra bits of it to make

drumroll please

slag

this sucks and i hate it so here's a fix, probably
(fun fact: in ore_datum_vr.dm, alloy = 1 was set, but i guess someone on polaris forgot it?)


Novacat (changes requested):
Please send to Polaris.

Novacat (approved):

PR #10028, opened 2021-03-27
Makes the event HUD work

Meghan-Rossi:
*Finishes up the partially-implemented event HUD (The one the event pref in the VORE character setup tab is for) and adds a verb to let event managers give it to people (can be used by right-clicking them, or from the "Fun" verb tab).
Tested.

### DIFF
--- a/code/modules/admin/admin_verb_lists_vr.dm
+++ b/code/modules/admin/admin_verb_lists_vr.dm
@@ -393,6 +393,7 @@ var/list/admin_verbs_mod = list(
 )
 
 var/list/admin_verbs_event_manager = list(
+	/client/proc/toggle_vantag_hud,
 	/client/proc/cmd_event_say,
 	/client/proc/cmd_admin_pm_context,
 	/client/proc/cmd_admin_pm_panel,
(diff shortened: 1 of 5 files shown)


code/modules/mob/living/carbon/human/human_helpers.dm
```diff
@@ -190,6 +190,10 @@
 	if(nif)
 		compiled_vis |= nif.planes_visible()
 	//VOREStation Add End
+	//VOREStation Add - event hud
```
TheFurryFeline, line 193: Since an Add End is already before this line, might just remove it and leave the trailing end tag at the bottom. Minor nitpick, yes, but lesser number of lines as a whole. Can probably just leave this here at the suggested line.
```suggestion
	//event hud
```

code/modules/admin/admin_verb_lists_vr.dm
```diff
@@ -393,6 +393,7 @@ var/list/admin_verbs_mod = list(
 )
 
 var/list/admin_verbs_event_manager = list(
+	/client/proc/toggle_vantag_hud,
```
Novacat, line 396: I would recommend adding this to admins as well since none of the Vorestation staff roles use the event role.

Novacat (changes requested):

Novacat (approved):

PR #10033, opened 2021-03-29
Trait Rebalance and Refactor

Novacat:
- New typepaths added for Positive, Neutral, and Negative Traits
- Colorblindness traits moved to Neutral traits, no longer refund points
- Custom species now start with 2 trait points
- Major Conductivity only refunds 1 point now
- Major Hardy only reduces item slowdown by 75%, costs 3 points
- Hardy cost upped to 2 points
- Reduced some extremely high/low values in some traits (flash_mod is the most notorious offender)
- Hard Feet and Flight reduced to 0 points


Very-Soft:
You should list out the things that you're changing, it's a little hard to read and look at, and some of the things are pretty silly, and at risk of sounding rude, looking at this seems like needless meddling just because.

Like just off the top I can see it looks like you made weakling slow you down less, but also made it give more negative points.

I also don't really agree with bringing the positives and negatives closer to normal. None of them get all THAT far from normal in any kind of especially useful manner. Traits haven't even come up in terms of balancing things in more than a year. I think as is you can get up to 25 extra hitpoints, 25% and 15% resistance in your choice of brute and burn if you really pushed it.

In hostile places these traits are going to equate to three or four extra hits from a lot of the hostile mobs. 

You also made the equivalent positive trait for hardy more expensive than the negative version. 

I'm not crazy about these changes as they are, and think they ought to all be listed out plainly in an easy to parse format so we can see it better and discuss it.

Wickedtemp:
In terms of gameplay, I don't feel like I noticed much of a change from having Major Hardy and not. Same for Haste. The only traits that seem to actually do anything are ones that give you extra abilities, like the antiseptic mlems and glowing. Overall, outside of antiseptic mlems, I don't really think the traits system matters much as it is, and with decreased values and emphasis on staying as close to moderate as possible, it'll matter even less.

Novacat:
Major Hardy's biggest impact is if you use a lot of items with slowdown, such as dufflebags and voidsuits.

dwinters99:
My major complaint with traits has always been that they don't change enough. A custom species doesn't really feel much different than a default human as-is. This seems to make them change even less and so I'm not amazingly happy to see this. 
Furthermore, this seems to exist to discourage people using the trait system to negate the downfalls of the core gameplay. Virgo is already painfully slow (presumably because of the tickrate?) so forcing people to be slower on average will not go over well. 

Also, if the goal of this is to prevent minmaxing, it has failed horribly. I already find myself running out of points before running out of traits so making positives cost 0 will just allow more beneficial traits while not taking the niche/unneeded traits like winged flight. In fact, I don't think I've ever seen any of the 0-cost positive traits be used outside of scenes where mechanical benefits are unneeded because of the ability to roleplay. Weaver and Winged flight are the only ones I think are applicable to 'balance' and winged flight is almost entirely fluff because of the lack of meaningful places to use it.

I suppose this will benefit downstreams? I'm not sure, though, because the attitude of downstreams is firmly speed-based and they will tolerate speed reduction even less. I don't know what this is aiming to do other than resolve a perceived imbalance that doesn't actually exist because of how little traits change gameplay already.

I might just be cynical but traits for the most part exist for solely gameplay so there is no reason to balance so strictly on a low-gameplay server such has virgo. This feels more like a Bay PR to me. Don't remove space-lube, please.

Haery70:
This feels more like an encouragement towards min/maxing over a discouragement, if you ask me. The majority of the negative traits have been removed and many of the positive traits that actually has advantages are basically very close to neutral traits.

Novacat:
Can you give specific examples? The ones I reduced to 0 points are mostly, in my eyes, either scene traits or so situational that just taking up a trait slot is cost enough, IMO.

Very-Soft:
> Major Hardy's biggest impact is if you use a lot of items with slowdown, such as dufflebags and voidsuits. That being said, one of the main things I have been aiming for is to make custom species traits mostly flavor instead of mechanics.

If that's what you wanted to do, you should just try to do that, rather than sneaking it in gradually over time.

I do not understand the purpose of this. If you want to remove the mechanical differences and leave only flavor, then suggest removing custom species, because we can already do that with the normal species being able to take neutral traits. 

I don't agree with that prospect though. I honestly think custom species is in a fine place. Most of the EXTREMES are in the negative traits, and if people want to move super slow or get punched in the face one time and die, then I don't really see why that's a problem. As for positives, they can take up to a few extra hits, which, against mobs is fine, and against players, doesn't actually really matter much since most of the most effective ways to take down a player don't care about resistances or health.

Now, I would be all for a change where like, what you take restricts you from taking other things. Like, I don't think you should be able to take Haste and Hardy at the same time, as an example. Or brute resist and burn resist at the same time. Things like that I think are fine.

Haery70:
Sharp Melee: Grants a melee attack buff when unarmed, reducing that to 0 means that it's a free attack buff.
Antiseptic Saliva: Was literally made a positive trait because it grants the ability to heal minor external damage at the cost of an easily replaceable resource: Nutrition. It had to be revised multiple times before getting accepted since the initial setup was basically healbellies without having to do vore on another codebase.
Snow Walker: Literally what it says on the tin, allows you to walk on snow with no penalties. I get that we don't have a lot of snow when V-3B isn't snowing, but I can guarantee that some players will still take the trait as a free trait.
I don't know if Winged Flight would have any benefits to being at 0 cost, but the fact that it's difficult to move around in the current map without running into a wall already balances itself out.

In the case for the removal of the listed negative traits I see that a lot of those traits are traits that actually have mechanical risks to take when in combat outside of Major Burn Weakness and Conductive Plus, which were often the go-to for dump stats right above Major Slowdown. With all the changes that's on the list it looks more like it's a huge buff for min/maxxers.

ItsTheKet:
While I will fully accept homogeny in traits and have, behind the scenes made some proposals to lower the heavy impact this would create. I personally am not very into this pull, as it directly goes against a personal ideal of mine of how something CUSTOM should be. Much like DWinters has mentioned, there isn't very much of a difference between a regular canon species and a custom one, barring some of the more extreme ones. 

Personally, while I am very well aware this would create much more work than this rebalance in of itself, I would much more propose to DOUBLING DOWN on the extremism. And allow me to explain:
SS13, despite it being a game I am relatively new to. Is one I hold rather dearly, with my own personal canon that many things, new and strange, or mundane and common can be seen all over. You can run into a many tentacled creature who can, in fact stab you with a chemical that will in fact heal you. Or Run into a mineral based entity with no need for atmosphere yet frail enough to shatter into pieces with a strong enough twack. It's a system that, while extremely faulty, brings way to so many things that can be difficult to achieve in other game platforms.

This PR goes against that vision of mine.

I want to see more species who Breathe Phoron, Get poisoned by being exposed to Oxygen or use the hidden mechanic that Alraunes have where they can subsist off of a high intake of Carbon Dioxide.

I wish to see Promethians being able to blob a ball of jelly, which without it's atmosproofing allows it to just be a cute little jelly blob about. I wish to see the vastly underused Skrell have the head pockets they have in paradise, giving them a SINGULAR ante above other species. On the flipside, While we will never see ventcrawling as a commnplace thing, I wish to propose allowing the OVERLY USED Promethians ability to modify their appearance, albeit much more limited to their nanite counterpart, be more common place.

I wish to see the feral mechanic be used more often. I wish to see Crystaline creatures that reflect lasers but crumble by a wrench to exist.

**I wish for more wild and meaningful changes to traits be something that is possible.**

This is obviously more long term, and will take much more work, but I want THIS more than I want things to become much more "normal"

I don't want "normal" I want strange and bizarre.

Novacat:
Keep in mind that 0-cost positive traits still count against your total number of traits, that is their 'cost' so to speak.

Also, Snow specifically refers to snow on the ground, which, Virgo 3B does not have anywhere except a few patches in the wilderness.

That being said, I am going to revert most of the changes.

Heroman3003:
Custom Species overshadowing regular species in aspects those regular species are supposed to be known for, be it health, speed, endurance, specific resists, etc, is very dumb tbh. It removes the point of regular species existing, especially since some of those are hard to justify as is. I always viewed Positive/Negative traits as something of a minor fluff with little mechanical impact, rather than severe balancing act of minmaxing all the best stuff. I feel like this PR moves towards that pretty well. Having custom species basically invalidate the balancing points of regular species by allowing you to pick different, vast majority of the time way less impactful or way more situational downsides while keeping all the strong upsides is a big problem. And mitigating it by making sure Custom Species cannot reach same extreme premade regular species can is a good solution, simply toning back the amount of impact traits have without rendering it completely nil.

Characters being more fragile than teshari, more slow than diona, as hardy as unathi, as vulnerable as promethean all remove the reason the base species have to exist other than 'the lore', most of which is still in WIP-paragraph-summary hell for majority of them. And the fact that you can balance out traits that would otherwise be extremely strong by just taking stuff that will never have any significant effect, like Conductive only makes issue worse.

~~Also, I still don't understand how Weaver is worth more than Sharp Claws and as much as an actual resist trait.~~

Meghan-Rossi:
I don't see a reason to remove major conductive.  Just make it cost the same as conductive.  Some people will still choose the worse version for the same reason they take allergies and stuff.  

I don't think removing major hardy is a good idea with the default movement speed this slow.  Maybe reduce the effect to a .25 or .3 multiplier instead of .1, and make it cost three points.  Similarly blocking haste from being taken with major hardy would be very bad for any role that requires spending much time in a voidsuit, unless the default movement speed is increased.

I can see the point of not overshadowing other species' unique advantages, but I don't think the advantages that need to be balanced include the fragility of teshari, the slowness of diona, or the vulnerability of prometheans.  I do not think we have a problem where it is too easy to create characters that are fragile, slow, and vulnerable.  I would say the advantage of teshari is speed, the advantage of diona is chemical and heat resistance, and the advantage of prometheans is radiation resistance and regeneration, and there are no traits that let custom species match those.

Heroman3003:
> I do not think we have a problem where it is too easy to create characters that are fragile, slow, and vulnerable. I would say the advantage of teshari is speed, the advantage of diona is chemical and heat resistance, and the advantage of prometheans is radiation resistance and regeneration, and there are no traits that let custom species match those.

I was meaning less so advantages and more so iconic and very much unique aspects of those rather than specifically mimicing advantages.

Verkister:
Perhaps some people might want to play a fragile character without having to force their character to be exclusively either a slime blob or a space chicken.

Verkister:
In my opinion, the "extremes" are what makes the trait system _interesting_. If shit has to be borderline normal with minimal effects on the ingame experience, what even is the point? No one's gonna _see_ the traits _listed_ on another person's character and figure out _"hey this totally fits this character's vibes"_ if the traits are hardly noticeable outside of the hidden mob variables. If someone wants to play a buff fuck or a scrawny squisher, by all means let them. A character _designed_ to be buff or squishy is far more _interesting_ when the ingame experience matches their design than a character whose **physical features** are nothing but fluff.

Verkister:
Honestly if we were to leave the "extreme features" to be exclusive to vanilla species, _this would not increase the use of those species for any lore-friendly purpose_. However it would _increase the reasons for people to pick those species just for mechanical minmaxing_. Like not playing a _teshari to play as a teshari_, but just to play **fast**, or not playing _unathi to play an unathi character_, but just to play **generic tank** without any correlation between the character and their species.

Verkister:
Inb4 "but muh visual balancing". If someone's gonna try and justify it by having the _species sprite_ to be _the_ "balancing factor" between custom and vanilla species, I'm just gonna remind them of that whole _"non-transparent prommie is a game breaking breach of mechanical balance"_ discourse we had a while ago.

Admiralkosmov:
This feels a bit pointless. SS13 is a game made by bored people for other bored people - removing content that exists in an already limited environment isn't where we should be going as community, imho.

You want teshari to be known as the speedy species? Then buff _their_ speed. Give them a dash. Give them gliding or whatever. You want diona to be known as the slow tank species? Then buff _their_ tankiness. Make them require actual sunlight to get nutrition. This _builds_ more content for them. Removing the speediness from everyone else is just cutting out code that is already there

code/modules/mob/living/carbon/human/species/station/traits_vr/positive.dm
```diff
@@ -29,124 +26,124 @@
...
-	var_changes = list("darksight" = 5, "flash_mod" = 2.0)
+	var_changes = list("darksight" = 4, "flash_mod" = 1.1)
 
-/datum/trait/darksight_plus
+/datum/trait/positive/darksight_plus
 	name = "Darksight (Major)"
 	desc = "Allows you to see in the dark for the whole screen."
 	cost = 2
-	var_changes = list("darksight" = 8, "flash_mod" = 3.0)
+	var_changes = list("darksight" = 7, "flash_mod" = 1.2)
```
Very-Soft, line 51: Also, pretty sure this is gonna make darksight not work at the edge of the screen. It's 7 tiles from the center to the edge of the screen, but would 1 mean you could see adjacent to yourself, or just the tile you're on? If it counts the center, then it should be 8.

KillianKirilenko:
I'm with winters and verki on this one. Custom traits could do with some reworking, but this isn't the right direction.

0-cost positives is an interesting idea though, definitely worth playing with in future versions of the trait system.

code/modules/mob/living/carbon/human/species/station/traits_vr/positive.dm
Novacat: Ye, that was intentional. Major Darksight was not intended to be perfect darksight.

Novacat:
Most of the big changes have been reverted, this is mostly going to be a balance pass for Darksight/Hardy/Conductive plus, as well as reducing the cost of a few positives to zero.

code/modules/mob/living/carbon/human/species/station/traits_vr/positive.dm
Very-Soft: Are  you sure about that? The description disagrees with you.

KasparoVy:
Gonna miss being Sergminer that can actually bear the burden of their equipment. Oh well, guess it wasn't meant to be.

SingoDingo:
From what I've found, movement speed seems to work on thresholds anyway, so normal hardy basically isn't even worth using in the first place. Which sucks, because the reason I use major hardy is that I'd rather be able to carry more than resist damage. I'd rather reduce the inconvenience of having to order a lighter voidsuit and find a black dufflebag (because that's basically all that is), than have any other mechanical advantage.

TheFurryFeline:
What on Earth happened with the dme?

Meghan-Rossi:
If engineering and mining had Vey-Med or equivalent voidsuits, losing the haste+hardy combination might be reasonable.  But without that, I really don't like this.  Losing haste+major hardy is painful enough.  Is haste+hardy really causing that much of a problem? A haste-compatible version of hardy that cost 5 points instead of 2 might work too.

Novacat:
I think it is more of an issue that haste+hardy has become meta in the first place. That being said, staff had a discussion about this and the end result is that item slowdown in general needs a nerf, which is the direction I was going anyway.

KillianKirilenko:
> If engineering and mining had Vey-Med or equivalent voidsuits, losing the haste+hardy combination might be reasonable.

EVA *hard*suits do have zero slowdown once fully engaged, and can be ordered from cargo like veymed suits. Those are, admittedly, not *void*suits.

> I think it is more of an issue that haste+hardy has become meta in the first place. That being said, staff had a discussion about this and the end result is that item slowdown in general needs a nerf, which is the direction I was going anyway.

I was mulling over this and I think one way to handle it could be making it so that traits can work off `+=/-=` rather than straight `=`. This would allow for more combinations of traits, and for traits to interact by more than just excluding each other.

Meghan-Rossi:
> EVA _hard_suits do have zero slowdown once fully engaged, and can be ordered from cargo like veymed suits. Those are, admittedly, not _void_suits.
I didn't know about that, that does work pretty well for mining.  It might be good if the tesla submap had one of those instead of the extra voidsuit.

Wickedtemp:
I mean I was stacking hardy and haste because my character's a giant feline taur, so speed and physical capability are just kinda implied with that type of build

Wickedtemp:
Actual opinion on this rather than personal anecdote though:

I think this is going in the wrong direction. Haste and Hardy barely made a difference as it was, and now it's even less. At this point why even bother having traits if they don't..do anything. I'd probably just remove the traits other than antiseptic saliva and call it a day rather than have damage-weaknesses, because the only reason I made fluff for Tempest having those weaknesses was to get points for hardy/haste, which were traits shi'd most likely have as a result of physiology and size. 

If our trait system is going to be 'rebalanced' until it's effectively just IC fluff with no impact on gameplay, then it's not a trait system, it's flavor text, and we have plenty of that already in char setup.

Sebbe9123:
Honestly, if this even has a remote chance to go through, it would be much better if you just put everything into the neutral catagory and let people pick and choose what they want overall with opposite traits locked out from being picked at the same time.
It is a roleplay server, why worry about balance at all if everything is just going to be the exact same for everyone except people playing canon species.

KasparoVy:
RIP

PR #10034, opened 2021-03-29
Implements a 'system' to allow whitelisted job alt-titles.

Heroman3003:
An alt-title can now have a ckey attached to it, and only that person will be able to select the alt-title. Currently no whitelisted alt-titles of course, but who knows if some will end up being put up/approved.

### DIFF
--- a/code/game/jobs/job/_alt_title_vr.dm
+++ b/code/game/jobs/job/_alt_title_vr.dm
@@ -1,0 +1,2 @@
+/datum/alt_title
+	var/whitelisted_ckey = null
--- a/code/modules/client/preference_setup/occupation/occupation.dm
+++ b/code/modules/client/preference_setup/occupation/occupation.dm
@@ -222,7 +222,7 @@
 	. += "</tt>"
 	. = jointext(.,null)
 
-/datum/category_item/player_setup_item/occupation/OnTopic(href, href_list, user)
+/datum/category_item/player_setup_item/occupation/OnTopic(href, href_list, mob/user)
 	if(href_list["reset_jobs"])
 		ResetJobs()
 		return TOPIC_REFRESH
@@ -237,7 +237,13 @@
 	else if(href_list["select_alt_title"])
 		var/datum/job/job = locate(href_list["select_alt_title"])
 		if (job)
-			var/choices = list(job.title) + job.alt_titles
+			var/choices = list(job.title)
+			for(var/I in job.alt_titles)
+				var/datum/alt_title/AT = job.alt_titles[I]
+				if(!initial(AT.whitelisted_ckey))
+					choices += I
+				else if(initial(AT.whitelisted_ckey) && user.ckey == initial(AT.whitelisted_ckey))
+					choices += I
 			var/choice = input("Choose a title for [job.title].", "Choose Title", pref.GetPlayerAltTitle(job)) as anything in choices|null
 			if(choice && CanUseTopic(user))
 				SetPlayerAltTitle(job, choice)
--- a/vorestation.dme
+++ b/vorestation.dme
@@ -752,6 +752,7 @@
 #include "code\game\jobs\whitelist.dm"
 #include "code\game\jobs\whitelist_vr.dm"
 #include "code\game\jobs\job\_alt_title.dm"
+#include "code\game\jobs\job\_alt_title_vr.dm"
 #include "code\game\jobs\job\assistant.dm"
 #include "code\game\jobs\job\assistant_vr.dm"
 #include "code\game\jobs\job\captain.dm"


SweetBlueSylveon:
I don't really like the idea of fluff job titles. Not only would NT not issue specialized job titles, but if you wanted to get one, you can just go to the HoP. I don't feel like it's reasonable to code in a way to circumvent.

Heroman3003:
> I don't really like the idea of fluff job titles. Not only would NT not issue specialized job titles, but if you wanted to get one, you can just go to the HoP. I don't feel like it's reasonable to code in a way to circumvent.

I'm not sure its right to call this 'circumvention', since its a whitelist-only and predefined thing. Especially since changing title at HoP's has awful tendency of yeeting you into Miscelaneous cateogry, making you pretty much invisible on the manifest.

Screemonster:
I've wanted to be the quartermonster since I started playing this damn game.

SweetBlueSylveon:
> > I don't really like the idea of fluff job titles. Not only would NT not issue specialized job titles, but if you wanted to get one, you can just go to the HoP. I don't feel like it's reasonable to code in a way to circumvent.
> 
> I'm not sure its right to call this 'circumvention', since its a whitelist-only and predefined thing. Especially since changing title at HoP's has awful tendency of yeeting you into Miscelaneous cateogry, making you pretty much invisible on the manifest.

This is the fault of individual HoP's. If you find this happening a lot, make sure to mention it in your interactions. But allowing fluff title requests would completely remove any need for it, and I fear, would create a trend of "Well I might as well just go whitelist for it".

SweetBlueSylveon:
Having this for silly or meme titles would be a problem that I'm not happy to have. But if we have a standard for this, then that's something different I suppose. Someone with a gimmicked character who makes pastries, and wanting to have a title "pastry chef" would be good.
My big question would be if this would allow for a system of applying this to in game spawned characters. Such as spawning event characters and having the option to set their ID job title.

Heroman3003:
> Having this for silly or meme titles would be a problem that I'm not happy to have. But if we have a standard for this, then that's something different I suppose. Someone with a gimmicked character who makes pastries, and wanting to have a title "pastry chef" would be good.

Honestly, that sounds better as a general alt-title than a personal one. I don't have strong opinion on memey alt-titles one way or the other though, but I do think that there is place to potentially have player/character-exclusive alt-titles.

>My big question would be if this would allow for a system of applying this to in game spawned characters. Such as spawning event characters and having the option to set their ID job title.

That is already possible tbh? I mean, it'd be more about changing IDs themselves as this whole ordeal is pretty much unrelated to spawning in event stuff specifically, at least as far as I can see, unless you want to whitelist people for event-exclusive alt-titles or something?

lm40:
i am not too much of a fan of this tbh, mostly because i don't see why one person would get unique access to a job title unless it's a meme title and i don't really see a reason for the company to issue a meme title. hop can if they wantn.

i know this  technically doesn't add it but seriously what other whitelist job titles would you apply for other than memes? if it's a non-meme title it could just be added as a universal alt-title.

InterroLouis:
The entire point of it being a whitelist is that it would require approval beforehand. If something too memey gets through, that would be on the approvers, wouldn't it? There are definitely possibilities for alt titles that are both not entirely memes, and are more personalized to a person, such as Scree's Quartermonster that they mentioned.

Wickedtemp:
(and also Chief Mlemical Officer)

Wickedtemp:
So, now that I'm on desktop instead of mobile, I can type a lil somethin-somethin up.

I think this is fine. I'm also biased, because I actually put in an app for a whitelisted alt-title of Chief Mlemical Officer that was called "elitist" by someone for some reason and then basically closed instantly without much discussion. I would have liked to talk about it a bit more to understand the concerns that people may have.

I think that most of these concerns would be handled through the whitelisting process. The only real issue would be the parameters for acceptance. The level of 'seriousness' in this server varies **wildly**, and it's not uncommon for NT to be depicted as being borderline incompetent buffoons one day, silly and scheming corporatists the next, and then completely level-headed and serious business owners the day after. I recall one admin, years back, saying that NanoTrasen would ship shuttles "full of fire" as a means of receiving large payments due to a "dangerous shipment fee". I think this is probably the best way to have it, though, because it allows greater freedom to play things out in different ways, with no hard-defined script to stick to when dealing with CC. 

We've had 'meme' characters that were done relatively well, and while I do generally argue on the side of higher RP standards, I don't see why this sort of thing should be disallowed as a whole. At the end of the day, this just seems like it could add a bit of fun here and there without taking away from the server or experiences of others. Provided that the title is a more or less clear indication of what that role is, does not include offensive terms, and doesn't try to list itself as a rank higher than it really is (IE "The Actual Director" for HoS or "Master Of Life And Death" for CMO) then I don't see why not. Especially if it's something that relates to the character that kinda makes you go "Yeah, that figures" when you see them with that ID.

If enough people like a whitelisted title and like five or so all want it, then just make it available. Now, the reason why I think this is a good feature instead of relying on the HoP is more or less the same reasons we have fluff items that allow mundane objects instead of just saying "just go to x area and ask for one every single shift you play from here on out". If it's something you may be doing **often** then going through that process may become a bit of a hassle for yourself, and for the HoP you fax for every single round for the sole purpose of an ID change.

lm40:
> Discussing this with staff, I've come to the consensus that we, majority do not want to fluff accept silly or meme names. If this PR you linked is added, we would only be accepting serious job alt titles. Such as "Pastry Chef" for a chef alt. I'm going to deny this application.

This is honestly why I was basically against the PR. It kinda opens the floodgates for people to apply for meme/nonsensical alt titles. I mean this is a feature technically so it is not really a bad thing to have inherently in case someone does actually apply for a unique job with an appropriately flavored alt-title as part of some SCA or something, but as far as I know no one has actually SCA'd a unique job before so it feels kinda cart before the horse to actually add it. IMO anyway. Most alt-titles that are just mild reflavors could be public. Like 'pastry chef', for instance.

What would be good with the job changes in general tho is if the last non-custom job assigned is what determined your manifest listing but there is a lot of funny stuff with the HoP in general, especially as it relates  to PTO/assigned  jobs etcetera so it is probably well out of the scope of this discussion.

Arokha (changes requested):
Seems to be not desired by staff at the moment, so only dubious merge potential, but:

As written, allows client manipulation to join as whitelisted job titles, needs additional checks at join time like alien whitelist etc.

JoanRisu:
This is a meeting thing but my general thought on it is; It involves a whitelist, which means oversight is by default necessary and would already filter out the more blatantly silly titles. Putting aside the application workload, the question becomes how we should handle inside-joke/meme titles? 

We'll discuss this further for a meeting.

PR #10036, opened 2021-03-29
Makes modifying ID cards not change the department if rank wasn't changed

Heroman3003:
Title. Means HoP-modified IDs won't be defaulted to Miscellaneous.

Polaris version: https://github.com/PolarisSS13/Polaris/pull/8018


TheFurryFeline:
Brilliant, now I can make a return of Site "Pudge" Manager. :3

PR #10041, opened 2021-03-29
Unwhitelists xenowear stuff

Heroman3003:
Polaris be lame with those restrictions.


Novacat:
tbh, this is not necessary since those gear restrictions only apply if the species in question is whitelisted.

PR #10043, opened 2021-03-29
Smole world beta Quick fix.

H0lySquirr3l:
Lowers the price WAY down to a more acceptable level for the bricks.

### DIFF
--- a/code/datums/supplypacks/recreation_vr.dm
+++ b/code/datums/supplypacks/recreation_vr.dm
@@ -93,7 +93,7 @@
 	contains = list(
 			/obj/item/weapon/storage/smolebrickcase, /obj/item/weapon/storage/smolebrickcase,
 			)
-	cost = 49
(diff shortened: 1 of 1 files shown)


code/datums/supplypacks/recreation_vr.dm
```diff
@@ -93,7 +93,7 @@
 	contains = list(
 			/obj/item/weapon/storage/smolebrickcase, /obj/item/weapon/storage/smolebrickcase,
 			)
-	cost = 200
+	cost = 49
```
Heroman3003, line 96: WHy an unround number? Almost every other thing in cargo is priced at multiple of 5

H0lySquirr3l: I did it based on the price of a standard tub of legos.

H0lySquirr3l:
price is now just 50

PR #10061, opened 2021-03-31
Enables Novel Posting

R3dtail:
Doubles Max_message_len from 2048 to 4096

Because I hate having to break up posts and this would at least mitigate it somewhat , additionally making it easier for folks to get their posts out without worrying AS MUCH about the 2048 character limit should? be decent.

### DIFF
--- a/code/__defines/misc.dm
+++ b/code/__defines/misc.dm
@@ -77,7 +77,7 @@
 #define DO_AUTOPILOT 5
 
 // Setting this much higher than 1024 could allow spammers to DOS the server easily.
-#define MAX_MESSAGE_LEN       2048 //VOREStation Edit - I'm not sure about "easily". It can be a little longer.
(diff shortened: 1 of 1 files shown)


Very-Soft:
I'm fine with this personally, unless there's some kind of technical reason this is a bad idea or whatever. It doesn't happen often, but it's annoying when you hit the text cap, so I'm fine with it being raised.

Wickedtemp:
Totally support this. Text caps are annoying.

dwinters99:
image
Thank you.

TheFurryFeline:
Also supporting this as a novel poster >w>

Tenryumi:
Yes please. I WANT this.

Arokha:
> image
> Thank you.

Hey be glad I added that message instead of it just vanishing into the ether which is what it did before |3

PR #10077, opened 2021-04-04
Turns things back to normal

Novacat:


dwinters99:
april fools tether was unironically SO MUCH BETTER than normal tether

F

PR #10082, opened 2021-04-05
Nerfs Protean Blobform

Novacat:
People really should not be trying to go into battle with this.


Verkister:
With the slowdown they won't be getting out of the battle either.

PR #10085, opened 2021-04-05
More Protean Tweaks

Novacat:
Going to try increasing simplemob nom times instead.

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/protean_vr/protean_blob.dm
+++ b/code/modules/mob/living/carbon/human/species/station/protean_vr/protean_blob.dm
@@ -36,7 +36,7 @@
 	max_n2 = 0
 	minbodytemp = 0
 	maxbodytemp = 900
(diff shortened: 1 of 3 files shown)


Screemonster:
does this affect only proteans or everyone? Because I can see this making mouse-eating an absolute pain.

Novacat:
Basically unless I made another var specifically for mice it would maybe effect mice as well?
I have no idea if it affects object holders though.

Verkister:
This affects everyone and every mob

Verkister:
and yes, it also affects scooped holders since those just relay the vore onto the held mob

Verkister:
should just leave the vore time stuff untouched and make attacks cancel the vore timer because this is fucking over everyone literally only because some proteans went battle-blob in order to abuse vore for combat advantage, which is already against rules.

Very-Soft:
Yeah let's not make vore things take three times as long across the board.

PR #10092, opened 2021-04-07
Update resize pref and some resizing code

Arokha:
Tweaked some things about resizing, nothing that really affects gameplay aside from moving the 'Toggle Resizing Immunity' to the vore panel and having it save with your characters.

- Admins who use the two admin tools to resize characters (admin size gun or admin verb) will produce several outcomes:
  - _Mob was in an area that allows extreme sizes, admin set them to an extreme size:_ Same outcome as if they had resized themselves (they will be resized upon leaving the areas that allow extreme sizes).
  - _Mob was not in an area that allows extreme sizes, admin set them to an extreme size:_ They are flagged as having an uncapped size and will not be automatically resized. Note that they still can't resize THEMSELVES to extreme sizes after. If they try, they're still limited to 25-200% unless they're in an extreme sizes area.
  - _Mob was previously flagged for extreme size allowance, admin sets them to a non-extreme size:_ Flag for extreme size allowance in any area is removed, mob returns to normal
- 'Toggle Resizing Immunity' moved to vore panel and added to character vore prefs JSON, defaults to allowing resizing (as it does now)
- Added things that can ignore your resizing immunity pref and resize you anyway, generally because they're things you do to yourself, or admin tools to adjust things:
  - Bluespace jumpsuit
  - Protean resizing
  - NIF sizechange software
  - Sizegun when target and firer are the same (aka you shoot yourself)
  - Admin resize verb
  - Admin shrinkray (be careful gosh)
  - NOT macro/microcillin, sorry. Can't know who injected you with it.
- Macro/microcillin now grow you up to 600% or down to 1% if you're in an area that allows such instead of stopping at 25% or 200%.
- Added some defines for resize min and max amounts so I don't have to remember that 'tiny' is the smallest normal amount and 'huge' is the largest normal amount.


seahen (approved):

Leshana (approved):
🥚

PR #10096, opened 2021-04-08
Lets Talon Crew Select Some HUDs in Character Creator

KasparoVy:
Allow Talon crew to select some relevant HUDs in the character creator. Left off the AR glasses in case that was too teched out or get this PR contested

### DIFF
--- a/code/modules/client/preference_setup/loadout/loadout_eyes_vr.dm
+++ b/code/modules/client/preference_setup/loadout/loadout_eyes_vr.dm
@@ -1,9 +1,12 @@
 /datum/gear/eyes/medical
-	allowed_roles = list("Medical Doctor","Chief Medical Officer","Chemist","Paramedic","Geneticist", "Psychiatrist", "Field Medic")
+	allowed_roles = list("Medical Doctor", "Chief Medical Officer", "Chemist", "Paramedic", "Geneticist", "Psychiatrist", "Field Medic", "Talon Doctor")
 
 /datum/gear/eyes/meson
 	display_name = "Optical Meson Scanners (Engineering, Science)"
-	allowed_roles = list("Station Engineer","Chief Engineer","Atmospheric Technician", "Scientist", "Research Director", "Explorer", "Pathfinder")
+	allowed_roles = list("Station Engineer", "Chief Engineer", "Atmospheric Technician", "Scientist", "Research Director", "Explorer", "Pathfinder", "Talon Engineer")
+
+/datum/gear/eyes/security
+	allowed_roles = list("Security Officer", "Head of Security", "Warden", "Detective", "Talon Guard")
 
 /datum/gear/eyes/arglasses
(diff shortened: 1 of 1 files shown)


code/modules/client/preference_setup/loadout/loadout_eyes_vr.dm
```diff
@@ -1,9 +1,12 @@
...
-	allowed_roles = list("Medical Doctor","Chief Medical Officer","Chemist","Paramedic","Geneticist", "Psychiatrist", "Field Medic")
+	allowed_roles = list("Medical Doctor", "Chief Medical Officer", "Chemist", "Paramedic", "Geneticist", "Psychiatrist", "Field Medic", "Talon Doctor")
 
 /datum/gear/eyes/meson
 	display_name = "Optical Meson Scanners (Engineering, Science)"
-	allowed_roles = list("Station Engineer","Chief Engineer","Atmospheric Technician", "Scientist", "Research Director", "Explorer", "Pathfinder")
+	allowed_roles = list("Station Engineer", "Chief Engineer", "Atmospheric Technician", "Scientist", "Research Director", "Explorer", "Pathfinder", "Talon Engineer")
+
+/datum/gear/eyes/security
+	allowed_roles = list("Security Officer", "Head of Security", "Warden", "Detective", "Talon Guard")
```
Novacat, line 9: tbh Talon guard has no use for security HUD since its mostly people for seeing who is set to arrest and only Tether can do that.

Novacat (changes requested):

code/modules/client/preference_setup/loadout/loadout_eyes_vr.dm
KillianKirilenko: there's also no way talon would have access to NT's security or medical records, more to the point

perhaps a better solution would be to make some talon-specific eyewears that have some of the overlay features and other perks enabled (flash resist for marine/guard, etc.) but none of the records access?

KasparoVy: Good call, I think it'd be easy enough to make a child subtype of those HUDs with restricted functionality

Cheers for the suggestions.

KasparoVy:
Converted to draft until I sort this out, thanks again for the reviews.

dwinters99:
I like this, it would be nice to have this expanded to (or to have a seperate PR for) adding Talon as an allowed role for various explo stuff like drop pouches and engi stuff like welding-masks and such.

code/modules/client/preference_setup/loadout/loadout_eyes_vr.dm
KasparoVy: Note to self: 'legacy'/'surplus'/'retrofitted' HUDs for namesake.

Arokha:
Any progress?

PR #10103, opened 2021-04-08
Remove unnecessary 'signatures'

Arokha:


Novacat (approved):
Some of these are on Polaris as well so I will mirror those parts.

PR #10110, opened 2021-04-09
Updates Listening Post gateway submap

Heroman3003:
Title. Refluffs it in the merc-removing flavor.

Also, properly removes snow outpost submap from the pool, which @Novacat seems to have tried to do, but not in the right place? If it is wanted to stay, just say so.


Meghan-Rossi:
Those dates are all in the future compared to station time (currently 2321).

Heroman3003:
I intentionally left the dates on old logs wrong. I also wrote in a small excuse as to why that is (and why they reference now-noncanon events)

Sebbe9123:
> Those dates are all in the future compared to station time (currently 2321).

Gateway isn't neccessarily tied to station lore, you can have magic and shit inside the gateway maps if you want to, just not on the actual station.

Heroman3003:
> Gateway isn't neccessarily tied to station lore, you can have magic and shit inside the gateway maps if you want to, just not on the actual station.

Gateway 100% is tied to station lore, or at least lore of same universe. No magic in the setting.

KillianKirilenko:
Wrong files there, I think. You (accidentally?) altered the april fools tether submaps & defines not the main tether ones.

Heroman3003:
Ah, whoops, lemme just--

PR #10125, opened 2021-04-10
Adds thicc thighs markings.

Verkister:
Who needs a consistent silhouette when you can have killer thighs!


TheFurryFeline:
Thicc Thighs Save Lives

PR #10127, opened 2021-04-10
Species cleanup

Nyksia:
Cleans up the messy species produceCopy proc, which was completely overriden by xenochimera and alraune, ignoring the parent proc entirely.
Now all of the children species use produceCopy of base species which makes a thorough copy, while ignoring lists (because they may contain more lists, very good BYOND.) so there are no trailing references instead of proper copies.
This also goes to fix the adjustment to organs I've done in #10107 which didn't apply to custom species, xenochimera and alraune, because I did not expect the children species to not use that proc.


Novacat:
This seems to edit a *lot* of different files unrelated to the PR, what is up with that?

Nyksia:
They are, in fact, all related! produceCopy proc was used in all of the edited files. Since I also removed the need for first argument, I had to alter the proc calls accordingly. There is really no point in providing a species datum to a proc inside a species datum, if it's supposed to copy itself.

Novacat (approved):
Optimization is always good.

PR #10154, opened 2021-04-14
Silence, cleanbots

Little-119:
Silences cleanbots by just commenting out the line that makes them play the sound, removing the behavior entirely.
The cartridges that janitors get for their PDAs can be used to locate cleanbots, if I remember correctly. So this beeping isn't needed for locating cleanbots.

Why not instead add a toggle to them, like the ones medibots have? Because if one or more of them are beeping and annoying me, I'll have to hunt them down to shut them up. Also I'm too lazy to figure out how to make the needed changes to their UI

Why not instead add a preference? Because I'm too lazy to figure out how to add more preference settings.

My decision to create and submit this PR may be motivated in part by the headache I am currently suffering from.

### DIFF
--- a/code/modules/mob/living/bot/cleanbot.dm
+++ b/code/modules/mob/living/bot/cleanbot.dm
@@ -28,7 +28,7 @@
 /mob/living/bot/cleanbot/handleIdle()
 	if(!wet_floors && !spray_blood && prob(2))
 		custom_emote(2, "makes an excited booping sound!")
-		playsound(src, 'sound/machines/synth_yes.ogg', 50, 0)
+		//playsound(src, 'sound/machines/synth_yes.ogg', 50, 0) // VOREStation removal
 
 	if(wet_floors && prob(5)) // Make a mess
 		if(istype(loc, /turf/simulated))


SweetBlueSylveon:
It would be an exceptionally good idea to give cleanbots a sound toggle like medibots have, but like you said, you aren't doing that. I really dislike the idea of silencing medibots, as I personally enjoy their little beeps, and the voicelines from the medibot. It'd make me incredibly sad if this change were to go through.

TheFurryFeline:
I could take a crack at this when I get home after classes. Probably just dupe code from medbot including the tgui for its speaker

dwinters99:
Making a toggle would require faffing with tgui which I COULD do and have already dabbled with (vis a vis making a bot) so I could probably do it but it'd likely be better for someone who is actually good at tgui to do it.

Little-119:
@SweetBlueSylveon This PR isn't changing anything about medibots, no need to worry about 'em.

Heroman3003:
The speech should just be a toggle, but off by default, imo.

SweetBlueSylveon:
> @SweetBlueSylveon This PR isn't changing anything about medibots, no need to worry about 'em.
That was a mistype, I meant to say that I would be sad if you silenced cleanbots, and think that it would be a massive shame if this pr were to go through.

TheFurryFeline:
https://github.com/PolarisSS13/Polaris/pull/8035 will probably be a good replacement of this PR if/when it gets here.

Wickedtemp:
Totally fine with this.

SweetBlueSylveon:
> [PolarisSS13/Polaris#8035](https://github.com/PolarisSS13/Polaris/pull/8035) will probably be a good replacement of this PR if/when it gets here.
> 
> It effectively gives the option to silence cleanbots like with medbots on top of the main goal of converting all bot UIs to TGUI

This is a much better alternative.

Little-119:
If I hear one more cleanbot boop then I'm gonna go on a rampage against them

TheFurryFeline:
Bwoop~!

Novacat:
Made redundant by: https://github.com/PolarisSS13/Polaris/pull/8035

Little-119:
I'm a little disappointed that cleanbots won't be silent by default, but this is better than nothing

TheFurryFeline:
Bwoop bwoop~ ;p

PR #10164, opened 2021-04-19
Probably adds deletion ability to library

Arokha:
As title


Novacat:
This cannot be tested in private environments so yolo

SigholtStarsong:
Sweet Gods above, please and thank you. I have a comprehensive list of things that do NOT need to be in the library.

PR #10178, opened 2021-04-20
New belt items

dwinters99:
Toolbelts can now hold nanopaste (+ advanced) and gardening spades

- Nanopaste is the size of a tube of superglue and I never leave the house without at least one tube of superglue on my toolbelt
- Hatchets and mini-hoes can be put on the belt so why not spades

Janitor belts can now hold the light replacer

- I see no downside to this except for the possibility of disrupting the incredibly intricate janitor end-game balance loop


Arokha (approved):
Top two make sense. Light replacer itself is a belt slot item, having it fit in a belt is *weirdy* but I agree it's inconsequential.

JoanRisu:
My initial thought is "We should add MOLLE-system pouches of varying sizes"
On a more focused note, if its small and makes sense to fit on a belt, why isn't it able to go on belt?
This one is fine to go through, I can't see why it wouldn't and its been 16 days, so if someone had a problem with it, they should've commented by now.

PR #10197, opened 2021-04-23
Comments out obsolete xenowear whitelist check.

Verkister:
Felt like It'd be better to just comment it out rather than trying to "fix" it for custom species even after upstream fix.


Novacat:
Conflict, and this check is going to have to be redone both upstream and downstream.

PR #10199, opened 2021-04-24
The Personal Computer Unit

klaasjared:
Adds the new Personal Computer Unit (PCU) to the game to replace the static laptops in atmos, command, and medical. I always thought that the static laptops made no sense considering that there is already a portable laptop in the game - one that looks vastly different from the static one. So to replace the black brick that is the static laptop I dug through the code and found the 'personal' salvage computer in `salvageable.dmi` which I made rotateable. I hope this will be a more "makes sense" version of what is currently in the game. It's certainly better than a generic black box with a screen and buttons.

Pictures of the new computer are found below:

Test3

rotate_showcase


Novacat (commented):
Why are the sprite files in computer_vr.dmi when it seems like the base computer object is still pulling from computer.dmi?

klaasjared:
> Why are the sprite files in computer_vr.dmi when it seems like the base computer object is still pulling from computer.dmi?

I... I'mma be real with you, I'm not smart

klaasjared:
> Why are the sprite files in computer_vr.dmi when it seems like the base computer object is still pulling from computer.dmi?

Because I am not a very smart person - fixed it thought

Iroquois-Pliskin:
Aside from looking a little out of place in the art direction (if you can even say we have one) these seem pretty good.

klaasjared:
> Aside from looking a little out of place in the art direction (if you can even say we have one) these seem pretty good.

I tried to make it match the art style as best I could - going as far as to rip a sprite from another .dmi and making it rotatable for the sake of looking decent - but I admit it's not perfect

Novacat (changes requested):
Honestly? I would recommend just mirroring the Polaris PR down here, then re-opening the Polaris PR. That would make it much easier to deal with future code conflict, and, I would be happy to earlymerge it here.

PR #10203, opened 2021-04-25
Switch Xenoarch AR-R for Science Goggles

TheFurryFeline:


Novacat (changes requested):
tbh I would rather just change the AR-R glasses out for science goggles since science goggles are airtight already.

TheFurryFeline:
> tbh I would rather just change the AR-R glasses out for science goggles since science goggles are airtight already.

That'll be a map edit, which is technically out of scope here but can do the needed change when home.

TheFurryFeline:
PR was made under a misassumption. D'oh.

PR #10204, opened 2021-04-25
[MIRROR] Teshari Specific Medals sprites

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/8048
--------------------
image
Left is PR, Right is current.

Added teshari specific sprites for medals so that they are no longer hidden inside teshari character's head.

To note, the only difference is that the medals are 4 pixel down. Nothing was done to resize them (We all get the same medal after all!).


AstraEther:
=D

PR #10218, opened 2021-04-27
Restricts the suit sensors monitoring to alert levels

Heroman3003:
The amount of information provided by Crew Monitoring Console is now based on combination of suit sensors and alert level. Even if sensors are set higher than current alert's provided level, they will just gives less info to console. If they are set lower, they will not give any extra info just because code was raised.

On Code Green only information up to binary is provided.

On Code Yellow and Code Orange only information up to vitals is provided.

On Code Violet, as well as Code Blue and above full information is provided.

This is more of a proposal kind of thing, but the way its now, it functions precisely right. I've actually had idea to throw this up for a long while now, just never got the motivation to look at that section of the code.

### DIFF
--- a/code/datums/repositories/crew.dm
+++ b/code/datums/repositories/crew.dm
@@ -32,22 +32,29 @@ var/global/datum/repository/crew/crew_repository = new()
 
 				var/list/crewmemberData = list("dead"=0, "oxy"=-1, "tox"=-1, "fire"=-1, "brute"=-1, "area"="", "x"=-1, "y"=-1, "ref" = "\ref[H]")
 
-				crewmemberData["sensor_type"] = C.sensor_mode
+				//VOREStation Edit Start
+				var/real_sensors = C.sensor_mode
+				if(real_sensors > SUIT_SENSOR_BINARY && seclevel2num(get_security_level()) <= SEC_LEVEL_GREEN && seclevel2num(get_security_level()) != SEC_LEVEL_VIOLET)			//Code Green (and maybe lower if that ever happens?) only gets binary.
+					real_sensors = SUIT_SENSOR_BINARY
+				else if(real_sensors > SUIT_SENSOR_VITAL && seclevel2num(get_security_level()) < SEC_LEVEL_BLUE && seclevel2num(get_security_level()) != SEC_LEVEL_VIOLET)		//Codes Yellow and Orange get the vitals but not tracking.
+					real_sensors = SUIT_SENSOR_VITAL
+				crewmemberData["sensor_type"] = real_sensors																										//Code Blue and above + Code Violet get all the sensors
+				//VOREStation Edit End
 				crewmemberData["name"] = H.get_authentification_name(if_no_id="Unknown")
 				crewmemberData["rank"] = H.get_authentification_rank(if_no_id="Unknown", if_no_job="No Job")
 				crewmemberData["assignment"] = H.get_assignment(if_no_id="Unknown", if_no_job="No Job")
 
-				if(C.sensor_mode >= SUIT_SENSOR_BINARY)
+				if(real_sensors >= SUIT_SENSOR_BINARY)	//VOREStation Edit
 					crewmemberData["dead"] = H.stat == DEAD
 
-				if(C.sensor_mode >= SUIT_SENSOR_VITAL)
+				if(real_sensors >= SUIT_SENSOR_VITAL)	//VOREStation Edit
 					crewmemberData["stat"] = H.stat
 					crewmemberData["oxy"] = round(H.getOxyLoss(), 1)
 					crewmemberData["tox"] = round(H.getToxLoss(), 1)
 					crewmemberData["fire"] = round(H.getFireLoss(), 1)
 					crewmemberData["brute"] = round(H.getBruteLoss(), 1)
 
-				if(C.sensor_mode >= SUIT_SENSOR_TRACKING)
+				if(real_sensors >= SUIT_SENSOR_TRACKING)//VOREStation Edit
 					var/area/A = get_area(H)
 					crewmemberData["area"] = sanitize(A.get_name())
 					crewmemberData["x"] = pos.x


User-U-U:
I really don't think this is a great idea, people get hurt on different levels and still will want help, miners, xenobiologists, people just roaming maintenance. I personally don't see the benefit, thats just my two cents though.

dwinters99:
It's actually really hard to get people to raise the alert level even when an emergency is happening. Not only is command usually "busy", but they also usually don't understand the problem enough to set the alert level. Even spiders don't get people to raise the alert level to even yellow. And that's pretty much the only time people die other than miners.

KillianKirilenko:
An interesting idea, and one that would probably work well for a server with more strictly enforced RP standards, but I agree that getting the alert level changed outside of serious crises is unlikely and this'll probably just lead to more miners/xenobio/xenoarch not getting rescued and/or not getting resleeved in a timely fashion.

Natje0:
If an optional sensor option, yes, it'd be a great addition!
As is, I don't want it. I max my sensors for a reason, I want to be found if I get messed up.

Rykka-Stormheart:
I'd only agree if it was an additional sensor option "only display my info at x alert level", but as-is? Hell no.

This will lead to more deaths and more stress on Medicals part, especially because alert level changing isn't really enforced, and admittedly can be difficult to do in the heat of the moment, plus with how casual a lot of the server is, and how calm it usually is, you'll have miners die out in the underdark and now we can't find 'em because we're on green. Etc etc etc.

ZodiacShadow:
Hard disagree with this one. On our server - in my experience - most damage is taken while on Green. Which would make suit sensors almost entirely useless.

AstraEther:
I would like to see this as optional thing someone can purposfly do, and not be on by default. Have default values be displayed on green.

Iroquois-Pliskin:
I like being able to have my sensors set to vitals on green. Tying it to alert level doesn't make sense to me.

No thanks

YuskoTheCoyote:
I hold the opinion that this change will help add to the roleplay environment. Medical would not have to set one person up at the sensor console all round long, as is what I typically see whenever I see or play as medical. This change would also encourage people to communicate and engage with each other through comms. And typically, it's the injured who's able to get medical to go where they are needed. So I doubt there will be that much of a change in the grand scheme of things. Medical will adapt to this change, especially when they still have all the other tools at their disposal.

pontifexminimus2:
Usually, the people getting hurt during green alert levels are the ones who get bumped into crit by animals which severely cuts down on communications.

Considering they can't speak in crit.

Furthermore, medical has plenty of opportunity to both check the sensors and not be glued to the front desk - AR medical glasses have crew monitors, so do the NIFsoft they get. In fact, we should include the Vegeta health glasses to this too, just so they have the same functionality as dorky AR glasses. But that's my opinion.

In my experience, alert levels are the most neglected functionality of staff, so I don't really approve this change. It took 45 minutes and the assertion of medical staff that I am indeed dead to raise the level to orange after the entire first floor of the asteroid was wiped by a Tesla.

R3dtail:
yeah there's plenty of tools to enable medial players to *not* stay at the front desk, it's just good to have someone in the department .  all this does is make the one tool folks to find injured people near useless. THis just makes shit more difficult and annoying. and much, much less fun.

Wickedtemp:
This wasn't even a serious suggestion. It was a joke, and for some reason now its a PR. 

I like having my sensors on tracking at all times. Its saved Tempests life. Alert levels are usually neglected by Heads, and this turns one of the most useful tools Medical has into something completely useless. 

So uh... no, lol

lm40:
Yea I don't like this change mostly because raising to code violet any time a single person gets iced  silently is gonna be annoying for everyone. It also is gonna cut down on medical being able to save people rather than doing corpse recovery because there's less to identify warning signs on at-risk people, like a miner spiked into crit by sudden brute. The Tether is also ENORMOUS, so trying to find people without sensors is something I would basically instantly give up on; even when they give decent information ('HELP ENGINEERING') there is a fair bit of space to cover and forget about trying to find one dead person across 7+ z-levels if they don't even have the chance to scream that.

Besides these things are opt-in so tying them to alert levels doesn't really make any sense.

Wickedtemp:
> I hold the opinion that this change will help add to the roleplay environment. Medical would not have to set one person up at the sensor console all round long, as is what I typically see whenever I see or play as medical. This change would also encourage people to communicate and engage with each other through comms. And typically, it's the injured who's able to get medical to go where they are needed. So I doubt there will be that much of a change in the grand scheme of things. Medical will adapt to this change, especially when they still have all the other tools at their disposal.

actually fully read this -- yeah we already don't need to sit at the reception table all day. We have methods of watching sensors through Nifs and AR-M glasses. These were added before Tether was even a thing. Having sensors be useless is a **huge** change and I don't even think Medical would necessarily "Adapt" to them being useless all the time. At best, we'd "get by". All this does is make the job annoying to play. This PR has no benefit.

TheFurryFeline:
As a past med player, I am very much against this for the reasons that've been stated already: this will make medical's job nigh impossible where most injuries can happen while on green alert. No support on this PR.

dwinters99:
Revisiting this to put quickly into words something I think most will agree with. Having an "automatic" setting that was dynamic by alert level *would* be nice. Having it be mandatory is NOT nice.

Wickedtemp:
Yeah, like.. if the systems were:
- At green/yellow/orange nothing changes
- Blue, minimum setting becomes Vitals
- Violet/Red, all tracking

While still allowing Tracking on Green, etc, then sure.

DemicusMaximus:
I don't agree with this one. People can already turn off their sensors, and medical isn't required to sit there and monitor vitals anyway. While a nice idea, I think it only makes for more problems than it alleviates.

Arokha:
May be better handled by https://github.com/VOREStation/VOREStation/pull/10218

Arokha:
I'd think a more lore-friendly way to implement this is to just let people turn their sensors on and off, and the alert level determines what information is available on the consoles. As written, this will leave people with a bunch of suit settings that say 'vitals' or 'binary' or whatever, but don't *actually* provide that information. It's a misleading interface at that point.

I also think that people who want, should be allowed to provide whatever info they want, if they want to. Might also be useful for people performing dangerous work or explorers or whatever, and you don't wanna change the alert level for that, so... ehhhh. I don't really like this idea.

JoanRisu:
Mmmm, not sure I like this. 

By that I mean, it's well-intended but design-wise doesn't take into consideration what can happen in green alert and the issue of who is in command and ready to raise the alert to the necessary level.
For now, it's a no. We can revisit this if necessary but currently I don't this will improve QoL significantly enough to justify it.

PR #10221, opened 2021-04-27
Disables Suit Sensors in Dorms

Novacat:


Wickedtemp:
Probably needs some sort of description. Basically this would block suit sensors in dorm rooms - the reasoning for this is:
- No one has to see the damage from different scenes in dorms,
- People aren't gonna see someone dying on sensors and potentially freak out and it turns out they've been in a dorm the whole time
- Dorms are supposed to be private and so this kinda fits that theme anyway

also can we extend this to a sensors blocking belly mode and just have that on by default
-

Rykka-Stormheart:
I'd prefer just sensors being blocked in dorms. If someone gets grabbed-n-gulped and digested outside dorms but ICly wants to resist, having sensors dampened would be counterintuitive, etc?

Wickedtemp:
This may be a hot take, but this:


> If someone gets grabbed-n-gulped and digested outside dorms but ICly wants to resist..

shouldn't be done through suit sensors. That'd be involving other people. I don't want to be dragged into it. Plus, this would be a good tool to prevent prey from using sensors to scream for help when the pred had no idea that it was going to happen and didn't want this to go down. We already have a stripping mode that accomplishes the sensors disabling, but this would be more direct.

Iroquois-Pliskin:
just make it a toggle in char options as to whether you have suit sensors auto-disable in dorms, enabled by default.

dwinters99:
This would also cover abandoned dorms which is a semi-common place for people to OD on floorpills. Make it a toggle.

Wickedtemp:
if you eat floor pills, you accept your fate

dwinters99:
image

Arokha (changes requested):
Laaaaaaaaaanx this is such a hack of a way to implement this :C
Do it a good way and not using thattttttt

code/datums/repositories/crew.dm
```diff
@@ -24,7 +24,8 @@ var/global/datum/repository/crew/crew_repository = new()
 	var/tracked = scan()
 	for(var/obj/item/clothing/under/C in tracked)
 		var/turf/pos = get_turf(C)
-		if((C) && (C.has_sensor) && (pos) && (pos.z == zLevel) && (C.sensor_mode != SUIT_SENSOR_OFF) && !(is_jammed(C)))
+		var/area/B = get_area(pos) //VOREStation Add: No sensor in Dorm
+		if((C) && (C.has_sensor) && (pos) && (pos.z == zLevel) && (C.sensor_mode != SUIT_SENSOR_OFF) && !(is_jammed(C)) && !(B.block_suit_sensors)) //VOREStation Edit
```
Arokha, line 28: Sorry I didn't notice this before, this whole section could be done better. Mind touching it up while you're in there?

`var/area/B = pos?.loc` works, can save a get_area() call because the turf's loc is always the area it's in (and the nullsafe operator in case the turf is null)

The big if() chain can have a couple optimizations to just be...

`if((C.has_sensor) && (pos?.z == zLevel) && (C.sensor_mode != SUIT_SENSOR_OFF) && !(B.block_suit_sensors) && !(is_jammed(C)))`

C will definitely exist, there's no potential for pausing between getting the list of clothing/unders and iterating over it (maybe there should be...). Pos only has one thing accessed on it, can just use nullsafe dereference there. I put the block suit sensors test before jammed test because I think it's more likely to be in dorms than to be bluespace jammed, and more likely things should come earlier... could probably be optimized more, maybe C.has_sensor is needless, iono

Arokha (approved):

PR #10226, opened 2021-04-27
Expanding the Public Garden significantly, or: Introducing the Public Leisure Minidome. 

ResidentCody:
### **Disappointed by how depressingly enclosed the Public Garden is and _still_ seeing people reminisce about the April Fool's map with its wide open spaces, grass, trees, et cetera, I decided to take it upon myself to do something about it.**

Thus, the Expanded Public Garden. At roughly four times the area of the original, it includes revolutionary concepts such as:
- Grass you can actually lie down on without breaking into Hydroponics or rolling in farming soil.
- A number of benches, tables, and chairs for relaxing at or perhaps bringing some food for a picnic.
- An assortment of colorful flowers, bushes, and healthy green foliage in general to provide a relaxing environment instead of only ever seeing the sadness that is Virgo 3b's flora.
- +5 soil patches (13 total), rearranged to better follow the structure changes. One main line along the left and two pairs on the right.
- Exterior surface tiles reworked and catwalks/railings added to maintain the ability to navigate outside safely.
- All z-level transitions unaffected. Stairs and the gas/power lattice on the right remain untouched to preserve compatibility with the Surface 2 z-level.

**Current Public Garden**

**Expanded Public Garden**

### DIFF
--- a/maps/tether/tether-03-surface3.dmm
+++ b/maps/tether/tether-03-surface3.dmm
@@ -5691,9 +5691,16 @@
 /turf/simulated/floor/tiled/monotile,
 /area/tether/surfacebase/surface_three_hall)
 "aiq" = (
-/obj/machinery/portable_atmospherics/hydroponics/soil,
-/obj/machinery/camera/network/civilian,
-/turf/simulated/floor/grass,
+/obj/structure/grille,
+/obj/structure/window/reinforced/full,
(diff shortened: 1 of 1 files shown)


ResidentCody:
Closing PR to rework my side of some github spaghetti, will resubmit once done with that.

PR #10227, opened 2021-04-28
Expanding the Public Garden significantly, or: Introducing the Public Leisure Minidome.

ResidentCody:
### **Disappointed by how depressingly enclosed the Public Garden is and _still_ seeing people reminisce about the April Fool's map with its wide open spaces, grass, trees, et cetera, I decided to take it upon myself to do something about it.**

Thus, the Expanded Public Garden. With significantly more area than the original, it includes revolutionary concepts such as:
- Grass you can actually lie down on without breaking into Hydroponics or rolling in farming soil.
- A number of benches, tables, and chairs for relaxing at or perhaps bringing some food for a picnic.
- An assortment of colorful flowers, bushes, and healthy green foliage in general to provide a relaxing environment instead of only ever seeing the sadness that is Virgo 3b's flora.
- Soil patches rearranged to better follow the structure changes.
- Exterior surface tiles reworked and catwalks/railings added to maintain the ability to navigate outside safely.
- All z-level transitions unaffected. Stairs and the gas/power lattice on the right remain untouched to preserve compatibility with the Surface 2 z-level.

**Current Public Garden**

**Expanded Public Garden**
Screenshot 2021-04-30 153130


ResidentCody:
Sorted out my GitHub-side stuff. Reopening PR for this.

AstraEther:
This Anon totally not Teshari tushies approve of this PR

ResidentCody:
Second commit adding an intercom near the APC and two AI Holopads for sufficient room coverage.
Screenshot 2021-04-27 202357

TheFurryFeline:
> This _does_ look really nice, but it is also making the tether bigger. I don't see this specific example as much of an issue, but it's something to keep in mind.

Yeah, it's only filling in some otherwise empty area on the surface which is the whole focus behind the asteroid downsizing thing.

Iroquois-Pliskin:
This absolutely needs at LEAST 2 more vents and scrubbers. That much glass is begging for a breach and a single pair of atmos equipment is nowhere near enough.

Andromeda-K22:
>1 vent, 1 scrubber

You're asking for engineers to hate you.

pontifexminimus2:
At first, I thought it is neat and cute but then I realized it is a massive room with only one vent and scrubber.

I curled up in a fetal position, whimpering softly.

Novacat (changes requested):
Would like more vents and scrubbers since it is such a large room.

pontifexminimus2:
Also re: Tether size

I think it's ok because this is just enhancing a common hangout spot (I use it at least, so this might be more anecdote than fact.)

I do would like to see at least 2-3 more scrubbers and vents before merging this, however.

ResidentCody:
All demands for more gas succ (air scrubbers and vents) to account for large room size have been noted. Will be adding more later today.

SweetBlueSylveon:
I feel like this further dilutes hang out spots on tether 3. And it only really adds a bunch of open, rather empty space in the public garden. As nice as this looks in dmm, standing on one side you will not be able to see the other, by quite a number of tiles. I'd vastly prefer this area to be more compact, and not repeat the issues with the old bar. I'd just prefer this change not go through.

ResidentCody:
Air scrubbers reworked. Added +3 pairs of vents/scrubbers to make 4 total, all hidden out of sight underneath the wooden tables. Main post image updated to reflect the latest.

Additionally, 
> And it only really adds a bunch of open, rather empty space in the public garden. As nice as this looks in dmm, standing on one side you will not be able to see the other, by quite a number of tiles.

I freely admit that this is kinda intentional. It's _meant_ to have open-ness and feel like a large area as a place intended to counteract what crew stationed here see every single day: steel walls in 3-wide hallways. I acknowledge that it is big, but the Public Garden being compact is one of the reasons I started making this change in the first place.

Iroquois-Pliskin:
With the addition of further atmos objects I've got no real issues anymore.

ResidentCody:
All window tiles have firelocks, yes. They are full grille/reinforced window/reinforced pane/firelock sets.

Iroquois-Pliskin:
> All window tiles have firelocks, yes. They are full grille/reinforced window/reinforced pane/firelock sets.

Based. Full support.

DemicusMaximus:
When I first heard of this plan, I was against it. But looking at it... I think I kinda like it. It gets a yes from me.

Very-Soft:
I am torn on this. I like it but also, indeed, it's massive. It being massive like this actually undoubtedly is going to cause a few problems. 

One being, as Pip suggested, it's big enough that you could have several groups of people in here while not having them on the screen, which indeed, was the problem with the old bar. 

AND. Practicality. The water and machines are REALLY REALLY far from the plants. If you actually want to do plant things without it being like, your responsibility, which is what this place was added for, it's gonna be a large pain in the ass. I can tell you from experience, it's already a bit of a pain. This change would probably make it kind of unbearable.

I don't mind so much the width of this change, but the tallness is too extreme. It's 21 tiles tall in places, which is about a full screen and a half. It should be less than 15 tall, considering it is slightly more than 15 wide. Basically, the philosophy I prefer to go with for social spaces like this, is that they should not be so big that one can get away from it being busy by moving across the room. Which, this would allow. I don't want the whole station to be able to sit in the same room without it being really busy in that place.

ResidentCody:
New adjustment, now showing Squashed Version, hopefully a friendly middle ground between the two extremes. Compare and discuss.
- Verticality squashed in response to "Too Big" critiques, dimensions at longest points are now 16x18. 
- -5 soil beds (returning to the original 8 total), repositioned to be on the right side nearer to vendors/supplies.
-  70% less amogus, because apparently somebody realized that the tall version looked like a crewmate. Stupid pattern recognition.
- Removed/reworked exterior catwalk section, removed most of them along the north as the extra outside turf provides ample safe walking space.

Very-Soft:
Much better!

Enzo-Leon:
@ResidentCody As a heads up when the map PR is merged here there were SMALL fixes to Surface three since I moved the nuke disk down to it following the removal of the vault. Be sure to save your draft of this map, at least of your project area here, so if you need to re-do it it's not 100% scuffed.

TheFurryFeline:
Updoot time here. Interesting, though, seems the merged asteroid squish didn't conflict with this. Yay.

TheFurryFeline:
Also, not sure if my Syringe Gun move PR will affect this, but maybe? Darn shuttle turfs may be a slight issue, but we'll see.

ResidentCody:
Not sure why it would, this area has precisely nothing to do with anything pertaining to any syringe guns and it had no issue with the minor changes brought about by the Asteroid Squish. Of course what do I know, it may actually have some cursed spaghetti interaction that I'd never consider.

TheFurryFeline:
Just realised you may wanna update the OP post for the squished version of the upgraded Garden in that it's not as tall.

ResidentCody:
There, main post image updated to latest "not as expanded" version image.

Novacat (approved):
Has a conflict but is fine otherwise, going to merge as soon as conflict is resolved.

Ketrai:
I can only say it feels a little weird to have the planting plots tucked away in the corner there. On most servers and the way the garden was structured before, plant plots serve both as decoration and practical use for the less dedicated botanists. Especially nice with a couple flowers. I'd spread them out a bit across the space. Water proximity shouldn't be a huge concern.

ResidentCody:
Conflict resolved, ready to go.

> Honestly the main issue with watering in general is that no one has bothered making it easier to fill up at water tanks yet. Unless I'm missing a secret technique to fill a bucket from them instantly.

Right click the bucket, set transfer value to desired number (120u is max) and now it'll fill/dump that much. Also use a sink, it's a stationary but infinite water source.

Ketrai:
> Conflict resolved, ready to go.
> 
> > Honestly the main issue with watering in general is that no one has bothered making it easier to fill up at water tanks yet. Unless I'm missing a secret technique to fill a bucket from them instantly.
> 
> Right click the bucket, set transfer value to desired number (120u is max) and now it'll fill/dump that much. Also use a sink, it's a stationary but infinite water source.

The idea of using water tanks is that they're portable. Setting the bucket to 120 u works for sinks. But you'll still only draw 10 u per time from watertanks.

PR #10229, opened 2021-04-28
Exploration Expansion: First POI

SleepingZi:
This is a test merge for a new POI to make sure I'm doing this correctly.

This adds a new POI to V4, the Speakeasy, a place to chill from the long day of exploring.
unknown

### DIFF
--- a/maps/submaps/surface_submaps/mountains/mountains.dm
+++ b/maps/submaps/surface_submaps/mountains/mountains.dm
@@ -42,6 +42,7 @@
 #include "excavation1.dmm"
 #include "spatial_anomaly.dmm"
 #include "speakeasy_vr.dmm"
+#include "Campsite_vr.dmm"
 #endif
 
 // The 'mountains' is the mining z-level, and has a lot of caves.
@@ -365,7 +366,7 @@
 	cost = 20
 	fixed_orientation = TRUE
 
-/datum/map_template/surface/mountains/normal/Speakeasy //VOREStation Edit
+/datum/map_template/surface/mountains/normal/Speakeasy //VOREStation add
 	name = "Speakeasy"
 	desc = "A hidden underground bar to serve drinks in secret and in style."
 	mappath = 'maps/submaps/surface_submaps/mountains/speakeasy_vr.dmm'
--- a/maps/submaps/surface_submaps/mountains/mountains.dm
+++ b/maps/submaps/surface_submaps/mountains/mountains.dm
@@ -41,6 +41,7 @@
 #include "Cliff1.dmm"
 #include "excavation1.dmm"
 #include "spatial_anomaly.dmm"
+#include "speakeasy.dmm"
 #endif
 
 // The 'mountains' is the mining z-level, and has a lot of caves.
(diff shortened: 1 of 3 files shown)


ShudderEve:
What about the existing beach bar + kitchen + shore area? Are those not enough? I'd have thought that V4 PoIs would be more exploration and puzzles than more bars.

dwinters99:
>What about the existing beach bar + kitchen + shore area? Are those not enough?

I would say that more places to hang out aren't bad at all. 

I would also say, in my opinion, that it would probably be a nice to have this PoI be in a partial state of disrepair. People have responded very well in the past to things needing construction such as the debris field ships. Also, this does seem a tad blocky, but that's an easy issue to fix.

SleepingZi:
> I would also say, in my opinion, that it would probably be a nice to have this PoI be in a partial state of disrepair. People have responded very well in the past to things needing construction such as the debris field ships. Also, this does seem a tad blocky, but that's an easy issue to fix.
I see what you mean here, however I like the finished state a little more, it gives the 'still in use' vibe that I'm trying to go for. I want this to be something that gives the caves some notion that they're still in use for various purposes, and not just a place of abandoned structures and hostile creatures.
I do see what you mean by it being blocky, and I have to agree, though I don't see a way to fix that without making the POI larger.

SleepingZi:
Revision 1.1: Made structure more 'natural' and added a jukebox to get the 2220's vibe

maps/submaps/surface_submaps/mountains/mountains_areas.dm
```diff
@@ -163,3 +163,7 @@
 /area/submap/spatial_anomaly
 	name = "POI - Spatial Anomaly"
 	ambience = AMBIENCE_FOREBODING
+
+/area/submap/Speakeasy
```
Novacat, line 167: //VOREStation Add

maps/submaps/surface_submaps/mountains/mountains.dm
```diff
@@ -363,3 +364,11 @@
 	mappath = 'maps/submaps/surface_submaps/mountains/spatial_anomaly.dmm'
 	cost = 20
 	fixed_orientation = TRUE
+
+/datum/map_template/surface/mountains/normal/Speakeasy //VOREStation Edit
```
Novacat, line 368: This should probably be changed to //VOREStation Add

Novacat (changes requested):

maps/submaps/surface_submaps/mountains/mountains.dm
```diff
@@ -41,6 +41,8 @@
 #include "Cliff1.dmm"
 #include "excavation1.dmm"
 #include "spatial_anomaly.dmm"
+#include "speakeasy_vr.dmm"
+#include "Campsite_vr.dmm"
```
Novacat, line 45: I think you added this by accident?

Novacat (changes requested):

maps/submaps/surface_submaps/mountains/mountains.dm
SleepingZi: yes, that was a result of me using upload to make a pull, I've fixed the issue.

Novacat (approved):

PR #10198, opened 2021-04-23
Food for Thought

klaasjared:
Tired of the lack of vending machines that offer delicious food options? Sick of Getmore and their tiresome slogans? Well here is the PR for you!

This PR adds several Baystation vending machines and their food items to Polaris as well as a brand new machine to aid in bloat removal. Below is a lenghty list of all current food vendors and the snacks taht each will carry.

Getmore Chocolate Corp

-Tau Ceti Bar
-Hundred-Thousand Thaler Bar
-Aerostat Bar
-Lars' Saltlakris
-Andromeda Bar
-Mocha Crunch
-Cronk Bar
-Shantak Bar
-Grandma Ellen's Hard Candy
-AlliCo Gummies
-BBQ Chips
-Chips
-4no Raisins
-Gum
-Admints
-Skrellsnax 
SnL Co. Snacks

-Tuna Snax
-BBQ Chips
-Chips
-Pistachios
-Smeki
-Salo
-Vobla (dried fish)
-4no Raisins
-Snack Cake
-Cheesie Honkers
-Syndi-Cakes
-Carps Ahoy! miniature cookies
-Bread Tube
-Beef Jerky
-Beef Jerky (True)
-Skrellsnax 
SweatMAX

-ChewMAX fruit bar
-SwoleMAX protein bar
Sol-Snacks

-Tidal Gobs (Triton)
-Saturn-0s -SPRITE DONE
-Jove Gello
-Plutonian Rods
-Frouka (Mars)
-Venusian Hot Cakes -SPRITE DONE
-Oort Cloud Rocks
-BBQ Chips
-Chips
-Skrellsnax 
-Gum
-Admints
-Sun Snax (Spicy chips)
-Appleberry Bits (Canned Cereal)
Nippon-tan!

-Rice Cake
-Pokeys
-Red Alert Nuts
-Choco Banang
-Hanami Dango
-Goma Dango
-Kaju Mamma! Bar
-TaroMilk Bar
-Calamari Crisps
-Skrellsnax 
-Ramen
-Namagashi
-Dorayaki
-Wasabi Peas
-Daifuku
Snix

-Suhariki (croutons)
-Canned Caviar
-Canned Caviar (True)
-Pistachios
-Smeki
-Salo
-Vobla (dried fish)
-Calamari Crisps
-Gum
-Admints
-Skrellsnax 
-MAPS
-Hakarl
-Pretzels
[...]

PR #10230, opened 2021-04-28
Brig Flashes

### LINKED ISSUE #10014: Surfacesec cells have no wall flashes, not in holding or in proper cells.
Not a bug but a suggestion. Security cells should, if not MUST have wallflashes to deal with bad prissoners.

Have greytider doing dumb stuff, click flash, greytider on floor to make life not cancer for officers. Instead we have no flashes, making dealing with prissoners dangerous for the status of your job banned-ness because of SOP.

85674dc3f51a4a1d5de87a1c105e6afdc314f14b

TheFurryFeline:
🆑 
maptweak - Adds mounted flashers to both brigs and in the visiting rooms brig-side.
/🆑 
Resolves #10014 
 Note, am fine if this ain't wanted. Another note, on testing, there's a substantial cooldown on the button so ya can't spam it too much.
 
 Another note: could conflict with #10198

### DIFF
--- a/maps/tether/tether-01-surface1.dmm
+++ b/maps/tether/tether-01-surface1.dmm
@@ -24363,6 +24363,13 @@
 	},
 /turf/simulated/floor/wood,
 /area/crew_quarters/sleep/Dorm_6)
+"aOO" = (
+/obj/machinery/flasher{
+	id = "SurfaceBrigFlash";
+	pixel_y = -20
+	},
+/turf/simulated/floor/tiled/dark,
+/area/tether/surfacebase/security/brig)
 "aOP" = (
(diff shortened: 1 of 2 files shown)


ShudderEve:
I don't understand why they weren't there in the first place. The brings have flashers that you can control with either a button or from the time-screen on most servers.

TheFurryFeline:
> I don't understand why they weren't there in the first place. The brings have flashers that you can control with either a button or from the time-screen on most servers.

Yeah, literally had to have Polaris's map open to compare. Probably at one point we did have 'em but got removed. Or maybe never had 'em? I dunno, but now this is here as a suggestion.

TheFurryFeline:
Huh, surprised this didn't need an admin discussion or something. Yay

PR #10234, opened 2021-04-29
Bringing Back the Bepsi

klaasjared:
Way back when - there was a soda vendor that Virgo used that sold Bepsi. It was phased out when we ported the vending code from Polaris. I found this sprite and I've decided to breathe new life into this dusty sprite. I even added some new soda cans to sell in the vendor that are Virgo specific. 

Everything that I've added is displayed below.

vr_can_showcase

vr_bepis_showcase


pontifexminimus2:
Bepis...

PR #10239, opened 2021-04-30
Fixes #10237

### LINKED ISSUE #10237: Character creation is borked for tails/ears/wings
The character setup panel's tails, ears, and wings are no longer able to be changed at all since the recent update #10223 

To be able to change a character's tail and or keep a pre-existing character's tail.

#### What actually happened
I've now got a tail-less, earless red panda.

#### Steps to reproduce
- (Step 1)
- Load up character setup and attempt to adjust the ears style for something more cute.
- (Step 2)
- Suffer as it fails to load anything on all fronts due to defaulting to "normal".

Heroman3003:
Title. No edits because it probably should be backported at some point... Just later maybe

Fixes #10237


MistakeNot4892:
Would suggest https://github.com/PolarisSS13/Polaris/pull/8057 as an alternative to this, as discussed on Discord earlier.

PR #10240, opened 2021-05-01
Removes access loss on ID contamination/digestion.

Heroman3003:
More proposal than anything. Does what it says, IDs no longer lose their access on digest_act or contaminate_act. They still change appearance to scuffed and half-digested. Honestly, I still don't know why it was a good idea in the first place. Especially with how many workarounds had to be added around that anyway, to the point where it's very rare to see it even get contaminated in the first place, but is very annoying when it does.

ID restoration consoles are NOT removed with this to avoid conflicting with existing mapping PRs, but are reduced to basically decor pieces. If this PR is accepted, then once map stuff is resolved, ID restorers will be removed as separate action.

### DIFF
--- a/code/game/machinery/computer/card.dm
+++ b/code/game/machinery/computer/card.dm
@@ -200,7 +200,6 @@
 					modify.access -= access_type
 					if(!access_allowed)
 						modify.access += access_type
-				modify.lost_access = list()	//VOREStation addition: reset the lost access upon any modifications
 			. = TRUE
 
 		if("assign")
@@ -225,7 +224,6 @@
 					modify.access = access
 					modify.assignment = t1
 					modify.rank = t1
-					modify.lost_access = list()	//VOREStation addition: reset the lost access upon any modifications
 
 				callHook("reassign_employee", list(modify))
 			. = TRUE
@@ -283,7 +281,6 @@
 			if(is_authenticated())
 				modify.assignment = "Dismissed"	//VOREStation Edit: setting adjustment
 				modify.access = list()
-				modify.lost_access = list()	//VOREStation addition: reset the lost access upon any modifications
 
 				callHook("terminate_employee", list(modify))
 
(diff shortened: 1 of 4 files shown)


Iroquois-Pliskin:
On the surface this seems pretty good but my singular concern is if someone tries to abuse the fact that they digested a head to get something fancy.

Otherwise sure why not.

Heroman3003:
That is already kind of punishable OOCly, plus vast majority of prey just store IDs in a way that prevents them being contaminated anyway. And even if they do, I don't think that's way too critical or balance-destroying, unless its used to do griefy rulebreaky stuff, which, again, is just to be punished OOCly. If someone digests HoP and uses the newly found access to rename their job into Ass Padding or to set Ian free, then good for them, imo.

I always thought ID contamination has never been anything more than annoyance for everyone involved and I will stand by it. Sure, maybe if vore was like a mechanic on antag-based server where it's primarily used as tool of mechanics, rather than a supporting structure for RP, maybe contamination would make sense, but it doesn't make any on VOREStation the way it is, IMO.

Iroquois-Pliskin:
While it's punishable OOC, sometimes things slip through the cracks, but if IDs are being stored in a manner that bypasses this anyways then fair enough.

If I had to guess from an outsider's perspective maybe it was meant to be a trophy thing.

lm40:
yes, i fully agree with this change, it never made any sense to me. like if all you want is to nick an id it's not like you can't strip someone as part of your scene, lmao.

Screemonster:
It'd be neat if this were a config or something but I think the servers that run antags also have rules forbidding the use of vore for antaggery anyway.

Wickedtemp:
Yeah, scrap it. There's no point behind the access breaking.

DemicusMaximus:
I quite like this. And while I feel it will increase admin work load slightly, I personally like preds abusing their preys ID's so it gets a yes from me.

Novacat (approved):

Arokha:
I believe this was originally mandated by Ace back in yon day, and that's why it was a thing.

PR #10241, opened 2021-05-02
Makes soulcatcher a pre-installed software

Heroman3003:
Title. NIFs loaded from savefiles will have soulcatchers included same way as commlink. Newly installed ones still won't, also same way as commlink.

Also makes soulcatchers NOT catch prey by default to counteract accidents that could be brought upon with this change. Now both catching self AND catching prey needs to be turned on if you want to do that. Other settings have same defaults.

### DIFF
--- a/code/modules/nifsoft/nif.dm
+++ b/code/modules/nifsoft/nif.dm
@@ -88,8 +88,9 @@ You can also set the stat of a NIF to NIF_TEMPFAIL without any issues to disable
 				qdel(src)
 			return FALSE
 		else
-			//Free commlink for return customers
+			//Free commlink and soulcatcher for return customers
 			new /datum/nifsoft/commlink(src)
+			new /datum/nifsoft/soulcatcher(src)
 
 	//Free civilian AR included
 	new /datum/nifsoft/ar_civ(src)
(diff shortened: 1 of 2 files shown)


DemicusMaximus:
So if I am reading this right, if they have a soulcatcher it'll default to being there the next time. Otherwise it does not. Right?

If that's so, than this seems like a good one to me. Just a little anti-frustration feature.

Heroman3003:
> 
> 
> So if I am reading this right, if they have a soulcatcher it'll default to being there the next time. Otherwise it does not. Right?
> 
> If that's so, than this seems like a good one to me. Just a little anti-frustration feature.

Nnnnot quite. It will work akin to the commlink, as in, you won't have it on shift you install the NIF on, but will automatically be given it on every shift afterwards. So even if you didn't ever have it before, you'd have it.

Novacat (approved):
This was suggested in the past but the issue was accidental soul catching happening prior. This fixes that.

Arokha:
~~The SC was basically the whole reason I made the NIF cough cough so ya~~

Seems OK to me, though the unsetting of the catching others flag might throw some people off. I get why you did it that way, and I think that's probably the right thing to do (there's only two options here, so).

Might be cool if you made it so the trash NIFs don't include one, and the normal (and higher) do.

PR #10242, opened 2021-05-02
Reduces cost of Sci and Eng AR programs.

Heroman3003:
Title. They are practically no different from Civilian AR bar the name. Shouldn't cost any more than that either then. Also adjusts otherwise misleading description of engineering AR, as it does NOT in fact provide the alerts (need separate program for that).


Iroquois-Pliskin:
Good PR.

Little-119:
I'm unsure of why these exist in the first place, to be honest

PR #10261, opened 2021-05-06
[MIRROR] Personal Crafting & Dependencies

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/8060
--------------------
This adds tgstation/tgstation#17341 and converts the stunprod over to use it as a proof of concept. 
This should eventually replace most other crafting systems that involve smushing stuff together, like the entire cooking system/etc.
Also replaces the is_x tool helpers with a generalized trait system, though the old procs are still there because effort.


Novacat:
Duplicate of an existing PR

PR #10266, opened 2021-05-07
Implement planetary solar behavior

Arokha:
Adds logic for solar panels on planets to behave differently (not using SSsun).

They base their power on the angle to the sun. I can't make the panels aim 'up' and tilt in a cool manner (assuming you're always on the equator and the sun passes overhead), so instead 'north' is where the sun is at noon, and all the 'south' angles (from 90 to 270, assuming 0 is north) are where it is on the other side of the planet. So day is from 270 to 90 basically, with noon being 0/360.

Sun brightness affects power output, so this means on v3b you get slightly sub-optimal power output since it's a bit darker than somewhere like v4, which will produce more power with solars. Bonus points if you figure out a way to build an array on v4 and mail the power back to the station.

Weather also affects solar output, so when overcast and whatnot it's decreased. I don't know if there's weather that makes the sun BRIGHTER but that would also work.

The goal would be to have the power output of the v3b solars produce the same amount of power on average after this PR as they do before, so it will require some tweaking to numbers and testing, and probably mapping in some extra panels since I suspect there will be too few to keep the same power output.

### DIFF
--- a/code/modules/planet/planet.dm
+++ b/code/modules/planet/planet.dm
@@ -67,3 +67,7 @@
 	sun["color"] = new_color
 	needs_work |= PLANET_PROCESS_SUN
 
+/datum/planet/proc/get_sun_angle()
+	var/relatime = current_time.hours_until_noon()
+	var/pm = (relatime <= 0)
+	return (180 + sun_position * (pm ? 180 : -180))
--- a/code/modules/planet/time.dm
+++ b/code/modules/planet/time.dm
@@ -42,6 +42,27 @@
 /datum/time/proc/make_random_time()
 	return new type(rand(0, seconds_in_day - 1))
 
+// 'day' as in not hours since the epoch
+/datum/time/proc/get_day_hour()
+	var/day_seconds = seconds_stored % seconds_in_day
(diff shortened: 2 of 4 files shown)


Arokha:
> As long as general power production values of V3B solar arrays don't actually decrease that's fine

I said that's the goal in the PR!

>  V3B is FAR from 'less bright' than V4

I think that's probably down to the color. Despite the appearance, noon on V3B has a brightness of '7' and V4 has a brightness of '10'. Or maybe that's enough of a difference to say that? I dunno! But they're not so different. Not even 50% different.

Greenjoe12345:
I'd guess this is being ported up to polaris? will be quite useful for us up there too

Arokha:
> I'd guess this is being ported up to polaris? will be quite useful for us up there too

Very likely, yup. Though I need someone to test these values and tell me how the work out. I ain't no engineer.

Arokha:
I'm not planning on finishing this at this time. Someone else can handle it!
In the end, it would be easier to get the lumcount to do solar stuff, or ask the sun for brightness, now that we've rewritten like everything about it.

PR #10270, opened 2021-05-09
Add phoron and nitrogen breather traits, and water breather trait

Little-119:
Adds the phoron and nitrogen breather traits, which both make one breathe phoron/nitrogen and make oxygen be toxic to them (instead of phoron). Of course, if one has these traits, they spawn with a breath mask and a full-size tank of the gas they need to breathe.

Also adds the water breather trait. Which just makes it so you can lay down in water without suffocating. Not nearly as interesting, maybe not even worth a point, only included in this PR because I decided to make some new traits and this is all I came up with.

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/blank_vr.dm
+++ b/code/modules/mob/living/carbon/human/species/station/blank_vr.dm
@@ -80,6 +80,3 @@
 			H.internal = H.r_hand
 			if(istype(H.internal,/obj/item/weapon/tank) && H.internals)
 				H.internals.icon_state = "internal1"
-
-	for(var/datum/trait/T in traits)
-		T.equip_survival_gear(H)
--- a/code/modules/mob/living/carbon/human/species/station/blank_vr.dm
+++ b/code/modules/mob/living/carbon/human/species/station/blank_vr.dm
@@ -56,7 +56,8 @@
 //Called when face-down in the water or otherwise over their head.
 // Return: TRUE for able to breathe fine in water.
 /datum/species/custom/can_breathe_water()
-	return ..()
+	return /datum/trait/positive/water_breather in traits
+
 
 //Called during handle_environment in Life() ticks.
 // Return: Not used.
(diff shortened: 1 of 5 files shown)


ItsTheKet:
You bet your ass I am happy to see this happening.

Wickedtemp:
Okay uh... This is when it'd be super helpful to have some of these traits pop up by default in medical records, or like... an automated popup in the resleeving console if one of these traits is detected. Otherwise people are gonna be resleeved without whatever tank they need, and I'm probably not gonna be able to find a phoron or nitrogen tank quickly enough. 

Also, do those with these traits get any benefit from phoron/nitrogen injections? That's a thing for Vox, so I'd recommend it.

Little-119:
> Okay uh... This is when it'd be super helpful to have some of these traits pop up by default in medical records, or like... an automated popup in the resleeving console if one of these traits is detected. Otherwise people are gonna be resleeved without whatever tank they need, and I'm probably not gonna be able to find a phoron or nitrogen tank quickly enough.

Yeah, I agree, though that's probably out of the scope of this PR

> Also, do those with these traits get any benefit from phoron/nitrogen injections? That's a thing for Vox, so I'd recommend it.

Probably not. Maybe for the phoron if it checks for the breath_type being phoron and not the species itself.

KillianKirilenko:
> Yeah, I agree, though that's probably out of the scope of this PR

For now, put a big old "NOTE THIS IN YOUR MEDICAL RECORDS" disclaimer when taking the trait. If people don't read it, don't add it to their records, and then complain when Medical don't realize they needed special care procedures it's on them not on Medical.

> > Also, do those with these traits get any benefit from phoron/nitrogen injections? That's a thing for Vox, so I'd recommend it.
> 
> Probably not. Maybe for the phoron if it checks for the breath_type being phoron and not the species itself.

I think phoron checks `species_tag` or something like that. You could do something similar quite easily by adding new species tags with the traits and/or by checking breath_type but either way you'd need to go through and edit a bunch of reagents.

Wickedtemp:
Wouldn't be hard to fluff it as "tanks are preloaded  around the base of the resleeving pod and are automatically put on". Its not the most crazy thing out there. I'd be totally fine with that.

Little-119:
Considering how probably most resleeves are just the aftermath of vore-themed snuff scenes and apparently something people just want to get over with, I'm opting for the most ~~lazy~~ convenient option and just automatically equipping breathing equipment for phoron/nitrogen breathers when they're created in the grower pod

Little-119:
My other ideas were:
- have the grower pod draw from a limited quantity of tanks actually stored in the machine
- add the functionality where nitrogen and phoron reagents benefit nitrogen/phoron breathers when injected, have the grower pod inject the new sleeve with a bunch of those reagents to give time for the doctor to give them a breathing mask and tank, which I guess would be supplied nearby

I also considered having the growing pod equip the smaller emergency tanks but actually those are kind of better because they fit in pockets instead of having to go on the back. ...this is going to be a rather inconvenient trait to have, isn't it?

Wickedtemp:
The first option is probably best. If there's no warning beforehand, what's likely to happen is the person gets resleeved, starts suffocating and taking toxins, the doctor isn't geared and has no idea anything is even going on until they see the health-bar change and might not know what to even do. Plus, how would we even know which tank to give them?

When it comes to Nitrogen and Phoron healing oxy, the values may need to be tweaked. Phoron, as a 'toxin', metabolizes much more slowly than the default metabolism rate. I think it was originally intended to more or less match Dexalin's healing rate, but if I'm honest, Dexalin sucks and isn't that useful if you want to actually save someone. Outside the scope of this PR, but yeh.

Little-119:
Is there anything else I need to do on this pr that the maintainers are waiting for?

Novacat (approved):
Looks fine code-wise, but may need to be discussed by staff?

Little-119:
I think I had a dream where this was merged. As well as a non-existent PR that I had marked as DNM but neglected to finish for a month

PR #10278, opened 2021-05-10
Unrestricts sprite accessories by default

Verkister:


Novacat (changes requested):
This should not be needed since they should already be unrestricted in the _vr files?

Verkister:
Yeah, turns out the _vr overrides with the species lists were missing prommies and customs

PR #10281, opened 2021-05-10
Move Syringe Gun

### LINKED ISSUE #10274: Surface medical syringe gun missing
**Brief description of the bug**

No syringe gun in surface medical

**What actually happened**

No syringe gun placed in medical

**Steps to reproduce**

Just go look in medical, no syringe gun.

TheFurryFeline:
Title. And yes, I know there's a mapping freeze, but I think this is so minor as to be fine.

🆑 
maptweak - Move the Syringe Gun from the grenade crate onto the table, remove extra box.
/🆑 

Resolves #10274

### DIFF
--- a/maps/tether/tether-03-surface3.dmm
+++ b/maps/tether/tether-03-surface3.dmm
@@ -29981,10 +29981,24 @@
 /turf/simulated/floor/tiled/white,
 /area/rnd/outpost/xenobiology/outpost_first_aid)
 "aZg" = (
-/obj/structure/shuttle/engine/propulsion,
-/turf/simulated/floor/reinforced,
-/turf/simulated/shuttle/plating/carry,
-/area/shuttle/tether)
+/obj/structure/closet/crate{
+	icon_state = "crate";
+	name = "Grenade Crate"
+	},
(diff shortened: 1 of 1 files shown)


Heroman3003:
Considering upper levels are on map freeze and are getting purged completely with next PR for them, maybe just adding new one into surface medical without removing it from upper one would be better?

TheFurryFeline:
As stated in the issue, the Surface Medical and Asteroid Medical both have a syringe gun. But the former was tucked away inside a grenade crate along with an extra box of cartridges. I've only just moved the thing out of the box. As for asteroid's syringe gun, I've left that alone, hence lack of an edit to the dmm for asteroid 2

Novacat:
This will conflict with the mapping freeze unfortunately.

TheFurryFeline:
Alas. Such a minor tweak, though.

Arokha:
If it conflicts, you'll have to redo it after freeze anyway, so no point in leaving it open. Gonna close for now, 'without prejudice' as it were.

TheFurryFeline:
Alrighty, then.

TheFurryFeline:
Can I get this re-opened?

PR #10298, opened 2021-05-11
Arokha/synthtweak

Arokha:
isSynthetic() on humans was severely overloaded with performing work when it's just an accessor function, so I moved the work to the appropriate places (and also reverted the lynx quick fix):

- Mechanical stomach provides the reagent purge verb (orchestrator on proteans)
- Diagnostic organ provides the self-diagnostics verb (orchestrator on proteans)
- Head organ provides monitor set verb if robotizing to a head that supports it, otherwise removes it
- Torso sets the `synthetic` var on owner to robolimb singleton (previous isSynthetic was handled by checking torso, so all cases of a human being isSynthetic() true should be the same: ones where torso is robotized)

### DIFF
--- a/code/modules/emotes/definitions/exertion.dm
+++ b/code/modules/emotes/definitions/exertion.dm
@@ -5,7 +5,7 @@
 	emote_message_3p = "is sweating heavily."
 
 /decl/emote/exertion/biological/check_user(mob/living/user)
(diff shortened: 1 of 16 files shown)


Novacat (approved):
This looks better but should be ported to Polaris to avoid future conflicts.

Arokha:
https://github.com/PolarisSS13/Polaris/pull/8077

PR #10302, opened 2021-05-12
Fix two character spawn bugs

Arokha:
- Fix missing nme/nsay verbs for using with soulcatcher
- Fix missing *vwag emote for those with *vwag tails

(Garbage NIFs don't start with any software anymore but I doubt anyone cares)


Little-119:
*vwag

PR #10304, opened 2021-05-12
Fix wrong radio path for talon

Very-Soft:
The game was crying about it so I fix


KillianKirilenko:
oops, I thought I caught all of those but I guess I didn't

PR #10309, opened 2021-05-12
Fixes Scream sounds

Novacat:


Arokha:
How is this different from the norm, and why's it a VR thing?

PR #10315, opened 2021-05-13
[MIRROR] Adds persistent smartfridges for garden products, material sheets

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/8042
--------------------
Adds framework for other persistent item storage mechanisms, hopefully it's abstracted enough that it should be fairly straightforward to use this feature for other storage media. The only major design choice is how to store the information, and while that seems daunting, it's pretty straightforward once you establish what information you need to persist.
Refactors persistence controller so it only initializes _named_ persistence datums, allowing unnamed datums to be used as abstract types.
Refactors base smartfridge type to a /produce subtype, so now the parent type is an abstract type (Typechecks are needlessly complicated if I have to exclude subtypes, but then use a subtype to specifically enable persistence (Other methods possible, this is the least work for mappers)). Reorganizes smartfridges into several files by general category.
Adds material sheet smartfridge, it stores each individual stack that you put into it (lazy but simple approach), but will merge/split them as appropriate when dispensing. Pulls from stacks in order they were put in, rather than searching for particular amounts. I'm pretty sure that's generally less overhead, but at least it's simpler than trying to implement a search function for particular stack sizes.

All numbers open for debate.

Persistent storage datums function by storing the type of the thing to spawn, and the number of things of that type to spawn, as individual tokens space-delimited, on the entry line for a single target storage object. Produce storage instead stores the seed name of the fruit, since that's where all the important information comes from. They can be configured to store up to some maximum amount of items in total, or a maximum per each type of thing.
The sheets storage as-written can store up to 50 sheets of each material (Not so much that you can in a single round stock it for a month, but enough that you can stock rarer materials for research).
Produce storage as-written can store up to 20 grown fruits total. It strictly checks the plant controller for what chems to put into the fruit, so xenobotany will find it totally useless for storing their god-apples. Ye olden mutant plants (eggplant etc) _can_ be stored, however, so those may see more use since you can more feasibly transfer them between rounds without needing chemistry or research toys. May require some sort of restriction in the future.

Sheet storage is mapped into Cargo Lobby, and is public-access
Persistent produce storage has been mapped into the place of the current garden/kitchen smartfridge, any access restrictions should remain in place.

This has been tested:
Persistence [before](https://puu.sh/HArRZ/ad3a0b7539.png) [after](https://puu.sh/HAsQl/c4e42616c0.png) (.png)
Sheet storage [operation(.mp4)](https://puu.sh/HvZq1/1bef1864bf.mp4)


Arokha:
Adds persistence for smartfridges across rounds

Arokha:
It's actually a subtype that has persistence so fine to merge and we can talk about it on a PR for mapping them in.

PR #10325, opened 2021-05-13
Add persistent smartfridges with lossiness

Arokha:
- Fixes up some persistence datum things, converting the files to JSON and fixing a few bugs
- Sticky notes have a 10% chance to fall onto the floor each shift
- Graffiti now keeps its icon so it looks the same between shifts
- Adds MAP_LEVEL_PERSIST so you can define persistent levels (like the talon), uses station_levels if none
- Creates some 'lossy' inter-round persistent smartfridges for sheet storage and produce storage.

Lossy sheet storage (for now):
- Up to 250 sheets per material persist (though it can hold 999 sheets total)
- Between rounds, if you have:
  - 1 stack or less, lose ~10 sheets
  - 1-10 stacks, lose ~1 stack
  - 10+ stacks, lose ~3 stacks

Lossy produce storage (for now):
- Up to 50 produce items persist between rounds (though it can hold up to 999 total)
  - 12.5% loss at random between rounds
- Only for basic produce, crazy xenobotany stuff will not persist

They are here:
StrongDMM_g7kmN1OJwo
StrongDMM_YGdVhQoFmX
StrongDMM_0tmyvqaHe9


Arokha:
Also yes I know blah blah map freeze. Fite me! The z-level I changed at the top of the tether isn't going to even *exist* after the squash, so it can't conflict. The one on z3 might, but I can fix it myself, just overwrite it, and the talon one isn't part of the tether.

Leshana (approved):

Arokha:
Upstream is https://github.com/PolarisSS13/Polaris/pull/8087

PR #10329, opened 2021-05-15
Talon + SME Edits & QOL

VictorZisthus:
Added a NIFshop and NIFsoft boxes to the Talon.

Added a preset console to the SME engine to allow borgs and the less financially well-off crew to access the supermatter monitor for more advanced engine setups without having to run back and forth between the CE's computer and the engine monitoring room.

Switched the Talon's engine to something more fitting for a spaceship and with a much easier to obtain fuel source. It even has a couple Radioisotope Thermoelectric Generators to power the ship through those long space voyages where the crew has to be in cryo! Updated signage to warn of the new radiation hazard.

Added a Cyborg Recharging station to the Talon Generator Core room. A last bastion for any synthetics in a zero-power situation and the top priority for the RTG's to recharge before their power gets to the SMES units.

### DIFF
--- a/maps/offmap_vr/talon/talon_v2.dmm
+++ b/maps/offmap_vr/talon/talon_v2.dmm
@@ -47,6 +47,28 @@
 	},
 /turf/simulated/floor/tiled/techmaint,
 /area/talon_v2/central_hallway/fore)
+"af" = (
+/obj/machinery/atmospherics/unary/vent_scrubber/on,
+/obj/effect/floor_decal/industrial/warning{
+	dir = 9
+	},
+/obj/structure/railing/grey{
+	dir = 8
+	},
+/obj/structure/table/steel,
+/turf/simulated/floor/tiled/techfloor/grid,
+/area/talon_v2/engineering)
+"ag" = (
+/obj/structure/railing/grey{
+	dir = 4
+	},
+/obj/structure/table/steel,
+/obj/effect/floor_decal/industrial/warning{
+	dir = 5
+	},
+/obj/item/weapon/paper/talon_shields,
+/turf/simulated/floor/tiled/techfloor/grid,
+/area/talon_v2/engineering)
 "ah" = (
 /obj/structure/cable/green{
 	d1 = 4;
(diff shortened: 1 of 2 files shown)


KillianKirilenko:
..right, NIFshop, I knew I'd forgotten *something*.

I'm not sure messing with the generators is a good idea. The potato puts out more power, has the advantage of more plentiful fuel, and lasts longer at max-safe, yes, but the RTGs put out barely anything (1250W each) and won't really even keep the lights on as the ship's idle draw is about 40kW*. The potato throwing out rads and the charger blocking the APC are also very questionable. Nevermind that rads mean instability for fbps last I checked, so the charger would be better placed in main engineering which also feeds straight off the generator.

*Seriously. One RTG can't even keep the *restrooms* running.

VictorZisthus:
> ..right, NIFshop, I knew I'd forgotten *something*.
> 
> I'm not sure messing with the generators is a good idea. The potato has the advantage of more plentiful fuel, and lasts longer at max-safe, yes*, but the RTGs put out barely anything (1250W each) and won't really even keep the lights on as the ship's idle draw is about 40kW**. The potato throwing out rads and the charger blocking the APC are also very questionable. Nevermind that rads mean instability for fbps last I checked, so the charger would be better placed in main engineering which also feeds straight off the generator.
> 
> *The safe/max for a tritium generator is 8/10, the safe/max for a potato is 4/5, with 12500 and 25000 per step respectively. The overall power output thus differs by precisely nothing, though the potato is much more fuel-efficient at the cost of being a radiation hazard whilst running.
> **Seriously. One RTG can't even keep the *restrooms* running.

Good observations! I thought the same when looking at the code. But as it turns out from testing, each RTG actually outputs 2.5kw instead of the expected 1.25kw. Spaghetti magic! With their powers combined they produce a respectable 5kw/h, enough to keep the core APC and the new recharger powered as long as the latter isn't under constant use.

I disconnected the engine room APC from the generator line because without SMES power or an actively running generator it depowers in minutes. The recharger itself wouldn't operate for long or be able to recharge with all of the other machines in the room pulling a whopping 20kw/h. The RTG's could never handle that, and there's no way to get them to prioritize the recharger itself in a zero-power situation.

I could easily move the new core recharger or APC if that's bothersome.

I also tested for radiation leakage, intentional and otherwise, both from the RTG's cores being exposed and from the potato. The potato itself only generates 3bq in intermittent bursts when running at 200kw/h generation, easily contained within the core itself. The maximum radiation leakage possible seemed to be from exposing both RTG cores for 120bq. Even then, radiation leakage only managed 13bq right outside the core hatch with moderate doses spreading throughout engineering and ending at 0bq before they hit the northern engineering wall. In short, even with intentional rad release only engineering was affected!

The cyborg recharger is capable of draining enough rads to keep a synth stable in the charger even at 60bq, so 3bq is nothing. Plus, that recharger is for a zero-power synario for when the round has run awhile without anyone setting up the potato, so the room should have 0bq in it to begin with.

Speaking of the potato, I found it's entirely possible for an organic or synth to waltz into the core unprotected, enable the potato, then walk away without getting hit with any radiation whatsoever. The potato is incredibly safe, far safer than what I was expecting when playing around last night.

If you have any further questions comments or concerns about these choices please let me know!

VictorZisthus:
image
New Core room layout.

VictorZisthus:
Looks like CI failed on account of my PR using code that it isn't aware exists. Should be good to force-merge!

KillianKirilenko:
> If you have any further questions comments or concerns about these choices please let me know!

There is one thing that needs fixing; the power documentation in the engineer's quarters; `maps/offmap_vr/talon/talon_v2.dm` lines 111 through 127.

VictorZisthus:
> 
> 
> > If you have any further questions comments or concerns about these choices please let me know!
> 
> There is one thing that needs fixing; the power documentation in the engineer's quarters; `maps/offmap_vr/talon/talon_v2.dm` lines 111 through 127.

Will take a look tomorrow!

PR #10340, opened 2021-05-16
Fixes ghosts breaking dropnoms

Verkister:
yeah


Novacat (approved):

Novacat (changes requested):
Actually, further down I have noticed there are redundant ismob and isliving checks that could probably be deleted?

Novacat (approved):

PR #10360, opened 2021-05-19
Update alienwhitelist.txt

Chemlight:
Simply put, adds Chemlight to the XenoChimera whitelist.

https://forum.vore-station.net/viewtopic.php?f=46&t=1955


Chemlight:
Although I'm unsure if I ran into a problem with the commit, this would be my first time doing something on vorestation.

If something is wrong do please correct me as I feel like I may have made a mistake or two.

Chemlight:
Actually considering this it's probably best I close this PR and try again from what someone is telling me.

PR #10366, opened 2021-05-20
Changes Print vore belly settings to X-print vore belly settings.

lbnesquik:
The reason for this change is that it this verb is really cool, but it also messes with speed typing 'Vore panel'
Given you use the 'Print' verb maybe once per two weeks, i added the X to make it show last.


lbnesquik:
> 
> 
> ~Make an export-vore-settings-to-json function~

image

PR #10370, opened 2021-05-21
[MIRROR] SSpersistence tweaks

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/8087
--------------------
- Convert output files to JSON instead of tab-delimited
- Adds MAP_LEVEL_PERSIST to allow specific maps to be persisted (if set on none, it will use station_levels)
- Adds a couple of 'lossy' sheet storage options (will not affect your existing ones)
- Fixes sticky notes and paper not aging at all
- Adds some small admin info lines for storage persistence (just coordinates and name)

More tangental:
- Converts smartfridge max storage var to not be a global so different types can have different values. Of all the vars to make global... this one? Really?
- Convert smartfridge New() to Initialize() due to how it was interacting with the timing of SSpersistence. It should be Init anyway.

FYI semantically a "token" is now a JS object literal loaded from the file. Previously it appeared to mean "cell" in the tsv file, or have mixed meanings in various places. I'm also not 100% sure it would persist more than one smartfridge before now? Maybe it would, I dunno. It does now.


code/modules/food/kitchen/smartfridge/smartfridge.dm
```diff
@@ -278,3 +279,285 @@
 			to_chat(usr, "<span class='warning'>Access denied.</span>")
 			return TRUE
 	return ..()
+=======
```
Arokha, line 282: Looks like line-endings diff

PR #10393, opened 2021-05-23
Update example configs from live server (sani)

Arokha:


config/example/config.txt
```diff
@@ -197,25 +167,20 @@ DEFAULT_NO_VOTE
...
-# RESPAWN_MESSAGE Remember to play a different character or something!
+## disable abandon mob
+# NORESPAWN
 
 ## disables calling del(src) on newmobs if they logout before spawnin in
 # DONT_DEL_NEWMOB
 
 ## set a hosted by name for unix platforms
-HOSTEDBY yournamehere
+HOSTEDBY Spades
```
Novacat, line 183: This should probably be changed on the live server.

Novacat (changes requested):

config/example/config.txt
Arokha: Don't request changes on a PR that has nothing to do with your requested changes

Novacat (approved):

PR #10395, opened 2021-05-23
Lore Fixy

TheFurryFeline:
"Proper" comment here: 5 centuries < 3 centuries = Virgo specific lore. One of many things that need to be fixed here.


Novacat (changes requested):
It would be easier to just override this in the _vr file.

Novacat (approved):

PR #10413, opened 2021-05-25
Expedition Medic Cargo Crate

KillianKirilenko:
Adds an orderable crate to cargo that contains two exploration medical voidsuits. 'nuff said.


Novacat:
The price will need to be changed more than likely when the voidsuit rebalance goes downstream but for now this will do.

KillianKirilenko:
> The price will need to be changed more than likely when the voidsuit rebalance goes downstream but for now this will do.

Probably. I just based it on the existing costs: single suit was 20, triple was 50, so 35 was the obvious middle ground for double.

PR #10414, opened 2021-05-25
Fixes pizzaguy uniform

### LINKED ISSUE #10404: Pizza Guy clothes shows no sprite when worn and cannot be rolled up/down
Attempting to wear pizza guy clothing (The hat is fine at least) does not update the worn sprites on the character.

Wear pizza guy clothes.

#### What actually happened
Not wearing anything at all.

#### Steps to reproduce
- Be you.
- Decide "It's pizza time."
- Manifest pizza guy clothes.
- Wear said pizza guy clothes.
- Cry as you are not with a fancy outfit.
- Attempt to see if rolling down the outfit might save your situation.
- Cry as that doesn't work either.

Heroman3003:
Fixes #10404


Novacat (changes requested):
Bad Icon Operation runtimes

KillianKirilenko:
> Bad Icon Operation runtimes

I suspect `index` expects a number, so trying to point it at the _vr file the way Hero has here won't work.

Might need to move everything to the /modules/vore/ subfolder with some of the other fluff stuff.

Heroman3003:
> 
> 
> > Bad Icon Operation runtimes
> 
> I suspect `index` expects a number, so trying to point it at the _vr file the way Hero has here won't work.
> 
> Might need to move everything to the /modules/vore/ subfolder with some of the other fluff stuff.

The main reason I did NOT do it with other vore clothing is because none of vore clothes use 'roll down sleeves/outfit' feature. Specifically because its just not compatible with way pretty much every _vr uniform is coded.

"vr" is 100% a viable index and issue actually lies elsewhere. I did a commit, let's see if it fixes it.

Novacat (approved):

PR #10422, opened 2021-05-25
Adds a fake sun object for mapper usage

Arokha:
If you put one of these on your map, it'll pick a random color/brightness from your choices, to pretend the map has a different time of day each time it's loaded. JUST PUT ONE THOUGH OKAY??? You can change that possible_light_setups to whatever you'd like for your options! Hooray.

Also randomly fixes ice thinking it has edges but it doesn't.


Very-Soft:
YAY! <3

vorestation.dme
```diff
@@ -2425,6 +2425,7 @@
 #include "code\modules\lighting\lighting_area.dm"
 #include "code\modules\lighting\lighting_atom.dm"
 #include "code\modules\lighting\lighting_corner.dm"
+#include "code\modules\lighting\lighting_fake_sun.dm"
```
Novacat, line 2428: This should probably have _vr appended to the end so I know it is a Vorestation specific file.

Novacat (changes requested):

vorestation.dme
Arokha: So piiiiicky

Novacat (approved):
The snow thing should probably be ported to Polaris but that is very smol

PR #10430, opened 2021-05-26
Add secret zlevels

Arokha:
Adds a landmark that allows you to mark zlevels that should not allow ghosts on them. Mobs/areas there will also not show up in the jump lists.


Very-Soft:
C: <3

PR #10435, opened 2021-05-26
Seperates chirp and teshchirp

Very-Soft:
Teshchirp was overriding chirp for non tesh and me no like so tesh chirp is tchirp now

also hypno and mrow were missing keys, so I added keys to them


Novacat (approved):

Little-119:
what if you' `*chrip`ed and it just played C418 - Chirp

PR #10437, opened 2021-05-26
Nerfs Storage Briefcases

Novacat:
Station ML3M/NSFW briefcases are now itemsize_huge (cannot fit in bags)
Non-station multi-cells are itemsize_large (takes up twice the current amount of space)

### DIFF
--- a/code/modules/projectiles/guns/energy/cell_loaded_vr/cell_loaded.dm
+++ b/code/modules/projectiles/guns/energy/cell_loaded_vr/cell_loaded.dm
@@ -242,7 +242,7 @@
 /obj/item/weapon/storage/secure/briefcase/nsfw_pack_hybrid
 	name = "hybrid cell-loaded gun kit"
 	desc = "A storage case for a multi-purpose handgun. Variety hour!"
-	w_class = ITEMSIZE_NORMAL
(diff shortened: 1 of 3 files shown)


Wickedtemp:
So...instead of balancing, you're just making it too much of a hassle to carry around entirely?

TheDavestDave:
as is they are a sercure breifcase you can put in a bag. You can't do that with normal sercure breifcases.

the cell nerf might cause a problem

Novacat:
They were not really supposed to be something you carried around.

dwinters99:
What is the reason for this other than "balance"? Is there some rogue element using these briefcases to powergame? Breaking into offices to steal the HoS's unique briefcase? A MUCH better way to implement this would be making the briefcases only be able to hold specific items (i.e. the gun and their cells + the mag maybe). This seems like a balance change that doesn't fix anything but makes things less convenient.

Wickedtemp:
So people carrying the ML3M and spare cells around was an issue? Is there a reason why that is a problem?

Novacat:
Going to go with winters idea of just restricting the suitcases to just ml3m/nsfw and associated cells/mags.

Wickedtemp:
yay
The ML3M lives on

PR #10409, opened 2021-05-24
explo vender stuff

TheDavestDave:
adds a bunch of job gear to the explo vender. Suits, both normal phasers, catalougers, headsets, and now glow sticks, flares, and a new bluespace glow stick. A flare that doesn't run out and burn stuff more or less.

PR #10442, opened 2021-05-27
Orderable Replacement Job Gear

KillianKirilenko:
WIP/Proof of Concept for ordering lockers of replacement job gear via cargo, as suggested by @Novacat in #10409.

image

Currently uses the *actual* job lockers. Pros;
- Easy to maintain, access restriction(s) and contents always match whatever the server's local settings are.

Cons;
- Doesn't show contents when viewed at the cargo console.
- Can include items that may be better put in their own crates or simply not be available.

Wanted to PR this up on Polaris but their cargo console wasn't cooperating when I tried to test it, something about missing templates.

Prices are placeholders. Contents and lockers may need to be totally overhauled (see cons, above).

### DIFF
--- a/code/datums/supplypacks/joblockers.dm
+++ b/code/datums/supplypacks/joblockers.dm
@@ -1,0 +1,202 @@
+/*
+*	Here is where any supply packs related
+*	to replacing a job's gear live.
+*/
+
+
+/datum/supply_pack/jobgear
+	group = "Replacement Equipment"
+
+
+/*
+TEMPLATE
+we don't actually need to 'contain' anything, using the proper containertype means it comes with all the stuff anyway!
+/datum/supply_pack/jobgear/
+	name = " Locker"
+	contains = null
+	cost = 30
+	containertype = 
+	containername = " Locker"
+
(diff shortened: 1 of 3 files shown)


Heroman3003:
Including some supposedly-very-unique pieces of job gear (like SM sword and RD harpoon) INCLUDING ones meant to be completely unique (CMO disk or CE blueprints) seems like a recipe for powergamey disaster of people just ordering those to get extra copies of a single piece of gear, especially with how cheap they are.

KillianKirilenko:
Yeah, I'm not sold on that either. As for the cost: I left the price at 30 as a placeholder. 'cause who knows, maybe staff will be fine with it if the head lockers cost 200+ points or something. /shrug

Rather throw it out, get the feedback, then tune accordingly.

Novacat:
Vanilla SS13 already has orderable replacement job gear crates, and usually the way it handles special equipment is to just not include them in the crates.

KillianKirilenko:
..no idea how I missed those. Closing this for now- I'll come back to the idea when I'm thinking straight.

PR #10444, opened 2021-05-27
Slopes! ◤◢◣◥

KillianKirilenko:
Adds some sloped bits to the talon, von braun (ERT ship), typhon (merc ship), and various small ships.

Also fixes a silly mistake @Novacat made on the ERT ship. ERT mapping units are not a valid substitute for important wires!


Novacat:
gdi my mouse must have misclicked.

PR #10450, opened 2021-05-27
Allow multiple SStranscore databases

Arokha:
So mercs can have their own separate one, or whatever.

### DIFF
--- a/code/controllers/subsystems/transcore_vr.dm
+++ b/code/controllers/subsystems/transcore_vr.dm
@@ -85,7 +85,8 @@ SUBSYSTEM_DEF(transcore)
 		src.current_run.Cut()
 		for(var/key in databases)
 			var/datum/transcore_db/db = databases[key]
-			for(var/datum/transhuman/mind_record/mr as anything in db.backed_up)
+			for(var/name in db.backed_up)
+				var/datum/transhuman/mind_record/mr = db.backed_up[name]
 				src.current_run[mr] = db
 
 	var/list/current_run = src.current_run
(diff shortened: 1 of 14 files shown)


code/controllers/subsystems/transcore_vr.dm
```diff
@@ -37,49 +44,60 @@ SUBSYSTEM_DEF(transcore)
...
 			return
 
 /datum/controller/subsystem/transcore/proc/process_backups(resumed = 0)
 	if (!resumed)
-		src.current_run = backed_up.Copy()
+		// Create a flat list of every implant in every db with a value of the db they're in
+		src.current_run.Cut()
+		for(var/key in databases)
+			var/datum/transcore_db/db = databases[key]
+			for(var/datum/transhuman/mind_record/mr as anything in db.backed_up)
```
Leshana, line 88: Looking at code elsewhere, I think  `db.backed_up` contains names (string) not mind_records.   The mind_record is the associated value.

Leshana (changes requested):

code/controllers/subsystems/transcore_vr.dm
Arokha: Fixed

code/modules/mob/living/carbon/human/life_vr.dm
```diff
@@ -42,8 +42,11 @@
 					if(istype(I,/obj/item/weapon/implant/backup))
 						if(!mind)
 							holder.icon_state = "hud_backup_nomind"
+						/* Pretty expensive with the addition of separate databases...
+						// is it really that important?
```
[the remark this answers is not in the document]
Arokha: Done!

PR #10458, opened 2021-05-28
sounds/emotes for spider-type characters

quixoticstargazer:
I was looking over the various emotes, and there's nothing for spider people, so I picked out a good-sounding wav from freesound.org, clipped out chittering and purring noises, adjusted the latter's volume (LUNA/Fizzle noted how it was barely audible), and edited in the emotes.

...

That's all that's needed, right?
The sound/voice files and audible_furry_vr.dm entries.

Sorry. I actually haven't really done anything with github before, though I'm learning. I was told that basically I should fork everything, put in the additions, and then do the pull request. Aside from editing the .dm file and putting the two sound files where they belong, nothing else has been edited.


Arokha:
Seems fine to me! Keys are a little long but could be changed in the future.

PR #10460, opened 2021-05-28
Another protean blobfix

Heroman3003:
blobbing needs sanity checks!


Arokha:
I'm sure there's a better solution I just can't think of one this second.

PR #10468, opened 2021-05-28
un-breaks unseen combat ai

Verkister:
haha lack of exit condition go brrrrr

kuva

### DIFF
--- a/code/modules/ai/ai_holder_combat_unseen.dm
+++ b/code/modules/ai/ai_holder_combat_unseen.dm
@@ -16,6 +16,8 @@
 			var/obj/O = find_escape_route()
 			if(istype(O))
 				return give_destination(get_turf(O), 0, TRUE)
+			else
+				return find_target()
 		else
 			return find_target()
 	else


code/modules/ai/ai_holder_combat_unseen.dm
```diff
@@ -16,6 +16,8 @@
 			var/obj/O = find_escape_route()
 			if(istype(O))
 				return give_destination(get_turf(O), 0, TRUE)
+			else
+				return find_target()
```
Arokha, line 20: I don't think this (or the other find_target()) are properly used here.

It appears that if there are no other targets in range, they'll set target to null, but NOT unset the last seen turf or change stance or anything. I'm pretty sure that this will only fix them attacking the ground in the case of there being other targets nearby. If there aren't, I think they'll just attack the ground forever.

Up on line 7, that condition is wrong, it should be <= world.time instead of >= world.time. That's what's keeping them attacking the ground forever.

Arokha: I made another PR to fix that part: https://github.com/VOREStation/VOREStation/pull/10471

Verkister: It's the melee_on_tile proc that keeps them spamming the attack, and since this part is underneath another else if, it wouldn't have been able to reach either return if there was no suitable O, so I think just flipping the sign would still keep them spamming the attack until the target times out.

Arokha: Ye though at least the time runs out after flipping the sign!
I wonder if give_target() should remove existing target stuff if you already have a target. Like as it is, it doesn't remove last seen turf or anything, that's all left over from the previous target, which is a bit naughty.

PR #10474, opened 2021-05-28
Small sprite fixes/tweaks

Verkister:
-Fixes synth lizard wag tail having messed up centering on the mirrored part of its wag. Also fixes it overlapping the legs from the side.
-Fixes female unathi having concave ass.
-Adds a blank state to ear file to avoid missing icons slapping squirrel ears on top of everything.


Arokha:
I see what you mean with the tail. Why's the whole thing move a pixel!

PR #10478, opened 2021-05-29
New item (Cyber Tech PCM)

VictorZisthus:
Fixed a typo in the Chameleon.dm

Added the Cyber Tech PCM, an invisible gas mask. This sensor suite conforms to the user and creates an invisible nano-weave around vital facial components without obstructing their sense of fashion or ability to eat. Comes at a cost.


KillianKirilenko:
Not sure I'm getting the point of an FBP-exclusive *gasmask*. They don't need gas filtration or an air supply last I checked, so the only point would be eye protection? Plus if you want invisible eye protection, there should be NIFsofts for that already, no? 🤔

You also consistently mispelled "nanoweave" in all of the descriptions by dropping the a.

dwinters99:
I'm also unsure as to the purpose of FBPs needing gas-masks. Unless of course you mean humans with synthetic heads, but that can't be made in the character editor and I'm unsure if the surgery saves between rounds or if it even exists.

VictorZisthus:
> Not sure I'm getting the point of an FBP-exclusive *gasmask*. They don't need gas filtration or an air supply last I checked, so the only point would be eye protection? Plus if you want invisible eye protection, there should be NIFsofts for that already, no? 🤔
> 
> You also consistently mispelled "nanoweave" in all of the descriptions by dropping the a.

Whoop, that's copy-paste for ya.

And no, sadly the Nifsoft does not protect from toxins, just vacuum. Or at least it didn't back when I tested it ages ago. I'll have to test again though!

VictorZisthus:
Removed TODO's, ready for review.

Novacat (approved):
Code wise is fine but I would like to run it by staff in the next staff meeting.

JoanRisu:
Talked about during our meeting. This is good as a loot item that people find out in the field. We don't think this should be a loadout or research item, however, due to the benefits afforded by this item. Those benefits are why we think it would be really good as a loot item.

VictorZisthus:
> Talked about during our meeting. This is good as a loot item that people find out in the field. We don't think this should be a loadout or research item, however, due to the benefits afforded by this item. Those benefits are why we think it would be really good as a loot item. 

Sounds good! I'll have to see about making it so you can change the filter with a screwdriver or multitool in hand to counter it's rarity then.

Arokha (changes requested):
Fix merge conflicts and implement requested changes from Joan

PR #10479, opened 2021-05-29
Ports Runechat

Arokha:
nHh9Qn2yqd
B9RQrk1izR


SigholtStarsong:
This will, indeed, probably be a mistake.

But a welcome mistake, to be sure.

dwinters99:
Make it opt-in. Otherwise, very good.

code/modules/emotes/emote_mob.dm
```diff
@@ -192,6 +194,7 @@
 					if(isobserver(M))
 						message = "<span class='emote'><B>[src]</B> ([ghost_follow_link(src, M)]) [input]</span>"
 					M.show_message(message, m_type)
+					M.create_chat_message(src, "* [formatted["subtext"]] *", FALSE, list("emote"), !!(m_type = AUDIBLE_MESSAGE))
```
Leshana, line 197: dis bit

Leshana (changes requested):

Very-Soft:
\o/

code/_helpers/unsorted.dm
```diff
@@ -1691,3 +1691,19 @@ GLOBAL_REAL_VAR(list/stack_trace_storage)
 		return "CLIENT: [D]"
 	else
 		return "Unknown data type: [D]"
+
+/*
+	is_holder_of(): Returns 1 if A is a holder of B, meaning, A is B.loc or B.loc.loc or B.loc.loc.loc etc.
+	This is essentially the same as calling (locate(B) in A), but a little clearer as to what you're doing, and locate() has been known to bug out or be extremely slow in the past.
+*/
+/proc/is_holder_of(const/atom/movable/A, const/atom/movable/B)
```
ShadowLarkens, line 1699: why not use A.contains(B)? it supports atoms too

code/datums/chat_message.dm
```diff
@@ -0,0 +1,299 @@
...
+	message.plane = PLANE_LIGHTING_ABOVE
+	message.appearance_flags = APPEARANCE_UI_IGNORE_ALPHA | KEEP_APART
+	message.alpha = 0
+	message.pixel_y = owner.bound_height * 0.95
+	message.maptext_width = msgwidth
+	message.maptext_height = mheight
+	message.maptext_x = (msgwidth - owner.bound_width) * -0.5
+	message.maptext = complete_text
+
+	if (is_holder_of(owner, target)) // Special case, holding an atom speaking (pAI, recorder...)
```
ShadowLarkens, line 163:
```suggestion
	if (owner.contains(target)) // Special case, holding an atom speaking (pAI, recorder...)
```

code/datums/chat_message.dm
```diff
@@ -0,0 +1,299 @@
...
+  * Arguments:
+  * * text - The text content of the overlay
+  * * target - The target atom to display the overlay at
+  * * owner - The mob that owns this overlay, only this mob will be able to view it
+  * * extra_classes - Extra classes to apply to the span that holds the text
+  * * lifespan - The lifespan of the message in deciseconds
+  */
+/datum/chatmessage/New(text, atom/target, mob/owner, list/extra_classes = null, lifespan = CHAT_MESSAGE_LIFESPAN)
+	. = ..()
+	if (!istype(target))
```
ShadowLarkens, line 51: personal style opinion: go through every single `if` statement added in this PR and make it `if(`, and not `if (`

code/datums/chat_message.dm
```diff
@@ -0,0 +1,299 @@
...
+  * * extra_classes - Extra classes to apply to the span that holds the text
+  * * lifespan - The lifespan of the message in deciseconds
+  */
+/datum/chatmessage/proc/generate_image(text, atom/target, mob/owner, list/extra_classes, lifespan)
+	set waitfor = FALSE
+	// Register client who owns this message
+	owned_by = owner.client
+	RegisterSignal(owned_by, COMSIG_PARENT_QDELETING, .proc/qdel_self)
+
+	var/erp_king = owned_by.is_preference_enabled(/datum/client_preference/runechat_long_messages)
```
ShadowLarkens, line 85: I'd prefer this be called extended_preference in the code and just keep the silly terminology in the preferences menu

code/datums/chat_message.dm
```diff
@@ -0,0 +1,299 @@
...
+		text = copytext_char(text, 1, maxlen + 1) + "..." // BYOND index moment
+
+	// Calculate target color if not already present
+	if (!target.chat_color || target.chat_color_name != target.name)
+		target.chat_color = colorize_string(target.name)
+		target.chat_color_darkened = colorize_string(target.name, 0.85, 0.85)
+		target.chat_color_name = target.name
+
+	// Get rid of any URL schemes that might cause BYOND to automatically wrap something in an anchor tag
+	var/static/regex/url_scheme = new(@"[A-Za-z][A-Za-z0-9+-\.]*:\/\/", "g")
```
ShadowLarkens, line 100: thoughts on having this as a static variable in this proc vs having it as a static global, or static on /datum/chatmessage?

code/datums/chat_message.dm
```diff
@@ -0,0 +1,299 @@
...
+		message.plane = PLANE_PLAYER_HUD_ABOVE
+
+	// View the message
+	owned_by.seen_messages.Add(src)
+	owned_by.images += message
+	animate(message, alpha = 255, time = CHAT_MESSAGE_SPAWN_TIME)
+
+	// Prepare for destruction
+	scheduled_destruction = world.time + (lifespan - CHAT_MESSAGE_EOL_FADE)
+	spawn(lifespan - CHAT_MESSAGE_EOL_FADE)
```
ShadowLarkens, line 173: This spawn and the one on line 149 cause a race condition that can allow end_of_life() to be called more than once

code/datums/chat_message.dm
```diff
@@ -0,0 +1,299 @@
...
+	if (is_holder_of(owner, target)) // Special case, holding an atom speaking (pAI, recorder...)
+		message.plane = PLANE_PLAYER_HUD_ABOVE
+
+	// View the message
+	owned_by.seen_messages.Add(src)
+	owned_by.images += message
+	animate(message, alpha = 255, time = CHAT_MESSAGE_SPAWN_TIME)
+
+	// Prepare for destruction
+	scheduled_destruction = world.time + (lifespan - CHAT_MESSAGE_EOL_FADE)
```
ShadowLarkens, line 172: I don't think this variable needs to exist anymore

code/game/atoms.dm
```diff
@@ -513,6 +522,8 @@
 		var/mob/M = mob
 		if(M.see_invisible >= invisibility && MOB_CAN_SEE_PLANE(M, plane))
 			M.show_message(message, VISIBLE_MESSAGE, blind_message, AUDIBLE_MESSAGE)
+			if(runemessage != -1)
+				M.create_chat_message(src, "* [runemessage || message] *", FALSE, list("emote"), audible = FALSE)
```
ShadowLarkens, line 526: I don't think this works correctly, there doesn't appear to be any condition where `message` is possible to print. I assume this should just work if you don't specify runemessage and instead right now it'll print just an eye.

code/modules/mob/living/say.dm
```diff
@@ -40,18 +40,18 @@ var/list/department_radio_keys = list(
 
 	  //kinda localization -- rastaf0
 	  //same keys as above, but on russian keyboard layout. This file uses cp1251 as encoding.
-	  ":�" = "right ear",	".�" = "right ear",
```
ShadowLarkens, line 43: you nuked the file encoding, need to go back in vscode, select encoding in the bottom right

and set it to

code/modules/mob/living/silicon/ai/ai.dm
```diff
@@ -1,1002 +1,1003 @@
-#define AI_CHECK_WIRELESS 1
```
ShadowLarkens, line 1: this file also looks weird, some kind of whitespace or encoding changes

code/modules/mob/living/silicon/say.dm
```diff
@@ -1,126 +1,127 @@
-/mob/living/silicon/robot/handle_message_mode(message_mode, message, verb, speaking, used_radios)
```
ShadowLarkens, line 1: Some kind of spurious whitespace or encoding again

code/modules/mob/living/say.dm
ShadowLarkens: The other ones below don't super matter except for visibility, this one **must** be fixed or it breaks shit

interface/skin.dmf
```diff
@@ -1282,7 +1282,7 @@ window "mapwindow"
 		saved-params = "icon-size"
 		on-show = ".winset\"mainwindow.mainvsplit.left=mapwindow\""
 		on-hide = ".winset\"mainwindow.mainvsplit.left=\""
-		style=".center { text-align: center; } .maptext { font-family: 'Small Fonts'; font-size: 7px; -dm-text-outline: 1px black; color: white; line-height: 1.1; } .small { font-size: 6px; } .big { font-size: 8px; } .reallybig { font-size: 8px; } .extremelybig { font-size: 8px; } .clown { color: #FF69Bf;} .tajaran {color: #803B56;} .skrell {color: #00CED1;} .solcom {color: #22228B;} .com_srus {color: #7c4848;} .zombie	{color: #ff0000;} .soghun {color: #228B22;} .vox {color: #AA00AA;} .diona {color: #804000; font-weight: bold;} .trinary {color: #727272;} .kidan {color: #664205;} .slime {color: #0077AA;} .drask {color: #a3d4eb;} .vulpkanin {color: #B97A57;} .abductor {color: #800080; font-style: italic;} .his_grace { color: #15D512; } .hypnophrase { color: #0d0d0d; font-weight: bold; } .yell { font-weight: bold; }"
```
ShadowLarkens, line 1285: you nuked some of the classes in here
they're not currently used but we should probably combine them properly

ShadowLarkens (changes requested):
I'm overall a little skeptical of this timerless port because I see some issues that spawn() is causing. It may be best to switch back to timers.

code/modules/mob/living/say.dm
Leshana, line 43: Did this file get re-encoded?

code/datums/chat_message.dm
Leshana: static variable in this proc is fine I imagine, that keeps it close to area of use but also doesn't require compiling the regex every time.

code/datums/chat_message.dm
Leshana: This one is absolutely required.   The one on line 149 is to cause an early-EOL if the stack of messages gets too tall.
end_of_life() does check if its deleted before running the animation again.

code/modules/mob/living/silicon/ai/ai.dm
Leshana: I think this one is to fix the file to bring it into correct whitespace.   However, asking that be done in a separate PR would be a reasonable argument.

code/datums/chat_message.dm
Leshana: This variable exists so the acclerated-despawn effect caused by later messages has a baseline for how much time was going to be left anyway.

code/datums/chat_message.dm
ShadowLarkens: it checks if it's deleted but doesn't delete it until after another spawn
hence race condition
also this means that qdel can be called twice which is very bad iirc

code/_helpers/unsorted.dm
Arokha: Didn't know it existed! I'll switch to that.

code/datums/chat_message.dm
Arokha: I'm gonna do changes and rebase so not gonna commit this here.

code/datums/chat_message.dm
Arokha: I also prefer if() though I guess vg has a lot of if () so I kept it matching.

code/datums/chat_message.dm
Arokha: As far as I'm aware, static variables in procs are the same speed as static globals and such. Or do you just want it for use elsewhere?

code/datums/chat_message.dm
Arokha: Yes, it's needed for that feature.

code/game/atoms.dm
Arokha: Legacy, I'll remove.

code/datums/chat_message.dm
Arokha: It's not very bad, it just prints a warning, but I'll add logic.

code/modules/mob/living/say.dm
Arokha: The eye is a UTF character, though.

interface/skin.dmf
Arokha: If they aren't used: why?

code/modules/mob/living/say.dm
Arokha: Eeyup. Not sure how to keep the eye character plus have russian encoding. Byond supports UTF now, so maybe it should just be moved to UTF and have those characters moved to UTF.

code/datums/chat_message.dm
Arokha: Truuuue

code/modules/mob/living/say.dm
Arokha: Nevermind I read this wrong, the eye is in another file. STILL. I'm gonna convert that to UTF because EH.

Arokha: Nevermind I read this wrong, the eye is in another file. STILL. I'm gonna convert that to UTF because EH.

code/modules/mob/living/silicon/ai/ai.dm
Arokha: Gonna commit with the fixed endings.

Hatterhat:
i respect that

interface/skin.dmf
ShadowLarkens: yeah i guess

code/datums/chat_message.dm
ShadowLarkens: nah this is fine, I wasn't sure which is why I asked

code/controllers/subsystems/runechat.dm
```diff
@@ -0,0 +1,238 @@
...
+			cm.prev.next = cm
+		if (i)
+			second_queue.Cut(1, i + 1)
+		cm = null
+
+/datum/controller/subsystem/runechat/Recover()
+	bucket_list |= SSrunechat.bucket_list
+	second_queue |= SSrunechat.second_queue
+
+/datum/controller/subsystem/runechat/proc/reset_buckets()
```
Leshana, line 128: Kindly call this from `/world/proc/on_tickrate_change()`

Leshana (changes requested):

Leshana (approved):

PR #10495, opened 2021-05-29
Ghosts can only go through the gateway if it's turned on

Very-Soft:


code/_onclick/observer.dm
```diff
@@ -66,12 +66,21 @@
...
+// VOREStation Edit Begin
+
+/obj/machinery/gateway/centerstation/attack_ghost(mob/user as mob) 
+	if(awaygate) 
+		if(user.client.holder) 
+			user.loc = awaygate.loc
+		else if(active)
+			user.loc = awaygate.loc
+		else
+			return
```
Novacat, line 78: This should probably have a to_chat instead of a return, similar to the line below this. Probably "[src] is not powered up."

Novacat (changes requested):

code/_onclick/observer.dm
MarinaGryphon: Ghosts interacting with things they're not able to typically doesn't give a message, to my knowledge. Giving a message here would be a bit odd.

Novacat: ...Good point.

Novacat (approved):

PR #10532, opened 2021-06-02
Fixes #10482

### LINKED ISSUE #10482: protens don't drop holsters when blobing if it's in your pocket like they do when attached to uniform
if you put a holster in your pocket, then put a gun in it. When you blob the holster will stay in your  pocket

it to get thrown to the floor like it does when attached to your uniform

#### What actually happened
can blob with gun

#### Steps to reproduce
- (Step 1)be protean
- (Step 2) get gun holster in your bocket
- (step 3) use holster to put gun in your pocket
- (step 4) blob

#### Code Revision
 

Arokha:
Holsters are not a gun compressor


Arokha:
Only upwards, it seems! Let's just start them at ITEMSIZE_NORMAL so they are at least that big and won't fit in pockets with tiny guns, which is not the intent of holsters. If the gun fits in your pocket just put the gun in your pocket.

PR #10535, opened 2021-06-02
Chessboard Holodeck TG-Style.

SpaceVampire:
This updates the Chessboard Holodeck to use TG sprites.
Because they look a lot better, in my opinion.
After hearing other people also voice issues with em, I made this PR.
**Old Chess**
Chesss
**New Chess**
NewChess
Black and White just makes more sense then Purple and Blue.
Plus I just feel the board is easier to read with the new tiles.


Arokha:
They do seem a lot easier to see

Very-Soft:
Yeah I like this way more

PR #10551, opened 2021-06-03
Exploration Gear Expansion DLC

klaasjared:
Adds long-awaited exploration department-based bags/backpacks as well as revamps the belts for both Explorer and Pathfinder. I also fixed the radio backpack on-mob sprite and did a little touching up like adding a dedicated handset sprite as opposed to the old signaler.

Below are the sprites that were added/changed.

pack_showcase

belt_showcase

radio_showcase


lm40:
handset sprite looks amazing :disbelief:

Ketrai:
image
image
Pointers for if you want to do in-depth shading.

klaasjared:
Do not merge this yet, gonna modify the sprites

Novacat (changes requested):
I would like to keep the old sprites (at least for the explorer backpack/satchel and the explorer belt) around because I probably will repurpose them for other things.

Novacat (changes requested):
Still needs the onmob sprites restored. Also, rather than using (old), I would recommend using _old instead, to make it easier for varediting.

Arokha:
Dang, these're pretty

Arokha:
Missed explorer_belt (old) and explorer_belt (old) in last commit, should do what Ascian said and do 'em to _old and I'll dismiss his change request

klaasjared:
closing really quick to reset the test - there's no Map Diffs so there shouldn't be an error

PR #10553, opened 2021-06-04
Make Talon's sensors part of its bridge instead of its captain's office

Little-119:
So that the sensors pull power from the bridge APC instead of the captain's quarters. Having them draw power from the captain's bedroom is a little weird, isn't it? Certainly seemed weird to me when I first saw how much power the captain's quarters were drawing


Arokha:
This is how every other ship is, so, makes sense

VictorZisthus:
Good catch! I definitely never noticed it.

PR #10559, opened 2021-06-06
Fix fake space edge turfs

Arokha:
@Very-Soft pointed out you could see walls through them. Was due to an optimization in precomputed appearance use. On mapper-placed ones I'll have to break that optimization (I mean I could set up another list but I dun wanna, it's ok).

Adds a few spacey things to shipself.

### DIFF
--- a/code/game/turfs/space/space.dm
+++ b/code/game/turfs/space/space.dm
@@ -136,7 +136,7 @@
 /turf/space/Entered(var/atom/movable/A)
 	. = ..()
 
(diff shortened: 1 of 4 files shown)


VictorZisthus:
"// !density so 'fake' space turfs don't fling ghosts everywhere"

Aww, I thought it was amusing getting flung into windoors as a ghost, having them open up for me and people wondering why that door just opened.

JK lol, thanks for the fix!

PR #10560, opened 2021-06-06
Update jukebox.json - Add 1ea secret song.

ResidentCody:
Adds the song "I Wupped Batman's Ass" to the secret songlist on the jukebox on behalf of Drate.

Song is quite niche and not exactly suitable for the main list, Should be flagged to be only visible if the jukebox is hacked.


config/jukebox.json
```diff
@@ -3499,4 +3499,14 @@
 "jukebox": true,
 "genre": "Rock"
 }
```
Arokha, line 3501: Missing a comma

PR #10561, opened 2021-06-06
Fixes the character creator by: Revert "Fixes #10321"

### LINKED ISSUE #10321: 'protean' appears as a prosthesis manufacturer
'protean' appears as a prosthesis manufacturer. As in, it appears in the list when you're prompted to select what manufacturer you want your prosthetic to be made by when changing your character's limbs in character setup.

#### Code Revision
- 34657a5820a3541ad7a9b71b3da5a97820d8eb9c 
 

image

Detective-Google:
Reverts VOREStation/VOREStation#10537

Breaks creating characters with prosthetic limbs because almost every limb has unavailable_to_build set to 1. Find a proper solution and PR that instead.

For example: 
image


Arokha:
https://github.com/VOREStation/VOREStation/pull/10562

Let's use this one instead since it fixes the original problem as well.

PR #10575, opened 2021-06-06
moves mat amour to VR file and adds bunch of chompers traits because github won't let me open multiple PRs for some reason

TheDavestDave:
I regret not saving stuff in a word doc, now I have to do this again.

undoes a fix here, it's been uploaded to polaris but it's in a VR file here until we get that fix.

a lot of intresting traits, some got dicthed becuase I couldn't make them work, others felt cheesy


code/modules/mob/living/carbon/human/species/station/traits_vr/positive.dm
```diff
@@ -156,3 +156,90 @@
...
+	cost = 2
+	var_changes = list("pain_mod" = 0.8)
+
+/datum/trait/positive/linguist
+	name = "Master Linguist"
+	desc = "You are a master of languages! For whatever reason you might have, you are able to learn many more languages than others."
+	cost = 0
+	var_changes = list("num_alternate_languages" = 12)
+
+/datum/trait/negative/hollow
```
Very-Soft, line 179: Shouldn't this be /positive/hollow?

Very-Soft:
In my opinion:

-Slippery dirt should be -1 instead of -2: 2 is a lot for something that situational.

-Linguist should definitely not be free, and DEFINITELY not allow you go take 12 extra languages. Like, cost 1 and maybe 2 or 3 extra languages.: You already start with the ability to take up to 3 languages (assuming your species doesn't come with extra languages), and, not only do we not let people take translators for free, we don't let them take sneaky translator headsets at all. I'm fine with this giving you significantly more languages, but I think, it shouldn't be HALF the total available languages extra.

-Table passer should probably cost 3: That can be a pretty powerful ability, especially when coupled with haste

I really like the nutrition grow/shrink ones, that's a great idea

Arokha:
`because github won't let me open multiple PRs for some reason`
image

PRs are requesting a merge of all the commits on a branch into another branch. You can make one PR per branch. You have to do separate projects on separate branches and make separate PRs for those.

code/game/objects/items/weapons/material/material_armor_VR.dm
```diff
@@ -0,0 +1,4 @@
+/obj/item/weapon/material/armor_plating/insert
+	unbreakable = FALSE
+	name = "plate insert"
+	desc = "used to craft armor plates for a plate carrier. Trim with a welder for light armor or add a second for heavy armor"
```
Arokha, line 4: This can't be here, it was already added in another PR from upstream presumably by you, so no need for this

TheDavestDave:
I'll start again with differnt branchs

PR #10578, opened 2021-06-06
Put the music speakers in crates!

Very-Soft:
You can order them, but they don't appear on the shuttle! This should fix that.


Arokha:
:O oops

PR #10602, opened 2021-06-09
Round-point Midpoint Upgrade

klaasjared:
Expands the midpoint map to be significantly larger than its current design by taking pages from existing real-life places such as the Seattle Space Needle. Nothing significant was added, this is simply an expansion that allows for a 360 view of Virgo 3b's atmosphere.

A picture of the new map can be found below:

new_mid_showcase


Zeldazackman:
Weird question, but could there be a jukebox added to the area since they work again?

Arokha:
Jukebox would be pretty cool, yeah

PR #10603, opened 2021-06-09
Adds Absorbed Examine settings to Vore Panel

Heroman3003:
Adds two new settings:

Examine Text (with absorbed victims) - examine texts that will show if you have people absorbed in your belly and have Show Absorbed Examine on.
Show Absorbed Examine - what it says on the tin, a toggle for whether the absorbed examine texts are supposed to show for that belly at all or not.

Mostly adding this because it's kinda really lame how once person is absorbed, they straight up disappear, not showing up in examines, not making a bulge or anything. This kind of thing allows to keep person mechanically absorbed, while still 'showing off' in same way you would a regular prey bulge.


code/modules/vore/eating/belly_obj_vr.dm
```diff
@@ -379,11 +386,32 @@
...
+
+		formatted_message = replacetext(raw_message, "%belly", lowertext(name))
+		formatted_message = replacetext(formatted_message, "%pred", owner)
+		formatted_message = replacetext(formatted_message, "%prey", english_list(absorbed_victims))
+		formatted_message = replacetext(formatted_message, "%countprey", absorbed_count)
+		if(display_absorbed_examine && absorbed_count)
+			return("<span class='warning'>[formatted_message]</span>")
+		else
+			return ""
+
```
Arokha, line 409:
```suggestion
/obj/belly/proc/get_examine_msg_absorbed()
	if(!display_absorbed_examine || !examine_messages_absorbed.len || !contents.len)
		return ""
		
	var/raw_message = pick(examine_messages_absorbed)

	var/absorbed_count = 0
	var/list/absorbed_victims = list()
	for(var/mob/living/L in contents)
		if(L.absorbed)
			absorbed_victims += L
			absorbed_count++
	if(!absorbed_count)
		return ""

	var/formatted_message
	formatted_message = replacetext(raw_message, "%belly", lowertext(name))
	formatted_message = replacetext(formatted_message, "%pred", owner)
	formatted_message = replacetext(formatted_message, "%prey", english_list(absorbed_victims))
	formatted_message = replacetext(formatted_message, "%countprey", absorbed_count)
	return("<span class='warning'>[formatted_message]</span>")

```

Not a huge fan of 'entire proc is inside a condition'. You should check display_absorbed_examine and absorbed_count earlier (in different places) to save all the work of doing those replacetext's and the loop and stuff! (You should test this change I'm suggesting)

Heroman3003: The reason I did it like that is mostly cuz thats how regular get_examine_msg does that, so I just wanted to go with consistent logic. I'll take a look at that later once I'm back home.

Arokha: I figured it was copypasted. I'm not askin' you to change the other one (though if it's so similar, I'd maybe say you COULD fix the other one, for bonus points), just the one you added.

Heroman3003: Fixed for both messages. I think.

PR #10604, opened 2021-06-09
Adds a different version of bloodsucker trait

Heroman3003:
Renames existing Bloodsucker into Bloodsucker (Obligate)

Adds a different version just called Bloodsucker which still gives suck blood verb, but does not make other sources of nutrition useless.

Because while it's cool for people who want to play full vampire, for those who want a more mixed diet, its way too restrictive. So, options!


Very-Soft:
yes pls

Arokha:
Sure why not

PR #10608, opened 2021-06-10
Port emitter sprites

Arokha:
From /vg/, though I'm not entirely sure of their ORIGINAL original origins!

It has little wheelies that deploy when you un-anchor it c:

https://user-images.githubusercontent.com/15028025/121567620-cb95f780-c9ec-11eb-9740-196ac6629a6b.mp4


Novacat:
Rather than poofing in thin air, I wonder if maybe they should be replaced with a broken machine object.

dwinters99:
Extremely personal preference but I must say I strongly dislike this icon.

Arokha:
> Rather than poofing in thin air, I wonder if maybe they should be replaced with a broken machine object.

What would you do with it? Also WYSI

Novacat:
yes I was dead tired when I wrote that and did not realize I was posting in the wrong PR

Very-Soft:
I love how this sprite looks, we might want to adjust the beam sprite for it eventually though

PR #10612, opened 2021-06-11
Ports suit dispenser from /vg/

Arokha:
It dispenses ~~suits~~ anything, and especially voidsuits. I rewrote a lot of it, oops. I dunno what "porting" is anymore.

They can customize voidsuits automatically as they dispense, and equip magboots, helmet, air/cooler for you depending on your species/syntheticness, etc etc.

You can also limit items by ckey or character name, or access or like a billion other things because EH. For voidsuits, it will (optionally, usually) refit them to your species and put an air tank/suit cooler on them for you, and magboots if you're a cool person.

PDTAHxmJRH


Heroman3003:
This also removes tether better in same PR? Not that it's not reasonable move or anything, but I'd think it'd be a separate PR or at least have mention in description

Arokha:
> This also removes tether better in same PR? Not that it's not reasonable move or anything, but I'd think it'd be a separate PR or at least have mention in description

Done

PR #10613, opened 2021-06-11
Fixes some telecomms things

Heroman3003:
Makes cycling a bit more reasonable again

Adjusts the entrances so that you can access both teleporter and messaging server without going into telecomm area

Adds dust shielding to walkway

Adds more turrets and symmetry

Removed extra unnecesary turret controller

Makes turrets not shoot all non-synthetics by default

Also, not telecomms, but removes id restorer in the hallway cuz deprecated device and puts another guestpass there cuz why not.


Arokha:
> Makes turrets not shoot all non-synthetics by default

Aren't they supposed to? This is traditional in the tcommsat that the turrets try to murder anyone unless you turn them off.

pontifexminimus2:
I think the people authorized to be there shouldn't be shot at by default.

Seems silly for me, since we do have ID checking for turrets. It's used in the AI upload.

Heroman3003:
> 
> 
> I think the people authorized to be there shouldn't be shot at by default.
> 
> Seems silly for me, since we do have ID checking for turrets. It's used in the AI upload.

Actually, ID checks on turrets currently don't QUITE work the way they are intended. It could be fixed in two different ways, but better one would be a different PR to make entirely.

Basically... When they have access check enabled, the turrets check if people in viccinity have access to TURRET, not their control panel. And all turrets simply have basic security and basic heads access set to them. Combined. But turret calculations are also a mess.

Novacat (changes requested):
Conflict!

Heroman3003:
Conflict resolved

Novacat (approved):

PR #10617, opened 2021-06-11
Adds PROPER tajaran ears as char setup option in ear category

Heroman3003:
The current option for tajaran ears is some weird STUB that looks same on both versions and doesnt' even have the back sprite and the real ears are still only available as hairstyle... Well, this literally copies that 'hairstyle' into ears option so you can pick actually good-looking hair along with authentic tajaran ears.


KasparoVy:
Heeelll yess. Cheers

PR #10631, opened 2021-06-13
Material Container Component & Autolathe QoL

ShadowLarkens:
This ports /tg/'s Material Container component, a generic component for any machine that wants to hold materials and harvest them from items. It is currently only implemented on the Autolathe, but should eventually be ported to the Exosuit Fabricator and other things like that.

The Autolathe has also been touched up with this PR, both code and QoL-wise.
 - You can now select x5 and x10 bulk printing on any item that's not a sheet (the old input window remains) 
 - When there are no materials, it doesn't just collapse the section and look awkward, it shows a proper empty state. 
 - Sheets now show in the "All" category.


ShadowLarkens:
Note: On /tg/, stacks are very awkward with this new system, because if your full stack goes over the cap, the machine just refuses to insert it at all, forcing you to manually split the stack. I've fixed this for this port.

code/modules/materials/materials/_materials.dm
```diff
@@ -49,6 +49,24 @@ var/list/name_to_material
...
+ *
+ * Used when recycling items, specifically to turn alloys back into their component mats.
+ *
+ * Exists because I'd need to add a way to un-alloy alloys or otherwise deal
+ * with people converting the entire stations material supply into alloys.
+ *
+ * Arguments:
+ * - flags: A set of flags determining how exactly the materials are broken down. (unused)
+ */
+/obj/proc/get_material_composition(breakdown_flags=NONE)
```
Arokha, line 64: Doc isn't accurate to arg name

PR #10633, opened 2021-06-13
Fixes tons of empty lines showing up in borgs' examines

Heroman3003:


code/modules/mob/living/silicon/robot/examine_vr.dm
```diff
@@ -3,7 +3,9 @@
 	var/list/message_list = list()
 	for(var/belly in vore_organs)
 		var/obj/belly/B = belly
-		message_list += B.get_examine_msg()
-		message_list += B.get_examine_msg_absorbed()
+		if(B.get_examine_msg() != "")
+			message_list += B.get_examine_msg()
+		if(B.get_examine_msg_absorbed() != "")
+			message_list += B.get_examine_msg_absorbed()
```
Arokha, line 9:
```suggestion
		var/bellymessage = B.get_examine_msg()
		if(bellymessage) message_list += bellymessage
		bellymessage = B.get_examine_msg_absorbed()
		if(bellymessage) message_list += bellymessage
```
Even better if you declare bellymessage outside the for() so it doesn't redeclare it each loop. Also "" is 'falsey' in byond so you can just do that, no need to compare to empty string.

Arokha (dismissed):

code/modules/mob/living/silicon/robot/examine_vr.dm
Heroman3003: I wasn't sure about strings' falseyness, so I didn't dare take rissks. In other place its used, its checking if(string == "") return kind of way, so I decided to go along.

PR #10652, opened 2021-06-15
Porting Over De-Husking Surgery

Enzo-Leon:
What it says on the paper. Adds a De-Husking surgery & an extra de-toxify repair "surgery" similar to our external repair surgery currently.

Dehusking surgery steps
(ENSURE THE PATIENT IS HEALED OF BURNS BEFORE STARTING)
- Aim the biogenerator at their chest and click to start
- Hemostat
- Finish with biogenerator

De-toxify works the same with aiming at the chest and aiming the biogenerator at them to pull toxins out of their body.

To help encourage the use of the de-husking process I've also increased the healing amount per use of the external repair surgery by 100%. This increased the healing from 25 per use to 50.

A note was also added in the OR rooms for the new surgery item to help players out.


Wickedtemp:
So just to confirm, burn damage has to be 0, yeah?

I think this is long overdue. There's no real reason that I can see as to why we shouldn't merge this. The person's gonna get resleeved or will be allowed to respawn anyway, might as well give us the capabilities to fix it with surgery and get that sweet medical gameplay.

Enzo-Leon:
> So just to confirm, burn damage has to be 0, yeah?
> 
> I think this is long overdue. There's no real reason that I can see as to why we shouldn't merge this. The person's gonna get resleeved or will be allowed to respawn anyway, might as well give us the capabilities to fix it with surgery and get that sweet medical gameplay.

Burn damage does NOT have to be 0. I tested it with the body still having some left over, but it's best to ensure they're at least <100 for damage before going through the surgery.

lm40:
don't really see the need to double the external  repair surgery then

Wickedtemp:
Probably cause husking means they have around 400 burn damage, usually higher than that even, but taking 400 as the example, you'd need to do the 'heal burn damage' step about 13 times, minimum. This would just bump that down to 6-7. When it comes to people who were thoroughly cooked, it isn't really that uncommon to see burn damage above 700, which would take 25 surgical steps to heal if the numbers aren't adjusted.

At first, I really didn't see the point in that, either, but...now I kinda do.

lm40:
i don't really  see that as a problem, if you have 700 damage than yes  fixing it entirely with surgery will be slow. there are ways to mitigate that, but it's 700 damage lmao.

Wickedtemp:
Yeah, but I'd rather do that with 13 surgical steps instead of 25. Still a substantial amount, just not to the point where I want to pull my hair out.

Novacat (changes requested):
This edits a polaris DMI, which means it will have to be sent upstream, or changed to a _vr dmi.

Also, not really too keen on making external repair surgery even stronger.

Wickedtemp:
> This edits a polaris DMI, which means it will have to be sent upstream, or changed to a _vr dmi.
> 
> Also, not really too keen on making external repair surgery even stronger.

It really isn't that strong as it is. It's basically about the same speed as having reagents in your system. For burns, pretty sure it's slower than dermaline. It's almost exclusively used for fixing up corpses anyway.

Novacat (changes requested):
Conflicts, and, the new biogenerator should be put in surgery kits rather than mapped in separate, otherwise this should be fine.

Heroman3003:
If it gets put in the surgery kits, the conflicts should resolve themselves on their own, probably.

Novacat (approved):

PR #10654, opened 2021-06-15
Decreases notification timing for transcore, allows delaying

Heroman3003:
Decreases initial notification timer to 6 minutes, down from 15

Decreases cooldown on repeat notification to 5 minutes, down from 10

Changes repeat notification message to be same as regular

Added a verb for ghosts that 'cancels' transcore message without removing the record or backup. Good if you want your pred to be one resleeving you or stick around for longer before doctors get to you. Notify Transcore can be used manually to send it later.

Mostly all of it is a proposal, since there's not much reason for the backup timer to be that high now that defib timer is an hour.

### DIFF
--- a/code/controllers/subsystems/transcore_vr.dm
+++ b/code/controllers/subsystems/transcore_vr.dm
@@ -15,7 +15,7 @@ SUBSYSTEM_DEF(transcore)
 	init_order = INIT_ORDER_TRANSCORE
 
 	// THINGS
-	var/overdue_time = 5 MINUTES
+	var/overdue_time = 6 MINUTES			// Has to be a multiple of wait var, or else will just round up anyway.
 
 	var/current_step = SSTRANSCORE_IMPLANTS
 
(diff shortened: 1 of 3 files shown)


Arokha:
Keep in mind that this subsystem only fires every 3 minutes. Non-multiples of 3 will not work like you might expect. So this will be a minimum of 6 minutes, despite you setting it to 5.

code/controllers/subsystems/transcore_vr.dm
```diff
@@ -113,8 +113,11 @@ SUBSYSTEM_DEF(transcore)
 			curr_MR.dead_state = MR_NORMAL
 		else
 			if(curr_MR.dead_state != MR_DEAD) //First time switching to dead
-				db.notify(curr_MR.mindname)
-				curr_MR.last_notification = world.time
+				if(curr_MR.do_notify)
+					db.notify(curr_MR.mindname)
+					curr_MR.last_notification = world.time
+				else
+					curr_MR.last_notification = world.time - 5 MINUTES
```
Arokha, line 120: Janky

JoanRisu:
This is fine. Green light to pass.

PR #10657, opened 2021-06-15
adds the phaser handbow

TheDavestDave:
adds a resprited phaser holdout. Instead of a beam it does a bolt that does the exact same damage. it also has no non-taser replacement becuase it needed it's own spirte and I already modded it twice for lethal and low power.


code/datums/supplypacks/munitions_vr.dm
```diff
@@ -3,8 +3,19 @@
...
+			/obj/item/weapon/gun/energy/locked/frontier/handbow=2
 			)
 	cost = 35
 	containertype = /obj/structure/closet/crate/secure
 	containername = "frontier phaser crate"
-	access = access_explorer
\ No newline at end of file
+	access = access_explorer
+	
+/datum/supply_pack/munitions/expeditionguns
```
Novacat, line 13: This will end up overwriting the previous entry. You have to change the name.

code/datums/supplypacks/munitions_vr.dm
```diff
@@ -3,8 +3,19 @@
 	contains = list(
 			/obj/item/weapon/gun/energy/locked/frontier = 2,
 			/obj/item/weapon/gun/energy/locked/frontier/holdout = 2,
+			/obj/item/weapon/gun/energy/locked/frontier/handbow=2
```
Novacat, line 6: If this is going to have its own crate, it should probably not be included in this crate.

code/datums/supplypacks/munitions_vr.dm
```diff
@@ -3,8 +3,19 @@
...
-	access = access_explorer
\ No newline at end of file
+	access = access_explorer
+	
+/datum/supply_pack/munitions/expeditionguns
+	name = "Frontier bows (station-locked) crate"
+	contains = list(
+			/obj/item/weapon/gun/energy/locked/frontier/handbow=2
+			)
+	cost = 35
```
Novacat, line 18: If all it has are handbows it could probably be cheaper. 20 points?

Novacat (dismissed):

code/datums/supplypacks/munitions_vr.dm
TheDavestDave: aparantly discard changes doesn't mean the same to be as it does to github

Arokha:
There's a stray pixel on the pump animation

Novacat:
This reminds me that I need to slightly nerf low power for phasers but not related to this PR.

PR #10661, opened 2021-06-17
Add a new contamination

lbnesquik:


code/_helpers/global_lists_vr.dm
```diff
@@ -435,6 +436,22 @@ var/global/list/contamination_flavors_musky = list("drenched",
...
+				"gunky",
+				"slimed",
+				"mucky",
+				"viscous",
+				"dank",
+				"glutinous",
+				"syrupy",
+				"slippery",
+				"gelatinous",
+				"thick")
```
Very-Soft, line 453: These are good, but if I might offer a point of critique, I think 'thick' might look a bit funny

'thick headband'
'thick Soanso's ID card', etc.

code/_helpers/global_lists_vr.dm
```diff
@@ -232,6 +232,7 @@ var/global/list/contamination_flavors = list(
 				"Dirty" = contamination_flavors_dirty,
 				"Musky" = contamination_flavors_musky,
 				"Smelly" = contamination_flavors_smelly,
+				"slimy" = contamination_flavors_slimy,
```
Heroman3003, line 235: probably should be Slimy instead of slimy to match all other?

code/_helpers/global_lists_vr.dm
lbnesquik: thicc

Leshana (approved):

PR #10662, opened 2021-06-17
Port current /tg/ lighting system

Arokha:
Funny enough I just wanted this for directional lighting, but it turns out in the past week there've been a few large changes to their lighting to improve performance and maptick speed, so those are here too. Let me know if I forgot to credit anyone.

## There Will Be Bugs
Things will look different, and you will find bugs. Something that touches this many files is going to leave some issues. It's been tested as well as I can test with a few helpers. When you find a bug, report it in a **polite and calm** manner. Whining about how you hate SS13 forever because you couldn't see the right color on your ID or because your dorm had no lights is not the appropriate method. Just report the issue you saw, like, using facts.

## Potential Advantages
- The actual visible part of lighting is no longer an atom that hides on every turf. It's an underlay on every turf, and this reduces client maptick time as there's one less atom in every turf to send to clients. (Kylerace w/ nod to zewaka)
- Corners now delete themselves if they are fully unaffected by light, reducing their memory usage for dark areas (could be maint, or a planet at midnight, or whatever). They also avoid performing some work if the light from them reaching a turf/corner is effectively 'dark'. (MrStonedOne)
- Rather than corners and turfs keeping lists of NW,NE,SW,SE corners/masters, they have these in separate vars, as this is more performant and less memory-consuming. The list was always going to be 4 entries long in both cases, so it was pointless to keep as a list rather than just as 4 vars. (MrStonedOne)
- An emissives system. Rather than doing that whole 'just put an overlay above lighting', which has the downside of being unable to be blocked by mobs, we have a system for 'emissives' (i.e. things that should show up despite the area being dark), but are also able to be blocked by mobs/objs using filters and masking. (actioninja, TemporalOroboros)
- Directional lighting, done by using images to mask away the darkness. Usually two images, one for a cone of light coming out of the light emitter('s parent atom) and one for the 'out in front' atom that holds a sprite. This is so that you can't push the light through walls or anything, and you can (in most cases) see it around corners and whatnot. (Azarak)

## Potential Disadvantages
- ~~The planetary lighting system is a little more clumsy than before, it does keep SSlighting busy for a few cycles as it updates lights, since SSplanets no longer directly tampers with light corners (not it's business). There's a fake /datum/light_source that the planet holds, that 'shines' on every planet turf basically. Open to suggestions on how to improve.~~ I rewrote it, now it uses no CPU.
- Light colors may be different than before, because Bay's lights had a strange bit of math when calculating lums for corners, where instead of applying the lum for each color as-is, it would take the square root and half it, then apply that instead. I'm unsure why, aside from 'it looks nice', but doing sqrt*0.5 on every color on every corner seemed expensive, so it was removed and just uses /tg/'s direct application of the lums. Of course, all the light colors are still POSSIBLE, but the color defines might need to be tweaked. ~~I did this for the two most 'popular' light colors, the tube and bulb.~~ Soft helped review and we came up with what I think are some good values.

## Tweaks I've made
- I thought it would be nice if the directional lighting cones could take hints as to how to offset the cone, so now it looks like the cone is coming out of your hardhat, or flashlight (aka it has a Y offset appropriately, see video below).
- Various Polaris things needed to be tweaked to have the emissives rather than using the above lighting plane, so I swapped those around.
- Created a fake lighting source for planets to use, that helps pin their corners in place so they don't evaporate (see aforementioned light-less corners self-destroying). In testing it seems to work fine.
- Adjusted light colors to try and replace pre-/tg/-lighting color scheme of the two main light sources (tubes and bulbs). Some additional work may need to be done to tweak light colors to whatever you prefer.
- ~~I'm using Bay's falloff math rather than /tg/'s falloff math for lighting. They appear to have similar performance characteristics and in the interest of having things look relatively similar, I did that. /tg/'s is in there, just commented out for now, unless a change is desired on that front.~~ I ended up using /tg/'s falloff math.

## Things you should know
- `light_system` var on atoms decides what light system they will use, where `STATIC_LIGHT` is the traditional system, where lights affect light corners in range, which in turn update an overlay (or, now, an underlay) applied to turfs. `MOVABLE_LIGHT` uses an alpha masking filter to remove darkness in an area, which is **far** less expensive, however only looks good for smallish lights. You can only see the light if you can see it's source, so imagine thinking a hallway is dark because you can't yet see the light. `MOVABLE_LIGHT_DIRECTIONAL` casts the light out in front. Both of these use a set of pre-made sprites that try to replicate the 'range' settings of the traditional system.
- For emissives (e.g. computer displays, APC lights, etc) that should be above darkness, you can use the new emissive_appearance() helper. Just add both a mutable_appearance() and an emissive_appearance() in most cases to the overlays (using add_overlay() of course... don't touch overlays directly) and you should end up with a nice glow over darkness that's also blockable by most objs and mobs.

https://user-images.githubusercontent.com/15028025/122092638-26a66080-cdd8-11eb-9406-0268223a23ee.mp4

image


Razgriz1032:
Just clone another codebase, or clone Citadel

Arokha:
> Just clone another codebase, or clone Citadel, or wait for SS14

or I can improve our codebase now! \o/

Leshana (changes requested):

KillianKirilenko:
> WcpA1xU27V.mp4

Funny that you add this when I took a shot at porting the Eris flashlights a couple weeks back. Unfortunately it threw so many errors I couldn't figure them out.

erislight

The main thing about it that I like is that if you have the flashlight in your active hand you can actually aim it around by clicking on turfs to highlight specific things. It snaps back to the 'resting' position 3 tiles ahead of your mob when you move. The spotlight sprite is also switched between icon states and has x/y scaling depending on how close or far it is from your mob, and rotates nicely as well.

Arokha:
> > WcpA1xU27V.mp4
> 
> Funny that you add this when I took a shot at porting the Eris flashlights a couple weeks back. Unfortunately it threw so many errors I couldn't figure them out.
> 
> erislight
> 
> The main thing about it that I like is that if you have the flashlight in your active hand you can actually aim it around by clicking on turfs to highlight specific things. It snaps back to the 'resting' position 3 tiles ahead of your mob when you move. The spotlight sprite is also switched between icon states and has x/y scaling depending on how close or far it is from your mob, and rotates nicely as well.

That'd be fun to add! Port that after this is in if you want? Or if you need help, happy to help.

Leshana (approved):

Novacat (approved):
Once the runtimes are fixed I will merge.

KasparoVy:
Oh holy shit nice! That was some heavy lifting, daaamn
Great work

PR #10663, opened 2021-06-17
Change span macros

Arokha:
Never one to be outdone, byond has decided that this is a more betterer way to write these macros.
Normal text strings written to world are 3 ops.
This concat style macro is the same 3 ops.
The 'text replacement', despite the thing it's doing being a macro, compiles to something that takes 4 ops.


Arokha:
EH nevermind

PR #10666, opened 2021-06-17
Add Bridge Bunny alt-title

Arokha:
Per Joan and staff meeting

### DIFF
--- a/code/game/jobs/job/captain_vr.dm
+++ b/code/game/jobs/job/captain_vr.dm
@@ -48,7 +48,7 @@
 	disallow_jobhop = TRUE
 	pto_type = PTO_CIVILIAN
 	alt_titles = list("Command Liaison" = /datum/alt_title/command_liaison, "Command Assistant" = /datum/alt_title/command_assistant, "Command Intern" = /datum/alt_title/command_intern,
-						"Bridge Secretary" = /datum/alt_title/bridge_secretary, "Bridge Assistant" = /datum/alt_title/bridge_assistant)
+						"Bridge Secretary" = /datum/alt_title/bridge_secretary, "Bridge Assistant" = /datum/alt_title/bridge_assistant, "Bridge Bunny" = /datum/alt_title/bridge_bunny)
 
 /datum/alt_title/command_liaison
 	title = "Command Liaison"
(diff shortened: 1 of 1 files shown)


Heroman3003:
So, memey alt-titles are fine when its staff that wants them then?

Novacat:
I do think Bridge Bunny as an official title is... kind of odd? "Bridge Officer" sounds like a much better official title while still being very bridge bunny-esque.

Arokha:
> So, memey alt-titles are fine when its staff that wants them then?

> I do think Bridge Bunny as an official title is... kind of odd? "Bridge Officer" sounds like a much better official title while still being very bridge bunny-esque.

You will need to take that up with @JoanRisu

JoanRisu:
We can add Bridge Officer too, though I would have to brief the staff on how to handle the implied authority of the name. It should probably also go under a separate PR. Bridge Bunny follows the particular trope of "Feminine Operations Officers who practically live on the bridge and ensures smooth operation of the ship/station."

Since we're on the topic of meme titles and this was part of a previous alt-title PR: #10034 uses the whitelisting system, implying the use of a fluff title that is _unique_ to the person that applied for the whitelist and not openly available to others as an official title. Whitelist also implies "staff wants/okay with them." I commented in the PR that I was generally fine with a whitelist given that whitelists by the very process of application weeds out whatever we don't want, including whatever titles are marked as inappropriate. 

Unfortunately, the uniqueness of holding a special, recognized title that you had to sign up and be approved for was not an appealing prospect to have to navigate, so it was turned down in favor of, "Let the HoP do unique titles and we'll just add titles that is more general and can be openly used." 

I have additional thoughts on titles but I want to keep this comment on-topic.

Very-Soft:
I think bridge bunny is a little silly, but it's fine imo.

lm40:
i don't really like it but apparently it's already been decided so yolo

User-U-U:
I really am not a big fan of memey stuff personally, there has been a lot of it ingame and this seems like it will only add to it.

Novacat:
Based on github comments I have a feeling that there might have been some miscommunication. That should probably be cleared up before this gets merged.

Wickedtemp:
So can we add Chief Mlemical Officer

JoanRisu:
Not unless it's an openly available title. Otherwise, personalized titles go through HOP.

Wickedtemp:
> Not unless it's an openly available title. Otherwise, personalized titles go through HOP.

Yes, that's what I'm referring to. As an openly available title.

JoanRisu:
Rejected by vote on 06/28/2021

PR #10668, opened 2021-06-17
Adjusts some APCs to have night mode disabled by default

Very-Soft:
Adjusts APCs in areas that it makes sense (to me at least) to have night mode disabled by default, such as:

-Equipment rooms
-Surgery rooms
-Medical's common patient treatment areas in general
-The brig
-Atmos
-The engine room AND both engine submaps
-Hydroponics and Xenobotany
-Shooting/weapons testing ranges

I may have forgotten to mention a few, and it's possible I missed some things too, but I thought these areas made the most sense for night mode to be off by default.

### DIFF
--- a/maps/submaps/engine_submaps_vr/tether/engine_sme.dmm
+++ b/maps/submaps/engine_submaps_vr/tether/engine_sme.dmm
@@ -219,13 +219,13 @@
 /obj/machinery/power/apc{
 	dir = 4;
 	name = "east bump";
+	nightshift_setting = 2;
 	pixel_x = 28
 	},
 /turf/simulated/floor,
 /area/engineering/engine_room)
 "aL" = (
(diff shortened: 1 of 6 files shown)


Heroman3003:
I'd suggest public garden as well maybe, as it is also area involving plant growth, even if indirectly, and is meant to be outdoorsy-looking in the shiny sunny way.

Novacat (approved):

Very-Soft:
> I'd suggest public garden as well maybe, as it is also area involving plant growth, even if indirectly, and is meant to be outdoorsy-looking in the shiny sunny way.

I thought about it, but, it's 1 APC, and it's right next to the plants. If someone actually wants to grow stuff there, I don't think it's too unreasonable to have them pap the APC. The public garden tends to be more of a hangout zone these days than a grow zone.

PR #10675, opened 2021-06-18
Changes persistent smartfridge loss to a flat rate.

SpaceVampire:
The current issue with lossy sheet storage is that the jump from .99 stacks to 1-10 stacks is way too punishing.
It makes it so the best option is to put in less then one unless you have a ton of it, because you are likely to loose more overall.
This PR makes it so you don't have to follow weird meta-logic, but making it a flat 20% loss at all levels.
Still following the standard overall cap of of to 250 sheets per material.
I just picked the first number that seemed good to me, but I am open to hear other ones for sure.
I also did this at like 5:00am under a request, so if the way I coded it is dumb let me know.
It passed the does it compile test at the very least.

### DIFF
--- a/code/modules/persistence/storage/smartfridge.dm
+++ b/code/modules/persistence/storage/smartfridge.dm
@@ -51,8 +51,8 @@
 
 		// Delete some stacks if we want
 		if(stacks_go_missing)
-			var/fuzzy = rand(75,85)
-			count = round(count*0.01*fuzzy) // loss of 15-25% with rounding down
+			var/fuzzy = rand(55,65)
+			count = round(count*0.01*fuzzy) // loss of 35-45% with rounding down
 			if(count <= 0)
 				continue
 		
--- a/code/modules/persistence/storage/smartfridge.dm
+++ b/code/modules/persistence/storage/smartfridge.dm
@@ -51,14 +51,7 @@
 
 		// Delete some stacks if we want
 		if(stacks_go_missing)
-			var/fuzzy = rand(-5,5)
-			switch(count / max_amount)
-				if(0 to 1)
-					count -= 10+fuzzy // 1 stack or less, lose 10
-				if(1 to 10)
-					count -= max_amount+fuzzy // 1 to 10 stacks, lose a stack
-				if(10 to INFINITY)
-					count -= max_amount*3+fuzzy // 10+ stacks, lose 3 stacks
+			count -= count*0.2 // flat loss of 20%
 			if(count <= 0)
 				continue
 		
--- a/code/modules/persistence/storage/smartfridge.dm
+++ b/code/modules/persistence/storage/smartfridge.dm
@@ -51,8 +51,8 @@
 
 		// Delete some stacks if we want
 		if(stacks_go_missing)
-			var/fuzzy = rand(-5,5)*.01
-			count -= floor(count*(0.8+fuzzy)) // loss of 15-25% with rounding down
+			var/fuzzy = rand(75,85)
+			count -= round(count*0.01*fuzzy) // loss of 15-25% with rounding down
 			if(count <= 0)
 				continue
 		
--- a/code/modules/persistence/storage/smartfridge.dm
+++ b/code/modules/persistence/storage/smartfridge.dm
@@ -52,7 +52,7 @@
 		// Delete some stacks if we want
 		if(stacks_go_missing)
 			var/fuzzy = rand(75,85)
-			count -= round(count*0.01*fuzzy) // loss of 15-25% with rounding down
+			count = round(count*0.01*fuzzy) // loss of 15-25% with rounding down
 			if(count <= 0)
 				continue
 		
--- a/code/modules/persistence/storage/smartfridge.dm
+++ b/code/modules/persistence/storage/smartfridge.dm
@@ -51,7 +51,8 @@
 
 		// Delete some stacks if we want
 		if(stacks_go_missing)
-			count -= count*0.2 // flat loss of 20%
+			var/fuzzy = rand(-5,5)*.01
+			count -= floor(count*(0.8+fuzzy)) // loss of 15-25% with rounding down
 			if(count <= 0)
 				continue
 		


code/modules/persistence/storage/smartfridge.dm
```diff
@@ -51,14 +51,7 @@
...
 		if(stacks_go_missing)
-			var/fuzzy = rand(-5,5)
-			switch(count / max_amount)
-				if(0 to 1)
-					count -= 10+fuzzy // 1 stack or less, lose 10
-				if(1 to 10)
-					count -= max_amount+fuzzy // 1 to 10 stacks, lose a stack
-				if(10 to INFINITY)
-					count -= max_amount*3+fuzzy // 10+ stacks, lose 3 stacks
+			count -= count*0.2 // flat loss of 20%
```
Arokha, line 54: Firstly, this will end up setting fractional amounts on stacks. It should be at the very least rounded to a whole number.

The problem with this math is that you'll never remove it all and you'll end up in zeno's paradox land.

<google-sheets-html-origin>
[Current] | [This PR] | Also won't work | Also won't work
-- | -- | -- | --
Minus 10 | Minus *0.2 | Minus round(0.2) | Minus floor(0.2)
50 | 50 | 50 | 50
40 | 40 | 40 | 40
30 | 32 | 32 | 32
20 | 25.6 | 26 | 26
10 | 20.48 | 21 | 21
0 | 16.384 | 17 | 17
&nbsp; | 13.1072 | 14 | 14
&nbsp; | 10.48576 | 11 | 12
&nbsp; | 8.388608 | 9 | 10
&nbsp; | 6.7108864 | 7 | 8
&nbsp; | 5.36870912 | 6 | 7
&nbsp; | 4.294967296 | 5 | 6
&nbsp; | 3.435973837 | 4 | 5
&nbsp; | 2.748779069 | 3 | 4
&nbsp; | 2.199023256 | 2 | 4
&nbsp; | 1.759218604 | 2 | 4
&nbsp; | 1.407374884 | 2 | 4
&nbsp; | 1.125899907 | 2 | 4
&nbsp; | 0.9007199255 | 2 | 4
&nbsp; | 0.7205759404 | 2 | 4
&nbsp; | 0.5764607523 | 2 | 4
&nbsp; | 0.4611686018 | 2 | 4
&nbsp; | 0.3689348815 | 2 | 4
&nbsp; | 0.2951479052 | 2 | 4
&nbsp; | 0.2361183241 | 2 | 4
&nbsp; | 0.1888946593 | 2 | 4
&nbsp; | 0.1511157275 | 2 | 4
&nbsp; | 0.120892582 | 2 | 4
&nbsp; | 0.09671406557 | 2 | 4
&nbsp; | 0.07737125246 | 2 | 4
&nbsp; | 0.06189700196 | 2 | 4

</google-sheets-html-origin>

code/modules/persistence/storage/smartfridge.dm
```diff
@@ -51,14 +51,7 @@
 
 		// Delete some stacks if we want
 		if(stacks_go_missing)
-			var/fuzzy = rand(-5,5)
```
Arokha, line 54: It would be nice if it still had a tiny bit of randomness so it's not super obvious number-wise ingame. Like having 100 of everything turning into 80 of everything seems visually unappealing.

Arokha (dismissed):

code/modules/persistence/storage/smartfridge.dm
Meghan-Rossi: Good points.  I think setting count to 0.8 * itself and then taking the floor of that should work though.  That would make it go 50,40,32,25,20,16,12,9,7,5,4,3,2,1,0.  And if we want to keep the randomness we could have the multiplier vary from 0.75 to 0.85 or something like that.

Arokha: Floor(count*0.8) would reach 0, but it would take more than twice as many shifts to reach 0 as the current implementation (on average). If the concern stated is the focus ('people can metagame it to keep sheets longer') then more than doubling the length of persistence seems at best tangential and at worst contrary to that goal.

code/modules/persistence/storage/smartfridge.dm
SpaceVampire: I agree with that.
I just wanted to make it so you didn't have to meta-game the correct amount put into it.
Having it be random would make it look a lot better, especially because otherwise everything that hit the cap.
Might end up with exactly the same amount. I'll see if I can find a way to work that into the new formula.

code/modules/persistence/storage/smartfridge.dm
SpaceVampire: Good point about the whole rounding thing.
The fact they could keep sheets longer wasn't really as much my issue.
It was more that "People felt like they had to follow weird logic in order to do stuff the best way."

SpaceVampire:
Hopefully this should have taken the feedback above into account.
With making it 15-25% loss, and rounding down so that it trends to zero at a faster rate.
Someone should still double-check the code for sure, to make sure I did it correctly.
Staff can also give feedback if they want the number tweaked.
I thought about doing a more bracket based system, but I couldn't figure out a good method for it.

Very-Soft:
I should point out that we made it change because wanted to make it so that mining wouldn't just, not have a job for a long time. With the 15-25% rate, if you have 200 of something in there, it's going to be more than 4 shifts before you've even burnt through half of that. 

4 shifts is 24 hours. 

I personally would prefer that the thing landed in such a place that, if no one touches mining for a day, the thing's probably not got much of anything in it. The persistent sheet storage should, in my opinion, be a thing that makes it so that, work done late in the shift isn't totally worthless, and some shifts you don't really need to go out. It shouldn't make it so that someone can mine so good that it means no one has to mine anymore for potentially days at a time.

If someone maxxes out the sheet storage (which is 1000 sheets of everything, which, not easy, sure), it will take around 14 shifts before it will be down to under a full stack. 

That's more than 3 full days. 

Obviously, this doesn't account for science actually using it also, but, I deffo think that's way, way too generous.

SpaceVampire:
> Obviously, this doesn't account for science actually using it also, but, I deffo think that's way, way too generous.

What do you think is a more reasonable loss range? I just picked the first one that came to mind here, and am completely on board with making it deplete faster. No strong feelings about the actual balance, I just disliked the fact that you had to meta the current setup in order to get the best outcomes from it.

Arokha:
image

Very-Soft:
> What do you think is a more reasonable loss range? I just picked the first one that came to mind here, and am completely on board with making it deplete faster. No strong feelings about the actual balance, I just disliked the fact that you had to meta the current setup in order to get the best outcomes from it.

I am not 100% sure, but I am generally of the opinion that, stuff in there shouldn't last in significant quantities for much more than a day. 

Like, from a gameplay loop standpoint, for people who actually like to play mining, it seems ridiculous to me that one would want to make it so that they don't need to go mining pretty regularly. I think that the persistent storage is good for short term. If you get into a shift late and destroy the mountain, it'd be good if that contributed at least some to the next few rounds. 

But if you come on today at 3PM, delete the mountain and put all the extras into the thingy, it seems a little silly to me for one to have been able to do such a good job at that, that when you come on the following day at 3PM, there's still so much in there that you don't even need to consider going out again.

I think if people didn't want to go mining for such long stretches, they probably just wouldn't play mining except when they want to go out.

But! As to what rate exactly it should be, I am not sure, but I think it should be relatively steep, such that if you want to maintain the materials, you need to go out somewhat regularly. I PERSONALLY would prefer that it was 100% under 1 stack (50 pieces) in 4 shifts or less. But I get that that might be -too- steep.

Very-Soft:
We talked about it in a staff meeting and we decided that a flat rate of 40% was a sweet spot

SpaceVampire:
> We talked about it in a staff meeting and we decided that a flat rate of 40% was a sweet spot

Done. I made it between 35-45% because earlier people said they thought it looked nicer if there was some random element to it.

PR #10703, opened 2021-06-19
Fixes observers being able to open vore eggs

Verkister:
Being a nosy ghost is even worse when the egg gets noisily smashed open from ghost interaction.


code/game/objects/items/weapons/storage/egg_vr.dm
```diff
@@ -17,6 +17,8 @@
 	use_sound = 'sound/items/drop/flesh.ogg'
 
 /obj/item/weapon/storage/vore_egg/open(mob/user as mob)
+	if(isobserver(usr))
+		return
```
Arokha, line 21: Should be `user` instead of `usr` since you've got it passed in right there

PR #10710, opened 2021-06-20
Wild West!

Very-Soft:
I touched up the Wild West gateway. 

image

This gateway is probably just as, if not more lethal than it ever was before. You probably -NEED- a medic for this one.

This is pretty close to the way it was in our files at least. Some parts of it might work a little different. I replaced all of the conflicting things, ported sprites for a few things that were missing, and compacted it a little so that it would fit within our 140x140 map size. The wishgranter is not present at the end, because I have no idea if the stuff it gives you even works. I want to make a more appropriate one later anyway, so for now I just put some treasure at the end.

__**This map features mercenaries**__ of the varieties that would have been present for their syndicate equivalents, meaning, there are basic ranged and melee mercs on the map, so you'll want to make sure you bring some armor because WEW do they pack a wallop.

I tested the placements a bit and you should be able to engage many of them without aggroing the whole area and getting super overwhelmed as long as you don't drag them into other mercs. 

Otherwise, I replaced the existing weapon spawns on the map with random weapon spawners, because the ones that were there were guns that seemed like they wanted to be shotguns but didn't accept any ammo, and that's stupid. SO. Random guns instead.


Very-Soft:
WHAT DO YOU MEAN IT PASSED THERE'S AN ACTIVE EDGE IN THERE SOMEWHERE! >:OOO

Very-Soft:
I guess I fixed it on accident somehow it's not there anymore... = w=

Zeldazackman:
Thank you so much for doing this.
I loved the hell out of this whenever it popped up in the gateway years back and was sad that it fell into disuse due to it being so far outdated with code changes, especially with all the fun spots that it has for scenes (After the bodies and bullets are swept away heh)

code/game/turfs/flooring/flooring_vr.dm
```diff
@@ -26,4 +26,37 @@
...
 /turf/simulated/floor/tiled/freezer/cold
-	temperature = T0C - 5
\ No newline at end of file
+	temperature = T0C - 5
+
+/turf/simulated/floor/redgrid
+	name = "processing strata"
+	icon = 'icons/turf/flooring/circuit_vr.dmi'
+	icon_state = "rcircuit"
+	initial_flooring = /decl/flooring/reinforced/circuit/red
```
Arokha, line 35: Did we not already have the red grid circuit floors? I thought we did

Novacat: Nope we did not!

PR #10721, opened 2021-06-21
Allow another variation of maploading

Arokha:
```
lateload_single_pick = list(
  list("Something Z1", "Something Z2"),
  list("Something else"),
  list(
    list("Something A1", "Something A2", "Something A3"),
    list("Something B1", "Something B2", "Something B3"),
    list("Something C1", "Something C2", "Something C3")
  ),
  list("Mlem Z1", "Mlem Z2")
```


Very-Soft:
I like loading maps!

Leshana (approved):

PR #10724, opened 2021-06-22
[MIRROR] waste doesnt exist

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/8152
--------------------
backport of VOREStation/VOREStation#10711, removes all mention of the "waste" material from code because it doesn't exist, like, at all


Arokha:
0 diff

PR #10727, opened 2021-06-22
Medicine Tweaks

Novacat:
Mostly, these are based off of notes I have been collecting over the past year or so, and I just got the motivation to actually act on those notes.

- Adminordrazine buffed again
- Myelamine buffed slightly, to reduce the chances of a standard clotting injector failing to treat IB
- Bicaridine Overdose slightly more effective at treating IB
- Inaprovaline metabolizes slower, to make it more useful in its role of patient stabilization
- Bone repair autoinjector changed to a 5u autoinjector with 5u of osteodaxon. Considering how osteodaxon works, 10u is overkill and leaves the patient feeling the side effects for longer than they should.
- Clone loss autoinjector added, patterned off the new bone repair autoinjector. 5u of Rezadone to deal with clone loss. Added to uplink as well as random unknown autoinjector spawners.
- Tables and rollerbeds now have a 50% success rate of surgery.


Wickedtemp:
Guess I'm neutral towards all of these, don't see much of a point in buffing myelamine or nerfing surgery odds on rollers. 

Inaprovaline already had a slow metabolization rate and ODs at 60u so you can use quite a lot of it. The autoinjector size change makes sense, adminchem buffs are fine since its whole point is to be OP, butyeah.

PR #10729, opened 2021-06-22
Netgun Upgrade

klaasjared:
Add the required code and icons for the status indicator upgrade on the weapon. Simple upgrade but I feel it'll help people with weapon ammo status.


Arokha:
Is the net supposed to blink like that

KillianKirilenko:
> Is the net supposed to blink like that

Icondiffbot doesn't handle animated transparency very well sometimes. It should be fine if you view it in dreammaker.

PR #10735, opened 2021-06-23
Adds a coffee maker to the bar.

DigitalSquirrel95:
Adds a coffee machine to the bar.  Things that were where it is now were moved to the back room.


Novacat:
I am not really sure how this will stop people from stealing the coffee dispenser? Though, I think a better solution might be to just add a coffee dispenser to the bar.

DigitalSquirrel95:
These three cartridges are the main reasons people steal the coffee maker in the first place.  Adding them should prevent the issue entirely.

Novacat (changes requested):
Rather than changing the dispenser itself, can you just map in another coffee dispenser to the bar, please?

DigitalSquirrel95:
> 
> 
> Rather than changing the dispenser itself, can you just map in another coffee dispenser to the bar, please?

Uhhh gimme a bit to learn how to map and I'll see what I can do.

Novacat (approved):

PR #10738, opened 2021-06-24
Asteroid Janitor Closet and Bathroom

klaasjared:
Adds the requested asteroid bathroom and janitor closet with some scatted vendors that were never added.

That's it.

That's the PR.


Novacat (changes requested):
Conflict

klaasjared:
Soft ported this

PR #10739, opened 2021-06-24
Minor Map Stuff

klaasjared:
Meh stuff that may or may not get merged. One was requested (chute to freezer) and one was something that I saw that could be changed easily so I did (makes the barsign an exterior sign so you can't see it from inside the bar). Meh, ten minutes of work max.


Novacat (changes requested):
Conflict

PR #10743, opened 2021-06-24
Adds Mug with Text to loadout

Heroman3003:
Felt a momentary moment of inspiration from soneone in discord. Its a mug with text for loadout! Pretty much meant to be customized used Name and Description customization fields.


Novacat (approved):
#1 rat

PR #10756, opened 2021-06-26
The One Where I Port Modals

Arokha:
Replaces `alert()` with `tgui_alert()` and `input() in list` with `tgui_input_list()`
Also unifies the signature of input() everywhere (as in, not just the list ones) with regards to the first argument being a user.

cAsgzOlJBY

mFRpy8MZ9E

dreamseeker_0tYty3YOW3


KasparoVy:
Looks awesome! I bet it performs well too

PR #10764, opened 2021-06-27
Port /tg/ vis_contents openspace

Arokha:
Removes the openspace subsystem, as it's not needed now!
Much less CPU intensive on the server, though get_mobs_and_objs_in_view_fast() could use some looking at.
Also allows multiz lighting (sort of).

Also ports glass floors.

zrJbeRe6WG


Heroman3003:
Does examining mobs between z-levels still work?

Arokha:
> Does examining mobs between z-levels still work?

Dunno! Maybe!

PR #10769, opened 2021-06-27
Allows removing items from storage easier

Arokha:
Clicking in the area around the item also counts.

YRqEPj4XYv

This was REALLY annoying and hard to do. The storage is stupidly different between belts with their pre-sized slots and the dynamic size backpacks. UGH.

I've totally lost track of the lifecycle of some of these things I've added, someone should review that to make sure I'm not leaving them around.

### DIFF
--- a/code/game/atoms_movable.dm
+++ b/code/game/atoms_movable.dm
@@ -63,6 +63,10 @@
 
 	moveToNullspace()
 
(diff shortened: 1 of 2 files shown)


KasparoVy:
Holy shit yes, awesome! Thanks, this is handy.

PR #10777, opened 2021-06-28
Antique Stun Revolver

klaasjared:
Adds the new Vintage Stun Revolver to the game in prep for the new break-action stun revolver that is coming to the server soon by You Matter#7158. Below is a revamped version of the current revolver that will serve as an homage to the old stun revolvers and for those that prefer a 'classic' look over a sci-fi look.

stun_revolver_0-export-export

(Note: The huge chunk of code that was removed was all duplicate code that was causing errors in the vendor.)


code/game/machinery/vending_machines_vr.dm
```diff
@@ -1528,110 +1529,6 @@
...
-	..()
-
-/obj/machinery/vending/engivend/New()
-	products += list(/obj/item/clothing/glasses/omnihud/eng = 6)
-	contraband += list(/obj/item/weapon/rms = 5)
-	..()
-
-/obj/machinery/vending/medical/New()
-	products += list(/obj/item/weapon/storage/box/khcrystal = 4,/obj/item/weapon/backup_implanter = 3,
-					/obj/item/clothing/glasses/omnihud/med = 4, /obj/item/device/glasses_kit = 1,  /obj/item/weapon/storage/quickdraw/syringe_case = 4)
```
Arokha, line 1553: Um, but, these vendors won't dispense these things anymore... if you do this...?

Arokha (dismissed):

code/game/machinery/vending_machines_vr.dm
Arokha: These were duplicates!

PR #10782, opened 2021-06-28
Port Eris Trashbag of Holding

Arokha:
image


dwinters99:
I'm all for a trashbag of holding but god is that sprite ugly. Not only that, it doesn't match the other X of holding sprites.

SplinterGP:
agree with dwinters a bit, i woulndt say its ugly but feels a bit scuffed all over

Arokha:
> I'm all for a trashbag of holding but god is that sprite ugly. Not only that, it doesn't match the other X of holding sprites.

They are also being replaced.

Very-Soft:
I like it

PR #10783, opened 2021-06-28
Fixes mechfab busy overlay getting stuck

Verkister:
Fixes mechfab busy overlay not getting cleared after finishing the task.

### DIFF
--- a/code/game/mecha/mech_fabricator.dm
+++ b/code/game/mecha/mech_fabricator.dm
@@ -219,7 +219,7 @@
   * Removes the overlay to show the fab working and sets idle power usage settings. Additionally resets the description and turns off queue processing.
   */
 /obj/machinery/mecha_part_fabricator/proc/on_finish_printing()
-	cut_overlay("fab-active")
+	cut_overlays("fab-active")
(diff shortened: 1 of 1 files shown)


code/game/mecha/mech_fabricator.dm
```diff
@@ -219,7 +219,7 @@
   * Removes the overlay to show the fab working and sets idle power usage settings. Additionally resets the description and turns off queue processing.
   */
 /obj/machinery/mecha_part_fabricator/proc/on_finish_printing()
-	cut_overlay("fab-active")
+	cut_overlays("fab-active")
```
Arokha, line 222: `cut_overlays()` does not accept an iconstate as an argument. This isn't a good fix for whatever is going on. This is just passing in a string as the `priority` argument.

Arokha (changes requested):
Can't accept since this is going to cut priority overlays like the emissive blockers, etc.

Verkister:
Yeah. There's a lot of this kinda crap still left all over the code while the cut_overlay also seems to be broken.

PR #10784, opened 2021-06-28
fix stealthmins on staffwho

### LINKED ISSUE #10768: Stealthminning seems broken?
An admin was stealthminned, but they (in their normal name) showed up on the staffwho nonetheless. Also, when they used deadchat adminsay, it said their current 'fake' ckey instead of Administrator.

Stealthmins not showing up in staffwho at all, and admin deadchat messages sent 'as admin' not doxxing the current disguise

#### What actually happened
It happened.

#### Steps to reproduce
- Be normal player
- Have a stealthminned admin speak in deadchat
- Be surprised that HEADMIN(fakeckey) showed up
- Check staffwho
- Real admin is there
- Check regular/advanced who
- Only fake ckey is there

lm40:
fixes the bug part of #10768

dsay still uses your real rank which could potentially be revealing for uncommon admin ranks (such as dogmin, host, etc.), tbh i would just not use dsay if you're trying to stealth

Resolves #10768

### DIFF
--- a/code/modules/client/verbs/who.dm
+++ b/code/modules/client/verbs/who.dm
@@ -69,7 +69,7 @@
 		var/temp = ""
 		var/category = R_ADMIN
 		// VOREStation Edit - Apply stealthmin protection to all levels
-		if(C.holder.fakekey && check_rights(R_ADMIN|R_MOD, FALSE, src))	// Only admins and mods can see stealthmins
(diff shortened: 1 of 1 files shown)


Arokha:
The dsay thing is separate, so I'm going to close that issue with this, since "don't combine a feature request and a bug report into one issue, so that's your fault"

Ketrai:
I always thought this was intentional lol.

PR #10785, opened 2021-06-28
Port Eris burning barrels

Arokha:
https://user-images.githubusercontent.com/15028025/123682950-a7724d00-d819-11eb-8ac8-5293da2379de.mp4


Very-Soft:
Me like!

PR #10786, opened 2021-06-28
Updates various sound files

ShadeAware:
Updates a few sound files which were either very out of date or didn't sound all that good. The following sounds have been updated:

image
image


Very-Soft:
I should note that, some of these sounds are absurdly, ear hurtingly loud, red alert especially is turbo loud AND it's nearly thirty seconds long. This wouldn't be so bad if Byond had decent volume controls, but the program itself is pretty absurdly loud. SO, I would suggest, go through these, listen to them at 100% volume, and tone them down to not kill your ears, cuz that's how it'll be in byond.

ShadeAware:
> 
> 
> I should note that, some of these sounds are absurdly, ear hurtingly loud, red alert especially is turbo loud AND it's nearly thirty seconds long. This wouldn't be so bad if Byond had decent volume controls, but the program itself is pretty absurdly loud. SO, I would suggest, go through these, listen to them at 100% volume, and tone them down to not kill your ears, cuz that's how it'll be in byond.

Ah, yeah, I didn't really notice that at first. Ill get to that in a bit.

Arokha:
Undraft when done

ShadeAware:
Oh I completely forgot about this. I should've marked it for review ages ago...

PR #10788, opened 2021-06-28
Port various lighting sprites, directional floodlights

Arokha:
Flashlights from Eris, lamps from /tg/
Swapped the floodlights to directional lighting, given that makes sense.


Very-Soft:
Biiiiiiiiiiiig fan of the new lamps

the new flashlights are okay, better than ours

Heroman3003:
The flashlight handle looks really wonky the way it is. Full loop version woulda been better.

PR #10789, opened 2021-06-28
Port Bay microwave, utensil/kitchen knife sprites

Arokha:
dreamseeker_Nv3U88LlFW
dreammaker_ttQbZCmTw2


Ketrai:
Chef knife is lovely. I kinda wanna keep the old microwave though. Just with a cleaner shade painted over it. The eris one lacks that third dimension and looks pretty thin height wise as well. A lot of games using 2 sides vs 3 sides tend to still use a third side for objects where it makes sense to indicate a rectangular shape.

Arokha:
/tg/'s microwave looks closer to ours but is an updated sprite, go check there and see if you like that one more? Then we can have staff weigh in.

Very-Soft:
Yeah the knives look good 

I'm not crazy about the microwave. Like, I think it's a technically nicer sprite, I would be okay with it, but it does look rather flat compared to the old one, and that kind of really stands out to me in a negative way.

Heroman3003:
The new microwave looks so FLAT.

Also, just as a note, iirc we do use material system for cleavers, so might need to account for the fact that the butcher knives will be WAY darker.

Arokha:
> The new microwave looks so FLAT.
> 
> Also, just as a note, iirc we do use material system for cleavers, so might need to account for the fact that the butcher knives will be WAY darker.

The existing cleaver is DARKER, so I think it'll be fine.

For the microwave, are these better?

/tg/, TGMC, Bay:
image

Very-Soft:
This microwave doesn't stand out to me in a negative way now so it seems okay!

PR #10792, opened 2021-06-28
Eris portable freezer, medical briefcase

Arokha:
nkFXKyFotx


Ketrai:
A case where the sprites still look sufficiently 3 dimensional. The handles help a lot to add depth.

Very-Soft:
I like the look of these more than the older ones. They perhaps don't speak 'freezer' to me, but the old one didn't really either, and the new one looks nicer imo, SO!

Heroman3003:
I like them better than other proposal in the past tbh.

Side suggestion: Use medical briefcase sprite for ML3M case.

PR #10794, opened 2021-06-28
Port various backpacks

Arokha:
Mostly from Eris, with a few gaps filled in by TG, plus a new variety of backpacks that are 'sport backpacks', though it's missing some job types. ~~Suggestions on what to do? Someone can fill in the sprites, or is there a way to make them available as a choice but with a grey pack for those jobs that don't have one?~~ 🤔 Nevermind I just made it a backpack choice, people can deal with the grey ones until someone decides to sprite the missing ones!

StrongDMM_LmWEhRKtZg


ShadeAware:
The only problem with Eris sprites is that they tend to make things feel a bit too shiny and plastic-like, especially on objects they really shouldn't such as cloth.

Arokha:
> The only problem with Eris sprites is that they tend to make things feel a bit too shiny and plastic-like, especially on objects they really shouldn't such as cloth.

Sometimes true, though our current ones lack the shading of these, even if the shading could be un-lightented slightly. There's also that ALL the bags have in-hands and on-mob sprites, whereas we're missing many of those and just use the default grey or sometimes none.

SplinterGP:
> 
> 
> > The only problem with Eris sprites is that they tend to make things feel a bit too shiny and plastic-like, especially on objects they really shouldn't such as cloth.
> 
> Sometimes true, though our current ones lack the shading of these, even if the shading could be un-lightented slightly. There's also that ALL the bags have in-hands and on-mob sprites, whereas we're missing many of those and just use the default grey or sometimes none.

i'd say get someone to paint those, they look completely out of place with that color, otherwise the shapes seems fine

Ketrai:
Satchels can really use a re-imagining as opposed to this very bright and shiny shading on them. You can just desaturate them em masse with tweakpng.

Arokha:
> Satchels can really use a re-imagining as opposed to this very bright and shiny shading on them. You can just desaturate them em masse with tweakpng.

I'll let someone who knows l'Art do that, like you. :P Nobody'll die from saturated satchels for a while in the meantime.

Very-Soft:
I think these look great. I like how bright the colors are. They seem perhaps a smidge shinier than I would personally make, but really like them.

Arokha:
> I think these look great. I like how bright the colors are. They seem perhaps a smidge shinier than I would personally make, but really like them.

Can probably be fixed with a pass in an art program, taking the whites down a slight amount

Heroman3003:
I feel like they look way too shiny, and engineering/industrial ones are bit too washed out (with engineering wrench marking on what's also mining backpack) which is meh.

Very-Soft:
I really like the backpacks! I think they are brighter and way more interesting to look at. I think if we decide we want to make some adjustments to them that's fine, but outside of the scope of this particular PR, since that would take a lot of discussing about different art opinions relating to them. 

I PERSONALLY don't think they are too shiny, they are just shinier than I would have probably made them if I had made the sprite. 

I think these are a big upgrade from most of our bags.

Verkister:
Where are the straps???

PR #10798, opened 2021-06-29
Space heater and porta scrub icons

Arokha:
From /tg/

pES2wCM2dM


Very-Soft:
I really like the new ones. 

I am fond of the old heater too. I wonder how reasonable it would be to make it like an 'old' heater, for including in away things or whatnot?

Heroman3003:
I like the idea and shape, but feel like colors of 'casing' are a bit too light. Darker color was good for it, imo.

PR #10799, opened 2021-06-29
Labeller, paper bins from Eris

Arokha:
See icon diff bot.


Very-Soft:
The labler looks neat

I don't really like the new paper bin though. It looks mega flat without it's angley bit. Prefer the old one to the new one personally.

Arokha:
> The labler looks neat
> 
> I don't really like the new paper bin though. It looks mega flat without it's angley bit. Prefer the old one to the new one personally.

I am indifferent in the 3/4ths to iso war, except to say that I think most servers doing real good sprites are doing the 3/4ths thing, so that's what most of them end up being, but personally the paper bin sprite has ALWAYS bothered me, because instead of comparing it to the perspective of nearby machines it's compared to the desk it's on and it feels EH.

Verkister:
kuva
suggestion to make them more recognizable

Arokha:
> kuva
> suggestion to make them more recognizable

You are totally correct. That does look better. I went ahead and implemented that.

Heroman3003:
The new bins still look way too flat, imo. Better, but waaay too flat.

Very-Soft:
I think is better with the change to the sprite

PR #10800, opened 2021-06-29
Swap chemmaster, port beaker sprites

Arokha:
They have little label sprites as well for when you label them. And non-wooden lids I dunno why they were wood before.
Beakers from Eris, though /tg/'s are very similar, but a headmin liked these more than /tg/'s.

image


Ketrai:
Can we please have a condimaster with labeled 60 u bottles again T-T

SplinterGP:
does it have transparency ? would be good to have

Very-Soft:
I do like these more than the old ones. They look real good and depthy!

Heroman3003:
I feel like the liquid itself inside the jar shouldn't be as 'shiny' but that's kinda nitpicky.

PR #10803, opened 2021-06-29
Port Eris walkietalkie sprite

Arokha:
See icondiffbot.

I know IDB looks like it has a stray white pixel on the top left but that's not actually there:
image

I guess that's an IDB bug


SplinterGP:
for me it seems just a bit little way too block, it could be slimmier for something thats like mobile and holded on hand

Very-Soft:
I like this one, it seems more retro-tech-ey to me than the ones we currently have, which is good imo. Might be worth going through and making there be different colors of them later though.

PR #10804, opened 2021-06-29
Port Eris stock parts icons

Arokha:
Much more purdy.

Tier 4 uses their 'debug' parts, and tier 5 uses their 'alien' parts. Also have their faction parts, which are in there, but unused currently.

dreammaker_O7bu84WIo3


Leshana (approved):

Very-Soft:
they're nicer and they're bigger, so better in looks and function imo

Heroman3003:
They look good, but I do think Omni looks a bit too different from the rest for a technically people-made tier. Still generally an improvement.

PR #10805, opened 2021-06-29
Port /tg/ supermatter sprite

Arokha:
iE3tJq4iL0


Very-Soft:
New one isn't a scaled down PNG of a crystal, looks like it's actually made of shiny semitransparent material and has depth and stuff, I love it

dwinters99:
I know I've been complaining about most of these sprite changes but god I actually just can't stand this one. It's so soulless and it touches such a sacred sprite. I will genuinely be extremely sad if this goes through. These sprites don't fit and they look out of place. Just the very idea that most of these are going to be merged anyways makes me extremely sad. I must be the asshole because god I actually just can't stand any of these sprites. I can't think of any other way to say how much I will cringe every time I see this and know that it didn't have to be this way. I know it's going to get pushed through anyways but god I have to say something.

Heroman3003:
I do like the sprite as a whole, but I think adding at least some sense of otherworldness to at least active state would be big improvement. The 'downscaled photo of rocks' of old one was very low quality, but it also made SM incomprehensible and conveyed "this thing is literally painful to look at is how wrong it is" rather well.

Arokha:
> I do like the sprite as a whole, but I think adding at least some sense of otherworldness to at least active state would be big improvement. The 'downscaled photo of rocks' of old one was very low quality, but it also made SM incomprehensible and conveyed "this thing is literally painful to look at is how wrong it is" rather well.

Someone should!

PR #10806, opened 2021-06-29
Port Eris, /tg/ vendors

Arokha:
Replace smartfridge sprite so it isn't "a wall", and give boozeomat a unique sprite (previously it just stole the fridge sprite)

Adds various /tg/ vendors for wardrobes (not mapped in)


Very-Soft:
The new ones are way better!

Did you actually fill out the job vendors or nah?

ArgobargSoup:
Are we replacing our current walls with Eris or /tg/ walls?  Part of the reason the smartfridge sprite, and possibly the booze-o-mat sprite too, looked like walls is because they were essentially *used* as walls in most maps... specifically, how you'd have a smartfridge as part of the wall between hydroponics and the kitchen, so both the chef and the botanist could access the same fridge.  Even these new smartfridge and booze-o-mat sprites still follow that... they're built to fit Eris walls.

Arokha:
> The new ones are way better!
> 
> Did you actually fill out the job vendors or nah?

I did fill them with whatever is in the job's wardrobe lockers, they can be placed on the map if you wish.

Arokha:
> Are we replacing our current walls with Eris or /tg/ walls? Part of the reason the smartfridge sprite, and possibly the booze-o-mat sprite too, looked like walls is because they were essentially _used_ as walls in most maps... specifically, how you'd have a smartfridge as part of the wall between hydroponics and the kitchen, so both the chef and the botanist could access the same fridge. Even these new smartfridge and booze-o-mat sprites still follow that... they're built to fit Eris walls.

I agree with the first part, but they don't appear to be built to fit eris walls or anything. Why would the back be visible then? It should have like, the top of the machine visible (see also the air alarms and apcs and such for that wall style). Seems to be styled just like a normal vendor. Also it'll work just fine as is, just look a little weird, until some map tweaks to make it look nice.

Personally between "This HAS to go on a wall, you can't put it anywhere else, and it only works in 1 direction" and "This CAN'T go on a wall, it can go anywhere there's floor and works in all 4 directions", I'll go with the latter and change the map if necessary to accommodate those 'shared' situations with windoors or whatever's required.

ArgobargSoup:
> > Are we replacing our current walls with Eris or /tg/ walls? Part of the reason the smartfridge sprite, and possibly the booze-o-mat sprite too, looked like walls is because they were essentially _used_ as walls in most maps... specifically, how you'd have a smartfridge as part of the wall between hydroponics and the kitchen, so both the chef and the botanist could access the same fridge. Even these new smartfridge and booze-o-mat sprites still follow that... they're built to fit Eris walls.
> 
> I agree with the first part, but the new ones don't appear to be built to fit eris walls or anything. Why would the back be visible then? It should have like, the top of the machine visible (see also the air alarms and apcs and such for that wall style). Seems to be styled just like a normal vendor. Also it'll work just fine as is, just look a little weird, until some map tweaks to make it look nice.
> 
> Personally between "This HAS to go on a wall, you can't put it anywhere else, and it only works in 1 direction" and "This CAN'T go on a wall, it can go anywhere there's floor and works in all 4 directions", I'll go with the latter and change the map if necessary to accommodate those 'shared' situations with windoors or whatever's required.

image
...ah, completely disregard my "looking like it was built for a wall" thing.  I somehow completely misread the icondiff check, good lord brain.

Arokha:
> ...ah, completely disregard my "looking like it was built for a wall" thing. I somehow completely misread the icondiff check, good lord brain.

No problem. Also if you have suggestions on how to fix the map around the existing ones, do tell!

Heroman3003:
The maps should be changed where applicable to account for fact that they ARE actually used in place of walls in quite a few places (even as a literal transit point of resources between rooms, in case of kitchen and chemistry).

PR #10807, opened 2021-06-30
Replace all the overmap icons

Arokha:
Like all.

Also, redoes how 'unknown' things work to be nicer. Now if you're flying the ship, it seems 'known' to you, and has the right icon! To others it looks mysterious. Also changed it so the ships just use 4 directions but have a vector arrow.

RXcizwl9bj
dreamseeker_qistSzG146
dreamseeker_WV4f6Zv7dI


Very-Soft:
I like these way more, I can't wait to see them in game

Ketrai:
image
Clearly this should be a planet :^)

PR #10808, opened 2021-06-30
Replace conveyors, ORM sprites

Arokha:
See icondiffbot.


Very-Soft:
The new conveyors look better, smoother animation, looks nice. 

I'm less crazy about the new ore redemption machine, it looks like the same thing but more flat. I can take or leave that one. I like that it's got less hard outlines, and it matches the style of many of the other machines so it's probably fine.

Arokha:
> The new conveyors look better, smoother animation, looks nice.
> 
> I'm less crazy about the new ore redemption machine, it looks like the same thing but more flat. I can take or leave that one. I like that it's got less hard outlines, and it matches the style of many of the other machines so it's probably fine.

I did it purely for the matching, since it's the only machine that has different perspective in that area

Heroman3003:
Why's it even called the ORM when we don't reallly have ORM tho, we have old smelter, and we have an exchange vendor...

PR #10809, opened 2021-06-30
Port some /tg/ barsigns

Arokha:
Catgirl maid cafe time.

See icondiffbot.


Very-Soft:
Moar signs good, I think the cleaner style for some of the replacements tends to look better-ish

I think the syndi-cat one might be messed up on its rotation though
image

Arokha:
> Moar signs good, I think the cleaner style for some of the replacements tends to look better-ish
> 
> I think the syndi-cat one might be messed up on its rotation though
> image

Fixn't

Heroman3003:
I dun really like the adminbuse one... seems like a bit TOO meta of a joke for the server.

Also am I blind or where did the giant stripper fairy go

Arokha:
> I dun really like the adminbuse one... seems like a bit TOO meta of a joke for the server.
> 
> Also am I blind or where did the giant stripper fairy go

You are blind!

PR #10812, opened 2021-06-30
Newisher /tg/ rcd/rsf, inducer sprites

Arokha:


Very-Soft:
Most of these look mostly the same or better, me like!

PR #10813, opened 2021-06-30
Port /tg/ fireplace sprites

Arokha:
ICONDIFF!!!


Very-Soft:
Looks waaaaaaaay better than the existing one, maybe we'll actually use this thing now!

PR #10814, opened 2021-06-30
Ports TG posters and adds some new ones too!

Very-Soft:
I wrote all the descriptions too for funzies.


Ketrai:
Only thing that's a little weird here and like with the barsigns is that there's a couple of very specific /TG/ culture that feels a little out of place when the meme in question isn't really relevant on virgo. Otherwise I'd love posters one day you can walk into to get a bigger image a la holomap.

Very-Soft:
> Only thing that's a little weird here and like with the barsigns is that there's a couple of very specific /TG/ culture that feels a little out of place when the meme in question isn't really relevant on virgo. 

It doesn't really matter to me if something is specific to one server's culture or not. 'Relevancy to virgo' I don't think is really all that important when it comes to there being some funny pictures that can be on the wall.  

> Otherwise I'd love posters one day you can walk into to get a bigger image a la holomap.
I do think this would also be cool, though, I wouldn't want it to be mandatory.

Ketrai:
Definitely optional. It's just nice to be able to make more high resolution artwork to put into the game.

PR #10816, opened 2021-06-30
5 new Eris floors

Arokha:
dreamseeker_7AthRILYl5
dreamseeker_Z8NS0VjZZk


Very-Soft:
New floor good!

Ketrai:
I don't remember seeing these used anywhere on eris. The bulky corners make them feel very repetitive. They've got a plethora of floor patterns with much lighter corners or centered patterns that aren't as intrusive. Eris floors also work great with their lighting system, it's very smooth. Everything is generally full lit, and objects cast shadows. Not to mention, these don't work with a large amount of our existing flooring. Whilst station tiles are decently smooth and mesh with just about anything, when you go off station you quickly run into grittier floors with more contrast.

Arokha:
> I don't remember seeing these used anywhere on eris. The bulky corners make them feel very repetitive. They've got a plethora of floor patterns with much lighter corners or centered patterns that aren't as intrusive. Eris floors also work great with their lighting system, it's very smooth. Everything is generally full lit, and objects cast shadows. Not to mention, these don't work with a large amount of our existing flooring. Whilst station tiles are decently smooth and mesh with just about anything, when you go off station you quickly run into grittier floors with more contrast.
> 
> Edit : it does seem the wire tiles are used sparingly in the tcomms room and to bridge gaps with single tile thick strips in space. They still look somewhat out of place there.

I-I don't see how that's really relevant... we already HAVE all their other floors, so talking about how you like the other ones more dun make much sense. I'm completing the set, as it were, so mappers can use them if they want. I'm not going to switch any floors to these or anything.

Ketrai:
I mean, usually it's best only to port something if it makes sense to use. We already have a huge amount of code bloat. Unless anyone is actually going to use these, why have them at all? I'd ask around if other folks like em. I mean, they're passable if you simplify them a bit more. Eg removing the center squares.

Arokha:
> I mean, usually it's best only to port something if it makes sense to use. We already have a huge amount of code bloat. Unless anyone is actually going to use these, why have them at all? I'd ask around if other folks like em. I mean, they're passable if you simplify them a bit more. Eg removing the center squares.

This is not our development philosophy.

PR #10817, opened 2021-06-30
Port Eris rogue drone sprites

Arokha:


Very-Soft:
New good, looks scary, me like

PR #10818, opened 2021-06-30
Replace my terrible UAV sprite

Arokha:
kz1SeMgQB3

Courtesy of Eris


Very-Soft:
Wauw this looks good~ Lots of customization options it looks like!

Arokha:
> Wauw this looks good~ Lots of customization options it looks like!

I haven't implemented any of those T_T
I wanna get the UAV to fly in a little window like how cameras work now, rather than moving your view to it...

PR #10819, opened 2021-06-30
Swap some guns for Eris ones

Arokha:
Also give proper inhands for shrink, oliv and gisk and ge_
And dump all the Eris icons in there for use as people see fit.

Remember: Set both `icon` and `icon_override` to the file if it has the inhands in the same file. It will use the `item_state` if set, or the `icon_state` if it's null, with _r and _l at the end to get the inhand sprites.


SplinterGP:
screenshot pls

Very-Soft:
I like de gunz

PR #10820, opened 2021-06-30
/tg/ and TGMC sinks, toilets, replace mirror

Arokha:
They look nicer, and also have partially constructed sprites for showers and sinks so we can make them constructable and stuff.


Heroman3003:
Sprite offsets in some places will need to be adjusted on the map probably.

Very-Soft:
Me like

Arokha:
> Sprite offsets in some places will need to be adjusted on the map probably.

Can just nudge the icons or pixel_x/y on the definitions if so!

PR #10821, opened 2021-06-30
Add pews, corp sofas

Arokha:
Pews were already in sprites, but not coded, dunno where they're from.
Corp sofas from /tg/


Very-Soft:
MOAR decorations MOAAAR

PR #10822, opened 2021-06-30
Eris multitool sprite

Arokha:
image


dwinters99:
Please no. It doesn't fit. Please. These sprites just don't fit.

Heroman3003:
I dont like the yellow color of it. Old sprite's yellow would look better on it.

Very-Soft:
I do like the new sprite more than the old one. The old one is a classic, but the new one looks way more interesting and looks like a kind of evolution to the old design, with better material properties and shading going on to it. 

Me like.

ShadeAware:
Very good, looks like something you'd actually use as a tool and not just an electrified banana.

Screemonster:
This looks like an actual multimeter and not... whatever the old multitool is.

PR #10824, opened 2021-06-30
Replace cells with Eris cells

Arokha:
I took vg's recharger sprite and used vis_contents to make it show the thing you're recharging now. I think it looks pretty good. Been meaning to do that for a while.

https://user-images.githubusercontent.com/15028025/124011026-8e020a00-d9ad-11eb-94d9-8de11cc33015.mp4


Ketrai:
Since discord discourse isn't enough have a comment. These look like scifi-things. The industrial look of the old machine grade cells especially fit in better with the general feel of engineering as opposed to science. They're also plenty readable and decently grabbable as well. I don't see the eris sprites as a strict improvement here, the general improvement is too neutral to really consider it a worthwhile replacement for device cells, and negative design wise for industrial cells. The added energy indicators are nice, but only a meager benefit over the simple informative indicators of red, yellow and green.

Arokha:
If the commit isn't obvious, I've added stripes to all of the cells. Normal has no stripes, 'hi' has 1 red stripe, 'super' has 2 red stripes, and 'hyper' has a wider gold stripe. This should make it easier to identify what 'tier' cells are at a glance:
tier 1 - no stripe
tier 2 - red stripe
tier 3 - double red stripes
tier 4 - wide gold stripe

At a suggestion I also darkened the 'off' parts of the lights using a dark semi-alpha area on the overlays

Heroman3003:
They look almost nothing like power cells, at least way less than old ones. I dun like em.

TheToaster98:
I really like Eris sprites, but I don't feel like these fit at all with our other sprites. Even if the charger looks amazing, it just feels like they don't mix at all with the rest of the sprites around the server.

Very-Soft:
I do think that functionally these will be more easy to see and click and such. I'm still not crazy about the overall designs, but I do think they look more interesting than the current cells, and with the stripes on them, it should make the transition to these way easier. 

I like them more than our current cells! I would deffo like to have an even better design in the future though.

lm40:
i actually prefer the aesthetic of our current cells more in terms of shape/color schemes but these sprites are better

PR #10825, opened 2021-06-30
/tg/ stools, fix stool code

Arokha:
Look, directions

dreamseeker_NnkMpuA3fY


Heroman3003:
New shuttle chair is way too grey, drab and boring. Also, since stools are getting fixed, constructible barstool when?

Arokha:
> New shuttle chair is way too grey, drab and boring. Also, since stools are getting fixed, constructible barstool when?

There's two shuttle chairs actually, that one and the one that's placed on things like the explo shuttle. Kinda confusing, one has an underscore and the other doesn't. This isn't the one that's on the explo shuttle, I don't think!

For constructing, WYCI! Just make it sheets or whatever!

Heroman3003:
> 
> 
> > New shuttle chair is way too grey, drab and boring. Also, since stools are getting fixed, constructible barstool when?
> 
> There's two shuttle chairs actually, that one and the one that's placed on things like the explo shuttle. Kinda confusing, one has an underscore and the other doesn't. This isn't the one that's on the explo shuttle, I don't think!
> 
> For constructing, WYCI! Just make it sheets or whatever!

I believe the replaced chair is not used on explo shuttle but it IS used on tram, which makes it even worse, imo.

Arokha:
> > > New shuttle chair is way too grey, drab and boring. Also, since stools are getting fixed, constructible barstool when?
> > 
> > 
> > There's two shuttle chairs actually, that one and the one that's placed on things like the explo shuttle. Kinda confusing, one has an underscore and the other doesn't. This isn't the one that's on the explo shuttle, I don't think!
> > For constructing, WYCI! Just make it sheets or whatever!
> 
> I believe the replaced chair is not used on explo shuttle but it IS used on tram, which makes it even worse, imo.

HMM! Ok, I'll leave that sprite, and add a new chair that uses this other sprite. Also I want to rework the COOL shuttle chairs anyway, because TGMC has some that retract into the floor and that's kinda awesome.

Very-Soft:
These stools are pretty gud!

PR #10829, opened 2021-06-30
Update character slot limits, load dialog

Arokha:
Increases the number of character slots and switches the slot choice to a tgui list. Note that this doesn't actually increase the character slots, someone has to edit the real config.txt on the server, but we'll do it at the same time ~~if we remember, but we won't~~

ln5jK5zbK9


Very-Soft:
ENBIGGEN THE SLOTS

PR #10831, opened 2021-07-01
Further Map Changes

Very-Soft:
So basically I went through and moved around vending machines so that they can face south where able. I probably didn't get =all= of them, but all of the ones in public places at least.

FOR SOME OF THEM, some somewhat large mapping changes occurred, and I ended up sort of needing to retool a couple of civilian areas a little bit as a result, NAMELY the garden/kitchen area. 

I've had a few ideas for little ways that area could be improved without completely redesigning the whole thing, and a few ideas came from feedback on the current setup. My solutions definitely are not perfect, but I think they address a few extra things that people have talked about.
 I'll list the things off here:

The Bar!
Is a little wider. now, allowing the tables in the center to be 3 wide. I COULD make them 4 wide but I would rather not have them crowded up to the bar. A BIG PART of this change is two fold: The kitchen gets a larger counter to play with, since, it was admittedly pitifully small, AND, the vending machines are off to the side where they can face south. This also adds a little booth with a window and a couple of chairs passed the bathroom. Perhaps not the most romantic place, but it's a smol place to tuck away and hang with a view. I made floor tube lights a thing and put them under the tables to light the center of the bar!

image

The Kitchen!
Not a WHOLE LOT changed, but, as mentioned, the kitchen counter is 2 tiles longer now, AND, there's a 3rd spot to place food for people to get, out on the new hall table. This comes with its own shutter control so you can still open each side individually.

image

Hydroponics
Mostly I added a bunch of lights to the main area and set the trays out where they can all be used without altering their default position, as it had been in the past. There's just as many as before, and this is probably not the ideal positioning for them, but it's not really meant to be. As ever, you can always wrench and move them however you like. I also knocked out the wall leading into the back hallway break room area thingy a bit so that the dirt grow plots could be moved south a bit and out of the path into the other back rooms. Finally, the garden's front desk was moved forward so that someone sitting behind it will be able to be seen from most of the hallway, so, little bits of what I am hoping will be some hydro quality of life and RP discoverability. This does cut down a lot of the seating over there, but I couldn't think of how to move the desk forward without compromising the seating, or cutting into other areas, and cutting into other areas was the worse option for me. There's nothing to stop botanists who want to run a little cafe in there or what have you from building seats inside anyway, so, it's probably fine. One of the doors to the phoron shelter was given the same access as the hydro/kitchen door, since, it's behind the hydro desk now.

Ultimately this makes the garden a bit bigger than before, and it's already kind of huge. Not really super intentional, but it doesn't take away from any other functional area's space, so it's probably fine. I know that's probably one of the main criticisms I'll get over the garden, 'it big', and yeah, I agree, but, it's not actually taking away from anything else's space by being that way, so, I don't really see it as an issue. 

image

The Back Area!
That hallway... break room... delivery area? It's kind of smaller now, in that its connective hallway spaces have sort of been absorbed by the areas they connect to. Functionally it should be no different than before, and maintains its killer view.

image

The Tram Station!
It has some more seats now and I moved the table from IN FRONT OF THE DOOR off to the side of the south platform so it's not quite so in the way

image

Otherwise! Many of the spaces that have vending machines have been wiggled around or expanded a little bit to make space for the vending machines to be able to face south. A FEW vending machines were removed where I couldn't find space for them (particularly where there were other vending machines of the same type close by), and a select few important ones which I couldn't think of how to position better were deliberately left in their original positions.

ULTIMATELY, this shouldn't change too much functionally, a few things have been wiggled around, hopefully for the better.


Detective-Google:
Looks really good! One minor thing I'd appreciate if you could add while you're at it; 
image
Either adding windows to these red spots or removing the current one and adding them would block the tram 'airlock' doors, which would be much appreciated to stop from walking onto the tracks and getting zapped(or worse!) during lag moments.

Big fan of the new kitchen stuff though :3

Ketrai:
image
Feels really weird to have a random wall cut up the kitchen like that. 
image
one table being vertical opposed to horizontal (or vice versa) here, is pretty space-inefficient. Feels more logical to have the frequently used kitchen machines on one line, and then the tables in general two spaces away but in a nice line as well. So you can just walk up one tile and have three tables to mess with in general, that aren't spread all over the place. There's a bunch of two-tile wide walk space on the left of the kitchen that's wasted opposed to the busy area around the cooking machines, which is simply more cramped as a result. Removing the aforementioned wall would allow for the kitchen to be stretched horizontally more, making it a bit more compact and closer to the counter in general. Can the ingredient closets be moved onto one line as well? I unno if anyone else has the problem but I always forget where the flour closet it before I move it next to the fridges. These are just suggestions, despite the extra space the new kitchen simply doesn't feel as easy to navigate as before, so rejigging it a bit may help?.

Hydro looks great, the farm animal space doesn't feel awkward anymore, and the tray are finally set up in advance. (Even though some people will definitely complain about that.) Is it possible to kindly move them down one tile and move the most vital machines next to the trays up north? Easy access to seeds and fertilizer can save some trips back into the storage room due to forgetfulness etc. Of course, then the storage room will start mostly empty, and feel rather empty. Maybe box up the animals so botanist/chef can choose to release them in the little ranch or even the big hydro area? Otherwise potted plants for deco or whatnot. I'm sure something else can be added to create a fun little storage room.

I'm really not a fan of using full wall blocks at random in the middle of the room. We've got standing lamps, or windows like in the tram area. Some kind of lamp-pillar could work too, and if you don't feel up to it I don't mind making a sprite for a pillar/rack that looks bulky enough without blocking vision. I guess this pr also means my suggested alternate designs for the bar are pretty redundant? Unless it's valid to use them for any PoI's, or in an away mission that can still use a little extra content in the way of a bar design. Or for there to be some way to be a department design rotation.

Very-Soft:
> one table being vertical opposed to horizontal (or vice versa) here, is pretty space-inefficient. Feels more logical to have the frequently used kitchen machines on one line, and then the tables in general two spaces away but in a nice line as well. So you can just walk up one tile and have three tables to mess with in general, that aren't spread all over the place. There's a bunch of two-tile wide walk space on the left of the kitchen that's wasted opposed to the busy area around the cooking machines, which is simply more cramped as a result. Removing the aforementioned wall would allow for the kitchen to be stretched horizontally more, making it a bit more compact and closer to the counter in general. Can the ingredient closets be moved onto one line as well? I unno if anyone else has the problem but I always forget where the flour closet it before I move it next to the fridges. These are just suggestions, despite the extra space the new kitchen simply doesn't feel as easy to navigate as before, so rejigging it a bit may help?.

I've fiddled with a bunch of the things here but I wanted to make note of the setup for the kitchen, because I think it is mostly good as is:

image

So it might be slightly confounding at first but bare with me

The cyan lines in this are, to me, critical travel lines. The life blood of the space. People are VERY LIKELY to be moving across or parked within this space. So, having some space to move around if there's more than one person in there is pretty ideal. Notably, the vertical cyan line I have never been crazy about being 1 wide, but, having a table to put things on after you cook them that's accessible from all three of the cooking machines there and not occupied by other critical machinery is a priority, but having ways to escape this line in case it should become clogged is important logistically. Similarly, the microwaves have a table parallel to them so that things made with them can be easily set aside. There's a fair amount of 2 wide space behind the kitchen counter itself to accomodate more than one person working and serving people in that area. I should also note that when I set this up I gave chefs a wrench, because I figured some might prefer a different setup of tables and machines. There's space to customize as you wish.

The red lines are spots where, moving through isn't as critical, but are good alternatives to the cyan lines in case of blockages, as opposed to there being large sections of 1 wide space where you can't pass someone without trading spaces, especially if they are moving something you can't move through. 

The only real -wasted- space I think the kitchen has is the green line section. There could stand to be something else in that space. I'm not sure quite what. For deadshift chefs, this might be a good place to set up a few hydro trays? I think I will leave it open to customization. I do agree though that leaving the wall there was not the best call, so as you can see, I knocked it out.

PR #10835, opened 2021-07-02
Ports lots of sprites for decoration purposes

Arokha:
Just decoration for mappers. I added a new /obj/structure/decor type that allows easy varediting and setting of 'state'. A mapper can just read the comments in the file and know what states they can switch to with minimal coding. I try to credit each server in comments above each item.


Arokha:
GDI, that stupid multitool commit made it in even though I meant to undo it. I'll rebase it out tomorrow

Cyantime:
Worth noting that `/obj/structure/prop` already exists for various fluff structures, adding another that does effectively the exact same thing is a bit strange.

Arokha:
> Worth noting that `/obj/structure/prop` already exists for various fluff structures, adding another path that does effectively the exact same thing is a bit strange.

It's not strange at all when you consider the human factors of 'I had no idea that existed'. I'll move them under that type and update that type with the changes I made.

Very-Soft:
Me like decoration!

PR #10836, opened 2021-07-02
Fixes vorepanel content pick menu

Verkister:
Fixes clicking on any individual target inside the belly on vorepanel just opening an empty popup.

### DIFF
--- a/code/modules/vore/eating/vorepanel_vr.dm
+++ b/code/modules/vore/eating/vorepanel_vr.dm
@@ -552,7 +552,7 @@
 	var/list/available_options = list("Examine", "Eject", "Move")
 	if(ishuman(target))
 		available_options += "Transform"
-	intent = tgui_input_list(user, "What would you like to do with [target]?", "Vore Pick", "Examine", available_options)
+	intent = tgui_input_list(user, "What would you like to do with [target]?", "Vore Pick", available_options)
 	switch(intent)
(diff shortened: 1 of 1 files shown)


code/modules/vore/eating/vorepanel_vr.dm
```diff
@@ -552,7 +552,7 @@
 	var/list/available_options = list("Examine", "Eject", "Move")
 	if(ishuman(target))
 		available_options += "Transform"
-	intent = tgui_input_list(user, "What would you like to do with [target]?", "Vore Pick", "Examine", available_options)
+	intent = tgui_input_list(user, "What would you like to do with [target]?", "Vore Pick", available_options)
```
Arokha, line 555: Actually can you change this to a tgui_alert? It makes more sense than a list. Alerts are just the boxes with buttons. No need for a 'list picker' to pick from 3 static choices.

```suggestion
	intent = tgui_alert(user, "What would you like to do with [target]?", "Vore Pick", available_options)
```

TheFurryFeline: Huh, looks like that the previous version had a double examine option, one being useless, pfft.

Verkister: Yup, and the useless option overriding the spot where the code expects a list :v

PR #10838, opened 2021-07-02
Makes non-decal effects not use obj plane

Verkister:
This makes effects such as projectiles, sparks, and flashlight projections not cast shadows when fake ao is enabled.

### DIFF
--- a/code/__defines/_planes+layers.dm
+++ b/code/__defines/_planes+layers.dm
@@ -90,6 +90,8 @@ What is the naming convention for planes or layers?
 	#define ON_WINDOW_LAYER			3.3 // Ontop of a window
 	#define ABOVE_WINDOW_LAYER 		3.4 //Above full tile windows so wall items are clickable
 
+#define ABOVE_OBJ_PLANE				-30
+
 // Mob planes
 #define MOB_PLANE				-25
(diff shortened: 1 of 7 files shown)


Arokha:
But then sparks will be on top of mobs

Verkister:
Should they not be?

Arokha:
I don't think so, I mean they're sliding along the ground, right? That's how I've always parsed them anyway.
There's probably another way to make them not have AO, hmm... I'm rummaging around in things I learned from the /tg/ lighting port about planes, HMMM...

Arokha:
I think I'd rather you just move the planes of those particular things like /obj/effect/abstract for the flashlight overlay thing to like... a new plane above OBJ_PLANE, just add one in _planes+layers.dm that's like -30 called like ABOVE_OBJ_PLANE.

I think it's safer if it's just those 3 or so things to adjust them rather than the whole effects tree. Especially because that'd affect things like /obj/effect/dummy/lighting_obj and /obj/effect/overlay/vis which might cause shenanigans.

Verkister:
I'm pretty sure the number of things that should cast shadows on the effects tree outside the decal branch is way fewer than the number of things that shouldn't. Did some code diving and pretty much everything solid is already either a decal subtype or has their plane overriden by their own vars

PR #10846, opened 2021-07-03
Make smol holders use vis_contents

Arokha:
Much more visually appealing. You can see their status icons, runechat, they can lay down and be all cute and turn different directions or whatever.


Arokha:
I wanna improve/fix throwing smols first, and a couple other things

Arokha:
Eh fuggit. Later

PR #10847, opened 2021-07-03
Adds nail polish.

MarinaGryphon:
Available in the loadout, Cosmetics category. Can be recolored.
Currently compatible with:
- Base humanoid nails/toenails
- Hand/foot-only markings (i.e. hooves, claws, vas toes)
Future changes to make:
- Add taur compatibility. This is a large undertaking and will require custom sprites for each supported taur body, most likely. ~~or just mask the bottom 4 pixels and pray it's not a snake~~
- Repath to `/obj/item/weapon/cosmetic/nailpolish`, repath lipstick to `/obj/item/weapon/cosmetic/lipstick`, replace `lip_style` mob var and `nail_polish` organ var with a generic organ `cosmetic` var, unify the systems.
Required before merge:
- [x] Sprite nail polish remover.
- [x] Add nail polish remover.


code/modules/makeup/nailpolish.dm
```diff
@@ -0,0 +1,158 @@
...
+	regenerate_icons()
+
+/obj/item/weapon/nailpolish/proc/regenerate_icons()
+	top_underlay = icon(icon, "top")
+	top_underlay_open = icon(icon, "top-open")
+	color_underlay = icon(icon, "color")
+	color_underlay.Blend(colour, ICON_MULTIPLY)
+	color_underlay_open = icon(icon, "color-open")
+	color_underlay_open.Blend(colour, ICON_MULTIPLY)
+	update_icon()
```
Arokha, line 28: Icon manipulation is realllllyyyyy slow and CPU intensive on the server. I'd suggest making these mutable appearances (use the `mutable_appearance()` helper) and setting the `color` var. The byond default use of the `color` var is multiply, so you should achieve the same result.

code/modules/makeup/nailpolish.dm
```diff
@@ -0,0 +1,158 @@
...
+
+/obj/item/weapon/nailpolish/update_icon()
+	. = ..()
+	underlays.Cut()
+	icon_state = "[initial(icon_state)][open ? "-open" : null]"
+	if (open)
+		underlays = list(color_underlay_open, top_underlay_open)
+	else
+		underlays = list(color_underlay, top_underlay)
+
```
Arokha, line 47: Why underlays and not overlays? Would be nice if SSoverlays was used for this so the time the appearance churn takes can be accounted for by the MC.

Doing `add_overlay(list(thing1,thing2))` is also atomic since SSoverlays queues it and handles it all at once.

code/modules/makeup/nailpolish.dm
```diff
@@ -0,0 +1,158 @@
...
+/obj/item/weapon/nailpolish/proc/get_polish(var/mob/living/carbon/human/target, var/bp)
+	var/static/forbidden_parts
+	if(!forbidden_parts)
+		forbidden_parts = BP_ALL - list(BP_L_HAND, BP_R_HAND, BP_L_FOOT, BP_R_FOOT)
```
Arokha, line 51: This is not necessary in static vars local to procs. The declaration only runs once:

This:
```
/obj/proc/do_something()
    var/static/list/testlist = list("one","two")
    world.log << testlist
```
Decompiles to:
```
// path: /obj/proc/do_something
// display/inheritance name: do something
// signature (WIP): /obj/proc/do_something()
	GETVAR subvar world field "log" // L3
	GETVAR global 67 /* testlist */
	OUTPUT
	END
```
You can see the static is actually secretly a global var. It's just a sort of 'namespaced' way of having global vars. It's identical in terms of how byond uses it, anyway.

code/modules/makeup/nailpolish.dm
```diff
@@ -0,0 +1,158 @@
...
+	
+	var/bp = user.zone_sel.selecting
+	var/obj/item/organ/external/body_part = target.get_organ(bp)
+	if(!body_part)
+		to_chat(user, SPAN_WARNING("[target] is missing that limb!"))
+		return
+	if(body_part.nail_polish)
+		to_chat(user, SPAN_NOTICE("[target]'s [body_part.name] already has nail polish on!"))
+		return
+	var/datum/nail_polish/polish = get_polish(target, bp)
```
Arokha, line 90: You could probably tweak your process here a little and have `body_part.get_polish()` instead, and have it return null or the datum. That way you're not reaching into the bodypart's vars directly.

code/modules/makeup/nailpolish.dm
```diff
@@ -0,0 +1,158 @@
...
+	if(user == target)
+		user.visible_message("<b>\The [user]</b> removes their nail polish with \the [src].", "You remove your nail polish with \the [src].")
+	else
+		if(do_after(user, 2 SECONDS, target))
+			user.visible_message("<b>\The [user]</b> removes \the [target]'s nail polish with \the [src].", "You remove \the [target]'s nail polish with \the [src].")
+		else
+			to_chat(user, SPAN_NOTICE("Both you and [target] must stay still!"))
+			return
+	body_part.nail_polish = null
+	target.update_icons_body()
```
Arokha, line 147: With the above comment, I'd probably also change this to like `body_part.clear_polish()` or something.

code/modules/mob/living/carbon/human/update_icons.dm
```diff
@@ -271,6 +271,11 @@ var/global/list/damage_icon_parts = list() //see UpdateDamageIcon()
 
 			for(var/M in part.markings)
 				icon_key += "[M][part.markings[M]["color"]]"
+			
+			// VOREStation Edit Start
+			if(part.nail_polish)
+				icon_key += "_[part.nail_polish.icon]_[part.nail_polish.icon_state]_[part.nail_polish.color]"
+			// VOREStation Edit End
```
Arokha, line 278: I'm not sure doing [icon] will do something reasonable. ... What does that actually get replaced with? I see they do it up a few lines, but I'm not sure about that either.

code/modules/organs/organ_icon.dm
```diff
@@ -131,6 +131,15 @@ var/global/list/limb_icon_cache = list()
...
 				mob_icon.Blend(limb_icon_cache[cache_key], ICON_OVERLAY)
+			
+			// VOREStation edit start
+			if(nail_polish)
+				var/icon/I = new(nail_polish.icon, nail_polish.icon_state)
+				I.Blend(nail_polish.color, ICON_MULTIPLY)
+				add_overlay(I)
+				mob_icon.Blend(I, ICON_OVERLAY)
+				icon_cache_key += "_[nail_polish.icon]_[nail_polish.icon_state]_[nail_polish.color]"
+			// VOREStation edit end
```
Arokha, line 142: I think this use of icon manipulation is unavoidable given how the body sprites are done, so the previous comment about avoiding icon manipulation doesn't apply here.

I wish it wasn't so!

Arokha (changes requested):

code/modules/mob/living/carbon/human/update_icons.dm
MarinaGryphon: It converts the file path to a string. I get the feeling it'd be an internal path in the rsc or whatever if it was the result of icon ops rather than a file operation.

code/modules/organs/organ_icon.dm
MarinaGryphon: That's why it's cached; it's only done once.

Arokha (approved):

PR #10848, opened 2021-07-03
Increases the character limit on character smell and taste.

SpaceVampire:
I had a couple of cases where I wanted to add more detail to my characters smell/taste, and found that the 40ch limit made it really impossible to do that. This PR currently raises it from 40 to 400 because I had to pick some larger number and decided to add a zero, it was really pretty arbitrary. The max limit on text for a belly is 4096 to help give some comparison. I suppose that it was smaller because its easier to lick/smell people, but I really don't see anyway this could be abused. 

Hopefully this PR let me let me know if anyone had a reason that it shouldn't be raised that I hadn't thought about, or if they want to see it changed to something other then the random number I came up with first. I mean, I already doubt I would ever use all of that space, but I don't see an issue with it being larger if people want to take advantage of it.

### DIFF
--- a/code/modules/vore/eating/vorepanel_vr.dm
+++ b/code/modules/vore/eating/vorepanel_vr.dm
@@ -6,7 +6,7 @@
 #define BELLIES_NAME_MIN 2
 #define BELLIES_NAME_MAX 40
 #define BELLIES_DESC_MAX 4096
-#define FLAVOR_MAX 40
+#define FLAVOR_MAX 400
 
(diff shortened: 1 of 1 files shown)


Very-Soft:
I think this should be fine. 400 is probably a bit overkill, but like, you know, if someone doesn't want to be spammed by a 400 character message, they probably shouldn't spam the button.

Arokha:
Seems fine to me

PR #10853, opened 2021-07-03
fixes spawning characters with size traits

### LINKED ISSUE #10762: Tall trait doesn't seem to work when character is adminspawned
See title!  Had a character with the tall trait, but after being adminspawned, didn't have the appropriate change to my character sprite's scaling.

To have a tall character!

#### What actually happened
Character got adminspawned, ended up normal-sized instead of tall.

#### Steps to reproduce
- Make a character with the "Tall" neutral trait.
- Shift your scale a bit in character creation/refresh the icon so you can see the vertical stretch.
- Get adminspawned.
- Find that your character doesn't have the vertical stretch that Tall should've given them.

#### Code Revision
Commit: 82d7db23c1ee8c7c278dd8c27426b46b3dd181d5 

- This could possibly affect the other traits that affect the character scaling.

lm40:
fixes #10762 

honestly im not entirely sure that this is the best fix, but it does /work/

### DIFF
--- a/code/modules/admin/verbs/randomverbs.dm
+++ b/code/modules/admin/verbs/randomverbs.dm
@@ -273,12 +273,12 @@ Ccomp's first proc.
 	var/target = tgui_input_list(usr, "Select a ckey to allow to rejoin", "Allow Respawn Selector", GLOB.respawn_timers)
 	if(!target)
 		return
-	
+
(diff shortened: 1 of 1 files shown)


Arokha:
It's a fine fix

lm40:
WELL when looking through the stuff i didn't like that regenerate_icons doesn't update_transform and I also didn't like that there are at least three separate ways of creating a new body (latejoin, resleever, admin spawn command) but fixing through either of those two ways would have been a lot more involved

PR #10855, opened 2021-07-03
Various HUD job icons additions and adjustments

Heroman3003:
Talon now has proper HUD job icons - grey versions of normal job icons, to represent visitor-esque status

Explo icons are now dark green instead of other departmental colors

Paramedic now uses old Field Medic icon

Psych now has brand new icon, a speech-bubble-esque kind.

Replaced SM's job icon with golden hat instead of a letter, thanks @Very-Soft 

Added ability to have job icons for alt-titles. Added icons for all Intern alt-titles, Electrician and Barista

Also deletes unused 'gardener' icon since its literally same as botanist anyway.


Very-Soft:
Looks good, me like. Off duty explorers might get mistaken for off duty service I think, but I also don't care about that.

Very-Soft:
these are good yes yes

Arokha:
Also removed intern? Doesn't seem documented in your PR message

Heroman3003:
> 
> 
> ~Also removed intern? Doesn't seem documented in your PR message~ Nevermind that's just the MAIN title

Yeah during testing I realized that there was one job leftover that still had main title defined as alt-title and with new tgui stuff in char setup it showed as a bug.

PR #10860, opened 2021-07-04
Adds tgui-style input modals for text, message, and num

Leshana:
- Enhanced TextArea components to support autoFocus
- Added component and datum for accepting text, mesage, num input
- Added supporting procs to open the component with signature matching the byond input() proc
- Converted a few cases in character setup to use the new procs as a demo of capability.
tgui input text
tgui input message


Arokha:
Looks good to meeeee

PR #10908, opened 2021-07-05
Converts visible_message notice spans to look more like emotes

MarinaGryphon:
* Changes a lot of blue-text third-person `visible_message`s to look more like emotes.
* Also fixes a few issues with `\the` after an html tag instead of `\The`. Technically they're autoinserted properly if you just use `[atom]`, but still.


Arokha (approved):
I think it looks nicer this way, personally.

PR #10916, opened 2021-07-05
tgui input tweaks

Arokha:
- Tab works in text areas to select buttons instead of inserting tabs (Thanks @Leshana!)
- Removed the 'eye' and 'NT' icons from the modals


JoanRisu:
Did you happen to add ctrl+enter in for submitting a prompt?

Issue #9880, opened 2021-03-08
Pushing canisters isn't smooth anymore


Very-Soft:
#### Brief description of the bug
when you push something it jumps to the next tile, rather than smoothly sliding with you

#### What you expected to happen
pushing things got changed some time ago to move  them smoothly along with you

#### Steps to reproduce
- Push a canister

#### Code Revision
- B:release D:
Commit: df54d624d8083743d84055dc81d83b02c61ae601

Heroman3003:
I'm actually a bit uncertain about this fix, in part because of the fact that ONLY canisters were objects smoothly pushable, while everything else was pushed same way as before. Having consistency in whether you want to move with objects you're pushing or just pushing it away from you is nice, as would be having actual option to choose which you want to do via preference.

PR #10928, opened 2021-07-06
SSInput and Diagonal Movement

Leshana:
- SSInput subsystem for running movement inside the tick.
- Switch skin.dmf to use KeyUp/KeyDown procs for movement keys.
- Logic to do responsive diagonal movement.
- Timing of glides so they take twice as long as cardinal moves.
- Fix smooth glide of pulled objects when moving diagonally (pass movetime parameter to Moved)
- Fix tgui to not try and use the KeyUp verb
- Fix tram door's use of usr instead of user
- Add code to use SSInput with cardinal-only movement behind a compile flag.
- Fixes VOREStation/VOREStation#9880 - smooth pushing of canisters etc.


Very-Soft:
I like moving diagonally

ShadeAware:
Y E S

PR #10930, opened 2021-07-06
Vore fixes

Arokha:
- Fix dumby mistake making hold mode not have emotes when it's supposed to
- Remove NT branding on vore panel, HMM
- Rename panel to 'Vore Panel'
- Optimize belly processing slightly to bail REAL early if there's literally no contents


Very-Soft:
I like when vore works

PR #10931, opened 2021-07-06
Makes ghosts able to play as some mobs sometimes

Very-Soft:
Like, when chicks grow up, they can play as the chicken
they can also play as mobs that are released from cargo crates

### DIFF
--- a/code/game/objects/structures/crates_lockers/largecrate.dm
+++ b/code/game/objects/structures/crates_lockers/largecrate.dm
@@ -31,7 +31,14 @@
 		for(var/atom/movable/AM in contents)
 			if(AM.simulated)
 				AM.forceMove(T)
-
+			//VOREStation Add Start
+			if(isanimal(AM)) 
+				var/mob/living/simple_mob/AMBLINAL = AM
+				if(!AMBLINAL.mind)
+					AMBLINAL.ghostjoin = 1
+					AMBLINAL.ghostjoin_icon()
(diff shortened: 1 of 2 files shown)


Arokha:
Chaos reigns

Ketrai:
Yesss. Will these stay as open options to possess or do you have to be a ghost at the time?

Heroman3003:
At this point, having a 'proper' ghost pod panel would be great, imo. If only to keep track of it all.

Very-Soft:
> Yesss. Will these stay as open options to possess or do you have to be a ghost at the time?

They will stay open until someone claims it or the mob dies, and will be accessible from the ghost pod verb

Very-Soft:
> At this point, having a 'proper' ghost pod panel would be great, imo. If only to keep track of it all.

That may be good! Though, I think the way it is will probably be pretty fine even if things get a bit crowded in there. 

I think it will still be rare for it to get crowded though, since these are some sort of specific and not super common interactions, all of which are player initiated and can never happen spontaniously.

A clearer way to look at things though might be good at some point though.

Ketrai:
> At this point, having a 'proper' ghost pod panel would be great, imo. If only to keep track of it all.

I unno how that could happen. You could just like, hit the uh, the uh thingie. The bird crate button. Of course! That would add juuuust a couple mobs to the list. Nodnod.

PR #10944, opened 2021-07-07
Lazy attempt at eating micros inside food

Arokha:
If they're in your food you should eat them when you finish the food. Glurp.

### DIFF
--- a/code/modules/food/food/snacks.dm
+++ b/code/modules/food/food/snacks.dm
@@ -39,12 +39,20 @@
 	if(nutriment_amt)
 		reagents.add_reagent("nutriment",(nutriment_amt*2),nutriment_desc)		//VOREStation Edit: Undoes global nutrition nerf
 
-	//Placeholder for effect that trigger on eating that aren't tied to reagents.
+//Placeholder for effect that trigger on eating that aren't tied to reagents.
 /obj/item/weapon/reagent_containers/food/snacks/proc/On_Consume(var/mob/M)
 	if(!usr)
 		usr = M
 	if(!reagents.total_volume)
 		M.visible_message("<span class='notice'>[M] finishes eating \the [src].</span>","<span class='notice'>You finish eating \the [src].</span>")
+		// Embedded-in-food smol vore
+		for(var/obj/item/weapon/holder/holder in src)
+			if(holder.held_mob?.devourable && isliving(M))
+				var/mob/living/L = M
+				holder.held_mob.forceMove(L.vore_selected)
+				holder.held_mob = null
+				qdel(holder)
+		
 		usr.drop_from_inventory(src)	//so icons update :[
 
 		if(trash)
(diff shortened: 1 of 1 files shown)


Heroman3003:
I've said in discord, but will reiterate here:

This has quite a bit of prefbreak potential towards preds, and I think checking for Feeding and Spontaneous Pred preferences should be done here too, same way micro itself is checked for Devourable.

Arokha:
> I've said in discord, but will reiterate here:
> 
> This has quite a bit of prefbreak potential towards preds, and I think checking for Feeding and Spontaneous Pred preferences should be done here too, same way micro itself is checked for Devourable.

Discussed with staff, decided against.

code/modules/food/food/snacks.dm
```diff
@@ -39,12 +39,20 @@
...
-	//Placeholder for effect that trigger on eating that aren't tied to reagents.
+//Placeholder for effect that trigger on eating that aren't tied to reagents.
 /obj/item/weapon/reagent_containers/food/snacks/proc/On_Consume(var/mob/M)
 	if(!usr)
 		usr = M
 	if(!reagents.total_volume)
 		M.visible_message("<span class='notice'>[M] finishes eating \the [src].</span>","<span class='notice'>You finish eating \the [src].</span>")
+		// Embedded-in-food smol vore
+		for(var/obj/item/weapon/holder/holder in src)
+			if(holder.held_mob?.devourable && isliving(M))
```
MarinaGryphon, line 50: This is redundant for several reasons:
- `holder.held_mob` is defined as `var/mob/living/held_mob` on holder
- `var/devourable` is defined on `/mob/living`
You can just check `if(holder.held_mob?.devourable)` although if held_mob is null that implies something has gone terribly wrong somewhere and an invalid holder hasn't been qdel'd.

MarinaGryphon (changes requested):

code/modules/food/food/snacks.dm
Arokha: I think you may have misread. The check is `isliving(M)`. `M` is the mob eating the food, no relation to holder.

MarinaGryphon: Aaaaughhhhh yeah I misread. `On_Consume()` on snacks is only ever called for carbons anyway, though, so it's of dubious veracity. Could probably just change it to `On_Consume(var/mob/living/carbon/C)`.
Oh also if you do keep the isliving check please move it OUT of the loop, it only needs to be called once.

Arokha: True about the last part! Where's the only call? I want to go look

Ketrai:
> I've said in discord, but will reiterate here:
> 
> This has quite a bit of prefbreak potential towards preds, and I think checking for Feeding and Spontaneous Pred preferences should be done here too, same way micro itself is checked for Devourable.

Can't really stuff yourself into a piece of bread so this should be fine. Otherwise you can literally just spit the micro back out. If there's any prefs broken here they're very mild breaks and easily controlled. Mob vore for example, can be a lot more intrusive.

code/modules/food/food/snacks.dm
Arokha: There's also snacks.dm:290 where /mob/living can eat snacks by attacking them, so it's not only carbons.

This whole system is dumm. Should probably just give mobs an attack_food() or somesuch

PR #10945, opened 2021-07-07
4 fixes

Arokha:
- Bluespace rift admin dialog fix
- Posibrains deffo go into guts now (fix for protean brains mostly)
- No multi-queuing vorings cus you could click the same person 999 times
- Posibrains get runechat


Very-Soft:
image
>:C

PR #10946, opened 2021-07-07
Makes time/tram use REALTIMEOFDAY rather than world.time

MarinaGryphon:
REALTIMEOFDAY means that at 11:59:59 -> 12:00:00, the time will go from 864000 to 864001 (Or 863999 to 864000?) and continue. Basically, it takes into account midnight rollover and is strictly monotonic.
Needs testing. I can't really just... AFK on a local server for 6 hours, so like... Iunno.

### DIFF
--- a/code/_helpers/time.dm
+++ b/code/_helpers/time.dm
@@ -119,7 +119,8 @@ GLOBAL_VAR_INIT(round_start_time, 0)
 /var/rollovercheck_last_timeofday = 0
 /proc/update_midnight_rollover()
 	if (world.timeofday < rollovercheck_last_timeofday) //TIME IS GOING BACKWARDS!
-		return midnight_rollovers++
+		midnight_rollovers += 1
+	rollovercheck_last_timeofday = world.timeofday
 	return midnight_rollovers
 
 //Increases delay as the server gets more overloaded,
--- a/code/_helpers/time.dm
+++ b/code/_helpers/time.dm
@@ -44,7 +44,7 @@ var/next_station_date_change = 1 DAY
 
 #define duration2stationtime(time) time2text(station_time_in_ds + time, "hh:mm")
 #define worldtime2stationtime(time) time2text(GLOB.roundstart_hour HOURS + time, "hh:mm")
-#define round_duration_in_ds (GLOB.round_start_time ? world.time - GLOB.round_start_time : 0)
+#define round_duration_in_ds (GLOB.round_start_time ? REALTIMEOFDAY - GLOB.round_start_time : 0)
 #define station_time_in_ds (GLOB.roundstart_hour HOURS + round_duration_in_ds)
 
 /proc/stationtime2text()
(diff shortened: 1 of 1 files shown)


MarinaGryphon:
Did local testing by changing my system time, tram was in fact called exactly six hours after roundstart.

Little-119:
Wait, what happens without this? Like what does this fix? Asking out of curiosity, mostly

Arokha:
I have *some concerns* about this, since I see that subsystems use REALTIMEOFDAY and yet at 8pm (my time zone) every day if the server is starting, you'll end up with zany results like the server reporting 864000+ seconds to do subsystem init.

Arokha (changes requested):
This is not monotonic, and will break shift lengths significantly, resulting in shifts that may never end, despite the 'test'. These macros aren't complete. `rollovercheck_last_timeofday` is never set anywhere, despite being checked in various spots by this macro. That's why this is broken at 00:00:00 GMT (8pm my time) for subsystem inits which report zany times for their startups.


```
#define MIDNIGHT_ROLLOVER		864000
#define REALTIMEOFDAY (world.timeofday + (MIDNIGHT_ROLLOVER * MIDNIGHT_ROLLOVER_CHECK))
#define MIDNIGHT_ROLLOVER_CHECK ( rollovercheck_last_timeofday != world.timeofday ? update_midnight_rollover() : midnight_rollovers )

/var/midnight_rollovers = 0
/var/rollovercheck_last_timeofday = 0
/proc/update_midnight_rollover()
	if (world.timeofday < rollovercheck_last_timeofday)
		return midnight_rollovers++
	return midnight_rollovers
```

world.timeofday will be 0 at 00:00:00 gmt, REALTIMEOFDAY will be...
`(0 + (864000 * MIDNIGHT_ROLLOVER_CHECK)`
while MIDNIGHT_ROLLOVER_CHECK will be...
`( 0 != 0 ? update_midnight_rollover() : midnight_rollovers )`

Either way (if it's 1 second past midnight or 0 or anything) this will return 0 because in the former it will compare some positive value to rollovercheck_last_timeofday and that can't be true so it will return midnight_rollovers, which will always be 0 because the only place it's incremented is impossible to ever trigger as rollvercheck_last_timeofday will always be 0 and world.timeofday will never be less than 0.

So this breaks the monotonic-ness of this. It will jump from 846000 to 0 at the GMT date boundary. Which will mean that shift durations will suddenly become negative, and shifts will last for who-knows-how-long.

MarinaGryphon:
```
/proc/update_midnight_rollover()
	if (world.timeofday < rollovercheck_last_timeofday) //TIME IS GOING BACKWARDS!
		midnight_rollovers += 1
	rollovercheck_last_timeofday = world.timeofday
	return midnight_rollovers
```
Aurora's proc, to show proper functionality.

MarinaGryphon:
Tested a midnight rollover, autocall works now. Went from 11:30->5:28 (where I saw the hour-left warning)->5:30 tram call with no issues.

Arokha (approved):
I guess this is technically monotonic now, but I feel compelled to point out that it wouldn't be accurate in the case of waiting more than 1 day between checks. Waiting 24 hours and 1 minute between checks (because it won't have updated rollovercheck_last_timeofday in over 24 hours) will only have increased by one minute, no rollover, and you'll get a difference of 1 minute despite being 24h01m difference.

It doesn't really matter in our case, I guess. Just being PEDANTIC.

Verkister:
kuva
Might be causing this kinda thingy to happen. Real date still 14th and round duration 1:20. insta-called transfer just 50 minutes into the round

Verkister:
Also to add to the above, it flipped the date on the **noon** station time. We use 24h clock here, not 12

Verkister:
I just realized this PR doesn't sync the round time with the real time, but instead it syncs the round **_duration_**, which isn't even supposed to be synced to anything, only counting time since round start

MarinaGryphon:
> I just realized this PR doesn't sync the round time with the real time, but instead it syncs the round **_duration_**, which isn't even supposed to be synced to anything, only counting time since round start

yes, it uses real time since round start rather than byond ticks since round start, which means it ignores time dilation.

PR #10949, opened 2021-07-07
More painting tweaks

Arokha:
- Allows cross-id saving, so you can unframe a painting from a public frame and slap it in a librarian frame and it will get saved into both pools.
- Prevents paintings from loading more than once. We have a lot of nice paintings now, let's show 'em all off rather than having chances to get duplicates around the station. And I'd rather have the frames empty if there's not enough, so we can get people to draw even more!

Holy heck people are doing a lot of art and it's all awesome.


Very-Soft:
\o/

Leshana (approved):

PR #10954, opened 2021-07-07
Protects Poppy and chicks from being vored

Very-Soft:
Continuing pet and baby protections. You can still kill chicks to get meat from them or whatever, you just can't be horny with them

### DIFF
--- a/code/modules/mob/living/simple_mob/donteatpets_vr.dm
+++ b/code/modules/mob/living/simple_mob/donteatpets_vr.dm
@@ -24,6 +24,10 @@
 	digestable = 0
 	devourable = 0
 
+/mob/living/simple_mob/animal/passive/opossum/poppy
+	digestable = 0
+	devourable = 0
+
 /mob/living/carbon/human/monkey/punpun
 	digestable = 0
 	devourable = 0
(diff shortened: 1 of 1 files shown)


Ketrai:
I get this is a thing for other pets. Definitely makes sense for the rabbits who just free roam and aren't being bred for meat. But it feels a little weird for farm animals. Even the fact they're a named pet to begin with. I have worked on farms before, animals tend not to get named unless they're being kept as a pet instead of for industry. Like a personal chicken. Restricting vore is also a bit odd because it's like. "You can chop em up and eat them. But swallowing them whole is just wrong!" I know this is probably meant to mainly prevent other kinds of vore,  but you can't exactly specify. 

TL;DR. Feels oddly specific, but in the end I dont really care too much, as I'm not here to vore chickens. You do you.

Very-Soft:
> I get this is a thing for other pets. Definitely makes sense for the rabbits who just free roam and aren't being bred for meat, or poppy the possum here. But it feels a little weird for farm animals. I have worked on farms before, animals tend not to get named unless they're being kept as a pet instead of for industry. Like a personal chicken. Restricting vore is also a bit odd because it's like. "You can chop em up and eat them. But swallowing them whole is just wrong!" I know this is probably meant to mainly prevent other kinds of vore, but you can't exactly specify.
> 
> TL;DR. Feels oddly specific, but in the end I dont really care too much, as I'm not here to vore chickens. You do you.

The protection only applies to chicks. When chicks become chickens, it spawns a chicken and deletes the chick, so it loses the protection at that time.

I mostly did it because we already have other baby animals not able to be eaten. If you need meat, sure, whatever, kill them. But I don't think people should be being horny with underaged characters, and, chicks are that.

PR #10978, opened 2021-07-08
Port of tg 60064

Arokha:
Not that our openspace changes all that much. But this should be set so that it registers a listener for parent destroy and removes itself.

### DIFF
--- a/code/datums/elements/turf_transparency.dm
+++ b/code/datums/elements/turf_transparency.dm
@@ -1,4 +1,5 @@
 /datum/element/turf_z_transparency
+	element_flags = ELEMENT_DETACH
 	var/show_bottom_level = FALSE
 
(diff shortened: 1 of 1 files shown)


LemonInTheDark:
I doubt this is really needed, since you guys to my knowledge don't persist turf signals through destroy, which is the only real reason I had to do this

Arokha:
> I doubt this is really needed, since you guys to my knowledge don't persist turf signals through destroy, which is the only real reason I had to do this

I don't see anywhere we remove them, but I guess if you new() a turf, despite having the same ref and whatnot, it must reset all the vars including the signal lists and such, hmm.

Well it shouldn't hurt anything I imagine, and will prevent the future case of us porting turf changing procs and forgetting to port this.

Arokha:
Actually we don't appear to even call Destroy() on changing turfs.............

PR #10987, opened 2021-07-09
Adds PAI spawn points.

H0lySquirr3l:
Just Added some PAI spawn locations by request of a PAI player. The current only spawn point not directly carried by a player is in the RD's office which it does not have access to escape from.

image

image

image

image

image

image

### DIFF
--- a/maps/tether/tether-01-surface1.dmm
+++ b/maps/tether/tether-01-surface1.dmm
@@ -7487,6 +7487,9 @@
 	dir = 1
 	},
 /obj/structure/closet/crate,
+/obj/item/device/paicard{
+	pixel_x = 4
+	},
 /turf/simulated/floor/tiled/techfloor,
 /area/tether/surfacebase/public_garden_one)
 "amC" = (
@@ -34023,6 +34026,9 @@
 /obj/machinery/atmospherics/pipe/simple/hidden/supply{
 	dir = 9
 	},
+/obj/item/device/paicard{
+	pixel_x = 4
+	},
 /turf/simulated/floor/tiled,
 /area/crew_quarters/visitor_laundry)
 "jCv" = (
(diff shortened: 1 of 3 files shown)


H0lySquirr3l:
Pre emptive waiting to be told my maps are wierd.

Arokha (changes requested):
`Your maps are weird`

Needs to be in tgm format. I suggest using StrongDMM just set to tgm format. It's included in the repo so you can just run strongdmm.exe from the tools/strongdmm folder. You can reopen these maps in it, set it to tgm format, and save them again, and update your PR with those. It should be better.

H0lySquirr3l:
> `Your maps are weird`
> 
> Needs to be in tgm format. I suggest using StrongDMM just set to tgm format. It's included in the repo so you can just run strongdmm.exe from the tools/strongdmm folder. You can reopen these maps in it, set it to tgm format, and save them again, and update your PR with those. It should be better.

Wana know the funny thing? .... I did exactly everything you just told me to do. Even making sure it saved with tgm.

H0lySquirr3l:
Oh well... I'll just give it another shot. Is not the end of the world. Coding seems to just not understand me. But mapping actively hates me.

Arokha:
In looking, your z03 map saved in tgm and is fine... the other two are not.

Maybe you need to pick tgm on each map before saving? Usually strongdmm uses whatever format it was in when you loaded it, but I suppose if you edited them in dreammaker instead of strongdmm, saved, then opened in strongdmm z1, z2, z3, changed to tgm, and hit save, it probably just saved z3 in tgm.

I don't know if those are the motions you did, but it's a possible set of actions that might result in this z1=dmm, z2=dmm, z3=tgm

H0lySquirr3l:
> In looking, your z03 map saved in tgm and is fine... the other two are not.
> 
> Maybe you need to pick tgm on each map before saving? Usually strongdmm uses whatever format it was in when you loaded it, but I suppose if you edited them in dreammaker instead of strongdmm, saved, then opened in strongdmm z1, z2, z3, changed to tgm, and hit save, it probably just saved z3 in tgm.
> 
> I don't know if those are the motions you did, but it's a possible set of actions that might result in this z1=dmm, z2=dmm, z3=tgm

Nope then I know exactly what happened. "Save all" doesn't save maps as tgm in strongdmm. Z3 was the only one I hit the normal save on. Very wierd....

H0lySquirr3l:
ok I was sure to hit SAVE on each map individually. lets see what happens now.

PR #10992, opened 2021-07-09
Allows players to name their mobs.

Ketrai:
Quick chompstation port.

- You can now name a vore mob you control, and change the description.
- Only works once, need admin support if you mess it up.
- Also includes some tweaks for the weretiger I forgot to add, such as hands to pick stuff up, and not being super-duper slow.


Ketrai:
Gimme a mo before merging this. Going to see about porting pounces too~

Ketrai:
Oh well I won't actually have time to port the pounce for a week. Holidays. Go ahead and merge then.

PR #11002, opened 2021-07-09
Move like every clothing icon file

Arokha:
MMMMMM. That was fun.

Note to downstreams: This removes the `_s` required at the end of every uniform-slot-clothing icon state. You might need to just rename some states in your dmis. Shouldn't need any code changes. If this DOES mean you need to do code changes, then you were probably unnecessarily specifying what icons clothes use when they could have inherited it.

image

I think this will be easier for people to understand. Glasses in eyes/item.dmi for the items, eyes/mob.dmi for the mob, etc.
No more "Well it's icons/obj/glasses.dmi for the item and icons/mob/eyes.dmi for the mob sprites" or just having zany names.

I'm sure uh, I broke a thing or two, it was a lot of lines changed. I tested some 'sensitive' stuff like teshari sprites, and fixed up a few backpack state names while I was in there, and merged some of their needlessly separate files.


Little-119:
what was the _s for?

Very-Soft:
Tested and reported some issues, after some fixxes things seemed to work as I expected.

Arokha:
> what was the _s for?

No idea, it's been that way for at least 6 years. I remember when I wrote the current on-mob icon system that I had to add a little bit of code to just append '_s' to every uniform icon to keep it that way.

If I had to guess, maybe it was from a time where uniforms and something else were in the same .dmi file. Maybe they used to be (as in like 7 years ago) in the same file with their item icons and the _s versions were the onmob ones? Don't know. There doesn't seem to be any reason at all right now other than to annoy people.

PR #11023, opened 2021-07-11
Adds a bunch of unit tests.

MarinaGryphon:
Ports a bunch of unit tests from an old Bay fork I had laying around, and fixed them up for Polariscode.

PR #11046, opened 2021-07-12
Tracks random c_tags on areas to prevent duplicates.

MarinaGryphon:
Should fix the last error in #11023.

### DIFF
--- a/code/game/machinery/camera/camera.dm
+++ b/code/game/machinery/camera/camera.dm
@@ -65,6 +65,12 @@
 	if(!c_tag)
 		var/area/A = get_area(src)
 		c_tag = "[A ? A.name : "Unknown"] #[rand(111,999)]"
+		if(length(used_ctags) >= 888) // The max number of unique ids from 111 to 999 inclusive
+			error("[get_area(src)] has over the maximum number of tagless cameras, and has run out of IDs!")
+		else	
+			while(c_tag in A.used_ctags))
+				c_tag = "[A ? A.name : "Unknown"] #[rand(111,999)]"
+			A.used_ctags += c_tag
 	..()
 	// VOREStation Edit End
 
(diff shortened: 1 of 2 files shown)


code/game/machinery/camera/camera.dm
```diff
@@ -65,6 +65,12 @@
 	if(!c_tag)
 		var/area/A = get_area(src)
 		c_tag = "[A ? A.name : "Unknown"] #[rand(111,999)]"
+		if(length(used_ctags) >= 888) // The max number of unique ids from 111 to 999 inclusive
+			error("[get_area(src)] has over the maximum number of tagless cameras, and has run out of IDs!")
+		else	
+			while(c_tag in A.used_ctags))
+				c_tag = "[A ? A.name : "Unknown"] #[rand(111,999)]"
+			A.used_ctags += c_tag
```
Arokha, line 73: This is scary since it will take longer and longer for each camera as they init in order, and can, if you're really unlucky, never finish.

Isn't there a better way?

Arokha: Maybe if you collide then just subtract/add to the number repeatedly until you find a free one, and if you wrap and arrive back at the original one, you can break and give up

MarinaGryphon: Could also just have them go in sequence, but then they wouldn't be random.

Arokha: I'm ok with that honestly. I don't care thaaaaaaat much. Though that does mean if someone adds ONE new camera to the map, the order gets lost if people have memorized it. But eh.

Arokha:
Tested?

MarinaGryphon:
> Tested?

Not yet, currently on vacation.

Arokha:
Teeeested?

MarinaGryphon:
> Teeeested?

No, currently taking a break.

PR #11053, opened 2021-07-13
Convert air alarm proc into define

Arokha:


Arokha:
Gonna close this for now, there's better ways

PR #11060, opened 2021-07-13
Couple vore fixes

Verkister:
Fixes a fucked up indentation left behind by the shitty conflict editor.
Fixes bellies requiring a mob inside to make any sound.


Arokha:
More commits than lines changed?

Also what's the point of doing all the emoting and stuff if there are no mobs...?

Verkister:
the sounds. altho they could probably be moved to somewhere that doesn't get skipped completely by the lack of mob inside

PR #11066, opened 2021-07-14
Fixes #11027

### LINKED ISSUE #11027: Cryoing as Talon jobs does not clear the job slot
Cryoing (or leaving via the tram or teleporter) does not clear your job slot if you are a Talon Captain/Guard/Medic/Pilot/Engineer and does not clear your name from the manifest. It is as if you did not leave, besides the fact that your body is gone.

Using 'quit the round' might be a workaround if one knows about this issue and cares to work around it. I haven't tested.

For cryoing as Talon to work like cryoing as any other job

#### What actually happened
Not that, see description of bug

#### Steps to reproduce
- Join as Talon
- Cryo as Talon
- Observe that the job slot has not opened, among other things

MarinaGryphon:
I believe that it did actually free the slot, but left the (OOC) manifest. This properly removes offmap spawn jobs from the hidden datacore.

(Also, this was partially made with Github Copilot!)

### DIFF
--- a/code/game/machinery/cryopod.dm
+++ b/code/game/machinery/cryopod.dm
@@ -491,6 +491,19 @@
 	for(var/datum/data/record/G in data_core.general)
 		if((G.fields["name"] == to_despawn.real_name))
 			qdel(G)
+	
+	// Also check the hidden version of each datacore, if they're an offmap role.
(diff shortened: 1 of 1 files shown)


MarinaGryphon:
oh no

PR #11069, opened 2021-07-14
Adds Hair Tie to all Species

Mewchild:
- Adds the tie hair inherent verb to all species, as before it was restricted to human only
           Caveat: Didn't give it to Diona or Xenohybrid due to enforced baldness. Didn't add it to shadekin because they don't have inherent verbs, and I was worried I would break something by assigning them
- Fixed the fact that Zorren had inherent verbs defined twice for them for some reason (shred limb was in one list, lick wounds was in another.


Ketrai:
For when you want something to really accentuate your suit.

Arokha:
why not

PR #11077, opened 2021-07-15
[ready] being jittery or having a chemical speedboost removes spaces, commas, and semicolons from speech

Hatterhat:
idea stolen from vgstation13/30001, see title
image

### DIFF
--- a/code/modules/mob/living/say.dm
+++ b/code/modules/mob/living/say.dm
@@ -110,10 +110,6 @@ var/list/channel_to_radio_key = new
 			S.message = stutter(S.message)
 			verb = pick("stammers","stutters")
 			. = 1
-		if(reagents.has_reagent("hyperzine")) // motor mouthed hyperzine user
-			S.message = replacetext(S.message," ","")
-			S.message = replacetext(S.message,",","")
-			S.message = replacetext(S.message,";","")
 		//VOREStation Edit Start
 		if(muffled)
 			verb = pick("muffles")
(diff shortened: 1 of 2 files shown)


PastelPrinceDan:
as a fellow hyperzine (over)user, i approve.

code/modules/mob/living/say.dm
```diff
@@ -110,6 +110,10 @@ var/list/channel_to_radio_key = new
 			S.message = stutter(S.message)
 			verb = pick("stammers","stutters")
 			. = 1
+		if(reagents.has_reagent("hyperzine")) // motor mouthed hyperzine user
```
Arokha, line 113:
```suggestion
		if(reagents?.has_reagent("hyperzine")) // motor mouthed hyperzine user
```
`reagents` is lazy

Also unfortunately this won't work for humans. We have a few different reagent holders, declared on /mob/living/carbon, such as `bloodstr` (the one you should check in this case, I imagine), `ingested`, and `touching`.

You'd have to put this in `carbon/handle_speech_problems()` or, add it to the existing `human/handle_speech_problems()` and read `bloodstr`.

Hatterhat: somehow, `reagents` on `carbon` actually does point towards `bloodstr`, but `ingested` and `touching` aren't covered, so fair point

i've got a better idea for this, though, now that i've realized that there was a `handle_speech_problems` for humans

Arokha: Ah, it sets reagents to reference bloodstr in carbon/Initialize(). That's `interesting`. Fair enough.

Screemonster:
I think this'd be funnier as a jitteriness effect so it happens from drinking a shitload of coffee too

Hatterhat:
> I think this'd be funnier as a jitteriness effect so it happens from drinking a shitload of coffee too

you have a fair point

Arokha (approved):

PR #11083, opened 2021-07-15
Adds possibility of borg module whitelisting

Heroman3003:
A bit of overhaul to how borg module list building works and allows possibility of borg module whitelisting in future. More detailed + less related changes.

Reunites total list of borg modules available to station into single one
Makes the emergency modules into global list instead of two hardcoded ones
Adds whitelisted modules list. Currently only contains Lost and Stray modules (mostly for example's sake). To access them, you need to be in alienwhitelists file with module name whitelisted for you.
Adds BoozeHound to AI shell options just because.

Also adds Gravekeeper and three syndie borgs to list of 'named module entries', although they aren't even whitelist-accessible thing.

### DIFF
--- a/code/game/jobs/whitelist.dm
+++ b/code/game/jobs/whitelist.dm
@@ -72,6 +72,23 @@ var/list/whitelist = list()
 			if(findtext(s,"[M.ckey] - All"))
 				return 1
 
+/proc/is_borg_whitelisted(mob/M, var/module)
+	//They are admin or the whitelist isn't in use
+	if(whitelist_overrides(M))
+		return 1
+
+	//You did something wrong
+	if(!M || !module)
+		return 0
+
+	//If we have a loaded file, search it
+	if(alien_whitelist)
+		for (var/s in alien_whitelist)
+			if(findtext(s,"[M.ckey] - [module]"))
+				return 1
+			if(findtext(s,"[M.ckey] - All"))
+				return 1
+
 /proc/whitelist_overrides(mob/M)
 	if(!config.usealienwhitelist)
 		return 1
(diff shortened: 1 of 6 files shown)


Arokha:
T-there already is basically a whitelisting system for borg modules, I get that they're sliiiightly different, one is for modules and one is for custom sprites, but... `code/modules/mob/living/silicon/robot/custom_sprites.dm`

If you're going to write your own, you should remove the existing one at the very least. Don't want two whitelist systems for the same thing.

Heroman3003:
The entire point of the system is that the two are definitely not same thing. One is allowing to select an extra sprite option for existing toolset, another is allowing to select unique toolset (potentially with separate set of sprites of its own) entirely. I do think that the points are different enough that, without a grander rewrite of how borg modules work, the two are different enough to warrant being separate. Especially in regards to code.

Arokha:
Can you elaborate more on what the whitelisted borgs would have?

Very-Soft:
I can say about this, that like, if this is for letting people have unique sprites, then that's cool, I don't have a problem with that, but like, if this is the kind of thing that would let someone have a super cool science surgeon borg that only they can play, then I'm probably gonna be super negative on every instance of that kind of thing.

Heroman3003:
> 
> 
> I can say about this, that like, if this is for letting people have unique sprites, then that's cool, I don't have a problem with that, but like, if this is the kind of thing that would let someone have a super cool science surgeon borg that only they can play, then I'm probably gonna be super negative on every instance of that kind of thing.

This is less meant for super powerful/extra modules, and moreso for... unorthodox ones, if someone wills so. Currently ones included, Stray and Lost, are mostly so that 'Lost Borgs' could potentially SCA to be station character while keeping Lost/Stray module toolset and unique spriteset. That toolset is not particularly powerful and is kind of an alternate variety of Standard in generalist way. Generally, I'd imagine stuff like clownborg, or other 'unique and RP focused' or 'event silicon character turned station regular' modules players may or may not  come up with in the future could fall under umbrella of whitelisted modules. Upholding balance of it would be its own topic of course, but its in no way is meant to allow 'better versions of existing modules'.

PR #11084, opened 2021-07-15
Fixes #10244 (or attempts to)

### LINKED ISSUE #10244: Railings Allow you to Climb/Clip Through Cliffs.
You can use hopping over railings to go into the tile you're not supposed to be able to enter in POIs using cliffs.

Hopping over a railing doesn't allow be to enter the tile.

#### What actually happened
My character hopped over the railing, and bypassed the fact that you're not supposed to be able to walk up a cliff.

#### Steps to reproduce
- (Step 1) Load up a POI that uses a cliff, I recommend the merc-with-a-shotgun surrounded on all sides by a cliff POI (and will be using it as an example,) (kill the merc if you do,) and have 6 steel sheets, (Or, to be exact, two metal rods, two steel flooring tiles, and four steel to make a railing,)  then, I recommend pressing into the north cliff as far as it will allow, (though it should work on all sides with this POI.)
- (Step 2)First, you place a metal rod on the ground right next to a cliff. Then, place a flooring on that, both of which can be made using steel sheets. Then, while standing on that tile and facing towards the cliff you want to clip through, make a railing using two steel sheets.
- (Step 3)Drag your character sprite onto the railing sprite. Your character will climb and clip through the invisible wall of the cliff. You can then loot, etc.
- (Step 4)You can even get back down without damaging your character using the same process, basically allowing you to loot for free. I believe you can do the same thing, or atleast do it similarly by throwing a translocator-beacon of any kind into the area that you normally can't move through, and slapping yourself to the beacon, and can get down much the same.

- (Any POI using cliffs.)

MarinaGryphon:
Untested. Should stop people from moving into places they shouldn't; may also stop people from crossing ledges/railings/etc. at all, hence the need for testing.

### DIFF
--- a/code/game/objects/structures/fitness_vr.dm
+++ b/code/game/objects/structures/fitness_vr.dm
@@ -142,7 +142,7 @@
 	if(!can_climb(user))
 		return
 
-	usr.visible_message("<span class='warning'>[user] starts climbing onto \the [src]!</span>")
+	user.visible_message("<span class='warning'>[user] starts climbing onto \the [src]!</span>")
 	LAZYDISTINCTADD(climbers, user)
 
 	if(!do_after(user,(issmall(user) ? 20 : 34)))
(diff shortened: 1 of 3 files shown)


Arokha:
Any progress?

PR #11094, opened 2021-07-15
Fixes holderless mobs being un-eggable

Verkister:
Fixes egging not working on mobs that don't have a holder type defined.

### DIFF
--- a/code/modules/vore/eating/bellymodes_datum_vr.dm
+++ b/code/modules/vore/eating/bellymodes_datum_vr.dm
@@ -175,10 +175,8 @@ GLOBAL_LIST_INIT(digest_modes, list())
 				return
 			if(isliving(C))
 				var/mob/living/M = C
-				var/mob_holder_type = /obj/item/weapon/holder
+				var/mob_holder_type = M.holder_type || /obj/item/weapon/holder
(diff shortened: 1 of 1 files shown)


code/modules/vore/eating/bellymodes_datum_vr.dm
```diff
@@ -175,8 +175,11 @@ GLOBAL_LIST_INIT(digest_modes, list())
 				return
 			if(isliving(C))
 				var/mob/living/M = C
+				var/mob_holder_type = /obj/item/weapon/holder
 				B.ownegg.w_class = M.size_multiplier * 4 //Egg size and weight scaled to match occupant.
-				var/obj/item/weapon/holder/H = new M.holder_type(B.ownegg, M)
+				if(M.holder_type)
+					mob_holder_type = M.holder_type
+				var/obj/item/weapon/holder/H = new mob_holder_type(B.ownegg, M)
```
Arokha, line 182: It won't let me do a cool commit suggestion thing because there's 1 line removed and github is picky, but you could do:
`var/mob_holder_type = M.holder_type || /obj/item/weapon/holder`

It will set mob_holder_type to the left side if it's true-y (aka if it's set) and to the right side if it's not!

PR #11100, opened 2021-07-16
Convert some bools to TRUE/FALSE instead of 1/0.

MarinaGryphon:
Prerequisite for #11086.
At some point I'll replace that with a parser-based solution, if possible.


Arokha:
Sometime you should probably change all the changes to density, opacity, and anchored to their set_X()

PR #11111, opened 2021-07-16
Fixes space heaters spamming runtimes

Verkister:
Also it seems the runtime viewer is completely busted. No wonder it's been looking suspiciously clean lately

kuva

### DIFF
--- a/code/_helpers/turfs.dm
+++ b/code/_helpers/turfs.dm
@@ -146,6 +146,9 @@
 			O.loc = X
 			if(O.light_system == STATIC_LIGHT)
 				O.update_light()
+			else
+				var/datum/component/overlay_lighting/OL = O.GetComponent(/datum/component/overlay_lighting)
+				OL?.on_parent_moved(O, X, O.dir, TRUE)
 			if(z_level_change) // The objects still need to know if their z-level changed.
 				O.onTransitZ(T.z, X.z)
 
--- a/code/_helpers/turfs.dm
+++ b/code/_helpers/turfs.dm
@@ -144,7 +144,8 @@
 	for(var/obj/O in T)
 		if(O.simulated)
 			O.loc = X
-			O.update_light()
+			if(O.light_system == STATIC_LIGHT)
+				O.update_light()
 			if(z_level_change) // The objects still need to know if their z-level changed.
 				O.onTransitZ(T.z, X.z)
 
--- a/code/game/atoms_movable.dm
+++ b/code/game/atoms_movable.dm
@@ -233,8 +233,7 @@
 		riding_datum.handle_vehicle_layer()
 		riding_datum.handle_vehicle_offsets()
 	for (var/datum/light_source/light as anything in light_sources) // Cycle through the light sources on this atom and tell them to update.
-		if(light.source_atom.light_system == STATIC_LIGHT)
-			light.source_atom.update_light()
+		light.source_atom.update_light()
 	return TRUE
 
 /atom/movable/set_dir(newdir)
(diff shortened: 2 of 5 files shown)


Arokha (changes requested):
Um, no, I mean they were changed to movable lights because that's what I wanted them to be for performance reasons.

Just fix the update_light calls, wherever those are happening.

Verkister:
Alrighty. Found and fixed a bunch of them.

No luck on the runtime viewer tho so I suppose disabling vchat is currently the only way to catch runtimes ingame at least with the example configs. Does it work properly on the live server tho? Kinda suspecting some of the provided config stuff might be gamebreakingly outdated (kinda makes it a pain in the ass to test anything with every loggable function freezing the game for several seconds)

Arokha:
The config should be relatively current, we updated it a couple months ago (the example configs, I mean).
What freezes the game for a few seconds? Not sure what you're referring to with 'loggable functions'.

image

It's possible that CRASH() doesn't get logged in the runtime viewer since it's not a 'real' runtime, but I dunno.

Arokha (approved):

code/game/atoms_movable.dm
```diff
@@ -233,7 +233,8 @@
 		riding_datum.handle_vehicle_layer()
 		riding_datum.handle_vehicle_offsets()
 	for (var/datum/light_source/light as anything in light_sources) // Cycle through the light sources on this atom and tell them to update.
-		light.source_atom.update_light()
+		if(light.source_atom.light_system == STATIC_LIGHT)
+			light.source_atom.update_light()
```
Arokha, line 237: I don't think this will ever happen and you can probably leave it as it was to keep performance.

I'm pretty sure the overlay lights are not added to the parent's light sources.

code/_helpers/turfs.dm
```diff
@@ -144,7 +144,8 @@
 	for(var/obj/O in T)
 		if(O.simulated)
 			O.loc = X
-			O.update_light()
+			if(O.light_system == STATIC_LIGHT)
+				O.update_light()
```
Arokha, line 148: In this case since it sets loc, the movable light might not come with, so you might want to do like...

```suggestion
			if(O.light_system == STATIC_LIGHT)
				O.update_light()
			else
				var/datum/component/overlay_lighting/OL = O.GetComponent(/datum/component/overlay_lighting)
				OL?.on_parent_moved(O, X, O.dir, TRUE)
```

I thiiiiink that will make sure it comes with?

code/modules/lighting/lighting_atom.dm
```diff
@@ -63,7 +63,7 @@
 		if(!isnull(set_light_on(l_on)))
 			needs_update = TRUE
 
-	if(needs_update)
+	if(needs_update && light_system == STATIC_LIGHT)
```
Arokha, line 66: Don't do this, it will prevent the stack_trace from happening, which we want to see.

Arokha (changes requested):

Verkister:
> 
> 
> The config should be relatively current, we updated it a couple months ago (the example configs, I mean).
> What freezes the game for a few seconds? Not sure what you're referring to with 'loggable functions'.
> 
> image
> 
> It's possible that CRASH() doesn't get logged in the runtime viewer since it's not a 'real' runtime, but I dunno.

I mean like anything that sends logs to database I think, like messages, attack logs, etc. With the vchat disabled those all would print out errors in the chat window about bad ports and failing connections and lacking write rights on the local machine in localized system language. Kinda suspecting a similar cause killing runtime viewer logs as well since the viewer is broken on local run code, both virgo and chomp, and even on live chomp.

Verkister:
Any chance to compare the live configs with the example configs to figure out what might be missing?

Arokha:
> Any chance to compare the live configs with the example configs to figure out what might be missing?

Using only example configs on my local server, it works fine, so the example configs aren't the problem. I'm using 514.1561 for reference.

image

Verkister:
Could it be possible that some part of the code might be relying on something external or otherwise just locally existing file that's not included in the repository or configs and therefore missing from affected setups?

PR #11112, opened 2021-07-16
Xenomorph Overhaul and fixes from Gurgstation

MisterGrimm:
This is a collection of fixes, changes, and overhauls from Gurgstation that fix the sorely untouched and dust filled code of xenomorphs.

~~**BIG NOTE HERE** : I haven't touched any .dmm files for this port, so any .dmm that has obj/effect/alien/egg in it will either need to have the egg removed, or replaced by something else like a xenomorph vore locker. I do not think staff would like the potential cascade of random players getting their hands on xenomorph queens. Thank god .dmm files doesn't spawn stuff that don't exist.~~ 
DMM files fixed and ready.
- Aliens_Allowed config changed to 1 so queens can lay eggs. Can be toggled by staff ingame.
- Replaces old broken obj/effect eggs with new obj/structure ghost pod eggs, complete with health and grow progress.
- Ghosts can spawn a single larva from a full grown egg after confirmation. Ghosts are notified that they can spawn a larva, and the egg's location in the chat screen. Clients banned from playing xenomorphs cannot spawn in eggs.
- Deleted old facehugger and embryo code, due to being broken and useless.
- Increases larva health from 25 to 50, while increasing their growth time from 200 to 325.
- Removed the carbon/alien growth status, and added a new one to alien/diona and alien/larva . For some reason it does not work right if put in carbon/alien, this was the only way I could fix it.
- Removed nutrition gain from radiation to prevent obese xenomorphs and diona.
- Tweaked xenomorph health, speed, damage, and damage resistance.
- Gave Xenomorphs adaptive sight. Xenomorphs can see turfs, mobs, and themselves through walls, but have to adapt to the darkness to be able to see well within it. Xenomorphs in darkness can see in the dark very well, xenomorphs in light cannot see in the dark very well.
- Gave xenomorph spit better sound effects.
- Removed neurotoxin spit burn damage, so sentinels can capture people for... Reasons.
- Fixed xenomorph interaction with doors, welded and bolted doors now take longer to open, but break open the door when finished.
- Removed the ability to put on hats and outer gear as xenomorphs, it caused alot of problems that required admin intervention to fix. It is still possible to put on some clothing through admin black magic.
- A few tweaks to descriptions, naming systems, and ability names for xenomorphs.
- Changed the light color on the weed node, white was not a good color.


Smdytb1:
*gurgstation
not gurglestation. play on words for GURPS, thought I should clarify.

PR #11122, opened 2021-07-17
Updated Vermin Infestation

Enzo-Leon:
Removes the vault from the possible locations for the vermin to spawn (No I didn't increase the funny number of mice or spiders)

This ALSO adds a few new locations

- Site Manager's office (We had the vault before so.. this shouldn't be an issue to replace it with such)
- Triage
- Research (Specifically the testing range and main RnD lab by the elevator

I didn't want to add too many more for now, but this was just a small little QOL fix with some extra additions.

### DIFF
--- a/code/modules/events/infestation.dm
+++ b/code/modules/events/infestation.dm
@@ -3,10 +3,12 @@
 #define LOC_CHAPEL 2
 #define LOC_LIBRARY 3
 #define LOC_HYDRO 4
-#define LOC_VAULT 5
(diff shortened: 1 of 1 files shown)


Novacat (changes requested):
Make sure to copy the #define changes to the #undef as well

Novacat (approved):

Heroman3003:
Wait a diddly darn moment. There is no option for what happens if that one dice rolls 10! There are only options for 0 through 9.

PR #11123, opened 2021-07-17
Add more explo name bloat

TheDavestDave:
Offsite Coordinator added because I think that is what Explo manger should be named.

Off site supported added for thoose non combat explos

Off site first aider for FM's that are more explo than medic

### DIFF
--- a/code/game/jobs/job/exploration_vr.dm
+++ b/code/game/jobs/job/exploration_vr.dm
@@ -47,13 +47,17 @@
 	minimal_access = list(access_eva, access_maint_tunnels, access_external_airlocks, access_pilot, access_explorer, access_gateway)
 	outfit_type = /decl/hierarchy/outfit/job/pathfinder
 	job_description = "The Pathfinder's job is to lead and manage expeditions, and is the primary authority on all off-station expeditions."
-	alt_titles = list("Expedition Lead" = /datum/alt_title/expedition_lead, "Exploration Manager" = /datum/alt_title/exploration_manager)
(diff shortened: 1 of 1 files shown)


Detective-Google:
...first aid**er**?? What. If you want to be "more Explo than medic" as an FM then maybe... Play explo?

TheDavestDave:
https://www.hse.gov.uk/firstaid/first-aider.htm

idk man, pretty sure a first aider is a nice stepping stone form explo to FM

Very-Soft:
I'm really not feeling these names. I think distancing the explorer role from the explorer role by adding an alt title for it is only going to make people salty when they inevitably find that they do in fact still have to do normal explo things if they join an exploration mission. 

Offsite coordinator sounds more like something you'd call the school teacher on a field trip

and first aider sounds gross, and I don't like it.

Novacat:
I have the same opinion as soft, and, I kinda feel that we have maybe gone a little too far with alt titles anyway.

lm40:
offsite coordinator i might be persuaded of, but 'first aider' i just don't like the sound/idea of, and 'offsite support'? really? when there's already 'offsite scout'?

that's why i thumbs-downed

KillianKirilenko:
First Aider is linguistically awkward and highly redundant -- mostly because I would expect *every explorer* to know the most fundamental steps of first aid as part of basic job training. It's like calling yourself a Chef because you threw together some buttered toast for breakfast. Anyone* can do that.

*Okay, *almost* anyone...

As for the two Offsite titles, they sound more like CentComm-side jobs. Major computer system broke down and engineering can't figure it out? Time to contact Offsite Support!

Heroman3003:
Support and First Aider just kinda sound dumb. Coordinator may work, but not in conjunction with Offsite. Maybe Expedition or Exploration Coordinator would look better?

Novacat:
We had a small discussion in staffchat and decided that we do not really want these alt titles.

PR #11136, opened 2021-07-18
Prevent synths from taking temp traits

Arokha:
Hot adaptation functions as an ad hoc 'I'm immune to space hooray' trait for synths which is not ideal. Ideally it'd be neat to have these specifically not affect that aspect of synths but that requires more work than I'm willing to try and figure out.


Heroman3003:
Are there no species with default stats similar to what Heat Adapted provides? If there are a better solution might be required either way, as synths do borrow stats like that from species.

Arokha:
> Are there no species with default stats similar to what Heat Adapted provides? If there are a better solution might be required either way, as synths do borrow stats like that from species.

Only unathi

PR #11170, opened 2021-07-20
Finishes purge of Industrial Expansion

Heroman3003:
Removes all remnants of ores added in industrial expansion. Materials originally produced from ores still exist and are still spawnable, only the ores themselves are gone.

Supposedly, despite uselessness, ores were kept to 'make people come up with use for them that is actually original', but that is doubtfully happening at this point.

Also Rutile is back to being an expedition area exclusive, as it was meant to be prior to that update.

### DIFF
--- a/code/game/objects/random/mapping.dm
+++ b/code/game/objects/random/mapping.dm
@@ -464,7 +464,7 @@
 
 /obj/random/multiple/ore_pile/item_to_spawn()
 	return pick(
-			prob(10);list(
+			/*prob(10);list(
 				/obj/item/weapon/ore/bauxite,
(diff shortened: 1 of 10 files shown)


Hatterhat:
i agree with everything but removing rutile from spawning on v3b

Heroman3003:
> 
> 
> Titanium was supposed to be harder to get. Having it be on v3b was a merge mistake, basically.

Actually, it wasn't merge error, but they were intentionally added to v3b. It's just that it was only done due to titanium being added to way more 'regular' recipies. And now that that's undone, no reason to keep it that way.

PR #11176, opened 2021-07-20
Allows hiding mobs/areas from observers without completely forbidding

Very-Soft:
Basically for more common usage for places we would prefer ghosts couldn't peek into before people are there ICly, like gateways.


Arokha:
Hidden forbidden holy ground

PR #11177, opened 2021-07-20
Reworks Shadekin eyecolors (again)

Heroman3003:
Removes all eyecolor traits

Makes shadekin 'type' determined by hue value of eyecolor in HSV automatically. For specific 'angles' of color wheel:
		if(0 to 20)
			eyecolor_type = RED_EYES
		if(21 to 50)
			eyecolor_type = ORANGE_EYES
		if(51 to 70)
			eyecolor_type = YELLOW_EYES
		if(71 to 160)
			eyecolor_type = GREEN_EYES
		if(161 to 260)
			eyecolor_type = BLUE_EYES
		if(261 to 340)
			eyecolor_type = PURPLE_EYES
		if(341 to 360)
			eyecolor_type = RED_EYES

Eyecolors will automatically increase Value/Saturation if its too low. Saturation needs to be 10 or above, Value needs to be 40 or above.

Also removes 'harmbap' attack which kind of was redundant and instead makes bite and claw sap energy based on damage dealt.


Very-Soft:
do you have any kind of graphy graph to show where these thingies fall?

Screemonster:
lemme get the boundaries off a thingy

anything below this is considered red
image
between these is orange
image
between these is yellow
image
between these is green
image
between these is blue
image
between these is purple
image
anything above this is red

values from https://alloyui.com/examples/color-picker/hsv.html

Screemonster:
would maybe have 20 as boundary between red and orange:
image

50 for orange and yellow?
image

70 is fine for the yellow/green boundary, likewise 160 for green/blue

up the blue/purple to 270?
image

could keep 330 for the purple/red, or maaaybe 340?
image

Heroman3003:
Colors adjusted based on Scree's feedback. I was picking them out basically on eyeball. Although I did go with 260 raise for blue/purple split instead of 270 as 270 seems a bit too 'obviously purple' to me.

Very-Soft:
looks good to me

Arokha (changes requested):

PR #11178, opened 2021-07-20
TTS Devices can now be used in chairs, and by alt clicking from pockets

Mewchild:
Port of https://github.com/Citadel-Station-13/Citadel-Station-13-RP/pull/3121
Quality of life for TTS devices. We can use other devices from pockets and chairs, why not TTS?

### DIFF
--- a/code/game/objects/items/devices/text_to_speech.dm
+++ b/code/game/objects/items/devices/text_to_speech.dm
@@ -7,7 +7,7 @@
 	var/named
 
 /obj/item/device/text_to_speech/attack_self(mob/user as mob)
-	if(user.incapacitated(INCAPACITATION_ALL)) //Are you in a state to actual use the device?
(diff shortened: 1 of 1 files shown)


code/game/objects/items/devices/text_to_speech.dm
```diff
@@ -7,7 +7,7 @@
 	var/named
 
 /obj/item/device/text_to_speech/attack_self(mob/user as mob)
-	if(user.incapacitated(INCAPACITATION_ALL)) //Are you in a state to actual use the device?
+	if(user.incapacitated(INCAPACITATION_KNOCKDOWN) || user.incapacitated(INCAPACITATION_DISABLED)) // EDIT: We can use the device only if we are not in certain types of incapacitation. We don't want chairs stopping us from texting!!
```
Arokha, line 10:
```suggestion
	if(user.incapacitated(INCAPACITATION_KNOCKDOWN|INCAPACITATION_DISABLED)) // EDIT: We can use the device only if we are not in certain types of incapacitation. We don't want chairs stopping us from texting!!
```

You can do it this way and it saves a second proc call! Those are bitflags so you can combine them with `|`. In fact, INCAPACITATION_ALL is just the combination of all of them.

PR #11197, opened 2021-07-21
Expands spider features

Heroman3003:
Adds some things for the future spider mechanics


Enzo-Leon:
Ah yes, more danger and terror. I love it <3

PR #11198, opened 2021-07-21
Increase inducer vendor cost

Very-Soft:
90 points is so cheap that it's a no brainer, especially considering that the scanners share points in gaining, but not in spending, Somethings as potentially powerful as a portable recharger should not be so trivial to acquire.

### DIFF
--- a/code/modules/mining/ore_redemption_machine/equipment_vendor.dm
+++ b/code/modules/mining/ore_redemption_machine/equipment_vendor.dm
@@ -136,7 +136,7 @@
 		EQUIPMENT("Fine Excavation Kit - Hand Pick",		/obj/item/weapon/pickaxe/hand,					375),
 		EQUIPMENT("Explosive Excavation Kit - Plastic Charge",/obj/item/weapon/plastique/seismic/locked,	1500),
 		EQUIPMENT("Industrial Equipment - Phoron Bore",		/obj/item/weapon/gun/magnetic/matfed/phoronbore/loaded,			3000),
-		EQUIPMENT("Industrial Equipment - Inducer",			/obj/item/weapon/inducer,						900),
+		EQUIPMENT("Industrial Equipment - Inducer",			/obj/item/weapon/inducer,						5000),
 		EQUIPMENT("Industrial Equipment - Sheet-Snatcher",	/obj/item/weapon/storage/bag/sheetsnatcher,		500),
 	)
 	prize_list["Hardsuit"] = list(
--- a/code/modules/mining/ore_redemption_machine/equipment_vendor.dm
+++ b/code/modules/mining/ore_redemption_machine/equipment_vendor.dm
@@ -136,7 +136,7 @@
 		EQUIPMENT("Fine Excavation Kit - Hand Pick",		/obj/item/weapon/pickaxe/hand,					375),
 		EQUIPMENT("Explosive Excavation Kit - Plastic Charge",/obj/item/weapon/plastique/seismic/locked,	1500),
 		EQUIPMENT("Industrial Equipment - Phoron Bore",		/obj/item/weapon/gun/magnetic/matfed/phoronbore/loaded,			3000),
-		EQUIPMENT("Industrial Equipment - Inducer",			/obj/item/weapon/inducer,						5000),
+		EQUIPMENT("Industrial Equipment - Inducer",			/obj/item/weapon/inducer,						3500),
 		EQUIPMENT("Industrial Equipment - Sheet-Snatcher",	/obj/item/weapon/storage/bag/sheetsnatcher,		500),
 	)
 	prize_list["Hardsuit"] = list(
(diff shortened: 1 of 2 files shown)


TheDavestDave:
5000 points is a just delete form the vender list kinda cost.  Science can still make them and that kinda points it's very likely a full team couldn't even pull together to get one

ZodiacShadow:
I'm with Dave on this one. At that cost, it'll likely only ever get used near the end of a shift and only if we don't have Science. At which point most of its use cases are already gone.

Very-Soft:
> 5000 points is a just delete form the vender list kinda cost. Science can still make them and that kinda points it's very likely a full team couldn't even pull together to get one. Without custom traits or vore

5000 is the mining price. Mining gets way more points, so their stuff is more expensive.

The explo one is 750.

Arokha:
I dunno about the 5000 relative to the other prices (I know that's the mining price). Compared to other mining items, that's more than a luxury survival capsule. So instead of an entire base you can pay more for an inducer?

That's also more expensive than four 1-beacon translocators, which I'd consider about equivalent to a real translocator (which only has 3, but I'll equate the annoyance of 4 different items to the extra beacon benefit). Saying an inducer is more useful than a translocator (assuming prices are based on usefulness) is eh. What about 3000?

Very-Soft:
I picked 750 because that's the same price as the carbine, a bigger snazzier phaser that is what many explorers try to get before they go on the shuttle or into the gateway, and with things like the telecubes and other high price items being not super uncommon, getting over a thousand is quite doable.

I think the inducer should be expensive enough that, if you want another expensive item it at least has the potential to be a choice between this thing and that thing, because again, unlike mining, explo shares acquisition of points as long as you have a catologer, and the inducer can be an extremely powerful item depending on what you bring. It can charge synth people and mechs and APCs and weapons. The inducer has the power to turn a pocket full of full hyper cells (which science regularly creates two power charging racks worth of to put into the public hallway) into crazy amounts of ammo for energy weapons, basically.

As for the mining price, it seems like the the explo vendor when it was added was trying to go with essentially removing a zero from the end. Some of the things in the mining vendor are a bit cheaper in that conversion though, and I think that makes sense considering that the fact that mining does not share points on acquisition. I came up with the explo price first, comparing it to the phaser carbine (750). And then considered the mining one, which was 900. 900 is more than the increased explo price, but is chump change in mining terms, so I considered the precious arrangement of (90 - 900), but thought that 7500 was definitely too much. I worried some about making it too cheap, because of explo again. They are not below begging people to use their resources to acquire things they want, and if mining can get it for a pittance, they will beg them for it. And, considering it can be reasonably powerful, charging drills and mechs and things of that nature, it is perhaps not quite as sought after for mining as it is for explo, but it still has the potential to be pretty mechanically powerful in their hands. But! They don't share points, so I thought that, knocking it down a ways from the potential 7500 to what I felt like was a more reasonable 5000. Mining folk sort of regularly boast about having several tens of thousands of points, and I feel like the inducer has a lot more potential mechanical use and benefit than most of the other things there, depending on what you bring. A mech for example, can make pretty short work of a lot of mobs and a lot of rocks, the only real limiting factor is power. But if science has 2 racks of fully charged hyper cells out front, as they commonly do these days, then, it wouldn't be hard for you to pick up a couple, carry them off, pick up an inducer, and use that to clear the entire mining Z and the underdark, and likely still have juice to go on yet more adventures. (Of course, you don't -need- an inducer for this specific scenario, but it certainly makes the prospect easier than swapping mech cells. POINT IS, the inducer can power just about anything that uses cells, and I think that has more potential utility than a stationary base with a medkit and some food in it.)

You could admittedly do similar with the phoron bore. I've seen it used to great effect in both a mining and a weapon capacity. It is likely also in for a price hike to make it similarly expensive to the carbine, because of its growing prevalence as a desirable weapon. Basically, I'm not super totally sure how it works, and wanted to play with it before I made changes,  but it seems to have an upgrade mechanic, and when upgraded, it basically instantly deletes a line of rocks, stuns any enemies it hits for 5 or so seconds, and sets them on fire. It doesn't do huge damage on hit, but it's enough that, when used effectively, it can totally lock down one or two enemies essentially indefinitely until they die, which, with the combined damage of the fire and the shot, isn't all that long. It has limited ammo, but it can also be powered by, -phoron crystals -void cells (any cell?), and inducers.

I'm not too against the idea of lowering the mining version some, but, I think for explo, 750 is expensive, but not prohibitively so, considering that they regularly pick up a gun for that much, and I think an inducer has a lot more potential use than a gun. (considering it can be used to power basically everything, including guns)

Heroman3003:
I feel important thing to consider is that mini-translocator and capsules cannot be acquired by means other than the vendor, while Inducer can. Stuff that can be acquired with help of other departments being overpriced is a fine balancing point, imo, in part because it encourages cooperating with other departments over just grinding points for yourself. Industrial Inducer being extra special, as it requires mining off-planet titanium and bringing it back and then research using it.

PR #11207, opened 2021-07-22
[kinda wip but not really] bags boots and suit storage - a second pass on the bay suits

Hatterhat:
bags: all the funny commonwealth hardsuits now hold bags on suit storage, because every other hardsuit does (except the robotics printable one but i might change that in a later commit idk)

boots: all the commonwealth suits also have advanced magboots because they're supposed to be sleek and fast, why would they not be

suit storage: kinda rolls in with the above bags thing, honestly, but - makes them able to hold more stuff in the suit storage.

### DIFF
--- a/code/datums/supplypacks/misc_vr.dm
+++ b/code/datums/supplypacks/misc_vr.dm
@@ -85,7 +85,7 @@
 /datum/supply_pack/misc/com_medical_rig
 	name = "commonwealth medical hardsuit (loaded)"
 	contains = list(
-			/obj/item/weapon/rig/baymed = 1
+			/obj/item/weapon/rig/baymed/equipped = 1
 			)
 	cost = 250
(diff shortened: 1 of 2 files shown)


Arokha:
Seems fine for most of the changes but I'd rather not have the base /storage type as suit-slottable on those, since I think it'd let you put pouches (of holding, etc) there? They're supposed to only fit in pocket slots or your hands. The ability to have like 8 backpacks worth of storage is somewhat horrifying (duffle of holding, 5 pouches of holding).

Can you check if you can put pouches there? The slot flags might save them from being placed there. What's the intent with making it /storage? The only other thing I can think of is boxes, which also seems pretty weird to be able to on a suit.

Novacat:
The robotics printable one does not have bags in suit storage because it goes on the belt instead of the back. So instead it should have belts in suit storage (which it should have already)

PR #11215, opened 2021-07-22
Small changes to AI Core

Mewchild:
Adds a few holoplants to the AI Core. Computers need pets too, and the crew isn't always willing.
Adds a jukebox to the AI core. It's a new variant that's pre-hacked, so the AI can listen to any jukebox song while they play.


Arokha:
Seems fine

PR #11220, opened 2021-07-23
New Rack and Wardrobe

GhostActual:
In the spirit of sprite updates, I'll be yoinking the `underware_wardrobe` sprite from Baystation and the `rack` sprite from Fulpstation. Minor change, but it helps with quality of life.

New Sprites:


Heroman3003:
These racks just look bad IMO

Arokha:
I've never been able to understand the racks. Can someone explain them to me? Like, I don't understand what they're supposed to be. To me, they look like one of these:

image

Like I should be using it on a desk to store papers. S'why I usually map in shelves instead.

Arokha:
code/game/objects/structures/under_wardrobe.dm:4:error: 'icons/obj/undies_wardrobe.dmi': cannot find file

GhostActual:
> I've never been able to understand the racks. Can someone explain them to me? Like, I don't understand what they're supposed to be. To me, they look like one of these:
> 
> Like I should be using it on a desk to store papers. S'why I usually map in shelves instead.

I do the same thing, but this is for existing map objects. It is infinitely easier to sprite in a replacement than it is to go through and find each instance of a `rack` and replace it with a `shelf`

Arokha:
> These racks just look bad IMO

I think all racks look bad. Replace 'em with shelves! I'd probably approve a PR for that.

PR #11221, opened 2021-07-23
New Core Sampler

GhostActual:
Ports over the Core Sampler sprites from Baystation that have been slightly modified for use.

New Sprites:
125798312-471006e6-7bd0-48ac-af56-b4a4bb3afcf7


Arokha:
Bigger! Easier to click! I like it

PR #11224, opened 2021-07-23
Add recycling machinery

Arokha:
Step 1: Crusher
- Put things with matter into crusher, via conveyor or hand or whatever you feel like
- It will output a bundle of processed materials, useless for now
- The crusher can be upgraded to preserve more of the matter from the items (matter bin and manipulator)

Step 2: Sorter
- Put the material bundles into the sorter
- It will accumulate them until it has enough for 1 sheet's worth of that material
- It will output some ground material dust

Step 3: Stamper
- Put the material dusts into the stamper
- It will output one sheet per dust

They can all accept from any direction, and they output to the direction the frame was when they were built.

https://user-images.githubusercontent.com/15028025/126848184-b0e55e21-3108-4c83-8580-a8060b288a40.mp4


Huenererschrecker:
Don't autolathes already work that way but as one machine?

Arokha:
> Don't autolathes already work that way but as one machine?

For some items, yup. Though that might change in the future.

PR #11253, opened 2021-07-26
Makes sizegun sprite actually look like a size ray

Heroman3003:
Since this 'new and improved' sizegun looks like fallout 4 institute pistol and completely ditched what cool-sci-fi-magic-gun aesthetic old one had, and since we're clearly 'newness over quality' in terms of sprite preferences, have new and improved sizegun sprite, cobbled together by absolutely-not-an-artist which looks more like a cool sizeray you'd see in shrinking porn art than all eris gun sprites combined. Also actually has the classic 'size knob' that was mentioned in the description of the gun, but nobody bothered to actually sprite onto that white rectangle of a replacement.


dwinters99:
I dislike the new sizegun as well but I don't think this is an improvement. Sorry.

SplinterGP:
i like the one from this PR since it reminds the old one.

lbnesquik:
I liked the old one from before.

PR #11255, opened 2021-07-26
Fixes photo code still using step_x/y

Verkister:
Ancient sins


Arokha:
Pixel_x and Pixel_y are handled on the next line, this will double them up

PR #11270, opened 2021-07-27
You can use health analyzers on simplemobs now

Very-Soft:
It shows their health %
now you can play vet when Ian comes in and whines at you.


Wickedtemp:
This is good. I like this.

Is there any way to help them if they're hurt?

Leshana:
Excellent idea.   I am equivocating on whether its best to do it here, or do it in `scan_mob()`

Arokha:
> This is good. I like this.
> 
> Is there any way to help them if they're hurt?

I believe you can use bandages on them

PR #11303, opened 2021-07-30
Supposed to be NO MORE SHADOWS

Arokha:


Heroman3003:
Wait, are zshadows being removed? What about ability to examine people on tiles below? The change to open space already butchered ability to shift-click zshadows, but being unable to examine entirely is pretty much last nail in the coffin for idea of dropnoms.

Arokha:
> Wait, are zshadows being removed? What about ability to examine people on tiles below? The change to open space already butchered ability to shift-click zshadows, but being unable to examine entirely is pretty much last nail in the coffin for idea of dropnoms.

zshadows already were removed, these still existing are a mistake

PR #11314, opened 2021-08-01
Warn before overwriting vore prefs as simple mobs

Arokha:
If you attempt to save your vore prefs while playing anything other than a human or a silicon, you get this message:

7wlbZI80cS

Should hopefully help prevent people wiping their bellies while playing a morph or something with another character slot loaded.

### DIFF
--- a/code/modules/vore/eating/vorepanel_vr.dm
+++ b/code/modules/vore/eating/vorepanel_vr.dm
@@ -302,6 +302,10 @@
 			return set_attr(usr, params)
 
 		if("saveprefs")
+			if(!ishuman(host) && !issilicon(host))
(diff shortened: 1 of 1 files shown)


Heroman3003:
Additional suggestion: Check if current mob's/character's name matches name in character setup and pull up a warning in that case. I do believe it would cover wider variety of potential issue-causing cases.

Arokha:
> Additional suggestion: Check if current mob's/character's name matches name in character setup and pull up a warning in that case. I do believe it would cover wider variety of potential issue-causing cases.

Ssssomewhat difficult since the mob's name var can be arbitrary if using masks and stuff, `real_name` is more reliable but nnnn sleeve shenanigans and stuff I dunno. I'll think about it, or, you're welcome to implement that.

PR #11327, opened 2021-08-02
Skyrat wallmed vendor

Arokha:
Old vs new

image


Very-Soft:
Looks okay, I think with the animation the new one has it will be k, I was gonna complain that it didn't have the cross thingy, but it does when it's animated.

GhostActual:
There is a stray pixel on the deny sprite

Arokha:
> There is a stray pixel on the deny sprite

There isn't, that's the icondiffbot being bad. That sprite is animated too and it decided to not animate it in the preview.

PR #11328, opened 2021-08-02
Skyrat field generator sprites

Arokha:
Old vs new

gUSdk1eSSE

F26MDeL1Mm


Very-Soft:
these look shiny me like

GhostActual:
Very nice, like these a lot

PR #11329, opened 2021-08-02
Skyrat emitter sprites

Arokha:
Old vs new

dreammaker_MBWUEDZYQv
dreammaker_CkNKMxFeXy

Also removes the Contain_F sprite from the file because the field generator PR moves it to it's file, so these either both need to be merged or one can be merged and the other edited.


Very-Soft:
I like this but aren't our emitters not this one anymore? I thought we were using the barrel one with wheels and different states for securedness

Arokha:
> I like this but aren't our emitters not this one anymore? I thought we were using the barrel one with wheels and different states for securedness

We are but I like these even more

GhostActual:
> We are but I like these even more

I actually like the ones that are currently used and I was going to port them upstream if Polaris didn't have them. They make more sense being on wheels.

Arokha:
Hm! Well, alright then.

PR #11331, opened 2021-08-02
Skyrat benches

Arokha:
I modified them so the padding can be colored, and the padding layers under mobs sitting on them when they face north.

dreamseeker_P7MqGvxnBu

Would be nice to have like, tails stick out the back, you know? Someone will need to play with tail layering a little for that.


Very-Soft:
this is cool I like the extra detail

PR #11335, opened 2021-08-02
Skyrat rack sprite

Arokha:
Also adds gunrack, but no fancy icons. It's just a different looking shelf basically, unless someone wants to gussy it up.

Old and older for comparison to new (and shelves and gunrack for context):
## Old/older
dreammaker_sYz7fhQ4pI

## New
dreammaker_PwJhdrIEZv


Very-Soft:
I don't think anyone will ever make a rack that I don't think looks like nonsense, but this one is a nicer looking nonsense than what we have.

KasparoVy:
Rack city.

PR #11337, opened 2021-08-02
SMART Rifle

Arokha:
Sprites are from Skyrat, but the rifle is coded from scratch by me, along with modifications to the sprites, and the creation of some sounds to accompany it (two public domain sounds where cut up and went into the creation of each sound).

Ki3ZBWUZhf

Alt-Click to ready/unready the rifle. Can only be reloaded when unready, and takes 2 seconds after readying to be prepared to fire. Needs two hands to be very accurate at all.

Mostly does A LOT of stun damage, along with some relatively minor brute damage (though may cause bleeding). If you click a mob to fire at, the projectile will home in on that mob. The homing may not be significant enough to compensate for one-handed firing. The weapon itself is fairly inaccurate if not targeting mobs.


Very-Soft:
Looks... 

OP ... 15

:p

LordSaladin:
so how do we get this?, don't see it in the protolathe?

PR #11341, opened 2021-08-04
Fancy sprited shuttles

Arokha:
A combination of Skyrat, Eris, GhostActual, and my work.

I think on other servers they just set all the icon_states by hand and that sounds like a nightmare so we're not doing that.

To map a fancy shuttle, do thus:
1. Place a /obj/effect/fancy_shuttle on the map
2. Set a fancy_shuttle_tag on it, anything unique will do
3. Map over the walls with /turf/simulated/wall/fancy_shuttle and set the tag to the same as step 2
4. If you want the shuttle's floors, map in /obj/effect/floor_decal/fancy_shuttle on all the floors you want to look like the shuttle's floors. If you want to know what those look like, you can place a /obj/effect/fancy_shuttle_floor_preview

There are invisible versions of overmap shuttle engines and sensors for placing wherever appropriate on your sprite. Note that they still have to follow the same rules as their non-invisible couterparts... sensors need vacuum, engines need clear space.

All four directions are supported for all shuttles.

If you want to edit a fancy shuttle's sprite, do thus:
1. Open the preview DMI, do whatever you want to it, though it must remain a square, and have all four directions. *See Note 1
2. Delete all states in the non-preview dmi (so the 'generic.dmi' to your 'generic_preview.dmi')
3. Import the preview dmi into the non-preview dmi with the 'Use size from file' unchecked, and the size set to 32x32. This will cut it up for you.

* Note 1: For the directions, rotating it 90 degrees four time is not good enough. All four directions have to have the bottom left corner as the origin of the ship with turfs starting there. So if you draw a north facing ship, which starts in the bottom left, you can do... 1) Rotate CCW 90 = west facing sprite, 2) Rotate 180, slide to bottom (keep 32x32 alignment!) = east facing sprite, 3) Rotate CW 90 = south facing sprite.

dreamseeker_i6hPlL5NAf

dreamseeker_Tn04kXvsrY


GhostActual:
Off, I made the Tourbus a bit too long, did I

Arokha:
> Oof, I made the Tourbus a bit too long, did I

HOW DARE nah it's fine

GhostActual:
> HOW DARE nah it's fine

D:

Detective-Google:
cool port but MAN this implementation seems cursed

Arokha:
> cool port but MAN this implementation seems cursed

I invite you to look at the incredibly cursed Skyrat dmi files for these shuttles. Each one is different, some start at 1,1 and some start at 0,0, some are just backwards, like the escape shuttle which appears to start in like, the top right corner. It's like, turbo nightmare.

At a few points I considered doing the icon_states thing but in the end gave up due to tedium. TRUST ME, this is MUCH faster than even normal walls, which cut/apply at least 4 overlays every time they update their icons, after looking up the materials, doing strbuilding to get names for states, etc. This 'get one icon' thing is trivial in comparison.

Very-Soft:
Looks cool!! C:

dwinters99:
This is a good step in the right direction, I feel. All my complaints leveraged at the Eris PRs is washing away now that they are starting to fit the overall style. Well done. Oh, but please don't add Eris' walls. They are *so* thin.

Arokha:
> This is a good step in the right direction, I feel. All my complaints leveraged at the Eris PRs is washing away now that they are starting to fit the overall style. Well done. Oh, but please don't add Eris' walls. They are _so_ thin.

Already did! And Bay's! And TGMC's! They're all options for mappers already, lol.

KillianKirilenko:
The tram actually looks like a tram? What sorcery is this?

Arokha:
> The tram actually looks like a tram? What sorcery is this?

ROCKET tram

PR #11342, opened 2021-08-04
Skyrat connected consoles

Arokha:
Modular and normal computers connect:
dreamseeker_zDbIYeVIsJ

Advanced (aka different sprite) version of (L-to-R) engines, helm, sensors consoles, the three you need for overmap flight:
ji22SIbzGk


Very-Soft:
ME BIG LIKE

KillianKirilenko:
Oh, fancy. Should make ship cockpits and some workspaces look a lot nicer.

The N/E/W adv sensors overlays appear to include the rest of the console though.

Might look even better if there were inner-corner angle pieces for joining consoles together... 🤔

Arokha:
> Oh, fancy. Should make ship cockpits and some workspaces look a bit nicer.
> 
> The N/E/W adv sensors overlays appear to include the rest of the console though.
> 
> Might look even better if there were inner-corner angle pieces for joining consoles together, even if they're just dummies/placeholders.

Fixed, and that sounds neat, if you wanna do!

KasparoVy:
Loving this round of graphical rework, place's looking better all the time

KillianKirilenko:
> > Might look even better if there were inner-corner angle pieces for joining consoles together, even if they're just dummies/placeholders.
> 
> Fixed, and that sounds neat, if you wanna do!

Maybe sooooon. I'll experiment with a couple different designs as well, I have some Ideas. just need to set some time aside to mash pixels together and see what comes out.

Novacat:
Thought: We should have this as a subtype instead of replacing the old one, since I can see a few places where the singleton consoles would work better.

Arokha:
> Thought: We should have this as a subtype instead of replacing the old one, since I can see a few places where the singleton consoles would work better.

That sounds pretty difficult as a subtype, since it'd be, y'know like... you'd have to make a subtype of every computer subtype.

Novacat:
A var would work!

Arokha:
> A var would work!

You do it! It would be good practice to get back into it ;p

PR #11344, opened 2021-08-05
Cortical Borers - +100 points to Steath Attributes

ResidentCody:
In the exceedingly rare instance somebody wants to play a brain sluggo, they are afforded five minutes before their presence is rudely announced with a very loud and public:

`"Unidentified lifesigns detected coming aboard the station. Secure any exterior access, including ducting and ventilation."`

This announcement is also what sounds out when there are giant spiders, and when no said spiders are found the meta-knowledge kicks in and everyone starts shouting about how people should watch out for cortical borers immediately.

This is understandably a huge mood killer and makes it super awkward to even be around. So! This PR comments out the four lines pertaining to "spawn_announcement" and consequently makes it so that the entire universe is not informed when a single borer is spawned for a single scene.

### DIFF
--- a/code/game/antagonist/alien/borer.dm
+++ b/code/game/antagonist/alien/borer.dm
@@ -20,10 +20,10 @@ var/datum/antagonist/borer/borers
 	initial_spawn_req = 3
 	initial_spawn_target = 5
 
-	spawn_announcement = "Unidentified lifesigns detected coming aboard the station. Secure any exterior access, including ducting and ventilation."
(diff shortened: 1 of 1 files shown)


Arokha (approved):
Seems fine to me

Very-Soft:
finally, the chosen one shall return

Arokha:
DELET. Let's do https://github.com/VOREStation/VOREStation/pull/11349 instead because it won't interfere with downstreams who might actually use them and want that.

PR #11347, opened 2021-08-05
Phoron Bore - Adds Hefty Recoil When Firing

ResidentCody:
Okay no it doesn't actually add physical recoil, but the phoron bore is now well-known to be a basically necessary Exploration support tool these days. Bringing at least one phoron bore for the purposes of clearing rock and stunlocking _multiple_ dangerous enemies at once is too good to pass up.

Consequently, propose adding a refire delay of two seconds to the phoron bore with no additional changes. The stun, damage, costs, all the same, but now you cannot _clickclickclickclickclickclickclickclickclick_ to hard stunlock a single enemy as oppressively, and the ability to juggle stunning multiple targets is severely hindered.

As a side effect this does somewhat nerf its ability to be used as a tool by actual Miners doing their jobs, however the two-second delay is much less of an issue in this environment and since the process looks something like "[Fire] - [Move] - [Fire] - [Move]" the majority of the delay is spent in the [Move] step. Long story short, sorry miners but the tool you can use with T4 parts to annihilate an entire mountain in record time is slightly slower at doing so.

### DIFF
--- a/code/modules/projectiles/guns/magnetic/bore.dm
+++ b/code/modules/projectiles/guns/magnetic/bore.dm
@@ -82,7 +82,7 @@
 		return "<span class='notice'>It has [mat_storage] out of [max_mat_storage] units of [ammo_material] loaded.</span>"
 	else
 		return "<span class='warning'>It\'s out of [ammo_material]!</span>"
-	
+
 
 /obj/item/weapon/gun/magnetic/matfed/attackby(var/obj/item/thing, var/mob/user)
 	if(removable_components)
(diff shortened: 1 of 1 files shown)


Very-Soft:
Being as you regularly use this I trust your judgement on reigning it in to a more balanced place.

KasparoVy:
Rock the bore & DD mining the asteroid field all the time back when I could play more regularly and see no problems w/ the delay at all, had no idea folks were using it in explo now. Sick

PR #11348, opened 2021-08-05
Port/add throwing subsystem

### LINKED ISSUE #10742: Overmap Meteor event causes tons of lag with the talon (or maybe just in general?)
#### Description of the bug

So, the talon is actually quite resistant to meteors what with all the point defense turrets it has, if it's set up, it can quite happily slowly fly through a meteor field without a crazy amount of risk, especially if you turn the shields on.

But, I have found that if you do that, the whole server gets CRAAAAZY laggy until it's out of the field, while the ship itself might take some minimal damage, if it takes any damage at all. (So, it's not like it's processing atmos changes or explosions or anything)

I was doing this because I noticed that when the meteors get destroyed, they drop some ore, so. Was attempting to mine using a spaceship, since, it at least seems somewhat intentional for that to be a thing you can do.

Not ruin the round for everyone with lag while seeking space profit

#### What actually happened

Everyone had a bad time

#### Steps to reproduce
- (Step 1) Set up the talon
- (Step 2) Fly the talon into a meteor field

#### Code Revision
Commit: c901b017ef9faaa5fab34126b088d695295f8715 

 

I should note there are kind of 3 moving parts here, and I think it could be any or all of them collectively causing this:

-The meteors spawn and get thrown in a direction
-The turrets shoot beams at the meteors, destroying them
-When destroyed, the meteors spawn a bunch of ore which continues on the meteors' original trajectory

I have noticed that the server seems to hitch for a while sometimes while the turrets' beams are firing, but, really I am not sure what part of this is causing the whole server to slow to a crawl, but it's happening, and I think it might be a source of some of our mysterious random server chuggings that happen sometimes

I haven't tested it further yet, so I'm not sure if this is a problem that just the talon has, or if it just always happens. (people usually avoid flying into meteors, and when they do fly into them, the shuttle usually blows up, so it seems normal that the server might lag.)

### LINKED ISSUE #11268: Flying the ITV Talon through asteroid fields lags the server
The server tends to lag when the Talon flies through an asteroid field.

To 'mine' asteroids using the point defense/shields.

#### What actually happened
Everything works, it just lags the server to an absurd amount until the Talon leaves the asteroid field.

#### Steps to reproduce
- Bring Talon shields up
- Enter an asteroid field
- Watch in awe as the server stutters to a crawl
- Leave the asteroid field and everything returns to normal

Arokha:
Port Bay's version which is an older version of /tg/'s, because /tg/'s current is a bit newer with features that we don't particularly have, and Bay's is literally a drop-in replacement. Copy-pasted and compiled and works fine.

Fixes the doom meteor lag from going through meteors on ships and stuff.

Fixes #11268
Fixes #10742

### DIFF
--- a/code/controllers/subsystems/throwing.dm
+++ b/code/controllers/subsystems/throwing.dm
@@ -1,0 +1,208 @@
+#define MAX_THROWING_DIST 1280 // 5 z-levels on default width
+#define MAX_TICKS_TO_MAKE_UP 3 //how many missed ticks will we attempt to make up for this run.
+
+SUBSYSTEM_DEF(throwing)
+	name = "Throwing"
+	wait = 1
+	flags = SS_NO_INIT|SS_KEEP_TIMING
+
+	var/list/currentrun
+	var/list/processing = list()
+
+/datum/controller/subsystem/throwing/stat_entry()
+	..("P:[processing.len]")
+
+/datum/controller/subsystem/throwing/fire(resumed = 0)
+	if (!resumed)
+		src.currentrun = processing.Copy()
+
+	//cache for sanic speed (lists are references anyways)
+	var/list/currentrun = src.currentrun
+
+	while(length(currentrun))
+		var/atom/movable/AM = currentrun[currentrun.len]
+		var/datum/thrownthing/TT = currentrun[AM]
+		currentrun.len--
+		if (QDELETED(AM) || QDELETED(TT))
+			processing -= AM
+			if (MC_TICK_CHECK)
+				return
+			continue
+
+		TT.tick()
+
+		if (MC_TICK_CHECK)
+			return
+
+	currentrun = null
+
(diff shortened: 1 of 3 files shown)


code/controllers/subsystems/throwing.dm
```diff
@@ -0,0 +1,208 @@
...
+	var/dy
+	var/diagonal_error
+	var/pure_diagonal
+	var/datum/callback/callback
+	var/paused = FALSE
+	var/delayed_time = 0
+	var/last_move = 0
+
+/datum/thrownthing/New(var/atom/movable/thrownthing, var/atom/target, var/range, var/speed, var/mob/thrower, var/datum/callback/callback)
+	..()
```
Leshana, line 62: We have a policy that /datum/New will never exist, so can safely remove.

Leshana (approved):

PR #11349, opened 2021-08-05
Make a borer subtype for adminbuse

Arokha:
The /non_antag version doesn't let you spawn more borers, doesn't ask for a ghost volunteer, and doesn't do antag role stuff.


KasparoVy:
Good thinking, courteous!

PR #11360, opened 2021-08-07
Fancy exploration shuttle

Arokha:
dreamseeker_2FQJrgHvvG


TheToaster98:
It looks pretty cool, but could it be possible to add some handrails?

KillianKirilenko:
Yeah, it could do with getting a couple of handrails back. I'd at least place a pair on the W/E sides of the rear hatch, and a couple by the sleeper.

Where's the shuttle sensor array now though?

Not sure I'm sold on the overall layout changes too. The cockpit was moved to the back to reduce the risk of asteroids trashing the consoles and making the shuttle a pain to recover/repair. It's either going to need point defense turrets or shields if you're really going to move stuff back up to the front. There's also quite a bit of dead/wasted space, and small vessels should be as space-efficient as possible.

ShudderEve:
The spritework for the exterior is great, but I have some mapping suggestions:

- There is just a pilot seat, nothing for the co-pilot, the current shuttle has two seats, and room for one more in the cockpit,
- There are just walls between the pilot and the rest of the ship, most pilots I see will be talking or planning with explorers, so consider adding some glass anywhere about here:

image

EDIT:

Just realized, but I think this is another ERIS port? Here's the ERIS mining shuttle sprite:

image
and the rear:
image

Maybe we should adjust it so that it fits the server's sprite style better, as of now, it kind of stands out as an obvious port from a different spritebase.

Arokha:
> * There is just a pilot seat, nothing for the co-pilot, the current shuttle has two seats, and room for one more in the cockpit,
> * There are just walls between the pilot and the rest of the ship, most pilots I see will be talking or planning with explorers, so consider adding some glass anywhere about here:

There's generally only one pilot, and it's physically impossible to have two seats that can reach three consoles unless the consoles are in the middle with seats on both sides (that's just the nature of geometry, nothing to be done about that). Having piloted the ship before, it was annoying to have to move between the left and right sides of the cockpit, so I personally find this easier to fly. I'll add a jumpseat though, since no reason not to!

> Just realized, but I think this is another ERIS port?

In a way. GhostActual took the shuttle and modified it heavily. Their mining shuttle has a different floorplan and is much larger.

> Maybe we should adjust it so that it fits the server's sprite style better, as of now, it kind of stands out as an obvious port from a different spritebase.

All the shuttles are being replaced, so they will match.

Arokha:
> Yeah, it could do with getting a couple of handrails back. I'd at least place a pair on the W/E sides of the rear hatch, and a couple by the sleeper.

Ok, I added 6 more handrails, two at the back and 4 in the general crew area

> Where's the shuttle sensor array now though?

See image below, it's outside on the topright. It's invisible for now (along with the engines) because there's not really any interactivity to be done with them. However, if it comes to pass that you can repair engines and sensors and stuff, the plan would be to make the sensors have a sprite sort of like a forward-facing antenna array below the turf plane, so it looks like it's attached to the bottom of the ship, and the engines can assume the large ship sprite's engine sprites.
 
> Not sure I'm sold on the overall layout changes too. The cockpit was moved to the back to reduce the risk of asteroids trashing the consoles and making the shuttle a pain to recover/repair. It's either going to need point defense turrets or shields if you're really going to move stuff back up to the front. There's also quite a bit of dead/wasted space, and small vessels should be as space-efficient as possible.

The front is actually two layers of walls and I think you'll find it more resilient than the windows of shuttles past. The walls are just marked as transparent so you can see through them. They will resist meteors better than windows. Additionally, having recently watched the explo shuttle get trashed despite cockpit movement by other means (side meteors, since when stationary they are split 25/25/25/25% between each cardinal side, and 90%/10% preferring the front when moving, and the destruction of APCs and the SMES and other things nearer the front), I'm not convinced it actually helps much. Besides, as I said, this should be more resilient now. Also staff requested that it still be destroyable, just not with good flying, so.

Now that you mention it, I'm going to replace the walls with stronger hull materials.

Handrails and sensors:
image

Engines, FYI:
image

SplinterGP:
since killian mentioned getting hit by asteroids and all, how does the walls look when some of them are broken ?

Arokha:
> since killian mentioned getting hit by asteroids and all, how does the walls look when some of them are broken ?

They break like other walls, but don't leave girders, so you end up with plating where there used to be walls.

Sometimes it looks a little strange, because bits that don't have a lot of ... well, pixel volume in them turn into a full 32x32 turf on breaking, but honestly since that benefits people trying to fix the shuttle by giving them more walking/safe area to work with, I'm inclined to leave it.

I had a brief period where I had it so that only walls adjacent to normal floors left plating, so you'd end up with a single layer of plating wrapped around the floors, assuming you destroyed every wall, which was fine too, but eh. Decided to undo that.

Arokha:
Gonna merge, since I fixed the specific criticisms about handrails and added another copilot seat. Other mappers can tweak further if they so wish, or can ask nicely somewhere!

PR #11243, opened 2021-07-25
Switches smartfridges around in hydroponics

Huenererschrecker:
- Deletes the extra smartfridge in the hydroponics backroom that noone ever uses.
- Changes the smartfridge between hydroponics and the kitchen into a produce persistent one (previously used exclusively in the never used backroom fridge)
(I hope I did this right and haven't broken something)

PR #11362, opened 2021-08-07
Smartfridge Fix

GhostActual:
Fixes the Hydroponics fridge issue that players were having problems with.

Also adds another airlock to the south of the surface-3 hangar because I felt that it didn't quite make sense for a hangar to only have one airlock and the rest be open to the elements.

### DIFF
--- a/maps/tether/tether-03-surface3.dmm
+++ b/maps/tether/tether-03-surface3.dmm
@@ -18345,7 +18345,9 @@
 /obj/effect/floor_decal/corner/lime/border{
 	dir = 1
 	},
-/obj/machinery/smartfridge/produce/persistent_lossy,
+/obj/structure/table/standard{
+	name = "plastic table frame"
+	},
 /turf/simulated/floor/tiled,
 /area/tether/surfacebase/botanystorage)
 "aDY" = (
@@ -21015,6 +21017,9 @@
 /area/crew_quarters/kitchen)
 "aIl" = (
 /obj/item/weapon/stool/padded,
+/obj/effect/floor_decal/spline/plain{
+	dir = 1
+	},
 /turf/simulated/floor/carpet/turcarpet,
 /area/crew_quarters/bar)
 "aIm" = (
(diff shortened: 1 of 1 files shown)


GhostActual:
https://github.com/VOREStation/VOREStation/pull/11243 This PR fixes this issue

Arokha:
So they will have two produce smartfridges then?
image

GhostActual:
> So they will have two persistent produce smartfridges then? I think that's not desired and would require some balancey talks maybe. Not sure though.
> 
> image

That's not the map I changed, I also added a second chicken, you're looking at an older file it seems.

Very-Soft:
> > So they will have two persistent produce smartfridges then? I think that's not desired and would require some balancey talks maybe. Not sure though.
> > image
> 
> That's not the map I changed, I also added a second chicken, you're looking at an older file it seems.

it's a small thing, but they don't need another chicken, they can get more chickens through eggs or through cargo. Pls no chicken powercreep

maps/tether/tether-03-surface3.dmm
```diff
@@ -29854,7 +29844,7 @@
 "aZg" = (
 /obj/effect/floor_decal/corner/grey/diagonal,
 /obj/machinery/door/firedoor/glass,
-/obj/machinery/smartfridge/produce,
+/obj/machinery/smartfridge/produce/persistent_lossy,
```
Arokha, line 29847: Yes, this is the map you changed. What are you talking about???

Arokha (changes requested):

Huenererschrecker:
Since I've done the original pr and all that, the original idea was to just move the persistency fridge into the position between botany and the kitchen and delete the one in the backroom. The reason being that the one in the backroom is never used anyways. The plan was never to be 2 persistency fridges.

Arokha:
> Since I've done the original pr and all that, the original idea was to just move the persistency fridge into the position between botany and the kitchen and delete the one in the backroom. The reason being that the one in the backroom is never used anyways. The plan was never to be 2 persistency fridges.

I'll note that moving them erases their contents, just FYI.

PR #11382, opened 2021-08-10
Add default zorren tail to character setup menu

barrabobabo:
The "default" Zorren tail is not available for selection in the character setup menu, despite its potential use for various characters that may not like the other vulpine/canine tails available. This seems to fix that!


barrabobabo:
oof, something stupid with the edit tracker. Going to re-open this.

PR #11383, opened 2021-08-10
Add default zorren tail to character setup menu 

barrabobabo:
The "default" Zorren tail is not available for selection in the character setup menu, despite its potential use for various characters whose players may not like the other vulpine/canine tails available. This commit fixes that.

### DIFF
--- a/code/modules/mob/new_player/sprite_accessories_tail_vr.dm
+++ b/code/modules/mob/new_player/sprite_accessories_tail_vr.dm
@@ -634,6 +634,14 @@
 	icon_state = "nevreantail_hc_s"
 	do_colouration = 1
 	color_blend_mode = ICON_MULTIPLY
+	
+/datum/sprite_accessory/tail/special/foxdefault
+	name = "default zorren tail, colorable"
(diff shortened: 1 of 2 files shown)


Arokha:
Er, no need to define the same thing twice

code/modules/mob/new_player/sprite_accessories_tail_vr.dm
```diff
@@ -634,6 +634,14 @@
...
 	do_colouration = 1
 	color_blend_mode = ICON_MULTIPLY
+	
+/datum/sprite_accessory/tail/special/foxdefault
+	name = "default zorren tail, colorable"
+	desc = ""
+	icon = "icons/mob/human_races/r_fox_vr.dmi"
+	icon_state = "tail_s"
+	do_colouration = 1
+	color_blend_mode = ICON_MULTIPLY
```
Arokha, line 644:
```suggestion
```

KasparoVy:
Glad to have this tail back, was missing from one of my characters after it got accidentally removed some time ago

PR #11387, opened 2021-08-10
Fixes infinite material duplication in R&D

Heroman3003:


code/modules/research/rdconsole_tgui.dm
```diff
@@ -423,6 +423,14 @@
...
 						return
 
+					if(istype(linked_destroy.loaded_item,/obj/item/stack))//Only deconsturcts one sheet at a time instead of the entire stack
+						var/obj/item/stack/ST = linked_destroy.loaded_item
+						if(ST.get_amount() < 1)
+							playsound(linked_destroy, 'sound/machines/destructive_analyzer.ogg', 50, 1)
+							qdel(ST)
+							linked_destroy.icon_state = "d_analyzer"
+							return TRUE
+
```
Arokha, line 433: Why can the DA take sheets at all?

code/game/objects/items/stacks/stack.dm
```diff
@@ -354,6 +354,7 @@
 /obj/item/stack/attack_hand(mob/user as mob)
 	if (user.get_inactive_hand() == src)
 		var/N = input(usr, "How many stacks of [src] would you like to split off?  There are currently [amount].", "Split stacks", 1) as num|null
+		N = round(N)
 		if(N)
 			var/obj/item/stack/F = src.split(N)
 			if (F)
```
Arokha, line 360: You should fix this in split() instead.

Arokha (changes requested):

code/modules/research/rdconsole_tgui.dm
Heroman3003: Because many sheets carry science values. Mostly related to 'materials' research, but also some others, like medical stacks or extra advanced/cool materials.

Arokha: I see. Well, your code here doesn't appear to ever reduce the amount in the stack.

Heroman3003: Nope, the code for that exists already and happens later in same proc. This is basically a contingency check for "do we even have enough of a stack for it to contribute?"

code/game/objects/items/stacks/stack.dm
Heroman3003: Done

LordSaladin:
So you're saying we need miners and drill techs to be present for the duration of every shift now?.
_cries in downstream_

Arokha:
> So you're saying we need miners and drill techs to be present for the duration of every shift now?.
> _cries in downstream_

Just code your downstream different?? Have the persistent sheet storage start with some sheets or something lol

LordSaladin:
> Just code your downstream different?? Have the persistent sheet storage start with some sheets or something lol

It's not the base materials I'm worried about it's the rare ones

LordSaladin:
> 
> > Just code your downstream different?? Have the persistent sheet storage start with some sheets or something lol
> 
> It's not the base materials I'm worried about it's the rare ones

What I'm trying to say is that the duplication was helpful on deadpop shifts

PR #11391, opened 2021-08-11
Orange Line Upgrade

GhostActual:
A more finalized and fitting version of the Orange Line tram.

orangeline-n

cctram


GhostActual:
Might conflict with other PRs - will fix as needed

Zeldazackman:
Some hopefully constructive criticism that I've got on this, 

It seems to have a limited amount of seating for folks (that aren't currently residing within a gut) and looks to be very geometrically focused for what it's worth (more windows might make it feel more open).
There seems to be good margin of open space reserved explicitly for showcasing the nanotrasen logos upon the floor to either side of middle seating that feels like it could be better utilized with some maneuvering for more seating space and or decorative additions.

However saying that, I do like the idea of sofas/benches right next to the accessible spots located on either side as it's quite tasteful on aesthetics, and ontop of that I do like the focus on benches rather than shuttle chairs as it's a nice change to see.

GhostActual:
> Some hopefully constructive criticism that I've got on this,
> 
> It seems to have a limited amount of seating for folks (that aren't currently residing within a gut) and looks to be very geometrically focused for what it's worth (more windows might make it feel more open).
> There seems to be good margin of open space reserved explicitly for showcasing the nanotrasen logos upon the floor to either side of middle seating that feels like it could be better utilized with some maneuvering for more seating space and or decorative additions (A wall mounted fire closet might be nice too for a "just incase sam burns the entire planet, again" event).
> 
> However saying that, I do like the idea of sofas/benches right next to the accessible spots located on either side as it's quite tasteful on aesthetics, and ontop of that I do like the focus on benches rather than shuttle chairs as it's a nice change to see.

To address these concerns:
-I can expand this both to the north and to the south to add additional space to the tram car if need be. This will be dependant on testing live, however, as most of the time less than twenty people escape on the tram.
-The extra space over the NT logo is walking space to get to the middle of the tram. on both the North and South sides of the tram car, there is a 3x7 gap. It technically can be crammed to 2x7, however, the only explansion that would be able to do that is the addition of tables and chairs to the far North and far South of the tram car. In my own testing, I found this to be quite cramped and uncomfortable.
-There are three windows in total - North, West, and East - those are really the only places that could fit windows without breaking the sculpt of the tram car. I can't add one to the south due to the instrumentation that makes the tram car go.
-There are two (2) wall-mounted fire closets on the opposing sides of the monitors mounted to the support wall to the North and South of the tram car.

Important to note:
The only way to expand the tram is to expand it North or South without expanding the tram stations as they are now. I will look into previous trams to determine an alternate route

PR #11395, opened 2021-08-12
Increase the round restart timer from 3 to 4

lbnesquik:
## About this pull request 
This simply adds another extra minute to the restart timer, like on polaris.

I often feel like the end of the shift can be a rushed affair. A little extra minute would be nice to exchange a few words before DC'ing, and so on. It also gives you a bit more time to take a pause if you'd like to queue for the next roundstart. It isn't a dramatic increase in time either for people who just want to get on with it either.


Arokha (approved):

Arokha:
Seems fine from code perspective. I'll leave it open a couple days in case anyone wants to comment from a gameplay perspective.

lm40:
end of the round is long enough already with tram call, tram waiting, and tram travel, you have tons of time to start rounding things off already

Very-Soft:
Seems kind of silly to me. Like, ultimately it changes so little that it doesn't matter, so, fine, but like, 1 extra minute at the end of the round is so little time that I'm not actually really sure what the point is.

Of course, if it added a lot of time to the end of the round I'd probably be resistant to this, so!

PR #11398, opened 2021-08-13
Przy creations port from Chompers

Arokha:
Przy/Serdy was like 'psst look at my hotties' and I was like :0 and he was like 'you COULD port them' and I was like OK

- Replaces ERT borg sprite he made in 2015 with a nicer one he made in 2021! Removes it's baton because it has ZAPJAWS now. Also a unique sleeper.
- Leopardmander! Normal glassbelly and bloo.
- Dragons! Boss mobs that let you pick sprite accessories for them and have telegraphed attacks sometimes.
- Big waffs! They're big and they're wolfeh.

To Chompers: I didn't change any of the functionality of these except for a commented-out `belly_fullscreen_color` in bigdragon.dm because we don't have that feature and I'm too lazy to port it. So! You should just be able to delete your _ch files. I also ported autotransfer so you can remove that from the belly obj file you have.

dreamseeker_t4o1BSMDJL


User-U-U:
Why not keep both ERT borgs?

GhostActual:
> Why not keep both ERT borgs? Not everyone wants to be a drake, I for one like the hound one.

I agree with this person, keep both and leave it up to the choice of the player

Rykka-Stormheart:
Chiming in, keeping the old ERT design as an alternative would be amazing. It's still a cute as heck borko sprite, and canines, 'nuff said.

KasparoVy:
Holy shiiit what a dope lot of creatures. Great spritework from the artist

Arokha:
> Why not keep both ERT borgs? Not everyone wants to be a drake, I for one like the hound one.

> I agree with this person, keep both and leave it up to the choice of the player

> Chiming in, keeping the old ERT design as an alternative would be amazing. It's still a cute as heck borko sprite, and canines, 'nuff said.

Ok, added a "Classic" choice

Rykka-Stormheart:
> > Why not keep both ERT borgs? Not everyone wants to be a drake, I for one like the hound one.
> 
> > I agree with this person, keep both and leave it up to the choice of the player
> 
> > Chiming in, keeping the old ERT design as an alternative would be amazing. It's still a cute as heck borko sprite, and canines, 'nuff said.
> 
> Ok, added a "Classic" choice

TY! <3

Rykka-Stormheart:
Oh! Is it feasible/easy to make the old ERT borg use the old icons (for blade, etc)? If not, dw about it. <3

Arokha:
> Oh! Is it feasible/easy to make the old ERT borg use the old icon (for blade specifically as that's the only model-specific sprite)? If not, dw about it. <3

Not really, 'm afraid, not without making the classic one an entirely separate module from ERT so it gets a different equipment for that, that has a different sprite. I mean you could do some hacky stuff maybe but enh. These don't show up super often anyway!

Rykka-Stormheart:
> > Oh! Is it feasible/easy to make the old ERT borg use the old icon (for blade specifically as that's the only model-specific sprite)? If not, dw about it. <3
> 
> Not really, 'm afraid, not without making the classic one an entirely separate module from ERT so it gets a different equipment for that, that has a different sprite. I mean you could do some hacky stuff maybe but enh. These don't show up super often anyway!

yeye! Totally understandable, it's why I asked if it was feasible, no worries since it's not + they're rare, the sprites for just the bork was plenty good. Thank you c:

PR #11408, opened 2021-08-13
[MIRROR] Stun Revolver Update

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/8178
--------------------
Adds a brand new Stun Revolver sprite that is modernized while keeping a ported version of the original that has been heavily modified. Both have full animation and should work as intended. The third one is a recent addition, a snub nose 'gangster' model that doesn't spawn normally and is mainly for adminbus.

New Sprites:

snub_stun

PR #11426, opened 2021-08-15
Stun Revolver (Virgo Edition)

GhostActual:
Adds a brand new Stun Revolver sprite that is modernized while keeping a ported version of the original that has been heavily modified. Both have full animation and should work as intended. The third one is a recent addition, a snub nose 'gangster' model that doesn't spawn normally and is mainly for adminbus.

New Sprites:
stun_rev

vintage_stun

snub_stun

NOTICE:
There was a recent merge (https://github.com/VOREStation/VOREStation/pull/11408) that was an upstream PR I created a while back. The sprites I created were meant to match Polaris, I had set aside a set for Virgo but didn't manage to catch the upstream merge before it was implemented.


Very-Soft:
I like the sprites, they actually look like revolvers which is cool

what's with the vending machine removals though?

GhostActual:
> I like the sprites, they actually look like revolvers which is cool
> 
> what's with the vending machine removals though?

It was all duplicated code I discovered while testing. It's still live, if you go to SecTech and look at the options you'll see two separate options to vend Stun Revolvers. I combed through the file and found more duplicated code so I removed it all.

R3dtail:
Question /suggestion would it be annoying to add a spinnable chamber like other revolvers have?  If only for flavor reasons of course.

GhostActual:
> Question /suggestion would it be annoying to add a spinnable chamber like other revolvers have? If only for flavor reasons of course.

I don't think so, the code files for Stun Revolvers and Revolvers are different. The most I could probably do is make it play a sound and do a chat message. I will look into it.

Arokha:
> > Question /suggestion would it be annoying to add a spinnable chamber like other revolvers have? If only for flavor reasons of course.
> 
> I don't think so, the code files for Stun Revolvers and Revolvers are different. The most I could probably do is make it play a sound and do a chat message. I will look into it.

If you do, do a separate PR since it's probably not worth creeping this sprite 'fix' into adding new features in the same PR

PR #11433, opened 2021-08-16
Removes emergency markings from surface shuttle

Very-Soft:
It's for going up and down, not for escaping!


GhostActual:
They never told me this!

PR #11438, opened 2021-08-17
Adds *dwoop to all synths. 

Torque4607:
Rejoice, all synthetics who were envious of station bounds getting to chirp happily. Now all silicons can partake in *dwoop. Just one line was missing from the human/emote.dm to make this work.


KasparoVy:
Woooo synth rights!

PR #11439, opened 2021-08-17
Ports RD labcoat from CitRP

PastelPrinceDan:
because a friend has been asking for it a while
- is in loadout
- also spawns in rd locker


Arokha:
Looks nice

PR #11442, opened 2021-08-18
Fulpstation Briefcases

GhostActual:
Adds Fulpstation Briefcases (modified to match the code) as well as changes some bags around to match the art style of the new bags that were ported by Aronai. As a bonus, I also ported the Fountain Pens from Fulpstation as well to complement the Briefcases.

This was made while a Frenchman whispered in my ear, I was not in the correct state of mind.


Very-Soft:
Seems cool, I like the book bag~

Arokha (changes requested):
I like these sprites more, but you deleted the "gunbox" state from storage.dmi, which is used by /obj/item/gunbox

Either change /obj/item/gunbox to use a different state or add a new/previous gunbox state.

Arokha (approved):

PR #11450, opened 2021-08-19
Catslugs and Simplemob QoL

Very-Soft:
Adds Catslug mob!
-Can pick up and use/throw many things.
-Can ventcrawl, can carry various materials and objects into the vents
-Speaks in sign
-Can eat and be fed foods, which allows them to naturally heal from injuries.

Also makes it so:
-All simplemobs understand galactic common without being able to speek it regardless of other settings or languages unless otherwise disabled with the understands_common var, for facilitating RP mostly, since there's basically no situation I can think of where I would prefer that a player controlled mob couldn't at least understand common.
-Simplemobs with hands can pick up and handle micros like human mobs can.
-Simplemob grabs will now disappear appropriately when dropped rather than waiting for when they feel like it.


JoanRisu:
Cute species kills man

Personally, it gets my vote.

code/modules/mob/mob_grab.dm
```diff
@@ -34,6 +34,7 @@
 	abstract = 1
 	item_state = "nothing"
 	w_class = ITEMSIZE_HUGE
+	destroy_on_drop = TRUE	//VOREStation Edit
```
Novacat, line 37: ...why was this never done before I wonder?

Arokha: It's not strictly necessary since it has it's own dropped() proc that deletes it, but it does make sense to have it set anyway.

PR #11454, opened 2021-08-20
Make stack amount var private

Arokha:
Am tired of negative, fractional, crazy sheet amounts. Just make the var private. No touchy.

### DIFF
--- a/code/game/objects/items/stacks/stack.dm
+++ b/code/game/objects/items/stacks/stack.dm
@@ -17,7 +17,7 @@
 	center_of_mass = null
 	var/list/datum/stack_recipe/recipes
 	var/singular_name
-	var/amount = 1
+	VAR_PROTECTED/amount = 1
 	var/max_amount //also see stack recipes initialisation, param "max_res_amount" must be equal to this max_amount
 	var/stacktype //determines whether different stack types can merge
 	var/build_type = null //used when directly applied to a turf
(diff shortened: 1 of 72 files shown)


code/game/objects/items/stacks/stack.dm
```diff
@@ -265,6 +273,27 @@
 			var/datum/matter_synth/S = synths[i]
 			S.add_charge(charge_costs[i] * extra)
 
+/obj/item/stack/proc/set_amount(var/new_amount, var/no_limits = FALSE)
+	if(new_amount < 0 || new_amount % 1)
+		stack_trace("Tried to set a bad stack amount: [new_amount]")
+		return 0
+	
+	// Clean up the new amount
+	new_amount = max(round(new_amount), 0)
```
Heroman3003, line 282: Why do this when we just exit out if its either negative or decimal anyway?

Arokha: I don't necessarily want to leave the stack_trace forever. It doesn't reaaaally need to be done though.

Novacat:
How will this affect admin varediting?

Arokha:
> How will this affect admin varediting? Amount is also edited a few times on map (The expedition shuttle's tritium for its backup generator)

It wouldn't affect either. Admins can still set it to arbitrary values. Unless you want me to prevent that?

Novacat:
I would say no need! Just wanted to make sure admins can still edit it!

dwinters99:
I'm unclear what this actually changes. Does this make it impossible to see the amount of items in a stack? Or is this under-the-hood stuff that won't effect a user? I would imagine it's the second, because I CAN'T imagine the first being someone anyone in their right mind thinking up.

Heroman3003:
> 
> 
> I'm unclear what this actually changes. Does this make it impossible to see the amount of items in a stack? Or is this under-the-hood stuff that won't effect a user? I would imagine it's the second, because I CAN'T imagine the first being someone anyone in their right mind thinking up.

This is basically sweep over all the places where stacks are made/split/etc to prevent potential of negative or decimal amounts in stacks.

PR #11461, opened 2021-08-20
Can now give custom species names during vorebelly transformations

### LINKED ISSUE #11406: Cannot give custom species names during vorebelly transformations
Not a bug but an oversight. There is no way to change a custom species names during vorebelly transformation.

Some window to pop up where I can enter what race to display as the custom species name

#### What actually happened
Species: Custom Species

#### Steps to reproduce
- Acquire prey
- Nom Nom Nom into a belly
- In the vore tab, Contents -> click the prey -> select Transform
- Click race, select custom species
- Notice how you cannot change the name of Custom Species

#### Code Revision
- 7ee7e8d0b0b87bf5c1aa5b9cc202bbf0c80e6e0c

KeenShayne:
Prompts you for a custom race name when transforming people in a vorebelly.
Fixes #11406
CustomRaceFix

### DIFF
--- a/code/modules/tgui/modules/appearance_changer.dm
+++ b/code/modules/tgui/modules/appearance_changer.dm
@@ -92,7 +92,8 @@
 			if(can_change(APPEARANCE_RACE) && (params["race"] in valid_species))
 				if(target.change_species(params["race"]))
 					if(params["race"] == "Custom Species")
-						target.custom_species = input(usr,"Custom Species name:","Species Name")
+						target.custom_species = sanitize(input(usr, "Input custom species name:",
+							"Custom Species Name") as null|text, MAX_NAME_LEN)
 					cut_data()
 					generate_data(usr)
 					changed_hook(APPEARANCECHANGER_CHANGED_RACE)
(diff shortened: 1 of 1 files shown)


KeenShayne:
Seems I need to access custom_species a different way.

KeenShayne:
> Seems I need to access custom_species a different way.
> Back to the drawing board!

I think this is reason enough to take a break for now.

code/modules/tgui/modules/appearance_changer.dm
```diff
@@ -91,6 +91,8 @@
 		if("race")
 			if(can_change(APPEARANCE_RACE) && (params["race"] in valid_species))
 				if(target.change_species(params["race"]))
+					if(params["race"] == "Custom Species")
+						target.custom_species = input(usr,"Custom Species name:","Species Name")
```
Arokha, line 95: This needs sanitization. As it is, you can put crazy things in here, like html img tags or anything. Run it through sanitize, and allow users to cancel if they change their mind as well.

Arokha (changes requested):

code/modules/tgui/modules/appearance_changer.dm
KeenShayne: I added the cancel option, but in this case there's no good way to cancel selecting the race, thus if you do hit cancel it simple defaults the custom race to "Custom Race".

I'm probably going to do a rework of this panel at some point to make it easier to use.

Arokha (approved):

PR #11462, opened 2021-08-20
Crusher Gaming

Hatterhat:
image
image
adds proto-kinetic machetes, daggers, and gauntlets. the crusher itself is unused because twohand code Sucks.

## HOW TO GET:
the dagger and machete are buyable from both mining and exploration vendors. points are placeholder values
however, the gauntlets are mining exclusive

## GIB NUMBERS PL0X
current damage numbers for the -
MACHETE (for running into things to stab them):
- cleaves
- sharp/edged
- 24 force
- 26 mark detonation damage (50 damage)
- 40 backstab damage (90 damage, but you have to hit the mob from behind)
- 20 thrown damage (110 damage on a backstab throw. why would you go for a backstab throw)

DAGGER (the backup plan):
- does not cleave
- sharp/edged
- 15 force
- 25 detonation damage (40 damage)
- 35 backstab damage (75 damage, see above backstab caveat)
- 35 thrown damage (105 damage on a backstab throw, but see above caveat)

GAUNTLETS (hardsuit required, for single-target damage):
- does not cleave
- not sharp nor edged
- 30 force
- 35 detonation damage (75 damage)
- 55 backstab bonus (120 damage! wow!)

original implementation on polariscode derivatives from Citadel-Station-13/Citadel-Station-13-RP#3015


Arokha (changes requested):
I like this but it seems to duplicate some stuff. Like the whole status effect system is cool and stuff but appears to just be another implementation of modifiers. Just write it using modifiers (or if you really think status effects are better, convert modifiers to status effects).

You shouldn't implement update_icon_state() and update_overlays() as stubs, just put the code in update_icons() like all the other ports from /tg/ stuff.

Basically: Don't add duplicates of existing systems for just this item unless you're going to replace those systems with these.

KasparoVy:
These weapons throw runtimes if you attempt to reblade them with a whetstone since they don't got no material/hardness property but it's inconsequential gameplay wise from what I've found in testing

KasparoVy:
Additionally, when firing you get a debug overlay: 
image

Hatterhat:
ah fuck

Hatterhat:
also they're. not supposed to be whetstoned. they arent material weapons. theyre more for "i want to mark things, and then hit them" instead of "i enjoy mashing LMB"

KasparoVy:
Cool! Runtimes bad

PR #11468, opened 2021-08-22
tree's haunted

Hatterhat:
port of vgstation-coders/vgstation13#28377. contains a haphazard find-replace of New() to Initialize() in the flora .dm.

however, this has some issues with big trees, because those use a Completely Unrelated Overlay to form the upper half. i don't know what i wanna do with that

https://user-images.githubusercontent.com/31829017/130339225-a7489d23-d09f-4cdb-8ec1-df7995976098.mp4

### DIFF
--- a/code/game/objects/structures/flora/flora.dm
+++ b/code/game/objects/structures/flora/flora.dm
@@ -21,6 +21,7 @@
 	var/max_harvests = 0
 	var/min_harvests = -1
 	var/list/harvest_loot = null	// Should be an associative list for things to spawn, and their weights. An example would be a branch from a tree.
+	var/image/transparent
 
 /obj/structure/flora/Initialize()
 	. = ..()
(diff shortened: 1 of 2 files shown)


Very-Soft:
I'd personally prefer if it only revealed what was behind the tree if you were directly adjacent to it. Part of the fun of the trees is being able to hide behind them, so you should have to get REAL CLOSE to check behind them.

Hatterhat:
yeah ok i guess i can do that

Arokha:
In re: Soft's comment, if you want to do that, you can do crossed() and uncrossed(). Only if your non-transparent pixels overlap the tree's non-transparent pixels will it call those, I believe. So you can do that instead of the signals

Very-Soft:
Feel free to undraft this when it's updated as requested.

Hatterhat:
man i genuinely forgot i had this pr up

PR #11469, opened 2021-08-22
Port some more clothes from CitRP + updates 'off' sprite for mesons & co to match new style.

PastelPrinceDan:
- Ports a buncha clothes from CitRP that I missed dearly.
Includes: command dress jacket, verglas dress, fashionable miniskirt (colorable), cowbell collar, cow bikini, paramedic uniforms

- Also updated the 'degoggles' sprite to match the new mesons/thermals/etc, because it reverted to the old sprite when off, which was kind of annoying. I recolored the mesons in three seconds so if anyone wants to make it better at some point, they should.


Arokha:
Er, do we not already have that paramedic uniform center-right?

Arokha:
The dreamchecker fail is due to:
image

PastelPrinceDan:
> Er, do we not already have that paramedic uniform center-right?

I don't believe we do? Looking at uniforms/items or mob.dmi, I'm not seeing it anywhere. I can't recall seeing anyone using it in game either, ever.

I'll see to fixing what failed that check, though.

GhostActual:
> Er, do we not already have that paramedic uniform center-right?

You have a similar colored jacket which looks identical, however the para uniform you have is this.

Screenshot 2021-08-22 020117
Screenshot 2021-08-22 020139

PR #11478, opened 2021-08-23
[MIRROR] Added tabloid magazines.

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/8228
--------------------
Adds trashy tabloid magazines with some headlines adapted from Europa.

Most of the mapchanges are SDMM pruning var setting that are the same as the object's compiled value.


GhostActual:
Merge this cause I was going to do something similar and to have a baseline would be lovely

PR #11485, opened 2021-08-23
Removes duplicate files related to taurs

lbnesquik:
## About this pull request
This PR removes all the taursuit files which do not have _vr.
None of them had anything unique about them, and sometimes lacked some files as well.

It also deletes a duplicate .dm file. Both it and the _vr version shared some variables and defines for some reason, but it isn't like this was a polaris file, so why have the seperation?

This clears up the folders a bit for us, making it easier to deal with clothing sprites in general.


Arokha (changes requested):
> but it isn't like this was a polaris file

https://github.com/PolarisSS13/Polaris/blob/master/code/modules/mob/new_player/sprite_accessories_taur.dm

lbnesquik:
Huh. Color me surprised.

PR #11488, opened 2021-08-24
New Flashlights (from scratch)

GhostActual:
Made some flashlights to practice diagonal sprite making and it spiraled from there. All are based on existing flashlights, however, all are from scratch. I even did the Chemlights/Glowsticks just to switch things up with animations.

Chemlight Example:
chemlight_red

Flashlights:
Showcase


Heroman3003:
No offense, but pretty much all flashlights just straight up look worse than current versions to me.

PR #11489, opened 2021-08-24
New ears!

Very-Soft:
Adds some new ears, and the ability to make ears that are taller than would normally be possible with ears (since from the top of the human sprite to the top of the sprite is 3 pixels)

I plan to add more to this later but I am going to bed. UWU


KasparoVy:
That's bloody cool, I really like the addition of the sprite offset so you can have CN Tower ears
Loads of flexibility, I can easily see this being used for horns/synth antennae

PR #11491, opened 2021-08-25
Make some NIF items apply wear only when used, and not when purchased.

lbnesquik:
## About this pull request
As the title says, really. I do note that the strongest modules (illegal ones) still have wear from installation.

It makes more sense to have your wear and tear come from usage rather than just downloading a program. That's it.

### DIFF
--- a/code/modules/nifsoft/software/05_health.dm
+++ b/code/modules/nifsoft/software/05_health.dm
@@ -5,7 +5,6 @@
 	cost = 1250
 	p_drain = 0.05
 	a_drain = 0.1 //This is messed with manually below.
-	wear = 2
 	activates = FALSE //It is automatic in emergencies, not manually controllable.
 	tick_flags = NIF_ALWAYSTICK
 	applies_to = NIF_ORGANIC
@@ -16,6 +15,7 @@
 /datum/nifsoft/medichines_org/activate()
 	if((. = ..()))
 		mode = 1
+		nif.wear(1)
 
 /datum/nifsoft/medichines_org/deactivate(var/force = FALSE)
 	if((. = ..()))
@@ -47,6 +47,7 @@
 				var/turf/T = get_turf(H)
 				var/obj/item/device/radio/headset/a = new /obj/item/device/radio/headset/heads/captain(null)
 				a.autosay("[H.real_name] has been put in emergency stasis, located at ([T.x],[T.y],[T.z])!", "[H.real_name]'s NIF", "Medical")
+				nif.wear(2)
 				qdel(a)
 
 		//Handle the actions in each mode
@@ -78,7 +79,6 @@
 	cost = 1250
 	p_drain = 0.05
 	a_drain = 0.00 //This is manually drained below.
-	wear = 2
 	activates = FALSE //It is automatic in emergencies, not manually controllable.
 	tick_flags = NIF_ALWAYSTICK
 	applies_to = NIF_SYNTHETIC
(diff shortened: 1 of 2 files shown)


TheFurryFeline (commented):
I have certain concerns about what'll happen if someone were to be put into stasis for a long duration or if this will just be a 1 stasis action = wear. With the latter scenario in mind, it's possible that someone could drift in and out for a while, which could rack up the usage rate.

lbnesquik:
> 
> 
> I have certain concerns about what'll happen if someone were to be put into stasis for a long duration or if this will just be a 1 stasis action = wear. That said, it's possible that someone could drift in and out for a while, which could rack up the usage rate.

If that's too much of a concern, it could always be put simply on the activation rather than both activation and stasis.

TheFurryFeline:
> > I have certain concerns about what'll happen if someone were to be put into stasis for a long duration or if this will just be a 1 stasis action = wear. That said, it's possible that someone could drift in and out for a while, which could rack up the usage rate.
> 
> If that's too much of a concern, it could always be put simply on the activation rather than both activation and stasis.

Would depend on the situation, really. I hardly play anymore but I remember a few times where I went into stasis for several minutes on end, drifting in and out of consciousness and getting whacked by an Underdark mob each time until it roamed out of view and I was able to get away after waking up one more time, lol.

Heroman3003:
Size Alteration is now a dead program with this

lbnesquik:
> 
> 
> > > I have certain concerns about what'll happen if someone were to be put into stasis for a long duration or if this will just be a 1 stasis action = wear. That said, it's possible that someone could drift in and out for a while, which could rack up the usage rate.
> > 
> > 
> > If that's too much of a concern, it could always be put simply on the activation rather than both activation and stasis.
> 
> Would depend on the situation, really. I hardly play anymore but I remember a few times where I went into stasis for several minutes on end, drifting in and out of consciousness and getting whacked by an Underdark mob each time until it roamed out of view and I was able to get away after waking up one more time, lol.

The wear level is certainly not enough to kill your NIF.  There is over 30 uses of medimachines if your nif is full health. Even if it has only 50 of integrity, you still have 16 individual occasions to get put into stasis. This is plenty of reserve.

TheFurryFeline:
> > > > I have certain concerns about what'll happen if someone were to be put into stasis for a long duration or if this will just be a 1 stasis action = wear. That said, it's possible that someone could drift in and out for a while, which could rack up the usage rate.
> > > 
> > > 
> > > If that's too much of a concern, it could always be put simply on the activation rather than both activation and stasis.
> > 
> > 
> > Would depend on the situation, really. I hardly play anymore but I remember a few times where I went into stasis for several minutes on end, drifting in and out of consciousness and getting whacked by an Underdark mob each time until it roamed out of view and I was able to get away after waking up one more time, lol.
> 
> The wear level is certainly not enough to kill your NIF. There is over 30 uses of medimachines if your nif is full health. Even if it has only 50 of integrity, you still have 16 individual occasions to get put into stasis. This is plenty of reserve.

A'ight, that sounds fine then.

Detective-Google:
Not a fan of having your NIF take damage every time you charge off an APC as a synthetic. Means youre going to have to hunt down a charger more often which defeats the point of the program

lbnesquik:
> 
> 
> Not a fan of having your NIF take damage every time you charge off an APC as a synthetic. Means youre going to have to hunt down a charger more often which defeats the point of the program

It is one point of damage out of 100. You got plentyyyy of charges before you actually will need to head to an APC.

Novacat:
On one hand, this will result in less NIF wear for people who only occasionally use NIFSofts. On the other hand, this will result in a lot more NIF wear for people who heavily use the NIFsofts. Controversial. Can result in situations where someone accidentally deletes their NIF because they used it too much in the heat of the moment.

That said, I am personally not a fan of automatically activating NIFSoft programs having use wear because unlike the above situation, the user has no control over it activating or not.

lbnesquik:
I will return to this later on.

PR #11516, opened 2021-08-27
Fixes radio headset sound toggle not working

Arokha:
Also makes them a little less loud.

-- Earlyport of: https://github.com/PolarisSS13/Polaris/pull/8254


Little-119:
Why do we gotta have so many toggles?

PR #11518, opened 2021-08-28
Cleans a couple sprites around.

lbnesquik:
## About this pull request
This rodent sprite had a unique file with duplicate sprites when they are actually defined elsewhere. Removing.

Vangrants had its own dedicated file but for only 2 sprites, i believe it is better to have it in the main folder.

Easier and less cluttered sprite directory.


Arokha:
I think you might've forgotten to commit 'icons/mob/vore.dmi'. There's no "vagrant" state in there, and you didn't include any.

lbnesquik:
> 
> 
> I think you might've forgotten to commit 'icons/mob/vore.dmi'. There's no "vagrant" state in there, and you didn't include any.

Fixed. Forgot that bit.

PR #11519, opened 2021-08-28
Tool Time

GhostActual:
Pulls Skyrat's tool sprites into the fold. Many have been modified to deviate from their art style. There are some unique additions as well.

Ported the full brass set and a custom brass toolbox. The toolbox was set up in Surface-1 in a new hidden compartment on the northern side of the station.

Also added a bonus hydroponic tool kit for future POIs.


Detective-Google:
not a fan of how these look.

also 
Screenshot_20210828-172853_Chrome

GhostActual:
> not a fan of how these look.
> 
> also
> Screenshot_20210828-172853_Chrome

Box error was fixed

Very-Soft:
I'm okay with a lot of these, but some of them are quite small and will not be easy to click on, wirecutters particularly seem significantly smaller, and that will make clicking them annoying if they aren't in your inventory already.

PR #11535, opened 2021-08-31
Misc additions I made while sleep deprived

GhostActual:
Created a new explosive barrel sprite based on the burning barrel sprite.  Speeds up the burning barrel sprite animation. Adds a dirty bed sprite (the current system does not support the bed as an object, it tries to color it the material that the bed is made out of). Adds a new crude barricade I stole from Skyrat purely because when added on top of the other barricade it looks cool. And finally, I made a new critter cage which I posted below.

aaa


Screemonster:
icondiffbot says this is replacing the barrel rather than adding a new one

It's not that I dislike the new sprite, but the old one is specifically meant to be a doom barrel with the green swirly goop (and in code it's a welder fuel tank that can't be closed, so it always sloshes fuel everywhere when you try to move it)

GhostActual:
> icondiffbot says this is replacing the barrel rather than adding a new one
> 
> It's not that I dislike the new sprite, but the old one is specifically meant to be a doom barrel with the green swirly goop (and in code it's a welder fuel tank that can't be closed, so it always sloshes fuel everywhere when you try to move it)

Attempted to do the swirly, failed to do the swirly because tired. Gave up. Will attempt again so the burnie can burn.

GhostActual:
> icondiffbot says this is replacing the barrel rather than adding a new one
> 
> It's not that I dislike the new sprite, but the old one is specifically meant to be a doom barrel with the green swirly goop (and in code it's a welder fuel tank that can't be closed, so it always sloshes fuel everywhere when you try to move it)

Fixed et

Screemonster:
Hmmm. I'm torn. 'cause on the one hand, it's a technically more competent sprite than mine, it definitely looks like a barrel of goop, but...
I don't look at it and immediately go "that is a Doom barrel. The barrel from Doom. The barrel specifically chosen to look like the one from Doom. The Doom barrel."
When I plopped these down on an exploration map, the explorers _instantly_ knew what was up on the very first round.

KasparoVy:
> I don't look at it and immediately go "that is a Doom barrel. The barrel from Doom. The barrel specifically chosen to look like the one from Doom. The Doom barrel."

Lmfao, the reference made me laugh myself breathless. Now pull the lever!

PR #11562, opened 2021-09-03
Medical and utility belts of holding

PastelPrinceDan:
The recent-ish addition of pouches of holding + a friend's screencap of a durathread belt from another server gave me the idea of belts of holding. So, here we are!

- Both belts are available from RnD. They require gold, diamond and titanium (because we barely use titanium). Pretty much the same material amount as pouches of holding.
- They both have twice the space. So, 14 slots instead of 7.
- The tool-belt can hold what tool-belts hold, of course, but also: power cells, RPDs, RCDs. The logic is that it's a bigger belt, so it can hold bigger things. 
- The med belt can hold what normal medical belts can, no additions.
- ...The sprites for them are recolors of the PF and explorers belt that I made in two minutes because I'm not an artist by any means. If anyone wants to resprite them down the line, please do.


dwinters99:
As useful as they'd be, I worry about the amplification of the already long lists that people bring RnD at what is essentially roundstart. Even if it's just the one item more, it's still going to lengthen people's "shopping" lists. I'm somewhat confident that the existence of the utility belt will cause a massive uptick in RnD orders but that may not be a bad thing. Also, the normal bag of holding isn't requested very often anyhow. I *would* recommend a misteleport balance function similar to the bags to prevent this from being a mandatory ask.

PastelPrinceDan:
I'm gonna be honest, the 'mobile chemlab' issue is already a possible issue with pouches of holding (they can also hold 14 [if not 16? I forget] normal items, ie bottles or pill bottles); however, I haven't seen it happen, because medical doesn't care/is reasonable enough not to do it. I don't think adding a belt will make it an issue.

As for adding an item to people's lists... I don't.. see an issue? It gives research one more click to do, sure, but that's it. Medical and engineering barely ask for existing upgrades unless they're reminded of them anyways, so I doubt many people will ask for them.

Aaaand for the misteleport chance, I'll see about adding one if admins/maintainers ask for it, but pouches don't have one, and this is essentially just pouch-but-restricted-items-list.

PR #11574, opened 2021-09-08
CentCom Revamp

GhostActual:
It started as makin the walls of CentCom end-of-round map look better. Then it spiraled from there.


Very-Soft:
I will probaby expand on this more later, but looks like a good start! <3

GhostActual:
> I will probaby expand on this more later, but looks like a good start! <3

Yay!

GhostActual:
Forgot to make it a draft cause really this is still a WIP

KillianKirilenko:
Have you considered experimenting with the portal system? The station itself could be a subway/underground and then you could have a small 'surface' type area, without needing to make a whole extra z-level.

GhostActual:
> Have you considered experimenting with the portal system? The station itself could be a subway/underground and then you could have a small 'surface' type area, without needing to make a whole extra z-level.

Soft will have to be the deciding factor on that, they know more about the Colony as a whole than I do. The way they put it is that the elevator bank actually sends you up to the surface level. I dunno, I still need to read up on the lore of this world to get a proper foothold in world building. Especially with upcoming ideas I have about the little moon that could.

Very-Soft:
> Have you considered experimenting with the portal system? The station itself could be a subway/underground and then you could have a small 'surface' type area, without needing to make a whole extra z-level.

I was wanting to make CC be its own Z again for a while now. With things as they are on the server right now, there's practically no space on the Z level to actually expand CC in a significant way, and, CC is in the way of expanding other things too, and I personally would like CC to be more like how it used to be, with areas to do some RP in, and in general space to experiment with stuff, ideally with things pre-set to enable that (for admin experiments or the times where someone wants to try something wacky without risking blowing up the station)

I'd also like the thunderdome to be a thing there again as well. AND I would like the option of having more kinds of holodeck programs, and, because of CC being where it is, there's nearly no space for making more of them! SO! 

I think CC should be its own Z. I miss the utility of the old CC, the new smole one you can hardly do anything with, without significant modification. 

AS FOR portals and the outdoors areas. I'm not too sure. I don't really think we need to work in outdoors areas but they're also fine if they look nice aesthetically. I'm not too fussy about it.

KillianKirilenko:
> I was wanting to make CC be its own Z again for a while now. With things as they are on the server right now, there's practically no space on the Z level to actually expand CC in a significant way, and, CC is in the way of expanding other things too, and I personally would like CC to be more like how it used to be, with areas to do some RP in, and in general space to experiment with stuff, ideally with things pre-set to enable that (for admin experiments or the times where someone wants to try something wacky without risking blowing up the station)
> 
> I'd also like the thunderdome to be a thing there again as well. AND I would like the option of having more kinds of holodeck programs, and, because of CC being where it is, there's nearly no space for making more of them! SO!
> 
> I think CC should be its own Z. I miss the utility of the old CC, the new smole one you can hardly do anything with, without significant modification.

I should clarify then - I think making CC it's own z-level and splitting it off from the overmap/holodeck programs/etc. is a fine idea, The trick with the portals is that you can fake a map having *multiple* z-levels when it's really just the one with a little careful layout work.

As a hypothetical example; if you wanted to be really fancy you could probably replace the tram with a transport shuttle and have it land on the top of a transit terminal building which is then made up of several smaller floors, and use the portal and/or lookingglass systems to produce the illusion of a multi-story building within a single z-level. Or something showy like that.

> AS FOR portals and the outdoors areas. I'm not too sure. I don't really think we need to work in outdoors areas but they're also fine if they look nice aesthetically. I'm not too fussy about it.

I'm just thinking in terms of "oh, the station happens to be near the edge of the dome so you can see out in a few places" or the aboveground level feels more a bit like a suburban area with streets and buildings rather than an underground area. Checkpoints into other parts of the colony could then be exits besides the residential elevators, or something like that.

Very-Soft:
> > I was wanting to make CC be its own Z again for a while now. With things as they are on the server right now, there's practically no space on the Z level to actually expand CC in a significant way, and, CC is in the way of expanding other things too, and I personally would like CC to be more like how it used to be, with areas to do some RP in, and in general space to experiment with stuff, ideally with things pre-set to enable that (for admin experiments or the times where someone wants to try something wacky without risking blowing up the station)
> > I'd also like the thunderdome to be a thing there again as well. AND I would like the option of having more kinds of holodeck programs, and, because of CC being where it is, there's nearly no space for making more of them! SO!
> > I think CC should be its own Z. I miss the utility of the old CC, the new smole one you can hardly do anything with, without significant modification.
> 
> I should clarify then - I think making CC it's own z-level and splitting it off from the overmap/holodeck programs/etc. is a fine idea, The trick with the portals is that you can fake a map having _multiple_ z-levels when it's really just the one with a little careful layout work.
> 
> As a hypothetical example; if you wanted to be really fancy you could probably replace the tram with a transport shuttle and have it land on the top of a transit terminal building which is then made up of several smaller floors, and use the portal and/or lookingglass systems to produce the illusion of a multi-story building within a single z-level. Or something showy like that.
> 
> > AS FOR portals and the outdoors areas. I'm not too sure. I don't really think we need to work in outdoors areas but they're also fine if they look nice aesthetically. I'm not too fussy about it.
> 
> I'm just thinking in terms of "oh, the station happens to be near the edge of the dome so you can see out in a few places" or the aboveground level feels more a bit like a suburban area with streets and buildings rather than an underground area. Checkpoints into other parts of the colony could then be exits besides the residential elevators, or something like that.

I don't disagree with you necessarily, I just do want it to be relatively easy to navigate and to be generally functional. If someone wants to go and make it look all multi-z that's fine, but dedicating a lot of space and time to making it look like there's stuff outside is not as important to me as it being a generally functional place to experiment with mechanics and do a bit of RP.

GhostActual:
floors and walls were ported to a new PR

PR #11582, opened 2021-09-10
Adds large autoinjectors to autolathes

PastelPrinceDan:
Made at Tempest's request.

- 15u capacity
- Costs 2000 plastic on T1. Might be a steep cost, but it's large injectors + it'll be reduced once the lathe is upgraded anyways.
- Recyclable


Wickedtemp:
OKAY SO explanation and reasons time.

The reason why I think this should be added is that to my knowledge, there's no way to actually get these injectors, you can only get the small ones, and that's pointless when the large injectors with reagents are purchasable with points. That's fine, but it also invalidates the existence of Medical when it comes to supplying a team. Medical isn't always around but this allows a department to supply another with potentially useful, potentially _better_ injectors than what they'd be getting from the vendor, or at the least, they can spend points on other things since they've already been supplied injectors.

PR #11584, opened 2021-09-11
[MIRROR] :siffet:

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/8289
--------------------
Adds a siffet costume. Icon by @ Schnayy.


Arokha:
IDB says this modifies every state, probably an IDB bug but needs investigation

PR #11597, opened 2021-09-15
Atmospheric Revamp

GhostActual:
Adds in new custom sprites for various atmos machines and replaces the oxygen tank dispenser with the 'objects_vr' version which is more modern.


Arokha (changes requested):
You can't just remove those iconstates and not replace them with something. The space heater still uses them. It'll just become invisible now. Don't do that.

GhostActual:
Ignore the white pixel in the `sheater3` sprite, that's a Github error

PR #11598, opened 2021-09-16
Fix Hot/Cold Discomfort Message Strings

Rykka-Stormheart:
Cold Discomfort would not display if your body was covered - every /under/ item by default did so.
Heat discomfort would not display if your body was UNcovered.

This is fixed now - once you're uncomfortable, irregardless of what you're wearing, you'll start feeling it.
This should help prevent random deaths to freezing or overheating as you'll have warnings the area is too warm/cold before you start going ouchie.

### DIFF
--- a/code/modules/mob/living/carbon/human/species/species_getters.dm
+++ b/code/modules/mob/living/carbon/human/species/species_getters.dm
@@ -79,9 +79,10 @@
 
 /datum/species/proc/get_environment_discomfort(var/mob/living/carbon/human/H, var/msg_type)
 
-	if(!prob(5))
+	if(!prob(10))
 		return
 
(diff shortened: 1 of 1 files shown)


KillianKirilenko:
..seriously, (lack of) *clothing* was suppressing the warning messages even when you were actually dangerously hot/cold? Good grief.

Rykka-Stormheart:
> ..seriously, (lack of) _clothing_ was suppressing the warning messages even when you were actually dangerously hot/cold? Good grief.

Polaris is pushing back that it's "intended" and it should "just throw another hud alert for temp warning", when, like, the messages are intended as an RP method to notify you of temperature, just like pain messages in chat.

PR #11605, opened 2021-09-18
Begone, active edge errors!

KillianKirilenko:
Squashes active edge errors on the wilderness/plains by cladding the unsimulated north wall with rock in a few spots where it was exposed.

Also subtypes the areas for the edges in case someone gets downed/lost there but can be tracked on GPS or other sensors. The last thing Medical needs is to be running in the complete opposite direction of someone who needs help.


Very-Soft:
The active edges aren't coming from the north wall exclusively, the wall was exposed in a few spaces because something about some of the POIs is massively raising the temperature in the plains, which kills all the mobs and anyone not wearing a full space suit. (And still causes active edges.)

The exposed unsimulated wall is meant to correct messed up temp from those POIs. But someone's gonna have to go through them and figure out which ones are messing things up.

KillianKirilenko:
Temperature shenanigans, you say? Hmm. I didn't see anything that looked particularly like it would have been *increasing* temps when I did the audit. 🤔

My best guess is that the temperature gap between modified and unmodified turfs was able to cause it, for stuff like the boathouse which has a fairly sizeable pond for boating on.

Very-Soft:
Seems like after your POI audit, I haven't noticed any more active edges happening!

KillianKirilenko:
Good to hear. It got a bit messy with all the subtyped terrains (I'm sure there's a better way to do it by like, iterating through all `outdoors = 1` turfs on the level at init and setting their atmos there or something?) but it'll have to do for now.

Thankfully StrongDMM made the audit much easier thanks to its search-and-replace function.

PR #11606, opened 2021-09-18
Plains POI Audit

KillianKirilenko:
Works over every active Plains POI and adjusts their turfs to use new v3b-specific turfs (added in `plains_turfs.dmm` using `VIRGO3B_TURF_CREATE`). Some minor adjustments were made to them where necessary, but contents/etc. are otherwise unchanged.

Pairs with #11605 to (hopefully) squash active edge errors coming from the plains. Will cut down on weird cases of buildings starting full of breathable air despite being partially or completely exposed to the V3b atmosphere.

Also removes a couple of unused duplicates that were simply just "POI X but with no areas".


Very-Soft:
Oh you did this! I will just merge this one for now to see if it fixes the active edges, since, there's not actually any reason the north wall should be causing active edges, since it should be doing the same air as the air that exists there already, which is how all the other virgo 3b maps work.

PR #11610, opened 2021-09-18
adds audible belch emote

lm40:
this is obviously inspired heavily by what radiantaurora did on chomp with jit's sounds, although the implementation is somewhat different/simplified even though the sounds are the same

this just adds an audible belch emote


lbnesquik:
I feel like this deserves a preference toggle

Detective-Google:
most vore sounds can be turned off with audio settings in preferences. This ought to be it's own preference really because talking with friends I know I'm not the only one who doesn't really want to hear burping trying to play. There's a reason it's considered bad manners irl and all.

lm40:
it is not viable to make a preference for a single emote outside of the general audible emote preference because of the way the emote system works

i don't really see it as more disruptive than any sounds we already have

lbnesquik:
> 
> 
> it is not viable to make a preference for a single emote outside of the general audible emote preference because of the way the emote system works
> 
> i don't really see it as more disruptive than any sounds we already have

If you can turn off digestion noises, you should be able to turn off this one.  
We have a ton of preference options already, this is not going to impact the game in any way to have one more.

On top of all of this, the amount of prefs we already have gives you a lot of examples to base yourself on.

lbnesquik:
Chomp has this by the way, maybe copy that as well. 

image

Iroquois-Pliskin:
Audible emotes were in almost all cases a mistake and get abused time and time again. I know for a fact that plenty of players skip the tram for that reason alone.

And now you wanna add in burping as an audible emote? Please. No.

PR #11624, opened 2021-09-24
Adds audible belching preference toggle, makes belch emote respect it

Rykka-Stormheart:
Preferences can now handle a list being passed to them, and will require all items in the list to be true. Singular datums work to preserve legacy behavior.

Audible belching will REQUIRE "Emote Noises" and "Audible Belching" preferences to be enabled.

Figured there'd be a potential shitstorm, and honestly adding the preference was easier than I thought.

This will work if for w/e reason you decide to add farting noises - a simple pref toggle, and bingo.

Etc etc.

<3

### DIFF
--- a/code/modules/client/preference_setup/global/setting_datums.dm
+++ b/code/modules/client/preference_setup/global/setting_datums.dm
@@ -101,7 +101,6 @@ var/list/_client_preferences_by_type
 	key = "BELCH_NOISES"
 	enabled_description = "Noisy"
 	disabled_description = "Silent"
-	enabled_by_default = FALSE
 
 /datum/client_preference/emote_noises
 	description = "Emote Noises" //MERP
--- a/code/modules/client/preference_setup/global/setting_datums.dm
+++ b/code/modules/client/preference_setup/global/setting_datums.dm
@@ -95,6 +95,13 @@ var/list/_client_preferences_by_type
 	key = "DIGEST_NOISES"
 	enabled_description = "Noisy"
 	disabled_description = "Silent"
+	
+/datum/client_preference/belch_noises // Belching noises - pref toggle for 'em
+	description = "Burping"
+	key = "BELCH_NOISES"
+	enabled_description = "Noisy"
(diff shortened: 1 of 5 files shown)


code/modules/client/preference_setup/global/setting_datums.dm
```diff
@@ -95,6 +95,13 @@ var/list/_client_preferences_by_type
 	key = "DIGEST_NOISES"
 	enabled_description = "Noisy"
 	disabled_description = "Silent"
+	
+/datum/client_preference/belch_noises // Belching noises - pref toggle for 'em
+	description = "Burping"
+	key = "BELCH_NOISES"
+	enabled_description = "Noisy"
+	disabled_description = "Silent"
+	enabled_by_default = FALSE
```
Heroman3003, line 104: Should be true by default, but thats imo

Rykka-Stormheart: CHOMP has it off by default, and it works out fine. Y'hear a belch, decide you don't want it, so turn it off. <3

I'd rather it be off by default, let people decide if they dislike it to find the toggle to turn it off.

Heroman3003: Wait, correct me if I'm reading this wrong, but doesn't "Not enabled by default" mean nobody hears it unless they actively go to find pref and turn it on? Which is opposite of what you describe? Might be forgetting which way noise preferences go there, just making sure.

Very-Soft: It should be on by default. If people want to turn it off they can do that. If it's off by default most people won't even know to go find the toggle.

Rykka-Stormheart: I'm tired and misread it, yeah. My bad.
Will fix, one sec.

Rykka-Stormheart: Resolved.

PR #11641, opened 2021-09-30
Removes geese from immedeate Tether surroundings

Heroman3003:
They may not be dangerous to people going outside too much, but they are still hostile simplemobs that will break into station through the windows should they spot something inside, forcing people out of areas that are supposed to be perfectly safe.


Screemonster:
A goose migration might be fun as a ground-based version of the carp migration event, but specifically as an event that, like the carp, removes them once its done.

PR #11647, opened 2021-10-02
[WIP]Ship map!

Very-Soft:
This ship is designed to be an alternative to the tether. Kind of a counterpoint, it is deliberately small and stripped down of lesser used fluff areas and such in service of keeping the size small. 

The basic layout and design is largely done, with the majority of the remaining work on the ship itself being wires, pipes, areas, and general room essentials, and the general point defense turret network for the ship. (And making the shuttles actually work)

There are A TON OF OTHER GENERAL ISSUES with running a map other than the tether that need to be resolved an I will probably need help with.

The ship's design at its core is to be small and smartly designed, quick and easy to navigate. One of my major goals was to make sure that the department desks and such were easily visible from the main hallways to assist with discovery if someone's doing a job, while also doing my best to avoid dead ends as much as possible, especially in the main halls. WHILE keeping it all to a basic small vaguely space ship shaped and trying to pack it as much into a tight pre-built hull. 

THE NAME is still in limbo since I kind of hate what I've come up with so far. UWU


Heroman3003:
Can screenshots of departments or such be provided separately? Mapdiff bot loves to just sometimes not load bottom half of the image properly.

PR #11649, opened 2021-10-03
Fixes #11648

### LINKED ISSUE #11648: throwing items a single tile doesn't  work
if you throw an item to an adjacent tile, the item appears on the adjacent tile, but ALSO remains in your hand.

lm40:
for whatever reason forcemove to your tile or an adjacent tile does not make you drop the item


JoanRisu (approved):
From what I can tell, this should be good to go. Unless someone has concerns regarding this, I'll go ahead and merge this in the morning.

PR #11673, opened 2021-10-10
removes 'empty ore box' verb because i'm done with crashes

PastelPrinceDan:
what it says on the tin

just comments out the empty ore box verb because any time it's used with too-large amounts of ores the server crashes

i'll try to see if there can be another fix possible for it but for now this is my solution

### DIFF
--- a/code/modules/mining/ore_box.dm
+++ b/code/modules/mining/ore_box.dm
@@ -58,6 +58,34 @@
 	for(var/ore in stored_ore)
 		. += "- [stored_ore[ore]] [ore]"
 
+// /obj/structure/ore_box/verb/empty_box()
+//	set name = "Empty Ore Box"
+//	set category = "Object"
+//	set src in view(1)
+//
+//	if(!ishuman(usr) && !isrobot(usr)) //Only living, intelligent creatures with gripping aparatti can empty ore boxes.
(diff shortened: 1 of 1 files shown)


DigitalSquirrel95:
Given I made literally the exact same PR within seconds of this being posted as well, I'm all for this.

FartMaster69420:
Is it even possible to collect strange rocks in an orebag/box? I've never seen that happen.

Heroman3003:
> 
> 
> This leaves the problem of strange rocks not being recoverable if unable to grab before it goes into the smelter, or if the miner doesn't pick them up while not using the ore bag.

Strange rocks dont even go into the bag anymore, cuz IIRC that was changed years ago.

Heroman3003:
> 
> 
> > This leaves the problem of strange rocks not being recoverable if unable to grab before it goes into the smelter, or if the miner doesn't pick them up while not using the ore bag.
> 
> as stated above, strange rocks don't get picked up by ore satchels. the only thing that could be an issue with this is bluespace crystals, which.. the only genuine one i've ever seen is at a poi in the UD where you're likely to pick it up rather than just walk over it and empty it in your ore box.

Are bluespace crystals even considered ore on this codebase? Iirc that shouldn't affect them either.

PR #11687, opened 2021-10-17
Probably fixes fatal turf underlay stacking

Verkister:
uh oh chief
kuva


Verkister:
Didn't fix the whole thing, still looking for whatever i might've missed

Razgriz1032:
I think I actually fixed it here.
https://github.com/CHOMPStation2/CHOMPStation2/pull/3072

Unfortunately it didn't fix the actual lag with our shuttles afaics, but it stopped the underlay stacking.

PR #11692, opened 2021-10-18
Fix getting stuck in deep space because of bad creation of temporary sectors

Leshana:
Deep Space temporary sectors were being created on Z1 instead of the overmap, and with a bad map_z list.
- Initialize was taking the wrong parameters, resulting in the sector being always located at 1,?,1 instead of on the overmap
  Fix by passing the first parameter to new() as the location you want the object, as is proper for BYOND.
- Also fixed map_z containing a zero, it was put there by the sector Initialize().
  Fix by setting the appropriate values *before* calling parent.

This hopefully should also fix (or at least mitigate) the creation of infinite "temporary" empty z levels because:
A) Someone lost in a overmap tile with no real sectors will get recycled back into the same Z without making a new one.
B) Attempts to cleanup will no longer *always* think players are present because it is counting all players in nullspace.

### DIFF
--- a/code/modules/overmap/spacetravel.dm
+++ b/code/modules/overmap/spacetravel.dm
@@ -5,15 +5,16 @@
 	known = FALSE
 	in_space = TRUE
 
-/obj/effect/overmap/visitable/sector/temporary/Initialize(var/nx, var/ny)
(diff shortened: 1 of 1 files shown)


Very-Soft:
Praise~!

PR #11696, opened 2021-10-21
Shuttle Fixes/Additions

GhostActual:
Finally fixes the backup shuttle that takes you to the station (adds lights and changes the base-turf to get rid of the see-through floor) and adds 'shuttle' blast doors and shutters. Basically, blast doors/shutters that have no base sprite so they're invisible when open. Once again the pilots can feel safe in their cockpit on the tourbus when they spin out of control and head into that carp shoal.

tourbus

sdboat


GhostActual:
There are things I want to change and maybe play around with so I'm converting to a draft until I figure it all out.

PR #10076, opened 2021-04-04
Von Braun OFD

KillianKirilenko:
Adds a crude-but-functional implementation of the Obstruction Field Disperser to the ERT ship, the NRV Von Braun. Replaces the old mech bay.

vb_ofd

For more instructions on the OFD, refer to [this page](https://wiki.baystation12.net/Obstruction_Field_Disperser).

Mostly a proof of concept for later implementation on the mercenary ship and possibly sec/eng "shuttles". Does not appear to be usable as a ship-to-ship or ship-to-surface weapon at this time.

:cl:
tweak - added an OFD to the Von Braun (ERT ship)
/:cl:

PR #11721, opened 2021-10-28
Talon Shuttle Tweaks

KillianKirilenko:
Based on some comments in the offmap channel, it sounds like the Talon Shuttle's side-door doesn't function like I wanted and is actually quite unsafe.

As such I've quickly remapped it to remove the side door, and reorganized the center bay a little as a result. Two less handrails, but you've already got six passenger seats and ten handrails so it's not like there's a shortage of seating. ~~And even if there was I'm sure people would find creative ways to secure additional passengers.~~

I've also adjusted the airlock to add a dump system so it can pump out hazardous atmospheres (e.g. V2's) if need be. It'll probably be pretty slow though.

:cl:
fix - made talon shuttle safer, added dump vent
/:cl:

### DIFF
--- a/maps/offmap_vr/talon/talon_v2.dmm
+++ b/maps/offmap_vr/talon/talon_v2.dmm
@@ -3353,6 +3353,15 @@
 	},
 /turf/simulated/floor/tiled/techfloor,
 /area/talon_v2/maintenance/wing_starboard)
+"kv" = (
+/obj/effect/map_helper/airlock/atmos/pump_out_external,
+/obj/machinery/atmospherics/unary/vent_pump/high_volume{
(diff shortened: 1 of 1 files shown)


KillianKirilenko:
~~What if I'm super lazy and just set the starboard-side blast doors to closed by default?~~

I'll fiddle with it some and see what I can do. Though I'd also like to know the staff's stance on the Talon possibly getting an ORB/OFD (compare #10076) of its own at some point. Probably not more than two of each shot type standard-issue.

KillianKirilenko:
Closing for the moment due to some technical stuff, but still working on this.

PR #11728, opened 2021-10-30
Sensible Dragon Boss Tweaks for Virgo

BigBababooey:
A few minor tweaks to make the Dragon a more reasonable boss for Virgo purposes. Drastically cutting it's base speed from Movement Cooldown 2 to Movement Cooldown 5 and lowering health by 25%. Be warned, it still has it's pounce, so this doesn't magically become easy at all. However also fixed it's issue where it chokes to death so it can't just be cheesed. Still a bandaid fix, it's more suited for V2 admittedly, but something is better than nothing!

### DIFF
--- a/code/modules/mob/living/simple_mob/subtypes/vore/bigdragon.dm
+++ b/code/modules/mob/living/simple_mob/subtypes/vore/bigdragon.dm
@@ -71,7 +71,7 @@ I think I covered everything.
 	response_disarm = "shoves"
 	response_harm = "smacks"
 	movement_cooldown = 4 //Fixed from 2, given our slower natural speed any mob at movement cooldown 2 is a nightmare let alone a boss that has a pounce
-	maxHealth = 800 //Still double a Phoron Dragon, if it's gonna be on 3b, it shouldn't have broodmother level health
+	maxHealth = 800
 	attacktext = list("slashed")
 	see_in_dark = 8
 	minbodytemp = 0
--- a/code/modules/mob/living/simple_mob/subtypes/vore/bigdragon.dm
+++ b/code/modules/mob/living/simple_mob/subtypes/vore/bigdragon.dm
@@ -70,7 +70,7 @@ I think I covered everything.
 	response_help = "pats"
 	response_disarm = "shoves"
 	response_harm = "smacks"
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/simple_mob/subtypes/vore/bigdragon.dm
```diff
@@ -70,8 +70,8 @@ I think I covered everything.
 	response_help = "pats"
 	response_disarm = "shoves"
 	response_harm = "smacks"
-	movement_cooldown = 2
-	maxHealth = 800
+	movement_cooldown = 5 //Fixed from 2, given our slower natural speed any mob at movement cooldown 2 is a nightmare let alone a boss that has a pounce
```
Very-Soft, line 73: I think 5 will make it REAL SLOW, which will make it very easy to kite. If this thing is supposed to be dangerous, removing a fourth of its health AND making it crawl is gonna make explorers with normal gear laugh at it.

BigBababooey: This comes from a rather consistent trend of seeing it cause wipes, even against otherwise compitent explo teams. As I noted in the comments, it still has it's pounce abilites and the like for the sudden burst of speed, when considered with the fact a single hit from this things abilities will make you slow enough that it's gonna maul you. As for the health nerf, it's perhaps unneccsary, this was based on me observing dragon corpses from these wipes (when atmos finally got to them) and noting often well over half the damage wasn't even done by them. Perhaps kneejerk, I could see a case for reverting the health shift but I'd be incredibly wary of any lower than movement cooldown 4. As a 3b boss, this thing is insane and not in a way people seemed to be having fun with outside of GM control, and even outside it's pretty damn tough.

BigBababooey: Put simply, we took a boss mob designed for another server's speed and to be reasonably difficult there, and then threw it on ours with the same speed, any less than halving it is silly, and I'd rather not have a massive thing in the starting area that demolishes all but the most robust if they stumble into it, because especially with it's speed it can cascade into killing people who aren't even aware of the danger when it stumbles way, which I've also seen happen quite a few times and it usually kills all explo mood hard when it does.

PR #11734, opened 2021-10-30
Takes Stellar Delight apart

Heroman3003:
into different files.

Splits up general stellardelight_misc into several different, smaller files with more purpose focus. Also takes some things that are common between Stellar Delight and Tether and moves them into general code files.


Jakksergal:
Could you expand this to include talon jobs into a shared file? They're currently only defined in Tether's files and aren't selectable roles on the Stellar Delight despite it also being there.

Heroman3003:
> 
> 
> Could you expand this to include talon jobs into a shared file? They're currently only defined in Tether's files and aren't selectable roles on the Stellar Delight despite it also being there.

This PR does not change anything in terms of how game actually works intentionally, and any 'actual changes' such as that would have to be done separately.

And already were, looking at recently merged PRs.

PR #11754, opened 2021-11-02
Adjustments to mob falling mechanics

Heroman3003:
Makes simple mobs of size MOB_SMALL and smaller not take fall damage from non-planetary falls. Have you seen how high a height can small creatures fall off of and just run away afterwards?

Makes the softfall ability as well as the afromentioned new size-based change affect the falling on other mobs. If fall would be safe for the falling mob, then neither it, nor the la(d/ss) below is affected either. No change in situation where dropnom prefs match one way or another.

Reduced damage from one mob falling onto another. Instead of both recieving full fall damage individually, the damage of a single fall is now split between the two evenly.

In dropnoms, if pred is the one falling from above, then if the fall would have been safe, they are no longer stunned.

### DIFF
--- a/code/modules/multiz/movement.dm
+++ b/code/modules/multiz/movement.dm
@@ -467,13 +467,16 @@
 // Take damage from falling and hitting the ground
 /mob/living/fall_impact(var/atom/hit_atom, var/damage_min = 0, var/damage_max = 5, var/silent = FALSE, var/planetary = FALSE)
 	var/turf/landing = get_turf(hit_atom)
+	var/safe_fall = FALSE
+	if(src.softfall || (istype(src, /mob/living/simple_mob) && src.mob_size <= MOB_SMALL))
+		safe_fall = TRUE
 	if(planetary && src.CanParachute())
 		if(!silent)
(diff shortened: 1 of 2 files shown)


Screemonster:
Very nice. The only situation I can think that hasn't been covered here and probably should is a micro falling on a macro (maybe fall damage proportional to the size ratio rather than an even split?)

Heroman3003:
> 
> 
> Very nice. The only situation I can think that hasn't been covered here and probably should is a micro falling on a macro (maybe fall damage in that case should be proportional to the size ratio rather than an even split, or else just negate damage for the person being landed on if the person falling is small enough for them to scoop)

Main reason I didn't do it is to avoid adding any mechanical advantage/disadvantage to micro/macro mechanics. Since the situation when it would be happening is specifically "when anything vorny doesn't occur", I just thought to not touch on it. Sensibility and logic of macro/micro interacting with game mechanics is a deep rabbit hole that I'd personally prefer to steer clear of, hence why I didn't add any scaling-related checks to the PR

Screemonster:
yeah fair enough

PR #11755, opened 2021-11-03
neck gaiter second pass - thank you @dan

Hatterhat:
NECK GAITERS:
- repathed to /accessory
- still wearable in mask slot
- have an item sprite from CM so it's not just the onmob sprite
- has an action button
- can now be attacked with a breathmask to put it behind the mask, granting it the AIRTIGHT flag for breathing purposes
- also have FLEXIBLEMATERIAL so you can eat stuff even with it coverin your face. i think


Very-Soft:
I think one thing I would ask is that it becomes not flexible when you put the breathmask behind it, since IIRC you can't eat when you are wearing those, so, it wouldn't make much sense to make that not so.

Hatterhat:
it's failing thanks to an active edge

PR #11763, opened 2021-11-03
Talon Improvements, Take 2

KillianKirilenko:
Expanded version of #11721.

- Restructures the Talon shuttle to be safer. The 'quick access' door has been removed. The starboard airlock doors are now seperate from the port doors, and will not auto-open. They can be manually bolted with a switch just above them. The blast doors remain in place on both sides as an emergency/fallback option.
- A distress ~~bacon~~ beacon has been added to the shuttle.
- The forward hull has been extended a bit to add an OFD/ORB bay and Escape Pod! As a result the whole hull has been shifted 'down' a bit, but you shouldn't notice this part as a player
- The OFD/ORB comes with two EM charges and two EXP charges, and has a sensors console for aiming/checking targets. If you want more OFD ammo, trade with the Tether/SD! The firing port shutter is controlled by the bridge shutters.
- Adds custom emergency supply lockers which contain the following items; a softsuit (inc. helmet), a breathmask, an oxygen tank, an emergency suit cooler, a gps unit, a random MRE, a flashlight, a flare, a survival knife, and a bottle of water.
- A couple of autolok suit lockers on the talon's shuttle have been replaced with emergency supply lockers.
- The Escape Pod cannot fly on the overmap but can land at planets if ejected over them. It has seats for eight, and each seat comes with one of the new emergency supply lockers.
- The wingtip airlocks have been retrofitted into full docking ports for direct access. Whether they work properly or not remains to be seen.

Escape pod prototype preview;
image


Very-Soft:
I am anxious about the talon getting an orb launcher, but I need to play with it some to see. I don't personally really want people being able to just clear paths through space weather normally, since we tend to use the space weather in some event stuff, and when given a very simple solution (shoot it) players tend to pick that, no matter how long or annoying that might be over more creative solutions. I don't want people to just never engage with mazes and things of that nature because it's possible to just fly a straight line through a thing with no risk.

I'm not saying no to that for now more, I wanna sit down and play with it a bit before I say yes, because I've never messed with the orb launcher. I don't PERSONALLY feel like it's at all nessisary normally since you can almost always fly around space weather, and, when you can't, both of the big ships have shields that they could use to slip through weather if managed properly.

KillianKirilenko:
Keep in mind that the OFD/ORB has;
- A limited range (1 overmap tile)
- A long cooldown (120s between shots, trying to fire during cooldown makes explode!)
- A fiddly calibration system (it's basically bulls-and-cows/mastermind)
- Ammo limitations and type restrictions

If you don't want people using it to "solve" events, delete the ammo (or even the launcher) during roundstart and cite some kind of supply shortage or maintenance issues. The long cooldown and calibration stuff means it won't really help people solve problems any faster than shields would let them simply fly through said problems already.

Anyway, I agree it's not really necessary, but I wanted to try adding one since it would give the Talon crew something they'd probably want to buy (OFD ammo) and a service they could offer (clearing obstructions between overmap locations) that the Tether and Stellar Delight lack, giving them more reasons to trade and generally interact with the crew.

FartMaster69420:
Giving the Talon an ORB gives it a bit of a mechanical distinction from SD that I like.

KillianKirilenko:
I see CI is continuing to be a pain in the ass.

KillianKirilenko:
@Very-Soft y'all figured out what you wanna do with this one?

If the ORB is a no-go I'll probably switch the side the escape pod is on and extend the canteen. Someone suggested more cooking equipment and I think I can squeeze a couple things in if I use a bit of elbow grease. Either way it's holding up some QoL/safety changes and has been for a month now.

KillianKirilenko:
Emergency lockers now have coolers as well, for synth crew, since I initially overlooked that.

And here's a preview of the ORB room in its current arrangement;
image

KillianKirilenko:
@Very-Soft it's been two months now, can I get an answer on this one?

KillianKirilenko:
Prepped an alternative layout with no ORB/OFD. The pod has been moved across the hall and the bar/canteen has been rearranged to support basic kitchen services. Supplies are limited compared to a proper kitchen so aspiring chefs will have to get creative, but I think this fits well.

Very-Soft:
Sorry it's taken so long to get to. I played with it a bit and I'm not quite so concerned. If the ORB becomes a problem, I may want to remove it later! But, for now let's give it a shot.

PR #11772, opened 2021-11-04
Adds Soft Falling positive trait

Heroman3003:
Adds new 1-cost positive trait, similarly to Winged Flight in balancing idea.
Soft Falling - causes falling from non-planetary heights to deal no damage and mitigates fall damage when landing on people and short stun from drop-nomming people.

Trait is present by default for Teshari, Rapala and Nevrean

Teshari's old slowfall has been removed along with both its unique messages as well as restrictions.

### DIFF
--- a/code/modules/mob/living/carbon/human/species/species.dm
+++ b/code/modules/mob/living/carbon/human/species/species.dm
@@ -208,6 +208,7 @@
 	var/greater_form										// Greater form, if any, ie. human for monkeys.
 	var/holder_type
 	var/gluttonous											// Can eat some mobs. 1 for mice, 2 for monkeys, 3 for people.
+	var/soft_landing = FALSE								// Can fall down and land safely on small falls.
 
 	var/rarity_value = 1									// Relative rarity/collector value for this species.
 	var/economic_modifier = 2								// How much money this species makes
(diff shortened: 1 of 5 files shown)


Heroman3003:
Mostly inspired by my previous PR that adjusted how falling on other people works. Plus, with increase in multiz mechanics on Stellar Delight, I think something for safely moving down that's not flight would be cool, I think.

Very-Soft:
I'm not personally the biggest fan of this because, people falling down and getting hurt in a not fatal, relatively trivial manner was one of the main purposes behind making the Z3 maint stuff have spaces for people to fall down in to give medical more non resleeving related work and opportunities to interact with people relating to their job with damage scales tot he point where they're unlikely to be racing a clock, so more roleplay and work for them can happen.

If anyone can just take this for no cost, then it completely negates that opportunity for interaction.

Heroman3003:
It's not a neutral trait, but a positive one, which means Custom Species only, and it's balanced similarly to Winged Flight in regard of "it's a trait where cost is that it's a very negligible benefit in terms of mechanics that occupies a full slot'. And I wanted to put cost at 1, but seeing Winged Flight's cost at 0 it just felt VERY WEIRD putting a trait that is kind of a lesser version of Winged Flight at a cost more expensive than, well, Winged Flight itself.

Very-Soft:
Well I think it should be a +1 at least if it's to be a thing. I don't want it to be a freebie, because I do actually want people to get hurt if they're careless and fall down, so it should be one that has a somewhat significant cost.

If it has next to no cost like this then it will just be one of those gimmie traits that everyone picks up because it has an easily recognizable mechanical advantage and prevents necessity of needing to be cautious in falling hazard zones. 

A very large part of the SD's overarching design philosopy was to design things in such a way that people would need to interact with eachother and have at least a little work to do where that was reasonable to do, and falling damage from the lattices was a conscious part of that consideration. A 0 cost soft fall will significantly damage the possibility of this interaction.

Heroman3003:
Alright, point taken and added to the cost of the trait.

PR #11775, opened 2021-11-04
new particle smasher recipes

PastelPrinceDan:
adds recipes for quartz, painite and void opal via particle smashing
(also tweaks the value of painite for cargo exports)

why:
- more recipes for science to play with, because why not 
- gives more for mining to do bc you need phoron, uranium (albeit a little bit of those if you just make a few gems), and most importantly _diamond_ which is hard-ish to get
- a way to obtain these gems in game because they're pretty :)
- i'm also actually planning on adding a use for them down the line

### DIFF
--- a/code/modules/materials/materials/gems.dm
+++ b/code/modules/materials/materials/gems.dm
@@ -67,7 +67,7 @@
 	tableslam_noise = 'sound/effects/Glasshit.ogg'
 	sheet_singular_name = "gem"
 	sheet_plural_name = "gems"
-	supply_conversion_value = 4
+	supply_conversion_value = 12 //if they take diamond to make via psmasher, they should be worth more than diamond
 
 /datum/material/void_opal
 	name = "void opal"
--- a/code/modules/power/singularity/particle_accelerator/particle_smasher.dm
+++ b/code/modules/power/singularity/particle_accelerator/particle_smasher.dm
@@ -379,3 +379,37 @@
 	required_atmos_temp_min = 3000
 	required_atmos_temp_max = 10000
 	probability = 1
+
+/datum/particle_smasher_recipe/sandstone_quartz
+	result = /obj/item/stack/material/quartz
+	required_material = /obj/item/stack/material/sandstone
+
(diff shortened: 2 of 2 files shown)


Heroman3003:
The materials were removed from obtaining and need before precisely because they were added without actually substantial use being added along with the materials. So imo just making them obtainable should come along with said use to avoid industrial expansion's repeat where we're still waiting on addition of substantial use for all the materials it added.

PastelPrinceDan:
given that coding is pretty complicated for me i'd rather see them addednow  and put in effort to make them usable later, rather than doing it all now and possibly have it denied and having it done for nothing

Very-Soft:
So what -can- these be used for now. Do they have any mechanical use at all?

PastelPrinceDan:
> So what -can- these be used for now. Do they have any mechanical use at all?

as i said in the description, i'm planning on giving them a mechanical use once they're implemented (a use in either T6 parts or other toys for science, with a tech level locked to PoI-only items so explo has more reasons to interact with science). and as i said in my reply just above your comment, coding's hard so i don't want to put effort in something _now_ if it might just be denied

Very-Soft:
I mean I don't really have a problem with these things existing, if they have some kind of use. It would be a little silly for them to exist but for the thing they're for to never get finished, or for them to exist, and then for the thing they were made to make gets declined for whatever reason. 

Like if they were at least some kind of crafting, or building material, something. I'm only hesitant to merge these because what they're intended for seems like a 'maybe I'll finish it eventually' kind of thing, without a very clear idea of what that end goal will actually look like. 

I wouldn't want them to get added, and then for the fact that they were added to be a point of 'well this thing I want to do that got added is already in, so why not use them for this turbo mega OP thing???' or whatever. Not that I expect that will happen, I just think that if they are intended to be used for making a thing, the thing they're intended to be used for should ideally come with them.

PastelPrinceDan:
hngngn fair enough yeah, i see your point. i'll close this for now and reopen it with a full use for them and such later on

PR #11777, opened 2021-11-05
i hate the engraving antichrist

Hatterhat:
to engrave things with sharp objects now you need to altclick the floor or wall, instead of "called every time you attack the floor on harm intent with a sharp object"

edit: also it engraves on help intent. still maintains altclick potential tho if you want

### DIFF
--- a/code/game/turfs/simulated/floor.dm
+++ b/code/game/turfs/simulated/floor.dm
@@ -175,3 +175,10 @@
 			to_chat(user, span("notice", "You deconstruct \the [src]."))
 			ChangeTurf(get_base_turf_by_area(src), preserve_outdoors = TRUE)
 			return TRUE
+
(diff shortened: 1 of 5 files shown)


FartMaster69420:
OH MY GOD YES

Enzo-Leon:
JFC thank you.

PastelPrinceDan:
the hero we needed

Very-Soft:
As was discussed in dev chat, I'm on board with making this not be the default action for clicking floors with a sharp object since it's often disruptive, however, I am not convinced that putting it under alt click is the best option, because it is very unlikely that anyone unfamiliar with that will actually discover the function naturally.

I PERSONALLY would prefer if it only happened on help intent, or if there was a preference toggle (which defaulted to ON) to toggle whether your clicks did graffiti or not, that way it can still be stumbled upon, but prevented if you know what to do.

PR #11780, opened 2021-11-05
Improves Physics

Heroman3003:
## About this pull request

As perfectly realistic simulator of spaceflying stationing experience, we must strive to perfection in simulation of all things related to realistic portrayal of spaceflight experience. That, obviously, first and foremost, means accurate simulation of physics. Physics is very important in both spaceflight and realism, and as someone who studied physics extensively, I feel that there is a certain area that needs improvement. I'm talking obviously, about fall impacts being absolutely unrealistic in only damaging the falling object and not the area below! So this seeks to remedy it by adding a perfectly realistic calculation, which determines the weight of relevant falling object and creates appropriate level of impact.

## Why it's good for the game

It's important to keep up with latest in physics and realism, and it's important to implement now since amount of people more affected by gravity that should cause larger falling impact than they do is rather high, it's important to satisfy our realistic needs of spaceflying simulator.

## Example

This is a wonderful example of realistic spaceflight experience, involving a heavily g-force vulnerable person descending from reasonable height with minimal resistance. This is what this Pull Request seeks to acomplish.

image

## Changelog

- People above certain mass threshold now have physical impact upon falling

## PS

~~this is literally just a meme pr that makes things explode when someone fat falls pls dont merge this~~

### DIFF
--- a/code/modules/multiz/movement.dm
+++ b/code/modules/multiz/movement.dm
@@ -506,8 +506,11 @@
 /mob/living/carbon/human/fall_impact(atom/hit_atom, damage_min, damage_max, silent, planetary)
 	if(!species?.handle_falling(src, hit_atom, damage_min, damage_max, silent, planetary))
 		..()
-		if(weight > 325)
-			explosion(get_turf(hit_atom), 2, 3, 5)
+		if(weight > 325 || size_multiplier > 1.75)
+			explosion(get_turf(hit_atom), -1, 0, 0)
+			var/turf/simulated/floor/hit_turf = get_turf(hit_atom)
+			if(istype(hit_turf))
+				hit_turf.break_tile()
 
 //Using /atom/movable instead of /obj/item because I'm not sure what all humans can pick up or wear
 /atom/movable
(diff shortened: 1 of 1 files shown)


KillianKirilenko:
~~only valid and mergeable if 200%-scale taurs have the same explosive power as maxcap ttv bombs~~

Enzo-Leon:
I need this level of comedy in my life (maybe downscale the damage to just damage a tile and to cause a small explosion sound to be played for max funnies)

lm40:
I MEAN WHY NOT DO A ZERO ZERO ZERO EXPLOSION

BigBababooey:
I entirely support this if it's scaled back to a 0/0/0 and maybe leaves a funni decal on the floor. That sounds hilarious yet not too unreasonable.

Heroman3003:
Made explosion size -1/0/0 with guaranteed damage of floor below and included the macros.

~~its still just a meme and kinda feels too shitposty~~

Very-Soft:
This would be very funny to have as an option for memes, but as a default, no

PR #11792, opened 2021-11-07
fixes body scanner not being synced with display/station time

### LINKED ISSUE #11791: Station Clocks Aren't In-Sync
The station time is different depending on what source you use.

Time synced across the board.

#### What actually happened
The Status Tab's time and the station's lil display boards are in sync, and the [time] text thingy in paperwork matches that, but Advanced Body scanners, Times of Death from medical equipment are both showing different times. This makes it a little confusing if you're attempting to do any In-Character records keeping. Times on different documents don't line up.

#### Steps to reproduce
- (Step 1)Look at me. Now look at the Status tab. Now look at the Scan Report. Now back to me. Those times are different. My first thought was it may be lag, but this happens on private servers that were just set up seconds ago as well.
- (Step 2)
- (and so on)

#### Code Revision
Server revision: B:release D:
Commit: 50b89b38017f34443fb0869e651c5cd57688afdb 

 

https://gyazo.com/2da8704c04479e9b06fb6852988b4b99
Time display and status tab both show 18:42, scan report shows 14:37 and I believe the times of death would use the same time that the scanner uses.

https://gyazo.com/79980d90983330ef7928c6065a844c35 
This was from a private server seconds after it was booted up.

PastelPrinceDan:
or at least it does on the 4 tests i did.
image

issue was mentioned in #11791 (along with health analyzers not showing the right time of death, which they seemed to do on my test server so ?????)

**does not fully fix #11791** probably because some other machines are probably out of sync too, i didn't check everything

### DIFF
--- a/code/game/machinery/adv_med.dm
+++ b/code/game/machinery/adv_med.dm
@@ -349,10 +349,10 @@
 			var/obj/item/weapon/paper/P = new /obj/item/weapon/paper(get_turf(target))
 			var/name = occupant ? occupant.name : "Unknown"
 			P.info = "<CENTER><B>Body Scan - [name]</B></CENTER><BR>"
-			P.info += "<b>Time of scan:</b> [worldtime2stationtime(world.time)]<br><br>"
+			P.info += "<b>Time of scan:</b> [stationtime2text(world.time)]<br><br>"
 			P.info += "[generate_printing_text()]"
 			P.info += "<br><br><b>Notes:</b><br>"
-			P.name = "Body Scan - [name] ([worldtime2stationtime(world.time)]"
+			P.name = "Body Scan - [name] ([stationtime2text(world.time)]"
 		else
 			return FALSE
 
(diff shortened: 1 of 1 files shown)


code/game/machinery/adv_med.dm
```diff
@@ -349,10 +349,10 @@
 			var/obj/item/weapon/paper/P = new /obj/item/weapon/paper(get_turf(target))
 			var/name = occupant ? occupant.name : "Unknown"
 			P.info = "<CENTER><B>Body Scan - [name]</B></CENTER><BR>"
-			P.info += "<b>Time of scan:</b> [worldtime2stationtime(world.time)]<br><br>"
+			P.info += "<b>Time of scan:</b> [stationtime2text(world.time)]<br><br>"
```
Heroman3003, line 352: stationtime2text doesnt' actually have arguments... I'm surprised it didn't error-out but you can just remove the argument in both cases.

PR #11797, opened 2021-11-07
Makes fuzzycuffs easier to get out of

R3dtail:
Was messing around in game with them and fuzzycuffs have the same breakout timer as normal cuffs.  They're toys, that doesn't seem 'reasonable' , that they're the same as proper handcuffs.  So this gives them a significantly reduced breakout timer.   10 seconds? Seems reasonable for toys.

### DIFF
--- a/code/game/objects/items/weapons/handcuffs.dm
+++ b/code/game/objects/items/weapons/handcuffs.dm
@@ -141,7 +141,7 @@ var/last_chew = 0
 /obj/item/weapon/handcuffs/fuzzy
 	name = "fuzzy cuffs"
 	icon_state = "fuzzycuff"
-	breakouttime = 100 //these are toys like come on
+	breakouttime = 100 //VOREstation edit
 	desc = "Use this to keep... 'prisoners' in line."
 
 /obj/item/weapon/handcuffs/cable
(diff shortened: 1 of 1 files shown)


Very-Soft (changes requested):
You'll want to do something like a 
//VOREStation Edit
comment on this in case polaris changes anything with this file, so we won't lose track of what edits are ours~

R3dtail:
> You'll want to do something like a //VOREStation Edit comment on this in case polaris changes anything with this file, so we won't lose track of what edits are ours~

Sorry, can do, I was just going off of what i was told in discord, I'm flying by the seat of my pants and don't know what I'm doing

Screemonster:
if anything you kinda _want_ a decent timer on them for a scene since it gives your partner time to notice you resisting out and do something about it while they're typing their posts though

R3dtail:
> if anything you kinda _want_ a decent timer on them for a scene since it gives your partner time to notice you resisting out and do something about it while they're typing their posts though

That's fair , I was just looking at it from the view point of if someone actually wanted them to not get out, they'd use something more significant than toy cuffs.  and cable cuffs are even more plentiful to get  and those are the go-to "use these if you have nothing else".  Uh- I just went with a low number that was less than cable cuffs.  Y'all are the smarter more familiar people, I'll do whatever number you want.  I'd think ten seconds would be enough to click on someone to move them a bit to keep from wiggling free, but that's just like my opinion.

Very-Soft:
10 seconds is still a long time to interrupt someone, so fine imo!

PR #11805, opened 2021-11-08
Simplemob QOL

Very-Soft:
Gives all simplemobs a slow passive healing as long as they have adequate nutrition. 

Also gives catslugs some new interactions. Try [help]ing them in different parts of their bodies.


Heroman3003:
I want to add that mobs also lack any way of gaining nutrition other than vore except ones with special interaction for it (Teppi, slimes).

PR #11816, opened 2021-11-11
Vore mobs can eat you while unconcious

Very-Soft:
-Only if it's enabled on the mob's AI holder
-Only if your settings are configured to allow it.

Otherwise it acts normal, doesn't bother SSD people, doesn't maul you while unconscious (unless it is set to do that, since that's an option that exists)

Important for a mob I will be adding soon.

### DIFF
--- a/code/modules/ai/ai_holder_targeting.dm
+++ b/code/modules/ai/ai_holder_targeting.dm
@@ -130,8 +130,15 @@
 			if(L.stat == UNCONSCIOUS)	// Do we have mauling? Yes? Then maul people who are sleeping but not SSD
 				if(mauling)
 					return TRUE
-				else if(unconscious_vore && L.allowmobvore) //VOREStation Add
-					return TRUE								//VOREStation Add
+				//VOREStation Add Start
+				else if(unconscious_vore && L.allowmobvore)
+					var/mob/living/simple_mob/vore/eater = holder
+					if(eater.prey_excludes[L])
+						return FALSE
+					if(eater.vore_fullness >= eater.vore_capacity)
+						return FALSE
+					return TRUE
+				//VOREStation Add End
 				else
 					return FALSE
 		if(holder.IIsAlly(L))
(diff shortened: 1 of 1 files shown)


code/modules/ai/ai_holder_targeting.dm
```diff
@@ -129,6 +130,8 @@
 			if(L.stat == UNCONSCIOUS)	// Do we have mauling? Yes? Then maul people who are sleeping but not SSD
 				if(mauling)
 					return TRUE
+				else if(unconscious_vore && L.allowmobvore) //VOREStation Add
```
Screemonster, line 133: Could this create an edge case where the mob will eat unconscious people, the prey allows mob vore, but something else prevents the mob from eating (such as being full) so it reverts to the standard mauling behaviour and claws them to death?

Very-Soft: That seems possible. I also was gonna look into what OOC escape does to make sure they don't maul people after they do that too.

Screemonster: maybe return false if will_eat() fails or something?

Razgriz1032:
What was the point of this PR? To globally disable mobs being able to eat unconscious people by changing the flag from TRUE to FALSE?

Very-Soft:
> What was the point of this PR? To globally disable mobs being able to eat unconscious people by changing the flag from TRUE to FALSE?

By default mobs can't really do anything to unconscious people unless they started doing it before they became unconscious, unless you turn mauling on, on their AI holder, which allows them to just continue to attack until the person is dead. 

This uses the same argument as mauling, but only allows them to do anything if the person's vore settings allow them to be eaten by the mob in question. (which, when an AI controlled mob tries to attack something that's fallen down, it eats them if it can)

This was added to support enemy types that could knock you out without killing you and still allow for mob vore. (Such as the space jellyfish, which primarily deal halloss damage, which usually results in unconsciousness before knocking you over.)

PR #11818, opened 2021-11-12
Fixes Gravity Failure Random Event

Enzo-Leon:
Switches the active .dme to use the old gravity.dm file so we do not need to have a gravity generator on a ship map. While we could adjust weights to ensure this never fired off without an engineer, @Very-Soft wanted to still have this fire off, but not add the risk of it remaining scuffed for the ENTIRE round if there just so happens to be 0 engineers about, or available.


Very-Soft:
THANK!! <3

PR #11820, opened 2021-11-13
Adds vorepanel save doublecheck for non-slot chars

Verkister:
Makes the vorepanel save button also do the double check it gives for simplemobs when the user is playing a character whose real_name doesn't match their client prefs, such as during body swap shenanigans etc.

### DIFF
--- a/code/modules/vore/eating/vorepanel_vr.dm
+++ b/code/modules/vore/eating/vorepanel_vr.dm
@@ -304,7 +304,7 @@
 			return set_attr(usr, params)
 
 		if("saveprefs")
-			if(!ishuman(host) && !issilicon(host))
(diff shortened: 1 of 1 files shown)


code/modules/vore/eating/vorepanel_vr.dm
```diff
@@ -304,7 +304,7 @@
 			return set_attr(usr, params)
 
 		if("saveprefs")
-			if(!ishuman(host) && !issilicon(host))
+			if(host.real_name != host.client.prefs.real_name || !ishuman(host) && !issilicon(host))
```
Arokha, line 307:
```suggestion
			if(host.real_name != host.client.prefs.real_name || (!ishuman(host) && !issilicon(host)))
```

Just adding parens around the latter part. I think it would be fine without (&& binds stronger than || according to http://www.byond.com/docs/ref/#/operator) but still.

PR #11826, opened 2021-11-14
[MIRROR] /atom New() => Initialize() [MDB IGNORE]

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/8298
--------------------
WIP, DNM, opening for CI and diff.

Apologies for the giant diff, by nature of the problem this had to be done all at once to avoid introducing ordering and dependency bugs. I imagine this will need some discussion as philosophically Virgo seems opposed to use of Initialize over New and this will be a serious compatibility issue going forward if merged here and not downstream,

Replaces `New()` with `Initialize()` across the board. New() overrides frontload init time onto world/New and are more or less uncontrolled with regards to how and when an atom New()s, while Initialize() is tick-checked, handled by a subsystem, and is able to return a useful value (init hints). Init time for the server appears to be exactly the same as prior to this PR (more or less 110 seconds on my machine) with the major init time hog being random map generation.

Some priority-init atoms have been given INITIALIZE_IMMEDIATE().

The following New() overrides have been left alone:
- /atom/New, as it calls Initialize for post-SSatoms-init instances.
- /obj/effect/projectile, due to the complexity of how it is created and the use of loc as an additional arg for init.

**TODO**
- [X] Initial conversion/cleanup.
- [X] Work out why turfs and areas don't Initialize().
- [x] Fix pipes not hiding under their turfs.
- [x] Kill remaining runtimes.
- [x] Fix shuttle init ordering to be post-SSatoms init so they can actually find their docking controllers etc.
- [x] Proper shakedown testing.


KillianKirilenko:
I see the linter/CI has just straight up rebelled and is refusing to run this time around.

code/game/objects/items/weapons/material/twohanded.dm.orig
```diff
@@ -0,0 +1,224 @@
+/* Two-handed Weapons
```
Novacat, line 1: This file should be deleted.

Novacat (dismissed):
Not enough time to go through everything but it seems the issue is that old Vorestation-specific Initialize procs need to be updated.

KillianKirilenko:
Okay, I think something in the linter just needs changing at this point. @Atermonera did you guys change anything with the linters when you did your version of this PR, or around that time? So far as I can tell the code that's setting off the linters is mostly-identical (Polaris uses `sleep` where we do) but your linter clearly isn't taking offense to that whereas ours *is*.

This needs fixing sooner rather than later, and I'm running out of ideas on where to turn to untangle this mess.

Novacat:
Seems like the remaining issue are some atoms not being Initialized.

Novacat (approved):

Heroman3003:
I'm surprised this thing wasn't just closed way back when we were closing all the other pointless mirror refactors. Really should've been tbh.

PR #11841, opened 2021-11-16
Somatoray Modularity

FartMaster69420:
A buff for xenobotany, this makes the Floral Somatoray upgradable with micro laser stock parts.

By default, a Somatoray is now printed with a t1 laser installed, as a laser is now required for the weapon to fire.

Previously the chance of a mutation was 20% per shot, this changes it to be 15 multiplied by the tier of laser.

This leads to the following chances of mutation:

Somatoray w/ T1 Laser: 15%
Somatoray w/ T2 Laser: 30%
Somatoray w/ T3 Laser: 45%
Somatoray w/ T4 Laser: 60%
Somatoray w/ T5 Laser: 75%

The idea here is to make life less tedious and allow for mutations of plants a bit faster so it wont take several hours to get good genes.


FartMaster69420:
Woo, finally got the checks to pass.

PR #11848, opened 2021-11-18
EVENTS AND LANDMARKS - OH BOY MORE COMEDY

Enzo-Leon:
The following PR will do a FEW things

* Adds a new Jellyfish Migration Event
* Updated Vermin Infestation to now spawn via landmarks vs. areas and airvents (This is to allow it so there are more spots and unique areas we can drop the critters at)
* Updated Carp Migration & Rogue Drone event numbers, and also added landmarks around
* Updated Drop Pod Event with new locations and updated the flavor text of the event
* Disabled Meteor Strike RNG event as it didn't make much sense with the current set-up

### DIFF
--- a/code/game/objects/effects/landmarks.dm
+++ b/code/game/objects/effects/landmarks.dm
@@ -61,10 +61,6 @@
 			prisonsecuritywarp += loc
 			delete_me = 1
 			return
-		if("blobstart")
-			blobstart += loc
-			delete_me = 1
-			return
 		if("xeno_spawn")
 			xeno_spawn += loc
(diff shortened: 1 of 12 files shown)


Heroman3003:
About drone pod drop landmarks: Are they given sufficent distance from ship itself? Hard to gauge just looking at mapdiff with all landmarks being same.

Enzo-Leon:
Yes. I did a test on all of them and the MOST damage they'll do is take out one of the PDW turrets, but none of the ships actual walls or power lines to the facility will be harmed.

PR #11856, opened 2021-11-19
Fixes boosted quantum pads not using extra power.

Heroman3003:
It also was throwing out a warning on compile


Hatterhat:
i also threw a fix for this in the KAs PR but this is good too

Very-Soft:
Already fixed by another PR

PR #11860, opened 2021-11-20
Fixes disappearing morphs

### LINKED ISSUE #11846: Morphs can turn permanently invisible
Morphs become permanently invisible after resting while disguised as a mouse

To remain visible

#### What actually happened
It became invisible

#### Steps to reproduce
- Be morph
- Shift-click on mouse
- Rest
- Look on in awe as you are now permanently invisible

Heroman3003:
Fixes #11846

Might not be the only way to trigger this bug, but this fixes the way its triggered in the issue mentioned.

### DIFF
--- a/code/modules/mob/living/simple_mob/subtypes/vore/morph/morph.dm
+++ b/code/modules/mob/living/simple_mob/subtypes/vore/morph/morph.dm
@@ -9,6 +9,7 @@
 	icon_state = "morph"
 	icon_living = "morph"
 	icon_dead = "morph_dead"
+	icon_rest = null
 	movement_cooldown = 1
 	status_flags = CANPUSH
 	pass_flags = PASSTABLE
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/simple_mob/subtypes/vore/morph/morph.dm
```diff
@@ -188,6 +189,14 @@
...
 	return ..()
 
+/mob/living/simple_mob/vore/hostile/morph/lay_down()
+	if(morphed)
+		var/temp_state = icon_state
+		..()
+		icon_state = temp_state
+	else
+		..()
+
```
Arokha, line 199: Why do this instead of just if(morphed) return?
Honest question: Are there cases where you need to 'get up' as a morph while morphed so you need it to modify if they're laying or not or what

Arokha (approved):
He replied on discord

PR #11882, opened 2021-11-21
Fixes mobs dying twice on gib

### LINKED ISSUE #11880: Broodmother Spiders - Xzibit Meme or Twisted Surprise Birthday Stripper Prank?
Broodmother spider corpses, when butchered for meat, explode into a (second) burst of hostile spiders and drop a (second) Royal Egg.

Butchering the dead body of a Broodmother Spider to yield only meat and gore, not hot venomous death for everyone nearby.

#### What actually happened
Applying a cutting tool (knife, etc.) to butcher the corpse for meat causes the Broodmother Spider's death event where it spawns a small wave of hostile spiders to occur a second time when it really shouldn't.

#### Steps to reproduce
- Fight and defeat a Broodmother Spider
- Defeat the first wave of on-death spiders, discard the Royal Egg since we don't need it for this
- Apply butchering tool to dead body of Broodmother Spider
- Fight and (hopefully) defeat the second wave of on-death spiders
- Enjoy having two Royal Eggs.

Heroman3003:
Fixes #11880

### DIFF
--- a/code/modules/mob/death.dm
+++ b/code/modules/mob/death.dm
@@ -1,7 +1,8 @@
 //This is the proc for gibbing a mob. Cannot gib ghosts.
 //added different sort of gibs and animations. N
 /mob/proc/gib(anim="gibbed-m", do_gibs, gib_file = 'icons/mob/mob.dmi')
-	death(1)
+	if(stat != DEAD)
+		death(1)
 	transforming = 1
 	canmove = 0
 	icon = null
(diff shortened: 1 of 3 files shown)


FartMaster69420:
I dunno, this bug is kinda funny though.

Very-Soft:
Just want to say that the broodmother isn't the only thing with this issue. Anything that drops loot on death drops it again when gibbed, so that core issue should probably be fixed, rather than just slapping it on the broodmother. 

(I mostly know this because I ran into the same issue with the teppi and their horns. ORIGINALLY they dropped 2 horns on death, but then when you gib them they drop horns again, and I thought that was weird, but trivial to adjust.)

Heroman3003:
Done. Now all mobs only 'die' once. Also adds extra horn to teppi's butchery list because when I tried to add it to the loot list it kinda didn't work.

PR #11888, opened 2021-11-21
Adds contained vines spawner and a POI that uses the spawner

SleepingZi:
adds a landmark that immediately spawns space vines and deletes its self. 
Also adds a new V4 POI that uses this new landmark
image
this will be modifed a tiny bit before its done

### DIFF
--- a/code/game/objects/effects/landmarks.dm
+++ b/code/game/objects/effects/landmarks.dm
@@ -82,6 +82,11 @@
 			vinestart += loc
 			delete_me = 1
 			return
+		if("containedvinesspawner")
+			Containedvines += loc
+			delete_me = 1
+			spacevine_containment()
+			return
 		//VORE Station Add End
 
 	landmarks_list += src
--- a/code/game/objects/effects/landmarks.dm
+++ b/code/game/objects/effects/landmarks.dm
@@ -86,7 +86,7 @@
 			Containedvines += loc
 			delete_me = 1
 			spacevine_containment()
-			return
+			return INITIALIZE_HINT_LATELOAD
 		//VORE Station Add End
 
 	landmarks_list += src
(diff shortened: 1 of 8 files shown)


Heroman3003:
Seems like there's some init order issues as it first runtimes when it fails to generate random seed (seemingly due to trying to do so before plants subsystem is initialized) and then seemingly failing to try and process neighboring tiles due to the fact that generation of seed was not complete.

SleepingZi:
> Seems like there's some init order issues as it first runtimes when it fails to generate random seed (seemingly due to trying to do so before plants subsystem is initialized) and then seemingly failing to try and process neighboring tiles due to the fact that generation of seed was not complete.

oh dear, yeah that's what I'm getting when I try to boot it, heck. any clue on how to fix this? I'm thinking of seeing if I can execute it on round start, but I have no clue how to do that.

Heroman3003:
> > Seems like there's some init order issues as it first runtimes when it fails to generate random seed (seemingly due to trying to do so before plants subsystem is initialized) and then seemingly failing to try and process neighboring tiles due to the fact that generation of seed was not complete.
> 
> oh dear, yeah that's what I'm getting when I try to boot it, heck. any clue on how to fix this? I'm thinking of seeing if I can execute it on round start, but I have no clue how to do that.

I'm not too sure off the top of my head, but consider using LateInitialize for the landmark or whatever that proc is called?

Very-Soft:
So is this working correctly now or no?

SleepingZi:
> So is this working correctly now or no?

I wanna test a few more things before I push this because I think it might cause problems if pushed now

SleepingZi:
alright what I did there seems to have fixed the issue on my end, if someone can restart the dreamchecker that would be nice

Heroman3003:
> alright what I did there seems to have fixed the issue on my end, if someone can restart the dreamchecker that would be nice

Dream Checker or Linter failing means there's something wrong with coding standard not the functionality of it. It's listed out in the filediff like a review would. This time it seems like something about return with a value being put in spawn

Heroman3003:
I think making removing one tabulation in return's line should fix issues it has.

PR #11889, opened 2021-11-21
Mice are Nice

GhostActual:
Added a rest sprite to mice.

Adds custom `on_head` and `in_hand` as well as `on_ear` sprites for all mouses.

This was painful to make.


Heroman3003:
Have you actually tested the morph part of it? Because reason for their disappearing has nothing to do with mouse sprites, and its related to morphing code. I already have a fix up for that, awaiting review.

GhostActual:
> Have you actually tested the morph part of it? Because reason for their disappearing has nothing to do with mouse sprites, and its related to morphing code. I already have a fix up for that, awaiting review.

Fixed it

PR #11895, opened 2021-11-21
Makes Chocolate Bar recipies require 5 Enzyme catalyst

### LINKED ISSUE #10913: Recipe conflict for chocolate bars & choccy milk
image
image

Tested ingame as well.

Heroman3003:
Fixes #10913

Generally it shouldn't be that drastic of a change, enzyme is easily available in kitchen and most chemical recipies that produce item product in cooking already use it.


lm40:
rip chocolate bar plants

PR #11913, opened 2021-11-23
Active Edge Hunt Part 5 - Revelation

Heroman3003:
The true active edges was mistakes I made all along... Okay not really but also kinda. But hey at least I combed through every niche case imaginable before realizing I was using **wrong PoI files**. So they definitely shouldn't ever bother us again! R-Right...?


Zeldazackman:
There's a simple phrase that comes to mind when it comes with active edges.

"The ride never ends"

PR #11918, opened 2021-11-23
More appropriate stack digestion

Verkister:
Makes digested stack items use one sheet/piece/whatever when digested and resets the stack item "health" for the digestion process to rinse and repeat until the stack is depleted rather than deleting it as a single sheet size food item regardless of the stack amount.


Heroman3003:
I think we've had it so that linter will fail PRs that edit amount in stacks directly instead of using stack procs.

PR #11922, opened 2021-11-24
Swaps Yes and No for give popup

### LINKED ISSUE #11915: Yes and No options on popup window for "Give" prompts are swapped
Simply put, most other yes/no prompts have Yes on the left and No on the right, so this is possibly erroneous placement of the buttons. 

I go to click yes to accept on the left, where that normally is placed, and do not see it.

#### What actually happened
I decline the attempt to give an item, due to the swapped placement

#### Steps to reproduce
- Player attempts to give you an item
- You go to click yes, but click no instead because of the swapped buttons, looking like an idiot
- Be mildly inconvenienced

- There's a nonzero chance this isn't _actually_ a bug and was intentional placement, but I myself am consistently confused by it, and I see fairly often people get confused and gaslit by the abnormal button placement. So if it's intentional, but can be switched back without causing problems, that would be swell.

Heroman3003:
Fixes #11915

### DIFF
--- a/code/modules/mob/living/carbon/give.dm
+++ b/code/modules/mob/living/carbon/give.dm
@@ -20,7 +20,7 @@
 
 	usr.visible_message(SPAN_NOTICE("\The [usr] holds out \the [I] to \the [target]."), SPAN_NOTICE("You hold out \the [I] to \the [target], waiting for them to accept it."))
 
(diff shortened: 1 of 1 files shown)


Arokha:
The reason for having it that way (as explained in the issue) is still there, though I don't know if anyone cares too much about that reason anymore. Will poke staff to check.

Very-Soft:
I think this is fine, but I will give it time for others to weigh in

Novacat:
Seems fine to me

PR #11944, opened 2021-12-01
Makes NIFs not get automatically lost (hopefully)

Heroman3003:
~~Makes it so that instead of erasing NIF from savefile and re-saving them later, NIFs only get their savefile durability reduced. This is still minorly punishing for not backing up, but not nearly as harshly as full on deletion. Eases job for admins too.~~

~~Properly backed up people will have these durability changes overridden anyway, so that won't cause issues, and if you leave round after extracting a NIF, well, it will also be overridden with nothingness. So this should only affect situations where NIF just disappears.~~

Adjusted based on feedback. Now NIFs simply saves whenever durability changes, upwards or downwards. Saving from backup implants and end of round happens still too.

### DIFF
--- a/code/modules/client/preference_setup/vore/08_nif.dm
+++ b/code/modules/client/preference_setup/vore/08_nif.dm
@@ -38,11 +38,17 @@
 	if((character.type == /mob/living/carbon/human) && ispath(pref.nif_path) && pref.nif_durability)
 		new pref.nif_path(character,pref.nif_durability,pref.nif_savedata)
 
+		/*
 		//And now here's the trick. We wipe these so that if they die, they lose the NIF.
 		//Backup implants will start saving this again periodically, and so will cryo'ing out.
 		pref.nif_path = null
 		pref.nif_durability = null
 		pref.nif_savedata = null
+		*/
+		//No we do not, that's lame and admins have to re-NIF them later.
+		//If they leave round or get their NIF extracted, it will save as 'gone' anyway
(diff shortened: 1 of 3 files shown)


Arokha:
This makes it so people who don't get an implant have infinite durability between rounds, doesn't it?

Heroman3003:
> This makes it so people who don't get an implant have infinite durability between rounds, doesn't it? EDIT: As in, there's nothing to update the savefile with durability after installing software, if they don't have an implant

NIF is still saved by those people when round ends or they use cryo (or equivalent). Unless they specifically somehow every round die without having a backup, they cannot keep avoiding it saving. Also, that's why the new punishment for 'dying without implant' is loss of about 10-15 durability, its specifically to prevent infinite NIF use being a possible exploit.

Arokha:
> Unless they specifically somehow every round die without having a backup, they cannot keep avoiding it saving

If this were another server I'd consider that rare but that's the goal of some people here, lol.

It should probably save whenever durability changes, honestly, instead of just at certain times.

Very-Soft (changes requested):
Pls do the thing Aro suggested. It should just save durability when it changes tbh!

Heroman3003:
Done. NIF now should save every time durability changes (assuming its inside a human when it happens).

Also adds unified proc for repair of NIF to facillitate that; and removes the 'alternate punishment' (along with punishments altogether).

PR #11946, opened 2021-12-02
Fixes vertical noise spread penetrating floors

Heroman3003:
No more hearing rafterspeak on first deck

### DIFF
--- a/code/_helpers/game.dm
+++ b/code/_helpers/game.dm
@@ -279,7 +279,7 @@
 		var/turf/U = GetBelow(O)
 		while(istype(U))
 			hearturfs |= U
-			if(istype(U, /turf/simulated/open))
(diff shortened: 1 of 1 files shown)


Arokha:
This is intentional though so you can talk down at people who talk up at you...

Heroman3003:
> This is intentional though so you can talk down at people who talk up at you...

yes, and that still works. Issue with original implementation though was that if, say, like on Stellar Delight you have three floors, with bottom floor separated from middle floor entirely ceiling-wise, and second and third floor having open space connection, then talking on third floor made you be heard THROUGH the ceiling of first floor. It basically meant that literally everyone under you could hear you all the way down, even if there was floor between you and them at some point. This just makes it so that floors actually stop spread of sound, while open space still sends it down.

PR #11950, opened 2021-12-03
[DNM]Explains how to ring the doorbell take two

R3dtail:
I have no idea how to use github whatsoever this is just me trying to add some explanations on how to use certain things in the game that as far as I can see aren't explained anywhere else.

### DIFF
--- a/code/game/machinery/doors/door.dm
+++ b/code/game/machinery/doors/door.dm
@@ -3,7 +3,7 @@
 
 /obj/machinery/door
 	name = "Door"
-	desc = "It opens and closes."
+	desc = "It opens and closes.  There's a little doorbell on the side that you can ring by holding Alt and Left-clicking the door" //Vorestation edit
 	icon = 'icons/obj/doors/Doorint.dmi'
 	icon_state = "door1"
 	anchored = TRUE


Heroman3003:
I think this should only apply to airlocks, not doors in general, since only airlocks have the 'knocking' functionality.

R3dtail:
> I think this should only apply to airlocks, not doors in general, since only airlocks have the 'knocking' functionality.

I admit that's the ONLY thing i found that had the description that matches airlocks.  Because airlocks.dm doesn't...[have] a desc variable and that's way outside of my comfort zone to go messing with that because i only have the barest understanding of what i'm doing.

Very-Soft:
you can set a desc variable on anything. The airlocks one doesn't have one because it uses the same one as the doors. If you add one to the airlocks, then it will use what you write there instead~

KillianKirilenko:
Good idea!

I will say, there's also a couple other types of desc that you can use - IIRC fluff_desc and antag_desc, the latter only appearing to people with antag flags. Mostly I mention this because you could use fluff_desc (or whatever it is) to make it stand out a little more and be more noticeable. Similar to how examining frames or walls will tell you what you need to do next to assemble/disassemble.

R3dtail:
> Good idea!
> 
> I will say, there's also a couple other types of desc that you can use - IIRC fluff_desc and antag_desc, the latter only appearing to people with antag flags. Mostly I mention this because you could use fluff_desc (or whatever it is) to make it stand out a little more and be more noticeable. Similar to how examining frames or walls will tell you what you need to do next to assemble/disassemble.

so that's what that is that's pog okay thank you that's definitely far more what i wanted to do.

R3dtail:
Either way don't- don't merge this yet because haha i'm really wanting to try and get this to look (proper)

PR #11957, opened 2021-12-04
Reworks trash pile pool chances

Heroman3003:
This is definitely far from 'complete' version and will probably go some revisions based on staff and player feedback, but, well, it had to be done at some point. Main intent of PR is to finally cut away all the 'useless' stuff from the trash pile pools, making them less of a 'losable' lottery, to accomodate map layouts that use few sparing trash piles better (such as SD).

Currently it removes a lot of Alpha pool items, while reworking meaning of pools themselves.

Alpha pool is now meant to be 'legal, practical, minor' items pool. It contains stuff like power cells, broken emags, welding masks, and on more rare end, pizza vouchers, cabin capsule, THE GLOVES.
Beta pool is now meant to be 'scene-focused' items pool. It contains stuff like sizeguns, fuzzy cuffs, amorphorovir vials, sleevemate, bodysnatcher, etc. Also includes latex maid dress and HYPER jumpsuit.
Original Gamma pool as it was before is untouched, but is now renamed into Delta pool. Works same way, contains same stuff, no change.
New Gamma pool is illegal stuff. Knives, regular cuffs, drugs, syndie devices fall into that pool.

Pool chances are: 50 for Alpha, 30 for Beta, 19 for Gamma and 1 for Delta.

Ultimately, primary intent is for trash piles to be, ironically, less trashy and more reliable in the smaller amounts they are present now without removing RNG chances or adding pity timers or whatnot. Just culling the useless junk like clothes, very generic and easily producable items, etc.

### DIFF
--- a/code/game/objects/structures/trash_pile_vr.dm
+++ b/code/game/objects/structures/trash_pile_vr.dm
@@ -11,13 +11,15 @@
 
 	var/obj/structure/mob_spawner/mouse_nest/mouse_nest = null
 
-	var/chance_alpha	= 79	// Alpha list is junk items and normal random stuff.
-	var/chance_beta		= 20	// Beta list is actually maybe some useful illegal items. If it's not alpha or gamma, it's beta.
-	var/chance_gamma	= 1		// Gamma list is unique items only, and will only spawn one of each. This is a sub-chance of beta chance.
+	var/chance_alpha	= 50	// Alpha list is minor legal stuff. May be useful, but not particularly so.
+	var/chance_beta		= 30	// Beta list is primarily items used to enable scenes in one way or another.
+	var/chance_gamma	= 19	// Gamma list is illegal mechanical items. All the cool gamer loot is stored here.
+	var/chance_delta	= 1		// Delta list is unique items that will be spawned only once per round.
+								// The REAL cool gamer loot is in here. Mixed bag between exceptionally practical items that would fit in Beta or Gamma pools.
 
 	//These are types that can only spawn once, and then will be removed from this list.
 	//Alpha and beta lists are in their respective procs.
-	var/global/list/unique_gamma = list(
+	var/global/list/unique_delta = list(
 		/obj/item/device/perfect_tele,
 		/obj/item/weapon/bluespace_harpoon,
 		/obj/item/clothing/glasses/thermal/syndi,
@@ -27,7 +29,7 @@
 		/obj/item/weapon/gun/projectile/dartgun
 		)
 
-	var/global/list/allocated_gamma = list()
+	var/global/list/allocated_delta = list()
 
 /obj/structure/trash_pile/Initialize()
 	. = ..()
@@ -52,12 +54,12 @@
 
 /obj/structure/trash_pile/attackby(obj/item/W as obj, mob/user as mob)
 	var/w_type = W.type
-	if(w_type in allocated_gamma)
+	if(w_type in allocated_delta)
 		to_chat(user,"<span class='notice'>You feel \the [W] slip from your hand, and disappear into the trash pile.</span>")
 		user.unEquip(W)
 		W.forceMove(src)
-		allocated_gamma -= w_type
-		unique_gamma += w_type
+		allocated_delta -= w_type
+		unique_delta += w_type
 		qdel(W)
 
 	else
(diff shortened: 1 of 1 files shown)


Very-Soft:
I don't like the way the different pools are weighted, and I don't like a lot of the things you removed from the lists.

Heroman3003:
Most of removed things fall into category of being extremely trivial to obtain items, or minor clothing, or basically some loadout stuff, all of which I think isn't that important to most players. I did consider additional option of "trash pile can be searched a few times, one time giving item from the proposed pools with proposed chances, and all the other times giving trivial 'useless' items such as baseballs, clothes, broken components, minor things, etc". But I thought that'd be a counterintuitive addition relative to ways trash piles work right now and would take a while for playerbase to adapt.

Heroman3003:
While this wasn't intentional result while I was pruning the lists, I ran some numbers now and this proposed listing I made works out somewhat nicely if we talk about pure chance/balance stuff of items kept and the likelyhood of overall finding 'any of those'

Tether had 62 trash piles in final revision, while Stellar Delight has 19.  That's about 30% of Tether's trash pile amount.

Now if we take individual chances of all items that this PR kept in the pools as they are right now on live, include pool chances in the equation, and add them up, we get that we get that the chance those any of those specific items would get found.... about 30% of the time. So technically this does not making 'finding one of the items kept' more likely. Now proportions those items have within said suggested pools are undoubtedly way, way screwier and going over them individually is way too complicated, but I just wanted to point out that this does not make 'finding these cool things' on the overall of items listed particularly highly likelier, relative to Tether's trash pile amount.

code/game/objects/structures/trash_pile_vr.dm
```diff
@@ -114,166 +116,98 @@
...
+					prob(1);/obj/item/weapon/storage/box/survival/space,
+					prob(4);/obj/item/device/pda,
+					prob(4);/obj/item/device/radio/headset,
+					prob(2);/obj/item/device/paicard,
+					prob(3);/obj/item/weapon/storage/box/sinpockets,
+					prob(2);/obj/item/weapon/storage/pill_bottle/paracetamol,
+					prob(1);/obj/item/pizzavoucher,
+					prob(2);/obj/item/device/nif/bad,
+					prob(2);/obj/item/weapon/card/emag_broken,
+					prob(1);/obj/item/device/survivalcapsule/popcabin)
```
Novacat, line 153: This should probably be a beta item.

code/game/objects/structures/trash_pile_vr.dm
```diff
@@ -114,166 +116,98 @@
...
-					prob(1);/obj/item/device/flashlight/pen,
-					prob(1);/obj/item/device/paicard,
-					prob(1);/obj/item/weapon/card/emag,
-					prob(1);/obj/item/clothing/mask/gas/voice,
-					prob(1);/obj/item/weapon/spacecash/c100,
-					prob(1);/obj/item/weapon/spacecash/c50,
 					prob(1);/obj/item/weapon/storage/backpack/dufflebag/syndie,
-					prob(1);/obj/item/weapon/storage/box/cups,
-					prob(1);/obj/item/pizzavoucher)
+					prob(1);/obj/item/clothing/suit/storage/vest/heavy/merc,
```
Novacat, line 143: This should probably be a gamma item

code/game/objects/structures/trash_pile_vr.dm
```diff
@@ -114,166 +116,98 @@
...
 	var/obj/item/I = new path()
 	return I
 
 /obj/structure/trash_pile/proc/produce_gamma_item()
-	var/path = pick_n_take(unique_gamma)
+	var/path = pick(prob(4);/obj/item/weapon/material/butterfly,
+					prob(4);/obj/item/weapon/material/butterfly/switchblade,
+					prob(4);/obj/item/clothing/gloves/knuckledusters,
+					prob(2);/obj/item/weapon/material/knife/tacknife,
+					prob(2);/obj/item/clothing/shoes/syndigaloshes,
```
Novacat, line 179: This can probably be bumped up to Alpha, since if I recall correctly these are just brown shoes which are anti-slip.

Novacat (changes requested):

Very-Soft:
I'll look over and test this more soon.

Very-Soft:
I just really don't like this and want to think about a better solution, especially since we're starting a kind of faux rotation.

PR #11958, opened 2021-12-04
Makes Bluespace Harpoon capable of televore

Heroman3003:
Title. In both modes of operation, makes BSH capable of comitting vore crimes/happy accidents.

~~Currently untested because I don't have ability to do so right now, but I want to put it up to not lose track of this. I will remove [UNTESTED] from the name once I made sure all parts of this work as they should.~~

Okay, I tested it and fixed a pair of minor hiccups, it should be all ready to go now.


Heroman3003:
Tested, fixed, ready.

PR #11967, opened 2021-12-04
Replaces wish granter with old 'dark' version; removes antag role from it

Heroman3003:
Legacy code


Screemonster:
I love how this replaces the old "free antag" option. And I love what it replaces it with.

PR #11970, opened 2021-12-05
Makes holodeck turn power on/off depending on state

### LINKED ISSUE #11961: Holodeck Lacks Power
The holodeck room does not have an APC.

To be able to install a jukeboxx into the holodeck so we can have music while playing chess.

#### What actually happened
Installation of a jukebox failed because there was nothing to power it.

#### Steps to reproduce
- Be in the Deck 3 Holodeck
- Put down a machine that uses APC power
- Powerless

- There is one tile along the north edge that is recessed and has an air alarm/sink/light, but this one tile does not share an area with the main room so jukebox music does not travel to it
- A workaround in-game would be to use a portable boombox linked to a jukebox elsewhere, but eh.

Heroman3003:
Fixes #11961 

It's a holographic projection anyway, it makes sense if its powered.

### DIFF
--- a/code/game/area/Space Station 13 areas.dm
+++ b/code/game/area/Space Station 13 areas.dm
@@ -1311,6 +1311,7 @@ NOTE: there are two lists of areas in the end of this file: centcom and station 
 	name = "\improper Holodeck"
 	icon_state = "Holodeck"
 	dynamic_lighting = 0
+	requires_power = 0
 	sound_env = LARGE_ENCLOSED
 	forbid_events = TRUE
 	flags = AREA_FLAG_IS_NOT_PERSISTENT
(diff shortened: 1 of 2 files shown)


Very-Soft:
Does this make the place always have power no matter what, or only when the holodeck is on?

I like the latter, I don't like the former.

Heroman3003:
Currently, it makes it always, though it's I doubt it can be abused that hard. I'm not sure if its possible to set up so that it only has power if powered, I'd need to look into how it handles areas in general. I'll look into it further tomorrow.

Heroman3003:
Fixed, now it turns power off when 'off' mode of holodeck is chosen, but keeps it powered if any other setting is active.

PR #11975, opened 2021-12-05
Door Interact Tweaks

KillianKirilenko:
A little update to #11594 after giving it some thought. My initial PR was a bit shortsighted, so I've made two changes;

1. Knocking now only occurs if you're on HELP intent. DISARM and GRAB cause nothing to happen. HARM is still hammering on doors.
2. This code no longer intercepts/prevents the "view tile contents in a tab" functionality, so you can grab stuff under closed doors without using the rightclick menu.

On the one hand I think it's kinda silly that anything on the same tile as a door is basically in a superposition where it can be grabbed from either side, *but* IMO doors should also be like rimworld and be held open by larger items rather than whatever you call the current behaviour. Smaller items (wrappers, bottles, cans, etc.) should be crushed or just launched to one side, larger ones should cause the door to ping back open after it tries to close, similar to how they won't close on mobs.

Not that I can be bothered to code anything like that right now. So this is kind of a stopgap and an "oops, I overlooked something" I guess.

### DIFF
--- a/code/game/machinery/doors/airlock.dm
+++ b/code/game/machinery/doors/airlock.dm
@@ -979,6 +979,7 @@ About the new airlock wires panel:
 	return
 
 /obj/machinery/door/airlock/AltClick(mob/user as mob)
+	. = ..()
 	user.setClickCooldown(DEFAULT_ATTACK_COOLDOWN)
 	if(!Adjacent(user))
 		return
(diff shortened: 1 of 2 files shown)


Very-Soft:
Not that I'm particularly bothered by it, but I wonder why you have changed it so that it doesn't doorbell on grab/disarm?

I don't think there's any harm in it doing it on those ones, and it doing nothing sometimes might make folk who do it somewhat confused if they expected something to happen, but actually nothing happened due to their intent.

KillianKirilenko:
I wanted to leave them there as options for opening the tab to grab tile contents *without* pinging the door, if people didn't want to for whatever reason.

Very-Soft:
Fair enough

PR #11976, opened 2021-12-06
Standartizes names and clarifies descriptions of Preferences Tab verbs

Heroman3003:
Makes all toggles actually called toggle, clarifies and grammarizes some of descriptions.

### DIFF
--- a/code/modules/client/preferences_toggle_procs.dm
+++ b/code/modules/client/preferences_toggle_procs.dm
@@ -150,15 +150,15 @@
 	feedback_add_details("admin_verb","TLobMusic") //If you are copy-pasting this, ensure the 2nd parameter is unique to the new proc!
 
 /client/verb/toggle_admin_midis()
-	set name = "Toggle Admin MIDIs"
+	set name = "Toggle Admin Music"
 	set category = "Preferences"
-	set desc = "Toggles the ability to hear the music in the lobby."
+	set desc = "Toggles the ability to hear music played by admins."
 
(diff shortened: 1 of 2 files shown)


code/modules/client/preferences_toggle_procs.dm
```diff
@@ -152,7 +152,7 @@
 /client/verb/toggle_admin_midis()
 	set name = "Toggle Admin MIDIs"
 	set category = "Preferences"
-	set desc = "Toggles the music in the lobby."
+	set desc = "Toggles the ability to hear the music in the lobby."
```
KillianKirilenko, line 155: Name and desc should probably be changed wholesale. Admins can do a lot more than play midi files nowadays, and it sure as hell ain't just lobby music.

Heroman3003: I somehow missed that.

code/modules/client/preferences_toggle_procs.dm
```diff
@@ -382,7 +382,7 @@
 /client/verb/toggle_status_indicators()
 	set name = "Toggle Status Indicators"
 	set category = "Preferences"
-	set desc = "Enable/Disable seeing status indicators over peoples' heads."
+	set desc = "Toggles seeing status indicators over people's heads."
```
KillianKirilenko, line 385: `Peoples'` is correct here, because it refers to status indicators belonging to people. Plural possessives are fucky, welcome to English.

PR #11984, opened 2021-12-07
Replaces maint doors by solars with regular type

Heroman3003:
They had engineering restriction despite the fact that section of maint they lead to is already publically accessible by just entering through northern maint entrance and heading there by maint itself. Seems kinda redundant as restriction.


Screemonster:
I think the reasoning behind that access restriction is so that if some dumbass opens both doors of the airlock at once there's at least _one_ section of maintenance before you can open directly onto a public hallway.

Very-Soft:
It is the way it is, to make it less easy to suck all the air out of the main hallways. You can still do it, but it takes more on purpose effort to achieve if you don't have the appropriate access.

Issue #11995, opened 2021-12-09
Lemon Seeds icon is missing


Heroman3003:
#### What you expected to happen
Lemon Seeds icon

#### What actually happened
Generic question mark seed icon

#### Steps to reproduce
- Vend yourself some lemon seeds

PR #11996, opened 2021-12-09
Lemmon Seed Fix

GhostActual:
Fixes https://github.com/VOREStation/VOREStation/issues/11995


Heroman3003:
Lemon product icon exists and functions. It appears its used because regular 'treefruit' icon is too dark to be yellow-colored. What its missing though, is a seed icon.

GhostActual:
> 

It was pointing to the wrong sprite, there was no `lemon` seed sprite in the `.dmi` there was only the `treefruit`. I dunno what happened to the icon, but it didn't exist in the code so I ripped the one from TG and put it in. From my tests, it works alright now.

PR #12011, opened 2021-12-15
Fixes pupdozer overlay anims being a total disaster

Verkister:
I don't even know how it is even possible for someone to mess up like this after getting so far with them


GhostActual:
> I don't even know how it is even possible for someone to mess up like this after getting so far with them

To answer this, my only response is "Byond fuckery"

Verkister:
Closing because I made more fixes and didn't notice this hadn't even been merged yet to avoid conflicts.

PR #12015, opened 2021-12-16
more stuff that can fit on toolbelts of holding

PastelPrinceDan:
big belt = some big stuff

lets toolbelts of holding hold
- inducers
- plasma cutters bc engineering projects are !!fun!! + miners use toolbelts too 
- light replacers
- stacks of steel & glass (only those because they're the most commonly used by engi)


Heroman3003:
I don't think any mining tool needs to fit on toolbelt, mining exclusive or not. Plasma cutter is already smaller than normal drill as part of its advantages.

Same to stacks.

PastelPrinceDan:
some engi players asked about the stacks so that's why they're here. if an admin/maintainer wants them removed, i'll do so.

as for the plasma cutter, no one really ever asks for them except people doing projects (mostly because it kinda Sucks) which is mostly why i added them, but i suppose i could remove them too if asked

Very-Soft (changes requested):
Conflicts

PastelPrinceDan:
...i fucked something up somewhere, please hold

PR #12016, opened 2021-12-16
makes large autoinjectors cheaper

PastelPrinceDan:
i forgot that upgrading the autolathe doesn't make it require less mats per thingie printed
which made one large autoinjector cost one whole plastic sheet, even with upgrades
which, really, is way too much for an autoinjector, given how a regular one costs literally nothing*

*edit: now costs 10 plastic after suggestion


Heroman3003:
I still think that regular autoinjectors should have cost for both recycling and printing, but very cheap like 5-10 plastic, and suspect that back in some time when autoinjector was added to autolathe the lack of cost was not intentional due to how autolathe costs worked, and eventually it was just settled as something normal.

PastelPrinceDan:
i'll add that real quick while i'm at it

PR #12032, opened 2021-12-25
adds a fireaxe to atmos

PastelPrinceDan:
in the oxy gen room, next to the cyborg charger

### DIFF
--- a/maps/stellardelight/stellar_delight2.dmm
+++ b/maps/stellardelight/stellar_delight2.dmm
@@ -2289,6 +2289,9 @@
 /area/stellardelight/deck2/aftstarboard)
 "eU" = (
 /obj/machinery/atmospherics/unary/vent_scrubber/on,
+/obj/structure/fireaxecabinet{
+	pixel_y = 28
+	},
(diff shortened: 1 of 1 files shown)


Very-Soft:
Why does atmos need a fireaxe?

Like, I understand the real world parallels, but I actually never stopped to question what the fireaxe actually does for atmos, apart from act as a convenient spider fighting weapon.

PastelPrinceDan:
people requested it, idk i don't play atmos

Very-Soft:
Well, I don't really think atmos needs a strong violence do-er. The only reason I gave one to paramedics is because they likely to need to go into places where there are things that are hurting people. 

Atmos people don't really do that, and you don't actually USE the fireaxe to break through doors and walls usually like you would in real life. Like. You COULD break through a door I guess, but there's better tools for that, that you start with in your toolbelt. The fireaxe itself is mostly only useful for killin' stuff pretty good, as far as I am aware. It's a decent self defense weapon, and atmos's job isn't/shouldn't be to go into areas with active hostile forces. If that kind of situation comes up, they can generally probably obtain weapons through whoever they're working with.

PR #12045, opened 2022-01-04
Air Traffic Adjustments

Enzo-Leon:
This is sort of a small test for a future change on Air Traffic to give us more options for lore and events to take place as well as seeing how to establish more relations between different entities and Virgo itself. The first test is with the space rats and Vox with a general blanket no pirate event towards them. Once the species is closer to a finished state there will be more adjustments and unique events set in place that I would like to try and also have between the other entities. This update is pprreetttyyy much going to be just lore and fun little nuggets of data for folks, and just help with world-building more than anything else.


SigholtStarsong:
Would it be in the scope of this to adjust the Emergency Responder Frequency to stay static at 121.5? It seems odd that the emergency frequency swings around as much as any other frequency.

Enzo-Leon:
That is something that I was thinking of doing in the future once more changes in air traffic go through =3 The current system needs some love, and this is the first of many that I'll wanna do on it.

PR #12061, opened 2022-01-11
Removes the announcement for main pred and morph spawns

Very-Soft:
The announcement only really encourages people to go hunting for it, and usually not in the good way, which can in general ruin the surprise that these roles can employ.

So! I think it would be better if we just didn't announce them publicly, since they aren't a thing we ever want people to go and respond to mechanically, and, I think that announcements like these ought to be reserved for things we want people responding to.

### DIFF
--- a/code/game/objects/structures/ghost_pods/event_vr.dm
+++ b/code/game/objects/structures/ghost_pods/event_vr.dm
@@ -80,11 +80,6 @@
 	to_chat(M, "<span class='warning'>You may be a spooky space monster, but your role is to facilitate spooky space monster roleplay, not to fight the station and kill people. You can of course eat and/or digest people as you like if OOC prefs align, but this should be done as part of roleplay. If you intend to fight the station and kill people and such, you need permission from the staff team. GENERALLY, this role should avoid well populated areas. You’re a weird spooky space monster, so the bar is probably not where you’d want to go if you intend to survive. Of course, you’re welcome to try to make friends and roleplay how you will in this regard, but something to keep in mind.</span>")
 	newPred.ckey = M.ckey
 	newPred.visible_message("<span class='warning'>[newPred] emerges from somewhere!</span>")
-
-	if(prob(announce_prob))
-		spawn(2 MINUTES)
-			command_announcement.Announce("Unexpected biosignature detected in the maintenance tunnels of [station_name()].", "Lifesign Alert")
-
 	qdel(src)
 
 /obj/structure/ghost_pod/ghost_activated/maintpred/no_announce
@@ -119,11 +114,6 @@
 
 	newMorph.ckey = M.ckey
 	newMorph.visible_message("<span class='warning'>A morph appears to crawl out of somewhere.</span>")
-
-	if(prob(announce_prob))
-		spawn(2 MINUTES)
-			command_announcement.Announce("Unexpected biosignature detected in the maintenance tunnels of [station_name()].", "Lifesign Alert")
-
 	qdel(src)
 
 /obj/structure/ghost_pod/ghost_activated/morphspawn/no_announce


Runa-Dacino:
I feel we need a possible alternative way to let people know who do want to interact with maint preds the intended way to know when they're available.

As such, I suggest a new event/antag preference: "Maint prey".

If the character has that on, they are notified that there's a maint pred in the tunnels, either OOCly or using an IC narrative approach.

By making it a preference like antag prefs, it will become possible for admins to turn it off for problem players if they abuse it.

Heroman3003:
While  I agree that currently announcement does cause unnecesary stir, I also definitely think that the main reason for it originally existing becomes a way bigger point on SD. That being: Let people know a thing is even there.

It's not as big of an issue for morphs, who can ventcrawl and are lowkey expected to use that to travel around and find victims, but for maint preds... Well, if they actually stay confined in maint, the likelyhood of actually running into a victim is extremely low, due to mix of size of maint being overall and proportionally smaller; as well as less reasons to ever visit it. In addition to that, lack of access to comms in terms of listening as well as alternative ways to communicate with anyone, I do feel like maint pred's ability to naturally find prey gets really damaged.

But I also am not sure about adding a dedicated role for 'get ate by maint preds' especially with pref disparity being a thing.

I think a potential middle road solution would be to keep announcements gone from morphs; and for maint preds, change them up to be something very ambigous in terms of 'what happened' and but OOCly recognizably unique. That way if people use that to go "oh there's a spook in the maint" can just bwunk them for metagaming; while people who actually understand the purpose of announcement will either ignore it or use as quiet and convinient excuse to take a walk around maintenance.

Zeldazackman:
I'll say it's about time, as I've seen characters go from staring blankly in the bar for the past half hour waiting for a scene to "Unga must kill thing and get muh valids" in no time at all once the announcement is made, regardless of whatever creature was picked in the end or just how harmless said creature is.

lm40:
I mostly agree with heroman I don't want maintenance predators running around the hallways so it would be helpful to have some sort of message that OOCly clues in people who want to get maintenance predded that now is the time to go running around maintenance. I can see the case for consistency that we don't announce things that are not OOCly problems in need of resolution, but there should still be some kind of notification.

Heroman3003:
While I agree that the true best case would be for maint preds to get some sort of tools to either announce themselves if they wish (in a local way) or tools that allow them to better actually find/attract/notify *specific* crew (which is my biggest concern, inability to find any viable prey because like... why would any viable prey ever just wander maint for a medium chance event that requires someone picking it up and may not even happen over course of 6 hour round).

But all the other issues the event has, regarding the validhunting and people announcing danger over comms, seems to be more of a enforcement issue and rules documentation issue. There need to be better, stricter, clear guidelines related to simplemob roles for both preds and prey, put along with other rules on rules page, similarly to how security has their special rules. If people run into maint pred and start yelling over comms, even though pred is maint, that should be a bwoink. If people start entering maint with intent of hunting the maint pred, that should be a bwoink. And if maint pred picks giant scary monster and starts trying to be 'cute' to people wanting nothing to do with it in the bar, that should be a bwoink. Instead of having semi-vague rules on "the maint pred is kinda maybe valid if its not staying in maint properly", OOC rulings on what they should be and shouldn't be need to be defined and put with other rules, and ways in which crew is allowed to react to it just existing should all definitely be limited and documented.

As semi-related thing, it seems like most of later problems with maint pred come up from spider varieties of it, so maybe pruning out those would already be an improvement (if only because spider is objectively the noisiest one and impossible to be stealthy with, while also being an actual non-pred threat to ship on regular basis and certain options being capable of spamming obstructive webs)

Very-Soft:
I personally think that it would probably be good if main pred was divided up a little bit. 

I don't think we really need there to be main pred spiders in general. Like, I'm fine with there BEING spiders, but they should probably be easy to tell apart from normal spiders at a glance. It's easy for confusion to happen there if you only get a peek at them, so, it might be good if they were removed from the list until we perhaps get some unique spider sprites for maint preds or some such. 

But! Beyond that, I also think that not all maint preds are created equal. Some it makes more sense for them to be running around in public areas than others. Like, the wolf as an example. TBH I think that would be better as like, an alternative to the maint pred spawn, since that's way more likely to be treated like a pet, than, say, the sect queen as an example. 

I think if the selection was more nuanced between 'maint pred' and 'pet critters', then we would have a little better grounds for making proper rules for them. 

Cuz like, if someone goes and plays the wolf and is playing a dumb idiot puppy that eats people in the hallway. Like, OOCly and ICly that's potentially fine, and you wouldn't have to stretch the imagination far to say 'oh that's just someone's dog', where, the same situation with a weretiger is like, 'okay you're pushing it'. Enforcing the same rule for both of those things evenly is difficult and often undesirable, because they are two different things with very different general expectations and presence in the IC space. 

But, again, such things are outside of the scope of this specific PR. uwu

Heroman3003:
> 

Dividing it up is something I've considered; but ultimately I feel it would carry certain redundancy without fixing core issue that the 'spooky/dangerous' critters would be still having.

I do agree that we lack fun variety in what voremobs we have. Our spiders aren't voremobs, our slimes aren't voremobs; both are semi-common hostile crew dangers and so using them as replacements has problems, despite both being almost staple to the vorny shenanigan tropes. So that's why the regualr spiders and kinda-lame-looking jelly blob (no offense to whoever made it) ended up being used. Special more-vorny-oriented versions of such mobs would be good, and has been done already once with defanged xenomorph being such for regular xenos. But I do think that splitting the 'spooky monster' and 'buppy' categories of maint preds is redundant due to both aiming to serve same objective of "play as a simplemob and do vorny"; and if only difference is 'where am i allowed to go", that could be adjusted without need of entirely splintering maint preds.

It would also not fix issue more dangerous ones are having to just split them off, as I do believe that, well. As pred they need a way to find prey, not just sit around hoping prey happens to come by. Announcement was meant to help that, but people are rousing panic on account of misunderstanding purpose of the role, the event and server in general. Ultimately, main thing that would be needed for maint pred is something like morph's ventcrawl (except maybe not literal ventcrawl), that allows them to take a peek at what's happening, who is around, who is free, who is alone and who is edible.

PR #12064, opened 2022-01-12
Makes deconstructing and reconstructing SD shields not lose unique power consumption

### LINKED ISSUE #12057: Deconstructing, and then re-constructing the Advanced Shield Generator increases powerdraw massively
Normally, the Advanced Shield Generator, set to range 55, only draws around 3000-4000 kW of power while Active using settings of Hull, Warp and Hyperkinetic projectiles.

Due to the shield being "Overloaded", I had to deconstruct it, then reconstruct so to fix it. This skyrocketed power draw to 11 000+ kW

At the above described settings, Adv. Shield Gen should draw 3000-4000 kW

#### What actually happened
It drew 11 000+

#### Steps to reproduce
- Turn shield generator on, activate it
- Forget to deactivate it so it drains fully
- Shield says "Overloaded", cycling it doesn't do anything
- Use screwdriver, crowbar on shield that's turned off
- Place the parts it ejected inside, screwdriver closed
- Set: range 55, max powerdraw 6 000, Hull shield, multi-dimensional warp, Hyperkinetic projectiles, 10 passive power
- Turn on, notice it takes around 200-300 kW on stand-by
- Activate, it's now 11 000 kW

#### Code Revision
Commit: 82228c1d70685413010d476abf9f116a1d8562b3 
 

- The Advanced Shield Generator was not moved from its location near the HoP's office.
- The Advanced Shield Generator was hotwired to the engine using a separate cable running to the shield room. On previous rounds, this caused no issues.

Heroman3003:
Fixes #12057

Makes SD's shield generator a special subtype with unique board that is not obtainable any other way (other than admin spawning). Kind of silly that it's 'intentional' to lose all power when one accidental crowbarring with panel open can result in permanent loss of something rather important. So, special subtype of shield generator!

### DIFF
--- a/maps/stellardelight/stellar_delight3.dmm
+++ b/maps/stellardelight/stellar_delight3.dmm
@@ -9856,9 +9856,7 @@
 /obj/machinery/computer/ship/navigation/telescreen{
 	pixel_x = 32
 	},
-/obj/machinery/power/shield_generator/charged{
-	power_coefficient = 0.33
-	},
+/obj/machinery/power/shield_generator/ship_optimized/charged,
 /obj/structure/cable,
 /turf/simulated/floor,
 /area/stellardelight/deck2/central)
--- a/maps/stellardelight/stellar_delight_things.dm
+++ b/maps/stellardelight/stellar_delight_things.dm
@@ -174,3 +174,17 @@
 
 /obj/machinery/power/quantumpad/scioutpost
 
+
+
+/obj/machinery/power/shield_generator/ship_optimized
+	power_coefficient = 0.33
+	circuit = /obj/item/weapon/circuitboard/shield_generator/ship_optimized
+
+// Starts fully charged
+/obj/machinery/power/shield_generator/ship_optimized/charged/Initialize()
+	. = ..()
+	current_energy = max_energy
+
+/obj/item/weapon/circuitboard/shield_generator/ship_optimized
+	name = T_BOARD("ship-optimized shield generator")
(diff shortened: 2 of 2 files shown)


Novacat:
I was kind of out of the loop for that discussion but... why does SD have a special generator subtype?

Heroman3003:
> I was kind of out of the loop for that discussion but... why does SD have a special generator subtype?
> 
> My current line of thinking is that if the Shield Generator's power consumption is too much for the _Stellar Delight_ then it probably is too high in general and the whole component's power consumption should probably be lowered.

SD's shield generator currently uses a shield generator varedited to consume only a third of the power. Not sure about balance of shields in general and ultimately not quite the goal of this PR, but if someone decides that overall power draw needs to be thirded; sure I can do that.

Very-Soft:
> I was kind of out of the loop for that discussion but... why does SD have a special generator subtype?
> 
> My current line of thinking is that if the Shield Generator's power consumption is too much for the _Stellar Delight_ then it probably is too high in general and the whole component's power consumption should probably be lowered.

The shield generator's default values are achievable, but very, very demanding. The reason that I lowered the required power on it, was because people were sort of taking making the shield run indefinitely as a mandatory action (which it's not), and holding up other ship operations to achieve this. The power needs were lowered so that they would take less work to achieve, so that engineering would have less reason to refuse to turn the fuel pump on.

I've been thinking about reverting that change though, because it was never really intended for the ship to just always have shields all the time, and I'm not sure that this change has necessarily been a positive one.

PR #12079, opened 2022-01-16
Buffs speed of protean blobs

### LINKED ISSUE #12060: Protean Blob Advertised As Speedster - Empirical Testing Determined That Was A Lie
Protean blobs are advertised as "moving slightly faster" upon assuming blob form when they most definitely are not.

For the words that say that protean blobs more slightly faster to not be a lie.

#### What actually happened
The words are a lie. You in fact move noticeably slower.

#### Steps to reproduce
- Be a protean.
- Run in a straight line to see how fast you move (Optionally record your screen and count how many frames it takes to run the distance).
- Blob, see the info text that says you are now faster.
- Blob the same straight line and see that you move slower. (Optionally, have someone who's not a blob run parallel and see how they move faster).

#### Code Revision
Server revision: B:release D:
Commit: 82228c1d70685413010d476abf9f116a1d8562b3 
 

- This is not a case of the Haste trait or any equipment slowdown effects being involved.
- The fix is to either increase protean blob movespeed to match the text, or change the text to say that you move slower.

Heroman3003:
Fixes #12060 
doesn't make them turbo fast, but no longer as slow either.

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/protean_vr/protean_blob.dm
+++ b/code/modules/mob/living/carbon/human/species/station/protean_vr/protean_blob.dm
@@ -14,7 +14,6 @@
 	health = 200
 	say_list_type = /datum/say_list/protean_blob
 
-	// ai_inactive = TRUE //Always off //VORESTATION AI TEMPORARY REMOVAL
 	show_stat_health = FALSE //We will do it ourselves
 
 	response_help = "pets"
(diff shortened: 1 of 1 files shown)


Detective-Google:
https://github.com/VOREStation/VOREStation/commit/ae80bbb10a3b11c4b53a1498b198b8ea5eb8004a#diff-11cc82ed1a1315ba7d99ecf11f86676bd9dedcc56b31a2cfc35a8478d3d941eb back and forth and back and forth and back and forth and back and forth and-  (they are uselessly slow right now this is a good pr)

Heroman3003:
If people try to abuse it for combat again, it can be nerfed again. Either that or tooltips can be changed and this nerf re-applied. We shall see. I do hope however that attempts to abuse blobform will be punished with removing people from protean whitelist, not by nerfing proteans.

TheDavestDave:
you know there's ways to make the nerf blobs without makeing the non combat focused race unable to escape combat. I'd suggest looking into an inability to pull objects

PR #12082, opened 2022-01-16
[MIRROR] Adds event regions to GameMaster events

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/8334
--------------------
This lets you define an event as occurring in
```
EVENT_REGION_PLAYER_MAIN_AREA // The main areas that players are expected to be in
EVENT_REGION_SUBTERRANEAN // The mines
EVENT_REGION_PLANETSIDE   // Somewhere on a planetary surface
EVENT_REGION_SPACESTATION // Somewhere in space that players likely have ready access to
EVENT_REGION_DEEPSPACE    // Somewhere "far away" on the overmap
EVENT_REGION_UNIVERSAL    // Affects all players, everywhere (E.g., comms blackout)
```
Event selection will pick a random region, weighted by the activity of players in those regions. The universal region is the total activity of all players everywhere, and so is multiplied by [ARBITRARY FACTOR] (1/2) to favor events that target specific regions. This number is straight out of my ass.

This is useful, because it means engineers on a planetary map won't cause meteor storms for people in space, but we _could_ set up an event that does something similar to _planetary_ z levels.


maps/southern_cross/southern_cross_defines.dm
```diff
@@ -153,6 +153,7 @@
 	flags = MAP_LEVEL_STATION|MAP_LEVEL_CONTACT|MAP_LEVEL_PLAYER|MAP_LEVEL_CONSOLES
 	holomap_legend_x = 220
 	holomap_legend_y = 160
+	event_regions = list(EVENT_REGION_SPACESTATION, EVENT_REGION_PLAYER_MAIN_AREA)
```
Atermonera, line 156: Note: y'all will need to either disable this system, or copy these lists into your maps defines.

Heroman3003: We don't really use gamemaster either way, we still use old event containers.

PR #12088, opened 2022-01-18
Adds Temperature Adaptations to Xenochimera Trait List along with Autohiss.

Rykka-Stormheart:
Ran these by Scree (Creator of Xenochimera) and he said these are fairly reasonable adaptations.

They're neutral traits available to custom species, and given that Xenochimera are also able to pretty much mix and match a lot of stuff based on what they've eaten/evolved, this will allow greater roleplay/adaptations to occur - for instance, a xenochimera that is well-suited to the warmth of the desert is going to struggle in cold areas, and vice versa.

(It's conceivable that a Xenochimera might eat a wolf, and adapt/evolve a super-thick coat of fur suited for winter planets.)

Upstream port of https://github.com/CHOMPStation2/CHOMPStation2/pull/2948

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/xenochimera_trait_vr.dm
+++ b/code/modules/mob/living/carbon/human/species/station/xenochimera_trait_vr.dm
@@ -1,7 +1,10 @@
 /*
 **	For now, these are just neutral traits for Xenochimera only to take.
-**	These are only traits that they should reasonably be able to evolve to acquire themselves.
-**	I won't add the resistances though because those are kinda lame for a 'chimera to take!
+**	Traits defined as custom_only = FALSE in neutral.dm will be available for Xenochimera to take as well.
+**	As such, be careful not to duplicate the traits, and only add dupes where necessary.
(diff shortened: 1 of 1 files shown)


Novacat:
Scree mentioned that this might not work too well since Xenochimera have their own temperature handling code. This might need testing.

Rykka-Stormheart:
> Scree mentioned that this might not work too well since Xenochimera have their own temperature handling code. This might need testing.

Ye, I'm also mildly leaning towards finding a way to just rewrite/change the traits to be a little more complex - rather than just copy-paste heat and cold adapt, use the temperatures from those and apply the xenochimera handling to those new thresholds. I'll play around with it.

For now, on downstream where the traits are live, passing below the danger temp simply causes damage iirc.

Screemonster:
yeah, I think the traits override the danger temps, whereas xenochimera have a danger temp of 0K deliberately and only take damage when they're, like, awake

Rykka-Stormheart:
Closing this in lieu of finding a better solution.

Will re-open down the road once I find a better one. 7o

PR #12105, opened 2022-01-22
Adds Pred-Pref, Prey-Pref tags to list, standardizes colors

Rykka-Stormheart:
Adds Pred-Pref and Prey-Pref, self-explanatory.

Standardizes the colors to make it a little more legible, non-vore is now simply black so it blends with the background.
Basically, non-vore folks aren't colored for easier identification across the list.


Heroman3003:
Switch being purple is better because it matches the colors used on discord server for same roles, IMO.

TheToaster98:
It may also be a good idea to change Bottom and Top for Sub and Dom, maybe.
Sounds like it would make a little more sense.

Rykka-Stormheart:
> It may also be a good idea to change Bottom and Top for Sub and Dom, maybe. Sounds like it would make a little more sense.

Not always, honestly. It'd probably be better just to add a third tag - dom does not universally = topping, etc.
I'd be okay with adding a third tag in a separate PR.

PR #12128, opened 2022-01-29
Xenobotany Pass 1 (Of who knows how many)

TheObserver-sys:
### About This PR
Xenobotany's in a funky spot currently, least the way I see it. It.. is definitely lacking compared to every other department in science in terms of what it can provide to the station, with nothing really special outside gatchaing cargo for funny seeds. While I would love to do a massive blanket patch to make it at least as interesting as other server's botany departments, it's neither feasible nor a good idea to push that massive of a change all at once, so I've decided to hit just a few areas that heavily disappoint me.

What's added:
A "new" ambrosia mutation, that being Ambrosia Gaia.  It is a difficult to maintain plant that produces the chemical Earthsblood, which is an extremely powerful curative extract -- at a cost. Partaking in it inflicts heavy hallucinations and a drugged state, to the point where it **will** damage your brain. Balance with Alkysine, and your patient should be right as rain. A warning: Take only upto 15 units of the stuff, unless you wish to overdose severely upon it.

A new grass mutation, carpet! Like grass, the fibers can be used in hand to create tiles of carpet, which then can be used to pad other objects as seen fit. To note: This doesn't let you make any special carpets, just baseline carpet. Cargo still has a job when it comes to making nicer looking shag-dens.

A new chili variety, the ghost chili! This pepper is.. pain incarnate, if we're bein honest. Contains condensed capsaicin, and will (eventually) be used in some rather excruciatingly hot foods. For now, it can be used as bootleg teargas grenades in the event of a massive infestation of spiders, or given to Explo to clear out mobs without expending ammo.

An upgrade to the plant analyzers: They now can recognize if a plant is able to emit reagent carrying spore clouds, and will display such a warning to the user. No more being surprised by getting hit by dense cloud of phoron smoke and burning a hole in your liver.

What's changed:
The maximum electrical yield of a plant based battery has been radically boosted. The previous calculation was.. extremely lacking, to say the least, topping out at an honestly pathetic 2MJ. While on TGcode, this is fine, as potency is very easy to build, and there's a gene that unlocks the true power of such plants, it's kinda wacky for this to be the case here, where it's an active war to get anywhere remotely past 50 without T4/T5 lasers. The new calculation allows a dedicated Xenobotanist to produce a cell that is rated at 40K/40MJs, beating out normally produced cells by 10K, rewarding the botanist for their dedication to their work.

The "Sporing" trait has been moved appropriately, from being covered under BIOCHEMISTRY to being covered under OUTPUT. Makes little to no sense to have it stuck to biochemistry, as it not only means that 90% of sporing plants will only poison people, but biochemistry normally wouldn't pertain to a plants ability to eject such stuff into the air. I had also considered fruit -- but that has the exploding trait covered, which is fair enough. You still can't directly mutate into it, making it something a player would have agency in placing upon a plant.

### Why This Is Good For the Game
There should be some semblance of unique/useful rewards for attaining knowledge in xenobotany. Be it a chemical that you can only get from it that can be a great help to the people of the station, to just something goofy like "100% GMO FREE CARPET", to cells that can be used in a pinch.  It also makes no sense for one of the more dangerous side effects of a plant to not show up on plant scanners, considering it already alerts us when plants produce new gasses, and if it will genuinely try to eat you. Lethal surprises you can't know about before getting hurt by it are kinda unfun, especially when it's not by the will of another player, but by pure dice roll.

### Changelog:
add: Three new plants, and a new reagent have been recently discovered by the hard work of Xenobotanists aboard the Stellar Delight. Grow them in your garden today!

tweak: Heavily adjusts the formula that decides plant capacitance, from Potency * 20 to Potency * 200, making potential end products **better** than what a protolathe can produce if potency is near maxed out.

fix: After several incidents with faulty plant analyzers aboard the Stellar Delight, an update to the scanning capabilities was pushed by a senior Xenobotanist working with Central Command, revealing if a plant can release spores periodically. Incidents featuring spore related poisonings are estimated to drop -- and many lawsuits potentially averted.


code/modules/hydroponics/seed.dm
```diff
@@ -737,11 +737,11 @@
 		if(GENE_BIOCHEMISTRY)
 			P.values["[TRAIT_CHEMS]"] =        chems
 			P.values["[TRAIT_EXUDE_GASSES]"] = exude_gasses
-			traits_to_copy = list(TRAIT_POTENCY, TRAIT_SPORING, TRAIT_BENEFICIAL_REAG, TRAIT_MUTAGENIC_REAG, TRAIT_TOXIC_REAG)
+			traits_to_copy = list(TRAIT_POTENCY, TRAIT_BENEFICIAL_REAG, TRAIT_MUTAGENIC_REAG, TRAIT_TOXIC_REAG)
 		if(GENE_OUTPUT)
 			traits_to_copy = list(TRAIT_PRODUCES_POWER,TRAIT_BIOLUM)
 		if(GENE_ATMOSPHERE)
-			traits_to_copy = list(TRAIT_HEAT_TOLERANCE,TRAIT_LOWKPA_TOLERANCE,TRAIT_HIGHKPA_TOLERANCE)
+			traits_to_copy = list(TRAIT_HEAT_TOLERANCE,TRAIT_LOWKPA_TOLERANCE,TRAIT_HIGHKPA_TOLERANCE,TRAIT_SPORING)
```
Heroman3003, line 744: I don't think it makes much sense to move Sporing to Atmosphere (which is mostly plant's ability to endure atmosphere, not its interactions with it). Exhuding gasses is under Biochemistry and unique production features are under Output and Metabolism. It makes more sense under any of those three than it does there.

code/modules/hydroponics/grown.dm
```diff
@@ -289,6 +289,20 @@
 		qdel(src)
 		return
 
+	if(seed.kitchen_tag == "carpet")
```
Heroman3003, line 292: This is a good way to obtain otherwise obscure and rare construction item, but I do have suggestion for the future on that front: more subtypes of this plant. Bigger issue with carpets tends to be that subtypes that are not orderable in cargo (default-red and blue) are pretty much unobtainable without adminbus and thats sad. Maybe having some subtypes be xenobotany exclusive and some cargo exclusive as good solution for both sides? Just something to think about.

code/modules/reagents/reagents/medicine.dm
```diff
@@ -1501,3 +1501,23 @@
...
+
+/datum/reagent/earthsblood
+	name = "Earthsblood"
+	id = "earthsblood"
+	description = "A rare plant extract with immense, almost magical healing capabilities. Induces a potent psychoactive state, damaging neurons with prolonged use."
+	taste_description = "honey and sunlight"
+	reagent_state = LIQUID
+	color = "#ffb500"
+	overdose = REAGENTS_OVERDOSE * 0.50
+	scannable = 1
```
Heroman3003, line 1513: Scannable is mostly for normal/conventional medicines. It being a more obscure, xenobotany exclusive chem doesn't make much sense to have scannable.

code/modules/reagents/reagents/medicine.dm
```diff
@@ -1501,3 +1501,23 @@
...
+
+/datum/reagent/earthsblood/affect_blood(var/mob/living/carbon/M, var/alien, var/removed)
+	M.adjustBruteLoss (-4 * removed)
+	M.adjustFireLoss (-4 * removed)
```
Heroman3003, line 1517: adjustBrute/FireLoss should probably be replaced with heal_organ_damage, similarly to how bicard/derm handle it.

Heroman3003:
Honestly, looking at actual numbers, both the potato cell buff and Earthsblood don't look as bad balance-wise as I thought they would from way you were describing them in discord. No strong feeling one way or another in terms of those, therefore.

code/modules/reagents/reagents/medicine.dm
Heroman3003: Also should consider making earthsblood be affected by per-species chem efficency modifier that used in most medicines.

code/modules/hydroponics/grown.dm
TheObserver-sys: There are two ways I was thinking about going about that, but didn't feel right to put under the purview of this PR in particular:
Option A would be as you say, make more subtypes of carpet-plant, which I worry might end up being rather bloated in the long run, even if making it possible to have a bunch of fun carpets available to the players.

Option B, which I'm more partial to, would be to go the TG station approach. Make Carpet a reagent -- and then add unique recipes to change it further, which replaces tiles by being sprayed upon them. I feel like option B is probably the better approach, and would allow not only xenobotany, but other various disciplines to make unique carpets in the future.

code/modules/reagents/reagents/medicine.dm
TheObserver-sys: I'll undo that, then. Had it scannable because of how powerful it is.

code/modules/reagents/reagents/medicine.dm
TheObserver-sys: Would this also *heal* organs, though? I don't want it undoing the damage being dealt to one's brain, as the jacked numbers on the chem are only rewarded +because+ it eats away at the brain.

code/modules/hydroponics/seed.dm
TheObserver-sys: I can move it to Output instead -- that one has so few results as it is that it'd be nice to give it some more love. I don't like keeping it biochemistry, because it strictly limits it to luck of the draw to get a non toxic spore plant if you want to do neat things with it, mostly owing to how the plant hardware works. At some point, I'd like to make the bioballistic able to target certain subsectors of genes, but that is a very long ways out.

code/modules/reagents/reagents/medicine.dm
Heroman3003: heal_organ_damage applies to 'external' organs. Aka hands/arms/torso/etc.

code/modules/hydroponics/seed.dm
Heroman3003: That'd be nice as long as its in gene that, well, makes sense for it. Honestly, just more consistency in categories sake rather than balance.

code/modules/reagents/reagents/medicine.dm
TheObserver-sys: heal_organ_damage was put in, I'm thinking about keeping it equally effective across all organics, though.

KillianKirilenko: Welcome to unintuitively named procs, we hate it here just as much as you will.

KillianKirilenko: As an additional note: the per-species chem modifier isn't actually used by anything that I know of. It's just there as an option that could be used in species or traits. So hooking that up shouldn't change its performance in a significant manner.

PR #12134, opened 2022-01-31
Swap slot machines

Very-Soft:
I didn't realize there were multiple variants when I placed this! >:O


PastelPrinceDan:
aoeoeoe i didn't notice that and fixed it too lmao
i'll go ahead and close mine

PR #12135, opened 2022-01-31
"please let us gamble in peace"

PastelPrinceDan:
swaps the slot machine in commons with a station slot machine
takes Thalers instead of chips that aren't around anywhere, so people can actually gamble


PastelPrinceDan:
soft fixed it 5 minutes before me :noooooooooooo:

PR #12143, opened 2022-02-03
Return to Tether

Very-Soft:
Fixes up a few of the Tether's issues to get it working properly again, and swaps back to it.


Wickedtemp:
I really, really, really don't want to go back. Granted, I don't play as much anymore, but in my opinion, the Tether was just a bad map. SD is a flat-out improvement when it comes to roleplaying and gameplay.

PR #12146, opened 2022-02-03
skimpy cap "dress" probably more like a leotard idk

Hatterhat:
virgofemformal
notes:
- i'm not spriting teshari variants and will actively refuse to. i was rather tempted to make it wearable for everyone but tesh but i decided that'd be too much effort
- does not come with pantyhose, bring it yourself
- if you're going to wear it, remember that there's a verb in the object tab to toggle underwear
- only covers UPPER_TORSO, which is to say the chest. will not protect lower body, which should be kind of obvious

original pr Citadel-Station-13/Citadel-Station-13#7307, sprites by @Michiyamenotehifunana


Michiyamenotehifunana:
Oh hey I forgot this existed

Hatterhat:
> Oh hey I forgot this existed

tbf its a nice sprite

PR #12149, opened 2022-02-04
makes amour for lucky

TheDavestDave:
it took three people to get this to work correctly. Custom item with admin aprovel


code/modules/vore/fluffstuff/custom_items_vr.dm
```diff
@@ -1419,3 +1419,24 @@
 	icon_state = "stamp-midhorror"
 	stamptext = "This paper has been certified by The Council of Mid Horror"
 
+//thedavestdave Lucky
+///I know this is pretty bodgey but if it stupid and it works it isn't stupid
+/obj/item/clothing/suit/storage/hooded/explorer/Lucky
```
KillianKirilenko, line 1424: Minor standards thing, but you should avoid capitalization in item paths and icon/item state names.

TheDavestDave: low case l installed

PR #12151, opened 2022-02-05
adds a catslug crate to cargo

PastelPrinceDan:
in the spirit of cat/chicken/etc crates
the price is steep because catslugs are more easily abused than cats and such, should a ghost join them

requested by @BonniePandora


Very-Soft:
HMMMMMMMMMMMMMMMMM

PR #12153, opened 2022-02-06
Speeds up the initial NIF calibration timer.

SpaceVampire:
**About The Pull Request**

Lowers the NIF calibration timer from 35 minutes, to 5 minutes.

**Why It's Good For The Game**

I view this as more of a quality of life change then any kind of balance change, because I kept making new characters only to feel like I really didn't want to get a NIF for them because it sucks up so much of a given shift in that post-surgery state. Once you have one is it really easy to keep it healthy, and given that they are mostly just used for soulcatcher fun or other bits of creative flavor I really don't see why it should have to be so painful to acquire it. This should be enough time to give the same general feeling, but without dragging it out so long that it becomes really discouraging. There is also an exploit/oversight you can use to bypass most of the time already, and this helps stop me from getting into trouble by using it after I tired of installing it for like the 6th time already. Talked with a handful of people, and they all seemed to agree with me. So uh. Just waiting to see what the maintainers think of my logic here. It seemed like just getting people from science to make it, and people from medical to install it are the parts that can actually add RP to the shift, and they are both preserved here.

### DIFF
--- a/code/modules/nifsoft/nif.dm
+++ b/code/modules/nifsoft/nif.dm
@@ -276,7 +276,7 @@ You can also set the stat of a NIF to NIF_TEMPFAIL without any issues to disable
 			install_done = world.time + 1 MINUTE
 			notify("Welcome back, [owner]! Performing quick-calibration...")
 		else if(!owner)
-			install_done = world.time + 35 MINUTES
+			install_done = world.time + 5 MINUTES
 			notify("Adapting to new user...")
 			sleep(5 SECONDS)
 			notify("Adjoining optic [human.isSynthetic() ? "interface" : "nerve"], please be patient.",TRUE)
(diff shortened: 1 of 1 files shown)


Heroman3003:
This is quite literally the only downside or complexity in NIF getting process and it's something that only happens once. I don't think almost removing it is a good idea.

Runa-Dacino:
I like the NIF calibration time being 30 minutes. Gives you justification for cuddling and general sappy stuff.

SpaceVampire:
> This is quite literally the only downside or complexity in NIF getting process and it's something that only happens once. I don't think almost removing it is a good idea.

It's once per every character you have, which really adds up pretty quickly. I can see how it would be less of a problem if you only play one person. But I don't really think there is anything engaging about having to wait around for over half an hour in a state of lessened functionality for something that only really exists to add flavor. It can be pretty annoying, given that not every shift even has the people/resources needed to install it and if you show up too late into it you risk running out of time.

>I like the NIF calibration time being 30 minutes. Gives you justification for cuddling and general sappy stuff.

I do fully understand that appeal of it, but if you enjoy cuddling'n'stuff you are always free to just keep doing that once it finishes the mandatory amount. Like. I have done the exact same thing before, I just don't want to be "forced" to do it.

Very-Soft:
No thanks

SpaceVampire:
So you aren't interested in talking about lowering it at all?
It's only 15 minutes on Cit-RP which would still be better then this.
@Very-Soft

PR #12161, opened 2022-02-06
Allergy Rebuff

KillianKirilenko:
It has been brought to my attention that Polaris' tweaks to allergies made them a bit more anemic than I'd like. They were intended to be deadly, and now they should be. Damage severity has been tripled, and the old "reduced effect" trait is now the current baseline.

Also added an extra neutral trait that doubles the already increased lethalty, for people who really like to flirt with danger.


Wickedtemp:
HELL YEAH

PR #12174, opened 2022-02-07
Hastily gives the ERT their suits back

Novacat:
A previous update removed all the ERT's voidsuits and hardsuits. Probably not intentional.

Also adds some newer stuff to the ERT's equipment bay.


KillianKirilenko:
"Probably not intentional", says cat who was the last person to touch the map according to the file history log.

PR #12183, opened 2022-02-08
Removes several overly dangerous Wilderness PoIs

Heroman3003:
Removes Merc Tower and Dog Base for being mercenary-related PoIs and not at all fitting for wilderness. Also makes separate file for Wilderness PoIs so that polaris PoIs don't sneak past vetting for such on accident again.

Removes otie house and all 'big monster boss' PoIs as well, first for being guarded by turrets as dangerous, and others for being big dangerous monsters in supposedly 'mostly safe' Wilderness.

### DIFF
--- a/maps/submaps/surface_submaps/plains/plains_vr.dm
+++ b/maps/submaps/surface_submaps/plains/plains_vr.dm
@@ -1,0 +1,349 @@
+// This causes PoI maps to get 'checked' and compiled, when undergoing a unit test.
+// This is so CI can validate PoIs, and ensure future changes don't break PoIs, as PoIs are loaded at runtime and the compiler can't catch errors.
+// When adding a new PoI, please add it to this list.
+#if MAP_TEST
+#include "farm1.dmm"
+#include "construction1.dmm"
+#include "camp1.dmm"
+#include "house1.dmm"
+#include "beacons.dmm"
+#include "Epod.dmm"
+#include "Epod2.dmm"
+#include "PascalB.dmm"
+#include "bonfire.dmm"
+#include "Rocky5.dmm"
+#include "Field1.dmm"
+#include "Thiefc.dmm"
+#include "smol2.dmm"
+#include "Mechpt.dmm"
+#include "Boathouse.dmm"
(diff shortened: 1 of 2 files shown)


Zeldazackman:
On this PR.

The wolf den poi is filled with canines that retaliate _upon being attacked_, but otherwise have the custom description of being "pretty chill for a wolf" giving a hint as to their behavior along with a crumpled up note stating as such for some background history of the poi.

The lonewolf poi is an oddity in that, for all intents and purposes the greatwolf that is spawned cannot actually survive out in phoron and will die in the span of a few minutes after the round starts, but ontop of that explicitly states that it too is also passive unless attacked or "given a reason not to be".

The leopardmander poi has mobs that react in kind in that they are not hostile unless attacked with the tymisians and siffets, and other mobs being keyed to having their faction as neutral, this includes the named red ottie named "Tawny".

Otie Shelter poi contains a mixture of both friendly and feral oties located within, with the majority being made up of coties or "collared oties" which start off as tamed.

The pois which are actually overtly hostile despite being in an area which is already barred already behind shutters as a security measure is the "syndi sniper" poi, the "Dragon Plateu" poi, and the "Dog Base" poi.

I'm also not sure if I'm reading it right, but it doesn't seem like the plains_vr file actually contains any datums to spawn nor give weight to the empty cabin poi.

Very-Soft:
Remove the mercs, leave the monsters pls

you're removing plenty of neat vorny things that only really bother you if you engage with them.

Heroman3003:
> Remove the mercs, leave the monsters pls
> 
> you're removing plenty of neat vorny things that only really bother you if you engage with them.

Okay, will put other PoIs back. However I will leave Otie Shack and Dog Base off because it has turrets guarding it

Heroman3003:
> I'm also not sure if I'm reading it right, but it doesn't seem like the plains_vr file actually contains any datums to spawn nor give weight to the empty cabin poi.

I didn't look deeper into empty shack, but it turns out it was incorretly pathed as wilderness thing, not plains, which is why i removed it thinking it accidentally made its way into this specific file.

PR #12188, opened 2022-02-09
Size Standardization Bracelet

Very-Soft:
It makes you 100% size when worn. Similar to a bluespace jumpsuit, it only works when worn, but dis-similarly to bluespace jumpsuits, this has no control over what size you can be. 

You might say its

one size fits all.

I put this in the loadout and in the gadgets vendor, as I feel like the limitation to only changing your size to 100% and back doesn't offer the same utility as something that actually gives you full control over your size, like the bluespace jumpsuit and sizeguns. 

I think that this limitation, and somewhat expensive loadout cost, makes it not as likely to be abused for powergaming as the other size changing items would be.
bs bracelet


Screemonster:
these would be good for explo gear lockers too as a "you're an inconvenient size for being on an away team so here's how you can access the content" thing

Razgriz1032:
Why was an admin log message removed for emags?

KillianKirilenko:
If you expand the file a little, it looks like it was redundant. Either it successfully emags and goes "X emagged a Y", or it fails and goes "X tried to emag a Y". The removed line was making it do both.

PR #12191, opened 2022-02-10
strapless bag

TheDavestDave:
a bag without straps taking out the body spirte. A kit bash of other things but should work. Still has an icon when worn but just not straps.

also in the loadout


PastelPrinceDan:
why put it in the utility tab and not under the bag selection? assuming it's the same amount of space as a regular backpack

TheDavestDave:
becuase I don't know where that lives code wise. And I do know where duffle bags live

PR #12204, opened 2022-02-13
sword crate point reduction

TheDavestDave:
I've had a few people say it's rather expensive for what it is. Hopefully a real dev who is smarter than me will decide the points. Until then  it's reduce by 25% to 75 points.

I'm a little worried that it may give miners durasteel swords round start but I've not noticed anyone doing that and my own tends to stay steel the whole shift.


Very-Soft:
I think 100 is fine for weapons

TheDavestDave:
yep, nice to know my first guess at balanceing was correct

PR #12234, opened 2022-02-14
Now's your chance to be a [BIG SHOT]

GhostActual:
Spamton drip as someone put it.


Very-Soft:
The glasses are okay

I really really dislike the mask, and would prefer we did not get that. That is much too much blatent current year meme shit, and I do not want to encourage more of that.

GhostActual:
Soft has spoken

PR #12235, opened 2022-02-14
Adds vendor refill cartridges to cargo

Heroman3003:
Adds vendor refill cartridges (base sprite courtesy of @Very-Soft)! When used on a vendor they add a full set of default stock to vendor's product capacity.
Vendor doesn't need to be empty or even to have ever been used. You can just stock them up with extra product using cartridges regardless.

Cartridges come specialized and broad. Specialized only refill their specific vendor. Broad refill multiple vendor types.

Specialized cartridges come primarily for public vendors, and are all 10 points per one cartridge except cigarette vendor, which is 15.
Broad cartridges come for both public vendors and departmental ones. They only come in 5-packs of cartidges per 75 point crate (150 for departmental vendor cartridges). They are easier to use though, due to being applicable to any of vendor under their purview.

There is also cheap 35 point crate that gives 5 random specialized cartridges.

The specialized cartridges are rather obvious, as name of vendor they refill is in the name.
Broad cartridges refill the following:
Food - All snack/food vendors
Drinks - All drinks vendors, soft and hot both, except Booze-o-Mat
Clothing - All departmental wardrobes and all laundry room loadout machines except electronics one
Technical - Vendomat, YouTool and Electronics Loadout machine in laundry rooms
Wardrobe - All departmental wardrobes
Specialty - Work/Department specific practical vendors, such as Dinnerware, ToxicMate, SecVend, etc. Still does not include Booze-o-Mat.

Admins can spawn a special Universal cartridge that can refill any vendor.

Only the normal stock is refilled, not contraband or premium.

In theory, code allows to, in future, make specific vendors only refill partially, refill more than even their default stock or refill premium/contraband. For now its just all vendors refilling precisely their standard stock amounts.

Other minor changes:
Moves airlock vendor and airlock pills to not be tether specific (still only mapped on tether).
Renames a soft-sprited soda vendor for confusion prevention.
Removes duplicate definitions of loadout vendors in the vending file. Why were they even duplicated!


GhostActual:
Oh good, this saves me from having to do it in my PR

PR #12246, opened 2022-02-16
Fixes compile warning on main branch

Heroman3003:
Variable defined but never used. Dunno why it popped up considering it was part of polaris's refactors that are still in conflict hell.


KillianKirilenko:
Give me a few minutes and I'll fix it properly.

PR #12247, opened 2022-02-16
Fixes sofa paths

### LINKED ISSUE #12154: Corner sofas are on the wrong layer
The corner-sofa sprite is in some cases drawn on the wrong layer.

For players to be able to be on top of the sofa corner.

#### What actually happened
The sofa corner refuses to yield to our power!

#### Steps to reproduce
- (Step 1) Make sofa corner that you should be drawn on top of.
- (Step 2) Get on top of it
- (Step 3) Fail and be sat upon by the sofa.

#### Code Revision
Server revision: B:release D:
Commit: [88987e534559b58da8d770f20e54fb918becec97](http://127.0.0.1:51301/tmp7100/%3Ca%20href=%22https://github.com/VOREStation/VOREStation/commit/88987e534559b58da8d770f20e54fb918becec97%22%3Ehttps://github.com/VOREStation/VOREStation/commit/88987e534559b58da8d770f20e54fb918becec97%3C/a%3E)

ktccd0:
Sofas used to have really weird subtypes, now they correctly inherit from either the base (middle), left, right, or corner, which solves the issue with layers on the non-red corner sofas (which red sofas were exempt from due to them being the base object type which other sofas were *supposed* to inherit from).

This Fixes #12154.

### DIFF
--- a/code/game/objects/structures/stool_bed_chair_nest/chairs.dm
+++ b/code/game/objects/structures/stool_bed_chair_nest/chairs.dm
@@ -320,21 +320,24 @@
 	icon_state = "corp_sofamiddle"
 	applies_material_colour = FALSE
 
-/obj/structure/bed/chair/sofa/corp/left
+/obj/structure/bed/chair/sofa/left/corp
 	icon_state = "corp_sofaend_left"
 	base_icon = "corp_sofaend_left"
+	applies_material_colour = FALSE
 
-/obj/structure/bed/chair/sofa/corp/right
+/obj/structure/bed/chair/sofa/right/corp
 	icon_state = "corp_sofaend_right"
 	base_icon = "corp_sofaend_right"
+	applies_material_colour = FALSE
 
-/obj/structure/bed/chair/sofa/corp/corner
+/obj/structure/bed/chair/sofa/corner/corp
 	icon_state = "corp_sofacorner"
 	base_icon = "corp_sofacorner"
-	corner_piece = TRUE
+	applies_material_colour = FALSE
+	//corner_piece = TRUE //No need to re-state this, as the parent object already handles corners
 
 //color variations
-
+//Middle sofas first
 /obj/structure/bed/chair/sofa
 	sofa_material = "carpet"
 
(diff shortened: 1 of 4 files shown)


ktccd0:
I didn't touch benches, despite them also having this weird subtype issue, because they don't suffer from the layering-bug and I don't feel comfortable touching their weird mutable appearance layering I've no idea what it does.
Frankly, I wanna fix the source of this spaghetti nonsense in materials but that's beyond the scope of fixing the reported issue and would be better for another PR.

ktccd0:
uhm ok, something went wrong with the integration test..... don't merge yet I forgot to save my last edit!

ktccd0:
This hurts my programmer soul, but at least this compiles and isn't explicitly buggy (although anyone touching sofa code in the future is in for a headache-inducing surprise when pews, benches and corporate sofas begin showing unpredictable behaviours).

Heroman3003:
> This hurts my programmer soul, but at least this compiles and isn't explicitly buggy (although anyone touching sofa code in the future is in for a headache-inducing surprise when pews, benches and corporate sofas begin showing unpredictable behaviours). EDIT: Well, it seems that didn't matter in the end, as sofas exist somewhere on a map anyway, so I will need to learn how to edit a map before I finish this PR.

It appears the issue arose on one of PoIs, so you'll need to look through files for those for sofas.

ktccd0:
ok, actually, the corp sofas and benches and pews having their own subtyping nonsense doesn't cause errors, but changing them wouldn't really solve anything either.
This just reveals a more inherent problem with sofa-code: The fact that directions and style of sofa-piece is determined by object subtype instead of using the already-existing variables. It's entirely possible to make a sofa determine what sprite to use by simply using the direction and placement of nearby sofas to select the appropriate icon, the only thing necessitating subtypes would be colours and styles of sofas (blue, red, pew, bench, corporate, etc.).
But doing that is beyond the scope of THIS PR so I should just be satisfied that corners no longer layer incorrectly and finish this PR. If I wanna fix the inherent problems with sofa-code, I should make a new PR to refactor them entirely, but I think I would need 2-3 more sprites for additional states the sofa could be in, in that case.

PR #12248, opened 2022-02-16
Sizegun sprite returning

Heroman3003:
Brings back both old iterations of sizegun sprites since in the end, both are better than mine and people seem to agree.

Makes 'newer' plastic design the defualy sizegun.
Adds 'old' subtype specifically for maint-found sizeguns.

Also makes the whole 'change sizegun sprite based on what size its set to' mechanics actually a thing and not a sprite leftover.


Very-Soft:
I like the 'new' sizegun in this one more than ghost actual's one, but I like their 'old' sizegun sprite more than the actual old sizegun sprite. The old sizegun sprite looks like a gun that will kill you if you get shot with it. Their one looks more like your one but with nicer shading, and I like your one more than the old one. 

It'd be cool if you two combined your PRs.

PR #12252, opened 2022-02-16
Craftable matiral  shields

TheDavestDave:
Currently uses the same parry chance as swords made out of the same material. Using some math however they do a 1/10 of the damage offensively. They currently cost 30 sheets to make

They currently use the riot sheild spirte (coloured to the material you used) Hopefully someone who can code better can get them in the crafting menu and be a bit more complicated to make. And I still need to get round to a more swash buckiler sprite for them

### DIFF
--- a/code/game/objects/items/weapons/material/shield_vr.dm
+++ b/code/game/objects/items/weapons/material/shield_vr.dm
@@ -2,8 +2,7 @@
 	name = "makeshift shield"
 	desc = "A makeshift shield"
 	icon = 'icons/obj/weapons.dmi'
-	icon_state = "eshield"
-	item_state = "eshield"
+	icon_state = "riot"
 	force_divisor = 0.07
 	sharp = FALSE
 	attack_verb = list ("attacked", "bashed", "smacked", "bonked", "spanked")
--- a/code/game/objects/items/weapons/material/shield_vr.dm
+++ b/code/game/objects/items/weapons/material/shield_vr.dm
@@ -1,0 +1,16 @@
+/obj/item/weapon/material/sword/shield
+	name = "makeshift shield"
+	desc = "A makeshift shield"
+	icon = 'icons/obj/weapons.dmi'
+	icon_state = "eshield"
+	item_state = "eshield"
+	force_divisor = 0.07
+	sharp = FALSE
+	attack_verb = list ("attacked", "bashed", "smacked", "bonked", "spanked")
+
+/obj/item/weapon/material/sword/handle_shield(mob/user, var/damage, atom/damage_source = null, mob/attacker = null, var/def_zone = null, var/attack_text = "the attack")
+	if(unique_parry_check((user*10), attacker, damage_source) && prob(50))
+		user.visible_message("<span class='danger'>\The [user] parries [attack_text] with \the [src]!</span>")
+		playsound(src, 'sound/weapons/punchmiss.ogg', 50, 1)
+		return 1
+	return 0
--- a/code/modules/materials/materials/_materials.dm
+++ b/code/modules/materials/materials/_materials.dm
@@ -109,7 +109,7 @@ var/list/name_to_material
 	var/datum/material/key = arguments[1]
 	if(istype(key))
 		return key // we want to convert anything we're given to a material
-	
+
 	if(istext(key))	// text ID
 		. = name_to_material[key]
 		if(!.)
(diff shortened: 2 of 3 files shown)


Very-Soft:
I don't really like this. I think that if the shields should probably use the other crafting system and require multiple different kinds of components, rather than most of a stack of basic materials. 

This seems like it would just make stacks of materials be one more thing that exploration needs before they will actually go and do anything. I think the shields should probably be at least in part, made with something that is not very common on the station.

TheDavestDave:
> I don't really like this. I think that if the shields should probably use the other crafting system and require multiple different kinds of components, rather than most of a stack of basic materials.
> 
> This seems like it would just make stacks of materials be one more thing that exploration needs before they will actually go and do anything. I think the shields should probably be at least in part, made with something that is not very common on the station.

yeah I tried the crafting menu and it just made a steel shield. I'd use wood amour and steel shield.  Orignally it was gonna be an amour plate, cable, and some wooden planks. But it always made a steel shield. You could use plastic or wood amour and it would be still a steel shield

people don't take the current explo sheild as they don't view it as "good enough". As is I don't think there's a lot of ways to make a better one than other than maybe plasteel or durasteel. Durasteel can already be used to make machettes and knifes better but no one really demands mining for that ether.

code/game/objects/items/weapons/material/shield_vr.dm
```diff
@@ -0,0 +1,16 @@
...
+	name = "makeshift shield"
+	desc = "A makeshift shield"
+	icon = 'icons/obj/weapons.dmi'
+	icon_state = "riot"
+	force_divisor = 0.07
+	sharp = FALSE
+	attack_verb = list ("attacked", "bashed", "smacked", "bonked", "spanked")
+	w_class = ITEMSIZE_LARGE
+
+/obj/item/weapon/material/sword/handle_shield(mob/user, var/damage, atom/damage_source = null, mob/attacker = null, var/def_zone = null, var/attack_text = "the attack")
```
Hatterhat, line 11: overwrites shield handling for all material swords and subtypes, not just the shield, due to inheritance; consider extending to `/obj/item/weapon/material/sword/shield/handle_shield`. or repath it

code/modules/materials/materials/_materials.dm
```diff
@@ -347,7 +347,8 @@ var/list/name_to_material
 		new /datum/stack_recipe("[display_name] armor plate insert", /obj/item/weapon/material/armor_plating/insert, 2, time = 40, on_floor = 1, supplied_material = "[name]", pass_stack_color = TRUE),
 		new /datum/stack_recipe("[display_name] grave marker", /obj/item/weapon/material/gravemarker, 5, time = 50, supplied_material = "[name]", pass_stack_color = TRUE),
 		new /datum/stack_recipe("[display_name] ring", /obj/item/clothing/gloves/ring/material, 1, on_floor = 1, supplied_material = "[name]", pass_stack_color = TRUE),
-		new /datum/stack_recipe("[display_name] bracelet", /obj/item/clothing/accessory/bracelet/material, 1, on_floor = 1, supplied_material = "[name]", pass_stack_color = TRUE)
+		new /datum/stack_recipe("[display_name] bracelet", /obj/item/clothing/accessory/bracelet/material, 1, on_floor = 1, supplied_material = "[name]", pass_stack_color = TRUE),
+		new /datum/stack_recipe("[display_name] shield", /obj/item/weapon/material/sword/shield, 30, on_floor = 1, supplied_material = "[name]", pass_stack_color = TRUE),
```
Hatterhat, line 351: if you wanted to make the crafting recipe a bit more involved, look into how powered crossbows are built with the frame and additional items being added

code/game/objects/items/weapons/material/shield_vr.dm
```diff
@@ -0,0 +1,16 @@
+/obj/item/weapon/material/sword/shield
+	name = "makeshift shield"
+	desc = "A makeshift shield"
+	icon = 'icons/obj/weapons.dmi'
+	icon_state = "riot"
+	force_divisor = 0.07
```
Hatterhat, line 6: if you add `edge = FALSE` below here, force calculations pull from the weight of the material and not the hardness i.e. baseball bats - you'll have to remember to adjust the force divisor too

code/game/objects/items/weapons/material/shield_vr.dm
Hatterhat: also at first glance doesn't have blockchance scaling with any stat from the material, just the flat `prob(50)` in the line below

code/game/objects/items/weapons/material/shield_vr.dm
```diff
@@ -0,0 +1,16 @@
+/obj/item/weapon/material/sword/shield
```
Hatterhat, line 1: why does it specifically have to draw from the sword subtype of material weapons? define a new type, e.g.  `/obj/item/weapon/material/shield`

Hatterhat (changes requested):
a question to be proposed is "who does this PR really serve," and if the response is "just me and nobody else" after brief introspection then it's probably not a great PR

it's good practice to remember that you're not just coding for yourself, you're coding for a server and/or its downstreams, who may or may not have to handle your future ill-thought-out feature bloat if it can really be called that

Heroman3003:
> people don't take the current explo sheild as they don't view it as "good enough". As is I don't think there's a lot of ways to make a better one than other than maybe plasteel or durasteel. Durasteel can already be used to make machettes and knifes better but no one really demands mining for that ether.

I think this just goes to say that explo has no real use for shields in the first place, not excuse to make newer, stronger and more OP ones just to play a medieval knight in plate armor, with sword and shield, in what's a futuristic space game. Explo melee meta already has enough negative connotations just as a concept (cough, marble spears, cough).

KillianKirilenko:
I'm having a hard time seeing the point myself. If the exploration shield isn't being used as-is, why would these be used instead?

Ignoring that, *30* sheets of any material is nuts.

My advice? Take this one back to the drawing board and sharpen your skills, then try again when you have a better grip on things.

TheToaster98:
I didn't want to ask, because I don't want to discourage you from continue coding, but I need to ask at this point: _Why?_
By now, I just feel like you are adding content that you want, only for you, only for this specific character, only for a specific roleplay you want to develop.
I haven't seen a single person apart of Lucky ask for a sword crate, or wear the armor, neither it makes IC sense for NT to have both chainmail armor AND an easily obtainable sword, that just happens to need explo access, even though almost all the dangerous stabby/shooty things need security access.
And now shields? 
It feels like you just want a fluff item without the restrictions applied to fluff items, and it really makes me wonder what the point of all these items is.

PR #12256, opened 2022-02-17
Sizegun Revamp 2022

GhostActual:
Pulled from the dead Armory Revamp from last year.

Decided to make them into a separate PR.
I know Heroman (https://github.com/VOREStation/VOREStation/pull/12248/files) has his own sizegun PR up so I decided to let this one ride and see which one the maintainers/admins want - the old sprites or these new sprites. Up to them.

sizegun_showcase


Very-Soft:
I like the old variant in this one more than the other PR, but I like the dinkier snubbier new sizegun from heroman's one more (because your one, while it's a very nice sprite, looks like a glock with some plastic on it, and I would really prefer if the sizeray didn't look like it was going to kill you)

Heroman3003:
Feel free to just pull the plastic sizegun sprite from either my PR or older files and once you apply I will close mine. This compromise is more than acceptable for me, as while I agree with soft on 'new' sizegun sprites, the OG ones definitely look better here.

PR #12262, opened 2022-02-17
primative-shields

TheDavestDave:
a primitive shield, perfect for the next knights at castle mystic event.

it's much less cool, there's no metal ones. but it's literally the only low tech shield in our code right.

No really all other shields are made of plastic like the riot shield, or energy like halo jackles.

anyway it has 10 lower block chance than the explo shield and 20 less than riot. It's planks off wood and a bucket tied together. I think it should make the "bad" explo shield look good in the blocking department


TheObserver-sys:
One wonders why we need this, when we **have** riot shields and "terrible" exploration shields. What possible usecase does a shield like this even serve? It's not like we have revolution on rotation, so it can't be for tiders. Most jobs that could need shields wouldn't need to ask hydroponics to make them wood. Chaplain's also aren't liable to go full on crusader, as we don't have cults, so that is also right out. This just.. feels like a fluff item with a shoddy justification, much like the sword crate. Hell, I'd be happy with making the only way to get broadswords being to acquire a cultic sword and purifying it with a bible in the chapel, giving those accursed things an actual purpose outside screenshake and taking up a useless slot in the artifact pool.

TheDavestDave:
why use any of the primative items in the crafting menu because morden tech exists?  Why use make bow and arrow when gun? why make a flint knife when steel one exist? Why have makeshift amour in the game when riot suit exists?

It's almost as if there somekind of theme being made. Almost like there's a primative low tech one. As no historian I could well be wrong but I'm pretty sure shields were invented long before bows were, that are ALSO in the primative crafting menu. It seems odd that there's a void there. But you know I'm sure there's every possibilty no one thought to get a chunk of something, pick it up, and hide behind it before the invention of laser swords

Heroman3003:
> why use any of the primative items in the crafting menu because morden tech exists? Why use make bow and arrow when gun? why make a flint knife when steel one exist? Why have makeshift amour in the game when riot suit exists?
> 
> It's almost as if there somekind of theme being made. Almost like there's a primative low tech one. As no historian I could well be wrong but I'm pretty sure shields were invented long before bows were, that are ALSO in the primative crafting menu. It seems odd that there's a void there. But you know I'm sure there's every possibilty no one thought to get a chunk of something, pick it up, and hide behind it before the invention of laser swords

The primitive items do not exist to replace or even be viable alternative for onstation use, they exist for possibility of tribal RP/mechanics for offstation characters which is more a concept right now due to quite a few missing aspects.

Very-Soft (changes requested):
It should probably require a wooden bucket instead of a regular bucket, and perhaps plant fibers, something made from plant fibers, or fur instead of cablecuffs.

PR #12264, opened 2022-02-18
Buffs Morphs

Very-Soft:
Allows morphs a couple of neat new things

Resting:
While resting and morphed, morphs do not 'look strange'. 

If a morph stops resting while someone is on their turf, they may instantly, stealthily eat the person. (If prefs allow)
(Basically straight stolen from protean blobs)

Take over prey:
A morph may choose to attempt to completely take over their prey's form, actually assuming control of their prey's body. 

This interaction allows the morph to pick any non-simple mob in any of their bellies that has a ckey. The morph gets a confirmation prompt, and the prey gets a double confirmation prompt. The prompt can only appear if the prey has the preference for allowing body impersonation enabled. 

For lack of an easy better option, this interaction is undone by way of the transformed morph ejecting the prey from their belly. Instead of actually ejecting the prey, it will restore both to their original conditions. (And THEN you can eject them if that's what you want to do) There is text explaining this to the morph when they succeed at the takeover.

OOC escape does also work with this, despite its complicated interaction. OOC escape returns both parties to their original bodies and undoes the effects of the interaction, and ensures both parties are on the floor.

Morphs can also now change their color
Also their health has been nerfed significantly. I think they should be fast OR tanky, not both, and I think being fast enough to run away if caught makes more sense than them being tanky enough to win a fight.


Verkister:
Looks like you forgot to rename the dead icon state.

Very-Soft:
> Looks like you forgot to rename the dead icon state.

o shid you right

PR #12280, opened 2022-02-19
adds avida's clothes and alt bnnuy ears

Screemonster:
Adds a set of custom clothes: https://forum.vore-station.net/viewtopic.php?f=26&t=2100
The hat changes its icon depending on whether the new wearer has compatible ears, using a generic icon if they don't.

Adds a variant of the bnnuy ears, creatively named bnnuy ears 2, that's had the tufty fluffy bits trimmed off the top for a more smoothly tapered bnnuy experience.


Very-Soft:
bnunny

PR #12289, opened 2022-02-19
hiden storage book

TheDavestDave:
Did you know when a chaplin spawns the game changes all bibles to their holy book? And has some gibbing code.

Well this is just a box with smaller space, but looks like a book.  Useful if your charater has maint trash to smuggle or you charater just wants an important book to them.

It costs 3 points from the loadout menu, it's the same size as the 2 point fanny pack and is hidden/harder to spot so gets +1 like the bluespace badge of holding compared to webbing.


TheDavestDave:
I have just been informed you can turn any book into a hidden one with wire cutters

PR #12297, opened 2022-02-20
fixes hardsuit coolers, actually

Hatterhat:
no longer will synthetic hardsuit enthusiasts burn to death on v2 solely because their cooler fuckin choked


TheObserver-sys:
Thank you, based god.

Issue #8454, opened 2020-07-23
Radio communications lags the entire server


Razgriz1032:
#### Brief description of the bug
Any time a message is sent over comms, like on the common channel, including things said by players or station announcements, the game lags for a split second.

#### What you expected to happen
No lag. Normal conversation, emotes, and PDA do not cause this lag.

#### What actually happened
Lag.

#### Steps to reproduce
- Have people talking over the radio and;
- Notice any looping animations (such as scrubbers) jitter for a moment.
- Run back and forth down a hallway and every time a radio message occurs, you stop. Easiest done by ghosting.

#### Code Revision
Server revision: B:release D:
Commit: 4c340c5dac0a7afa8f5f219ec3ba812c5ed67033

Razgriz1032:
Slight update, I did a test today to disable vchat by turning off the core vchat files and such, reverting back to default interface - radio messages still lag the client. This might be due to the recent changes to communications in general.

Razgriz1032:
Another update, I tested it with people talking on supply radio, there is no lag. So it seems to be occurring over common radio.

Razgriz1032:
Tested with a communications blackout even. Seems to still lag when someone tries to communicate over common.

Razgriz1032:
I have set up macros to test in common, a couple departments, and normal chats.

Looking at the Status tab, I see CPU hanging out between 0 ~ 11 when standing around and/or moving. When I press the macro to send a radio message over common, CPU spikes to 35. When I press a macro for a message over a department radio, there is no change. When I press a macro for normal chat, there is no change.

There is a little bit of lag if I spam the normal chat fast enough, and the department chats, but that's because of vchat, and the saved message size is only 150. Setting it to minimum, 50, resolved that lag.

Even when doing a drive-by on an intercom, spamming the normal chat macros, as soon as I get within range of the microphone, it freezes.

I turned off my headset speaker, so I don't see green text, but it still lags.

Anything that produces a message over the common channel causes lag, including ATC, station event announcements, command radio announcements, etc.

PR #12298, opened 2022-02-20
Attempt to reduce sound based lag

Very-Soft:
So, I am not sure how much this will help or hurt. 

We have noticed that, there is a lot of lag that happens under heavy radio use, and Ascian noted that the playsound proc was seriously overloaded during those times. 

Notably, presently every radio in the game uses playsound to play the radio sound any time anything sends a radio message, even if the radio in question doesn't have the channel the message was sent to. 

And if you are unaware, playsound figures out who's in range of every source of sound and their distance from its source to adjust the sounds so that it sounds like it's coming from the right place in the world. 

Problem is, if the server's decently busy, that's possibly a hundred things doing that work every time any message comes over a radio.

I thought it would be better if, since the radio systems are ALREADY doing the work of figuring out who can see the radio message, that playsound doesn't actually need to do that work again for tons of different sources. 

ADDITIONALLY, the thing to determine if the sound should be played based on preferences is also a part of playsound, meaning that it STILL RUNS PLAYSOUND EVEN IF THE PREFERENCE IS OFF

So! I used byond's direct sound thing which, I THINK just makes your client play the sound and ought to use next to no extra server resources for it. 

The only downside is that we lose pitch shifting on the radio sounds, and we lose the sense of 'audio presence' for the sounds in the world. 

In short, this means that the radio sound will just, always sound the same no matter where it is in relation to your character. If you receive a radio message, the sound will play on your client.

The sound should also only ever play once for you per message, rather than stacking like it does when lots of people/radios are in the same place. 

ASSUMING THIS WORKS HOW I THINK IT DOES, this should reduce a lot of computer work for playsound on busy shifts. (And ideally, increase performance)

This change does also use and respect the same preference that already exists for radio sounds, so there will be no change in how that works.


Novacat (approved):
Hard to say how much this will improve performance, but it is worth a try.

Razgriz1032:
Ahh yes, #8454

PR #12300, opened 2022-02-20
New->Init refactor fixes

KillianKirilenko:
Fixes merge conflicts in #11826. Maybe. Hopefully.

Probably needs doublechecking and careful testing. No compile errors/warnings, no runtimes, no odd behavior on test server that I could notice with quick tests, but no guarantee *something* won't fuck up.


code/game/machinery/machinery.dm
```diff
@@ -119,17 +119,11 @@ Class Procs:
...
-<<<<<<< HEAD
 	blocks_emissive = EMISSIVE_BLOCK_GENERIC
 
-/obj/machinery/New(l, d=0)
-	..()
-	if(isnum(d))
-=======
 /obj/machinery/Initialize(var/ml, d=0)
 	. = ..()
 	if(d)
```
Heroman3003, line 126: It being isnum(d) seems like it was a vorestation change. Not sure why, but probably should keep it just in case?

code/game/machinery/doors/multi_tile.dm
```diff
@@ -7,22 +7,7 @@
 	open_sound_powered = 'sound/machines/door/WideOpen.ogg'
 	close_sound_powered = 'sound/machines/door/WideClose.ogg'
 
-<<<<<<< HEAD
 /obj/machinery/door/airlock/multi_tile/Initialize(mapload)
```
Heroman3003, line 10: Leftover (mapload)

code/game/objects/effects/alien/aliens.dm
```diff
@@ -427,9 +424,6 @@
 			visible_message("<span class='alium'>[src.target] begins to crumble under the acid!</span>")
 	spawn(rand(150, 200)) tick()
 
-<<<<<<< HEAD
-//Xenomorph Effect egg removed, replaced with Structure Egg.
```
Heroman3003, line 431: Eggs were removed entirely on vorestation side previously, I think the PR just brought them back because diff.

code/game/objects/effects/landmarks.dm
```diff
@@ -21,26 +21,16 @@
 			delete_me = 1
 		if("JoinLate") // Bit difference, since we need the spawn point to move.
 			latejoin += src
-<<<<<<< HEAD
-			simulated = TRUE
-		//	delete_me = 1
-			return
-=======
```
Heroman3003, line 28: The diff in this file doesnt' quite show it, but there is one landmark (vinestart) in this file defined below where return has appeared. Not sure if it was fixed or not, just making sure.

code/game/objects/structures/simple_doors.dm
```diff
@@ -22,12 +22,15 @@
 	hardness -= material.combustion_effect(get_turf(src),temperature, 0.3)
 	CheckHardness()
 
-<<<<<<< HEAD
 /obj/structure/simple_door/Initialize(mapload, var/material_name)
```
Heroman3003, line 25: Leftover mapload

code/game/objects/structures/simple_doors.dm
```diff
@@ -229,44 +221,15 @@
...
-	..(mapload, material_name || MAT_HARDWOOD)
-
-/obj/structure/simple_door/sifwood/Initialize(mapload,var/material_name)
-	..(mapload, material_name || MAT_SIFWOOD)
+/obj/structure/simple_door/hardwood/Initialize(mapload)
+	knock_sound = 'sound/machines/door/knock_wood.wav'
 
-/obj/structure/simple_door/resin/Initialize(mapload,var/material_name)
-	..(mapload, material_name || "resin")
+/obj/structure/simple_door/sifwood/Initialize(mapload)
```
Heroman3003, line 230: And lower down too.

code/game/turfs/simulated/wall_icon.dm
```diff
@@ -94,30 +94,23 @@
 		img.alpha = (i * alpha_inc) - 1
 		damage_overlays[i] = img
 
-
 /turf/simulated/wall/proc/update_connections(propagate = 0)
 	if(!istype(material))
 		return
 	var/list/dirs = list()
-<<<<<<< HEAD
-	var/inrange = orange(src, 1)
```
Heroman3003, line 103: I think keeping inrange at least would be better here, to not call orange multiple times down the line.

code/modules/mob/living/silicon/robot/robot_modules/station.dm
```diff
@@ -59,13 +59,7 @@ var/global/list/robot_modules = list(
 	handle_shell(R)
 
 	if(R.radio)
-<<<<<<< HEAD
-		if(R.shell)
-			channels = R.mainframe.aiRadio.channels
```
Heroman3003, line 64: Won't this break radio recalculations for AI shells or is it handled elsewhere properly?

code/modules/mob/living/simple_mob/subtypes/animal/borer/borer.dm
```diff
@@ -62,13 +62,8 @@
 
 	true_name = "[pick("Primary","Secondary","Tertiary","Quaternary")] [rand(1000,9999)]"
 
-<<<<<<< HEAD
-	if(!roundstart && antag)
-		request_player()
-=======
 	if(!roundstart)
```
Heroman3003, line 65: Antag var check is actually vorestation thing and should definitely be kept.

code/game/objects/effects/alien/aliens.dm
KillianKirilenko: not much point in deleting them then, if diffs are just gonna bring them back again

just don't use them in any of our own POIs

code/modules/mob/living/silicon/robot/robot_modules/station.dm
KillianKirilenko: note the addtimer callback to run the recalculate proc

polaris seems to have been doing just fine without it for a good year or two at least, too

code/game/machinery/machinery.dm
Novacat: Yes, this was a vorestation change to fix a bug. Should probably remain as such.

KillianKirilenko: done

should probably be upstreamed if it's a serious bug. I've flagged it as an edited line for now.

KillianKirilenko:
Right, I think that's everything, for now. No compile errors, no runtimes once launched.

PR #12307, opened 2022-02-22
Smol morph fix

Very-Soft:
When taking control of prey, if prey body is muffled, unmuffle them.


Heroman3003:
does prey body also get unabsorbed? Because if not then it might have issue with limited vision range.

PR #12319, opened 2022-02-24
Trader!

Very-Soft:
Sort of like a vending machine, but significantly more simple and more highly configurable on the fly for use in events and the like.

Still wip for now, need to make sprites for.


GhostActual:
I like this

PR #12321, opened 2022-02-25
A more elegant weapon from a more civlised age

TheDavestDave:
adds what is effectively a rebalanced sword.

instead of being in sword family it's in two handed, and allows some tricks.

Like most two handed weapons it gains a damage boost if wield. At 0.3 sharpness it will be as powerful as machete of similar material. Minus the cleave effect. If held with something in your other hand it does 0.1 sharpness, a third of the machete.

Also the parry ability ONLY works if wielded. So you can't use it with a shield or a second parry item. You can't even parry if you have a gun or pen in the other hand.

so pros
can parry

cons
needs an empty hand to parry
if used one handed is 1/3 as powerful as stock gear
starts out as steel and needs mats to upgrade


Runa-Dacino:
If it's a two-handed weapon, please don't call it a saber. I get the "lightsaber" reference but...

Why not call it something that was actually used with two hands?

TheDavestDave:
claymore is already taken

TheDavestDave:
it's been renamed to longsword, along with the sprite names

Very-Soft:
Would prefer if this was in a cargo crate as opposed to the explo vendor. Explo doesn't really need more kinds of weapons they don't need to talk to anyone to get.

TheDavestDave:
would you be okay with it haveing explo access, and would a 50 point crate be okay?

code/datums/supplypacks/munitions_vr.dm
```diff
@@ -57,4 +57,16 @@
...
+	access = access_security
+
+/datum/supply_pack/munitions/longsword
+	name = "longsword"
+	contains = list(
+			/obj/item/weapon/material/twohanded/longsword=2
+			)
+	cost = 50
+	containertype = /obj/structure/closet/crate/secure
+	containername = "frontier phaser crate"
```
Very-Soft, line 69: Should probably give this a less copy pasta name

Very-Soft:
> would you be okay with it haveing explo access, and would a 50 point crate be okay? it's getting far away form the civ iv explorer weapon than I imagined

The other swords you added are armory access locked, I don't see why these would be different.

TheDavestDave:
theese ones are a lot weaker. Anyway it's done and tested

PR #12323, opened 2022-02-25
Fixes arcades not being deconstructable

Heroman3003:
Was this really an issue for a whole year that nobody ever reported? Seriously?


Screemonster:
What happens if there are special prizes in it?

if I fill an arcade machine with ~geese~ funky loot then I don't want people just wrenching the thing apart to get them without playing

Heroman3003:
> What happens if there are special prizes in it?
> 
> if I fill an arcade machine with ~geese~ funky loot then I don't want people just wrenching the thing apart to get them without playing

I believe they just disappear.

PR #12334, opened 2022-02-26
[MIRROR] Ports high-res lobby screen system from Eris from Bay

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/8401
--------------------
Adds support for high-res images as lobby images. Currently includes nothing of the sort, but the support is there and we're no longer constrained to 490x490 images for Future Reasons.

Does lose the fade-in effect at roundstart but... Ech. It prevented you clicking anything until the image was fully faded anyway.

Also put the old-old lobby image back into rotation since we've had support for a random selection for years but never used it even though the current and old images are honestly fairly comparable... Except the old one is animated!


Very-Soft:
I've played with this a little bit to try to get it going, and, I can deffo get it going, but, doing so will remove staff's ability to change the title screen on the fly, and I don't really think that's worth giving up, since that's fun. 

So! Before this gets in, I'd like us to get an admin verb or some such that will allow us to change it.

PR #12344, opened 2022-02-27
New ringtones

BonniePandora:
Adds in a few new tones for PDA ringers, mostly engineering/technical based.

There's now "dust", "rad", "servo" and "spark" alongside all the existing ones, figured a little bit of variety wouldn't do any harm.


code/modules/pda/pda.dm
```diff
@@ -34,7 +34,11 @@ var/global/list/obj/item/device/pda/PDAs = list()
 								"honk" = 'sound/items/bikehorn.ogg',
 								"SKREE" = 'sound/voice/shriek1.ogg',
 								// "holy" = 'sound/items/PDA/ambicha4-short.ogg',
-								"xeno" = 'sound/voice/hiss1.ogg')
+								"xeno" = 'sound/voice/hiss1.ogg',
+								"dust" = 'sound/effects/supermatter.ogg',
+								"spark" = 'sound/effects/spark4.ogg',
```
Very-Soft, line 39: should be sparks

code/modules/pda/pda.dm
```diff
@@ -34,7 +34,11 @@ var/global/list/obj/item/device/pda/PDAs = list()
 								"honk" = 'sound/items/bikehorn.ogg',
 								"SKREE" = 'sound/voice/shriek1.ogg',
 								// "holy" = 'sound/items/PDA/ambicha4-short.ogg',
-								"xeno" = 'sound/voice/hiss1.ogg')
+								"xeno" = 'sound/voice/hiss1.ogg',
+								"dust" = 'sound/effects/supermatter.ogg',
+								"spark" = 'sound/effects/spark4.ogg',
+								"rad" = 'sounds/items/geiger/high1.ogg',
```
Very-Soft, line 40: should be sound

BonniePandora: Fixed in the new commit

code/modules/pda/pda.dm
BonniePandora: This is fixed now too

PR #12253, opened 2022-02-17
Donk Co. Toys

GhostActual:
Includes several toy modifications and additions that were purposed in the past as well as the addition of a new card game from TG. There are some minor references to external media such as Fallout or Doctor Who however nothing that breaks the meta.

-Wiz-Off Card game
-Action figures
-Sonic Driver (Sonic Screwdriver - works like a real screwdriver but has no advantages over them)
-Sus Co. Plushies
-New Plushies
-Reorganization of `toys_vr.dm`
-Addition of new Laser Tag set based on a Star Trek esq. theme (currently not available to order, meant for events)
-Revamp of current Laser Tag set

A while back I promised Donk Soft toys in the Armory Revamp that died. Those toy guns are added here along with all the stuff to add them into the game such as a vendor that can be ordered through cargo and a crate that can be ordered through cargo.

Included in this PR are new Ammo Boxes - three to be exact - Cap Box (for the Cap Gun), Foam Darts, and Riot Foam Darts. You can either click on the box to dump all the contents or Alt-Click on it to extract one bullet at a time. You can hold the box in a pocket so you can extract bullets, this is to prevent the annoying task of pulling the box into your hands, opening the box, realizing you have to drop your gun to free up a hand, or throw the box down to pull stuff out. This way you just need to switch hands and Alt-Click the box to pull out a dart.

_**Donk Co. to release new toy line!**_

_Donk Co. and affiliates have announced that a new line of toys will be coming to the Virgo system. After a huge success in the Core Worlds, officials at Donk Co. have confirmed that these toys will become available to other systems in the frontier. A spokeswoman for Donk Co. had this to say "We are overjoyed that we get to share our new toys with those on the farthest reaches of known space!" The launch date for these new toys has yet to be announced as Donk Co. is still in the process of securing transport for the new toys, however, it is likely the eager masses of the Virgo system and beyond will not have to wait much longer._

PR #11636, opened 2021-09-29
Armory Revamp 2022

GhostActual:
What it says on the tin. Revamps weapons - specifically ranged weapons. New sprites have been made for just about every gun that is inside `gun.dmi` as well as the addition of some new ones and new mechanics that might see their way into gameplay in the future. I would list some sprites I made for this PR but there are quite literally too many to post.

The PR isn't truly done, but I need to find a stopping point or I am going to lose my mind.

PR #12353, opened 2022-03-01
Ammo Boxes

GhostActual:
Added in new Ammo Boxes that replace the old ammo box system.
There are other, minor sprite changes and some code changes/organization in a vain attempt to combat the spaghetti.
Not much to this one, but it was a pain due to the map changes.


Very-Soft:
did you actually test and make sure all this works as you expect it to?

GhostActual:
Yes, this is originally from the Armory Revamp #11636, it uses the same code as the Ammo Boxes that were merged with Donk Co. Toys #12253

I'll run it through a test to be 100% sure, but as it stands it should work as intended

Very-Soft:
Lemme know when you've tested it. With big PRs like this it's difficult for me to see by skimming through if something's not as it should be, so, would like to ensure that it's been well tested and all that.

GhostActual:
It works

PR #12358, opened 2022-03-02
New PDA type, new sprites for other types & overlays!

BonniePandora:
Ports across the flick-open/clamshell looking PDA style from Citadel, as well as updating a few others they had that were missing from here, mostly transparent ones and the neko-style of PDA.
Also switched out a few of the icons used for exploration PDA's, to help them feel a bit more distinct, rather than sharing sprites with detectives (explorers) & bridge secretary's/HoP's/CCO's (field medics).
Some versions of HoP sprites were updated, as well as Captain sprites, details listed below.

Changelog:
PDA "Slider" style added
Overlays added to PDA's for pen & ID being inserted and for their flashlights turning on & off - TODO: sort the overlays so the pen one is applied upon spawning.
Transparent model PDA sprites added for; rugged, default
Neko model PDA sprites added for; old, rugged, standard
New default style Captain PDA sprite added, HoP recieves the old Captain one
New HoP old stype PDA sprite added
Explo medic PDA type changed from pda-h to pda-gene
Explorer PDA type swapped from pda-det to pda-lawer-old(the PF's sprite)
Pathfinder PDA type swapped from pda-lawer-old to pda-transp

Example of what the new style looks like -
aaaaaaaa


Cyantime:
Was that called clamshell where it came from, because that's definitely not a clamshell design, it looks like it slides out to the side rather than hinging open.

BonniePandora:
> Was that called clamshell where it came from, because that's definitely not a clamshell design, it looks like it slides out to the side rather than hinging open. Extremely glad to see expansion of the clear plastic cases to other styles

No, it was called "PDA-alt" on citadel I think, it's been some time since I last played on it so I don't remember exactly.
I'd asked a few folks whilst working on this whether "Flick-out" would be a good name for it and was suggested clamshell for it's name instead so *shrug
I could probably change it but it's hardly the end of the world, not like the PDA's state can be changed to be compacted.

Issue #12346, opened 2022-02-27
wearing the classic size gun on your belt slot turns you into a size gun


TheDavestDave:
#### Brief description of the bug
putting a size gun on your belt has the full sized sprite when worn

#### What you expected to happen
a mini sprit or not sprite

#### What actually happened
I have merdged with the size gun

#### Steps to reproduce
- (Step 1) get classic size gun sprite
- (Step 2) put on belt
- (and so on)

#### Code Revision
- (Found using the "Show Server Revision" verb under the OOC tab.)
 

#### Anything else you may wish to add:
- (Location if it's a mapping issue, screenshots, sprites, etc.)

PR #12363, opened 2022-03-03
Sizegun Changes

GhostActual:
Changes around some parts of the Alien Sizegun such as adding a catalogue entry and changing the color of the beam into the `darkmatter` beam in order to better match the color scheme.

Also meant to fix https://github.com/VOREStation/VOREStation/issues/12346


Heroman3003:
Alien sizegun is intentionally event-only to my knowledge, which is why it doesn't have lore or isn't accessible (and why it is a subtype of 'admin' sizegun). It's meant to basically be 'ic' version of admin sizegun spawned for when someone is playing abductor or such.

GhostActual:
> Alien sizegun is intentionally event-only to my knowledge, which is why it doesn't have lore or isn't accessible (and why it is a subtype of 'admin' sizegun). It's meant to basically be 'ic' version of admin sizegun spawned for when someone is playing abductor or such.

Undid the loot spawn

PR #12364, opened 2022-03-03
item bank - blacklists Alien Circuits

PastelPrinceDan:
storing those only allows people to get precursor tech roundstart, instead of having someone do xenoarch for it. 
that's stupid and goes against what the bank is for, especially when there's like 5 randomized ones that can be printed in rnd, allowing people to just put a different subtype back in once they've used the one retrieved.


Very-Soft:
Thanks for looking out for this stuff

PR #12369, opened 2022-03-04
Reduces the duration of spider footstep sound

Heroman3003:
From 5 seconds to about 0.8 seconds to actually match the average spider step time. No more chittering on top of chittering stacking when spider is player controlled or rushes towards objective.


Very-Soft:
Lemme review this here shortly.

PR #12385, opened 2022-03-06
Ports Honk-Hound module from downstream as whitelisted borg module

Heroman3003:
By extremely unpopular demand, this hellish beast from Chomp can be here now. Whitelisted, same as clowns are.

The module is mostly identical to servicehound, but has its own borgbelly, bikehorn and special confetti cannon that can fire pies, confetti or banana peels.

Also ports confetti cannon and confetti balls to make it work.

Most credit goes to Verkister for coding and Valekin for sprite.


Screemonster:
I'mma be real, the reason why clowns are whitelisted is at its core the reason why dogborg players have such a negative stereotype associated with them - namely, that people ignore what it's actually meant to be and instead leap on as "le funy grief licence" or "the cute pupy that vores things" and frankly I'm dreading the kinds of people who would actually apply to play this thing and the tantrums that would result when they're told no.

Skysteak:
This is an absolutely understandable concern which justifies keeping clown as a whitelisted job. However whilst a whitelist isn't some immutable barrier against bad actors, it both provides an explicit expectation of certain standards and a means as to which to retract access if needed.

In the case of clownborgs this is also applying a whitelist atop of a silicon job role, which whilst not held to the same intense scrutiny as some servers, is still expected to follow the spirit of their lawset and at least be benign to the crew. If playing a clownborg induces a flagrant disregard for this, it not only affects their right to play the whitelisted module in question, but also the right to play silicons in general.

Screemonster:
the other thing is that this would be the first borg that's outright "there is no regular borg equivalent of this"

there isn't an entertainer borg, or a clown borg, or a mime borg

so we're adding, effectively, a whole new silicon job

_and it's a goddamn pupy that vores things_

Skysteak:
> the other thing is that this would be the first borg that's outright "there is no regular borg equivalent of this"
> 
> there isn't an entertainer borg, or a clown borg, or a mime borg
> 
> so we're adding, effectively, a whole new silicon job
> 
> _and it's a goddamn pupy that vores things_

I will say that this is not for a lack of them out there! This isn't the most update to date image but you can see a few here:

https://i.imgur.com/r3DvOwe.png

I wouldn't be opposed to them being added in either, though that is beyond the scope of this PR.

BigBababooey:
I've only ever played on Virgo tbh. So my exposure to clowns is limited, to say the least, and frankly, I just don't find them funny. The only clown I ever notably enjoyed as a clown and not just as a generally funny person was COLIN, and that's at least in part cause I enjoy puns. I get they're a big part of SS13 culture and are something iconic, but I often see you, Heroman, be the one who argues we shouldn't be doing stuff just because other servers do it, so it's a small bit surprising to see this had your enthusiasm behind this to try to make it happen.

To be entirely honest I've for the most part not really seen a clown and went "this added to your funny" they either were already so, or they weren't and now the clown just added a bit of a tryhard aspect, as if I was being told oocly I should find them funny. So, the more I think about this, the more I can't personally see what it will add other than a big obnoxious sprite that screams "why aren't you laughing?" at me.

A funny borg can be funny without a clown module. Bike horns are just not very funny, even as a punchline to intentionally bad jokes, I **do not** want to see what people do with an on demand pie cannon, because the only outcome I can see at all is being annoying, and I shouldn't have to spell out why banana peels get old. 

It's built in by design to either run a joke that wasn't even that funny except in instances of shared neuron small brain goof around moments into the ground, or to inevitably have people just have to be an asshole to you and "law 2" you out of your ability to do the thing anyways.

I'd consider, maybe being chill with someone SCA applying for clown and being given the clown borg as a custom sprite for a service borg, but the clown specific stuff just leaves me wary. And even that leaves me wary too that it's not just going to be a confusing thing to have around. I don't really like having to tell borg players to leave me alone, because I try to treat them fairly and just the same as anyone else, but I don't see how I can treat this one like a normal person and idk if I want to expend the brainpower trying to figure out how.

YuskoTheCoyote:
I am support of this PR mostly thanks to the fact that such a role will be whitelisted, same as clowns. I personally do not share the same fears as the other staff. With the role being behind a whitelist, we have much more regulation over who we determine could be trusted to play said role properly. 

We do not get much applications for clowns to begin with, so my assumption is that it would be the same with these kinds of borgs. Few people would apply, and we as the staff can put in extra scrutiny toward the people who do apply to ensure that no issues arise from the role being played improperly.

Heroman3003:
I think whitelisted stuff is fine to have. Clowny aesthetics are clowny, and while extra tools are rather grief-capable, I doubt role will actually see any significant demand at all, dogborgs are niche, clowns are niche, overlap is minimal and easily filtered.

lm40:
i do not support mostly cuz i don't like adding stuff like this prematurely. if someone gets a sca accepted for clown dog borg or w/e then fine, merge, idc, but it just grants the idea unnecessary credibility to add it first.

Heroman3003:
> i do not support mostly cuz i don't like adding stuff like this prematurely. if someone gets a sca accepted for clown dog borg or w/e then fine, merge, idc, but it just grants the idea unnecessary credibility to add it first.

Tbh, I did want to sit on it until an app gets approved before trying to put it through, main reason I sent it 'early' was due to the fact that I just wanted to start doing some other minor fixes and didn't want to lose track of this branch so I just put up the PR. I did't expect it to generate this much controversy.

Skysteak:
> i do not support mostly cuz i don't like adding stuff like this prematurely. if someone gets a sca accepted for clown dog borg or w/e then fine, merge, idc, but it just grants the idea unnecessary credibility to add it first.

As Hero mentioned, this was (and still) is basically the plan in question,

pontifexminimus2:
As a loremaster and headmin Imma say it straight up.

I don't want those.

Verkister:
lol

User-U-U:
> on why clowns are whitelisted is at its core the reason why dogborg players have such a negative stereotype associated with them - namely, that people ignore what it's actually meant to be and

Still have yet to see any actual example of "Cute puppy vore things grief license"

Very-Soft:
Largely overwhelmingly negative staff reaction to this. I personally think this is more likely to lead to annoyance for many people, and fun only for the few who play this role. I think that a clown borg module misses a large portion of what makes clowns fun and funny, primarily being the counterplay. 

Clowns are funny in large part because, if they're doing the sillies mechanics wise, you can play back at those things. You can push an annoying clown over and steal their lunch money and swirlie them after they put space lube under the doors. Both sides laugh, no one dies. (Probably)

You can't slap a clownborg and steal their tools if they're being annoying about it. Your only real options are to use a flash as a temporary measure, and lethals as a permanent measure. Meaning, MOST PEOPLE will just have to sit there and take it, and not be able to do anything back. And that's not fun for anyone except the borg.

SO! I think this would require so specific a person, that it's almost not worth having around, and I think this will only really attract people who want to do funnies with little to no means of counterplay. And that's exactly the kind of person I would not want behind one of these. 

This module can come up again if we see someone who does fit it, but I do not think that in its present incarnation, it would be a benefit to the server.

PR #12403, opened 2022-03-08
Small PDA hotfixes

BonniePandora:
Starts PDA's with the pen-inserted overlay now, as well as renaming the slider style PDA's .dmi to one that fits it, rather than being called "pda_clam.dmi"


Very-Soft:
clam

PR #12404, opened 2022-03-09
Adds the hazard undersuit to the loadout menu

BonniePandora:
Because someone did a fucky-wucky & forgot to add it when slapping them up the first time round it seems.
Sorted now though


Hatterhat:
whoops

PR #12411, opened 2022-03-10
Fixes item digestion vore stat count

Verkister:
Makes the stat count the things when the item is actually digested instead of counting every tick of partial digestion that may have affected it.


Very-Soft:
Thank!

PR #12435, opened 2022-03-13
Change the default map to virgo minitest.

lbnesquik:
## About this pull request
The default map that is used by the game

## Why is it good for the game
Faster compiles for the coders. Much faster. 
I don't know how the map switch between tether and the new map work, but it should probably work fine?

### DIFF
--- a/vorestation.dme
+++ b/vorestation.dme
@@ -4108,6 +4108,6 @@
 #include "maps\submaps\space_submaps\debrisfield\debrisfield.dm"
 #include "maps\submaps\surface_submaps\wilderness\wilderness.dm"
 #include "maps\submaps\surface_submaps\wilderness\wilderness_areas.dm"
-#include "maps\tether\tether.dm"
+#include "maps\virgo_minitest\virgo_minitest.dm"
 #include "maps\~map_system\maps.dm"
 // END_INCLUDE


lbnesquik:
image

Lame

Heroman3003:
> image
> 
> Lame. Can we work around this, or just rename the map? It would really save hours of time over each of our coders year.

The way we handle releases and merges means that whatever map is going to be on master, will also be on live. If coders want faster compiles, they can just change the map for duration of their testing, it's not hard at all.

lbnesquik:
> > image
> > Lame. Can we work around this, or just rename the map? It would really save hours of time over each of our coders year.
> 
> The way we handle releases and merges means that whatever map is going to be on master, will also be on live. If coders want faster compiles, they can just change the map for duration of their testing, it's not hard at all.

Yeah, but we have to do it with every new project. That's something that take time and micro every time.

Heroman3003:
> > > image
> > > Lame. Can we work around this, or just rename the map? It would really save hours of time over each of our coders year.
> > 
> > 
> > The way we handle releases and merges means that whatever map is going to be on master, will also be on live. If coders want faster compiles, they can just change the map for duration of their testing, it's not hard at all.
> 
> Yeah, but we have to do it with every new project. That's something that take time and micro every time.

It's unticking one file and ticking another, which is way less time than it would be for the maintainers to make separate prs to change map to actual server map and then back to test/example every time a release PR would need to be pushed.

Very-Soft:
It is not hard to swap your personal project over to minitest. You only ever have to update the DME when something changes with it. It's quite easy to just not reset your DME when you start a new project. What you are doing here would mean minitest would be live on the main server, and would also break our way of '''map rotation'''

PR #12446, opened 2022-03-16
[WIP] Adds item overlays to sec & toolbelts

BonniePandora:
Or atleast, in theory it will.
Current issues with this are some overlays not displaying correctly, and the fact that the code I've got for this is incredibly jank. 

The biggest problem is that in testing the items put in to adjust the icon in the players inventory call the item_state name, but it doesn't seem to recognise those in the relevant dmi. So red wirecutters will display on the belt's sprite (with white handles, for whatever reason?), but yellow & blue simply don't show, despite it trying to apply "cutters_yellow" or "cutters_blue" & the corresponding sprites being named as such. Similar issue with screwdrivers and any other subtype, so the basic crowbar displays, but not the red one, or the jaws of life, etc etc.

I'd appreciate a measure of help with this, given it would be nice to have & opens up plenty of further opportunities for overlaying on other belts as well.

Here's a current WIP using the tools that _do_ display, albeit some of them incorrectly.
image


Cameron653:
Doing some PR cleanup and this has been sitting for six months now. If more changes are made and the PR is made ready for merge, do feel free to re-open this PR.

PR #12447, opened 2022-03-17
Fixing broken TGUI on DNA Modifiers

FartMaster69420:
Looks like a tooltip was trying to display in a bad position, that has now been fixed.

Geneticists Rejoice.


Very-Soft (changes requested):
conflicts

FartMaster69420:
Conflicts should now be resolved.

PR #12451, opened 2022-03-18
Promethean Blobform

FartMaster69420:
Creates a blob form prometheans can drop into.

Mechanically they're pretty weak and are limited to standard simplemob passive heals instead of promethean healing. The idea here is blob form is for RP only, with attacks dealing no damage and everything hurting the little blobos pretty easily. They also lack the ventcrawl ability, as that's an easily abused power.

Prometheans gain a new verb, 'Toggle Blobform', this lets them collapse into a blob.

Prometheans in blob form have several verbs, 

'Toggle Width' which allows for changing between a fat slime or skinny slime sprite.

'Toggle Shine', this gives the prommie the shine of a metal slime.

'Select Body Colour', works just like it does in humanoid form, allowing body color to change.

Hats will also carry over between humanoid and blob form, letting the little fellas look dapper.

Finally, they have an array of chat emotes that can set their sprite's face icon. This isn't an addition I've added, just the result of inheriting off of slime mobs who already had this.


FartMaster69420:
Whoops, serious bug found with rads.

FartMaster69420:
Should be good to go now.

code/modules/mob/living/carbon/human/species/station/prommie_blob.dm
```diff
@@ -0,0 +1,466 @@
...
+		rad_glow = CLAMP(rad_glow,0,250)
+		set_light(max(1,min(5,rad_glow/15)), max(1,min(10,rad_glow/25)), color)
+	else
+		set_light(0)
+	return
+
+//Constructor allows passing the human to sync damages
+/mob/living/simple_mob/slime/promethean/New(var/newloc, var/mob/living/carbon/human/H)
+	..()
+	if(H)
```
Very-Soft, line 87: These special promethean slimes should probably delete themselves if they don't get passed a human form.

FartMaster69420: Sure thing, should help prevent weird edge cases.

code/modules/mob/living/carbon/human/species/station/prommie_blob.dm
```diff
@@ -0,0 +1,466 @@
...
+		B.owner = src
+
+	//vore_organs.Cut()
+	if(blob.prev_left_hand) put_in_l_hand(blob.prev_left_hand) //The restore for when reforming.
+	if(blob.prev_right_hand) put_in_r_hand(blob.prev_right_hand)
+
+	Life(1) //Fix my blindness right meow //Has to be moved up here, there exists a circumstance where blob could be deleted without vore organs moving right.
+
+	//Get rid of friend blob
+	qdel(blob)
```
Very-Soft, line 463: It would probably also be good if we didn't delete the blob, and instead maybe stored it in the prommie body. Unless I'm reading things wrong, it seems like you'd need to redo any customization every time you flip flopped between them. Just as a suggestion~

FartMaster69420: Aye, on it

Heroman3003: I think easier is middle ground ,where just customization is saved as vars on species datum or something and re-applied when you return to blobform. Proteans delete their blobform too because of technical fangaling. Just a suggestion here too though.

FartMaster69420:
Okay, final push for this one. Should be good to go. There are two notable known issues, but they are not game breaking and can be resolved in future updates.

1) Hats change color to match the blob wearing them. This is also an issue with xenobio slimes.

2) Turning into a blob while walking in hotkey mode will get the movement stuck down, resulting in continued movement. This is also an issue with protean blobs, however.

Very-Soft:
did you make it so that they delete if they don't get a human form set?

FartMaster69420:
Aye, sure did. Tested with the game panel, they delete on spawn if you try to spawn a prommie blob without a prommie.

PR #12457, opened 2022-03-18
Adds Submit To Prey Trait

Heroman3003:
It's mostly identical in practical application to what Dominate Pred does, except the difference is that the pred is one initiating control switch, not the prey. They get to choose which if prey in the bellies get the offer. Otherwise, beyond slight difference in flavor texts here and there, the functionality and result is identical to Dominate Pred.

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/traits_vr/neutral.dm
+++ b/code/modules/mob/living/carbon/human/species/station/traits_vr/neutral.dm
@@ -500,3 +500,13 @@
 /datum/trait/neutral/dominate_prey/apply(var/datum/species/S,var/mob/living/carbon/human/H)
 	..(S,H)
 	H.verbs |= /mob/living/proc/dominate_prey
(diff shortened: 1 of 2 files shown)


FartMaster69420:
Should add this ability to borgs, since they can't select traits.

Heroman3003:
> Should add this ability to borgs, since they can't select traits.

That would be easy enough to do, but I am not sure how 'desirable' ability for borgs to give control to anyone would be.

FartMaster69420:
Maybe, but considering anyone can already become a borg at Robotics or at spawn I don't see this being used outside of scene stuff.

Very-Soft:
I do not really think this should be a thing for borgs, borgs are expected to play to certain rules and standards. Suddenly plopping someone in control of a borg body who isn't willing to accept those rules and standards is not ideal, and could kind of be an easy rules trap.

PR #12516, opened 2022-03-23
[WIP] Stonks

GhostActual:
This is a massive work in progress. Because the original code was removed over four years ago the UI was never ported to TGUI and thus does not work right. A new TGUI system has to be made in order for this to work properly and I can't do that honestly - I've never touched TGUI and it appears too complex for my usual standard of coding.

Regardless, at this point, the basics work, and the major stuff is ported. Figured I'd make a PR so no one else has to go through trying to find the code like I did.


ItsSelis:
Hey, I can help you out with the TGUI stuff if you want. I have been doing my own updates/additions to it recently and pretty much know how it works now :P

GhostActual:
> Hey, I can help you out with the TGUI stuff if you want. I have been doing my own updates/additions to it recently and pretty much know how it works now :P

If you can somehow get this to work, please, be my guest because this is the type of coding that is pretty much witchcraft to me

ItsSelis:
So this is a very rough result I got:
grafik
Now it only really needs to be fed with all the data and handling the act actions.

GhostActual:
> Now it only really needs to be fed with all the data and handling the act actions.

How does one go about doing that? From what I understand, this code is veeeeeery old and has actually been removed from /TG so there's no real reference on how it was supposed to function and interact with TGUI

ItsSelis:
This would be an example of passing data to TGUI:

`/obj/machinery/computer/stockexchange/tgui_data(mob/user)
var/list/data = list()

data["stationName"] = using_map.station_name

return data`

But yeah, the code is pretty old, and like the Admin Player Notes ui thing, it is required to kind of "translate" it. At least the variables are mostly the same so thats a plus :o

GhostActual:
Gonna close this and make a new PR to unbury this from the ton of mirrors.

PR #12522, opened 2022-03-24
Firearms 1

GhostActual:
Adds in new pistols and new pistol sprites as well as added a new CentCom Sidearm Box that has yet to be mapped in. Also adds in new Energy Rifle sprites, Burst Laser (now known as Energy Carbine) Sprites, and Energy Gun sprites. Also adds in new Energy Luger and PDW sprites. All new sprites were pulled directly from the old Armory PR that I did a while back.

Also removed the second gun from the HoP locker (honestly don't know why they had two guns while HoS and Captain both only had one)

Still working on this a bit, need to make sure that everything is done right.


GhostActual:
whoops, forgot to set as draft

Issue #12525, opened 2022-03-25
Pilots cannot unlock the Tour Bus Cockpit APC


Skysteak:
#### Brief description of the bug

Pilots cannot unlock the 'Tour Bus Cockpit APC' with their default ID access. They can unlock the rear APC named 'Tour Bus APC' but _not_ the one in the cockpit. 

#### What you expected to happen

For Pilots to be able to unlock the 'Tour Bus Cockpit APC' with their default starting access.

#### What actually happened

Pilots actually _cannot_ unlock the 'Tour Bus Cockpit APC' with their default starting access! 

#### Steps to reproduce
- (Step 1)Spawn in as a Pilot.
- (Step 2)Go to the Tour Bus and attempt to unlock the 'Tour Bus Cockpit APC' with your ID.
- (Step 3)Collapse onto your knees and scream at your newfound powerlessness.

PR #12529, opened 2022-03-25
Auto-Resleeving

Very-Soft:
Adds auto-resleeving!

Under normal conditions, this allows one to resleeve their character after they have died in the round. 

The auto-resleever requires a bit of time after death to be used:

-Characters who died due to vore reasons must wait 15 minutes

-Characters who died for other reasons must wait 30 minutes

The auto-resleever only requires power, there is no biomass requirement.

The auto-resleever also works for species that can ordinarily not be resleeved, since the potential for abuse there is minimal.

To use the auto-resleever, one only needs to click it as a ghost. If it is ready to revive you, it will pop up a confirmation dialogue.

Additionally, the automatic transcore notification, and the 'Cancel Transcore Notification' verb has been removed. Ghosts have a 'Notify Transcore' verb, which can be used around 5 minutes after death if they wish to notify medical.

This should eliminate false positives for medical, because no one likes resleeving people who don't need it.

This does not otherwise alter the original resleeving process or machines.

The auto-resleever for the tether is on Z1, just south of the psych office. It also features a return chute for items right next to the door. (I wanted to put it in the resleeving room's cleaning room, but it's really far into medical)

The auto-resleever for the stellar delight is in the resleeving's cleaning room.

I also fixed the access on the tourbus's APC while I was there.
fixes https://github.com/VOREStation/VOREStation/issues/12525

image


Heroman3003:
The only thing that makes me 'hmmm' at this really, is the fact that (mostly talking vore deaths here) it kind of presents a choice between waiting longer and bothering medical, which kind of seems to still present an issue when it comes to resleeving being a chore medical or extra long wait for the prey.

Also if you did fix the tourbus thing, pls mark PR as Fixes #12525 and I'll close my own for that.

lm40:
I mean IIRC when transcore timing was changed, we specifically considered 'lower transcore timing should NOT equate to dead people expecting/demanding five minute resleeve of medical'. And you can ping regardless, just if no one wants to do it then instead of bitching about medical being asleep at the wheel/furiously LOOCing/adminhelping/whatever, you can just do it yourself.

Very-Soft:
The timer being 15 mins is mainly to allow for CONSIDERING going to medical. I do not want this to just be 'the better resleeving'

it's automatic, it's slower, but not unreasonably so. If you don't want to deal with medical, or medical's not around or whatever, smol wait, then go.

Dahlular:
Pretty amazing stuff, tbh. I'd imagine the balancing works well.

Verkister:
Uhh just gonna point out that the thing underneath those flaps is a disposal chute, not an outlet.

PR #12547, opened 2022-03-26
Adds a logical consequence to milk overindulgence

drexample:
Alright, alright, I get it, everyone likes realism, but no one likes redundant code - so here's a better dairy beverage related feature.

You know how milk is supposed to make your bones stronger? Like really strong?
What if you drink too much of it? What if the bones become, y'know.
Too strong. That wouldn't be good. Your skeleton will become stronger than you.

### DIFF
--- a/code/modules/reagents/reagents/food_drinks.dm
+++ b/code/modules/reagents/reagents/food_drinks.dm
@@ -1219,7 +1219,7 @@
 	if(ishuman(M))
 		var/mob/living/carbon/human/H = M
 		if(SKELETON in H.mutations)	return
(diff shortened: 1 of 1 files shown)


Very-Soft:
Please do not open meme prs

drexample:
this wasn't a meme pr goddamn

BigBababooey:
I feel as if we live in very different realities. Also we already have this joke but at a more reasonable because milk heals brute slowly iirc.

Huenererschrecker:
Not anymore, that was removed some time ago if I remember right.

PR #12548, opened 2022-03-27
Micros can be holstered

FartMaster69420:
Allows micros to be holstered, opening a wide range of tactical tiny deployment opportunities.


BonniePandora:
I'm having nightmares of a matryoshka doll-esque stack of Security rocking up, only for each of them to pull out another member of the force & drop them.

PR #12549, opened 2022-03-27
Lets proteans crawl under and over tables.

FartMaster69420:
Considering they have the hide verb I'm assuming not being able to pass tables is an oversight.


Very-Soft:
I do not think it was an oversight though, the idea being, they're not any more coordinated or agile in their blobby form than they are in their human form, they are just better at hiding in their blob form than they are in their human form. Keep in mind, hiding is not only for tables, you can hide under all kinds of stuff, and I don't really like the idea of giving their blob for more utility than it already has, considering how strong it can be.

PR #12550, opened 2022-03-27
Autoresleever checks for whitelist

Rykka-Stormheart:
As it says.
Autoresleever will abort the spawn attempt if the client is not whitelisted for the species (as currently one can ghost, select an autoresleever, and spawn via it even on a whitelisted species they don't have access to normally.)
~Autoresleever will abort spawn attempt for species with NO_SCAN in their flags. This prevents Xenochimera/Proteans/etc from rejoining via Autoresleevers, as per whitelist expectations - Xenochimera even have specific lore stating they contaminate sleevers.~

Removed as per maintainer request here.

These changes will be applied to regular clonepod + resleeving pod eventually:tm:

Upstream port of sanity feature from https://github.com/CHOMPStation2/CHOMPStation2/pull/3938

### DIFF
--- a/code/modules/resleeving/autoresleever.dm
+++ b/code/modules/resleeving/autoresleever.dm
@@ -76,6 +76,18 @@
 
 	var/client/ghost_client = ghost.client
 	
+	if(!is_alien_whitelisted(ghost, GLOB.all_species[ghost_client?.prefs?.species]) && !check_rights(R_ADMIN, 0)) // Prevents a ghost ghosting in on a slot and spawning via a resleever with race they're not whitelisted for, getting around normal join restrictions.
+		to_chat(ghost, "<span class='warning'>You are not whitelisted to spawn as this species!</span>")
+		return
+	
+	var/datum/species/chosen_species
+	if(ghost.client.prefs.species) // In case we somehow don't have a species set here.
+		chosen_species = GLOB.all_species[ghost_client.prefs.species]
+		
+	if(chosen_species.flags && NO_SCAN) // Sanity. Prevents species like Xenochimera, Proteans, etc from rejoining the round via resleeve, as they should have their own methods of doing so already, as agreed to when you whitelist as them.
+		to_chat(ghost, "<span class='warning'>This species cannot be resleeved!</span>")
+		return
+	
 	//Name matching is ugly but mind doesn't persist to look at.
 	var/charjob
 	var/datum/data/record/record_found
--- a/code/modules/resleeving/autoresleever.dm
+++ b/code/modules/resleeving/autoresleever.dm
@@ -80,6 +80,7 @@
 		to_chat(ghost, "<span class='warning'>You are not whitelisted to spawn as this species!</span>")
 		return
 	
+	/* // Comments out NO_SCAN restriction, as per headmin/maintainer request.
 	var/datum/species/chosen_species
 	if(ghost.client.prefs.species) // In case we somehow don't have a species set here.
 		chosen_species = GLOB.all_species[ghost_client.prefs.species]
@@ -87,6 +88,7 @@
 	if(chosen_species.flags && NO_SCAN) // Sanity. Prevents species like Xenochimera, Proteans, etc from rejoining the round via resleeve, as they should have their own methods of doing so already, as agreed to when you whitelist as them.
 		to_chat(ghost, "<span class='warning'>This species cannot be resleeved!</span>")
 		return
+	*/
 	
 	//Name matching is ugly but mind doesn't persist to look at.
 	var/charjob


Heroman3003:
Autoresleevers very intentionally can resleeve whitelisted species, that is not an oversight. As for 'ghosting' I think the issue is generally the ability to use it without having joined round and died first, and that should be what's fixed and not the whitelist thing (although that form of sanity seems good).

Rykka-Stormheart:
> Autoresleevers very intentionally can resleeve whitelisted species, that is not an oversight. As for 'ghosting' I think the issue is generally the ability to use it without having joined round and died first, and that should be what's fixed and not the whitelist thing (although that form of sanity seems good).

I think you're misunderstanding the PR.
Whitelisting sleeving is ONLY not allowed if the PLAYER is not whitelisted for the species.
Example A:
Bob logs in, chooses his xenochimera slot. He's not made an app nor been approved, but he's able to ghost, and the autoresleever has latespawn on. He clicks to spawn, and is able to.

This check prevents that by giving feedback that they're not whitelisted and not allowing them to bypass the spawn restriction.

As for the NO_SCAN respawn:
This is a flag deliberately applied to certain races like Xenochimera that cannot be resleeved normally. Why do these ~special~ resleevers get around that restriction?

Skysteak:
Do note that as said on Chomp itself, the whole idea of manual notification is to remove 'false positives' wherein the the target's RP potentially can set it off -- hopefully preventing SSD/useless bodies being made. In addition history has shown that people tend to forget to opt out of something (Suit Sensors whilst being consumed), so having it be an active process to opt in is better imo.

Heroman3003:
> As for the NO_SCAN respawn:
> This is a flag deliberately applied to certain races like Xenochimera that cannot be resleeved normally. Why do these special resleevers get around that restriction?

Because this is also, in part, to allow species like black-eyed shadekin or protean (both of which are not normally resleevable) to return into round via more IC means than "offsite respawn and join" after digestion, and it kinda allows xenochimera to participate in digestion scenes too without being completely removed from round. The NO_SCAN mechanic exists primarily to prevent the bodies of that species from being printed out (as its undesirable to have random people be able to get into bodies of whitelisted species), as well as in part due to some of them having special revive methods for non-vore deaths. For vore deaths however...

Rykka-Stormheart:
> Do note that as said on Chomp itself, the whole idea of manual notification is to remove 'false positives' wherein the the target's RP potentially can set it off -- hopefully preventing SSD/useless bodies being made. In addition history has shown that people tend to forget to opt out of something (Suit Sensors whilst being consumed), so having it be an active process to opt in is better imo.

That hasn't been reverted or changed in any way (on this PR/here), so I'm not sure why this is commented here. I'll respond to the Chompstation one in a minute.

Skysteak:
> > 
> That hasn't been reverted or changed in any way (on this PR/here), so I'm not sure why this is commented here. I'll respond to the Chompstation one in a minute.

Ah you're correct, forgive me!

Rykka-Stormheart:
> > As for the NO_SCAN respawn:
> > This is a flag deliberately applied to certain races like Xenochimera that cannot be resleeved normally. Why do these special resleevers get around that restriction?
> 
> Because this is also, in part, to allow species like black-eyed shadekin or protean (both of which are not normally resleevable) to return into round via more IC means than "offsite respawn and join" after digestion, and it kinda allows xenochimera to participate in digestion scenes too without being completely removed from round. The NO_SCAN mechanic exists primarily to prevent the bodies of that species from being printed out (as its undesirable to have random people be able to get into bodies of whitelisted species), as well as in part due to some of them having special revive methods for non-vore deaths. For vore deaths however...

The preventing people from getting into whitelisted species is done NOT by NO_SCAN but by blanket disallowing all whitelisted species to be printed via a cloning pod + resleeved. I know 'cause I've been digging through the code.
This will be changed so that whitelisted species that do NOT have the NO_SCAN flag can be printed and sleeved into if the player is whitelisted for them, but that is coming in a separate PR as I wanted to keep scope focused on one thing.

I will also posit, again, that as per whitelist rules, and I'm noting specifically for Xenochimera here as they're the ones I know by heart, that it is your responsibility as per the whitelist, to my knowledge, to ensure that you do not die or use offsite respawn to get around the fact that you cannot be resleeved, as this invalidates the entire point of the species' restrictions.

Very-Soft (changes requested):
Checking to see if the person is actually whitelisted is good, I didn't think of that. 

However, for preventing certain species from being resleeved: I deliberately considered and did not add restrictions to species who cannot be resleeved, because the primary reason most species that can't be resleeved can't, is because they are whitelist species, and allowing them to be resleeved would allow anyone with access to a resleever to print out and sleeve themselves or others into those species.

The auto resleever cannot be used in this way, so it's not something we really have to worry about. 

Additionally, the majority of those species also do not have any way to return to the round apart from respawning, which we do allow. So this is hardly a stretch, since it doesn't open up much room for abuse.

Rykka-Stormheart:
Reopening, failed active edge test.

PR #12555, opened 2022-03-27
Adds siphon vent to Tether's autoresleever room.

foopwo:
Just like in the cargo office, the delivery chute pumps air to the room. This triggers the air alarm if it happens enough times, so I added a a siphon vent to maintain regular pressure.
The vent is placed under the clothes vendor (Where the scrubber was), and the scrubber moved in front of the resleever.
sleeber
Pipe icons are wonky in the editor, but I assure everything's connected and functional.

### DIFF
--- a/maps/tether/tether-01-surface1.dmm
+++ b/maps/tether/tether-01-surface1.dmm
@@ -30484,7 +30484,12 @@
 /obj/effect/floor_decal/corner/paleblue/border{
 	dir = 9
 	},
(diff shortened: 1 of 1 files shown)


Very-Soft:
did you test and see if the scrubber actually connects and works with the supply pipe right there? If not it would work fine underneath the auto resleever

foopwo:
All three units were tested, functional, connected to their relevant pipenets.

PR #12558, opened 2022-03-27
Xenochimera QOL PR

Rykka-Stormheart:
Massive QOL changes 'chimera have been needing for a while.
- Adds Regenerating + Hatch thrown alerts, similar to stunned/weakened/buckled, and clicking will show the remaining time to revive, or allow you to hatch.
- Defibrillators will now apply halloss to Xenochimera and set them feral on revive.
- ~Defibrillators will now apply halloss (pain) to everyone on revive.~
- Medscanners will now display if a Xenochimera is regenerating or needs more protein, and if a 'chimera is dead, will warn that they require sedation.
- Reconstitution cooldown reduced from 1 hour to 10 minutes.
- Regeneration will no longer handle ferality in the middle of it, preventing edge-case *twitches while clearly regenerating.
- Regeneration will now have an audible + visible cue that plays on a 2% chance.
- ~Xenochimera have gained the Lick Wounds Verb.~
- Lick Wounds has been modified to allow treating self-injuries, and injuries of the mob next to you. Given how limited the use is and how often bandages + chems are just flat-out better, this shouldn't be a huge balance issue, IMO.

Additionally, as per 2nd commit:
Xenochimera will now reduce ferality at a tickrate of 1 per Life tick while inside a belly or in darkness, same as the hunger reduction if they're full.
They will not experience hallucinations and will get safety messages while inside a belly, as well as the status indicator confirming they're safe even in a lit environment.

```
In total:
If sated, feral will reduce by 1 per Life() tick.
If in darkness _or_ a belly, feral will reduce by 1 per Life() tick.
A feral of 220 will take 110 seconds to clear if satiated, not in a belly, and in light.
Alternatively, it will take 55 seconds to clear if satiated **and** either in a belly _or_ in the darkness.
```

Upstream port of
https://github.com/CHOMPStation2/CHOMPStation2/pull/3906 and https://github.com/CHOMPStation2/CHOMPStation2/pull/3823 respectively.


lm40:
Thumbs down on defibs doing halloss to everyone

Novacat (changes requested):
There was a small admin discussion on this, and the two takeaways:

- Defib doing halloss to everyone should probably be split off into its own PR.
- Xenochimera probably should not have the Lick Wounds verb.

Rykka-Stormheart:
> There was a small admin discussion on this, and the two takeaways:
> 
> * Defib doing halloss to everyone should probably be split off into its own PR.
> * Xenochimera probably should not have the Lick Wounds verb.

Do you mind elaborating on the latter, as to why they shouldn't?
This is the only IC method of them treating themselves if no-one is around, as iirc 'chimera aren't capable of bandaging themselves (ICly, at least, mechanically they can.)

Novacat:
Lick wounds cannot be used on yourself (at least at the time of its implementation), and generally Scree thinks it is a little silly for Xenochimera to be using lick wounds on non-Xenochimera for... obvious reasons.

Rykka-Stormheart:
> Lick wounds cannot be used on yourself (at least at the time of its implementation), and generally Scree thinks it is a little silly for Xenochimera to be using lick wounds on non-Xenochimera for... obvious reasons.

Lick wounds can be used on yourself as of this PR.
If it would make things more comfortable, would you be comfortable with Xenochimera having a version of this verb where they can ONLY lick their own wounds?

If yes to above, should the regular Lick Wounds still have the capability to be used on yourself? As it is it's... basically never used.

Novacat:
Iono! That would be a question for Scree.

Rykka-Stormheart:
> Iono! That would be a question for Scree.
> 
> Personally I do not mind if lick wounds can be used on yourself. It was initially implemented very conservatively due to concerns that never became an issue.

Is Scree around to answer on the Xenochimera-specific lick wounds?

Rykka-Stormheart:
Removed pain on defib, removed lick wounds for Xenochimera, as per maintainer request. All other changes remain the same.

Screemonster:
I don't mind xenochimera being able to lick their own wounds. They're meant to be crazy-survivalist feral apexcritters. I can't even remember what lick wounds even does - doesn't it prevent infections which they're immune to anyway?

A xenochimera slavering all over a regular crewmember's open wounds would logically do the exact opposite of preventing infection.

Rykka-Stormheart:
> I don't mind xenochimera being able to lick their own wounds. They're meant to be crazy-survivalist feral apexcritters. I can't even remember what lick wounds even does - doesn't it prevent infections which they're immune to anyway?
> 
> A xenochimera slavering all over a regular crewmember's open wounds would logically do the exact opposite of preventing infection.

Okay but what if Xenochimera-special lick wounds, and if they use it on someone else it immediately gives them a max-strength infection /s

~Obviously we code roanoke syndrome and such here right /s~

All of the above is sarcasm. I'll take a look at adding a separate, 'chimera specific lick-wounds later as per request.
That can go in a separate PR, though.

PR #12561, opened 2022-03-28
SD - V2 aerostat xenobio cell

PastelPrinceDan:
someone in the discord brought up the idea of having a single xenobio cell on the outpost, for more dangerous xenobio activities that shouldn't happen on the ship (mob-spawning, releasing mobs in stasis cages, reviving mobs with a denecrotizer, etc)

so, this little unoccupied room on the aerostat was turned into a tiny xenobio lab

contains things such as:
-  very basic necessities (this is not supposed to replace general xenobio on the ship AT ALL) + a cell
- mass driver and outer blast door to eject mob (if you're lucky enough and they step on the MD)
- inner blast doors in case of oopsies
- interior field gens so mobs cannot pass through 

image


Very-Soft (commented):
You have an active edge at 90,71, looks like there might be a space turf in there somewhere?

PastelPrinceDan:
> You have an active edge at 90,71, looks like there might be a space turf in there somewhere?

fixed now!

PR #12569, opened 2022-03-31
Updates On-Death message to direct players to Autoresleever or wait.

Rykka-Stormheart:
Upstream port of https://github.com/CHOMPStation2/CHOMPStation2/pull/3963


Wickedtemp:
helps prevent the 
"Someone's implant notif went off I'lll handle it."
Printing the body, 
"Oh they just used the autosleever... well, the body is already printing. Guess I'll bag and tag it."

Very-Soft:
I like updating this, but I do not personally like the tone of the message, as it makes using the autoresleever for vore deaths sound mandatory, which it is not. I would prefer if it was more like 'If you died you can use the autoresleever, or wait for medical to resleeve you, you can use the ghost verbs for either thing'

PR #12571, opened 2022-03-31
Trait Tweak - Positives capped at 4, Negatives Uncapped

Rykka-Stormheart:
As above.
Traits now have an unlimited amount for negatives, given they are intended to harm you.
Positives have a cap of 4 total.
Prior to this, one could do 4 positives and 1 extremely expensive negative, but no other 'fun' negatives.

Now, one can still do 4 strong positives, but as many negatives as they like.
IDEALLY this will allow for custom setups with a lot of small-cost negatives for higher cost positives.

Upstream port of https://github.com/CHOMPStation2/CHOMPStation2/pull/3958

### DIFF
--- a/code/__defines/misc_vr.dm
+++ b/code/__defines/misc_vr.dm
@@ -20,7 +20,7 @@
 
 //For custom species
 #define STARTING_SPECIES_POINTS 2
-#define MAX_SPECIES_TRAITS 5
+#define MAX_SPECIES_TRAITS 4	// Cap positive traits at 4, given negatives are unlimited. Prior number was 5.
 
 // Xenochimera thing mostly
 #define REVIVING_NOW		-1
--- a/code/modules/client/preference_setup/vore/07_traits.dm
+++ b/code/modules/client/preference_setup/vore/07_traits.dm
@@ -165,28 +165,29 @@
 	if(pref.species == SPECIES_CUSTOM)
 		var/points_left = pref.starting_trait_points
 
-		for(var/T in pref.pos_traits + pref.neg_traits)
+		for(var/T in pref.pos_traits + pref.neg_traits) // Only Positive traits cost slots now.
 			points_left -= traits_costs[T]
+		for(var/T in pref.pos_traits)
 			traits_left--
 		. += "<b>Traits Left:</b> [traits_left]<br>"
 		. += "<b>Points Left:</b> [points_left]<br>"
 		if(points_left < 0 || traits_left < 0 || !pref.custom_species)
 			. += "<span style='color:red;'><b>^ Fix things! ^</b></span><br>"
 
-		. += "<a href='?src=\ref[src];add_trait=[POSITIVE_MODE]'>Positive Trait +</a><br>"
+		. += "<a href='?src=\ref[src];add_trait=[POSITIVE_MODE]'>Positive Trait(s) (Limited) +</a><br>" // More obvious/clear to players.
 		. += "<ul>"
 		for(var/T in pref.pos_traits)
 			var/datum/trait/trait = positive_traits[T]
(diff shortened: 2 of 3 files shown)


Very-Soft (changes requested):
I'm not a fan of this because while, I'm fine with letting people take as many negatives a they want, the point balancing is designed around you not having effectively unlimited points.

I would be fine with this if the negative traits still counted toward your total traits, but you could keep taking negative traits even if you exceeded that number, and then wouldn't let you spawn with positive traits until the remaining traits was 0 or more.

I'm fine with the concept of someone taking every negative trait in the game, but I don't want that also to mean they get 4 super strong traits just because they took every negative trait they don't care about,

Rykka-Stormheart:
> I'm not a fan of this because while, I'm fine with letting people take as many negatives a they want, the point balancing is designed around you not having effectively unlimited points.
> 
> I would be fine with this if the negative traits still counted toward your total traits, but you could keep taking negative traits even if you exceeded that number, and then wouldn't let you spawn with positive traits until the remaining traits was 0 or more.
> 
> I'm fine with the concept of someone taking every negative trait in the game, but I don't want that also to mean they get 4 super strong traits just because they took every negative trait they don't care about,

I don't personally see it as a point of concern, given if people want to pile on negatives, they pay in spades for it. Negatives on VORE Code are FAR heavier than the positives, and positives give very little.
If people want to cripple themselves with negatives to get 4 good positives, more power to them.
An example:


40% reduced incoming Burn damage
40% reduced incoming Brute damage
40% reduced incoming pain
175 HP (350 total before death)
Players cannot trade places with you, nor can mobs.

In return, negatives:
Hallucinate and struggle as long as there are not carbon mobs on your screen.
30% effective heals. This character took 170 toxin, for instance, and required overdose on Cartha and Dylo and still took over 90u to clear it.
-1.5 Slowdown, so they walk much slower, on a codebase that rewards dodging hits.
Bleeds 3x as much as normal.
Traumatic Shock lasts 2x as long (The CRIT warning + shock slowdown).
They sleep for 2x as long.

This is all CHOMP Traits, mind you, but the point is:
In return for being able to frontload damage, the crippling negatives I take on result in being extremely slow to heal, and basically die once the short-burst of combat is over.

On this codebase, the positives are FAR weaker, and the negatives, bar the Reduced Bicompatability + Deep Sleeper, are just as strong as downstream.
20% brute/burn resist here whereas the weaknesses go as high as 50% incoming damage.

My point is, people cannot make absurdly broken builds by way of High Endurance + Resists being blocked from being paired on this codebase, unlike downstream.

Screemonster:
Eh, there's a reason why most tabletop systems (like the old world of darkness games) capped the possible point gain from negative traits, because it's insanely easy to do some ludicrous minmaxing.
Other games, like the old tri-stat system, deliberately made negative traits give _very few_ points for the impact they had, so that they'd become something that people would _maybe_ squeeze out one or two if their character was _just_ over the points allowance, but building your character around having an absolute ton of flaws would leave you basically crippled for very little benefit.

I'm fine with people taking extra neg traits as effectively zero-point versions once they exceed their normal allowance.
One of the key things is that people shouldn't be able to use traits to make a species that's better than a stock species.

Rykka-Stormheart:
> Eh, there's a reason why most tabletop systems (like the old world of darkness games) capped the possible point gain from negative traits, because it's insanely easy to do some ludicrous minmaxing. Other games, like the old tri-stat system, deliberately made negative traits give _very few_ points for the impact they had, so that they'd become something that people would _maybe_ squeeze out one or two if their character was _just_ over the points allowance, but building your character around having an absolute ton of flaws would leave you basically crippled for very little benefit.
> 
> I'm fine with people taking extra neg traits as effectively zero-point versions once they exceed their normal allowance. One of the key things is that people shouldn't be able to use traits to make a species that's better than a stock species.

I s'pose. I'll take a look at that later, though, will close this one for now. Thank you for the feedback! 7o

PR #12575, opened 2022-04-01
The Grand /TG Port - Part 1

GhostActual:
Now in bite-sized portions, this PR set aims to port over a slew of features, sprites, and mechanics from various different /tg base codes and make them work with the Bay code. This is an ongoing effort and already things are changing as I make this more manageable for the moderators and admins. Nothing will be removed from the original PR, it will just be more organized.

PLEASE MERGE THESE IN ORDER OR IT MIGHT BREAK

Part one includes:
-Minor code changes/organization
-Spooky Ghosts now drop Ectoplasm so the cook can once again make ghost burgers
-The Advanced Mop has been ported (sprite modified) to the code. It works twice as fast as the original mop and requires research to obtain
-An alternative Space Skeleton sprite was added - directly ported from TG and modified to have a vore sprite
-Space Skeleton's catalog description was changed to be more lore accurate (can be touched upon by the lore-mins)


Very-Soft (commented):
Conflicts, I will focus on this next so shouldn't be more conflict after now

GhostActual:
> Conflicts, I will focus on this next so shouldn't be more conflict after now

Fixed conflicts

PR #12600, opened 2022-04-02
Tiny pipe fix & gas leak event fix

### LINKED ISSUE #12599: Virology and command are cut off from atmospherics
Upwards pipes in Engineering SMES room on NSB Rascal's Pass do not connect to the atmospherics system.

Virology and command to receive air as well as filter our bad stuff.  

#### What actually happened
Pipes weren't connected, so the atmos issue was resolved but the pipes ended up over pressure and supply wasn't filling.

#### Steps to reproduce
Go to SMES room in engineering. Go to the powernet module above the grid checker and turn on your T-Ray. Wall pipes are not connected at all :(

To fix this, add the pipes show in the image below, including replacing the straight pipes with manifolds at that intersection.

#### Code Revision
- c15e97c853511fb40119e845dd1301d5ebffe931
 

image

ItsSelis:
Fixes #12599

Fixes the issue of the gas leak event appearing on the outside (or in tunnels) of the groundbase map, spreading the gas everywhere on the outside.

The "Scrubber to Waste" pump was moved, due to it filling up the mobile scrubbers otherwise.

grafik
grafik

### DIFF
--- a/maps/groundbase/groundbase_areas.dm
+++ b/maps/groundbase/groundbase_areas.dm
@@ -369,6 +369,7 @@
 	excluded |= /area/groundbase/level1/westspur
 	excluded |= /area/maintenance/groundbase/level1/netunnel
 	excluded |= /area/maintenance/groundbase/level1/nwtunnel
+	excluded |= /area/maintenance/groundbase/level1/stunnel
 	excluded |= /area/maintenance/groundbase/level1/setunnel
 	excluded |= /area/maintenance/groundbase/level1/swtunnel
 	excluded |= /area/groundbase/level2/ne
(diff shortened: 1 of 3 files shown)


maps/groundbase/groundbase_areas.dm
```diff
@@ -358,3 +358,25 @@
 	icon_state = "orablacir"
 /area/groundbase/mining/explored
 	icon_state = "blublacir"
+
+// Exclude some more areas from the atmos leak event since its outside.
+/datum/event/atmos_leak/setup()
```
Very-Soft, line 363: you're probably going to want to do the level2 areas also, since it looks like you missed them

ItsSelis: Ah, I was certain that I did add them. I can correct it in a bit.

PR #12671, opened 2022-04-06
Makes mobs not spawn with id unless defined so

Verkister:
Makes mob ids only spawn on mobs that should come with them rather than jamming one into every single mouse and cockroach and so on.


BonniePandora:
Oh right, that was something that was requested of me by Soft when I was adding that in for those catslugs. They'd wanted it added so all mobs got it so admins could mess around with the vars for giving mobs access on the fly & whatnot.

Verkister:
There's an even easier way for that that doesn't require adding blank ids to every mob.
Simplemobs already have a var called "myid" that can be edited into an atom reference to any spawned ID card without even needing to dig the card up from within the mob's vars for varediting, and it doesn't even require the physical presence of the ID within the mob itself.

PR #12679, opened 2022-04-07
[STALE] Map Rotation

ItsSelis:
Currently pending integration of the server itself.

- [X] Ability to add maps to the rotation list (config/map_rotation.txt)
- [X] Compilation of chosen map (data/map.txt)
- [x] (Debug) Ability to instantly trigger map rotation
- [x] Ability to override map in current rotation (Admin Verb)
- [x] Keeping track of how many rounds passed after last rotation (required for the configurations below)
- [x] Configurable map swapping modes
  - [x] Automatic (x number of rounds)
  - [x] Vote (x number of rounds)
  - [x] Ability to also schedule it on a specific weekly day (e.g.: Sunday Evening) instead of x rounds
- [x] Triggering the Map rotation script, if one of the configured options above were met
- [x] Triggering the Map rotation script, after manual vote was started (and had a result)
- [x] Ability to use custom map rotation handling (custom scripts/software)

Currently, the script will look inside of `data/map.txt`. The server itself can set the next map in rotation in here, either by admin override, vote or automatic rotation.
It defaults to the very first entry of `config/map_rotation.txt`, in the case of this file missing.

The currently available maps are set in `config/map_rotation.txt`. This would allow, to completely disable a specific map from the rotation, for example for event reasons.
Invalid maps (at least ones that do not exist) will be ignored by the script.

The script will do the rest once called.
**Note:** The first time running the script, might prompt to accept that it will run in trusted mode. It is very recommended to also check "Remember this decision" in order for this to work.

It is also possible to enable custom map rotation handling, disabling the script entirely. In that case, it will set the number 0 to 1 (From FALSE to TRUE) inside of `config/map_rotation.txt` **on line 3**.


ItsSelis:
Added Verbs:
- Debug -> Trigger Map Rotation
- Server -> Override Map Rotation

Heroman3003:
I think there should also be option to have it happen on schedule irrelevant of round amount or voting. Our round timings are uneven and don't amount to perfect day/week, so the round-based schedule, if used, will slowly crawl and stop being at end of sunday and become early monday, then late monday, etc.

ItsSelis:
> I think there should also be option to have it happen on schedule irrelevant of round amount or voting. Our round timings are uneven and don't amount to perfect day/week, so the round-based schedule, if used, will slowly crawl and stop being at end of sunday and become early monday, then late monday, etc.

Yeah, it would really shift. Adding it would be a pretty good idea.

ItsSelis:
Note:
When I tested the script with the Ubuntu VM I set up, I had to use following command to favor Python 3.x for the 'python' command (since python is already 3.x on Windows and the script syntax is using that version at the moment)

`update-alternatives --install /usr/bin/python python /usr/bin/python3 10`

GhostActual:
Amazing work

Cameron653:
This needs conflicts resolved and IIRC someone with access to the serverbox to help.

ItsSelis:
> This needs conflicts resolved and IIRC someone with access to the serverbox to help.

Yeah, resolving the conflicts is paused until VGS integration is done (since it will just keep going and going with merged from master commits.

ItsSelis:
This may be re-opened once no longer stale or a better system implemented.

PR #12685, opened 2022-04-08
What's that material??? & Techdisk Bluescreen fix

FartMaster69420:
Fixes the issues people were having with Exosuit and Prosthetic Fabricators showing incorrect icons for some materials. (Only icons changed in this case are platinum and graphite.)

Further, the icons are now tooltips, you can hover your mouse over to see what material it is. This guarantees some of the weird grayscale icons can be deciphered.

What is still not changed (TODO if I feel not lazy in the future)
-Changing the icon color for grayscale icons
-Not all materials have icons set, though all materials that are used in recipes should be.


FartMaster69420:
As a bonus just pushed a fix to tech disks causing a blue screen when one tries to upload a tech to one in the RnD Console screen.

Putting it here because TGUI bundles are annoying, and I don't wanna wait to make another PR after this gets merged.

PR #12686, opened 2022-04-08
Firearms 2

GhostActual:
Ports over many sprits from the original Armory PR. Also moves some things around.


Very-Soft (changes requested):
Conflicts

GhostActual:
> Conflicts

fixed

PR #12687, opened 2022-04-08
Nerfs gravitons and morphium, adds lots of boards into autolathe

Heroman3003:
Gravitons now require first level Anomaly and 500 units (1/4th of a sheet) of diamonds to construct.
Morphium no longer gives Anomaly levels.

Adds following boards to autolathe: all three arcasdes, jukebox, mech recharger, borg recharger, battery rack, grid checker, breaker box, atmos heater and cooler, ARF.

### DIFF
--- a/code/datums/autolathe/engineering_vr.dm
+++ b/code/datums/autolathe/engineering_vr.dm
@@ -25,3 +25,51 @@
 /datum/category_item/autolathe/engineering/candymachine
 	name = "candy machine electronics"
 	path =/obj/item/weapon/circuitboard/candymachine
+
+/datum/category_item/autolathe/engineering/battle_arcade
+	name = "battle arcade machine electronics"
+	path =/obj/item/weapon/circuitboard/arcade/battle
+
+/datum/category_item/autolathe/engineering/orion_trail
+	name = "orion trail aracade machine electronics"
+	path =/obj/item/weapon/circuitboard/arcade/orion_trail
+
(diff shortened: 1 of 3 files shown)


TheObserver-sys:
As I said on the discord, I will say it here:
Gravitons probably don't need to be pushed behind anomalous, for while they give you good vision, that is.. all they do. Most jobs, not just exploration require it for better operation, including the selfsame job you say you're trying to make more relevant. (It also just encourages bug abuse that is difficult to fix instead of  encouraging people to play Xenoarch, a job that is in desparate need of content and purpose outside rush down any "anomalous material/long exposure particle" to check that off the list.) A better way to nerf gravitons, and visors in general would be to take the eris approach to them, make them require power to use. A small cell/device cell to be precise. That's what they do in the NIF, it makes sense that they might require it outside the NIF too.

For morphium, it makes absolutely no sense to strip away its value as anomalous. For one, it's rarely made often enough for it to matter, and for two, the effort to create it justifies its ability to unlock T4's and their assorted technology. A better way to nerf it, as I said, is to remove the persistence from it on the fridge, along with the other materials like supermatter and whatnot, since those too, if I am thinking right, probably can be fridged as well, with far more devastating results. However, if you insist upon removing anomalous from it as the most short sided fix for it, then I must argue that it acquires precursor in exchange, as all useful precursor items require it to create them -- to the point there is a recipe to turn alien junk INTO morphium with the particle smasher.

PR #12698, opened 2022-04-10
Xenobio Upgrades

FartMaster69420:
Makes several edits toward allowing for xenobio tools that will make life a tad easier.

1) A xenobio bag is added to all Science Level 3 Bio lockers, this item can hold monkey cubes, extracts, and potions

2) Dart guns can now hit xenobio extracts and transfer reagents to them. (For fun, the same can be done to food.)

3) A bluespace monkey delivery gun can now be built. This uses the bluespace harpoon sprite and can be loaded with a single monkey cube. When fired, it deploys a monkey at the targeted turf.

4) A portable slime processor can now be built. This can process a single slime or monkey at a time, grinding them up where they lay.  This uses the chainsaw sprite as a placeholder.

Both the Monkey Gun and Portable Processor use morphium in their construction, and require "endgame" tech levels. That said, consider the current tech requirements and mat costs placeholder. I haven't done much balancing on them, yet.

Feedback is very much welcome!


Heroman3003:
I feel like slime processor being portable is kind of lame tbh. Other stuff seems cool and fine, although the dart gun thing affecting all injectables may result in some hmmmmm.

FartMaster69420:
The portable processor has a timer for each extract it removes from a slime and can only process one slime at a time, so other than being portable it's kinda worse than a standard processor.

As for darts in food, I don't see it being any worse of an issue than syringes already are. Especially since it's not exactly subtle to shoot a dart at a food item. (The dart doesn't vanish, it just sits there in front of it.)

PR #12700, opened 2022-04-10
TGUI Additions

ItsSelis:
**TGUI Fallback Code**
TGUI will now send an "__fallback" act back to the tgui_act handler of the window that errored, giving the possibility to add legacy menu's or handle it in some way, not just leaving one stuck with that TGUI "Bluescreen".

Here is a simple demonstration:
grafik

**Admin UI Conversion to TGUI**
This work in progress will replace the old ui of the player notes with a new one made with TGUI.
The only thing missing currently is the ability to apply filters and the page system.

Maybe other interfaces can be updated to TGUI as well.

Current state:
grafik


ItsSelis:
Bunde Compliation Date/Time: 22.05.2022/21:23

PR #12718, opened 2022-04-12
Ports Span Macros from TG Station

FartMaster69420:
This is a direct port of https://github.com/tgstation/tgstation/pull/59685

This adds the ability to use a macro to simplify spans in text outputs.

Example:

This:

user.visible_message("<span class='notice'>[user] Does [thing]</span>", "<span class='notice'>You do [thing].</span>")

Becomes This:

user.visible_message(span_notice("[user] Does [thing]."), span_notice("You do [thing]."))

This is mostly a QOL change for development, and doesn't change anything in player experience. It should also make porting some TG code stuff slightly less tedious.

Note, the old method still works and no current uses have been changed in code. This is mostly just for edits going forward.


Cyantime:
These already exist here, in two different ways
https://github.com/VOREStation/VOREStation/blob/master/code/__defines/misc.dm#L400
for the direct "SPAN_X" macros, and
https://github.com/VOREStation/VOREStation/blob/master/code/_macros.dm#L1
for the generic version.
Duplicating these a *third* time just to copy TG seems unnecessary.

FartMaster69420:
This is certainly not a necessary change, but it brings a wider array of spans over into macros.

PR #12725, opened 2022-04-13
Update station_special_abilities_vr.dm

liache:
This removes the nutrition cap from winged flight! Now LORGE full flyers can dropnom and get lorger still. Imagine the size of the oncoming shadow.

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/station_special_abilities_vr.dm
+++ b/code/modules/mob/living/carbon/human/species/station/station_special_abilities_vr.dm
@@ -813,7 +813,7 @@
 
 /mob/living/proc/flying_toggle()
 	set name = "Toggle Flight"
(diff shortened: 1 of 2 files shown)


foopwo:
This only changes half of it.
https://github.com/VOREStation/VOREStation/blob/cdb1b8f43d430ebaf6f607b43f6a38f847172f27/code/modules/multiz/movement.dm#L246 will need to be changed as well.
There may be other flight-related nutrition checks too.

Very-Soft (changes requested):
see above comment

liache:
Thank you Foopwoo! My initial tests didn't have my stop flying no matter how much nutrition I had, but I found the stop flying procs you mentioned and edited them out. =) I found a few other little comments referring to the nutrition to get rid of too. I'll be more thorough in the future. Merging the additional cuts into this shortly.

PR #12738, opened 2022-04-15
Stumble Vore

FartMaster69420:
This allows for stumbling into someone from flips, confusion, or blindness to cause a spontaneous vore check between the two players. If prefs align then they will preform a dropnom, much like slipvore.

If prefs do not align for a dropnom, then they smack into each other, and both get knocked down on the same tile and take 0.5 brute damage.

Further, this feature also extends to flight. However, unlike the standard behavior for flight stumbles (walls, tables, ect) this behavior must be opted into. To this nature, a new verb is added 'toggle flight vore' allowing for dropnom checks when bumping into a player. This verb has been added in all places where the standard toggle_flight verb is added, and is to be found in game in the abilities tab.

By default flight vore is disabled, and will not change behavior of flight at all. When it is turned on, then bumping into players will work exactly the same as if you had stumbled into them.


ItsTheKet:
Let me say that I am speaking out of personal interest, rather than official capacity, so don't take my word as law.
Though, as funny as this is, and willing to roll with it. As the resident flipper, I can EASILY see more than the fair share of annoyances.

Thankfully I did look into the code a little and see that it's applying when running. Which fair play for that to happen someone is blind/confused; you already smack into anything if you're running while you're in those states! But at the same time I'm still kinda iffy. Mainly at the fact that if things don't align you smack into them anyways and get knocked over. I'm not sure how i'd feel having to suddenly concern myself with treating other players suddenly being obstacles. Regardless of their intent (or if I'm buckled to an office chair running at mac 5 via fire hydrant thanks to there not being a sanity check for being buckled. Which, again: funny.)

Honestly, at this point we're having so many things tied into spontaneous that we should probably have to divide this all into actual subcategories of itself.

Maybe I want teleporting shenanigans, maybe I don't want to deal with this. I personally don't want to opt in/out on both because they're both tied into the same options. 

At the very least, if prefs didn't align I didn't have to worry about knock down with anyone I run into and the sanity check of buckled people, I might not be AS critical but AS Is, **I do _not_ want to see this being pushed.**

Skysteak:
I admit I am mixed on this because it feels like it something which should be in its own sub category. I appreciate that we're slowly straying towards an endless and incomprehensible array of different toggles ,but at the same time we're getting to the stage wherein Spontaneous Prefs now dictate alot of options -- some potentially unwanted whilst keeping it on for others. I will say that I do appreciate the thoughtfulness taken to add a flight vore toggle, that's good! However I feel that stumbling is easiest condition to accidentally set off versus Drop/Sipnoms, which naturally is a boon but also a potentially irritating negative. For example:

I am playing a character with the Tracuer trait and I keep spontaneous vore - prey on for Dropnoms/Sipnoms. This means I will potentially be flipping alot and that means I potentially may be getting the confused condition alot, which can easily result in me going into someone when I had no intention or desire for a scene. 

Another example is that I am Lightweight and I run into someone -- ensuring I am immediately consumed instead of knocked down -- wherein if that person had wanted to get me for a scene, I would've been claimed anyway during the stun. It's also a bit abrupt too I suppose?

In both examples the only reliable way I could avoid this is removing the traits or turning spontaneous vore prey off entirely, which is not optimal.

Finally I am unsure what benefit is provided by having a stumbling player _knock themselves and someone else_ down? Perhaps there is some initial slapstick value but it's going to create a ton of annoying little situations. Someone stumbles into you whilst drunk, knocking you down. Someone flips over a table with Traceur gets confused and crashes into you ,knocking you down.  A mercenary simplemob throws a flashbang, causing multiple people to potentially bowl each other over in front of an armed simplemob, knocking them down at best. As Ket mentions it turns people into obstacles and whilst in the most mild of forms, punishes the non stunned player because...? If people wanted to experience that easily then they'd take the 'Lightweight' trait.

FartMaster69420:
The only reason I didn't include a specific pref for it is due to the aforementioned overload of pref options in the panel.

I could certainly limit the knockdown to only happen to those with light weight or remove it entirely.

FartMaster69420:
Preferences have now been added for drop vore, slip vore and stumble vore. Allowing for toggle of each. (They still require spontaneous pred or prey to be enabled) Televore wasn't touched, since that's harder to do accidentally.

Further, smacking into someone will not knock both players over, it will only knock over and stun the confused/blind/flying player.

The stumble interactions are different in two scenarios, if the player has the light weight trait, or if they are past the maximum examine weight level (over 474 weight, gives the biggest gut examine message.) If they are light weight, then both players get 0.5 brute damage and are knocked over. If they are at a high weight level then the stumbler bounces off of them and is tossed back up to two tiles away.

PR #12750, opened 2022-04-16
Pappershredder/Space Heater Revive

GhostActual:
What it says on the tin. No more bland grey boxes.

space_freezer
space_heater


Very-Soft (commented):
not sure if icondiff bot is just throwing a fit, or if you forgot to include the sprite changes, but it looks like the heater sprites just got deleted, rather than replaced.

GhostActual:
> not sure if icondiff bot is just throwing a fit, or if you forgot to include the sprite changes, but it looks like the heater sprites just got deleted, rather than replaced.

The ones circled are the ones that were replaced. The one that was deleted was removed because it was a duplicate.
Screenshot 2022-04-17 112115

PR #12754, opened 2022-04-16
[MIRROR] Historic weapon expansion

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/8502
--------------------
Adds a bunch of sword types, ported from Aurora, as well as some extra polearms. (pike, halberd, etc) Currently only admin spawnable, but could be used for POIs, randomspawners, if one were set up for swords.

Also adds a bow and arrows, ported from Vorestation. (who got the sprites from TG) Both are craftable via the crafting menu. Unlike metal rods for the crossbow, arrows cannot be stacked, instead, you can craft a quiver (also in the crafting menu) to store them in! Sprites from TG.
There is also an energy/hardlight bow, that is also only admin spawnable currently.

image


GhostActual:
Added then Reverted so I'm not going to bother fixing this

PR #12761, opened 2022-04-17
EnerGen Brand Batteries

GhostActual:
New yet somehow Old cell battery sprites.

Batteries

(Blue and White battery sprites belong to VerySoft)


Very-Soft (changes requested):
would appreciate it if the ? thingy got some kind of battery indication on it rather than just being a question mark. Veing able to tell the random spawners apart by peeking them is kind of nice UWU

GhostActual:
> would appreciate it if the ? thingy got some kind of battery indication on it rather than just being a question mark. Veing able to tell the random spawners apart by peeking them is kind of nice UWU

Fixed

PR #12764, opened 2022-04-17
Adds a disposal bin that washes clothes and other oddities

Very-Soft:
Adds a new disposal bin 
image

it is wall mounted and has a distinctive blue color, these are exclusively connected to resleeving, auto-resleeving, or to pipes with a destination tagger for resleeving.

In short, this means there are now special disposal bins around which you can anonymously deposit soiled equipment into and have it delivered to medical all cleaned without having to worry about anyone in the lobby screaming at you.

I have also made contamination default to disabled in the vore panel, and slightly changed the save key for that variable, so all bellies new or old will have the option disabled. 

If you want to use the feature, that's fine, just turn it back on and save your vore panel and you won't have to worry about it again.

This is intended to make contamination more of a conscious choice on the part of the individual, rather than having it be the default, and has been made in light of a string of complaints about the feature being abused.


code/modules/vore/eating/vorepanel_vr.dm
```diff
@@ -701,7 +701,7 @@
 			host.vore_selected.item_digest_mode = new_mode
 			host.vore_selected.items_preserved.Cut() //Re-evaltuate all items in belly on belly-mode change
 			. = TRUE
-		if("b_contaminates")
+		if("b_contaminate")
```
Verkister, line 704: Spotted a little whoopsie right here.

Verkister:
Honestly what's the point in changing the save key there? Like legit unable to come up with any other motive than to purposefully fuck over people's already existing setups just for the sake of fucking them over?

Heroman3003:
Secretely messing with everyone's savefile settings is no bueno

Verkister:
If you wanna sabotage your players' carefully crafter save files willy nilly with hasty underhanded speed selfmerges on a whim just because you don't feel like addressing the culprit themself, don't be surprised with playerbase hemorrhaging towards other better managed servers.

Very-Soft:
> Honestly what's the point in changing the save key there? Like legit unable to come up with any other motive than to purposefully fuck over people's already existing setups just for the sake of fucking them over?
> 
> Like come on are you really really intending to target the entire playerbase with this kind of punishment just to retaliate on some dork's ingame abuse instead of acting like an admin and slapping the perp without needing to pretend you're just a misguided dev pretending to be an admin with indiscriminate gungho code retaliation.
> 
> You know the abuser can just single click turn their singular nastifier vorgan back on and keep on making a mess, and that's literally all the "gain" that the save key change accomplishes at the cost of literally everyone else needing to fix their entire vore setups belly by belly, character by character, just to get their stuff back the way they intended. Is that really worth it?

The alternatives were
>taking administrative action on everyone who left goopy clothes anywhere anyone might find them, which is basically everywhere due to proposed area by area consent policy changes

I was strongly against this idea, because 
1. the feature is on by default
2. the VAST MAJORITY of digestion scenes where the prey gets resleeved lead to the contaminated item being given back to the prey. 
3. as a result of 1 and 2, this would be a continuing, ongoing rules trap that people would fall into on accident. 

I figured it would be better to conceive of a mechanical solution to this problem than making potentially infinite future work for ourselves, as that always leads to increasingly soured temperaments for the staff members who have to deal with that, which leads to harsher reactions to it happening.

My initial thought was 'if everyone's complaining about this stuff being left around and no one likes dealing with it, then why not just remove the feature?'

This was talked out, and we came up with 
>it should default to off, but that on its own will not do anything, since every belly ever made has it turned on unless someone went out of their way to turn it off, so the save key should be modified slightly to reflect the change.

>And an easy way to clean and return things to medical without having to barf stuff all over the lobby. 

So I did both of those things. It's not like I just went and did them for no reason, I don't care about goopy clothes one way or the other, I just don't want us making rules that will result in witch hunts because someone didn't know better than to turn something off from its default setting.

This is why this change was made.

Verkister:
Those people who wanted their option disabled were those who already had went out of their way to disable it. Everyone else got their setups diddled noncon and as someone who had a lots of characters with lots of bellies, I fail to see anything short of disasterous coming out of this decision.

Verkister:
Basically instead of taking action against the few, you decided to take action against everyone.

Very-Soft:
I would rather the people who feel like using an option re-enable the thing, than make rules that require us to hunt down people who didn't know that the default option is against the rules to leave around. 

I'm sorry that you might have to click a couple buttons to get your gooey clothing settings back how they were if you ever decide to play here again. But I would rather that, than have a rules trap we have to continuously moderate.

The change is also very small, and I know it's already been undone downstream. If you don't like our changes, you're welcome to not take them.

Verkister:
https://wiki.vore-station.net/Rules#Bystander_Consent_Policy

PR #12776, opened 2022-04-17
Toggle Overlay for the various kinds of goggles

Very-Soft:
image

image

image

image

### DIFF
--- a/code/modules/clothing/glasses/glasses_vr.dm
+++ b/code/modules/clothing/glasses/glasses_vr.dm
@@ -129,3 +129,78 @@
 		SPECIES_VOX = 'icons/inventory/eyes/mob_vox.dmi',
 		SPECIES_WEREBEAST = 'icons/inventory/eyes/mob_werebeast.dmi'
 		)
+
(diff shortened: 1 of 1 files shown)


Sageofaether12:
very poggers and very doggers, thank you soft

BonniePandora:
Thank god, now I can mine without a migraine-inducing overlay _having_ to be plastered across my screen.
Very noice

Novacat:
We talked about this a bit in staffchat, and, mostly want to bring a couple of points from there.

- There should be some visual indicator that you are wearing vision-enhancing goggles. Especially since it is a really, really bad idea to weld while having them on.
- You can already toggle the goggles on/off without having to take them off.

My own personal thought is that having enhanced vision should have downsides, and I am not a fan of the idea of just being able to turn this on at the start of the shift and just forget you have them.

That said, the colors of some of the special visions (esp mesons) are a bit irritating and could probably be changed to something less eye-grating.

Very-Soft:
Not desired per staff discussion

PR #12777, opened 2022-04-18
Makes Vision Overlays Not Painful

Novacat:
Tested and works.

Increases transparency of the vision overlays to make them not painful.


KillianKirilenko:
Why not  just change `alpha` to 30/20/10 or whatever other value?

Novacat:
That is what I am doing now.

PR #12778, opened 2022-04-18
Make meson vision less angry green

Very-Soft:
since we can't have the toggle


Novacat (approved):
Approved. Should probably be merged alongside https://github.com/VOREStation/VOREStation/pull/12777 to help make for a better overall user experience.

PR #12781, opened 2022-04-18
Removing Debug Text from Stumble Vore PR

FartMaster69420:
I'm an idiot and left this in here.


FartMaster69420:
Also re-enables stack_vr.dm, looks like another PR disabled it by mistake

PR #12785, opened 2022-04-18
[MIRROR] Removes /obj/item/weapon, /obj/item/device [MDB_IGNORE]

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/8317
--------------------
I'm gonna regret putting this up now because there's other big PRs that are gonna conflict with this massively but here goes:

Mechanical changes:
Any /obj/item can cleave. Previously just /obj/item/weapon.
Icons should all be in working order. I went through searching for every icon state in icons/obj/weapons.dmi and icons/obj/device.dmi.


Atermonera:
FYI, this will probably be easier to merge by doing the following steps yourself:
The `update icons` commit from the original PR was going through all the icons in `weapons.dmi` and `devices.dmi` and finding what things used them, then updating the `icon` vars of those things to explicitly reference the .dmi's

Find/replace
`obj/item/weapon/` with `obj/item/`
`obj/item/device/` with `obj/item/`
The trailing / makes sure you're only changing _paths_ and not _variable names_ (There's a couple instances of that). You could use a leading `/` but it's not necessary. I used the fully-qualified paths for safety.

PR #12801, opened 2022-04-19
Secret Tech - Radiocarbon Spectrometer Buff

FartMaster69420:
What this does is create a framework for designing hidden tech levels for items, that are only unlocked when that item is scanned in a radiocarbon spectrometer.

This allows for gating of techs behind some gameplay of the spectrometer mini game, reducing tedium of RnD, and giving the spectrometer an actual in game use.

So far the base system of adding techs works, this will require further discussion for what techs should be gated behind this system.

### DIFF
--- a/code/modules/research/secret_techs_vr.dm
+++ b/code/modules/research/secret_techs_vr.dm
@@ -1,0 +1,18 @@
+/*
+	The secret tech system! This is to add some more gameplay to RnD beyond the basic tech speedruns and deconstructing.
+
+	Basically, we setup origin techs on an item and also secret techs, when the item is successfully scanned in a radiocarbon spectrometer (code/modules/xenoarcheaology/tools/geosample_scanner.dm), then it will add secret techs to it's origin tech list.
+*/
+
+/obj/item
+	var/list/secret_tech = null
+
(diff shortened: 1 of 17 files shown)


code/modules/research/secret_techs_vr.dm
```diff
@@ -0,0 +1,18 @@
...
+
+// This is the proc for updating tech, calling it will add secret_tech into origin_tech, then set secret_tech to null so that one can't just infinitely increase an item's tech level.
+/obj/item/proc/update_techs()
+	for(var/I in origin_tech)
+		for(var/T in secret_tech)
+			if(CallTechName(T) == CallTechName(I))
+				origin_tech[I] = secret_tech[T] + origin_tech[I]
+				secret_tech.Remove(T)
+	origin_tech += secret_tech
+	secret_tech = null
```
silicons, line 18:
```
for(var/i in secret_tech)
    if(i in origin_tech)
        origin_tech[i] += secret_tech[i]
    else
        origin_tech[i] = secret_tech[i]
secret_tech = null

Cameron653:
Same reasoning as last PR. This has been sitting for 5 months now.

PR #12815, opened 2022-04-22
Secret Science

FartMaster69420:
This is a revamp to the science system, setting a cap for items tech origins.

For now, the tech cap is set to 4, however consider that a placeholder. I specifically made TECH_CAP a variable, so it can easily be tweaked to fit whatever cap the community finds ideal.

If an item has a tech level that's over the cap, it's dropped to the cap and the difference is added to a 'secret_tech' variable.

This secret tech can be unlocked by scanning the item in a radiocarbon spectrometer.

To facilitate finding techs, a handheld scanner has been added, the 'Portable Resonate Analyzer' (Yes, taken from unused borg code and revamped to actually work in this use case.) This item can be used to scan an item, getting a list of the origin techs, and a special message if there are hidden techs available on it.

This should hopefully make RnD a more engaging task to preform, adding more depth than just deconstructing a meta list.

There are several instances where the item cap is purposely avoided being used. If the item has the variable 'cap_override' set to true, it will not have it's origin_techs touched. The other scenario is if secret_techs are defined for an item. There's no point in creating a secret_tech list if one already exists, and if secret_tech is already defined I assume the developer set it for a reason.


Heroman3003:
Needing to use xenoarcheology machine to examine normal station stuff and finds and even stuff printed from the RnD just to unlock levels seems like bloat and attempt to force use of a mechanic that was specifically made to be fluff-only in the past BECAUSE it was a long, slow, clunky bloat seems like wrong way to approach fixing anything related to RnD's design imo. This is also not ADDING anything, this is restricting current stuff thats already ranges from "useless" to "maybe kinda cool". I stand by opinion I voiced in discord, that if hidden tech stuff is to be added, it should only be new items behind it and new items that have it, not locking tech of current items behind the system, nor require system to be used to make currently makeable items. This isn't a rework, this is a tack-on.

FartMaster69420:
Just giving more unlocks to science doesn't fix the underlying issue of speed running RnD giving all techs in about ten minutes.

There are large amounts of printables that are pointless because they're made as stopgaps between levels that just don't exist in the current system. (T2 techs, Upgraded T-Ray analyzer, A bunch of the medical items such as laser scalpels)

Heroman3003:
A lot of these items exist as stop-gaps for the "before materials but after pre-material RnD". This locks out more than that and IMO ultimately not a solution that makes RnD less speed-runnable as it practically changes nothing about the process or order or anything, other than putting in an extra step that was REMOVED from xenoarch as a necessety due to being unnecessarily tedious. It doesn't counter it being rushable or speedrunnable, it just makes it take longer.

FartMaster69420:
If it make something take longer, by definition it is less speedrunnable.

Like it or not, materials are no longer a viable system for locking progression due to the mat fridge, science will almost always have decent mats at shift start.

Heroman3003:
> If it make something take longer, by definition it is less speedrunnable.
> 
> Like it or not, materials are no longer a viable system for locking progression due to the mat fridge, science will almost always have decent mats at shift start.

Making something less speedrunnable and making something take more time to speedrun are two different definitions in my eyes. To make something less speedurnnable, you add variables, uncertainties, reliance on other people, etc. Something that prevents a person from following a perfect pre-made path on their own every single time. Adding extra steps doesn't make it less speedrunnable because steps just become part of the speedrun, and sure, it takes longer, but its still a same static way to follow every time with no variation. And adding plain RNG to current system the way it is is not a solution either tbh, because said system is not designed to handle it.

FartMaster69420:
>And adding plain RNG to current system the way it is is not a solution either tbh, because said system is not designed to handle it.

This PR doesn't add any RNG though?

FartMaster69420:
Now supports borgs loading the spectrometer with click-drag.

lm40:
yeah unless i'm misunderstanding this i'm not a fan either, making things more tedious is not really the ideal solution although i would like if something could be done to r&d to make it less of a ten minute shopping list blitz

code/modules/research/secret_techs_vr.dm
```diff
@@ -0,0 +1,73 @@
...
+		if(T in origin_tech)
+			origin_tech[T] = secret_tech[T] + origin_tech[T]
+		else
+			origin_tech.Add(T)
+			origin_tech[T] = secret_tech[T] + origin_tech[T]
+	secret_tech = null
+
+// This gives us a hardcap at initialization.
+/obj/item/Initialize()
+	..()
```
silicons, line 31: initialize absolutely needs to return an INITIALIZE_HINT

use
. = ..() for the default

code/modules/xenoarcheaology/tools/geosample_scanner.dm
```diff
@@ -91,10 +91,13 @@
 		if(scanned_item)
 			to_chat(user, "<span class=warning>\The [src] already has \a [scanned_item] inside!</span>")
 			return
-		user.drop_item()
-		I.loc = src
-		scanned_item = I
-		to_chat(user, "<span class=notice>You put \the [I] into \the [src].</span>")
+		if(!isrobot(user)) // VoreEdit, fixing a bug that allowed borgs to add modules to the scanner.
+			user.drop_item()
+			I.loc = src
```
silicons, line 96: I.forceMove(src)

code/modules/xenoarcheaology/tools/geosample_scanner.dm
```diff
@@ -357,3 +367,30 @@
...
+// VORE Add, gives borgs access to the sampler.
+/obj/machinery/radiocarbon_spectrometer/MouseDrop_T(var/obj/item/I, var/mob/living/user)
+	if(!istype(I))
+		return
+	if(user.stat || user.incapacitated(INCAPACITATION_DISABLED) || !istype(user))
+		return
+	if(!isturf(I.loc))
+		return
+
+	I.loc = src
```
silicons, line 379: I.forceMove(src)

Cameron653:
This has been sitting for 5 months with no further discussion and seems to be largely against the PR. Going to close this.

If more discussion is going to be had, feel free to re-open this PR.

PR #12821, opened 2022-04-24
Solar Tweak!

Very-Soft:
Makes the max distance a var instead of a define, that way it can be different depending on the situation.

It defaults to 40, so the default behavior is maintained.
I could probably define this on the map instead, but that doesn't really allow for one Z having one setting and another having another. So I decided to do this instead.


Novacat:
This could probably be simplified a bit by using var/SOLAR_MAX_DIST instead of var/max_solar_distance. Lessens the chances of future conflicts, as well.

PR #12865, opened 2022-04-30
[MIRROR] Atmospheric Retention Fields Port

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/8582
--------------------
Just a little downstream project of mine that I figure I might as well push upstream for compatibility's sake.

> A high-tech alternative to the humble firedoor, atmospheric retention fields activate when an area's air or fire alarms are triggered, generating a shimmering forcefield that prevents airflow across it. They are fairly power-hungry, but more convenient than firedoors as they prevent airflow without preventing movement. Alternately a more power-efficient impassable variant also exists. Their activation may not be instant, but it's quick enough that it can help save lives.
> 
> image
> 
> Adjacent fields link together to form more visually-pleasing barriers in any shape or size imaginable.
> image
> 
> If shut down by an EMP, fields will come back online with a slight randomized delay.
> staggered_repower
>
> ARFs run off the Environmental circuit, so they should be the last things to shut down when an APC starts to run flat. They can also be tinkered with using various tools to adjust their behaviour, or just disassembled entirely: ex. you can pop one open using a crowbar and either cut the wires to sabotage it or toggle it into always-active mode using a multitool.

Their functionality hasn't changed significantly since they were initially added, though since you guys don't have /tg/lights (yet) I've commented out the associated lighting settings. Once #8526 is merged, you can just uncomment that block and it should look nicer.

These are 110% Coder Art and could _really_ do with some nicer sprites for both the emitter module and the field itself, and maybe some SFX for shutdown/powerup, but otherwise are fully functional.

## Todo?
- [ ] Nicer Field/emitter sprites
- [ ] Power up/down & ambient SFX
- [ ] Wall-mounted subtype?

## Changelog
:cl:
add: Added atmospheric retention field generators, which can be built after science has done some research.
/:cl:


KillianKirilenko:
This can be safely closed rather than merged. It has some minor optimizations but I'll redo those manually later since it also has some conflicts due to other unmerged refactors.

PR #12869, opened 2022-04-30
Dosimeters

ItsSelis:
The first version of the dosimeters.

The radiation levels might need to be adjusted for the films (when they start to darken and turning black).

Thanks to @BonniePandora for the dosimeter sprites!


Novacat:
I like this.

code/modules/vore/fluffstuff/custom_items_vr.dm
```diff
@@ -620,6 +620,139 @@
 	new /obj/item/weapon/paper/khcrystal_manual(src)
 	new /obj/item/clothing/accessory/collar/khcrystal(src)
 
+/obj/item/clothing/accessory/dosimeter
```
Novacat, line 623: This should probably be moved to a different file. custom_items_vr.dm is very much a bloated file and realistically should probably only contain fluff items. I recommend VOREStation\code\modules\clothing\under\accessories\badges_vr.dm (You will have to create a few file so make sure to add it to the dmi)

Novacat (changes requested):

code/modules/vore/fluffstuff/custom_items_vr.dm
ItsSelis: Yeah, I noticed too that it seemed pretty bloated. The general idea was based around the life crystal so I figured it would be the right file.

I moved it to the badges_vr one now.

Novacat (approved):
Could probably be a little bit more available but that can come later.

ItsSelis:
Yeah its not perfect yet, the radiation threshold could be perfected too in the future as well.

PR #12877, opened 2022-05-01
Makes "Succumb" an IC verb, adds warnings

Runa-Dacino:
Places the "Succumb" verb into the "IC" tab with a description. Using it requires clicking "yes" twice to ensure no accidental use.

Tested on local:

Works for cases where health is over 0 (rejects, regardless of yes/no).
Works for cases where health is under 0 (rejects when either 1st or 2nd dialogue was a no)
Works for cases where dead (rejects, regardless of yes/no.)

### DIFF
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -79,14 +79,21 @@
 	return 1
 
 /mob/living/verb/succumb()
-	set hidden = 1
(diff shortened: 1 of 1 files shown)


Novacat (approved):
Seems fine to me. Just double check to make sure this works.

Runa-Dacino:
> Seems fine to me. Just double check to make sure this works.

It works on local for synthmorphs and organics alike.

PR #12879, opened 2022-05-01
Adds belt slot flag to proto-kinetic machete and subtypes

Runa-Dacino:
Enables wearing proto-kinetic machetes, daggers and future weapons that are below ITEMSIZE_HUGE on your belt-slot, like you can with plasteel machetes and various tools/knives.

ITEMSIZE_HUGE was specifically chose to prevent wearing gauntlets on belt, as doing so sounds ridiculous (greaves/gauntlets you quickdraw from your belt?).

Wearing a sword/dagger in your belt however makes more than sufficient sense to me.

Balance concerns:

- Miners/explorers may store their proto-kinetic sidearms in one more storage slot
- Caveat: Explorers can already store theirs in explorer belt.
- Caveat: Miners can NOT store theirs in mining belt
- Caveat: Since miners can NOT store theirs in their belt, and likely want to bring a belt with them AND a mining satchel - they have 3 items competing for belt slot.
  - Belt in belt slot: easier access to drill, GPS, umbrella and marker beacons. If not worn, must have 2 hands free to do the same.
  - Satchel in belt slot: Automatically collect dropped ore. If not worn, must be held in hand (does not collect from bag).
  - Plasteel/Proto-kinetic machete in belt allows for quick-draw if main weapon (KA, phaser) is dropped. If in bag, need to fumble for longer.
- Proto-kinetic dagger could ALREADY be carried in pocket slot.
- Plasteel Machete could ALREADY be carried in belt slot. 

As such, I do not believe allowing miners/exploration to wield their backup side-arm in an easy-to-reach slot is a balance concern. Doing so requires sacrificing convenience elsewhere.

Tested on local:
- Can wear machete/dagger on belt
- Can NOT wear crusher nor gauntlets on belt
- Crusher, machete, gauntlet and belt function as expected against mobs.

### DIFF
--- a/code/modules/mining/kinetic_crusher.dm
+++ b/code/modules/mining/kinetic_crusher.dm
@@ -256,11 +256,9 @@
 	backstab_bonus = 40 // 100
 	thrown_bonus = 20 // 120
 	update_item_state = FALSE
+	slot_flags = SLOT_BELT
 
-
-/obj/item/weapon/kinetic_crusher/machete/Initialize()  //Making sure you can't store gauntlets/greaves on your belts.
-	if(w_class < ITEMSIZE_HUGE)
-		slot_flags = SLOT_BELT
+		
 
 
 /obj/item/weapon/kinetic_crusher/machete/gauntlets
@@ -280,6 +278,7 @@
 	detonation_damage = 37 // 75
 	backstab_bonus = 55 // 130
 	var/obj/item/offhand/crushergauntlets/offhand
+	slot_flags = null
 
 /obj/item/weapon/kinetic_crusher/machete/gauntlets/equipped()
 	. = ..()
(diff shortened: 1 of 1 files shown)


Novacat (changes requested):
This is a really weird way to do it. I would recommend just setting the kinetic machete slot flags to SLOT_BELT and just add slot_flags = null to kinetic machete's subtypes where the slot flag is undesirable.

Runa-Dacino:
> This is a really weird way to do it. I would recommend just setting the kinetic machete slot flags to SLOT_BELT and just add slot_flags = null to kinetic machete's subtypes where the slot flag is undesirable.

Did as requested. However, doing so also removes ability to equip items on slot_back. Not an issue presently, since only gauntlet exists as inheritence and gauntlet on back doesn't make sense either. However, I made it based on w_class thinking in the future there may exist new weapons inheriting from machete that would make sense on the belt slot again (say, a longsword or katana).

Runa-Dacino:
> > Did as requested. However, doing so also removes ability to equip items on slot_back.
> 
> You should be able to specify multiples by seperating them with a pipe (|), so `SLOT_BACK|SLOT_BELT` ought to work.

Noted, and thank you! I do feel that at least it's fine not to allow gauntlets/greaves to be on backslot.

PR #12881, opened 2022-05-01
Pocket pouch rebalance + obtaining method

TankTheBirb:
You know those pocket pouches you can find sometimes except you don't know where you can get them? Well you couldn't aside from map spawns. This fixes that, you can now make them using cloth, fur, or leather. Leather lets you make the really nice ones because its harder to get, cloth and fur let you make the smaller and specialized ones. Botany is your friend, Wabbacks let you grow cloth, teppi provide fur, and any animal can give you leather. There are also alternate ways to get these things too if botany is kill (unfortunate), but they're more time intensive than just asking the botanist.

Also I changed the contents of specialized pouches to be somewhat more sane because the specialized pouches before were really restrictive and inferior in just about every way. So now there is a balance to it. Tools pouch is basically a belt for your pocket but half size. Supplies can carry actually useful supplies. Parts can carry the most of any pouch but literally only stock parts. Medical can draw stuff faster from their pouch and it can carry more medical related supplies than it could before. Ammo lets you carry grenades and seismic charges now. Baton pouch is now the melee pouch and can fit more than just batons in it (still only one of them tho).


code/game/objects/items/weapons/storage/pouches.dm
```diff
@@ -58,26 +58,46 @@
...
+		/obj/item/device/radio/headset,
+		/obj/item/device/robotanalyzer,
+		/obj/item/weapon/material/minihoe,
+		/obj/item/weapon/material/knife/machete/hatchet,
+		/obj/item/device/analyzer/plant_analyzer,
+		/obj/item/weapon/extinguisher/mini,
+		/obj/item/weapon/tape_roll,
+		/obj/item/device/integrated_electronics/wirer,
+		/obj/item/device/integrated_electronics/debugger, 
+		/obj/item/weapon/shovel/spade,
```
Novacat, line 97: Pretty sure this is itemsize large which is way too big.

Novacat (changes requested):
To preface: These are supposed to be put into pockets, which can normally only hold itemsize_small items. You are basically taking the drawback of having to take time to pull items out of your pockets, in exchange for being able to put four things in your pockets.

The medical pouch is way too strong. Faster item withdrawal + more items is kind of too much.

(That said, I also think the bluespace pouch should be toned back a bit)

code/game/objects/items/weapons/storage/pouches.dm
TankTheBirb: Also yeah medical pouch probably draws too fast for being a medium. I'm thinking I will make it small size and keep fast draw so its stuff like "I can put my health analyzer in here and and also have my pill bottle/auto injector in somewhere actually accessible". Or some other "I would put it in my pocket if I didn't need better things there" since that's vaguely the idea of pouches. Admittedly with the specialized ones I didn't feel too bad with giving them nice potential because the basic medium one is so much better in every way and the large size one is even better than that.

Also yeah the BS pouches are BS, full backpack carry size in your pocket with only a normal size item restriction. I didn't touch the balance on those because they're too good and I don't know how to fix that other than cut the storage size down.

> Pretty sure this is itemsize large which is way too big.

https://github.com/VOREStation/VOREStation/blob/5d9f0462c19c2b7fb53bab24ef4d54b87f455ba6/code/modules/mining/mine_items.dm#L171 its this one, which has ITEMSIZE_SMALL. You may be thinking of the normal/wood shovels that are normal size.

Novacat: Ahhh.

TBH, cutting the storage size down would be enough. BS Backpacks are, iirc, roughly twice the space of a regular backpack (might be 2.5x the space, I forgot exactly), so bluespace pouches should be INVENTORY_POUCH_SPACE*2, which would give it 8 slots which is slightly more than a box which is pretty big.

TankTheBirb: I dunno, I think I'll leave it alone for now since it was outside the scope I intended for this PR.

Medical stuff got updated, deleted a slot so its 1 normal 1 small or 3 smalls with the same quick draw, BUT, the quick draw is slow enough that you need to briefly stop to pull stuff out. Also removed the defib as an option and added lollipops.

PR #12886, opened 2022-05-02
[MIRROR] Remove carbonmob dependencies

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/8561
--------------------
Broken out from #8546 
This stuff is actually MOSTLY reviewable.
Changes:
- [x] Breathing is specific to humans
- [x] Diona nymphs are now simple mobs (No specialized AI)
- [x] Brainmobs are mob/living
- [x] Deleted carbonmob xenomorphs and larvae.
- [x] All mobs can pose.
- [x] Virus2 can only affect humans.
- [x] Deletes loyalty implants.
- [x] Borer procs are only on humans.
- [x] Only humans can be handcuffed and legcuffed.
  * This is feature-parity from previous, because carbonmob xenos are deprecated, how do you handcuff a brain, how do you handcuff a nymph.
  * I deliberated over making the handcuffed var actually be on mob/living, because simple_mobs can have hands by setting `has_hands = TRUE`, but decided towards feature-parity for simplicity.
- [x] Removed stomach_contents, because that should be on the stomach organ
- [x] Removed the declaration for metabolism vars, and broke some organ procs. This is intentional, I have to rewrite that logic from the ground up as part of the overall new system.
- [x] Species var is now on mob/living. There's definitely room to leverage it on simple_mobs, but I'm not going to implement that here.


Runa-Dacino (commented):
I strongly feel that moving brainmobs from carbon to simply living is going to break the transcore notification added for posibrains after gibbing/digestion...

Very-Soft (commented):
I am pretty sure this will break a lot of things. Many of the things this seems to be aimed at just aren't problems here, so I'm not even sure how worth it it is to fix.

izac112:
> I am pretty sure this will break a lot of things. Many of the things this seems to be aimed at just aren't problems here, so I'm not even sure how worth it it is to fix.

_not my repo but_ think its related to [https://github.com/PolarisSS13/Polaris/pull/8603](https://github.com/PolarisSS13/Polaris/pull/8603) which if you intended to mirror once/if it becomes a thing you might need the changes in this, though i have not read enough of the code to be sure, just something to keep in mind

KillianKirilenko:
Izac's right. This is just a precursor to more changes. It'll be necessary if we want those changes.

@Very-Soft I think at this point you and the rest of the staff need to sit down and have a serious discussion about the state of PRs and catching up to Polaris, and whether or not we should try to maintain parity or formally split and only port content selectively. At the very least there needs to be *some* kind of hold on merging new stuff (besides trivial things like rotating the map) until a formal decision is made, because every step taken is going to risk making it harder to fix the issues that currently exist.

Sypsoti:
This was not merged into our master, but into the Goosemed branch. Just making sure you're all aware.

Atermonera:
fyi, both this and the "deletes /mob/living/carbon" PR were done into a dev branch, because they're actively part of a medical rewrite that I'm working on

This definitely breaks reagent metabolism and you should not worry about merging these

KillianKirilenko:
Good to know. My point stands in general but it's one less thing to worry about for now at least.

Mirrorbot probably needs to be tweaked to not copy PRs you guys merge into anything that isn't Master at the very least.

PR #12887, opened 2022-05-02
[MIRROR] Deletes mob/living/carbon

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/8546
--------------------
Re-using this PR to accomplish the second part of the groundwork: deleting `/mob/living/carbon`. #8561 accomplishes the groundwork for this groundwork, by removing all of the other dependencies upon carbonmobs, such that the only thing left on carbon is now human mobs, and therefore, all carbon mobs _are_ humans. Thus, we can do some simple find/replaces to replace references to carbons with references to humans. There's probably a lot of typecasting from human->human now, that... will just have to be scrubbed out long-term. Because this is 16 _thousand_ lines already.


Atermonera:
fyi, both this and the PR to clean up the dependecies were done into a dev branch, because they're actively part of a medical rewrite that I'm working on

This definitely breaks reagent metabolism and you should not worry about merging these

PR #12897, opened 2022-05-04
Cocoon Metamorphosis 

FartMaster69420:
Adds a cocoon spinner trait, allowing for characters to place themselves into a little cocoon and transform their appearance.

Also makes a slight edit to the (non-vore) appearance menu, fixing the visual preview to actually show the player instead of the turf they're on.

Note, TF mechanics in the cocoon will not allow for changing race. Only for changing appearance.


Screemonster:
You know what I've been wanting this for xenochimera for _ages_ as an alternative to burning a 1-hour cooldown on their revive to change their ears or whatever

Very-Soft (changes requested):
I would prefer if this was not a neutral trait. It should be positive and cost 1 or 2 points.

It doesn't provide a big advantage or anything, but it's really one of those things I don't think just everyone should be doing for free.

FartMaster69420:
Done. Also found and fixed an issue leading to custom species not being able to access the list of wings/tails/ears in the appearance menu.

PR #12906, opened 2022-05-06
Making "Walkpods" a loadout option

ItsSelis:
Someone said it would be pretty cool if it was possible.

Generally some music is nice though.


Very-Soft:
Don't really like that all the cool things are going into the loadout but sure whatever

PR #12907, opened 2022-05-06
Integrated circuit camera can now use different networks, plus modular computer camera access change

foopwo:
Allows circuit tinkerers to change which network a circuit camera will connect to, no longer will science have all the drone making fun!
Also allows people with engine access to view the engineering camera network on their tablets/laptops, just like they already can over the fixed camera consoles.


foopwo:
Also adding qm access to mining cams since they've got a console for it.

PR #12910, opened 2022-05-07
NSB Ambrosia

Youtubeboy139:
Take a trip on the planet Accerion, a little known jungle planet located in the system near Virgo. Promise it'll be a real treat... or rather, make you into a real treat. (Lighting still needs fixed - other changes might be needed)

[Entry for the Map Contest - DNM]

164951496-894d4f9f-b4c3-4fed-8a5b-c399f5997fd0


Youtubeboy139:
Map itself is still busted so this PR dies with it's conflicts

PR #12941, opened 2022-05-09
[MIRROR] Removes dnaswitch artifact effect

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/8625
--------------------
It's outright broken for FBPs. The mutations of DNA2 are largely unsupported/legacy, and the more we cut out things that tie into DNA2, the more likely we are to actually be able to eventually cut out that tumor entirely to _actually_ replace it with a new system.


Very-Soft:
this only has polaris's dme in it

Heroman3003:
Fixed version was merged.

PR #12944, opened 2022-05-09
Make Cooking Great Again [Tedium Update, Rebalance some cooking aspects]

Somememeguy:
Background:
--------

Cooking, even prior to the Aurora cooking update, was always a bit of a tedious task to do.

Many clicks, very time consuming, and with the Aurora update it seemed to make things a bit more tedious (even if mass producing was nice).

This sentiment was shared with a few other chefs players that I know, so I decided to make this update to try to alleviate some of these issues.

What's changed?:
--------

Well, the important changes are as follows:

-Rebalances some of the recipes, primarily by either lowering the required amount of items, or by making them produce multiple of the items (The bun recipe now produces three buns instead of one, etc)

-Adds in Cake Batter, a reagent that replaces the some of the ingredients required to make some cakes. Recipe is 15 flour, 10 milk, 15 sugar, and 1 egg. Results in 60 units. Bit less ingredients in total, especially EGG.

-Makes it so the plant bag can be used on the microwave to dump plants into it

### DIFF
--- a/code/modules/food/kitchen/microwave.dm
+++ b/code/modules/food/kitchen/microwave.dm
@@ -147,6 +147,34 @@
 				"<span class='notice'>You add \the [O] to \the [src].</span>")
 			SStgui.update_uis(src)
 			return
+	else if (istype(O,/obj/item/weapon/storage/bag/plants)) // There might be a better way about making plant bags dump their contents into a microwave, but it works.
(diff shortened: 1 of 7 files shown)


Very-Soft (commented):
You're going to want to figure out how to update the recipe page for this, since I imagine some of the changes are going to confuse people!

PR #12950, opened 2022-05-10
Fixes heart damage/nonexistence having no impact on functional blood levels

lm40:
I haven't tested this exhaustively but at least for humans/normal species this causes you to actually die if your heart is removed.


Very-Soft:
I actually accidentally deleted everyone on the server's heart and no one noticed.

PR #12951, opened 2022-05-10
Silicons now have omnihud and EMT borgs have inflatables

TankTheBirb:
Changes the toggle_sensor_mode proc to toggle omnihud panes on or off. It gives the same panes as the AR-Bs do, so you get all the relevant information you could want. Also because sechud and medhud stinky and everyone and their [species of pet] has AR which is SO GOOD, so the fancy silicons get it too now.

Mostly messes around with silicon files but also changes the examine behavior for records on borgs to conform to the new system. Now its just "gib records if hud"


TankTheBirb:
I am not smart enough to un-merge these two with whatever I did so I can remove the inflatables change and slap that in a separate PR now that I know what I'm doing if asked to do that :v

PR #12953, opened 2022-05-11
Sleevecard simplification and features

FartMaster69420:
Guts most of the sleevecard/infomorph code and replaces it with a single file, making sleevecards a subtype of paicards. Likewise, infomorphs are now a subtype of pai. This will allow sleevecards to be used for anything PAI cards can be used for, such as controlling bots.

Functionally sleevecards aren't much different, the majority of their code was just copy/pasted from pAI code already. Directives software was removed from the software interface for sleevecards, since they aren't lawed AI.

The ability to add a mind to the sleevecard from a Sleevemate was also added.

The old files aren't deleted, just disabled.


Novacat:
Does the old file have anything worthwhile? It might just be better to delete/replace entirely.

FartMaster69420:
I don't think there's anything special in the old files, but I figured I'd leave that choice up to maintainers.

Novacat (changes requested):
Go ahead and replace it. If we desperately need the old file for whatever reason (and chances are we probably will not) we can just look at git history.

Novacat (changes requested):
Can you rename sleevecard2.dm to sleevecard.dm?

code/modules/resleeving/sleevecard.dm
```diff
@@ -1,60 +1,33 @@
...
-	if(!infomorph)
-		to_chat(user,"<span class='warning'>\The [src] does not have a mind in it!</span>")
-	else
-		to_chat(user,"<span class='notice'>\The [src] displays the name '[infomorph]'.</span>")
-
-//This is a 'hard' proc, it does no permission checking, do that on the computer
-/obj/item/device/sleevecard/proc/sleeveInto(var/datum/transhuman/mind_record/MR, var/db_key)
-	infomorph = new(src,MR.mindname,db_key=db_key)
+/obj/item/device/paicard/attack_ghost(mob/user as mob)
+	return
```
Very-Soft, line 12: Pretty sure that doing this will make it so ghosts can't join unattended pAI cards anymore, which is basically the only way anyone actually plays pAI regularly.

FartMaster69420: Ah heck, that was supposed to be paicard/sleevecard

FartMaster69420: Fixed

PR #12955, opened 2022-05-11
[WIP] Modular PDA Port

ItsSelis:
These will most likely not entirely replace the existing ones. 
But it would be neat to have something similar that could be worn in the ID Slot, using the same kind of cartridges etc.

Of course, I am very open to any feedback on this, or if this should be continued at all.


Novacat:
You should look up the early version of the Modular Computer PDA port which basically just added MC functionality to existing PDAs.

ItsSelis:
> You should look up the early version of the Modular Computer PDA port which basically just added MC functionality to existing PDAs.

Yeah, heard about that a few times already, will look it up soon (or tomorrow most likely).

Cameron653:
Work still being done on this? Been sitting for 4 months now. I don't want to close it since I know TGUI related things are fucky.

If it's still being worked on, do mention it here. If it's not being worked on/ feel free to close the PR. 
I'm going to look at this in a few days and close it otherwise.

PR #12969, opened 2022-05-14
[MIRROR] db flex tape

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/8627
--------------------
**A**. removes unused forum sql configuration

**B**. adds some up-front config guards to save a little work on spamming db connect attempts even when sql is disabled

**C**. Adds SSmisc_slow, a 5 minute wait SS.

The only behavior of SSmisc_slow at the moment is DoReconnects, which handles connection maintenance for dbcon and dbcon_old by:
- checking if they're connected and sending a tiny safe query to keep them open.
- or if not, making sure they're closed, then reopening them with new handles.
- and if it fails this too many times, giving up softly.

With logging at every step.

This should remove the need for someone to push the fix database connection button by simply not allowing valid connections to time out when the lobby's been sitting empty for a half day stretch.

~~Draft because I want to make sure of DoReconnects locally first.~~


Cameron653:
Closing this. If someone wants to come through later on and work on this, feel free, because I have no idea what this does and it's DB stuff.

PR #12972, opened 2022-05-14
[MIRROR] Nerfs Hyperzine

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/8634
--------------------
Heavy miscalculation on my part. You are now reasonably speedy.

Also makes psilocybin actually hallucinogenic and removes a very annoying hallucination interaction with airlocks.

IDK why drug.dm fiddled with spaces so much. My actual line changes are just 145, 160, and 169.


Skysteak:
Just so it's on record, I don't think a nerf is necessary on this server.

Wickedtemp:
Our move speed is ass. Let people be speedy.

Cyantime:
This is literally a buff to the current form of hyperzine here, bringing it from speedboost 1 to 1.2.
The "Nerf" is to a polaris change merged two weeks ago cranking it all the way up to three.
Stop freaking out about nothing

Skysteak:
> This is literally a buff to the current form of hyperzine here, bringing it from speedboost 1 to 1.2. The "nerf" is to a polaris change merged two weeks ago there cranking it all the way up to three. Stop freaking out about nothing

Ah that's fair enough then! Thank you for the clarification!

Cameron653:
This PR would be annoying as shit to fix and needless. It touches airlock.dm which can't be quick-fixed either.

If we really want to adjust hyperzine we can just make a PR to adjust it to 1.2

PR #12996, opened 2022-05-20
Ports card icons from Eris. Implements them for VOREStation

Runa-Dacino:
Per CoolCrow420's request and with TheToaster's provision of Eris icons for a card game inspired by UNO, the card game was added and is available from the loadout at roundstart. It is not mapped in to game rooms.

Loadout cost is "1".

Tested on local (dealing, discarding, shuffling, placing cards back into deck, drawing, making a hand, splitting hands, concealing hands, examining hands).

The icons:

Cardback, deck:
image
Wildcards (4 of each):
image
Red (2 of each):
image
Yellow (2 of each):
image
Green (2 of each):
image
Blue (2 of each):
image
Hand (Need to examine to see last card played, may fix later but does not prevent use. Currently displays First card played)
image
Hand (concealed)
image


Very-Soft:
I do think this is cute and good, but I think the name of the game should change. I would rather we didn't get DMCA'd by Mattel.

Runa-Dacino:
Should be free of copyright, unless the idea of "red +2" is also problematic, but should not be.

code/modules/client/preference_setup/loadout/loadout_general_vr.dm
```diff
@@ -96,3 +96,8 @@
 	display_name = "schnapsen playing cards"
 	description = "French-suit playing cards! Pre-picked for 2-player mode."
 	path = /obj/item/weapon/deck/schnapsen
+
+/datum/gear/egy_game
+	display_name = "EGY playing cards"
+	description = "A deck of cards for playing EGY! Be the firs tto lose all cards!"
```
Very-Soft, line 102: 'firs tto'

Runa-Dacino: Fixed! Apologies.

PR #12998, opened 2022-05-21
Fixes Phaseshift "Gravity"

### LINKED ISSUE #12986: Phased shadekin are stunned when moving from no gravity to gravity
Phased shadekin, when moving from a turf with no gravity (such as space) onto a turf with gravity are stunned and turned sideways. This also causes them to drop anything they're carrying. This does not affect movement speed, allowing you to slide around while sideways.

Phased shadekin to be unaffected by anything.

#### What actually happened
I fell over and dropped my stuff in a wall.

#### Steps to reproduce
- phase as a shadekin
- move out to a space tile
- move back onto a non-space tile

ItsSelis:
Fixes #12986

### DIFF
--- a/code/__defines/species_abilities_vr.dm
+++ b/code/__defines/species_abilities_vr.dm
@@ -1,3 +1,0 @@
-// Shadekin
-#define AB_PHASE_SHIFTED	0x1
-#define AB_SHADE_REGEN		0x2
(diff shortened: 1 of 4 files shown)


Heroman3003:
I think a better check for that is is_corporeal proc check, which is almost same thing but with accounting for future possibilities of phase-shift-like scenarios

ItsSelis:
> 

Oh, that could work. Looking it up I did find incorporeal_move getting set. That is used in other files for the ability to walk through walls and such.

Though, the area does not check for that value when making someone stunned or fall down, which could be used instead of the moving of ability defines. (Observers are excluded from it anyways because they are not a living mob which area actually checks for)

PR #13002, opened 2022-05-22
Plating can be built on most ground turfs.

foopwo:
Exactly what it says on the tin. This PR lets people lay plating on floor turfs as long as these conditions are met.

- The turf doesn't have an obtainable tile type.
- The turf cannot be deconstructed by normal means.
- It is not a specifically forbidden turf type, which currently includes lava, water, ice, and maglev tracks.

Mineral turfs also no longer need a support lattice.


foopwo:
Side note, while testing, I found that it's possible to set tiles over lava, ice, water, and maglev turfs, resulting in some odd visuals. But this isn't an issue this PR introduces, you can do that on the current master already.

PR #13018, opened 2022-05-29
Duke's Mapping Competition: Burrito's Lake Map

Burrito-Justice:
This map was started at the very beginning of this month and I worked on it on and off all the way up to now. Currently the map has no real "objective" or anything interesting as far as story goes. It's just pretty to look at (or so I hope) and generally just a blank-ish hangout place with a neat aesthetic.


Burrito-Justice:
Here's some images (all taken in DM because I'm too lazy to re-tick and host to show the compiled screenshots BLEH)
image
image
image

PR #13025, opened 2022-05-30
Makes stomach noises save

lm40:
Why does almost everything else in the vore panel save but not that? Hopefully I did this right. I also considered making digest/absorb/whatever modes save too.

### DIFF
--- a/code/modules/vore/eating/living_vr.dm
+++ b/code/modules/vore/eating/living_vr.dm
@@ -234,6 +234,7 @@
 	P.vore_taste = src.vore_taste
 	P.vore_smell = src.vore_smell
 	P.permit_healbelly = src.permit_healbelly
+	P.noisy = src.noisy
 	P.show_vore_fx = src.show_vore_fx
 	P.can_be_drop_prey = src.can_be_drop_prey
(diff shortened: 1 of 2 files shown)


Very-Soft:
> I also considered making digest/absorb/whatever modes save too.

I think that this would be good, BUT! It would be GREAT if you could toggle whether it saves the digest mode per belly.

Like I'd love for my absorb bellies to just always be set to absorb, but like, mouth and stomach bellies I'd prefer to be on hold until I set them otherwise. 

I'm not sure how complicated that would be, but! I think it's a good idea at least~

Poojawa:
> > I also considered making digest/absorb/whatever modes save too.
> 
> I think that this would be good, BUT! It would be GREAT if you could toggle whether it saves the digest mode per belly.
> 
> Like I'd love for my absorb bellies to just always be set to absorb, but like, mouth and stomach bellies I'd prefer to be on hold until I set them otherwise.
> 
> I'm not sure how complicated that would be, but! I think it's a good idea at least~

Wouldn't be a complicated thing, but it would require setting yet another var to be serialized and recorded. 

It'd be as easy as copy/pasting if the belly is 'wet', then making the save code skip that one's setting so that it would default to the default.

PR #13028, opened 2022-05-30
(badly) implements saved belly modes

lm40:
okay I am kinda just making this because I wasted so much time trying to get this to work that I'm yoloing, I don't really think this is the best way to do it but it /works/

There is a pref for each belly that can be enabled or disabled to enable or disable saving the belly mode for that belly, defaulting to on.


Very-Soft (commented):
Conflicts

PR #13033, opened 2022-06-03
lets chaplains choose a fluff title and removes global religion

lm40:


code/game/jobs/job/civilian_chaplain.dm
```diff
@@ -92,9 +94,19 @@
 		new_deity = deity_name
 	B.deity_name = new_deity
 
+
+	var/new_title = sanitize(input(H, "Would you like to change your title?", "Title Change", I.assignment), MAX_NAME_LEN)
+
+	if (length(new_title) != 0)
+		I.assignment = new_title
```
Screemonster, line 101: I'd like to see _some_ sort of sanitation applied here so people can't just put an existing title in, like "Chief Medical Officer" but I guess we can just yell at people who do that.

lm40: I don't THINK this is super difficult to implement let me see

lm40:
i made a mess of this branch and im not good enough at git to fix it hold on i remake this

PR #13056, opened 2022-06-08
Tiny PDA UI Fix & Communicator Changes

ItsSelis:
Changes in this PR:

- Adds Reply button on the top of the Clipboard-Mode of the PDA Messenger
- One can now reply directly to messages received by the communicator, by pressing reply in the chat (Like PDAs)
- Communicators now display the message in the chat (similarly how PDAs do it)
- Communicators also now display the sent message in the chat (as a kind of feedback of actually replying to it)


ItsSelis:
> Do communicators relay your sent messages into the chat as well, like PDAs do?

Oh, that they don't do yet. They only really show the received message (showing the sent one too is probably smart because of people thinking that it might not have sent at all).

PR #13066, opened 2022-06-11
Make vchat warning only fire if vchat is enabled

Very-Soft:
ME USE OLD CHAT

ME NOT LIKE THAT IT WARNS ME THAT VCHAT ISN'T WORKING EVERY TIME WHEN I HAVE THE PREFERENCE FOR IT DISABLED

So, this makes it only warn you about vchat not working if you have vchat enabled.


Heroman3003:
PRAISE

PR #13070, opened 2022-06-12
Restart vote tweaks

Very-Soft:
Makes it so you can only trigger restart votes if there are no staff on/if you are staff

still needs to be tested

### DIFF
--- a/code/controllers/subsystems/vote.dm
+++ b/code/controllers/subsystems/vote.dm
@@ -362,7 +362,8 @@ SUBSYSTEM_DEF(vote)
 
 		if(VOTE_RESTART)
 			if(config.allow_vote_restart || usr.client.holder)
-				if(!GLOB.admins.len || usr.client.holder)
+				var/admin_number_present = send2irc_adminless_only(initiator_ckey, name)
+				if(admin_number_present <= 0 || || usr.client.holder)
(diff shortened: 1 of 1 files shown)


Heroman3003:
Imo, checking for staff being active is better because we do occasionally get situations where staff member goes to sleep while logged in and staff is not affected by automatic timeout. I believe the activity check is already used in either adminhelp or staffwho, where it gives a "even if no staff active ,we send the ahelps to out-of-game chat" message if staff is there, but inactive; or not there at all.

lm40:
i honestly think this ability should maybe just be removed, i've never seen it legitimately used once and i have seen it illegitimately used numerous times. while in theory it serves a purpose there really should be no circumstance where the server is operational enough for people to vote but not a single admin can possibly come on to press a single button

Very-Soft:
> i honestly think this ability should maybe just be removed, i've never seen it legitimately used once and i have seen it illegitimately used numerous times. while in theory it serves a purpose there really should be no circumstance where the server is operational enough for people to vote but not a single admin can possibly come on to press a single button

I mean it's totally possible to disable it, but I do not have access to the config file afaik, and I would need that in order to do so unless I just decided to break this vote, which I would rather not do.

PR #13074, opened 2022-06-12
Adding myself to the CODEOWNERS for TGUI

ItsSelis:
For any TGUI related files.

This will just auto request my review on such files (or at least I will get a notification on GitHub about it).

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -4,9 +4,9 @@
 # In the event that multiple org members are to be informed of changes
 # to the same file or dir, add them to the end under Multiple Owners
 
-# MULTIPLE OWNERS
+# ItsSelis
 
-/code/__DEFINES/tgui.dm @ItsSelis @ShadowLarkens
-/code/controllers/subsystem/tgui.dm @ItsSelis @ShadowLarkens
-/code/modules/tgui @ItsSelis @ShadowLarkens
-/tgui @ItsSelis @ShadowLarkens
+/code/__DEFINES/tgui.dm @ItsSelis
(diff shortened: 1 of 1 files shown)


Heroman3003:
To note, last person that did this quickly undid this as iirc it will also affect all mirrors and downstream PRs, but I may be wrong in that regard.

ItsSelis:
> To note, last person that did this quickly undid this as iirc it will also affect all mirrors and downstream PRs, but I may be wrong in that regard.
> 
> EDIT: Actually, nevermind, that was something else in addition to this I think that caused it.

Oh?
I don't think that I would get them for forks, I mean, that would be pretty crazy.
Though I could imagine that maybe being the case if the downstream has their repo standalone (like we do) and getting their updates via a mirrorbot.

Very-Soft:
I asked Aronai about this and he indicated that he didn't think this would do what you wanted it to do, and doubted that it ever worked for the person who was on this originally either, since neither of you have write access to the git.

I don't know anything about these matters! Though! Since Shadowlarkens doesn't contribute here anymore you could actually remove them from the file since it is quite unlikely that they will return!

ItsSelis:
> I asked Aronai about this and he indicated that he didn't think this would do what you wanted it to do, and doubted that it ever worked for the person who was on this originally either, since neither of you have write access to the git.
> 
> I don't know anything about these matters! Though! Since Shadowlarkens doesn't contribute here anymore you could actually remove them from the file since it is quite unlikely that they will return!

Yeah I looked it up a bit. Apparently it indeed only works with those on write access (due to the reviews of such people being able to actually block the merge from happening at all if changes are requested).

A bit unfortunate but it is not that critical either.

PR #13075, opened 2022-06-12
Makes unathi breacher rig (NT version) orderable at cargo

Heroman3003:
At certain liznerd's request.

300 cost, armory access requirement, Misc category, single unathi NT breacher rig.

### DIFF
--- a/code/datums/supplypacks/misc_vr.dm
+++ b/code/datums/supplypacks/misc_vr.dm
@@ -107,7 +107,7 @@
 	contains = list(
 			/obj/item/weapon/rig/breacher = 1
 			)
-	cost = 300
+	cost = 250
 	containertype = /obj/structure/closet/crate/secure/gear
 	containername = "unathi breacher hardsuit crate"
 	access = access_armory
--- a/code/datums/supplypacks/misc_vr.dm
+++ b/code/datums/supplypacks/misc_vr.dm
@@ -102,6 +102,16 @@
 	containername = "Commonwealth engineering hardsuit crate"
 	access = access_engine
 
+/datum/supply_pack/misc/breacher_rig
+	name = "unathi breacher hardsuit (empty)"
+	contains = list(
+			/obj/item/weapon/rig/breacher = 1
+			)
+	cost = 300
+	containertype = /obj/structure/closet/crate/secure/gear
+	containername = "unathi breacher hardsuit crate"
+	access = access_armory
+
 /datum/supply_pack/misc/zero_rig
 	name = "null hardsuit (jets)"
 	contains = list(
--- a/code/modules/clothing/spacesuits/rig/suits/alien.dm
+++ b/code/modules/clothing/spacesuits/rig/suits/alien.dm
@@ -7,7 +7,7 @@
 	desc = "A cheap NT knock-off of an Unathi battle-rig. Looks like a fish, moves like a fish, steers like a cow."
 	suit_type = "\improper NT breacher"
 	icon_state = "breacher_rig_cheap"
-	armor = list(melee = 60, bullet = 60, laser = 60, energy = 60, bomb = 70, bio = 100, rad = 50)
+	armor = list(melee = 60, bullet = 45, laser = 35, energy = 35, bomb = 70, bio = 100, rad = 50)
 	emp_protection = -20
 	slowdown = 6
 	offline_slowdown = 10


BigBababooey:
Like a year ago I asked for it to potentially be able to be given an in game way to be obtained once it's erroneous mapping was removed since it was really fun to use even if eye gougingly slow, this seemed like the most reasonable way. It's cost is high even for a hardsuit and it's so damn slow it's wildly impractical, for the cost you really should just get one of the other hardsuits and be infinitely more mobile and still really well protected... but... well it's neat. And I see no harm in being able to order the NT version from NT for a large amount. Will it get use? Probably not! But if it does, that would be cool, options are cool.

_pleaseIwannaletmylizardlarp_

Novacat:
Keep in mind that the NT Breacher Rig provides 60% armor across the board, which is roughly twice as good as Hazard rigs (which is the Security hardsuit) and even better than ERT rigs.

Security rig:
armor = list(melee = 60, bullet = 40, laser = 30, energy = 15, bomb = 60, bio = 100, rad = 30)
	
ERT rig:
armor = list(melee = 60, bullet = 50, laser = 30,energy = 15, bomb = 30, bio = 100, rad = 100)

Unathi breacher rig:
armor = list(melee = 60, bullet = 60, laser = 60, energy = 60, bomb = 70, bio = 100, rad = 50)

Heroman3003:
Eh, its pretty expensive to order and with speed being the name of the game I don't think it's that strong, especially since its not authentic one (which IS actually OP despite having same downsides due to just how little damage you take with it). It is a powerful protection piece but it makes you move at speed of a dying turtle who is  taking its sweet time, which is an extreme tradeoff, which merc or ert equipment does not have, meaning rig itself is not even close to being on par with those.

Novacat:
The main issue is that while, on paper, the breacher rig is slow. Unathi have 75% item slowdown reduction so that 6 slowdown becomes a 1.5, which is not that bad (it adds up to 2 if you combine with their native slowdown). Comparatively speaking, they are not going to move that much slower than a normal human in a rig with 1 slowdown.

BigBababooey:
So, after some discussion and thoughts, I personally think seeing this some should be amended. I will say first and foremost that the slowdown point I think is a biiiit moot because when you're moving that slow you can no longer partake in the "meta" of combat, and need to fight in an entirely differently manner by that point. There's a reason Explo people always have haste/hardy and would never be caught dead with a default slowdown. And I can say confidently with experience back when the suit was mapped in, I didn't have a snowballs chance in hell of catching up, I forced them to slow down lol.

That said, I see no reason the stats of the breacher rig can't be brought back in line some if it's seen as a bit too good. There is definitely a point to be made that it's across the board solid, which isn't normally the case, and it is meant to be the shitty NT knockoff.

My proposal is as follows 

Unathi breacher rig:
armor = list(melee = 60, bullet = 45, laser = 35, energy = 35, bomb = 70, bio = 100, rad = 50)

And remove the species restriction. Of course, anyone without some sort of hardy on their character is going to be in hell, but if a human wants to roleplay incredible trouble using a suit fitted for Unathi bodies, I think that's funny, and I see no harm in it.

Heroman3003:
> The main issue is that while, on paper, the breacher rig is slow. Unathi have 75% item slowdown reduction so that 6 slowdown becomes a 1.5, which is not that bad (it adds up to 2 if you combine with their native slowdown). Comparatively speaking, they are not going to move that much slower than a normal human in a rig with 1 slowdown.

1 slowdown is a huge difference already that makes it impossible to kite mobs (at which point armor that high is pretty much necessary to not just instantly keel over). This is tanky gear. Also, comparing to sec gear, its only really better against ranged attacks, which are extreme minority in current explo ecosystem (with aerostat no longer hivebot infested and mercs not used anywhere).

Novacat (changes requested):
Unathi breacher rig:
armor = list(melee = 60, bullet = 45, laser = 35, energy = 35, bomb = 70, bio = 100, rad = 50)

And remove the species restriction. 

Yeah this would be fine.

Though looking at the numbers the Security/ERT rigs are hot garbage and should be buffed because there are station void suits with better defensive stats (at least against lasers/ballistics) but that is way out of scope.

Heroman3003:
I have no clue how removal of species restriction will work on voidsuit with only one species sprite in the first place, and non-human at that and I am no spriter to make change like that either.

Very-Soft:
Isn't this event equipment?

Why are we adding event equipment to cargo and talking about nerfing it. 

We also don't let normal station folk just go and order armored voidsuits so they can wear them around the station, so what makes you think that you should be allowed to do it? = w=

BigBababooey:
It's specifically armory access, it's not at all for being worn around normally. You can already order hardsuits for extreme circumstances, this would just be another hardsuit you could order for an extreme circumstance cause it's fringe but cool.

It's literally never going to be chosen to be given over anything else as Event equipment and never has, the only time it was used otherwise was when it was erroniously mapped in for a time.

Heroman3003:
> Isn't this event equipment?
> 
> Why are we adding event equipment to cargo and talking about nerfing it.
> 
> We also don't let normal station folk just go and order armored voidsuits so they can wear them around the station, so what makes you think that you should be allowed to do it? = w=

Regular breacher rig is definitely event-level gear, but afaik, NT version of it was created originally as specifically dumbed down, nerfed version that from the start was meant to be a more "actually balanced for station to use" kind of thing.


~~Also I am 99% certain that breacher rig, NT or regular, hasn't ever been used in an event without being explitly requested, and it was in EVA roundstart for a long time without ever really being 'abused' in an impactful way so imo that only goes to show that it is not really some OP gear nor something too special.~~

Heroman3003:
Adjusted the stats. I am not touching species resirictions since suit doesn't have sprites for any other species and I am in no skill to make them.

Cyantime:
The NT breacher was sitting in tether EVA storage for years, right next to the skrell voidsuits.

BigBababooey:
> Adjusted the stats. I am not touching species resirictions since suit doesn't have sprites for any other species and I am in no skill to make them.

admittedly I think it would be funny just to have it not adjust the sprite and very clearly not be actually fitted for other species. Don't some voidsuits use the same sprite for multiple different species? Is it not the same concept? _Genuine question I don't know many things_

InterroLouis:
But is it worth 300 points adjusted like this? Compare it to the industrial rig that costs 150 and has armor stats comparable to the adjusted version, but without the massive slowdown.

BigBababooey:
> But is it worth 300 points adjusted like this? Compare it to the industrial rig that costs 150 and has armor stats comparable to the adjusted version, but without the massive slowdown.

to be entirely truthful, no. It's a generally poor pick, but I'm not really of the belief everything needs to be a good pick, especially if this is going to stay species restricted. The commonwealth hardsuits are absurdly better (though perhaps too good) but for roleplaying as the party's tank for a big expedition that gets supplies ordered for it, nothing's better. And I think that's fun enough to have a niche harmless place on the cargo list.

I mean hell, the boomboxes that can be built by science for cheap are 300 for 2, not everything is well priced, and it makes sense that a more niche, limited production suit from NT, even if it's not very good, might be more costly.

Heroman3003:
I'm gonna drop points down to 250 to match other 'expensive' rig modules.

PR #13085, opened 2022-06-16
[MIRROR] Fixes recursive explosions

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/8645
--------------------
Recursive explosions finally updated to support brand spanking new features such as multi-z, and no longer uses global variables that limit the game into processing only one explosion at a time.

Aside from the tweaks, fixes a major bug where an explosion on open space turfs gets stuck in an infinite spreading loop, which locked up the loop break part there, and due to the vars having been global and the reset part being blocked by the loop break return, this would permanently break all explosions ingame.


Heroman3003:
*What are those diffs*

Cameron653:
This PR is FUBAR. Going to just make the actual PR myself.

Heroman3003:
> This PR is FUBAR. Going to just make the actual PR myself.

We dont even USE recursive explosions

Cameron653:
Yeah but we might as well have it fixed in case we ever decide to enable it/downstream wants to use it.

PR #13102, opened 2022-06-18
Tweaks biofuel processor trait

radiantAurora:
Increases nutrition multiplier for biofuel processer from 25% to 60%

**Why:**
As it stands now Biofuel Processor is unnecessarily bad. Yeah sure, FBPs can charge but they can _also_ take trash eater which, correct me if I'm wrong, provides 50% the nutrition from food compared to the 25% biofuel processor provides. So the intent of this tweak is to make biofuel processor worth using over vacuuming food like a glutton.
Also I want to be horny with foodplay scenes as a robot.

I'd also like to make a suggestion for someone who actually knows their way around SS13 code:
An 'obligate' version of biofuel processor that provides full nutrition from food while disabling the use of cyborg chargers.

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/traits_vr/neutral.dm
+++ b/code/modules/mob/living/carbon/human/species/station/traits_vr/neutral.dm
@@ -164,7 +164,7 @@
 	cost = 0
 	custom_only = FALSE
 	can_take = SYNTHETICS
-	var_changes = list("organic_food_coeff" = 0.6, "synthetic_food_coeff" = 0.9)
+	var_changes = list("organic_food_coeff" = 0, "synthetic_food_coeff" = 0.6)
 
 /datum/trait/neutral/glowing_eyes
 	name = "Glowing Eyes"
--- a/code/modules/mob/living/carbon/human/species/station/traits_vr/neutral.dm
+++ b/code/modules/mob/living/carbon/human/species/station/traits_vr/neutral.dm
@@ -164,7 +164,7 @@
 	cost = 0
 	custom_only = FALSE
 	can_take = SYNTHETICS
-	var_changes = list("organic_food_coeff" = 0, "synthetic_food_coeff" = 0.25)
+	var_changes = list("organic_food_coeff" = 0.6, "synthetic_food_coeff" = 0.9)
 
 /datum/trait/neutral/glowing_eyes
 	name = "Glowing Eyes"


Novacat:
I recommend double checking the trash eater values because my hunch is that they are a lot lower than 50%.

radiantAurora:
I seem to remember them being upped to 50% at some point, though I don't know where they'd be in the code so I can't really check.

lm40:
It is 50% for organics, i.e. 50% of the reagents of a food item will be transferred into an organic who trash ate something. I think 50% of the reagents would transfer for a synthetic too, but I doubt that matters since base synthetic with only trash eater would derive nothing from metabolizing nutriment.

lm40:
ALSO while there is a possibility that organic_food_coeff = 0 implies certain things wrongly for biofuel processing synths, if the value is not going to be zero it should probably be equal to synthetic_food_coeff.

radiantAurora:
> base synthetic with only trash eater would derive nothing from metabolizing nutriment.

I'm not sure why but this is actually wrong. Using trash eater will still derive nutrition even as a synth, with or without biofuel processor. I've tested this before but, as I've said before, I don't know where the code for trash eater nutrition gain is.

> if the value is not going to be zero it should probably be equal to synthetic_food_coeff.

As for this I disagree, I think the trait works better if synthetic foods provide more nutrition for synthetic bodies, even if they can still derive _some_ from organic foods.

lm40:
> I'm not sure why but this is actually wrong. Using trash eater will still derive nutrition even as a synth, with or without biofuel processor. I've tested this before but, as I've said before, I don't know where the code for trash eater nutrition gain is.

I'll check this, but it's more likely that they are gaining nutrition from trash-eaten non-food items which just provide a straight charge. Food items transfer their reagents which like I said should really do nothing. Or maybe when the food item is deleted it provides an element of straight charge or something, idk exactly.

> As for this I disagree, I think the trait works better if synthetic foods provide more nutrition for synthetic bodies, even if they can still derive _some_ from organic foods.

There's no such thing as synthetic foods, synthetic_food_coeff just determines how much synthetics receive from nutriment, which is why it would be inconsistent/weird to have it different from organic_food_coeff.

radiantAurora:
> There's no such thing as synthetic foods

Given the description of the trait I'd have thought this was the case, specifically:
> Energy-dense foods such as protein bars and survival food will yield the best results.

If it's not the case I'll go and match the coefficients

lm40:
Overall tho, trash eater should already be worse than biofuel processor from the perspective of eating food because it doesn't take into account the nutrition of items. You just get the flat value whereas with biofuel processor you get the flat value + 1/8 of the nutritional value of the item. Or if you just eat the item normally with only biofuel processor you get 1/4 the nutritional value of the item. So this doesn't really seem too necessary to me but yeah.

KillianKirilenko:
> > There's no such thing as synthetic foods
> 
> Given the description of the trait I'd have thought this was the case, specifically:
> 
> > Energy-dense foods such as protein bars and survival food will yield the best results.
> 
> If it's not the case I'll go and adjust the numbers to reflect that.

Very very belatedly (since I noticed the processor was buffed and wanted to find the PR which did it) this is probably my fault for naming the vars a little unintuitively.

How it works is that the organic coeff multiplies how much nutrition non-fbps get from food, and synthetic coeff multiplies how much nutrition fbps get from food. I forget why I set it up this way specifically, but it is what it is.

I did mull over making 'cyborg food' (which would be very efficient for synths, but bland and unpleasant for regular people) in the past but never really got around to making it or figuring out how to distribute it fairly, and it's become a bit moot with the perks just being buffed outright.

> An 'obligate' version of biofuel processor that provides full nutrition from food while disabling the use of cyborg chargers.

I've got a couple of ideas on how to do this fairly easily.

Issue #13118, opened 2022-06-19
Unable to use the 'Adjust-Mass' Promethean verb


CrowbarLamb:
#### Brief description of the bug
Promethean 'Adjust-Mass' verb does not work

#### What you expected to happen
Pressing the button would open up a window that gave you a 25-200% size choice

#### What actually happened
Nothing happens when you press the button

#### Steps to reproduce
1. Be a Promethean
2. Press 'Adjust-Mass' in the Abilities tab
3. Despair as you are unable to change your size

PR #13121, opened 2022-06-19
Adjust Mass and Hide fixes

Very-Soft:
Fixes adjust mass's input, and tweaks the hide verb so it tells you when you stop hiding (also makes it disable eye glow while hiding for pAIs)


ItsSelis:
Fixes #13118

PR #13126, opened 2022-06-19
Encoding Fixes

ItsSelis:
It would be most beneficial if someone else could also test some of the input.

This PR will disable the encoding by default for TGUI Input Boxes.
All text sanitation already encodes the html.

Static Data Refreshing is also disabled (the R&D Console has to be fixed up a bit for that to work properly without interrupting the scientists)


Very-Soft:
Me tested, seems okay now

PR #13164, opened 2022-06-25
[TGUI] Added prettierx

ItsSelis:
This replaces some of the eslint checking. It pretty much makes the code look a bit prettier without doing much (Saving the file).

Allows the developers to actually code the stuff instead of dealing weird syntax formatting.

The build script includes the prettier check and there are no functional changes to it, really just rather asthetics.


Razgriz1032:
Trying to rebuild the TGUI bundle (on another repo_ says "run Prettier." How do we do that for the bundle?

PR #13218, opened 2022-06-30
Ports some modular computer stuff from baystation

ItsSelis:
This really just includes mostly variables and procs that will be helpful on future ports, not having any major/breaking changes while doing so. (Also some comment cleanup)

Also enables the broadcast E-Mail Address (broadcast@virgo.local) to relay a message sent to everyone.


Very-Soft (changes requested):
Conflicts

PR #13244, opened 2022-07-02
Multi-Map Testing

ItsSelis:
The integration test workflow will now compile all 3 of the main maps upon creation/update of an PR (like the currently active map would have been tested, but in this case, all do).

For this to work, the stellardelight folder had to be renamed to stellar_delight, due to the other map .dm files having the exact same name as the folder itself.


Heroman3003:
Looks like remaining issues are on engine submaps

ItsSelis:
> Looks like remaining issues are on engine submaps

Yeah, that one is fixed too. I wanted to fix the wire tests a bit, doing the last check. If it doesnt work the two areas of the shuttles have to be exempted.

PR #13275, opened 2022-07-04
Fixes ashburndracki's whitelist entry

Heroman3003:
there are no underscores in whitelist files


faaaay:
hey gamer, cool PR, might I suggest we add a comment to the file saying there are no underscores?

PR #13277, opened 2022-07-04
Adds new alt titles for some jobs

Aroliacue:
Adds the following alt titles - Search and Rescue to Paramedics, Security Contractor to Security Officers and Talon Marine to Talon Guards.


Very-Soft:
Would prefer not to have 'talon marine' as it sounds much too much like a military/soldier type title, and I really don't want people thinking that's what the role is for

PR #13290, opened 2022-07-06
[MIRROR] Adds some core behavior code, uses it to fix a small bug (ABLE TO BE MERGED) 

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/8660
--------------------
Three commits for clarity.

a1dcc55be9f6edb91012f681cd79f42658209217 adds the abstract_type variable and IsAbstract method to /datum, and the global function is_abstract.
Abstractness signals whether something should be instantiated or not, and is useful for things like base behavior types and templates. Because there is no such language concept in DM, it's up to code authors to actually use it.
Refs:
<https://github.com/Baystation12/Baystation12/pull/32114>
<https://github.com/Baystation12/Baystation12/pull/32157>

e2b345a023fd3beca39004722c5dda913761f0da simply sets abstract_type for /datum/trait and template subtypes, and uses is_abstract to prevent their instantiation. This means the base "Physical" and "Mental" trait no longer show up in character setup.

b67fe40cc08e245c5bf126929e0524964a95dcba adds transform member variables to /atom and /image, and the ClearTransform and SetTransform members. These can be used to update atom and image transform fields in a generally cleaner way than directly setting or changing the field.
It also adds the Clear and Update methods to /matrix, which perform the underlying behaviors and allow a nicer pattern for matrix creation than multi-lining: `matrix().Update(...scale, rotation, etc...)`
Refs:
<https://github.com/Baystation12/Baystation12/pull/32067>

I did not actually update the codebase to make use of the matrix and transform handlers in this PR because I'm a little hung over this morning and there's a good few hundred matches to look over, but the thought's there.


Cameron653:
Going to hold on merging this until further testing is done.

Cameron653:
Hey admemes
This adds a new 'SetTransform' variable that can be called on **ANY ATOM**. Yes, you read that right.

# ANY ATOM

How to use:

/atom/proc/SetTransform(
	scale,
	scale_x = tf_scale_x,
	scale_y = tf_scale_y,
	rotation = tf_rotation,
	offset_x = tf_offset_x,
	offset_y = tf_offset_y,

Do an Atom-ProcCall on anything you want to change.
Call the proc **SetTransform** (CASE SENSITIVE!)
If you want to just adjust scale (both X and Y to be the same) set arguments to 1 and set it as a number to your liking.
If you want to adjust more, set arguments anywhere from 2-6 (2 will only let you adjust X, 6 will let you adjust X, Y, rotation, and X and Y offset.)
Set the first argument to NULL and set the others to your liking.
Example: SetTransform(NULL ,2.5, 3.75, 45, 10, 10)
This will make someone that is 2.5 wide, 3.75 tall, be turned 45 degrees, and have an X and Y offset of 10.

In reality they will look like this: 
dreamseeker_2022-09-24_21-20-53

Have fun. Some examples of fun that was had is listed below.


dreamseeker_2022-09-24_21-04-16

Cameron653:
As a side note this was also checked with Tall and Tall still functions. Forgot to mention that.
dreamseeker_2022-09-24_21-27-27

PR #13297, opened 2022-07-08
Fixes a typo for the random mouseray spawner.

Cameron653:
Typo had it as maintoyloot which overwrote the actual maintoyloot random spawner.

Presumably it was supposed to be mouseray, judging by what was above it.

This hotfixes it.


Very-Soft:
good catch, thanks!

PR #13310, opened 2022-07-09
Beginning of the Vore Export

ItsSelis:
This PR will add a dedicated export panel for all vore related data.

Currently it only will support exporting all bellies of the mob that is currently being played.
One can either choose to export it to a formatted HTML file, or export it to a so-called "Vore Database" (VRDB) file.

.vrdb files will be able to be opened/edited with an tool called the VRDBEdtior (Vore Database Editor), which is still currently under development.

Here are very early development screenshots (not the final product):
grafik
grafik
grafik


Very-Soft:
This is a big thing I do not feel totally confident in how it works, so would like other people to review it.

Further: the prospect of you putting a panel into the game which people can upload their own files to the server is something I'm deffo not touching with a 10 foot pole.

I PERSONALLY don't really get the whole 'let's export vore bellies' thing, since you can only really use them in places that adopt the same system, and is almost exclusively useful for people who don't want to play here. And the whole, making a special filetype that can be used for it seems like it would require users to be able to upload files to the server, and I really don't think that will go well. 

So, while I appreciate the work on this, I doooon't really think this is something I'm so much inclined or knowledgeable enough to handle.

ItsSelis:
> This is a big thing I do not feel totally confident in how it works, so would like other people to review it.
> 
> Further: the prospect of you putting a panel into the game which people can upload their own files to the server is something I'm deffo not touching with a 10 foot pole.
> 
> I PERSONALLY don't really get the whole 'let's export vore bellies' thing, since you can only really use them in places that adopt the same system, and is almost exclusively useful for people who don't want to play here. And the whole, making a special filetype that can be used for it seems like it would require users to be able to upload files to the server to actually be useful, and I really don't think that will go well.
>
> So, while I appreciate the work on this, I doooon't really think this is something I'm so much inclined or knowledgeable enough to handle.

The main reason of the exporting is to backup all the descriptions of the bellies themselves in one click instead of going to each belly, press every single button and copy them out one by one. This is pretty much how the X-Print to Chat thing works, pretty much just dumping all of the data into a formatted html, since the chat thing lags out like hell (even just on my development instance with some default bellies) if pressed (and even disappearing into the chat history depending of how many bellies are saved).

And the files are not uploaded to the server. The files are handled by the clients TGUI Bundle for the import and export part. Though, that stuff was just supposed to be a fun tool to just edit the bellies outside of the game with an editor or to copy paste the bellies between characters.

PR #13354, opened 2022-07-17
Allows Entertainers to have custom job titles

Heroman3003:
Chaplains get it, and Entertainers make a lot of sense to have it too, IMO.

### DIFF
--- a/code/game/jobs/job/civilian_vr.dm
+++ b/code/game/jobs/job/civilian_vr.dm
@@ -254,6 +254,37 @@
 						"Magician" = /datum/alt_title/magician, "Comedian" = /datum/alt_title/comedian, "Tragedian" = /datum/alt_title/tragedian,
 						"Artist" = /datum/alt_title/artist)
 
+/datum/job/entertainer/equip(var/mob/living/carbon/human/H, var/alt_title, var/ask_questions = TRUE)
+	. = ..()
+	if(!.)
+		return
+	if(!ask_questions)
+		return
+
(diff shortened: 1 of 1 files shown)


Very-Soft:
I'm much less okay with this for the simple matter of numbers. 

There's only one chaplain. It's not unreasonable to me to have one funky religion man running around with a funky job title.

Entertainers on the other hand, there's able to be up to four of them, and while it's easy to see what they actually are through AR, I feel like it would muddle things up much too much, and have too much potential to be memey, and push the line of what's acceptable in regards to these things.

Me do not like the idea of folks just setting your jobs to whatever, and I feel like opening it up to what essentially amounts to an alternative assistant role would evolve the question over time to 'well why can't THIS role pick their own title, why can't THAT role pick their own, why can't ALL the roles just name themselves whatever'. Like I can already see someone taking this and saying 'what about clowns and mimes and interns'.

I don't think that's a road we should go down. I'm okay with it being so on the chaplain, because there's 1 chaplain, and it makes good sense that the representative of funny space religions would have funny titles. I'm fine with them having that. I don't think it should be pushed further than that though.

lm40:
Yea I'm sort of iffy on it for that reason too, I wouldn't really mind clowns and mimes being able to do it actually because they're whitelist and uncommon but entertainer has no real role and is basically a visitor alt-title. While Chaplain is not really an important job, it at least is a distinct role, so I think it suffers slightly less from people taking extreme liberties with the custom title.

PR #13362, opened 2022-07-19
Change interpersonal belly transfers to a seperate option

Very-Soft:
Now instead of 'Move' asking you 'your belly or someone else's', there's 'Move', which moves things between your bellies locally, and 'Transfer', which moves things to people-who-are-near-you's bellies

### DIFF
--- a/code/modules/vore/eating/vorepanel_vr.dm
+++ b/code/modules/vore/eating/vorepanel_vr.dm
@@ -611,7 +611,7 @@
 	var/atom/movable/target = locate(params["pick"])
 	if(!(target in host.vore_selected))
 		return TRUE // Not in our X anymore, update UI
-	var/list/available_options = list("Examine", "Eject", "Move")
+	var/list/available_options = list("Examine", "Eject", "Move", "Transfer")
 	if(ishuman(target))
 		available_options += "Transform"
 	intent = tgui_alert(user, "What would you like to do with [target]?", "Vore Pick", available_options, strict_byond = TRUE)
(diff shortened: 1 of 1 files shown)


code/modules/vore/eating/vorepanel_vr.dm
```diff
@@ -611,7 +611,7 @@
 	var/atom/movable/target = locate(params["pick"])
 	if(!(target in host.vore_selected))
 		return TRUE // Not in our X anymore, update UI
-	var/list/available_options = list("Examine", "Eject", "Move")
+	var/list/available_options = list("Examine", "Eject", "Move", "Transfer")
 	if(ishuman(target))
 		available_options += "Transform"
 	intent = tgui_alert(user, "What would you like to do with [target]?", "Vore Pick", available_options, strict_byond = TRUE)
```
ItsSelis, line 617: A suggestion I'd have is using tgui_input_list instead of the alert one (which should actually swap to a list automatically though that was temporarily disabled), due to the list of buttons getting pretty long and sometimes overlapping stuff or stuff.

Otherwise it all looks fine! :D 

```suggestion
	intent = tgui_input_list(user, "What would you like to do with [target]?", "Vore Pick", available_options, strict_byond = TRUE)
```

ItsSelis (approved):

ItsSelis:
Oh right, I forgot that the tgui_input_list has no strict_byond variable in it's proc. Removing that part should be totally fine (that was some debug thing in the past which wasn't removed it seems a)

PR #13365, opened 2022-07-19
Adds pride pins to vending machines (Looty Inc.)

Runa-Dacino:
Dirty web-edit addition of pride pins to be obtainable in-round.


Very-Soft:
WEB EDITER!!!!!

PR #13385, opened 2022-07-23
Adjusts Space Ghost

Heroman3003:
Reduces life time of ghostling from 2 minutes to 35 seconds. This should reduce maximum amount active at any time from 17 to 5. And hopefully prevent lag vortexes from being nearly as bad if they happen to be created.

As additional measure, fixes factions of things in the whale to actually be universally matching and prevent mob infighting.


Novacat (approved):

Novacat:
Hopefully this might help fix the mob subsystem crashes, but I want Soft to take a look before this is merged.

PR #13387, opened 2022-07-24
Allows attacks to properly damage simple_mobs.

Cameron653:
Actually way easier than I thought it'd be. Took me longer to actually find where all the damnable code was at then it took to code it.
- Fixes a bug which makes unarmed attacks do burn damage to mobs.
- Makes it so when mobs are attacked by an unarmed human, they take damage properly, much like normal unarmed attacks. 

This means that the damage one deals while unarmed is no longer hardcoded to whatever the person making the mob mistakenly put the "harm_intent_damage" as thinking it was for something else.

For reference:
Normal unarmed attacks deal 1d5 damage + the damage the unarmed_attack itself has. (Bloodletters has 15 for example.)

The code confirms the person attacking is a human, checks what unarmed attack they have selected (or if they have bloodletters selected/on), rolls a random dice between 1 and 5 for "damage" then adds the bonus damage certain unarmed_attacks have.

It also checks to see what type of damage that unarmed_attack should do, applies it (Brute?Burn?Tox?Oxy?Brain? Etc) and gives it a proper attack verb instead of just whatever the creator of the mob you're attacking decided it should be.

Photos:
Attacking w/ bloodletters: https://i.imgur.com/BzbmZGj.gif
Using various attacks: https://i.imgur.com/QdMtMV8.png

### DIFF
--- a/code/modules/mob/living/simple_mob/defense.dm
+++ b/code/modules/mob/living/simple_mob/defense.dm
@@ -45,7 +45,18 @@
 
 		if(I_HURT)
 			var/armor = run_armor_check(def_zone = null, attack_flag = "melee")
-			apply_damage(damage = harm_intent_damage, damagetype = BURN, def_zone = null, blocked = armor, blocked = resistance, used_weapon = null, sharp = FALSE, edge = FALSE)
+			if(istype(L,/mob/living/carbon/human)) //VOREStation EDIT START Is it a human?
+				var/mob/living/carbon/human/attacker = L //We are a human!
+				var/datum/unarmed_attack/attack = attacker.get_unarmed_attack(src, BP_TORSO) //What attack are we using? Also, just default to attacking the chest.
+				var/rand_damage = rand(1, 5) //Like normal human attacks, let's randomize the damage...
+				var/real_damage = rand_damage //Let's go ahead and start calculating our damage.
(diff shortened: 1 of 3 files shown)


Cyantime:
What of things like simplemob mechs, which are using it properly to receive zero damage from fists with the "uselessly hits" custom text.

Cameron653:
For simple_mob mechs and the such, I would suggest they have the attack_hand proc manually changed in their appropriate files to make it say no damage. However that would more than likely be a bad idea.
I'm unsure of how many mobs have functionality like that, but it should be an edge case, especially given the fact that unarmed attacks aren't only low damaging attacks and can have up to 20 damage a melee hit, which would most definitely damage a mech.

Also yes, the harm intent flavor text being removed is intentional. 
It's still used for anything else that attacks them w/ the exception of humans. It doesn't exactly makes sense to "stomp on" something when you're using bloodletters or biting it.

Screemonster:
for something like simplemob mechs I'd rather just give simplemobs a damage threshold var or something that makes it ignore attacks that do less than that damage. 
note: I don't mean _reducing_ the damage by that amount - just a straight "if damage < threshold, return" check that does full damage if it pens.

Cameron653:
Let me go ahead and give them a damage threshold. Shouldn't be too difficult to do.

Cameron653:
W/o bloodletters (normal attack: https://i.imgur.com/fiC8zQQ.png)

With bloodletters: https://i.imgur.com/twbuJUL.png

PR #13393, opened 2022-07-25
[MIRROR] deco stuff

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/8667
--------------------
Ports a selection of decoration stuff from TG and Bay
image
(the black billboard is animated so that's why the stuff on it is faded. It fades between these)
image
image

Some billboards for POI use
A selection of new posters
Some new signs
A wall clock and a calendar
Some desk clutter, most of which can be toggled by clicking it inhand
Map edits were made to place some of these things in, this is my first PR editing maps so hopefully it's okay.


Cameron653:
This has too many conflicts. If we want this, it can be manually ported.

PR #13395, opened 2022-07-25
Ghost spawning as mice from trash piles

Fracshun:
Added the ability for observers to click on trash piles to trigger the mouse spawn prompt. Doing so spawns them on the pile with the same "crawls out of the trash" message a normal mouse would trigger.


Very-Soft:
why tho? There's a verb for spawning as mice already.

Very-Soft (changes requested):
Also you've donked up the nanomaps a bit, please remove those

Fracshun:
> why tho? There's a verb for spawning as mice already.

The verb random spawns you in the vents. Which are a mess at the best of times, and can lead to getting stuck in departments if you don't know your way around. This gives the option to spawn more freely in maintenance areas without the maze minigame.

PR #13398, opened 2022-07-25
adds "selective" bellymode for mobs

Screemonster:
Adds a new bellymode, _currently_ available only by varediting the belly's digest_mode var, intended for use on mobs.

The mode checks the prey's digestable and absorbable prefs and will attempt to digest, absorb, and drain in that order depending on what the prey allows.
Belly emote messages will _also_ display those for digesting, absorbing, or holding respectively.
Vore panel description is fairly generic because doing anything remotely complicated with TGUI beyond "copypasting a new entry into a list" is beyond my expertise.


Heroman3003:
Looks like tgui bundle wasn't actually recompiled, but beyond that PR seems fine. Also its okay if it wasn't as long as someone does it between merge of this and release.

PR #13399, opened 2022-07-26
Expands Server Status Info

Poojawa:
As per a long standing request on the discord, the information has been expanded.

Unable to personally test because I do not have access to the bot, but this is basically a 1:1 copy of checking admin lists so it should be fine. 

No names, no ckeys are revealed in these counts.

### DIFF
--- a/code/modules/tgs/v5/chat_commands.dm
+++ b/code/modules/tgs/v5/chat_commands.dm
@@ -4,7 +4,23 @@
 	admin_only = FALSE
 
 /datum/tgs_chat_command/status/Run(datum/tgs_chat_user/sender, params)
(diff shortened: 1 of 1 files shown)


code/modules/tgs/v5/chat_commands.dm
```diff
@@ -4,7 +4,23 @@
...
+	var/list/counts = 0
+	var/list/afks = 0
+	var/list/active = 0
+	var/list/bellied = 0
+
+	for(var/X in TGS_CLIENT_COUNT)
+		var/client/C = X
+		counts++
+		if(C && !(istype(C.mob,/mob/new_player) || istype(C.mob, /mob/observer)))
+			if(C && C.mob && isbelly(C.mob.loc))
```
Poojawa, line 16: You can never be too paranoid with byond. Don't worry about it.

PR #13405, opened 2022-07-28
Engineering & Medical Contractor alt-titles

BonniePandora:
Throws in both "Engineering Contractor" & "Medical Contractor" as new alt-titles for station engineer & doctor respectively, with small blurbs for the both.

Quick & dirty webedit, if I screwed something up I'll sort it later on


lm40:
I don't really care too much either way here but it seems a little odd to me. Maybe it's just the kinds of jobs I've worked but I've never known someone who was a contractor to have 'contractor' in their actual job title regardless of their technical status.

PR #13415, opened 2022-07-29
Slightly modifies Etheo's custom dogborg sprite to give the borg less booty.

Cameron653:


EtheoBoxxman:
The borg butt was too op, had to get it nerfed.

PR #13426, opened 2022-07-31
refactors feralness

Screemonster:
Xenochimera now remember how stressed they are from moment to moment, and only check to go feral when this value increases.

In short, if you get hit, you either go feral immediately or you don't at all.

Feralness values from damage should cap at much lower levels than before (important since this now gives them an attack buff)


Screemonster:
noticed during testing that mob attacks don't call apply_effects() on unarmed attacks, which meant the bonus damage wasn't being applied. So, screw it.
Yes, this means you apply your feral rage in full if you misclick and claw yourself. Don't claw yourself.

PR #13436, opened 2022-08-01
adds cool and fun and totally balanced new weapon

Screemonster:
I'll be honest, I mainly added this because I wanted to see if it was possible.

Yes, this thing does consume an entire weapon cell in a single shot for a bloody good reason. That and the wind-up delay make it fairly impractical to actually use but zoweeeeeemama are the results fun.

image

You know exactly what this thing is.


Screemonster:
soft made a cool new charging up sound for this that works perfectly so that's going in too

Cameron653:
This is a perfectly balanced weapon as far as I can see and would be perfect to be set in the lockers of every security officer.

It only does 25 agony per taser strike after all!

Has my support for server addition.

PR #13454, opened 2022-08-05
Expands mining&archeology belt storage.

Cameron653:
- Allows mining to carry items they get from the mining vendor.
- Allows mining to carry screwdrivers and multitools.
- Allows mining to carry binoculars
- Allows mining to carry ore scanners.
- Allows mining to carry radios.
- Allows xenoarch belts to carry multitools.
- Allows xenoarch to carry pickaxes on their belt.
- Allows xenoarch to carry marker beacons.
- Allows xenoarch to carry geiger counters (radiation is like...Part of their job.)
- Gets rid of needless /pickaxe duplication in the mining list.
- General cleanup. Sorts everything to look neater.


Cameron653:
If anything else should be added to either the mining belt or xenoarch belt, I am all ears. I doubt this is everything that _should_ be in there, but this expands the list enough that it should carry everything that most people want to be in it.

Some of the common complaints was being unable to carry tools and the ore scanner in mining belts. This addresses that and then some.

Also, I did not add the entire path of weapon/melee to the list. That was there before I even touched the file, I just rearranged the file and added onto it with things that would make sense (PK weapons for example.)

Heroman3003:
Idk, this kind of feels like it completely invalidates point of general 'utility' belt that carries actual tools by making both xenoarch and mining belts being able to carry nearly everything.

Cameron653:
I could reign it in some toolwise. 
Mining uses crowbars, screwdrivers, and wrenches (for the mining drill, although wrench is the most commonly used one) and occasionally multitools (for the abandoned crates and giving the drills an ID, but I don't think anyone even uses the ID feature on mining drills or even knows about it.) 

I could easily limit it to them being able to only hold those four and limit the xenoarch tools to being able to hold wrenches&multitools, which are the two tools commonly used.

I was more or less bringing it in line with how I saw the other belt in the file (exploration) belt being able to hold every /tool, while the utility belt seems to not be able to do that.

Cameron653:
Made it so mining can only carry wrenches, screwdrivers, and multitools and xenoarch can only wrenches and multitools.

Heroman3003:
Exploration belt itself is pinnacle of "explo is department that does every job ever" issue anyway, so its not a Good Example to follow imo.

Cameron653:
Yeah, tbh the exploration belt holds pretty much literally _anything_ you can think of looking at it closer.

But anyways, now they shouldn't encroach on the utility belt. Utility belt will still be the go-to for tools.

PR #13456, opened 2022-08-06
Adjusts dogborg modules to be inline with regular modules

TankTheBirb:
tl;dr these things were designed with less functionality than regular silicons, but that didn't apply to every module. I have no idea why, I am assuming oversights. I have fixed the oversights.

- K9, Kmine, Service-hound, Janihound, and Boozehound are unaffected by this. I'll come back to those at a later date if I recognize any major capability gaps.
- Pupdozer has a less glacial material synth, now only 1/4 as fast as the base model instead of 1/20th the speed.
- Sci-Borg now has their sub-department tools, as well as jumper cables. This means they can save the xeno-bio's work instead of killing it, and do xeno-arch (or at least assist a miner who can break rock for you while you do it). Also un-dying synthmorphs. 
- Medihound now has splints, the ability to interact with beakers and stands by having a gripper, a larger internal beaker, and a dropper. This puts them roughly on par with the crisis module now. They can use 1 patch to 5 splints as the medicine ratio at the moment, I may add brute/burn specific patches from the crisis module and then also add the liquid patch to the crisis module.

TODO FOR MAYBE A LATER PR
- Make a surgeryhound (sprite may be easier than other stuff)
- Address carbon/silicon and silicon/silicon capability disparities in the five unaffected modules (some exist, mostly minor)
- make it so borgs (all relevant ones) can do NIF implant surgery

### DIFF
--- a/code/modules/mob/living/silicon/robot/robot_modules/station_vr.dm
+++ b/code/modules/mob/living/silicon/robot/robot_modules/station_vr.dm
@@ -270,20 +270,27 @@
 	src.modules += new /obj/item/device/dogborg/boop_module(src) //Boop the crew.
 	src.modules += new /obj/item/device/healthanalyzer(src) // See who's hurt specificially.
 	src.modules += new /obj/item/weapon/reagent_containers/syringe(src) //In case the chemist is nice!
-	src.modules += new /obj/item/weapon/reagent_containers/glass/beaker(src)//For holding the chemicals when the chemist is nice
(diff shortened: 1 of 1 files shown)


Heroman3003:
Janihound's drawback compared to normal borg is the need to use water reserves in order to keep walk-cleaning and walk-cleaning in general being a module rather than inheirent feature.

User-U-U:
Been working on a surgery hound actually going to make a PR soon

PR #13458, opened 2022-08-07
Allows changing of the release verb when expelling stomach contents.

Cameron653:
- Allows for people to change their 'expel verb' much like their ingest verb.
- Example: 
image


SpadesNeil:
Psssst.

Add belching sound effects to the release sounds. :>

Cameron653:
Perhaps. In another pr though because I do NOT want to touch tgui again and
break this

PR #13467, opened 2022-08-08
[MIRROR] conk rete (ABLE TO BE MERGED) 

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/8671
--------------------
Adds concrete as a material, somewhat ported from bay/nebula (at least, in terms of the colour value and the texture overlay for it), obtainable via cargo or the stacks found in engineering, or by crafting it with 2 water, 2 silicate and 2 calcium per brick.
Cynosure and a number of POIs have been edited to have concrete walls, except in some special locations such as the toxins burn chamber and the supermatter core.
There is also a new type of flooring for mapping and POI use.


Cameron653:
[2022-09-21T06:36:27] Runtime in reader.dm,370: list index out of bounds
  proc name: parse grid (/dmm_suite/proc/parse_grid)

On the tether...Strange.

Cameron653:
There were a few submap changes made to incorporate concrete, but it caused a plethora of map problems.

If concrete is desired on the submaps it can be manually added in another PR.

PR #13471, opened 2022-08-08
Makes wounds disappear

lm40:
Have you ever noticed that if you get scraped once sometimes it literally never goes away throughout the entire round? Even when all the damage is totally healed? Sometimes external organs don't process to save resources, and as a result (it seems) to almost never if ever properly remove healed wounds from external organs.

HOWEVER, THIS FIX WILL CAUSE EXTERNAL ORGANS THAT BASICALLY DON'T /REALLY/ NEED TO BE PROCESSED TO SIT IN THE PROCESSING QUEUE FOR ~10 minutes (the time wounds take to heal), so there's a nonzero chance this could cause a performance decrease, so I also consider removing that ten minute timer entirely.

### DIFF
--- a/code/modules/organs/organ_external.dm
+++ b/code/modules/organs/organ_external.dm
@@ -732,8 +732,8 @@ Note that amputating the affected organ does in fact remove the infection from t
 		return
 
 	for(var/datum/wound/W in wounds)
-		// wounds can disappear after 10 minutes at the earliest
-		if(W.damage <= 0 && W.created + 10 * 10 * 60 <= world.time)
+		// wounds used to be able to disappear after 10 minutes at the earliest, for now just remove them as soon as there is no damage
+		if(W.damage <= 0)
 			wounds -= W
 			continue
 			// let the GC handle the deletion of the wound
--- a/code/modules/organs/organ_external.dm
+++ b/code/modules/organs/organ_external.dm
@@ -594,6 +594,8 @@ This function completely restores a damaged organ to perfect condition.
 		return 1
 	else
 		last_dam = brute_dam + burn_dam
+	if (number_wounds != 0)
+		return 1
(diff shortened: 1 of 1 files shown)


Cameron653:
Sounds good to me. Provided this was tested and works as expected, I 100% support this. The wounds not disappearing is a clear bug, especially low level wounds.

The only woulds that really make sense to persist are the ones that result from _MASSIVE_ amounts of damage, aka, the final-stage wounds. I don't think the game differentiates this when it comes to wounds disappearing, however, and it's such an edge case that it should be addressed in another PR.

lm40:
Alright, when I get the chance I'll just go in and remove it, I do sort of agree that there should be a possibility for a wound to hang around visually if it was something massive (like 150 brute to chest or w/e, you might heal the damage but still a message is fine), but I'm pretty sure the wound system currently doesn't distinguish much if at all between '150 brute wound that healed over time down to 5 brute' and '5 brute wound from punching yourself'.

Cameron653 (approved):
Looks good to me. We just need to make sure to remember to add a variable timer depending on the wound severity in a later PR.

Novacat:
I am slightly worried about the performance implication BUT there honestly should not be much unless we are doing a very combat heavy round where there is going to be 10-20 people injured at once.

lm40:
Wounds already process up until they are healed and then do basically one more process for good measure before their organ gets kicked from the queue (which is probably supposed to delete them but almost never did because of the ten minute timer; wound would be at 0 but not eligible for deletion and the organ would be out of the queue so it would never process again). Honestly removing the timer alone /should/ have fixed this issue but this is more explicit. To be honest I dunno how expensive organ processing even is, but the fact that someone made a whole selective organ processing system implies that it is can be at least somewhat costly.

PR #13474, opened 2022-08-09
Optimizes shower's turf washing a bit

Heroman3003:
One day, someone should rewrite wet_turf proc to not have long-ass sleep() calls.


Cameron653:
I had no idea reagents spilled onto the tile when splashed onto something on them. Wow. Now I realize why I couldn't figure out why the tile kept being hit with water.

Cameron653 (approved):

PR #13477, opened 2022-08-11
Adds Loopty Loo to Protean whitelist

Heroman3003:
Reportedly approved via modmail


Very-Soft:
https://forum.vore-station.net/viewtopic.php?f=46&t=2165

PR #13481, opened 2022-08-11
Medsensor borgs can analyze patients in their belly.

Cameron653:
Exactly what it says on the tin. Lets any dogborgs that have the medsensor variable analyze patients inside of their borg stomach.

Also fixes a bug that was causing runtimes that was found while I was coding this. If dogborgs didn't have a certain material for one reason or another and digested something that added to the material, it'd runtime due to it being null. This adds a check to make sure it actually has the materials before adding charges to them.


Cameron653:
Ran into some runtimes (https://i.imgur.com/qZdZGRq.png) and slapped in a quick fix for them as well.

Wickedtemp:
Can this be added to healbelly mode as well? Or the ability to use health analyzers on mobs in healbelly?

Cameron653:
I'd rather do that in a separate PR, as that would involve touching the actual stomach vore-code as opposed to a borg module, which is going to be _just_ a tad bit more difficult.

I do like the idea, however.

PR #13485, opened 2022-08-12
Condenses all the animal languages and deletes unused/unwanted languages

Cameron653:
- Condenses all animal languages into a general "animal" language
- Removes a few unused/low-used languages


code/modules/mob/language/monkey.dm
```diff
@@ -1,58 +1,13 @@
...
 	desc = "A collection of ooking noises made by monkeys"
-	speech_verb = "chimpers"
-	ask_verb = "chimpers"
-	exclaim_verb = "screeches"
+	speech_verb = "says"
+	ask_verb = "inquires"
+	exclaim_verb = "yells"
 	key = "C"
-	syllables = list("ook","eek")
+	syllables = list("chrp","tweet","squick","croak","hiss","gronk","meow","mew","bark", "woof", "bowwow", "yap", "arf") //This is a generalized animal language. If spoken by an animal, it comes out with the mob noises. This only occurs when spoken by a carbon (human).
```
Heroman3003, line 8: Wouldn't that mean that monkey species also will have all the noises instead of ones only relevant to their monkey languages?

code/modules/blob2/overmind/overmind.dm
```diff
@@ -25,7 +25,7 @@ var/list/overminds = list()
 
 	universal_understand = TRUE
 
-	var/list/has_langs = list(LANGUAGE_BLOB)
+	var/list/has_langs = list(LANGUAGE_ANIMAL)
```
Heroman3003, line 28: Blob might be unique enough to maybe have its own language? Idk that's subjective.

code/modules/mob/language/station.dm
```diff
@@ -125,19 +125,6 @@
 	flags = WHITELISTED
 	syllables = list("qr","qrr","xuq","qil","quum","xuqm","vol","xrim","zaoo","qu-uu","qix","qoo","zix")
 
-/datum/language/skrellfar
-	name = LANGUAGE_SKRELLIANFAR
```
Heroman3003, line 129: If far Skrellian is being removed; might wanna look into code. I think SOMEWHERE was implant specifically to allow non-skrell to speak it that might need cleaning up.

code/modules/mob/language/monkey.dm
Cameron653: Correct. It's an edge case, however. Monkeys don't speak on their own so it'd be a player controlled monkey that would make these sounds, and player-controlled monkeys aren't able to be obtained normally.

code/modules/blob2/overmind/overmind.dm
Cameron653: We were talking about blob in the meeting and it was deemed such a rare thing to occur (We only used it for april fools), so it having an animal language isn't an issue.

code/modules/mob/language/station.dm
Cameron653: Yeah there was a few places that added LANGUAGE_SKRELLIANFAR and I went ahead and removed them. The way I ensured it was that if LANGUAGE_SKRELLIANFAR was mentioned anywhere in the code after I removed its define it would error out (Which is what happened with the EVENT1 species).

code/__defines/mobs.dm
```diff
@@ -330,7 +330,7 @@
 #define SPECIES_SHADOW			"Shadow"
 #define SPECIES_SKELETON		"Skeleton"
 #define SPECIES_GOLEM			"Golem"
-#define SPECIES_EVENT1			"X Occursus"
+#define SPECIES_EVENT1			"X Anomalous"
```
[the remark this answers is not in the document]
Cameron653: At first I removed EVENT1 entirely, but then I realized it's used in quite a few places in the code, as was EVENT2 and EVENT3.

It required less touching the code to just swap EVENT1 over to the same as EVENT2 as opposed to removing either EVENT1/2/3 entirely and only having two EVENT species.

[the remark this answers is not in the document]
Heroman3003: Okay, but why not add the name back in properly? Because now in code there are two species with same name and game really won't like it. I mean, I know our staff pretty much doesn't use event species ever; since they require var-based customizations to work, but it is better to avoid that scenario in the first place.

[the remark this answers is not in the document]
Cameron653: Fair point, let me go ahead and change their name back. Thinking about it, it shouldn't really affect much if I leave the name as it was.

PR #13486, opened 2022-08-13
Adds mentors

lm40:
This is not really a great implementation so if you don't want it then just close it, but I spent a few hours with it so I figure I make the PR.

Adds mentors in a shameless copypaste of the admin ticket system. Mentors are pulled from the currently seemingly unused mentors.txt, and have the ability to see mentorhelps only through a verb under the admin tab, as well as chat in a mentor-only channel. Every client gets a mentorhelp verb that is basically like the adminhelp verb except for this secondary ticket system only. Admins can see the mentor channel and use mentor verbs through being an admin, but pure mentors do not interact with admin tools or code in any way, only their own mentor ticket system. Mentorhelps show the mentorhelper's ckey (and a mentor's response to a mentorhelp shows their ckey) but extremely little information otherwise compared to adminhelps since I rather start them off with too little than too much. Any admin with +ADMIN can make any non-admin a mentor at will (or remove anyone from mentor at will) but because mentor is attached to the client it will be removed if they disconnect (like how if you deadmin you readmin upon reconnecting). Mentors show up in staffwho but don't have ooc colors or anything. Mentors cannot (except maybe through some kind of incredible href abuse) initiate a pm, they can only respond to people who mentorhelp. The only things you can really do with a mentorhelp are resolve it or respond to it (or re-open it if it was previously resolved).


lm40:
i remake this

PR #13490, opened 2022-08-13
Makes brainzsnax actually snacks

Screemonster:
Missing callback meant brainzsnax didn't actually provide any nutrition. This fixes that.

Also fixes the red variant adding both red reagent _and_ normal reagent.

### DIFF
--- a/code/modules/reagents/reagents/food_drinks_vr.dm
+++ b/code/modules/reagents/reagents/food_drinks_vr.dm
@@ -536,6 +536,7 @@
 				H.feral = 0
 				to_chat(H, "<span class='info'>Your mind starts to clear, soothed into a state of clarity as your senses return.</span>")
 				log_and_message_admins("is no longer feral.", H)
+	..()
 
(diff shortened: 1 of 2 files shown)


code/modules/reagents/reagents/food_drinks_vr.dm
```diff
@@ -536,6 +536,7 @@
 				H.feral = 0
 				to_chat(H, "<span class='info'>Your mind starts to clear, soothed into a state of clarity as your senses return.</span>")
 				log_and_message_admins("is no longer feral.", H)
+	..()
```
Novacat, line 539: This should be probably at the top of the proc, to be consistent. At roughly line 256?

Novacat (changes requested):

code/modules/reagents/reagents/food_drinks_vr.dm
Screemonster: amended, also fixed the check above to be consistent with the recent feralness refactor

PR #13502, opened 2022-08-15
Fixes typo

Cameron653:


Very-Soft:
But it looks fine how it is, I don't see why this needs to change

PR #13515, opened 2022-08-16
taurs are no longer fat shamed by human_movement.dm

TankTheBirb:
why was weight even part of a mechanic in the first place? its like, not mechanical anywhere else :v


Cameron653:
I saw this before and skimmed over it without even realizing what it was for.

That's actually kind of humorous taurs moved slower with fat people on their backs.

PR #13519, opened 2022-08-16
Small Staff Request addition

Very-Soft:
It tells you it worked if it reaches the end of the proc now


Heroman3003:
Whoops, I thought about adding it but forgot. Good catch.

PR #13528, opened 2022-08-17
Fixes proteans disappearing up their own ass when confused

### LINKED ISSUE #13522: Proteans Can Trip Into Themselves And DIE
Proteans can slipnom themselves when turning back into humanoid form.

They turn back into humanoid form without issue.

If you are afflicted with the confused status effect, or some manner of triggering slipnoms, if you go into blob form then change back while in motion, you'll trip over your protean blob on re-spawn, and be eaten. The blob then deletes, and by extension, you get deleted along with it.

You need to have spontaneous pred AND prey enabled for this to occur.

Easiest way to reproduce
- Flip until confused
- Immediately blob form
- Restore humanoid form
- Immediately attempt to move

Heroman3003:
Fixes #13522


Cameron653 (approved):
no more protean AV allowed

PR #13530, opened 2022-08-18
Allows use of Virtual Avatars in game without breaking things.

Cameron653:
No this isn't as amazing as it sounds. It's just that it allows for use of VR people without having a "Breaks everything" button that is vastly inferior to our custom coded promethean opaque toggle.
Gives them all the normal, Virgo specific promie stuff as well.


Cameron653:
DNM while I add in a few extra verbs for them to use like normal prometheans have.

Cameron653:
Good for merge now.

PR #13531, opened 2022-08-18
Allows custom species altevians to take various bits

Heroman3003:
I really dont like how polaris up-ported our customization just to ban everyone from using it so that we now have to explicitly permit everyone to use it again.


Screemonster:
I wonder if we could refactor it to just have a SPECIES_ALL tag that skips the check

PR #13534, opened 2022-08-19
Makes the freezer able to hold more items at cost of size.

Cameron653:
- Increases the size of the portable freezer itself by 50%
- Allows the freezer to hold bloodbags, beakers/bottles, drinks, and food instead of just organs.

Was talking to someone about freezer bags and how they are completely unused currently due to the restrictiveness of them, rendering them useless. This was due to the fact that freezer bags simply held organs and nothing else, which being able to print organs means it's completely useless unless doing ghetto surgery in the field (bad idea due to failure %) or saving an organ from the field to put back on a person (again, useless due to being able to print organs on station)

This change expands them somewhat, allowing them to hold a wider variety of things at the cost of an increased storage size.

As far as I was informed, they were nerfed on Polaris due to their capacity vs normal boxes, as people were filling them with beakers/pillbottles. This change ultimately resulted in medical just emptying out a medkit instead and using that store their pillbottles/beakers instead since it held slightly less.

# The dreaded math:
#### Beakers hold 60u.
(Large and bluespace beakers exist, but aren't as strong as the pillbottle volume, so not being factored in)
#### Pillbottles hold 14 pills which can each hold 60u each pill.
This means pillbottles have, per bottle, 840u max of raw-reagent-storage per bottle.

Currently a medical kit has a storage space of 14 (7 beakers / pillbottles) beaker / pillbottle volume: **(420u/5880u)**
The Portable Freezer has a storage space of 20 (10 beakers) w/ a total volume of: **(600u)**

Both of these have a storage_cost of 4, meaning that the value of each storage unit of backpack space was:

**Medical Kit: (105u/1470u) per backpack storage space used**
**Portable Freezer: (150u) per backpack storage space used** (prior to being limited to only being able to hold organs)

This meant that the portable freezer was ~50% more efficient per backpack space used than the medical kit when using only beakers.

With this change, it expands the portable freezer's list of items able to be held while making it slightly weaker than the medical kit.

Now with a storage_cost of 6, it means that the portable freezer still can hold 10 beakers but at the cost of 1.5x the size of a medical kit in terms of backpack storage.

What this comes out to is (100u) per backpack storage space used, meaning it is _slightly_ less efficient than a medical kit in terms of raw-reagent storage when using beakers, while still 58x weaker than a medkit when using pillbottles while also allowing for more uses than solely organs.

The portable freezer can't hold pillbottles either, meaning that when it comes to raw-reagent storage, the medical kit is still the tool of choice for raw-reagent-storage.

### DIFF
--- a/code/game/objects/items/weapons/storage/boxes_vr.dm
+++ b/code/game/objects/items/weapons/storage/boxes_vr.dm
@@ -36,3 +36,6 @@
 
 /obj/item/weapon/storage/box/brainzsnax/red
 	starts_with = list(/obj/item/weapon/reagent_containers/food/snacks/canned/brainzsnax/red = 6)
+
+/obj/item/weapon/storage/box/freezer
+	can_hold = list(/obj/item/organ, /obj/item/weapon/reagent_containers/blood, /obj/item/weapon/reagent_containers/glass, /obj/item/weapon/reagent_containers/food)


Heroman3003:
just adjusting storage_cost manually and not per item-size-tiers is a terrible idea, IMO. Unless itemsize_large is itemsize_normal + 2; this will cause great inconsistencies.

Heroman3003:
A second thing of why the freezerbox is theoretically a lot better and part of why everything but organs was cut out is general fact of its ability to hold itemsize_normal items while medkit can only hold itemsize_small.

code/game/objects/items/weapons/storage/boxes_vr.dm
```diff
@@ -35,4 +35,8 @@
 	starts_with = list(/obj/item/weapon/reagent_containers/food/snacks/canned/brainzsnax = 6)
 
 /obj/item/weapon/storage/box/brainzsnax/red
-	starts_with = list(/obj/item/weapon/reagent_containers/food/snacks/canned/brainzsnax/red = 6)
\ No newline at end of file
+	starts_with = list(/obj/item/weapon/reagent_containers/food/snacks/canned/brainzsnax/red = 6)
+
+/obj/item/weapon/storage/box/freezer
+	storage_cost = ITEMSIZE_COST_NORMAL + 2 //1.5X the size of a medkit. You can carry organs in it!
+	can_hold = list(/obj/item/organ, /obj/item/weapon/reagent_containers/blood, /obj/item/weapon/reagent_containers/glass, /obj/item/weapon/reagent_containers/food/drinks, /obj/item/weapon/reagent_containers/food)
```
Heroman3003, line 42: /obj/item/weapon/reagent_containers/food implicitly includes /obj/item/weapon/reagent_containers/food/drinks. Why double definition?

Cameron653: Completely missed that, let me go ahead and fix it.

Cameron653:
> just adjusting storage_cost manually and not per item-size-tiers is a terrible idea, IMO. Unless itemsize_large is itemsize_normal + 2; this will cause great inconsistencies.

Itemsize_large is 8, and there's currently no itemsize that is equal to 6. 

I was looking around of how to increase the itemsize to something in between two size classes and the only way I could find was how the bag of holding does it, which is "storage_cost = INVENTORY_STANDARD_SPACE + 1" which I replicated here.

If there is a better way to do it (I could do itemsize_normal * 1.5 or doing ITEMSIZE_COST_NORMAL + ITEMSIZE_COST_SMALL but these result in the same) I'll gladly change it, but I don't believe there any better way outside of that due to the fact that the code involved either checks storage_cost or w_class (seen here: https://github.com/VOREStation/VOREStation/blob/6ac73f563c8552987ee8315896d89e7be3a11e0c/code/game/objects/items/weapons/storage/storage.dm#L749-L765

I went ahead and changed it to ITEMSIZE_COST_NORMAL + ITEMSIZE_COST_SMALL for the time being to dictate it is a "middleground" object, effectively 1.5x the size of a normal item.



> A second thing of why the freezerbox is theoretically a lot better and part of why everything but organs was cut out is general fact of its ability to hold itemsize_normal items while medkit can only hold itemsize_small.

I can see the reasoning there, but with its previous list (Blame only shows that it had "can_hold = list(/obj/item/organ, /obj/item/weapon/reagent_containers/food, /obj/item/weapon/reagent_containers/glass)") it really wasn't too big of an issue, given the fact that even if someone had the _largest_ glass in the game (bluespace beaker) it still pales in comparison to the raw capacity of a normal box/medkit utilizing pills for maximum storage potential.

Heroman3003:
Custom storage_cost is currently only used for 1 item and that's backpack, because they aren't really supposed to fit inside any container with very select exceptions (lockers basically); meaning they have an explicit reason to be outside normal boundaries (to not fit into any 'inventory container', but fit into lockers/crates); while also not really having any practically-noticable effect since you cant really see that they're unique size, ever. Making the freezerbox have unique snowflake size just to balance it that is still explicitly between existing sizes rather than a real size; and meaning its the only item in the game of that size, seems very dumb to me, especially since a lot of game still defaults to referring to item's w_class for calculations other than storing. Basically, current storage system is kind of... Very much designed around items fitting directly into one of the storage classes and not really being between them. Only exception is 'large storage' item that is only like that so that it can still fit into non-item storages, but isn't supposed to fit into any item ones.

Wickedtemp:
I just like the new cold box

Cameron653:
I personally don't care much about the efficiency but tried to modify it so its beaker-efficiency was on par with the medkit efficiency it's not really much of a problem since the medkit can still hold _way_ more reagents than the portable freezer possibly can ever dream of.

I had a concern that the PR would have problems due to balance concerns but, honestly, after running the numbers between the efficiency of a normal box/medkit and the portable cooler, it honestly doesn't need to be brought to the same level.

If there's concerns about it being too strong, I can just lower the capacity like you mention, but I highly doubt anybody will look at the portable freezer and try to use it for that purpose.

Heroman3003:
Well, main reason it was nerfed originally is because an explo situation where it was used as 'better medkit' occured at some point, but if that doesn't happen, I imagine that won't be problematic

Cameron653:
I don't know if I technically opened pandora's body by pointing out how they could boost their raw-reagent-value by using pillbottles & just dissolving them in a beaker and hold literal thousands of reagents in one box, but whatever.

Exploration shouldn't be a problem (and shouldn't be balanced around since it's being removed soon) and if they are trying to use this to min-max, they're minmaxing wrong.

PR #13535, opened 2022-08-19
Increases bluespace interferance

lm40:
This makes it so that trashbags of holding and pouches of holding (EDIT: and tool belts/medical belts of holding) generate teleportation interference in the same way as bags of holding.

### DIFF
--- a/code/datums/helper_datums/teleport.dm
+++ b/code/datums/helper_datums/teleport.dm
@@ -1,3 +1,10 @@
+var/bluespace_item_types = newlist(/obj/item/weapon/storage/backpack/holding,
+/obj/item/weapon/storage/bag/trash/holding,
+/obj/item/weapon/storage/pouch/holding,
+/obj/item/weapon/storage/belt/utility/holding,
+/obj/item/weapon/storage/belt/medical/holding
+)
+
 //wrapper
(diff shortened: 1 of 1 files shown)


Heroman3003:
I would suggest also belts tbh because they are very much 'expand storage absurdly' versions of existing items too, but that's about it. ~~Also maybe defining it as a list and running a for(var/entry in bluespace_interfering_items) list loop may be better code-wise solution, but while the pool is so low, the current version works~~

lm40:
Yea, I agree that is a very ugly solution that for some reason I chose not to question, this is much more future-proofed. I'm still kinda on the fence about belts just because they are limited in what they can hold so for now I'm not putting them in.

Novacat:
I agree with Heroman in that if Bluespace pouches and Bluespace trashbags are included, that bluespace belts should as well if for nothing else than consistency sake. The belts are limited in what they can hold but they can still hold twice the amount of things as normal belts, and they can hold a lot of the things that people put into belts (like medicine bottles). Being able to carry around 2x the amount of medicine is strong.

I do think it is very strange that, at the moment, bluespace pouches are more easily obtained than normal pouches (how do you even get normal pouches?), and have way, way too much storage capacity for what they are.

Bluespace badges are an exception because they do not actually expand your storage space, and just exist as an aesthetically minimal form of webbing.

lm40:
Well if you both say so it's easily enough done now anyway.

PR #13541, opened 2022-08-20
Adds customizable language keys to language menu and character setup

Heroman3003:
Allows you to set custom language keys for your languages.

Only keys set in character setup save and persist, inround changes do not.

Only letters (uppercase and lowercase) and numbers are accepted as keys. Inputting a single space (" ") will clear the key. Inputting anything else will result in error.

Custom keys do not replace existing key of language they're set for. If you set Tavan's (default E) language key to j, then both E and j will be usable as keys.

Custom keys DO override keys of other languages though. If you have both Tavan (default E) and Siik (default j) and set Tavan's custom key to j; you'll be able to use Tavan with both E and j, but not be able to use Siik at all until you give it its own custom key or change Tavan's custom key.

### DIFF
--- a/code/modules/client/preference_setup/general/02_language.dm
+++ b/code/modules/client/preference_setup/general/02_language.dm
@@ -46,6 +46,14 @@
 	for(var/prefix in pref.language_prefixes)
 		if(prefix in forbidden_prefixes)
 			pref.language_prefixes -= prefix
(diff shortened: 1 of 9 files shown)


Cameron653 (approved):
Provided this is fully tested and works, has my approval.

PR #13543, opened 2022-08-21
Allows painkillers to negate slowdown due to damage

Cameron653:
Prior to this change, painkillers worked on most causes of slowdown (reducing traumatic_shock, which when above 80, would start accumulating shock_stage) but not all.

Having 40 damage or more would cause you to have slowdown. 

Even if your arms and hands were hit with 10 damage each (40 damage total, no broken bones) and then you downed the strongest non-slowing down painkiller (tramadol) you would still be slowed down like it did nothing, even though the pain pills are taking off double the pain you're currently in.

With this change, it makes your pain take painkillers into account.

Additionally, as an unintended side effect, it allows for PAIN INCREASERS to be used. Meaning if the person is in any pain, it will amplify their pain slowdown.

I can't help but think this is a bug/an oversight as the code involved is **10 years old (https://i.imgur.com/kwt6CVw.png)** but it's been in the game so long that it can probably be viewed as a feature change as well.

### DIFF
--- a/code/modules/mob/living/carbon/human/human_movement.dm
+++ b/code/modules/mob/living/carbon/human/human_movement.dm
@@ -20,7 +20,11 @@
 			. += M.slowdown
 
 	var/health_deficiency = (getMaxHealth() - health)
-	if(health_deficiency >= 40) . += (health_deficiency / 25)
+	if(health_deficiency >= 40) //VOREStation Edit Start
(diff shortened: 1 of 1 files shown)


Skysteak:
Solid PR and I think it's a positive change to the game, rewarding forethought and permitting an avenue for easier escape from a bad situation. This also does not affect slowdown from broken legs/feet which means that cannot be simply ignored either, which is nice. If in the future we get some adrenaline or 'fight or flight' style mechanics then I think it'd be fair for this to be looked at, but that's not something I am privy to if it'll even happen.

PR #13546, opened 2022-08-21
Adds Toxins & Cloneloss digestion damage.

Cameron653:
- Adds Toxins as a digestion damage type.
- Adds Clone damage as a digestion damage type.

- Cloneloss gives 2x nutrition than brute/burn since it is exceptionally harder to heal from. (not that it matters)

So now slime characters can actually RP out sucking out your DNA and leaving you a DNAless husk while slamming you with toxins like xenobio slimes can.


BonniePandora:
One slight concern, this about halves the time you have to find and smack the OOC-escape button before entering crit if you end up in a blender-set belly by mistake

Cameron653:
Yeah unfortunately there's no way to really fix/avoid that without adding something like a grace period to entering stomachs, which is outside the scope of this PR.

Currently a blender stomach will do 6/6/12, meaning 24 damage a tick. That gives you 5 ticks before you're knocked into crit (which most of it is oxydamage so you will heal it back up if the pred does CPR). This means 9 ticks until you're killed.

With this, blender stomachs can be 6/6/12/6/6/, meaning 36 damage a tick. That gives you 3 ticks before you're knocked into crit, which should still be plenty of time, especially seeing as it would take 6 ticks to fully kill you.

So it decreases the time-to-crit by 2 ticks and the time-to-death by 3 ticks, which should still be plenty of reaction time for a pred to spit someone out and then toss them into a heal belly after they OOC escape.

Runa-Dacino (commented):
How does toxloss affect synthmorph prey?

Radiation damage for synthmorphs is handled as toxin damage, and is cleared by entering a cyborg recharger.

In this case, would toxloss applied to synthmorph prey cause them to lose "system integrity" (thus causing confusion etc.) or do nothing?

If it causes system integrity damage, it may be worth adjusting the relevant bits of code to prevent Confusion/Shutdown/stuttering like we already do for shock_stage.

Cameron653:
DNM while I need to fix some problems.
1. Checking how toxins affects synthmorphs
2. Ensuring that people don't puke in stomachs while afflicted by toxins.

Cameron653:
I actually found that both the synthmorph toxins damage and the human toxin damage (vomiting) is performed in the same location.

Going ahead and modifying it so that doesn't happen.

Cameron653:
Fixed! Now synths won't get confused and people won't vomit. Tested it by having the new digestion toxins damage give them 66 damage and letting them sit in the stomach for a long time. No vomiting to be had!

This is now ready for merge provided there are no more objections.

PR #13549, opened 2022-08-21
Why were these in gitignore

Cameron653:
It makes no sense and if someone ever wanted to make a change they would be incapable of doing so

Also overload was listed there twice and after looking back at it, this was included in a Polaris sync I did 6 years ago, meaning that this was due to some kind of mistake.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,9 +22,6 @@ cfg/
 !/.vscode/settings.json
 !/.vscode/tasks.json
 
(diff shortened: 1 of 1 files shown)


Cameron653:
As a side note, the .gitignore file does NOT seem to work on file paths like the three being deleted here, for some odd reason.

That should probably be looked at but it's an extremely low priority.

PR #13559, opened 2022-08-23
New sprite for dosimeter case + case capacity fix

FirstSgtDan:
It was using the default cardboard box sprite.
Also fixed the dosimeter case capacity so all items it comes with can be placed back inside without problems.

dosimeter_case


Cameron653:
DOSIMETER UPDATE
LET'S GO

BonniePandora:
That's a _lush_ looking sprite for it

PR #13577, opened 2022-08-25
Fixes linter fails in upstream-merge-8298.

MistakeNot4892:
Apologies for giant diff, I pulled from master to make sure no new ones had snuck in. This is meant to fix the reported failures with #11826.

Commits that are actually relevant to the fixes are:
https://github.com/MistakeNot4892/VOREStation/commit/a37e0701214b9bdbac609ff20163aa324fd64b7c
https://github.com/MistakeNot4892/VOREStation/commit/7082955d2362a3123a1e597dc1562f6b3d0aaa90


Cameron653 (commented):
The files I have skimmed through look fine, but this is way, _way_ over my head (as to what is actually going on here.) I kind of get what's happening here though so...

Should be fine to merge...? The two mentioned commits look fine.

PR #13596, opened 2022-08-27
[MIRROR2] /atom New() => Initialize() [DNM]

Novacat:
Converts a lot of New() to Initialize(). This was previously merged but there were issues, I am setting up this branch if people want to test/fix it.


code/datums/ghost_query.dm
```diff
@@ -20,10 +20,10 @@
 
 	// Then wait awhile.
 	while(!finished)
-		sleep(1 SECOND)
-		wait_time -= 1 SECOND
-		if(wait_time <= 0)
-			finished = TRUE
+		spawn(1 SECOND)
```
KillianKirilenko, line 23: revert this

code/game/objects/effects/landmarks_poi_vr.dm
```diff
@@ -13,7 +13,8 @@ var/global/list/global_used_pois = list()
 INITIALIZE_IMMEDIATE(/obj/effect/landmark/poi_loader)
```
KillianKirilenko, line 13: change to LateInitialize

code/datums/ghost_query.dm
Heroman3003: This was replaced with spawn because linter refuses to accept any form of sleeping in initialization.

KillianKirilenko: well then the linter needs a smack, 'cause polaris' version of this code still uses sleep

Heroman3003: Because Polaris I don't believe has any ghostpods that activate instantly upon spawning. We do. Hence the difference.

Cameron653:
Leaving this up since it's still being actively worked on as far as I can see.

PR #13603, opened 2022-08-29
Mouse hole!

Very-Soft:
Adds mouse holes. 

These are based on the kob tunnels I'm working on, but they work good and let micros/mice enter and traverse them!

Non-micros/mice can click the hole to attempt to scoop someone inside! If there's more than one thing in the hole, what they grab is random.

There's no way to build these or anything for the moment. Maybe that can be a thing eventually! For now it is intended to be used in an upcoming event thing! There will be some other things to follow.


code/game/objects/micro_structures.dm
```diff
@@ -0,0 +1,201 @@
...
+
+	anchored = TRUE
+	density = FALSE
+
+	var/max_accepted_scale = 0.3
+	var/magic = FALSE	//For events and stuff, if true, this tunnel will show up in the list regardless of whether it's in valid range, of if you're in a tunnel with this var, all tunnels of the same faction will show up redardless of range
+
+/obj/structure/micro_tunnel/Initialize()
+	. = ..()
+	if(name == "mouse hole")
```
Heroman3003, line 15: I think better way to do it is to make a variable called something like "autoname" and if its true, set our name upon init, otherwise skip, rather than checking if that specifically is our name.

code/game/objects/micro_structures.dm
```diff
@@ -0,0 +1,201 @@
+/obj/structure/micro_tunnel
+	name = "mouse hole"
+	desc = "A tiny little hole... where does it go?"
+	icon = 'icons/obj/structures/micro_structures.dmi'
+	icon_state = "mouse_hole"
+
+	anchored = TRUE
+	density = FALSE
+
+	var/max_accepted_scale = 0.3
```
Heroman3003, line 10: I'd suggest maybe a bit higher. Quite a few of people playing micro use higher size despite being almost minimal RP size because of sprite visibility. 0.5 might do.

code/game/objects/micro_structures.dm
```diff
@@ -0,0 +1,201 @@
...
+/obj/structure/micro_tunnel/attack_generic(mob/user, damage, attack_verb)
+	attack_hand(user)
+	return ..()
+
+/obj/structure/micro_tunnel/MouseDrop_T(mob/living/M, mob/living/user)
+	. = ..()
+	if(M != user)
+		return
+
+	if(!can_enter(user))
```
Heroman3003, line 166: Maybe error message on trying to enter when too lorg?

code/game/objects/micro_structures.dm
```diff
@@ -0,0 +1,201 @@
...
+			pixel_y = -32
+		if(4)
+			pixel_x = 32
+		if(8)
+			pixel_x = -32
+
+/obj/structure/micro_tunnel/attack_hand(mob/user)
+	if(!isliving(user))
+		return ..()
+	if(user.loc == src)
```
Heroman3003, line 52: Can you actually properly attack_hand the tunnel while inside it? If so disregard. Although I'd be mildly hmmm if you can interact with anything around from inside the tunnel as well.

Very-Soft: You can deffo click things while you're inside them!

code/game/objects/micro_structures.dm
Very-Soft: When you are too big to enter you put your hand in to try to pull things out!

code/game/objects/micro_structures.dm
Very-Soft: I don't think is necessary, either you name it yourself or it names itself after the area. When you go to select a destination, if there's things that are the same name, then it only puts that name on the list once, and picks one to go to at random. I'd prefer they get renamed than not.

Heroman3003: I am mostly proposing it for possibility of subtypes of this, or having different name but also adding area in front... Actually, in that regard, potentiall setting name to "[our_area.name] [name]" might work better. Still, its minor enough, just thought to suggest.

code/game/objects/micro_structures.dm
Heroman3003: I mean specifically for interaction I commented on, where you click-n-drag yourself onto the hole

PR #13605, opened 2022-08-30
Micro hiding! 

Very-Soft:
Adds a var and a proc to all objects in the game! 

If an object starts with the 'micro_target' var enabled, then it will obtain the 'micro interact' proc as a verb on initialize. 

The 'micro interact' proc, allows micros and other small mobs such as mice, to hide inside of whatever the object is. Additionally, those inside one object can use the verb to move to an adjacent object that has the 'micro_target' var. 

Non micros/small mobs can also use the 'micro interact' verb to attempt to extract anyone who might be within the given object.

### DIFF
--- a/code/game/objects/micro_structures.dm
+++ b/code/game/objects/micro_structures.dm
@@ -7,8 +7,8 @@
 	anchored = TRUE
 	density = FALSE
 
-	var/max_accepted_scale = 0.5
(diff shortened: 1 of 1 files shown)


Cameron653:
Neat. Can see this being fun and allowing for hide&seek stuff as well.

Heroman3003 (approved):

Cameron653 (approved):
Looks good

PR #13618, opened 2022-08-31
Adds colorized stomachs. [TEMP DNM]

Cameron653:
Primarily ported from CHOMP downstream, but modified to work better with VIRGO code and keep original sprites.

- Allows for colorizable stomachs
- Adds a new 'preview' button that allows you to see what prey will see when in your stomach. Fades after 2 seconds.
- Adds two new stomach sprites, 'synthflesh' and 'animated stomach' (Synthflesh: https://i.imgur.com/xadUJ5k.png) (Animated: https://i.imgur.com/dIBtCSp.gif)
- Adds framework for 'multi-layered' stomachs in the future.
- Touches TGUI and makes me cry.

The panel with the additions looks like this: https://i.imgur.com/Tm3EfNW.png.
When swapped to 'colorizable stomachs only' it looks like this: https://i.imgur.com/W1jOyH1.png

Example of the preview button being used with a normal stomach: https://i.imgur.com/6TX88Oa.png
Example of the preview button being used with a colored stomach: https://i.imgur.com/I9XggVe.png

That's kind of all there is to it.
DNM while waiting for more icons that can be recolored.
Current icons that are available to be used: https://i.imgur.com/43zcgBF.png

I did run into an issue where the 'vore_colorized' spritesheet refuses to be loaded, even though it's in the proper places. This isn't an issue as I used a workaround, but would it would be nice to not have to do such. However this isn't a problem as all it requires is any new stomachs to just be added to both files and excluded from the list of normal stomachs, which is simple to do.


Cameron653:
TGUI comes in yet again to kick me in the nuts

PR #13632, opened 2022-09-03
Fixes typo in stumblevore_vr.dm

Runa-Dacino:
Minor typo. May actually have been an intended feature. Still, quick web-edit in case it was an oversight.

carlessly -> carelessly


Very-Soft:
But, I don't want any carls in my slipnoms????

PR #13640, opened 2022-09-03
[MIRROR] Fix telecomms monitoring consoles (NEEDS TESTING)

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/8699
--------------------
Fixes #8610.
evidence


Cameron653:
This touches tcomms. I got rid of all the errors and the compile issues, but I don't feel good about pushing this through without being tested.

It does initialize stuff and changes a few var names around, mostly.

_It should be safe_ but I'd rather this be tested.

Cameron653:
We already fixed tcomms consoles apprently so this isn't needed.

PR #13659, opened 2022-09-04
Allows light manipulation via multitool.

Cameron653:
- Allows lights to be changed and manipulated via a multitool.

- Allows six things to be changed:
1. Normal operating Range
2. Normal operating Brightness
3. Normal operating Color
4. Nightshift Range
5. Nightshift Brightness
6. Nightshift Color

All these (barring color) have their max set to the light's individual variables. For example: You can't make a light tube brighter/shine further than it normally would.

image

image
image


Very-Soft (commented):
be cool if it changed the overlay color too!

Issue #12023, opened 2021-12-19
The snack item, "Sushiraki pack" lacks a sprite for its depleted version


Runa-Dacino:
#### Brief description of the bug

The snack item, "Suharaki pack" lacks a sprite once depleted of nutrients, making it seem that your hands are empty. 

#### What you expected to happen
Once you consume the snack item, "Suharaki pack", for it to show a sprite that depicts an empty pack of snacks rather than be invisible.

#### What actually happened
Invisible item without a sprite that makes you think your hands are empty,

#### Steps to reproduce
- Approach a"Snix" vending machine
- Purchase "Suharaki"
- With help intent, use it on yourself to consume. Repeat until out of nutrition.
- Watch as the sprite disappears

#### Code Revision
- Server revision: B:release D:
Commit: 07d00c670d81bd546b7b763444ddee105f7a18d6 
 

#### Anything else you may wish to add:
-Shop shows sprite
![image](https://user-images.githubusercontent.com/20523270/146695310-2616e154-dd14-41bd-8c47-52385c1ddba0.png)
-Examine image on ground/hand:
![image](https://user-images.githubusercontent.com/20523270/146695322-bb80a29f-de50-482a-ba9f-d0e3846aaa0f.png)
- Sprite while full on ground:
![image](https://user-images.githubusercontent.com/20523270/146695328-8c2bc7a2-7598-47af-a92a-a3a4f92fe727.png)
-Sprite in hands while full:
![image](https://user-images.githubusercontent.com/20523270/146695333-aa98d8f4-fef3-4153-9836-defaa66b2ff7.png)
-Examine sprite once empty
![image](https://user-images.githubusercontent.com/20523270/146695239-685a82a6-a73e-4555-b6a2-efc3b4f85c75.png)
-The (lack) of sprite in hands once empty.
![image](https://user-images.githubusercontent.com/20523270/146695346-39155fbb-74c3-40bd-a680-1f2d06aa9c3e.png)
- The (lack) of sprite on the ground once empty
![image](https://user-images.githubusercontent.com/20523270/146695354-ce9e791c-7fd9-4b2a-82a9-df47f673f097.png)

PR #13665, opened 2022-09-05
Fixes missing suhariki trash sprite

Runa-Dacino:
Fixes https://github.com/VOREStation/VOREStation/issues/12023

There was no "croutons" object within trash.dmi.

I copied suhariki from food_snacks.dmi into trash.dmi, and did some quick edit to the sprite.

Snack sprite:

image

Trash sprite:

image


Heroman3003:
I was actually looking into fixing that issue a while back. But the misspelling of item in the issue made me unable to find 'wtf is the sushiraki'.

Runa-Dacino:
> I was actually looking into fixing that issue a while back. But the misspelling of item in the issue made me unable to find 'wtf is the sushiraki'.

I did spell it right in the body of the bug report, alongside screenshots! I did also fail to find it by title alone.

PR #13683, opened 2022-09-07
Fixes bug regarding feral xenochimera and treating others/being treated

TheGreatKitsune:
This bug prevented feral xenochimera or other 'non-dexterous' creatures from being healed, and also allowed them to heal others.


Cameron653 (approved):

Screemonster:
This wouldn't have just been affecting xenochimera, wouldn't this also make it impossible to heal monkeys?

TheGreatKitsune:
It would, yes. Xenochimera are where I saw the issue happening most, though.

PR #13689, opened 2022-09-08
Implements rudimentary adrenaline system

Runa-Dacino:
Adds a new verb under "abilities" for all human mobs: "Toggle Adrenaline Response" - this is set to "OFF" by default, requiring manual activation.

Upon being activated, the player is notified that they are ready for trouble.

Should they get hurt by brute, burn, oxloss, broken bones, damaged organs, poison or even genetic damage - if their "shock" reaches 20 (for brute/burn, they need 15, for broken bones 1 broken bone is enough), adrenaline kicks in. Important to note: Tasers and holodamage and space jellies will NOT activate adrenaline.

Adrenaline scales with the amount of pain you have, lasting longer but also leaving you weaker for longer in turn.

Rather than adrenaline, I named the reagent "Epinephrine" as adrenaline is already a reagent used to stabilize hearts and I was afraid of touching it. This reagent sets "weakness" (being knocked down by carp, from too much pain, etc) to 0, allowing you to stand up far faster than otherwise. Furthermore, you become under effect of a tramadol-level painkiller, and move as if you were on hyperzine.

While you have epinephrine in your system, you start metabolizing it to metanephrine. Metanephrine in itself is harmless - it acts as an inaprovaline level painkiller. However, too much of it in your bloodstream indicates you've pushed your body beyond what is safe. In other words, when it overdoses at 30 units, you start to suffer a plethora of random effects:

Each life tick,
5%, you may suddenly suffer 10 brute damage - you pulled a muscle, tendon, whatever. This can only fire once. After firing, you will only have to deal with it after your body cleared all metanephrine.
2%, you may fall over and get jittery legs that make you run into walls. On the plus side, this clears some metanephrine from your system
10%, You might end up vomitting if you had some food before the incident
50%, you may get dizzy and light headed, your eyes blurring
10%, your hands may start shaking

These effects can occur simultaneously. However, any of them firing will make you safe from further effects until 60 ticks pass.

All numbers, side effects, strengths I am up for changing.

Tested thoroughly on local, found no outstanding bugs related to this system.

A fun note for special use case:

If you set "Toggle Adrenaline Response" to a macro, you can keep it toggled "OFF" until you're in a hairy situation, rather than have it trigger for your first mob encounter.

This can give you a quick boost to run away or fight back, albeit likely at risk of a severe overdose of metaneprhine if you do it after being badly hurt.


Heroman3003:
First of all, I really don't think this should be in 'abilities' tab if its universal to all humanmobs.

Then regarding balancing of it itself... it sounds like its way too extreme on both ends, to me at least. Epinephrine makes you pretty much immune to knockdown, and extra fast and have less pain effects. While metanephrine seems like something that will just make pain all the more miserable to deal with in every way imaginable. And it being a toggle that is so easy to use and free to be used whenever you want kinda feels to me like this is 'get out of jail free' card rather than actual mechanic to balance out the issues pain system has, since it only makes high pain stuff more extreme with exertion chem.

I'm honestly not sure this is good for combat at all, in terms of balancing.


Also, I'm not sure, but its worth looking into how the generation of new chems interacts with the whole 'togglable visual pain for synths' thing. I don't remember immedeately if they use shock for it or not, but if they do, you may get some epinephrine-clogged pipes.

Runa-Dacino:
> First of all, I really don't think this should be in 'abilities' tab if its universal to all humanmobs.
> 
> Then regarding balancing of it itself... it sounds like its way too extreme on both ends, to me at least. Epinephrine makes you pretty much immune to knockdown, and extra fast and have less pain effects. While metanephrine seems like something that will just make pain all the more miserable to deal with in every way imaginable. And it being a toggle that is so easy to use and free to be used whenever you want kinda feels to me like this is 'get out of jail free' card rather than actual mechanic to balance out the issues pain system has, since it only makes high pain stuff more extreme with exertion chem.
> 
> I'm honestly not sure this is good for combat at all, in terms of balancing.
> 
> Also, I'm not sure, but its worth looking into how the generation of new chems interacts with the whole 'togglable visual pain for synths' thing. I don't remember immedeately if they use shock for it or not, but if they do, you may get some epinephrine-clogged pipes.

Synths use pain.dm's mechanics, and completely ignore the shock.dm code.

This uses shock.dm's "traumatic_shock" var, and does not interact with synths!

For the balance, that's the idea - something for people to run away when they get into too much trouble, rather than be forced to "fight or die."

Runa-Dacino:
Added ability to sleep off over-exertion, potentially avoiding side effects if you get away from trouble in time! Resting, buckling onto chairs or outright sleeping all work.

Novacat (changes requested):
Toggling it on/off should not be a thing. Other than that, staff are going to need to look over this very carefully.

Runa-Dacino:
> Toggling it on/off should not be a thing. Other than that, staff are going to need to look over this very carefully.

I'll see if I can add a check for being in a vore belly and maybe increase threshold to 30 (bone break adds 30 shock)

Cameron653:
For the vore belly check, do a 
``if(isbelly(loc)) 
return ``
where the chemical would be injected, so it stops before injecting the chem.
Because as is, the button could easily be abused.

As mentioned before, I'd highly suggest looking into making this a modifier instead of a reagent w/ a metabolite.  This would allow for reduced/increased damage, increased evasion, decreased accuracy, etc etc. 
Example of how to apply a modifier:
add_modifier(/datum/modifier/underwater_stealth)
Example of a modifier that does some things already related to the things mentioned above: https://github.com/Cameron653/VOREStation-CLDev/blob/9952518c1068f933fc0d5f098ac1d78974f8b0a3/code/modules/mob/_modifiers/modifiers_vr.dm

Also would suggest upping the trauma requirement before you have the effect hit off. 20 is a very easy level of pain to achieve from simple injuries.

Runa-Dacino:
@Novacat 

Requested changes commited.

Adrenaline will NOT trigger while inside a belly. It will trigger upon leaving though.

Upped treshold to 30 traumatic shock, to ensure rafters won't trigger adrenaline. Reduced reagent amount to 3/4 of "realPain".

Treshold of 30 chosen as that's the amount needed to break a bone.

Runa-Dacino:
> For the vore belly check, do a `if(isbelly(loc)) return ` where the chemical would be injected, so it stops before injecting the chem. Because as is, the button could easily be abused.
> 
> As mentioned before, I'd highly suggest looking into making this a modifier instead of a reagent w/ a metabolite. This would allow for reduced/increased damage, increased evasion, decreased accuracy, etc etc. Example of how to apply a modifier: add_modifier(/datum/modifier/underwater_stealth) Example of a modifier that does some things already related to the things mentioned above: https://github.com/Cameron653/VOREStation-CLDev/blob/9952518c1068f933fc0d5f098ac1d78974f8b0a3/code/modules/mob/_modifiers/modifiers_vr.dm
> 
> Also would suggest upping the trauma requirement before you have the effect hit off. 20 is a very easy level of pain to achieve from simple injuries.

Adjusted trauma to require at least 25 brute/burn to trigger. 

I'd rather it be a reagent, if possible - since it _is_ a chemical the body produces!

BigBababooey:
I like this concept, but I don't like how it's being proposed here, in this iteration, I think it definitely has a lot of flaws that would create behaviors opposite to what's desired. 

First of all, I do not like the idea of it being activated at all, I believe a random chance past a threshold that scales with each time you get hit past that point might be a far more elegant solution. Or maybe just an assured trigger when a threshold is hit, that's a balance discussion that I don't think I'm suited to. I would never accept this as a button myself, it just is incredibly gamey with that iteration. Though I think you've heard that loud and clear.

Realistically, I think this shouldn't be having all these rolls of stuff, or have some complicated process. I just like the idea of pain/slowdown being offset once a threshold is hit, and also a tank to your ranged accuracy (and perhaps blurry vision) to indicate clearly you're fucked up and need to get out of there in an elegant manner. I even wouldn't mind if melee was given a slight (like 20% max) chance to miss, just enough to communicate to the person in melee too they're not fighting at peak efficiency and should consider falling back. Melee is never really optimal anyways. so I don't want to make it worse, but I don't like this current iteration and it's ability to just... well, pretty much let people go berserk. That strikes me as the exact opposite of what you're going for here.

Screemonster:
If this is taken away from being a toggle as some people are suggesting, I worry that it will interfere with xenochimera's already-existing response-to-injury mechanics in an annoying way. Would it be possible to tie it to a trait or else just make it _not_ apply to every species?

Enzo-Leon:
Concept wise I like it, but as it is right not it's a funny button anyone can press at any time to stunt on a mob or player just sits weird with me. If it's a % chance or something of the sort after conditions are met I'd feel this would be more aligned for something to implement.

Runa-Dacino:
> If this is taken away from being a toggle as some people are suggesting, I worry that it will interfere with xenochimera's already-existing response-to-injury mechanics in an annoying way. Would it be possible to tie it to a trait or else just make it _not_ apply to every species?

This should be doable, yes. Would it be a positive trait, or a neutral one anyone can take?



> Concept wise I like it, but as it is right now, it's a funny button anyone can press at any time to stunt on a mob or player just sits weird with me. If it's a % chance or something of the sort after conditions are met I'd feel this would be more aligned for something to implement.

The button was removed per Ascian's request already.

lm40:
I like this conceptually but I dunno if it should be a reagent. I also didn't like the button but I think that's already gone.

Skysteak:
Firstly I want to express my appreciation for you tackling what is a desirable but understandably delicate feature -- the effort and willingness to listen to feedback both most noticed and welcome! 

Secondly, with the button situation irrelevant now, I personally lean in favour of this triggering in such a way that it's either set or RNG but practically guarantied to fire if enough damage is taken, to the point that it can be considered reliable. I don't have the code know-how for that or an exact figure but I would not want a series of really unlucky rolls to completely deny someone the mechanic essentially. 

Thirdly, as stated in the Discord I feel that some telegraphing of the 'wearing off' state should be provided, particularly if those effects are to be left unchanged. Something akin to 'Adrenaline is wearing off, your drained body desires rest' would be lovely, as it also provides a clue to have a lie down for those whom may be ignorant of the system.

Finally, I do agree with Heroman and Ali in that some of the effects there are a bit rough in terms of side effects. The blurrines and proposed accuracy nerfs (if we must have Baymiss) are good ways to help communicate that this is a bailing out mechanic imo. The vomiting and fall downs could be slightly toned down imo, the latter seeing a tad high. This is assuming you stick with the current setup and don't simplify/streamline it.

Overall however good work! I think the foundation is solid and I am optimistic that the finished mechanic will look excellent!

BigBababooey:
My ideal mental image of this is literally just

- Hit threshold where this activate, whatever its decided to be and figured out, possibly a soft cap where it may trigger, and then a hard cap where it 100% will, get speed boost and pain reduction to allow you to continue moving
- Get blurry vision (and maybe an additional accuracy penalty) to further communicate this as a run away mechanic (but not making it impossible to keep fighting if you choose to)
- This lasts for however long, maybe even just a flat duration after taking the damage, and then you crash. Maybe the timer resets if you continue to take damage past a certain amount. Make the crash just be like a wave of pain so they do collapse after, after all, you're coming down from a massive adrenaline spike.

I think all of this making it a reagent and such is pretty heavily overcomplicating things, and while I like it conceptually, I don't think it's going to make combat any smoother or more fun in that form.

Also just like... make a neutral trait to allow people to turn it off, mostly for Xenochimera, but if someone else wants to, power to them I guess.

TankTheBirb:
Consolidating my feedback here now from testing an earlier version of this. I will not be retreading addressed feedback.

For reagent vs modifier, I think accomplishing timer based stuff may be easier to have as a modifer than as a reagent. Before I get into that, I like the way this metabolite system works but this may not be the best way to use it. Anyways, you also produce a lot of units of these reagents, so if you use dialysis, you're gonna suck out a lotta stuff that isn't the toxin. Modifier also reduces the abuse potential for powergaming what is supposed to be a "run away" thing. I think you can also do a lot more with modifiers, such as affect accuracy, outgoing damage, etc that you can't really do with a chem reagent. Or if you can, it makes less sense when removed from the body. You can do a modifier that gets proc'd by having a reagent in someone, if reagents are deemed the best approach to this.

Threshold-wise, I think the design-specific question is when this should kick in. Is this to counteract escape chance reducers from damage or is this to provide a burst of power when struck hard? How does this affect non-combat activities like being struck by an emitter or zapped by a door? I think more thought may need to be put into how the threshold is set and used, possibly either by applying stacks of a modifier or some other method. Ideally this is proportional as a response to the damage source and the number of sources. More damage and many harm sources (broken bones, big hit, high damage already etc)? More speed and inaccuracy! But maybe shorter with each stack so you can't just keep proccing a stack to be infinitely adrenalined.

Effects-wise, I actually like the concept of the downsides, I just think they won't be as appreciated on a thing every fleshy carbon will have. The upsides should give greater escape chances, and the downsides shouldn't make it harder to run away. HOWEVER, I recommend keeping the method of acquiring a metabolite drug like Meta is, because that might actually be useful for powerful chems that doctors administer. Generally, focus the effects and their scale on what it needs to do. Personally, this is best focused as a "GTFO timer" that gives people time to get clear of danger and get closer to help.

I think you could have neutral/positive/negative traits that modify this as a stretch goal as the system is already very complicated with all factors involved, and traits may be best as a response to feedback for how people may or may not like this.

Final thoughts because it's a lot... I think your conceptual approach to this much requested feature is good. It approaches the problem of how binary combat is in a way that could address the problem. I think this draft proves you have the right IDEA, just the approach needs more tuning. The attention it's getting proves that this is something very interesting, and I think something very good. Honestly this is a massive amount of progress on something pretty ambitious for one day, I'm impressed it's even this far along. And even if some things like the reagents get scrapped, what you used here has potential for more powerful chems to have downsides if you give someone too much too quickly.

lm40 (changes requested):
Also I just noticed, while I don't support this being a reagent regardless there is already a definition of /datum/reagent/epinephrine in https://github.com/VOREStation/VOREStation/blob/9855a5e3702355065fabe50bc131a9424e19b368/code/game/gamemodes/changeling/powers/epinephrine_overdose.dm

Heroman3003:
My personal thoughts on it:

1. Obviously button is bad, I know its been removed, but just pointing out, it being a button made it a lot more game-ified than it should have been.

2. Considering its just an effect occuring inside the mob, spawning a reagent just to give its effects for a certain duration, it really should be considered as a modifier rather than a reagent. Both for ease of use and to prevent unnecesary reagent bloat.

3. Should first two things be done, maybe the whole thing should happen with a scaling chance, rather than scaling duration and not be guaranteed on-cooldown adrenaline bursts, but rather chance that gets higher the more pain you have (and cooldown between checks to be higher than life tick)

4. It prooobably shouldnt have nearly as severe downsides as it does now. Instead it should have downsides that heavily disincentivize continuing to fight and instead use the opportunity to escape. Hamper ability to attack back in some way, but give speed and energy in very short burst instead.

5. Shouldnt be able to trigger past hardcrit threshold for obvious reasons. Also accounting for painless species/species with different handling like xenochimera.

Runa-Dacino:
Updated main post.

old post body:

No more button.

Upon being activated, the player is notified that they are ready for trouble.

Should they get hurt by brute, burn, oxloss, broken bones, damaged organs, poison or even genetic damage - if their "shock" reaches 20 (for brute/burn, they need 15, for broken bones 1 broken bone is enough), adrenaline kicks in. Important to note: Tasers and holodamage and space jellies will NOT activate adrenaline.

Adrenaline scales with the amount of pain you have, lasting longer but also leaving you weaker for longer in turn.

Rather than adrenaline, I named the reagent "Epinephrine" as adrenaline is already a reagent used to stabilize hearts and I was afraid of touching it. This reagent sets "weakness" (being knocked down by carp, from too much pain, etc) to 0, allowing you to stand up far faster than otherwise. Furthermore, you become under effect of a tramadol-level painkiller, and move as if you were on hyperzine.

While you have epinephrine in your system, you start metabolizing it to metanephrine. Metanephrine in itself is harmless - it acts as an inaprovaline level painkiller. However, too much of it in your bloodstream indicates you've pushed your body beyond what is safe. In other words, when it overdoses at 30 units, you start to suffer a plethora of random effects:

Each life tick,
5%, you may suddenly suffer 10 brute damage - you pulled a muscle, tendon, whatever. This can only fire once. After firing, you will only have to deal with it after your body cleared all metanephrine.
25%, you may fall over and get jittery legs that make you run into walls. On the plus side, this clears some metanephrine from your system
10%, You might end up vomitting if you had some food before the incident
50%, you may get dizzy and light headed, your eyes blurring
10%, your hands may start shaking

These effects can occur simultaneously. However, any of them firing will make you safe from further effects until 60 ticks pass.


All numbers, side effects, strengths I am up for changing.

Tested thoroughly on local, found no outstanding bugs related to this system.



Button was removed.


new post body:

New functionality:

If you have a traumatic shock of at least 5 (ergo: not under a strong painkiller) and outside a belly and have opted in with a trait...

We take the difference between your health and your max health.

Maximum once, a broken bone adds 55 to this (pushing you into large dose instantly), while a dislocated bone into 40 (pushing you into low dose).

Value, before evaluation, is multiplied by trauma_mod and reduced by painkiller effects (including adrenaline).

3 types of dosage exist:

Small (8 unit to 10 unit adrenaline):
This gives you paracetamol, which will delay pain slowdown from getting hit. This will accumulate you equivalent metanephrine without side effects as long as you don't get hurt more. This stops working once 5 metanephrine accumulates (you still get painkiller effect, but no further adrenaline is produced)

Large dose (20-30) - this occurs in case of bone break if no painkillers, or at 51 HP missing at least. This gives you a painkiller between tramadol and paracetamol, and if you don't have 15 metanephrine accumulated, also gives you hyperzine

Critical dose: if you took 81 damage, or broke a bone and got 30 damage or variations, you are injected a flat 35 units of epinephrine. This triggers the strongest version, giving you tramadol and hyperzine until metanephrine hits 30 units. At this point, you accumulate metabolite at double rate.

For metabolites, until 15 is accumulated no bad things happen.
Between 15 and 25, you take a minor aim debuff and a warning
Between 26 and 35, your aim is crippled

Over 36, you start to roll for dizziness (cosmetic, 1/3 chance), losebreath for 2 ticks (1/3 chance), confusion for 2 ticks (1/3 chance), 5 brute (10% chance) and vomitting (0.25% chance)

Over 61 the side effects get far worse
you lose breath for 8 ticks (30% chance), get confusion for 10 ticks (30% chance), take 15 brute (20% chance), vomit (5% chance), fall down (5% chance)

Runa-Dacino:
I've added a check to prevent adrenochrome farming by printing sleeves and beating them up. 

Since it only affects carbon/human , people won't be able to farm monkeys either.

Runa-Dacino:
Made it so if you rest/sleep/get vored/sit down with less than 1 metanephrine in your system, all modifiers are cleared alongside the "broken bone boost" flag.

Runa-Dacino:
New change: You cannot combine neural hypersensitivity with heightened adrenal response.

Combining them would make neural hypersensitivity into a bonus trait.

I think grit (need 20% more damage to trigger stages) is fine, since it essentially inhibits the adrenaline response.

Whereas, neural hypersensitivity allows triggering the strongest adrenaline response with just 40 damage which would be overpowered, as strongest response's goals are to get you moving and running away from potential instakill attacks.

Runa-Dacino:
Inaprovaline can now prevent epinephrine overdose for critical patients from building up before metanephrine can start inhibiting production.

Furthermore,  made the message for clearing unsteady state clear after being jittery for better clarity when resting.

Finally, made sure reagent_state is now liquid.

TankTheBirb:
Feedback based on most recent version:
- For some reason, adding the trait to a character does not appear to activate the fight_or_flight var, you have to activate that manually by changing the false state to true. That may be a bug. I also may be dumb.
- This works pretty well for "run tf away" but if you stay in combat or ignore the reagents in you, you're gonna suffer at least a little bit. This is good.
- I think the telegraphing of when you're under the effects of epi or meta should give some sort of chat message. "You feel fatigued" when Meta stacks up high, and "you feel your heart rate return to normal" for epi being out of your system may be good feedback to players who just got damaged as to what is going on. Right now it's sorta hard to tell without a health analyzer, and I'm not sure "you need to rest" will come as obvious to people as it should.

Things I have not tested yet:
- Effect on bone breaks (I am testing organically rather than forcing test cases and haven't managed to break one yet)
- Effect on teshari/low endurance species
- Effect when actually fully kitted out for combat
- Effects in prolonged fights

TankTheBirb:
Actually one additional feedback, I think the painkiller effects may be too strong, as I basically never had to use the tramadol I brought with me for testing. I'd reduce the effects of pain killing to like half of where they are to keep painkillers relevant.

PR #13696, opened 2022-09-10
Adds personal shield generators. [READY FOR MERGE]

Cameron653:
image

https://user-images.githubusercontent.com/15969779/190929909-40fe3acf-f5cb-4e5c-9128-3527e17c12d0.mp4

# CREDIT TO Identity Crisis#2599 ON DISCORD FOR THE SHIELD GENERATOR SPRITES!

## MAJOR:
- Adds 'Personal Shield Generators' (Mostly referred to as PSG for the rest of this PR.)
- Adds the ability for energy based modifiers, so further energy based items/clothing can add modifiers.
- Adds better fine-tuning effects for modifiers.

## MINOR:
- Adds coloration to modifiers
- Adds ability for modifiers to choose from a custom _vr file.
- Fixes a bug where modifiers didn't modify embed chance.

#### THE PSG
The PSG is a belt or backpack based item. They can be fitted with device cells or can be map spawned with a 'loaded' cell.
- The shield is recolorable
- It can come with or without a simple egun built in. Backpack units by default have the egun.
- Is completely variable, allowing for setting:
1. The 'hit-cost' (bonus charge used for blocking/reducing an effect)
2. The 'damage_cost' (How much charge is used per damage taken? (Ex; 30 damage at 25 damage_cost = 750 charge used))
3. The 'modifier_type' (Which modifier it chooses from. A lot of variant modifiers were made for a plethora or uses.)
4. The gun type. Must be coded in if anything more complex than a simple pistol is to be added, however.
5. The cell it comes loaded with.
6. The 'generator_active_cost' which is how much charge it uses per tick of being on. Set to 10 by default.

That's all for the PSG. The code itself is complex but the idea behind it is simple. Kind of.

#### CELLS
- The PSG can be loaded with a recharging cell or a normal cell.
- Cells current take 60 seconds from the last time the charge was used to start charging and then take 30 seconds afterwards to charge the cell back to full.
- Belt versions have 2400 charge. Backpack versions have 4800.
- Added an upgraded variant that is not in use. (30 second recharge delay, 15 second recharge)

#### VARIANTS
A variety of variants have been coded in. Some are adminspawn/unused while others are meant for use. 
ALL SHIELDS UNLESS OTHERWISE NOTED ARE 100% WEAK TO ANY ELECTRIC ATTACKS, MEANING DOUBLED DAMAGE.
If shot with an electric halloss damage weapon with the mining variant (stun revolver, 70 halloss normally for example,) the user will take 280 agony damage pre-armor.

ALL BELTS HAVE CELLS OF 2400. ALL BACKPACKS HAVE CELLS OF 4800
CURRENTLY, EVERY UNIT TAKES 25 CHARGE PER 1 DAMAGE TAKEN. (This belt units can take 96 damage before breaking.)
Additionally, every unit uses 10 charge per second as well. (With exception of adminbus and parry.)

MINING BELTS:
- Mining PSG. Provides 30% to 15% resistance to brute&burn along with a 25% disable duration reduction. Has -50% halloss resistance (50% extra halloss damage.)
- Strong Mining PSG. Caused by inserting an upgrade chip into the PSG. Provides 45% to 25% resistance to brute & burn, along with a 50% disable duration reduction. Has -100% halloss resistance (100% extra halloss damage.)

SECURITY:
- Backpack PSG: Provides 50% to 10% resistance to brute & burn & halloss, along with 25% disable duration reduction.
- Security Belt PSG: Provides 25% to 5% resistance to brute & burn & halloss.

MISC:
- Bruteburn. Scaling brute&burn resistance from 100% to 0% resistance depending on cell charge. Unused.
- Biohazard: Provides 100 to 75% resistance to toxin, oxy, and cloneloss damage types. Unused.
- Adminbus: 100% resistance to everything. Adminspawn only.
- Broken: Doubles brute & burn damage taken. Unused.
- Inverted: Like bruteburn but opposite. 0% to 100% resistance depending on cell charge. Lower charge = more resistance. Unused.
- Parry: Provides complete immunity for the tick that the shield is activated. Only lasts for that tick. Unused.

## SIDE NOTES:
- The idea (currently) is that mining will start with one belt and security will start with one backpack(in the armory). (This has been discussed frequently and a 'slow rollout' seems more appropriate. Once we see how the shield generators are working, we can start expanding their numbers/ways to acquire them.)
- Mining is technically able to get more of them, but it is a 1 in a 100 chance from abandoned mining crates, which are quite rare to find and then they have to get that 1% chance on top of that.

# TODO:
- [x] Test changing cells and ensure it works.
- [x] Finetune numbers. Make the backpack shield generator stronger and the belt one weaker. (Partially done. Cells need to be adjusted.)
- [x] SPRITES!!!! Currently uses the chronomancer backpack, but better sprites would be nice! Not essential, but wanted!
- [x] Make it so the mining belt can ACTUALLY be upgraded.
- [x] Make it so rigged cells don't immediately go BOOM.
- [x] Fix bug that makes it so removing the cell mid-use doesn't make it infinite.
- [x] Add sounds!


Cameron653:
Suggestions are greatly welcome!

How are the current numbers for the current 'base' backpack
How should the power cell be modified, if at all?
How effective should the 'belt' version be? (In terms of reduction/charge/etc)


What should the mining shield generator do?
What should the security shield generator do? (Should it be an armory only thing? How many should be in the armory?)
What should a research shield generator do? What mats should it take
Should shield generators be upgradable via research parts such as manipulators? (I don't personally like this but it was suggested)

InterroLouis:
> Suggestions are greatly welcome!
> 
> How are the current numbers for the current 'base' backpack How should the power cell be modified, if at all? How effective should the 'belt' version be? (In terms of reduction/charge/etc)
> 
> What should the mining shield generator do? What should the security shield generator do? (Should it be an armory only thing? How many should be in the armory?) What should a research shield generator do? What mats should it take Should shield generators be upgradable via research parts such as manipulators? (I don't personally like this but it was suggested)

No comment on numbers because I want to use this and so my opinion on how stronk they should be is biased. However...the mining shield generator should have some radiation blocking power, simply because artifacts. Don't know about the security one, idk, melee focus maybe? Since they'll most often have to deal with any carp migrations. 
As for the science one, maybe parts? Perhaps not stock parts since you don't like them, but shield generator parts. Give it a space limit and have each part take up a certain amount of space. Have it need one of each part type...shield emitter, shield capacitor, and power source. So then you could have a better emitter to get perhaps better power use per time you're hit, capacitor for more capacity, power source for better charging...but you would have to balance them out with the available part space within the device itself. So if you make one thing better than just the normal, you would have to put a less good version of another part in to make everything fit.

Cameron653:
- Changes how it mitigates damage.
- It now only affects external attacks (punches, mobs, items, guns, etc.)
- Now is PRE armor mitigation! https://i.imgur.com/CXx82J5.png
- Adds variants. Allows mobs to use the new variants, but not the pack itself.
- Adds an extended description and fluff text.
- Adds ability to have both min&max for each variable.

I still need to make them have specialized recharging batteries (especially for the belt unit), but currently the normal recharging batteries are fine.

Cameron653:
- Adds more variants
- Adds personalized power cells for the shield generator.
- Fixes a mixup of words.
- Rebalances some of the modifiers to make it so they ramp up slower & have at least some effect when nearing the end of the charge.
- Fixes a dividing by 0 bug on specialty shield generators, such as the 'parry' one.
- Adds a huge text wall to explain this stuff to future people.

- Power cells now take longer before starting to charge and now charge slower by default.

Cameron653:
- Adds ability for modifiers to use a custom _vr sprite file.
- Changes the appearance of the sprite used for the shield overlay.
- Adds ability to color the shield sprite via a multitool.
- More fiddling with numbers...
- Allows rechargers to recharge them to allow for non-recharging variants to recharge.
- Gives shields a default 'cyan' color.
- Adds a mining disk to the equipment vendor to upgrade it.
- Gave the mining variant and the belt itself new sprites!

Cameron653:
Renamed the 'energy_modifier' to 'damage_cost" for clarity as to what the var actually does.
Added a security version. It's a codersprite.

Novacat:
To start with: These are going to be very beneficial to the game, since one of the balancing issues we have had are carbonmobs are exceptionally fragile.

The numbers has to be very carefully thought out however. Especially since if the numbers start off overtuned at first than nerfing later will result in people throwing a fit.

I will have to put some more thought into these later.

Cameron653:
> To start with: These are going to be very beneficial to the game, since one of the balancing issues we have had are carbonmobs are exceptionally fragile.
> 
> The numbers has to be very carefully thought out however. Especially since if the numbers start off overtuned at first than nerfing later will result in people throwing a fit.
> 
> I will have to put some more thought into these later.

I completely agree. The whole reason I started to make these is I've constantly heard comments about how melee combat is difficult since mobs shred through people and can make a single hit turn into a long affair of fixing bones, fixing internal damage, fixing space suit punctures (if those even happen anymore?) and other unpleasant side effects. For miners, this can mean 30min-1hr in medical away from their job.

Meanwhile the mobs just have...Damage. That's it. No fractures, no slowdown due to damage, nothing.

When the adrenaline PR is merged, it'll deal with some of the common combat complaints (unable to run from mobs due to slowdown) but the damage issue and its consequences are still the same.



Someone mentioned shield generators that you'd fight, take a few hits, then duck off to let your shield generator recharge before going back at it. (Scree IIRC)

I went ahead and combined both of these into one idea. Something that'll make you sturdier in combat (melee & ranged) against mobs while rewarding hit&run play so you either chew through batteries or you have to take time in between fights to rest up if you want your shield at full strength. 

When it comes to PvE, electric type attacking mobs will shred through armor. Doubly so if someone tries to use a mining-specific shield generator to fight them. Using a mining shield against a yellow slime? You'll only make that mistake once!



I've been going back and forth with the numbers over and over and over again, as you can most likely see with the commit history here.

It's _exceptionally_ hard to figure out just _what_ the right numbers are, since there's so many working variables. I haven't even _touched_ the damage_cost variable for any shield generator so far (leaving it at base 25, so 40 damage per 1000 charge), but I've made these things completely able to be modified in every aspect, from their special effects to the damage they can take to the amount of effects they can block, so it's able to be finetuned to a T.

I'd gladly welcome suggestions for the numbers.
My current 'vision' is:
Personal Shield Generators are rare/valuable items/resources. Mining gets some since they're constantly dealing with hostile fauna, but outside of that they should be a "If you see this, shit is going down." type of item.
To start with: Security has 2 backpacks, Mining has two belts.

- Security has two backpack PSGs in the armory. The numbers&variables are up in the air, but these are meant to be more or less 'Sacrificing firepower/utility for tankiness.' 
Instead of being able to carry 4 guns and who knows what else in your backpack, the backpack slot is relegated to a shield that keeps you alive for longer and you have a simple energy pistol that can be used in a pinch but is a bad idea as it'll drain your (relatively low) battery.
There ARE belt versions for security, but...Security would just be able to always wear them as the belt doesn't hold too much, meaning it'd be an everyday item, which these aren't really supposed to be.

- Mining has two belt PSGs and can upgrade these via mining points, but at no point should there be more than two. 
Maybe they can make them more specialized for brute, or more specialized for burn, or less resistances but less knockdown, or just make it better in general (Current upgrade disk makes it better in general.) 
Not too sure yet, but I kinda want to make it like the kinetic accelerator where you can upgrade it in certain ways.

- Research is...Off the tables for now. These are too new to be added to research and could be balance breaking during PvE events. Let's see how it's received by mining & security first and how effective/ineffective they are.

- Exploration is soon to be scrapped, so not a concern.

- Medical could have a CMO with the 'biohazard' variant in their locker like how the RD gets a bluespace harpoon. It's so situational that it'll rarely come into play but could be a neat little toy in their arsenal that could come up sometime in case xenos come on board, since it has toxins resistance.

And that's all the departments.

Cameron653:
Anyway, this is [technically] ready for merge. Going to see what issues I have to fix between main branch&this branch.

The final step left is to edit the numbers and then in a separate map PR, put shield gens on the actual map.

Cameron653:
Huh no conflicts with the base branch. Neat.

Cameron653:
The numbers on this are more or less fine, at least for a test run. Going to go ahead and see how it works out, along with giving security one backpack in the armory along with mining one belt.

If all goes well, can make them accessible via other means OR up the amount placed on the map.

Currently just waiting on sounds for the shield turning on/off. Should be good to go after that!

Cameron653:
Added a 1 in 100 chance that abandoned mining crates will have a mining shield generator.

Makes it so if a miner is ultra lucky, they can get one for themselves if the map-spawn one is being used.

Cameron653:
Added sounds! The saberon/saberoff sounded like a good 'powering on & powering off' sound, so I used those for now!

The sounds can easily be swapped in the future, meaning if I am given more appropriate/fitting/just plain better sounds, I can use those!

I can now at long last say this is
#ready for merge.

PR #13707, opened 2022-09-10
[Fixed Mirror] Removes dnaswitch artifact effect

Heroman3003:
Original PR's description (#12941):
It's outright broken for FBPs. The mutations of DNA2 are largely unsupported/legacy, and the more we cut out things that tie into DNA2, the more likely we are to actually be able to eventually cut out that tumor entirely to actually replace it with a new system.


Heroman3003:
...gee byond.

PR #13716, opened 2022-09-11
Enables mobs to be able to be drop_preds by default. Also fixes prefs.

Cameron653:
Because if you're thrown into a mob/stumble into a mob/slip into a mob, it probably wouldn't think twice about gulping you down. If this is an issue I can just remove it and leave the bugfix in, but given all this does is allow people to yeet themselves/slip/stumble into mobs if prefs align.

Also makes it so throwvore respects mobvore prefs.


Cameron653:
I failed with the web client, let me open this on an actual branch.

PR #13725, opened 2022-09-12
Fixes bottle throwing

Heroman3003:
Now break
throwing is terrible


Cameron653 (approved):
At last.
I can fill a bottle with poly and yeet it at someone's face.

PR #13732, opened 2022-09-14
Adjusts emagged booze

Heroman3003:
Renames Service borg's special poison emag booze and changes description to better explain what it actually is

Makes pounce module be Boozehound's proper emag tool (losing poison booze in process as borgs currently dont support more than one emag tool, and poison booze was overriding pounce)

### DIFF
--- a/code/modules/mob/living/silicon/robot/robot_modules/station.dm
+++ b/code/modules/mob/living/silicon/robot/robot_modules/station.dm
@@ -669,7 +669,8 @@ var/global/list/robot_modules = list(
 	src.emag.reagents = R
 	R.my_atom = src.emag
 	R.add_reagent("beer2", 50)
-	src.emag.name = "Mickey Finn's Special Brew"
+	src.emag.name = "Auntie Hong's Final Sip"
+	src.emag.desc = "A bottle of very special mix of alcohol and poison. Some may argue that there's alcohol to die for, but Auntie Hong took it to next level."
 
 /obj/item/weapon/robot_module/robot/clerical/general
 	name = "clerical robot module"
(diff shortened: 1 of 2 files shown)


Cyantime:
What's the new beer name referencing? The emag beer is [disguised chloral](https://en.wiktionary.org/wiki/Mickey_Finn), not just outright poison.

Heroman3003:
> What's the new beer name referencing? The emag beer is [disguised chloral](https://en.wiktionary.org/wiki/Mickey_Finn), not just outright poison.

Never heard of original reference. New name is just something I made up on spot to be distinctly different from old name. Also because its administered via bottle, itll pretty much always be in doses that actually do toxins, not just knockout. Regardless, main intent is to make it BLATANTLY OBVIOUS that its not actually a normal alcohol, but something dangerous, because I doubt vast majority of people playing ever heard of whoever Mickey Finn is, and would assume its just another cool drink name like Tungushka Triple Distilled for Vodka. In fact an incident like that is reason for a rename.

Cyantime:
The phrase isn't that obscure, I don't think? It's been used [forever](https://en.wikipedia.org/wiki/Mickey_Finn_(drugs)) and is explicitly what the emagged beer is about. All the emag things are designed to mangle people and if service borgs are getting emagged and blindly feeding people their new stuff then that's an issue with the borg player.

Heroman3003:
> The phrase isn't that obscure, I don't think? It's been used [forever](https://en.wikipedia.org/wiki/Mickey_Finn_(drugs)) and is explicitly what the emagged beer is about. All the emag things are designed to mangle people and if service borgs are getting emagged and blindly feeding people their new stuff then that's an issue with the borg player.

Except emag is not the only way to get it. There's also an upgrade module that USED to be named 'illegal' but now has different, a lot 'softer' name in part because several modules (such as research, mining or any dogborgs) produce not griefy killing things, but either things practical to department or for scenes. Either way, I was asked to change it to be 100% obvious that it is poison, which I did, while keeping "haha fancy name on the bottle" theme going. I had no clue who Mickey Finn is, but also the player who got it and staff who asked me to change it had no clue wtf was it about either. So I am just gonna call it what I see it as - obscure reference that obfuscates what the drink is more than it explains.

PR #13733, opened 2022-09-14
Updates Xenoarch sprites with Polaris.

Cameron653:
We don't actually use the _vr file, so this is good.

Also fixes the bug of invisible artifacts.


Cameron653:
Honestly I should just copy & paste their xenoarcheology folder over to here to bring all the updates xenoarch has had since then, but...Eh. Later.

PR #13743, opened 2022-09-15
New->Init Pod Tweaks

KillianKirilenko:
Finally dug up those fixes I mentioned a little while back. This offloads the pod triggering the ghostprompt to be part of the event, rather than part of the pod being initialized and makes the poi_loader lateinit, so the linter *should* no longer get upset about there being a sleep in there.

Trigger chance is currently a prob but that can easily be removed to make it always alert ghosts if desired.

Tested and confirmed not to freeze local servers at least. I don't know if it'll actually still pass the linter, but it's worth a shot.

### DIFF
--- a/code/datums/ghost_query.dm
+++ b/code/datums/ghost_query.dm
@@ -20,10 +20,10 @@
 
 	// Then wait awhile.
 	while(!finished)
-		spawn(1 SECOND)
-			wait_time -= 1 SECOND
-			if(wait_time <= 0)
-				finished = TRUE
+		sleep(1 SECOND)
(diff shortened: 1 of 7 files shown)


Heroman3003:
The ghost notification should absolutely happen 100% of the time for events.

KillianKirilenko:
Okay, I see why that was. There was originally a notice as part of the `create_occupant` proc that would fire off a command announcement, but that seems to have been axed in #12061. I must've misinterpreted the vars being left behind.

Cleaned that up, and removed the now-redundant no_announce pod variants since they serve no purpose now.

PR #13757, opened 2022-09-18
Attempts to fix simplemob pref loading

Heroman3003:
Should have no practical changes, but also should fix runtime with simple mob /Login(). Why was it put where it was put anyway?


Very-Soft:
To be fair it probably was fine before I added prefs copying to Login

PR #13762, opened 2022-09-19
Fixes pakkun stuff

Heroman3003:
Fixes fire sound and allows escaping


Screemonster:
I've relayed the current state of the pakkun to a couple of people and they have, independently, suggested that the current "just eat people again immediately, no escape" implementation be saved for future use even if it isn't the default, just for... reasons.

Cameron653 (approved):
Looks good.

PR #13771, opened 2022-09-20
Removes hotfood vendors from station spawning

Heroman3003:
Its a vendor with poisoned food originally meant for spawning in pois with 'expired food' meme. It also randomly had carpotoxin aparently?! Basically makes it not appear in station spawning pool, and removes cartridges for it from cargo.

### DIFF
--- a/code/datums/supplypacks/vending_refills_vr.dm
+++ b/code/datums/supplypacks/vending_refills_vr.dm
@@ -18,11 +18,6 @@
 	name = "SweatMAX Vendor Refill Cartridge"
 	cost = 10
 
-/datum/supply_pack/vending_refills/hotfood
-	contains = list(/obj/item/weapon/refill_cartridge/autoname/food/hotfood)
-	name = "Hot Foods Vendor Refill Cartridge"
-	cost = 10
-
 /datum/supply_pack/vending_refills/weeb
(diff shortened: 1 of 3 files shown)


Cameron653:
_nooooo how am I supposed to get my carpotoxin and zombiepoweder now!!!_

Cameron653 (approved):

BonniePandora:
Isn't there a version of the hot food vendor that isn't like
Ancient & lethal? 
Why not swap this not-great one out for that instead, given it only really spawns at... the V2 surface outpost and the fuel depot, currently.

Heroman3003:
> Isn't there a version of the hot food vendor that isn't like Ancient & lethal? Why not swap this not-great one out for that instead, given it only really spawns at... the V2 surface outpost and the fuel depot, currently.



> Isn't there a version of the hot food vendor that isn't like Ancient & lethal? Why not swap this not-great one out for that instead, given it only really spawns at... the V2 surface outpost and the fuel depot, currently.

Those are event machines for use in specific places or events. Also its not 'same version' really, its jsut a machine full of fastfood for free.

Cafaling:
Kinda funny vendor, but probably for the best that it goes.

PR #13774, opened 2022-09-21
Adds a ton more altevian stuff

Heroman3003:
Adds 3 altevian guns, 2 energy and 1 ballistic.
Adds altevian armored spacesuit.
Adds 2 altevian melee weapons, regular and energy.
All of the above is adminspawn only atm.

Also adds 5 new jumpsuits in a selection in xenowear, only wearable by altevians.
And 1 new breath mask, also only wearable by altevians, available in xenowear.


code/modules/projectiles/guns/projectile/altevian_vr.dm
```diff
@@ -0,0 +1,39 @@
+/obj/item/weapon/gun/projectile/altevian
+	name = "Altevian Rivet Repeater"
+	desc = "An offensive designed by the altevians that is used for decompression and maximizes structural damage while also serving as a good method of personnel damage."
```
BonniePandora, line 3: I think you missed the word "weapon" near the start here.

code/modules/clothing/spacesuits/void/event_vr.dm
```diff
@@ -212,21 +212,49 @@
...
+
+	icon = 'icons/inventory/suit/item_vr_altevian.dmi'
+	default_worn_icon = 'icons/inventory/suit/mob_vr_altevian.dmi'
+	icon_state = "rig-heartbreaker"
+
+	armor = list("melee" = 90, "bullet" = 90, "laser" = 90, "energy" = 90, "bomb" = 90, "bio" = 100, "rad" = 80)
+
+	species_restricted = list(SPECIES_ALTEVIAN)
+	no_cycle = TRUE
+	slowdown = 2.5
```
Cameron653, line 247: In the future if this becomes an actual, obtainable item, either the armor values should come down some or the slowdown should go up some. Beacher for reference has slowdown 6 and the same armor values.

As an adminspawn item it's fine though.

Cameron653 (approved):
Everything else in here looks good.

code/modules/clothing/spacesuits/void/event_vr.dm
Heroman3003: Breacher is also unathi exclusive, who only get 1.5 slowdown from it due to inheirent trait. This suit is technically worse than breacher, both by not being a rig and by being exclusive to species that DOESNT reduce the slowdown.

Cameron653: Fair enough. Didn't realize that unathi have enough natural slowdown reduction to make it only 1.5

PR #13782, opened 2022-09-22
Buffs Vore

Very-Soft:
Adds the 'process' option to the list of things you can do to prey in your bellies!

This option will let you pick between digest and absorb, when you pick one, it lets the prey know what you are trying to do and has them confirm! If they confirm and are still in the belly, then that option will instantly happen to them.

The options available to you when you click it are dependent on the prey's prefs, so, if a prey has digest turned off, but absorb turned on, then it will only offer absorb! If they have both disabled, then it will tell you no!

This way if you want to gurg someone up real fast once your scene is done or whatever, you don't have to adjust all your gurg values or whatever.

This does also allow one to SELECTIVELY digest or absorb people even if they are in the same tummy, which I think is very radical.

### DIFF
--- a/code/modules/vore/eating/vorepanel_vr.dm
+++ b/code/modules/vore/eating/vorepanel_vr.dm
@@ -792,7 +792,7 @@
 						return
 					if(isliving(usr))
 						var/mob/living/l = usr
-						var/thismuch = ourtarget.health
+						var/thismuch = ourtarget.health + 100
 						if(ishuman(l))
 							var/mob/living/carbon/human/h = l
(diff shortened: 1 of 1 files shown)


ItsSelis (approved):
Looks good to me!
Nice work :P

Heroman3003:
This looks solid, however, one small thing missing is that nutrition normally acquired from the digestion process is lost in this option. Maybe, similarly to absorption calculating nutrition that absorption would take, have the 'remaining hp' be converted into nutrition before nuking the prey?

Cameron653 (approved):
finally

Vore 2

PR #13792, opened 2022-09-23
Fixes conk rete walls

Heroman3003:
Premapped ones that is. Fucky wucky.


Cameron653:
Ah-ha, I knew something would slip by.

Cameron653 (approved):

PR #13186, opened 2022-06-26
[MIRROR-FIX] "Nerfs" Hyperzine (Discussion Needed)

ItsSelis:
This actually buffs Hyperzine (increasing it from 1 to 1.2).

Additional Comment from original PR:
> Also makes psilocybin actually hallucinogenic and removes a very annoying hallucination interaction with airlocks.

This PR should close #12972

PR #13808, opened 2022-09-25
fixes psilocybin

Cameron653:
Does the bugfix that #13186 contains without adjusting hyperzine.

Adjusting hyperzine should be it's own individual PR.

I also want to fiddle with changelogs so this is the perfect opportunity

## changelog
🆑
fix: psilocybin now makes you hallucinate
/🆑


Cameron653:
Going to make this fix changelogs as well.

PR #13811, opened 2022-09-26
Fixes the changelog.

Cameron653:
- Fixes changelogs (It needed to be told which loader to use)
- Fixes instructions on how to generate changelogs
- Removes old changelog history (I attempted to preserve it and it would randomly nuke the changelog history, so instead we shall start fresh!)
- Makes changelogs point to Virgo's github, wiki, discord, etc instead of linking to Polaris.

Changelogs are an optional feature and easy to perform. 
I highly suggest them for new features/changes or critical bugfixes while small changes like behind-the-scenes bugfixes aren't really needed to be added to the changelog.

With this in mind, I'd also suggest AGGRESSIVE_CHANGELOG being turned on in the config.
This would make it so the first time someone logs in after a server update, they'd be given info on what changed!

How to create a changelog:

Step 1. Go to \html\changelogs in where the code is installed.
Step 2. Make a copy of example.yml
Step 3. Change the info inside of it. It walks you through the steps as to be foolproof.
image

Step 4. A dev will eventually run the 'makeChangelog.bat' when it's update time and it'll update the changelog with everyone's changes. (located in tools/GenerateChangelog)

If you don't want to do step 1-3, you can do the below as an example and it should automatically make a changelog when someone runs the .bat

### Changelog

🆑
rscadd: Changelogs added.
rscdel: Old changelogs deleted.
experiment: Testing out changelogs.
/🆑


Cameron653:
This is what the changelog currently looks like with this PR (YML aleady merged)
firefox_2022-09-26_00-21-41

ItsSelis:
Wasn't there an autochangelog?

And it is possible to preserve the changelog history. There is a script that I once repaired that put it in the right format.

Cameron653:
Closing this due to a better method that is going to be done in another PR.

PR #13815, opened 2022-09-27
Fixes and Improves the Changelogs

ItsSelis:
This PR is a merge of the changes made in #13811 by @Cameron653  as well as some that I had done in the past.
We decided to just open up another PR instead of dealing with merge conflicts that were caused (by both PRs editing the changelog history and such).

There are more script changes and cleanups done, as well as the modernization of the ui itself into an tgui interface (most of which were ported from tg due to the changelog system seemingly origination from there).

Afaik there is supposed to be a workflow that generates the changelogs, but there are also methods of making the automation part into VGS, so that has to be tested.

**Note:** The footer (shown in the 2nd screenshot under the "Failed to load" message) might have to be adjusted to show the right commit link and date for the license change to AGPLv3 and maybe missing credits.

Here are the screenshot examples of the new UI:
grafik

(It says failed to load since there are no changelogs, the old ones were moved into old_archive if anyone needs them. But they are not visible in-game unless moved to archive)
grafik

Note to future developers editing the example.yml:
You have to regenerate the MD5 hash and modify the ci script "validate_files.sh" that was modified in the pr
grafik

## Changelog
🆑
rscadd: Changelogs added.
rscdel: Old changelogs deleted.
experiment: Testing out changelogs.
/🆑


Cameron653:
For anyone looking at this, while it's true this does touch 155 files, these are more or less the only significantly changed files: https://i.imgur.com/amDT5g2.png

Seeing as the screenshots show it has been tested, this looks good to me.

Cameron653 (approved):
Looks good and given the fact it's tested, I have no worries about this.

I presume using 🆑 stuff /🆑 will still allow for PRs to auto-gen a changelog?

ItsSelis:
> 🆑 stuff /🆑

Only one way to find out

Cameron653:
closing and reopening because Travis shit itself

Cameron653:
Closing and reopening again because travis is not playing nice. Linter is coming up negative for some odd reason.

Cameron653:
Linter FUBAR and won't explain why

ItsSelis:
Due to #14661 no longer required (players do not see a ui iirc so it wouldn't make sense?)

KillianKirilenko:
> Due to #14661 no longer required (players do not see a ui iirc so it wouldn't make sense?)

Correct, we just point them to the (currently manually compiled) wiki changelog.

It would ultimately be up to staff to start harshly enforcing proper changelog creation before anything like this would be worth merging.

PR #13819, opened 2022-09-29
Complicates Pakkuns

Very-Soft:
Pakkuns now occasionally try to rest!

When they rest, they set their digest mode to unabsorb! So, if you end up absorbed you can still escape eventually.

Resting Pakkun also have their AI set to sleep, so they will not engage, so this may be able to be used to sneak past, or up to them!

Also makes it so that when you help click them while they are resting, they will wake up! and re-enable their AI, if they have one.

They also turn their digest mode back to the default setting when they stop resting.

All of this of course, only happens when they don't have a client.

### DIFF
--- a/code/modules/mob/living/simple_mob/subtypes/vore/pakkun.dm
+++ b/code/modules/mob/living/simple_mob/subtypes/vore/pakkun.dm
@@ -20,6 +20,7 @@
 	icon_dead = "pakkun-dead"
 	icon_living = "pakkun"
 	icon_state = "pakkun"
+	icon_rest = "pakkun-rest"
 	icon = 'icons/mob/vore.dmi'
 
 	faction = "pakkun"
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/simple_mob/subtypes/vore/pakkun.dm
```diff
@@ -46,8 +47,46 @@
...
+	vore_default_mode = DM_SELECT
 
 	var/extra_posessive = FALSE					// Enable if you want their tummy hugs to be inescapable
+	var/autorest_cooldown = 100
+
+/mob/living/simple_mob/vore/pakkun/Life()
+	. = ..()
+	if(autorest_cooldown)
+		autorest_cooldown --
+	else if(prob(5) && ai_holder.stance == STANCE_IDLE)
```
Screemonster, line 59: will this still trigger if there's a client? if it does then get_AI_stance() returns null if there's a client and ai_holder.stance otherwise.

PR #13820, opened 2022-09-29
Fixes build mode and makes it not terrible.

Cameron653:
- Makes Basic build mode warn you before you delete the ground and space everyone
- Makes Basic build mode turn the turf you deleted into the base turf for the area. If area base turf is not available, it goes with the base turf of the Z level.
- Makes Advanced build mode start with Null. No more spawning lockers and qdel()ing players!
- Fixes a longstanding bug where if you cancelled selection in build mode and then tried to select something from the list, it defaulted to a locker. Instead, if you cancel it, it defaults back to null and lists work like normal.

🆑
- bugfix: Fixes build mode. No more admins deleting or spacing you by mistake.

/🆑

### DIFF
--- a/code/modules/admin/verbs/buildmode.dm
+++ b/code/modules/admin/verbs/buildmode.dm
@@ -252,7 +252,7 @@
 			if(BUILDMODE_BASIC)
 
 				return 1
-			if(BUILDMODE_ADVANCED) //BUILDMODE-ADVANCED-DEBUGGING-COMMENT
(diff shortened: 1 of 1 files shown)


Aroliacue:
Very awesome bugfix, thanks Cam!

Cameron653:
This was merged in  PolarisSS13/Polaris#8719 and has a mirror that is #13833 

I'm just going to go ahead and merge this since it's already been reviewed and even merged by Polaris.

PR #13825, opened 2022-09-30
Sanitization

Cameron653:
- Makes RCON tags less exploitable.
- Sanitizes some more things so you can't put HTML into it.
- Utilizes TGUI better to show the user the max they can post.
- Limits some things like circuits so you can't insert 50000 characters.
- Allows badmins to send HTML in subtle messages but only if they do it properly.


Cameron653:
DNM while I add more fixes

PR #13828, opened 2022-10-01
Blueprint Revamp.

Cameron653:
dreamseeker_2022-10-01_01-57-27

## MAJOR:
- Completely revamps the blueprints.
- Ports /tg/ blueprints and bastardizes them.
- Finds how to fix lighting in an area This solves a longstanding build mode bug.
- Gives admins the ability to spawn a structure via build-mode with or without an area. With an area now has lighting.
- Gives engineers a 'writing blueprints' that are a weaker version of the CE's blueprints.
- Gives players the function to use a paper to draft up a new room in an airtight buildable zone. (Now you can make a hut in the wilderness and it'll actually work!)

### Blueprint functions:
- Create/modify an existing area (in a 3x3 square. Good for expanding an area to include walls or finetuning)
- Create new area or merge two areas. (Allows you to make a new area if the room is airtight or merge two areas together.)
- Change area name.
- Wire Legend. Allows you to see what wires do. (This has been made its own separate item but can be enabled via admin memery.)
- Allows you to swap between area/room mode like the old blueprint.
- Can come with charges or no charges. CE blueprints have no charges and are infinite. Normal engineer blueprints have charges and must be filled with the CE's blueprints if too many changes are made.
## MINOR:
- Adds a new global list "wire_name_directory" that is important for revealing wires.
- Gave vent pumps, vent scrubbers, and air alarms a new update_area proc that is involved when their area is updated due to a blueprints being used.
- Gave areas a 'setup' proc that is called when a new area is created via blueprints. Sets the area vars to what would be default (power off) Adds a new 'areasize' variable to areas that lets you know how large they are.
- Added procs to areas to check the power in an area and check change the name of an area properly.
- adds a new rect_turfs define. Unused for now.
- Fixes a bug with champagne by writing a new proc, 'get_current_area_turfs' which just checks the current area's turf.
- Fixes a bug that made it so areas made in space were permadark. Light now effects them if you make an area in it!
- Fixes a bug with buildmode that makes it so areas are now fully lit when placed down via build mode. No more forever-dark zones that light wouldn't touch.

### NOTE: I can't possibly know every area that should be blacklisted in the game. If you can think of an area that "You should never be able to do anything to" or "You should be able to build stuff in" then by all means, let me know in the comments!
The above is extremely important. Please give me info if I missed anything. This will work and people can't do too much horrible stuff with it barring combining areas into a new area or making a new area inside of an area if it isn't added to the blacklist, but it should still be done.

However, with how it currently is, it should prevent any major issues ATM as the important places that I could think of are blacklisted.

SCREENSHOTS:
The HUD itself:
dreamseeker_2022-09-30_21-02-33

Expanding an area (3x3):
dreamseeker_2022-09-30_21-15-42
image

Expanding an area (Merging):
dreamseeker_2022-09-30_21-55-41

dreamseeker_2022-09-30_21-56-38

What happens if you try to make a new area/expand an area into an area on the 'SPECIALS' list or an area that contains another area inside of the room that is on the 'SPECIALS' list:
image

What happens if you try to expand an area into space tiles: 
dreamseeker_2022-09-30_21-58-52

What happens if you try to make a new area based on a room in a non-airtight location:
image

However, you are able to make a new area and expand it outwards on normal turf even if the area doesn't have an atmosphere/fully boxed in via the 3x3 finetuning ability: (I see no reason to really remove this. You could just make an area and then knock down the walls if you really wanted to.)
image

You can define multiple areas seamlessly: 
dreamseeker_2022-09-30_22-03-13

image

Part one of the spaceship rework! (Extremely rough draft, only shown to show you can define an area and pilot it.)
dreamseeker_2022-09-30_22-08-27

dreamseeker_2022-09-30_22-09-11

A newly made area with working lighting: 
dreamseeker_2022-09-30_23-02-16

Making an area with paper:

Trying to make an area too large with paper:
dreamseeker_2022-10-01_23-24-12

NOTE: I can't possibly know every area that should be blacklisted in the game. If you can think of an area that "You should never be able to do anything to" or "You should be able to build stuff in" then by all means, let me know in the comments!

TODO:

- [x] Replace references to the blueprints with /obj/item/areaeditor/blueprints
- [x] Make the wire legend into its own item
- [x] Make 'notepads' for the engineers to use to expand areas and the like.
- [x] Make 'create an area' a verb anyone can use with a piece of paper. No modification, just creation. So if you're in the wilderness or space you can make a house or something!
- [x] Something else I'm forgetting idk.


Cameron653:
There we go. After the absolute insanity that is this PR, it is ready to good and ready for merge.

Took an absolute insane amount of time but finally got it fully functioning and as bug free as I could.

The biggest issue is going to be any new areas that are added that should/shouldn't be able to be built in/expanded need to be added to the global list, but it's extremely easy to do.

PR #13838, opened 2022-10-02
[MIRROR] Adds SSTyping

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/8722
--------------------
Alternative subsystem based typing indicator behavior with protection against multiple samples in flight. Takes the input bar into account, uses vis contents instead of image shenanigans, and gracefully supports species changes.

*might* break mob z mimic typing indicators; can't trivially test that alone.

closes #8707

also adds XMOUSE_OPACITY_* defines since byond doesn't ship MOUSE_OPACITY_* itself.


Cameron653:
Just was poking around with this to see what it does. We have Runechat so this seems irrelevant.

PR #13840, opened 2022-10-02
Smole vending tweak

Very-Soft:
Swaps the asian food vendor for a random vendor spawner, and adds altevian food vendor to the random vendor spawner


Heroman3003:
I was fairly certain I added it to random vendor selection already... Huh, go fig Well, good to have it more then.

Cameron653 (approved):

Very-Soft:
> I was fairly certain I added it to random vendor selection already... Huh, go fig Well, good to have it more then.

u put it only in the food one, but not the all one, I just added it to all

Heroman3003:
> > I was fairly certain I added it to random vendor selection already... Huh, go fig Well, good to have it more then.
> 
> u put it only in the food one, but not the all one, I just added it to all

I see! Good catch then.

PR #13843, opened 2022-10-02
Makes xenoarch triggers a little bit less teidous

Cameron653:
dreamseeker_2022-10-02_18-58-34

- Makes artifacts accept paths instead of just IDs of certain chemicals.

There's other changes I have as well but I'd rather just slowly roll them out one by one. Lets me test them more thoroughly like that.


Cameron653:
DNM I need to make this check for 'toxin' in the name/id of a chemical instead of lists.

Cameron653:
Closing this, going to be on the backburner.

PR #13844, opened 2022-10-03
Fixes parse-srcset checksum

ItsSelis:
The old checksum is invalid, the now one now works fine.

The bundle still compiles and the UIs work:
grafik
grafik


Cameron653 (approved):
This needs to be manually merged by someone with greater permissions than I have as it failing the CI test and can't be merged manually due to the master branch being protected.

ItsSelis:
This problem is no longer reproduce able for some reason (suggesting that the checksum error was either fixed or that the cache of it was very very broken). This PR for now is no longer needed.

PR #13845, opened 2022-10-03
New nice ticket(panel)

ItsSelis:
This PR adds an tgui version of the window of a ticket being displayed.
This means that the data and stuff inside live-updates (no more refresh button).

Of course there is still a button to switch to the old byond ui if wanted (which also automatically opens in case the tgui window crashes for some reason).


Cameron653:
Has a fallback. 

And this is what it looks like 
image

Cameron653 (approved):
I was observing while this was written and can confirm that this works properly and has the proper fallbacks in case of TGUI failure.

PR #13859, opened 2022-10-06
ALL TERRAIN WINGS

Cameron653:
- Makes flight let you fly over snow
- Makes flight let you fly over water
- Makes flight let you fly over leeches
- Makes flight let you fly off cliffs
- Makes flight let you fly over mines. (I'm not sure. This shouldn't really ever happen anyways since mines aren't generally in use. If it's a problem I can just remove it.)

### DIFF
--- a/code/game/objects/effects/mines.dm
+++ b/code/game/objects/effects/mines.dm
@@ -96,7 +96,8 @@
 		explode(M)
 
 	if(istype(M, /mob/living/))
-		if(!M.hovering)
+		var/mob/living/mob = M
+		if(!mob.hovering || !mob.flying)
 			explode(M)
 
 /obj/effect/mine/attackby(obj/item/W as obj, mob/living/user as mob)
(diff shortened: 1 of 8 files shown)


Skysteak:
I will say that this is coming from someone who hasn't used Flight before, so my input may not be as solid as those who have!

It all looks solid albeit the only point I'll raise is that Winged Flight was made a zero cost trait due its limitations iirc. Whilst it still occupies a trait slot, this _would_ provide a zero cost way of bypassing cliffs; the Soft Fall trait also providing this boon albeit it comes at the cost of 1 point. 

I admit it's not something I have a solid solution for, as this feels like an awkward midpoint between providing a (albeit extremely niche) mechanical advantage, but not really worth justifying being its own trait. However I thought it'd be something to keep in mind regardless! 

Side Note: This is blatantly a suggestion but it'd be nice if there was some sort of 'slow mode' one could engage to press up against objects without slamming into them, albeit that may be out of the scope of the PR!

Cameron653:
The cliffs from what I can obtain are an edge case and not really incorporated into the game other than a few PoIs. If cliffs were more common, then I feel like it'd be a bigger question of if it's fine or not.

It shouldn't really come into play (especially seeing as someone would have to toggle winged-flight on, move down, then toggle it back off) which if they're using it to get away from something up there would take a moment, vs soft-fall where they could just run off it and be fine.

PR #13860, opened 2022-10-06
Gird check fixes

Heroman3003:
Fixes grid check affecting subgrids that are properly separated

Makes grid check not occur at all when your TEG is not producing half its power or not connected to a powernet.

Also increases chance of announcement up to 80 from 30. Mostly for fluff, since with how rare it is already; announcement with that low chance plays even rarer. And effect of the event is obvious either way, really.

### DIFF
--- a/code/modules/power/generator.dm
+++ b/code/modules/power/generator.dm
@@ -261,11 +261,11 @@ GLOBAL_LIST_EMPTY(all_turbines)
 
 	src.set_dir(turn(src.dir, 90))
 
-/obj/machinery/power/generator/power_spike()
-//	if(!effective_gen >= max_power / 2 && powernet) // Don't make a spike if we're not making a whole lot of power.
+/obj/machinery/power/generator/power_spike(var/announce_prob = 30)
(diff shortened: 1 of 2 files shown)


code/modules/power/generator.dm
```diff
@@ -261,19 +261,19 @@ GLOBAL_LIST_EMPTY(all_turbines)
 
 	src.set_dir(turn(src.dir, 90))
 
-/obj/machinery/power/generator/power_spike()
-//	if(!effective_gen >= max_power / 2 && powernet) // Don't make a spike if we're not making a whole lot of power.
+/obj/machinery/power/generator/power_spike(var/announce_prob = 30)
+//	if(!(effective_gen >= max_power / 2 && powernet)) // Don't make a spike if we're not making a whole lot of power.
```
Cameron653, line 265: needs fixie

Cameron653 (approved):

Darlantanis:
Petition to rename grid check to gird check

Heroman3003:
> Petition to rename grid check to gird check

girth check

PR #13864, opened 2022-10-07
Adjusts vulpkanin heat/cold handling. Makes them mid-way between human and taj.

Runa-Dacino:
Getting cold-risk overlay at just 5 celsius is annoying. One would think a furry creature would be able to withstand temperatures until like -20.

With these changes, Vulpkanin will feel comfortable up until the temperature gets below -10 celsius. At which point, they get periodic warnings to get out. At -30 celsius, they start taking damage. 

Unlike Tajara, who specifically live in tundra conditions, Vulpkanin do not retain cold resistance in extreme conditions. Meaning, while tajara can survive - 70 celsius, vulpkanin start suffering as hard as humans do. Tajaran tresholds are -70 for first damage, -130 for second tier and -190 for third tier. 

To balance the increased cold resistance, Vulpkanin become uncomfortable in temperatures over 21.85 celsius (chosen to avoid spamming them, copied from taj). Thematically, vulpkanin being comfortable in cold is a nice mechanical contrast to zorren, who live in deserts. I won't touch them in this PR, but I think zorren could use a bit of nudge in opposite direction since they're supposed to be desert-adapted.

They start taking heat damage half-way between human and tajaran values. Across the board, it's reduced.

Also fixes vulpkanin somehow sweating in heat.

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/station_vr.dm
+++ b/code/modules/mob/living/carbon/human/species/station/station_vr.dm
@@ -235,6 +235,27 @@
 	wikilink="https://wiki.vore-station.net/Backstory#Vulpkanin"
 
 	catalogue_data = list(/datum/category_item/catalogue/fauna/vulpkanin)
+	
+	//Furry fox-like animals shouldn't start freezing at 5 degrees celsius.
+	//Minor cold is resisted, but not severe frost.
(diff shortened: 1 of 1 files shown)


Cameron653 (commented):
Seems good to me. Unless anyone has any objections, I see no reason not to include this in the next update.

PR #13867, opened 2022-10-07
[MIRROR] A -ton of Flags [READY FOR MERGE]

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/8721
--------------------
SCG, VGA, NT, Five Arrows, Both Almachs, Vystholm, and the four big political parties.
For mapping/admin use currently. Base sprite and basis of code from Aurora.

edit: and pirate flags.


Cameron653 (approved):
The flags are probably going to be unused, but the changing of things like the vague 'directions_sci' icon state for the toxins sign to be the proper 'directions_toxins' along with all the others is nice.

I dunno if we even use them, but I doublechecked the DMI and we have them, so we might as well have the icon_states updated.

Additionally, I added the chemdiamond and the other missing Polaris sprites from our decals.dmi into it.

PR #13872, opened 2022-10-07
[MIRROR] Shuttle Refit

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/8718
--------------------
Generally does away with the old shuttle wall and floor tiles which I think may be some of the oldest and most brightest sprites in the game, and makes shuttles look more nicer.
Occasionally shuttle corner-accents fail to pick up the colour of the surrounding hull, which is generally an issue with the decor piece as previously implemented, even with tweaks. [If anybody asks it got into a fender bender.](https://www.ebaymotorsblog.com/motors/blog/wp-content/uploads/2015/10/plastidip-3-800.jpg)

Includes a bunch of CM wall pieces because they look nice on ships in the same vein as the eris floors.
A couple of shuttles use elements of those, entirely those, or just a mix of the existing hull-wall pieces.
Also a bunch of floor decal stuff from CM and downstream, incorporated on shuttles and a couple of related spots.

Cargo:

Merc:

Departures:

Arrivals (Not pic of final version but close enough):

Also changed are various centcom shuttles, the Major Bill's PoI, and escape pods.


Cameron653:
Cynosure stuff and we can't simply port over the sprites.

Manually port this if desired.

PR #13881, opened 2022-10-08
[MIRROR] Ports Diagonal Movement [READY FOR MERGE]

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/8199
--------------------
https://i.gyazo.com/d74fba16abce0589d02408333528407c.mp4

for years, man has gone sideways but now
man go diagonal

https://github.com/VOREStation/VOREStation/pull/10928
https://github.com/VOREStation/VOREStation/pull/10975

:cl:
rscadd - Implemented diagonal movement when two directional keys are pressed simultaneously.
/:cl:


Heroman3003:
We literally had it for a long time. If nothing interesting is in the files, might just close

Cameron653:
Going to clean this PR up to see if it has anything we'd want.

code/modules/mob/living/living_movement.dm
```diff
@@ -202,7 +204,8 @@ default behaviour is:
 		if(!T2) // Map edge
 			now_pushing = 0
 			return
-		var/move_time = movement_delay(loc, t)
+
+		var/move_time = movement_delay(loc, t) * SQRT_TWO
```
Cameron653, line 208: So if you're pushing something, the move_time is ~40% longer?

interface/skin.dmf
```diff
@@ -176,6 +176,9 @@ macro "borghotkeymode"
 	elem
 		name = "S+UP"
 		command = "KeyUp S"
+	elem 
+		name = "CTRL+S+REP"
```
Cameron653, line 180: I'm unsure what REP is on the keyboard.

code/modules/multiz/movement.dm
```diff
@@ -355,6 +355,17 @@
 /obj/structure/catwalk/CheckFall(var/atom/movable/falling_atom)
 	return falling_atom.fall_impact(src)
 
+/atom/movable/proc/can_fall_to(turf/landing)
```
Cameron653, line 358: We already have this. https://github.com/VOREStation/VOREStation/blob/1331236464c830b93f09b3659c40d171d712731b/code/modules/multiz/movement.dm#L326-L337

code/modules/mob/living/living_defines.dm
```diff
@@ -1,6 +1,8 @@
 /mob/living
 	see_invisible = SEE_INVISIBLE_LIVING
 
+	appearance_flags = TILE_BOUND|PIXEL_SCALE|KEEP_TOGETHER|LONG_GLIDE
```
Cameron653, line 4: These are already applied on /atom/movable it'd seem?  https://github.com/VOREStation/VOREStation/blob/5195f715452d7a92f2b911f8217343a856f6ed13/code/game/atoms_movable.dm#L3

code/modules/multiz/movement.dm
Cameron653: Removed.

Cameron653 (commented):
All in all, the only 'major' change here appears to be that pushing an object makes you move 40% slower.

There's a few additions that allow use of the "REP" key, but I don't know wtf the REP key is.
It also standardizes a few things to be written in more 'common' ways. 
Ex: 
Instead of "IF(THING) XYZ += THING"
It does:
"If(THING)
XYZ += THING"

Also makes the following applied to mob/living instead of just being applied to /atom/movable:
appearance_flags = TILE_BOUND|PIXEL_SCALE|KEEP_TOGETHER|LONG_GLIDE

Adds a few procs like a sound multiplier that returns the inverse of ear deafness and some reagent proc as well.

I guess this is good to merge?

Heroman3003:
Honestly, I dont think we should have any of that. Pushing slowdown is dumb, REP is how things work already without it, standartization of ifs is nice, but ultimately purely visual, flags work good as is and we kinda may have things break if that changes and some random procs about deafness and reagents have no place in the movement PR at all

Cameron653:
Let me see about cleaning this PR up a bit more to just keep the IF standardization. Might as well make the PR into something useful, at least.

Cameron653 (approved):
Major (In terms of this PR this is the only actual change): 
The if standardization

Minor:
A DEBUG_INPUT being moved to the proper place
A VorestationEdit comment removal (Now in base Polaris)
The sqrt define (I'm sure _some_ engineering nerd can find a use of it)
Comments of things like key_down will show up in VSC now.

PR #13888, opened 2022-10-09
Makes PAIs able to continue healing even after nutrition depletion

### LINKED ISSUE #13885: [GENERAL] pAI vore belly 'heal mode' does not function.
When attempting to heal injured crew using a pAI belly, there appears to be no effect whatsoever. I believe this may have something to do with pAIs having no 'nutrition' stat.

1) Eat someone
2) They get healed over time

#### What actually happened
1) Eat someone
2) They stay unconscious and don't get healed at all.

#### Steps to reproduce
- Spawn in as pAI
- Create a healbelly
- Eat someone that has injuries
- Watch as they don't get healed

Cameron653:
See title.
- Fixes a bug where PAIs would run out of nutrition and not be able to heal any further.

PAIs don't use nutrition and start with 400 on spawn. After healbellying for enough time, they'd run out and not have any avenue to heal any further. Fixes  #13885

### DIFF
--- a/code/modules/vore/eating/bellymodes_datum_vr.dm
+++ b/code/modules/vore/eating/bellymodes_datum_vr.dm
@@ -131,7 +131,7 @@ GLOBAL_LIST_INIT(digest_modes, list())
 	if(L.stat == DEAD)
 		return null // Can't heal the dead with healbelly
 	var/mob/living/carbon/human/H = L
-	if(B.owner.nutrition > 90 && H.isSynthetic())
+	if((B.owner.nutrition > 90 || istype(B.owner, /mob/living/silicon/pai)) && H.isSynthetic()) //PAIs can't get any supplemental nutrition.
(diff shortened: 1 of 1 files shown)


Aroliacue:
Thanks :)

Very-Soft (commented):
GENERALLY pAIs shouldn't be acting as healer things. pAIs are little helper pals, not medical things. They DEFINITELY shouldn't be sources of UNLIMITED healing without any input or cost.

I think, it would be more reasonable to make it so that they have some way other than digesting people to restore their nutrition, rather than just making it so they have unlimited healing.

Such as the borg rechargers or some such.

Cameron653:
> I think, it would be more reasonable to make it so that they have some way other than digesting people to restore their nutrition, rather than just making it so they have unlimited healing.

Fair point. That'd be the best long term solution for this. I'll try and see if I can get it so PAIs can get nutrition somehow.

PR #13892, opened 2022-10-10
Update rechargestation.dm

KillianKirilenko:
Allows pAIs to recharge using cyborg chargers, as an alternative to #13888

Recharge rate is the same as for synths (10 per processing tick) and it caps at the pAI's starting amount (400).

Only gently tested, because pAI stuff is difficult to test alone.


Cameron653 (approved):
👍

PR #13894, opened 2022-10-11
Swaps soap to old sprites

Cameron653:
No more squares.
todo: make code that it has a 10% chance when it spawns in to add -alt to the sprite to make it a square. (because space soap is pretty cool)

Makes soap look like this: 
dreammaker_2022-10-10_21-11-01


Razgriz1032:
Have you also seen the monstrosity that is the shield generator sprites? :P

Here's an example of one
image

But wait, maybe THIS is the shield generator! Jk this is the mining conveyor. What is polaris doinggggg.
image

PR #13895, opened 2022-10-11
Random-Soap Addon

KillianKirilenko:
Addon for #13894 that adds a couple of extra vars and an initialize tweak so that soap occasionally appears using the new 'block' sprites rather than the classic oblong sprites. By default it's 10%, but this can be changed for mapped-in soaps to disable it entirely or increase the chance.

image


Heroman3003:
Imo, there should 100% be a way to override it and spawn a specific bar of soap, for both mapping and midround spawning, so I dont see a problem with dozens of subtypes. And if we're to have dozen subtypes, an /obj/random/soap would be more so in lieu for the purpose. Also the old soap should be included in that one too.

KillianKirilenko:
Try this one on for size. It's probably a bit cleaner? I have a hard time imagining someone wanting a specific *square* soap spawned mid-round for any other reason than pulling an admin's leg with weirdly specific requests, but if you do then it should probably just be spawned and var-edited to the square type.

There was already a random/soap but I added a new one that doesn't have the 'special' nt/deluxe/syndi soaps for more general purposes (there probably shouldn't be nt/deluxe/syndi soap on the talon for example).

Heroman3003:
That works, although I dont personally consider NT/Syndie/Deluxe soaps to be any 'special', this is a lot more flexible so I'll say this is good to go.

Cameron653 (approved):
Looks good to me.

I personally don't mind some of the square sprites (although some of them just look wonky), so it's nice to see they can still show up (with a low chance)

PR #13925, opened 2022-10-16
[MIRROR] ssatoms stuff

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/8757
--------------------
adds ssinit_misc_early
removes irrelevant stuff from ssatoms
vaguely improves maintainability of ssatoms

there are no particular performance improvements here, but the behavior and readability of ssatoms is improved. Also removes irrelevant cruft *from* ssatoms and (where relevant) adds it to a new early miscillaneous init subsystem.

probably most notably, the game won't hard-hang during startup atoms init anymore.


Heroman3003:
I'll leave this open for as long as #13596 is, because seems like continuation of fixes for that.

PR #13950, opened 2022-10-20
[MIRROR] Fixes laser rifle sprite when worn on belt

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/8773
--------------------
was reversed


Heroman3003:
This PR just won't work because of our good ol' refactored icon location code. Would need modification.

Note for the future: the bug is that the 'laser' state has west and east facing states swapped.

PR #13973, opened 2022-10-24
fluff item spawns now

TheDavestDave:
allowed_roles chaplain doesn't let you spawn in an item if you join as chaplain. I have no idea how to fix that

### DIFF
--- a/code/modules/client/preference_setup/loadout/loadout_fluffitems_vr.dm
+++ b/code/modules/client/preference_setup/loadout/loadout_fluffitems_vr.dm
@@ -1444,7 +1444,6 @@
 	display_name = "Lucky's amour"
 	ckeywhitelist = list ("thedavestdave")
 	character_name = list("Lucky")
-	allowed_roles = "Chaplain"
 
 /datum/gear/fluff/greek_dress
     path = /obj/item/clothing/under/fluff/greek_dress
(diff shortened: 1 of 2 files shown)


Runa-Dacino (changes requested):
Change 

`allowed_roles = "Chaplain`

to 

`allowed_roles = list("Chaplain")`

Very-Soft (dismissed):
Fixit, don't just remove the restriction.

Runa-Dacino (changes requested):

TheDavestDave:
> Fixit, don't just remove the restriction.

does it still need a job restriction as it's now a sub varriant of a custome with no amour vaules?

code/modules/client/preference_setup/loadout/loadout_fluffitems_vr.dm
```diff
@@ -1444,6 +1444,7 @@
 	display_name = "Lucky's amour"
 	ckeywhitelist = list ("thedavestdave")
 	character_name = list("Lucky")
+	allowed_roles = list("Chaplain")
```
Runa-Dacino, line 1447: This should hopefully fix it, and ensure it stays as a chaplain only thing @Very-Soft

Runa-Dacino (approved):

PR #13976, opened 2022-10-24
[MIRROR] Ports Aurora hijab sprite

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/8792
--------------------
Credit: https://github.com/Aurorastation/Aurora.3/pull/11433

I found this much easier to color properly because they're actually white, and I didn't like how the current ones become much darker in the side views - thus making it feel a lot harder to match with outfits.

Whether or not this is a good change should probably be considered, because they're actually fairly different.


Heroman3003:
Nonfunctional PR. If the hijab sprites are desired, gotta make a separate, manual port.

PR #13978, opened 2022-10-24
[MIRROR] Some Machine Sprites

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/8793
--------------------
Some via Aurora, some that we already had but were inconsistently used (e.g. viro centrifuge being duplicated in the hydroponics file, but an old version)
Impacts mining, hydroponics, and wherever good washing machines are sold.


Heroman3003:
We already had those lmao

PR #13979, opened 2022-10-24
[MIRROR] Selective Medical Resprites

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/8782
--------------------
Based on a few Bay PRs but being picky about a couple of things for art style reasons.

(Most) surgical tools, nanopaste and porta-freezer:
https://github.com/Baystation12/Baystation12/pull/32657
Sleeper/Scanner (yes, this still matches the existing console):
https://github.com/Baystation12/Baystation12/pull/32266

Bone gel is an older version from bay.

Also fixed body scanners never resetting to their 'empty' icon, and added 'actively scanning' states to scanners/sleepers for when the console is actually used.


Heroman3003:
We already use our own sprites for all the machinery, that we upgraded from elsewhere a while ago. This should have no practical effect, but also am considering just closing this because of such

PR #13989, opened 2022-10-26
Adds more nanomeds to gb

SingingSpock:
It's what it says on the tin. Rascal's Pass is a fun map, but it lacks nanomeds that can be used to emergency stabilize someone when medical isn't around. I added nanomeds to every public bathroom that didn't have one, as well as one to the resleeving lobby and the departure seating area.

### DIFF
--- a/maps/groundbase/gb-z1.dmm
+++ b/maps/groundbase/gb-z1.dmm
@@ -1003,6 +1003,9 @@
 /obj/machinery/light/small{
 	dir = 1
 	},
+/obj/machinery/vending/wallmed1/public{
+	pixel_y = 28
+	},
 /turf/simulated/floor/tiled/asteroid_steel,
 /area/groundbase/civilian/mensrestroom)
 "cq" = (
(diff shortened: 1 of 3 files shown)


Heroman3003:
I don't think there should be one in medical's lower entrance area, considering its both medical and public access, so if someone is there, doctors should be attending to it.

Also one in cargo seems to be kinda wack, although I cant tell if its actual map bug or just the mapdiff bot making it look wack.

SingingSpock:
The nanomeds are for when there are no doctors, and my primary purpose for putting one there was ensuring they're spread somewhat across the map, especially the level 1. If it being in there specifically is an issue, I could easily move it across the walkway to the cryo tubes and accomplish my same purpose. The cargo one... I just didn't know it existed and it got affected by some changes to the prefabs I made for directionality. I fixed it and will redo the commit soonish.

Heroman3003:
Unless there are not enough nanomeds in medical, I don't see the reason for moving one in, but I do think it'd be a pretty bad tone in general to have this vendor where people go and self-medicate right in front of actual doctors, because people can and WILL do that given the opportunity. So its better to avoid that even as possibility, IMO.

Screemonster:
I don't like the idea of random crew just self-medicating all the time. If there's no medical on then maybe now's not a good time to fight the wildlife.

Plus I can't help but feel like "I don't need to fax for a doctor there's a nanomed right there" is just going to lead to people bandaging over their untreated infections then keeling over vomiting mid-scene.

Hell, that _already_ happened a couple of weeks ago. Some dude kept getting into fights with dustjumpers, getting eaten, escaping, running off, standing next to the medibot for heals, then going off into the wilderness to try and defeat a chinchilla again - this was on a shift where there _were_ active doctors, he just didn't speak to them until he finally started vomiting everywhere and passed out outside the bar, at which point the med staff took him in and found he had like five separate necrotic infections.

SingingSpock:
Firstly, this is a feature on *every* other map. The Tether even has two mini-medbays with sleepers in it, SD has nanomeds in its one bathroom but it's small enough that that's always quickly accessible. Secondly, wall Nanomeds don't contain much medication. They have gauze, ointment, and some inaprov injectors. They are enough to save you from dying immediately of a traumatic injury, nothing more, especially when there is no actual medical staff to treat internal injuries. They don't have painkillers. They are, in fact, less useful than a medbot, which has unlimited supplies. Thirdly, people being dumb and not going to the doctor is, across every single server I've played on, completely independent of how much publicly available medication is. That guy would've fought fauna to the death with or without the medbot, the medbot just enabled him to do it a bit longer. Fourthly, the kinds of injuries you can treat with a Nanomed are precisely the kinds of injuries that, in real life, you would absolutely not bother a doctor with. If I get a minor cut, I'm going to go to my first aid kit, grab a bandaid, and put it on. This does not need the skills of someone who has dedicated a good portion of their life to medical school. Once you start taking organ damage (as the guy in your story did) it's beyond the reach of what can be self-treated and needs to go to medbay.

Issue #13992, opened 2022-10-27
[GENERAL] Asterisks break Me posts


SatinIsle:
#### Brief description of the bug
Using more than one * in a me post causes the entire post to be deleted and replaced with the error message: You may use only one "*" symbol in your emote. 

#### What you expected to happen
For the post to go through with the *'s visible in the message.

#### What actually happened
When using me (whether in the command line or through the hotkey 5 window), I was writing my message normally. I accidentally used * 's to bold some text rather than +'s. This meant that when I went to post my message, the entire message was lost, replaced with "You may use only one "*" symbol in your emote." I also died inside as I hadn't backed up the message out of game.

#### Steps to reproduce
- Simply post any "me" verb message that includes two or more *'s.

#### Code Revision
Server revision: B:release D:
Commit: [9d6828a0f1e8780052abb5a30e06339174448dd3](http://127.0.0.1:56698/tmp13184/%3Ca%20href=%22https://github.com/VOREStation/VOREStation/commit/9d6828a0f1e8780052abb5a30e06339174448dd3%22%3Ehttps://github.com/VOREStation/VOREStation/commit/9d6828a0f1e8780052abb5a30e06339174448dd3%3C/a%3E)
 

#### Anything else you may wish to add:
- This ONLY applies to the Me verb.
- This does NOT apply to Say, Whisper or Subtle. (I have tested all three)

PR #13993, opened 2022-10-27
Using asterisk reflects your message before returning

Very-Soft:
Not sure why asterisks are a problem in emotes, but! This will reflect your message back at you instead of just deleting it.
Fixes https://github.com/VOREStation/VOREStation/issues/13992


Heroman3003:
I'm merging this, but reminding myself to look deeper into 'me' asterisks to figure it out.

PR #13998, opened 2022-10-28
[MIRROR] Halloween Costumes

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/8797
--------------------
Adds the following costumes:

- Bunny
- Dog
- Hot Dog
- Snowman
- Bee
- Witch
- Owl (resprite)
- Griffin
- Jester
- Scarecrow
- Ice Fairy
- Lobster
- Roman Soldier
- Vampire (wear it with the high waisted pants!)
- Pharaoh
- Mummy (combine with above if you like)

Also:
- Re-sorts and re-paths a lot of costume items within their respective files, hence the stupid big diff. This not only makes it easier to do the loadout lists, but also way easier for admins to find Funny Costume Items when spawning stuff in-round for say, a halloween gala.
- Removed a bunch of costume items from loadout...
- ...Added those costume items and more to new costume selections instead (Uniform, suit, hat, and mask).
- Removed the scratch cueball (since the suit is gone and its just homestuck), and WH40K robes ~~for basically the same reason I'm not porting the seventy crusader outfits so many servers seem to have for totally normal reasons.~~
- Fixed ERT Ranger uniforms all having the same name and not listing properly smh

Costume loadout selections could be removed post-halloween if we felt like it was too silly but I think we can mostly trust people not to dress up as jesters and jingle to work too much.

A lot of these are via:
https://github.com/Aurorastation/Aurora.3/pull/14978
with credit to the author and the other people they credited (We had a bunch already, I think most of which are _really_ old.)
Vampire onmob is by me because it was missing from that PR.
Hot dog is from vore.


Heroman3003:
All of this is broken and will need manual porting closing until then.

PR #14012, opened 2022-10-29
Wormborgs

KillianKirilenko:
Ports UristMcStation's Wormborgs. Comes in eight flavours to cover most roles;
image
image

A couple of states run double duty: service covers butler and general clerical, janitor covers janitor and science.

Look for the `W02M` near the bottom of the sprite list for non-dogborg modules.


Screemonster:
Oh, worm?

Very-Soft:
🪱

PR #14028, opened 2022-11-03
Allergen Effect/Reaction Traits

KillianKirilenko:
This is another one of those little things that I should have put together a long time ago.

It adds a handful of new neutral traits that you can use to modify the actual reaction(s) you have to your allergens. Really uncomfortable with the idea of choking? You can disable oxydamage entirely. Want to remove the damage, confusion, and pain effects, and instead go into a kind of sleepy torpor-like state? You can do that too! Given the entire mechanic is opt-in I figure hey, more choice on what *kind* of reaction happens is good. The only effect/reaction that can't be disabled is the involuntary pale/shiver/etc. emotes, as I consider those an essential signal to bystanders that you haven't collapsed due to something (relatively) benign like SSD or whatever.

Just, you know, don't go crying to medical if you died from an itty bitty sip of milk because you decided to enable all the damage reactions _and_ added increased reaction strength too.

Tested locally and works as-intended so far as I can see.

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/traits_vr/neutral.dm
+++ b/code/modules/mob/living/carbon/human/species/station/traits_vr/neutral.dm
@@ -278,7 +278,7 @@
 	allergen = ALLERGEN_COFFEE
 
 /datum/trait/neutral/allergy_reaction
-	name = "Reaction: Disable Toxicity"
+	name = "Allergy Reaction: Disable Toxicity"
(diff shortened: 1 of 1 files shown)


Heroman3003:
I'd suggest putting word Allergy in the name of traits, for clarity.

KillianKirilenko:
Not a bad idea. I've renamed the intensity/strength traits as well so they're sorted with them.

PR #14033, opened 2022-11-04
Axes the axes

Heroman3003:
Fireaxe cabinets now spawn empty. End of an era.

### DIFF
--- a/code/game/objects/structures/fireaxe.dm
+++ b/code/game/objects/structures/fireaxe.dm
@@ -12,7 +12,7 @@
 	var/hitstaken = 0
 	var/locked = 1
 	var/smashed = 0
-	var/starts_with_axe = FALSE
+	var/starts_with_axe = TRUE
 
 /obj/structure/fireaxecabinet/Initialize()
 	. = ..()
@@ -189,5 +189,5 @@
 	icon_state = text("fireaxe[][][][]",hasaxe,open,hitstaken,smashed)
 
 
-/obj/structure/fireaxecabinet/full
-	starts_with_axe = TRUE
+/obj/structure/fireaxecabinet/empty
+	starts_with_axe = FALSE
--- a/code/game/objects/structures/fireaxe.dm
+++ b/code/game/objects/structures/fireaxe.dm
@@ -12,10 +12,13 @@
 	var/hitstaken = 0
 	var/locked = 1
 	var/smashed = 0
+	var/starts_with_axe = FALSE
 
 /obj/structure/fireaxecabinet/Initialize()
 	. = ..()
-	fireaxe = new /obj/item/weapon/material/twohanded/fireaxe()
+	if(starts_with_axe)
+		fireaxe = new /obj/item/weapon/material/twohanded/fireaxe()
+	update_icon()
 
 /obj/structure/fireaxecabinet/attackby(var/obj/item/O as obj, var/mob/user as mob)  //Marker -Agouri
 	//..() //That's very useful, Erro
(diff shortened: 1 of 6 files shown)


Very-Soft:
Why even have the cabinet if it's spawning empty? @ w @ 

Like I'm all for their removal from being spawned as normal map equipment, I just don't understand the thinking leaving he cabinet behind. That's just gonna make people think they're bugged for starting empty, or otherwise be a reminder about what isn't a thing anymore.

KillianKirilenko:
Has there been some spate of people abusing fireaxes recently or something? I can understand if Exploration keeps yoinking them, but I think Soft is right in that it would be better to just delete the lockers from the maps than do this.

Heroman3003:
There wasn't but topic was brought up and ultimately question of "why do even have the axes" along with it. This was a bit of a meme similar to the decorative ones but after some deliberation, all of them, including decorative will just get axed instead.

Zeldazackman:
If this was brought up as a matter of the fireaxes being abused then it'd be logical in a way to counter the abuse.

However if this is a removal just for the sake of it being a removal due to the axes not being utilized quite as much then I'm not quite so sure, as most of the time that I've seen one being used it'd be for clearing out spaceweeds when cargo isn't ordering goats.

Heroman3003:
> If this was brought up as a matter of the fireaxes being abused then it'd be logical in a way to counter the abuse.
> 
> However if this is a removal just for the sake of it being a removal due to the axes not being utilized quite as much then I'm not quite so sure, as most of the time that I've seen one being used it'd be for clearing out spaceweeds when cargo isn't ordering goats.

Simply put - after a bit of deliberation it became clear that fireaxe has no real use case. The space weeds clearing is probably THE only valid use for it, but even then there are more directly-intended ways to deal with space weed problem. Otherwise, as a tool, fireaxe is objectively worse than anything else departments that have access to it have already, and there is never a reason to use it unless you just want to be more destructive than you have to me. And as a weapon... Well, if its seen as more a weapon than a tool, then none of the places that have it right now shouldn't have it at all. Add to that a history of it being taken by random people to go ham on spider hunting, and there is not really a good reason for it to exist at all.

Zeldazackman:
> 
> Simply put - after a bit of deliberation it became clear that fireaxe has no real use case. The space weeds clearing is probably THE only valid use for it, but even then there are more directly-intended ways to deal with space weed problem. Otherwise, as a tool, fireaxe is objectively worse than anything else departments that have access to it have already, and there is never a reason to use it unless you just want to be more destructive than you have to me. And as a weapon... Well, if its seen as more a weapon than a tool, then none of the places that have it right now shouldn't have it at all. Add to that a history of it being taken by random people to go ham on spider hunting, and there is not really a good reason for it to exist at all.

As far as I'm following, the reason for the removal isn't due to abuse but rather disuse in that fireaxes aren't being utilized nearly as often to argue for their continual existence?

Very-Soft:
The fireaxe is intended to be used as a tool to help people get around in case of a fire, but in practice actually no one does that, and the alternatives for doing that, being hacking or dismantling windows, are generally faster and easier to do with the normal tools you use to do that, so everyone does that instead. You need a multitool to use the fireaxe anyway, so if you can get the fireaxe you PROBABLY already have tools. 

That just leaves its other function, which is a pretty good weapon. It kills a lot of stuff real fast! It isn't necessarily a problem for someone to pick up something they have laying around to go fight monsters with. But if its only practical mechanical function is that it's pretty good for fighting monsters, which is what people DO use the axe for, then I don't think it should necessarily be something that is normal department equipment. It's not atmos techs or paramedics or most bridge crew's job to go fighting monsters, but the second monsters appear on the station, the people in those jobs tend to rush for that axe rather than calling for the job who's job is to fight the monsters, which IS a bit of a problem.

One could argue 'just nerf the damage', and sure, we could probably do that, but that would just make it actually useless as an item. I think that it would be best to be a thing that can be found on away maps. In those cases it might even be used for its intended purpose. I like the axe in theory, it just seems misplaced on vorestation.

PR #14061, opened 2022-11-09
Species Alcohol Mod Cleanup

KillianKirilenko:
It's #14059 but adjusted for a couple of minor changes and the existence of alcohol tolerance/weakness traits. Also adjusted the trait descriptions a little.

Closes #14059.


Heroman3003:
Various trait/code cleanup seems good but I am personally unsure about default species being given default modifiers for the alcohol.

Heroman3003:
Ah, nevermind that, it was conversion of reagent_tag stuff from alcohol.

Though on that note I'd suggest extra trait that is 'normal liver' or something like that so that those species can reset their resistance to 0.

KillianKirilenko:
> Ah, nevermind that, it was conversion of reagent_tag stuff from alcohol.

Yeah, I never replaced the old strength_mod/alien `IS_THING` checks in ethanol when I first added it, which I probably should have.

> Though on that note I'd suggest extra trait that is 'normal liver' or something like that so that those species can reset their resistance to 0.

Done.

PR #14077, opened 2022-11-13
Makes some species specific emotes general

TankTheBirb:
Adds species specific emotes to the general carbon list. Included is: mchirp (audible)
gnarl (audible)
surprised (audible)
tchirp (audible)
trill (audible)
bounce
jiggle
light
vibrate


Heroman3003:
I wonder if it makes *help for those that have it naturally double copy them... Eh, not a big enough deal. Looks good.

PR #14078, opened 2022-11-13
Xenochimera Fix for Numbing Fangs/Sharp Melee overriding Claws

Rykka-Stormheart:
Claws were being reset to the default attacks list for numbing/etc, rather than using the chimera specific subset, because in order to update attacks we have to replace the entire list. CODE. <:


Rykka-Stormheart:
[oh yeah this is a hotfix I did 3 1/2 months ago oops](https://github.com/CHOMPStation2/CHOMPStation2/pull/4645)

PR #14093, opened 2022-11-17
equip_special

Very-Soft:
makes it so that when equipped, items call equip_special(). By default equip_special() does nothing. 

This also makes it so that if you use the hotkey to equip things, such as the size standardization bracelets, they will do the things they are supposed to.

### DIFF
--- a/code/modules/clothing/under/miscellaneous_vr.dm
+++ b/code/modules/clothing/under/miscellaneous_vr.dm
@@ -127,6 +127,25 @@
 	var/emagged = FALSE
 	var/target_size = 1
 
+/obj/item/proc/equip_special()
+	return
+
+/obj/item/clothing/gloves/bluespace/equip_special()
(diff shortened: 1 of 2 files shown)


code/modules/clothing/under/miscellaneous_vr.dm
```diff
@@ -127,9 +127,12 @@
 	var/emagged = FALSE
 	var/target_size = 1
 
-/obj/item/clothing/gloves/bluespace/mob_can_equip(mob/M, gloves, disable_warning = 0)
-	. = ..()
-	if(. && ishuman(M) && !disable_warning)
+/obj/item/proc/equip_special()
```
Heroman3003, line 130: The PR as a whole looks good. Although I do think this 'base' proc could use being defined in some more generalized place rather than hidden away here.

PR #14096, opened 2022-11-17
Trash pile adjustments

Novacat:
- Removes the zip gun from the gamma list
- Removes the emag from the alpha list
- Moves the weapon permit from the gamma list to the alpha list
- Adds the single-beacon translocator to the beta list
- Adds size bracelets (including an emagged variant) to the beta list

### DIFF
--- a/code/game/objects/items/weapons/id cards/cards_vr.dm
+++ b/code/game/objects/items/weapons/id cards/cards_vr.dm
@@ -8,3 +8,6 @@
 /obj/item/weapon/card/emag/used/Initialize()
 	. = ..()
 	uses = rand(1, 5)
+
+/obj/item/weapon/card/emag/one
+	uses = 1
--- a/code/game/objects/items/weapons/id cards/cards_vr.dm
+++ b/code/game/objects/items/weapons/id cards/cards_vr.dm
@@ -8,6 +8,3 @@
 /obj/item/weapon/card/emag/used/Initialize()
 	. = ..()
 	uses = rand(1, 5)
-
-/obj/item/weapon/card/emag/one
-	uses = 1
--- a/code/game/objects/structures/trash_pile_vr.dm
+++ b/code/game/objects/structures/trash_pile_vr.dm
@@ -22,7 +22,6 @@
 		/obj/item/weapon/bluespace_harpoon,
 		/obj/item/clothing/glasses/thermal/syndi,
 		/obj/item/weapon/gun/energy/netgun,
-		/obj/item/weapon/card/emag/one,
 		/obj/item/weapon/gun/projectile/dartgun,
 		/obj/item/clothing/gloves/black/bloodletter,
 		/obj/item/weapon/gun/energy/mouseray/metamorphosis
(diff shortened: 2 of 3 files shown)


Skysteak:
What is the reasoning behind the Emag changes?

Novacat:
It was generally agreed that they have little scene use and are mostly a kind of trap item that if someone stumbles upon they can cause issues and get into trouble with. They are also very powerful mechanically, especially when they have the full 10 charges.

Skysteak:
> It was generally agreed that they have little scene use and are mostly a kind of trap item that if someone stumbles upon they can cause issues and get into trouble with, as well as being the bad kind of secbait. They are also very powerful mechanically, especially when they have the full 10 charges.

I feel like if you're going to go from ten charges to zero changes then you may as well just remove them, as their utility is sharply curtailed to the point of nigh on uselessness.

Heroman3003:
I do think emag is just better off gone from trash pile obtainability entirely. This feels like compromisory appeasement attempt which... Is just unnecesary, as the waters will boil regardless.

R3dtail:
This feels like it's just aiming to baby proof the place even more, leading to the place feeling like a more sterile feeling place where nothing bad ever happens ever.  I get it, the place mostly caters to those who want calmness, but if there's too much of a sterile feeling then...it feels like nothing happens.  Which leads to reduced engagement, which leads to even less happening.  Sure, it's possible to be a shitter with some of the stuff being removed, but it's ultimately possible to be a shitter with, nearly everything available.  Zip guns are typically a one shot thing, science can make even worse things and can  (and has! ) handed them out or missused them.  Emags , sure are the main point  Isee a lot of people taking issue with being removed, because of their utility.  They're a useful scene tool to have if you're say, a robot, or have a NIF, or reall y there's a lot of creative uses one can have with the funny card, even though I imagine a not insignificant amount of people juse use them on a NIF shop or god forbid an arcade machine or pizza voucher.  Those latter two, sure, can be problematic and a little annoying, but it's a surprise that can lead to a dynamic interaction with other people, and a delightful time all around and ultimately a learning experience.  Since that's part of the key bits of ss13 is... learning by dying in incredibly funny ways.  If people are abusing the card deliberately to BE a shitter in a way that's negatively impactful to the health of a round or anyone else they should be banned, not remove a tool that people can creatively use for dynamic, and fun, roleplaying opportunities.

Wulfestwulf:
I agree with R3dtail, the EMAG adds some funny or interesting moments, however removing the Zip guns is fine imo, I have never seen them be used for anything ever. And also, instead of removing the EMAG from the trash piles, you could remove the arcade/holodeck effects (the two most harmful ones imo).

Novacat:
Zip guns are not being removed because they are a potential shitter item, they are being removed because they are completely worthless.

Emags are being removed for a multitude of reasons:
-They are very strong items, in normal SS13 they are only accessible by antagonists for a very good reason.
-We went through the logs over the past three months, the only time they were used in a scene involved a synth, and was done entirely through roleplay.
-They are generally only used for powergaming, memes, or griefing.
-The only 'Well maybe alright' use we could find was size bracelets, we took that feedback and are now adding emagged size bracelets as a possible maint drop.

As shown in the PR, it was considered adding a one use emag as a possible gamma reward, one use would limit the damage should someone go ham with it, while it would be rare enough that an emagged nifsoft vendor is no longer an ubiquitous presence, but this was removed after feedback.

PR #14111, opened 2022-11-20
removes meme reaction, adds sickness to regen

Screemonster:
Removes the meme monstertamer reaction that existed mainly as a means for medbay to force a chimera regen - it's no longer needed as the regen cooldown has since been massively nerfed and chimera can now be defibbed normally.

Speaking of things that were implemented since chimera regen was a thing - they get a variant of resleeving sickness now if they revive from dead, just like everyone else. Dying sucks. Try not to do it. The duration is SLIGHTLY shorter to account for the time taken by regenerating in the first place.

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/station_special_abilities_vr.dm
+++ b/code/modules/mob/living/carbon/human/species/station/station_special_abilities_vr.dm
@@ -118,10 +118,13 @@
 
 		//Dead when hatching
 		if(stat == DEAD)
+			var/sickness_duration = 10 MINUTES
 			//Reviving from ded takes extra nutrition - if it isn't provided from outside sources, it comes from you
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/carbon/human/species/station/station_special_abilities_vr.dm
```diff
@@ -168,6 +171,13 @@
 
 	revive_ready = world.time + 10 MINUTES //set the cooldown CHOMPEdit: Reduced this to 10 minutes, you're playing with fire if you're reviving that often.
 
+/datum/modifier/resleeving_sickness/chimera //near identical to the regular version, just with different flavortexts
+	name = "imperfect regeneration"
```
Heroman3003, line 175: Why is name not capitalized

PR #14129, opened 2022-11-21
Makes lootboxes a hell of a lot more random

Screemonster:
Overhauls the abandoned crates so it now uses a weighted list of things instead of the giant switch statement.

It now picks from a list of all possible junk. If the junk is low-value, it picks again until it has enough items to be worth packing in a crate!

Items that previously appeared on the list that spawned multiples of an object, for instance, will just spawn one without taking up a lot of the value allocation. Good stuff, regardless of how common it is, will take up a large amount of the allocation.

Some of the items that really needed to be kept together have been put into backpacks and dufflebags. If you get a dufflebag in a crate be sure to look inside!

Hydroponics tray spawns have been launched into the sun.


Runa-Dacino:
I like that you'll no longer possibly spend minutes decoding a harder lootbox, and be rewarded with just 1 horrible tie.

You now have (potentially) 11 opportunities for something fun.

Screemonster:
The horrible tie is _supposed_ to spawn alongside a chameleon jumpsuit in the old version. I've split those into two separate options.

liache:
This is a good PR. As per request, addition suggestions:

- More mutation injectors. Whichever ones don't have easily abused interactions together. Genetics is rare but very fun, and the x-ray inclusion is a big find. 

- Additional activated ghost pods. Posibrains and prommies as ancient awakened organisms could be scene starters, for example. The rare posi brain into ripley interaction might be slightly more common with the mining mindset already in play. 

- Mystery autoinjectors, if they're not too tough to add. Regular/joke chems good, but appropriately-weighted rare reagents like carpotoxin derivatives or vermicetol can be identified by chemistry. More interaction potential.

- Perhaps a very very rare arch find? It's very thematic, and allowing a slim chance for rnd to get into anomalous tech without archaeology- appropriate characters could be a nice touch.

- Teppi pops out. For the extra silly.

Screemonster:
I don't know which DNA injectors are particularly worth having besides xray. TK and telepathy were the ones that were mainly annoying.

Arch finds are already in there I think?

liache:
You're right, arch finds are there. Sorry, missed them!

As to genes, I would recommend nobreath, noshock, and/or cold_resist. Telepathy and tk can suck to bring to station, and regen is excessive b on this code I think, but getting one random environmental immunity component for a round should be harmless fun, if you like the idea.

Skysteak:
To summarize my thoughts and suggestions put on the Discord:

1. Overall I like  the approach and think it's better than the system we have now.

2. Making the system be able to take multiple types of sword (and even gun) would add variety without hopefully bloating or making the list too weapon heavy. Not sure how doable or not this is however!

3. The genes presented sound fair, although I'd advise double-checking them as genetics code feels like the coding version of a haunted house, and they may not work (as intended).

4. As suggested in Discord, a backpack with a Red Cap, Blue Jacket, White Trousers, Standard Explo Cataloguer and either a Capture Crystal or Net Gun will permit someone to be the very best (like no one ever was).

Screemonster:
As you can see from the mouseray thingy, it's possible to have the list contain sublists and _pick_ from a list, so I can make a single entry contain multiple possible guns. It already contains multiple subtypes of esword under the same entry, so it all goes in as "an esword".

Now I've figured that out I might re-consoldidate the ores into a single entry.

Screemonster:
I've added the DNA injectors and the capture crystal master outfit suggested above, along with a net gun. I've also fiddled with the weights since the list is a _lot_ bigger than it used to be.

I've sorted the list in a vaguely weight-then-value order, but keeping the spacecash and pickaxes grouped together.

I am well aware that when this goes live it'll need playtesting to figure out the ideal weightings and value figures for all this crap because some of the sub-lists being in as their own thing makes my brain hurt trying to figure out what the actual probability is for, say, a chunk of diamond ore.

Screemonster:
Removed the cataloguer from the pokemaster backpack as it will interfere with exploration (things scanned by one cataloguer cannot by scanned by others in the same shift, so if a miner finds one and actually uses it, it'll deprive the explorers of points even when the miner can't use them)

Added a few hacking devices - hacktools, emags, a fully loaded toolbelt with power tools, and random alien tools.

A sample of 10 crates with the current weights provided the following rolls:
image
image
image
image
image
image
image
image
image
image

PR #14136, opened 2022-11-23
Adds more altevian things

Heroman3003:
Microdosing on rats
Adds 1 food and a set of captainy suits.


Verkister:
I assume the cyberpunk vest species restriction was unintentional?

PR #14137, opened 2022-11-25
Simplified Hacktool

KillianKirilenko:
Something about the TGUI interface for the old hacktool just doesn't work, so fuck it, I've stripped out the fancy functionality for a much simpler approach.

Its function is now very simple; it can let you use doors that you don't have permissions for. Hacked doors can be opened or closed, and the tool 'remembers' a short list of recently-hacked doors so you don't need to rehack them constantly. In exchange for the reduced functionality, the overall hack time is now a lot shorter. It also can't raise lowered bolts unless you have the special subtype (so, if you want to hack into capital-S Secure areas, you need to do it the old-fashioned way).

Fixing the TGUI is too much of a headache and this isn't a traitor-oriented server, so I don't really see the need to go through the trouble of making it so people can go around messing with doors on a whim by electrifying/bolting/de-safetying them. Might as well just make it a nice easy *open sesame* tool.

They have not been added to any loot sources in this PR: I leave availability up to the staff's discretion.

### DIFF
--- a/code/game/objects/items/devices/hacktool.dm
+++ b/code/game/objects/items/devices/hacktool.dm
@@ -46,10 +46,12 @@
 
 	// Note, if you ever want to expand supported_types, you must manually add the custom state argument to their tgui_interact
 	// DISABLED: too fancy, too high-effort // A.tgui_interact(user, custom_state = hack_state)
-	// Just brute-force it open
+	// Just brute-force it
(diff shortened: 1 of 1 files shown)


KillianKirilenko:
Oh shoot, I knew I should have drafted this before I took a nap.

Going to revisit this in a little bit, I want to tweak how hacking works some more.

PR #14161, opened 2022-12-03
fixes "Revert "promethean fixes and QoLs""

Seris02:
did a big woopsie, all the sprites n stuff work and the original pr back to working as well
tested with custom species, shapeshifter species and normal species


Seris02:
fixes #14160

PR #13816, opened 2022-09-27
Browser Stat/Verb Panel

ItsSelis:
This PR is in a functional, yet experimental, state. The verbs still all work fine and nicely (with the MC tab for example even having better performance/no lag). Though some visual bugs could occur.
Please report such bugs in the comments or make a new isssue.

@Cameron653 has made a simple preview of the current state (as of commit [41c0470](https://github.com/VOREStation/VOREStation/pull/13816/commits/41c047008d573dfc84cd643efd35cfac178bd8a9)):

The stuff list:
- [x] (!) Fix categories disappear (but verbs stay) when joining the round (on the admin tab)
- [x] ( ) Categorizing the verbs (or some of them at least, if wanted)
- [ ] ( ) Pref for either splitting sub-categories of admin tab into tabs or leave it categorized under admin (Default: No split)
- [x] ( ) Removing the Interviews stuff (tickets stay of course)
- [x] (!) Globals/Master-/Failsafe-controller not showing any stats (but runtimes are fixed)
- [x] (!!) Implement switching to dark mode (necessary styles and functions do exist already)
- [x] (!!!) Make Examine panel work
- [x] (!!!) Less Icon lag (Turf examination)
- [ ] desc_panel_image should use icon2html instead of the old code to properly render interaction icons
- [ ] Loading the server with me as a host does not give me access to the tools in the admin panel until I load in as an observer. I otherwise only have what you see in the image below.
- [x] Observe button (at least as a host role) does not make you observe and instead just removes the menu, apparently spawns a ghost of your current slot, and leaves you on the main screen. Attempting to join the round normally does work, however.
- [ ] You cannot pick things up from the turf examine menu by clicking on them.

PR #14188, opened 2022-12-11
Adds admin tokens

ItsSelis:
This PR adds the admin tokens for any admin topic (aka admin HREFs).
Each admin will have a unique, randomly generated href token which has to be passed to the called topic. Should there be no token present or should it be malformed, it will block the href call and notify the admins.

Most of the calls should be already converted over, but there could be some or a few that could be missed. That's why a so-called debug mode is currently hard coded to true. This will still make the call, even if the token is not valid, so that admins can still do their thing. It will also log which href was called to the logs and tell the admins that they should notify a developer about this, so that the href token proc can be added to the href call.

In the end this will generally make href calls more secure, and prevent incidents like making everyone monkeys because an admin pressed on a piece of paper (which is rather more a harmless version of such exploits, making it theoretically possible to exploit other href calls like banning the ckey, a job or department etc.). It especially adds the layer of security, due to the admin token having to be the exact one that the calling admin has been assigend to.

I already have conducted tests and these are the results of a bad or a missing href token:
grafik
The logs will receive such a line for example (bad token):
`UAH: src=[0x2100121a];admin_token=101098109924524068057082;c_mode=1`
Same example with missing token:
`UAH: src=[0x2100121a];c_mode=1`

If the key is valid, nothing is logged or blocked of course.

**Note:**
The procpath, icon and md5 things are used for other upcoming PRs but also seem to maybe be generally useful. Since they do not really break or change anything I added them too.

This PR is also one of the atomizations of #13816


Heroman3003:
I do have mild concern with commands not going through because of href shenaniganry, but if it happens, it will be logged and also the admin can just do the command again, and hopefully its all just unfounded.

ItsSelis:
> I do have mild concern with commands not going through because of href shenaniganry, but if it happens, it will be logged and also the admin can just do the command again, and hopefully its all just unfounded.

As the screenshot already shows, the call is actually going through and it's really only just logging that it theoretically didn't. Though, I can triple check if it really really goes through just to be sure. That will only take a minute then.

Heroman3003:
I have no concerns about calls going through in general. My concern is moreso about scenario where href changes halfway through call or something and it doesnt go through and makes 'bad call' error or something. Again, mostly just noting it down here for myself in future than asking you to doublecheck or fix as screenshots indicate you tested properly as is.

ItsSelis:
> I have no concerns about calls going through in general. My concern is moreso about scenario where href changes halfway through call or something and it doesnt go through and makes 'bad call' error or something. Again, mostly just noting it down here for myself in future than asking you to doublecheck or fix as screenshots indicate you tested properly as is.

Ohhh, now I get what you mean, yeah. I would suggest leaving it on the debug mode for some while. After that it could maybe ask the admin if the call was actually correct (to prevent the paper trick but still make the call, well, callable if its so specific that the general debug mode didn't find it yet)

PR #14190, opened 2022-12-11
Adds mentor ticket panel

ItsSelis:
This panel is based of the admin ticket panel ui, but changed up in order to replicate the mentor byond ui.

There also is another tgui state, called tgui_mentor_state. All it really does, as the name implies, is checking if the client currently has mentor rights. If not, it will not open the ui (like tgui admin ui's with admin rights).

It also adds the mentor ticket list to the Tickets tab and categorizes them.
Here is a screenshot of how it looks (As admin):
grafik

Also has been tested as mentor but no other rights (aka regular User with mentor rights):
grafik


Heroman3003:
Looks good.

PR #14215, opened 2022-12-19
Adds scars and bandages from polaris

TankTheBirb:
this is from upstream-merge-8844 or whatever its called and I would've just added to that PR but it wouldn't let me so I instead stole the files from that and copied it into this PR so anyways credit goes to the people who made that PR happen

so basically there are scars and bandages and they exist for tesh and non-tesh. They look neat. Images stolen from original PR
image
image


code/modules/mob/new_player/sprite_accessories.dm
```diff
@@ -2090,7 +2090,7 @@ shaved
 
 	//Empty list is unrestricted. Should only restrict the ones that make NO SENSE on other species,
 	//like Tajaran inner-ear coloring overlay stuff.
-	species_allowed = list()
+	species_allowed = list(SPECIES_HUMAN,SPECIES_SKRELL,SPECIES_TAJ,SPECIES_HUMAN_VATBORN,SPECIES_UNATHI,SPECIES_ZADDAT)
```
Heroman3003, line 2093: Uh. No. We don't want markings as a whole restricted to just those. That's Polaris thing.

Heroman3003 (changes requested):

code/modules/mob/new_player/sprite_accessories.dm
TankTheBirb: Good catch because I didn't notice that at first, fix is up

Heroman3003 (approved):

PR #14221, opened 2022-12-20
creates a tab in the TF menu for changing markings

Seris02:
because I don't want to be finishing up tf'ing someone and then seeing it be like
"oh"
"you have markings"
"it looks v bad now."
"sadness."

feel free to absolutely blast me on the efficiency of the code, I've been very tired
however I did profile it and it's surprisingly efficient tbh
even when absolutely fucking mashing the buttons


Heroman3003:
At a glance value this looks good, but markings and human icons in general is a messy area to just glance, so I'll leave this hang a bit in case someone else spots something off.

Although, one thing that would be a cool addition with the system that actually allows dynamic changes to markings is maybe shapeshifters being able to edit their own, so possibility of that in future as well does make me rather interested.

PR #14245, opened 2022-12-25
Adds fireworks and their launchers

Heroman3003:
Adds fireworks to the game.

Science can now produce firework stars and firework launcher machinery. On RP and Tether they also start with one.

Firework launcher can be unwrenched to be moved around, but needs to be wrenched down to work.

After being loaded with a firework star, when put outside on weather-compatible planet, firework can be launched! After this the launcher enters cooldown period. Cooldown is lower the higher tier parts are used.

There are two types of firework stars - weather and aesthetic. Weather fireworks instantly change weather to their appropriate type. Aesthetic are just for fluff and send a message to everyone who can see the sky. There is also configurable aesthetic firework star where you can select color and shape it takes.

This also adds two new weather types accessible only through stars.

Admins have override that disables ability for fireworks to affect weather on certain planet for event purposes.


Heroman3003:
Btw, the icons are not really placeholder, but I am not a spriter, so they are low-quality codersprites and I totally welcome any higher-quality replacements someone may want to do.

PR #14247, opened 2022-12-25
Gives pilots their holdout back

BonniePandora:
image
Now that I got your attention with that title;

Replaces the tazer gun in pilots lockers with the LAEP97 "Defender" secure shuttle-protection pistol.
Basically a rehashed secure small energy gun with the phaser's facility safety interlock added & the lethal capabilities upon red-alert stripped out. 
Instead, it now has the option between 6 shots of tazer blasts, or 3 shots of (weaker) stun-revolver esque fire.

Haven't included any spare weapon cells, however the device cell in their locker is enough for a single additional tazer-shot from it in a pinch.
The interlock can be disabled by someone with armory access(Warden, HoS, Captain), so this should help prevent any hotshot pilots from using this to grief/job-steal security without deputization.

Sprites are basically the same as the secure small energy gun. Also got the flashlight to work too, so that's something. If someone wants to correct it to be the directional kind of light, they're welcome to do so.


BonniePandora:
Thank you Virgo-Flyer active edges, very cool

PR #14253, opened 2022-12-26
Makes improved health scanners able to pinpoint rads

Cameron653:
Because why not.
Currently science can make an improved health analyzer, the other are adminspawn/maintspawn/PoI spawn, and them having the ability to pinpoint rads really isn't worth it.

This gives science a reason to spend some resources on better health analyzers and medical a (slightly increased) reason to want improved health analyzers.

### DIFF
--- a/code/game/objects/items/devices/scanners/health.dm
+++ b/code/game/objects/items/devices/scanners/health.dm
@@ -103,7 +103,7 @@
 		OX = fake_oxy > 50 ? 		"<span class='warning'>Severe oxygen deprivation detected</span>" 	: 	"Subject bloodstream oxygen level normal"
 	dat += "[OX] | [TX] | [BU] | [BR]<br>"
 	if(M.radiation)
-		if(advscan >= 2 && showadvscan == 1)
+		if(advscan && showadvscan == 1)
 			var/severity = ""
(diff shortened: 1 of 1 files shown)


Novacat:
iirc, I usually ask for improved health analyzers because they can pinpoint broken bones/internal bleeding.

Cameron653:
The improved analyzers are good, but the ability for them to detect rads were left out.

When I initially coded the radiation rework, I was under the presumption that the higher tier analyzers could be made, not just the tier 1 advanced analyzer that can't use the mechanic to point out radiation.

I doubt anyone will be going and trying to get an improved health analyzer just because it detects radiation, but it's just adding on another reason it'd be beneficial to have one.

Novacat:
Hm. My current thought is rather than move the advanced analyzer's features to the improved analyzer, a better approach would be to just give R&D the ability to make the advanced analyzers. It is silly that the advanced analyzer is basically event character only and the phasic one just does not exist at all.

Cameron653:
That'd be another way of going about it as well. Let me see about doing that real quick

Cameron653:
Made the advanced analyzer able to be created in PR #14255

PR #14255, opened 2022-12-26
Makes advanced health analyzers creatable via R&D

Cameron653:
This allows for a health analyzer that indicates how many rads are actually still in someone to be created. (And also tells if someone has minor brain damage according to the comments.)

Not really anything too game changing for medical, but if they want to have a _slightly_ better scanner than the improved one R&D can make, they can get this.

### DIFF
--- a/code/modules/research/designs/medical.dm
+++ b/code/modules/research/designs/medical.dm
@@ -88,6 +88,15 @@
 	build_path = /obj/item/device/healthanalyzer/improved
 	sort_string = "KBAAB"
 
+/datum/design/item/medical/advanced_analyzer
+	name = "advanced health analyzer"
+	desc = "An even more advanced handheld health scanner, complete with a full biosign monitor and on-board radiation and neurological analysis suites."
+	id = "improved_analyzer"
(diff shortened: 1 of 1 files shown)


code/modules/research/designs/medical.dm
```diff
@@ -88,11 +88,20 @@
 	build_path = /obj/item/device/healthanalyzer/improved
 	sort_string = "KBAAB"
 
+/datum/design/item/medical/advanced_analyzer
+	name = "advanced health analyzer"
+	desc = "An even more advanced handheld health scanner, complete with a full biosign monitor and on-board radiation and neurological analysis suites."
+	id = "improved_analyzer"
```
Heroman3003, line 94: Same ID as improved_analyzer design.

Cameron653: Whoops, copy and paste error. Let me fix that

Novacat (approved):
Seems fine. Requiring diamonds does put it a step up above the Improved. The event character analyzers should probably be replaced with phasics, but that will be in a separate PR.

PR #14263, opened 2022-12-29
Allows SD-exclusive job titles to be used anywhere

Heroman3003:
I thought it'd be fun to keep ship-y titles to ship, but it seems like it just makes them really hard to actually use. Sad, but I think this is better, although if other staff wants, I could keep some or all exclusive, this is mostly a suggestion in current state.


Runa-Dacino:
Yay

Very-Soft:
I've thought they should all be available anywhere from the start

Heroman3003:
It seemed like a cool idea to have different titles on different map okay ;w;

PR #14266, opened 2022-12-29
Fixes "switch tail layer" verb

Verkister:
Fixes alternative tail layer applied by the "switch tail layer" IC verb that was intended to put the tail on top of other stuff.


Heroman3003:
I'm not sure of purpose of this other than overriding tails even when they are supposed to be in lower layer from the front view.

PR #14267, opened 2022-12-29
Move to Database Subsystem

ItsSelis:
# **!NOTE TO DOWNSTREAMS!**
Please make sure that both database templates from the SQL folder were put in the feedback database (otherwise the contents of the dbcon_old database (death, karma, karmatotals, library, population) have to be moved over to the feedback database containing all the "erro" tables manually!

When checking the database, it should look something like this (actual database name might differ of course!):
grafik

In this PR, the server **will not** try to connect to an old (seperate) database containing the death, karma, karmatotals, library, population tables anymore!

-----
## Note!
It is **highly** recommended to create new database backups, just to be sure, before merging this in any case. The tests below have shown that nothing should be wrong and when compared it is basically the same code that was just moved over to a subsystem.

This should be a regularity, at least to be done daily, but its better to be sure and create a manual one before merging.

-----

### Description
This PR basically does 2 major things:
1) Move database code into a subsystem
2) Remove dbcon_old stuff

The login credentials of the dbcon_old was removed from the database config example. Any existing configuration will be ignored.
Virgo on production will have no conflict by the removal of dbcon_old, which has been checked by one of the maintainers.

In the future, asynchronous querys can be implemented and systems to query tables, without blocking the rest of the code execution. This could potentially improve roundstart lags.

Due to dbcon_old being removed, the stuff from its sql template was moved over (and updated) into the feedback database template. The missing table erro_connection_log was also added, which had caused some SQL Query errors, due to it being unable to find the nonexistant table.

This PR also adds more SQL query logging, which is logged in a seperate log file (at the same location as the other log files, but called sql-error.log).

Here are some screenshots of the tests of the "new" system on a local development instance:

erro_connection_log:
grafik

erro_player:
grafik

erro_privacy:
grafik

population:
grafik

library:
grafik
grafik


Heroman3003:
Ngl, this is plainly kinda scary to merge as someone who doesn't know much about current database systems, much less new one you're proposing.

ItsSelis:
The system rather has "moved" to a subsystem, since if both added and removed database code files are observed, that they have pretty much the same or at least very similar code. Futhermore, prior of the creation of this PR, the production database has been checked in order to make the manual tests that were conducted are as reaslistic to production as possible.

But it is fair to wait for someone a little experienced in database stuff to review the debug query logs or, well, the database tables that have been created by the changes that were made in the PR.

JoanRisu:
Ach, this should be looked at soonish. I've forwarded this to the relevant code wranglers.

ItsSelis:
This may be re-opened once no longer stale.

PR #14281, opened 2022-12-31
un-nerfs fishing from previous change

Seris02:
PR #12142 changed the max volume of snacks, which affected fishing bait, to the point where even the deluxe bait only gives the minimum reduction, 0.75

this adjusts it to a completely different formula, based around what I think the general calculated bait values should be for the bait types.

Screenshot 2023-01-03 112213

### DIFF
--- a/code/modules/fishing/fishing_rod.dm
+++ b/code/modules/fishing/fishing_rod.dm
@@ -89,7 +89,7 @@
 		for(var/datum/reagent/re in Bait.reagents.reagent_list)
 			if(re.id == "nutriment" || re.id == "protein" || re.id == "glucose" || re.id == "fishbait")
 				foodvolume += re.volume
-
+		foodvolume = Bait.reagents.maximum_volume == 80 ? (min(foodvolume,50) * (Bait.reagents.maximum_volume / 50)) : Bait.reagents.maximum_volume // VOREStation edit: because of changes to max reagents in snacks, readjusts the foodvolume to have the same values as polaris again, but only when it has the same max volume as the changed amount in snacks (assuming the snacks reagent size stays the same...)
 		toolspeed = initial(toolspeed) * min(0.75, (0.5 / max(0.5, (foodvolume / Bait.reagents.maximum_volume))))
 
 	else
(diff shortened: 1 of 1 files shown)


Heroman3003:
Just change internal volume for the bait specifically back to what it was pre-nerf. This feels needlessly convoluted, and bait was just not something I even thought accounted for something like volume; nor do I think it needs to have expanded volume snacks do. Main reason for expansion was to fit all the reagent from cooking, modular or normal, when it comes to 'bigger' recipes.

Seris02:
> Just change internal volume for the bait specifically back to what it was pre-nerf. This feels needlessly convoluted, and bait was just not something I even thought accounted for something like volume; nor do I think it needs to have expanded volume snacks do. Main reason for expansion was to fit all the reagent from cooking, modular or normal, when it comes to 'bigger' recipes.

The thing is that you can use all snacks as bait, so if I want to just stock up on... I dunno
Super bite burgers for bait, which would be great on polaris code, about the same as deluxe worms, I wouldn't be able to here, and they'd have the minimum reduction still

Heroman3003:
I just really dont like the whole "if we are 80 do one version, if not do the other" option. Its very snowflakey. Changing formula to be more universal rather than snowflakign it up to account for our changes is preferred when we are a 'soft' downstream to polaris anyway. Ideally, there'd just be one universal formula that applies to all snacks still. Maybe make it require less 'percentage' of foodiness to top it off, but cap it or something. Or make it based purely on numeric amount of food. Something that's not "check if we've been buffed or not and calculate differently depending on that".

Seris02:
> I just really dont like the whole "if we are 80 do one version, if not do the other" option. Its very snowflakey. Changing formula to be more universal rather than snowflakign it up to account for our changes is preferred when we are a 'soft' downstream to polaris anyway. Ideally, there'd just be one universal formula that applies to all snacks still. Maybe make it require less 'percentage' of foodiness to top it off, but cap it or something. Or make it based purely on numeric amount of food. Something that's not "check if we've been buffed or not and calculate differently depending on that".

aight, will just go with the formula I was going to change it to beforehand then, will do tomorrow

Seris02:
is done
mlem

PR #14285, opened 2023-01-01
adds halos to loadout

Detective-Google:
a quick and dirty PR to add a customization option I found to be lacking.

image
image
image

if someone wants to donate a prettier sprite that's cool too. I spent a while trying to make it look better and everything looked worse than the basic. :)


code/modules/client/preference_setup/loadout/loadout_head.dm
```diff
@@ -126,6 +126,13 @@
 	..()
 	gear_tweaks += gear_tweak_free_color_choice
 
+/datum/gear/head/halo
+	display_name = "halo (colorable)"
+	path = /obj/item/clothing/head/halo
+
+/datum/gear/head/halo/New()
+	gear_tweaks += gear_tweak_free_color_choice
```
Heroman3003, line 134: Needs ..() before addition of color choice.

PR #14292, opened 2023-01-03
[MIRROR] Markings expansion 2: Teshari boogaloo

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/8865
--------------------
Ports a bunch of Teshari markings from downstream, adjusting them to work here.
Teshari also no longer by default have colored feathers on their hands/feet, those are now added as markings.
This is due to the fact that those organ-defined feather markings displayed above any others, making it very hard to actually have markings on your limbs as Tesh.


Heroman3003:
I would have liked this but at this point its better to just close.

PR #14320, opened 2023-01-10
adds a button for ingesting dogborg sleeper contents + couple other miscellaneous slightly related fixes

Seris02:
default is disposal now, which ejects it as normal
ingestion puts them in the current belly

before this, neither option made any difference whatsoever, the code didn't even actually use the eject_port for anything when ejecting

also fixes the laying down delay issue (lay down > three seconds later the sprite finally updates and lays down)
also fixes the fact that there was almost always a greyed out self clean button because the else statement was in the wrong spot
makes it so the buttons wrap around the window instead of being cut off so you don't have to constantly resize it

### DIFF
--- a/code/modules/mob/living/silicon/robot/dogborg/dog_sleeper_vr.dm
+++ b/code/modules/mob/living/silicon/robot/dogborg/dog_sleeper_vr.dm
@@ -14,7 +14,7 @@
 	var/cleaning = 0
 	var/patient_laststat = null
 	var/list/injection_chems = list("inaprovaline", "dexalin", "bicaridine", "kelotane", "anti_toxin", "spaceacillin", "paracetamol") //The borg is able to heal every damage type. As a nerf, they use 750 charge per injection.
-	var/eject_port = "disposal"
+	var/eject_port = "ingestion"
 	var/list/items_preserved = list()
 	var/UI_open = FALSE
 	var/compactor = FALSE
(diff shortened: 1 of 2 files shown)


Heroman3003:
The thing it was used for is for the release message. Thats what it was there for. So that people can know whether you were released from dogborg front hole or back hole.

Skysteak:
As much as I'll miss it, good job fixing the sliding bug.

Also yeah certain borgs have front or back ports which are cosmetically the same thing but can be used to flavour RP in certain ways.

Seris02:
> The thing it was used for is for the release message. Thats what it was there for. So that people can know whether you were released from dogborg front hole or back hole.

**huh**

should I just make another button for ingesting the contents then?

Skysteak:
> r

Personally I think that'd be a better option, unless there is something I'm missing.

Heroman3003:
> > The thing it was used for is for the release message. Thats what it was there for. So that people can know whether you were released from dogborg front hole or back hole.
> 
> **huh**
> 
> should I just make another button for ingesting the contents then?

I believe main reason its both worded so vaguely and gives no option to 'ingest' via other port is general pref protection. Idk if it should at all exist really, but either way, I am not opposed to button to transfer prey from borg belly to normal one and back; but I dont think it should be tied to this thing.

Seris02:
donnnnne

Heroman3003 (approved):

PR #14321, opened 2023-01-10
fixes the colorable qipao

lukevale:
I can fix more, this is just a brute force fix due to spaghetti code making the sprites not function for what ever reason.
_for all I know it's because we didn't give it blood sacrifices_


lukevale:
IDK intergration tests failed, because it says it failed with mapping errors? Which I didn't touch?

PR #14333, opened 2023-01-15
[MIRROR] Moves blob chunk effects to artifact effects.

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/8783
--------------------
This is _mostly_ a 1:1 transfer of effects, though there's a lot of vagaries required since artifact effects aren't necessarily attached to things holdable in a persons hands. Some things were redundant with existing things.
This code is completely untested, but provided with the warranty of "I will look into issues if they're actually reported."
... It compiles, at least.


Heroman3003:
We dont use blob in that way and this is so many conflicts out the ass.... Gonna close until it actually becomes needed.

PR #14344, opened 2023-01-16
better persistent dirt

Seris02:
what the title says
no longer will it just immediately make the station a filthy fucking mess, and instead it will actually properly transfer dirt


code/game/objects/effects/decals/Cleanable/misc.dm
```diff
@@ -38,11 +38,11 @@
...
 	mouse_opacity = 0
 
-/obj/effect/decal/cleanable/Initialize(var/mapload, var/_age)
+/obj/effect/decal/cleanable/Initialize(var/mapload, var/_age, var/dirt)
 	.=..()
 	var/turf/simulated/our_turf = src.loc
-	if(our_turf && istype(our_turf) && our_turf.can_dirty && _age)
-		our_turf.dirt = 101
+	if(dirt && our_turf && istype(our_turf) && our_turf.can_dirty)
+		our_turf.dirt = dirt
```
Heroman3003, line 45: There is a small problem with this, though its kinda pre_existing. This means pre-mapped dirt will not properly dirty-up the tirf and will be stacking. I think potential solution would be to self-destruct if there is dirt on tile already; and also make sure that if we are NOT persistant dirt, we just apply maximum dirtiness to turf anyway.

I'd also  suggest making it clamp between dirt level at which it appears on tile and 101 , just in case

Finally, there is also fact that this will make icon updating for the dirt very awkward - it will start at maximum alpha like it already has anyway and then when someone walks over it will be updated for first time. Forcing update of alpha here is important.

Seris02: all suggestions applied - even better dirt code created

except that you can't just apply max dirt if not persistent, otherwise when the turf creates dirt for the very first time (when turf.dirt goes over 50), it'd just jump instantly to the max

Heroman3003: mapload might be a good way to handle that? Im not sure if its relevant for init or not. Or maybe another var that is set to false when creating it 'naturally' rather than pre-mapped? Something along those lines can work.

code/game/objects/effects/decals/Cleanable/misc.dm
```diff
@@ -38,11 +38,17 @@
...
-/obj/effect/decal/cleanable/Initialize(var/mapload, var/_age)
+/obj/effect/decal/cleanable/dirt/Initialize(var/mapload, var/_age, var/dirt)
 	.=..()
 	var/turf/simulated/our_turf = src.loc
-	if(our_turf && istype(our_turf) && our_turf.can_dirty && _age)
-		our_turf.dirt = 101
+	if(our_turf && istype(our_turf) && our_turf.can_dirty)
+		our_turf.dirt = clamp(max(age ? (dirt ? dirt : 101) : our_turf.dirt, our_turf.dirt), 0, 101)
+		var/calcalpha = our_turf.dirt > 50 ? min((our_turf.dirt - 50) * 5, 255) : 0
+		var/obj/effect/decal/cleanable/dirt/alreadythere = locate(/obj/effect/decal/cleanable/dirt, our_turf)
```
Heroman3003, line 47: Won't this just always find ourselves and self-destruct as result?

Seris02: ah shit, true, thanks

Heroman3003 (approved):

code/game/objects/effects/decals/Cleanable/misc.dm
Heroman3003, line 47: a better solution than locate for this specific one may be for(var/obj/effect/decal/cleanable/dirt/alreadythere in our_turf)

Seris02: oh, true
will do that in the morning

PR #14347, opened 2023-01-16
[TEST FIRST] Miscellaneous ZAS Fixes and Optimizations!

RimiNosha:
So, ports  
`https://github.com/discordia-space/CEV-Eris/pull/7390`, `https://github.com/Baystation12/Baystation12/pull/32291` and `https://github.com/discordia-space/CEV-Eris/pull/1615`

In short:
- Fixes ZAS debug overlays not properly overlaying over everything they should. They use the ADMIN2 (15) layer now.
- Some minor weirdness fixes to logic.
- Makes gas overlays marginally less laggy by caching icons and using effects instead of atoms to render.
- Turns get_zone_neighbors into a define, saving precious performance on the extremely hot procs it's used on.
- Adds ATMOS_CANPASS_TURF, and replaces a couple of hot usages with it.

The performance uplift won't be amazing, but it should help with consistency at the very least.

Untested beyond compiling.

### DIFF
--- a/code/ZAS/Fire.dm
+++ b/code/ZAS/Fire.dm
@@ -115,7 +115,7 @@ If it gains pressure too slowly, it may leak or just rupture instead of explodin
 	icon = 'icons/effects/fire.dmi'
 	icon_state = "1"
 	light_color = "#ED9200"
-	layer = TURF_LAYER
+	layer = GASFIRE_LAYER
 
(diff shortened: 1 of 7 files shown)


Heroman3003:
Yeah this is gonna need to be remade if its to work... And testing was never done in the end. Gonna close this but if someone wants to revive and properly test it, I welcome that.

PR #14353, opened 2023-01-19
Tilted Duffles

KillianKirilenko:
Switches dufflebags to use a tilted angle rather than being horizontal over the lower back. Includes all current non-fluff duffles. Obviously a subjective thing, but I think this looks nicer and better reflects their higher capacity compared to regular bags.

Also fixes one of the subtypes having no on-mob icons

Little preview of the north-facing on-mob icons (east/west facings are also included):
duffs

Black, engi, med, EMT
Away-team, drone*, sec, talon
Science

*The missing type!


Very-Soft:
I would be more into this if it was more like, a verb on duffles that lets you change the orientation

KillianKirilenko:
> I would be more into this if it was more like, a verb on duffles that lets you change the orientation

I had to make another half-dozen icon states for this, conduct some Time Fuckery to fix a mistake, and dig through code to find the exact proc I needed since backpacks aren't clothing and have to use a different proc to update their on-mob icons, but it's done. 50/50 for the state to be tilted/nontilted when created, and you can swap with an Object-tab verb.

PR #14354, opened 2023-01-19
Event Stuff Bugfix/Balance Pass

Novacat:
- Fixes some pathing bugs on a submap
- Fixes the Frontier DMR to have the correct ammo cost and firing sound
- Fixes ERT voidsuits back slots
- Buffs armor values of ERT and Merc voidsuits
- Nerfs armor values of Heartbreaker voidsuits
- Buffs armor values of Asset Protection, Combat, and Military rigs, as well as the knockoff Breacher rig
- Nerfs armor values of the genuine Breacher rig


Skysteak:
This is purely quibbling about things from an RP/IC perspective, but I almost feel like the genuine rig should be left untouched and either have a (slightly) more common intermediary or have the fakey fake one brought up to the 'new' values of the genuine one. I dunno as said this is IC, but considering that Breacher Rigs in lore were meant to smash into and then take on capital ship defences, making them comparable to durasteel plate just feels like a shame.

Novacat:
The main issue is that the exponential nature of armor values means that there is not a big gap between 'Squishy' and 'Indestructible'. A dude with a Breacher and nothing else has 1000% effective health. Throw a tactical jumpsuit on top of that and you have immunity to melee attacks, and 2000% effective health against ballistic/laser attacks. Ideally nothing should go over 90% armor because of how nutty it gets at those scales.

Ideally, I want to see more armor pieces that straddle the 50%-80% parameter of 'Powerful, but not invulnerable', that is what this PR seeks to do.

Skysteak:
> The main issue is that the exponential nature of armor values means that there is not a big gap between 'Squishy' and 'Indestructible'. A dude with a Breacher and nothing else has 1000% effective health. Throw a tactical jumpsuit on top of that and you have immunity to melee attacks, and 2000% effective health against ballistic/laser attacks. Ideally nothing should go over 90% armor because of how nutty it gets past 90%.
> 
> Ideally, I want to see more armor pieces that straddle the 50%-80% parameter of 'Powerful, but not invulnerable', that is what this PR seeks to do.

I largely agree with you that you don't want to approach an effective immunity! However to once more cry 'But IC/lore', I feel that the genuine breacher is something that should be used rarely and included with the expectation that no is _not_ balanced, and yes this thing _is_ a power trip. You justifiably may say 'What's the point of having an item that will see single digit use over years?', and to what I'd say that it can help build immersion and give a sense of scale setting wise. It ironically enough helps underscore the civilian/PMC nature of what's dealt with by the crew, and sells the concept of something that was likely meant to be surviving no joke heavy lasers and pulse guns.

Don't get me wrong; Post nerf if you placed it in world it'd be extremely strong, and the appreciation for its old power level wont likely be taken in by many. However I suppose the best comparison I can give is that it's like nerfing shadekin powers (which I know has happened on some level); It's understandable but undercuts part of the concept's identity.

Runa-Dacino:
I could maybe see it as a middle path solution to indicate in our breacher RIG, either within the name or its examine (right click or info window) that it's used, retired or "defective".

Like how CPUs end up sold as multiple different products depending on production quality.

We indicate that the ones that the coreward periphery can reliably access are ones that "fell off the truck" as it were, or sold off because not up to par of the Hegemony.

This would satisfy "Actual unathi wear stuff way stronger than this", and we don't need to code/keep the original stats because... on the rare occasion someone might actually wear the millitary-grade stuff? A GM can surely edit the values on the fly, no?

Skysteak:
Note that when making my posts I did mix up the knockoff breacher and the fancy breacher a bit, which as embarrassing as it is to say, is important to admit for anyone reading this and being confused. I  do agree with Runa's points above but consider my arguments slightly to moderately diminished.

Novacat:
Keep in mind that I also buffed the combat/military rigs as well. The main issue is that once you get past 80% or so, each additional % has a much bigger impact. After 90%, it gets a little nutty which is why we want to avoid getting armor up that high.

PR #14376, opened 2023-01-24
Changes all dates to "-240" if higher than 2323.

Runa-Dacino:
In some cases, applies case-by-case logic (such as events referring to unathi war) to make sense.

Also fixed some names and words.

I avoided touching .dmm files for now.


Runa-Dacino:
I fucked up using vscode. Somehow it pushed more than I wanted.

PR #14378, opened 2023-01-24
Ports raptor borgs

SatinIsle:
Adds three raptor variants of dogborgs from Chomp. For the sake of making their names more descriptive, they are called; SecRaptor model V-4, MediRaptor model V-4 and SciRaptor model V-4.
Edit: No longer called the above, they are now all just called Raptor V-4.

image
image

For the sake of my own learning, I've moved the files and code over by hand rather than just doing a direct port. Tested locally and they seem to work fine!
Link to original PR: https://github.com/CHOMPStation2/CHOMPStation2/pull/5318


code/modules/mob/living/silicon/robot/robot_vr.dm
```diff
@@ -140,6 +140,8 @@
 	if(wideborg == TRUE)
 		if(icontype == "Drake") // Why, Why can't we have normal nice things
 			icon = 'icons/mob/drakeborg/drakeborg_vr.dmi'
+		else if(icontype == "MediRaptor model V-4" || icontype == "SecRaptor model V-4"|| icontype == "SciRaptor model V-4") //Added for raptor sprites
```
Heroman3003, line 143: I don't really like this jank. Maybe rename them all to be "Raptor V-4" or other 'universal' name across modules to avoid snowflaking like that?

SatinIsle: Fine by me. Also, I've noticed that these don't have resting sprites, causing them to go invisible when they rest. I'll fix that straight away.

Heroman3003 (approved):

PR #14384, opened 2023-01-26
Adds ability to customize discomfort

Runa-Dacino:
Both pre-made and custom species may now choose up to 10 messages each for warnings while overheated or cold.

If there are no custom messages, game falls back to defaults (which, if the pre-baked species lacks fluff, uses human)

There is a way to reset to defaults.

It is possible to go back and edit any of the 10 messages. The editor shows you the previously set message with an option to cancel out (this avoids over-writing the previous message).

### DIFF
--- a/code/modules/client/preference_setup/vore/07_traits.dm
+++ b/code/modules/client/preference_setup/vore/07_traits.dm
@@ -15,6 +15,9 @@
 	var/custom_ask = null
 	var/custom_exclaim = null
 
+	var/list/custom_heat = list()
+	var/list/custom_cold = list()
+
 	var/list/pos_traits	= list()	// What traits they've selected for their custom species
 	var/list/neu_traits = list()
 	var/list/neg_traits = list()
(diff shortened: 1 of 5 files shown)


code/modules/client/preference_setup/vore/07_traits.dm
```diff
@@ -313,6 +335,63 @@
...
+	else if(href_list["custom_heat"])
+		var/numMessages = pref.custom_heat.len
+		if(numMessages > 0)
+			var/alert = tgui_alert(user, "Edit current messages or add a new one?","Selection",list("Edit","Add", "Cancel"))
+			switch(alert)
+				if("Edit")
+					var/list/indices = list()
+					var/i
+					for(i=1, i<= numMessages, i++)
+						indices.Add(i)
```
Heroman3003, line 347: I'm uncertain about need for this to be an indexed list at all. Something more akin to how bellies handle it where its just a singular text list of descriptions with double linebreaks as separator would work better here. Main reason I used indexing for flavor texts of nutrition and fatness was due to the fact that you both HAD to have all 10 and because the order was very important, unlike here where neither amount you have nor the order matters.

Runa-Dacino: Requested changes done.

I did not make it into a whole proc of its own as that seems far too involved for something so minor.

Heroman3003 (approved):

PR #14386, opened 2023-01-26
sanitizes selection of custom bases in prefs

### LINKED ISSUE #14382: Can create and play as Monkey/wolpin/etc characters
Step 1. Create a promethean.
Step 2. Set your icon to monkey/wolpin or any of the other menagerie of playable species' primitive equivalents.
Step 3. Change your species to custom, you can now play a custom animal-carbon.

Verified that the character is valid to log in and play if all other conditions are met.

Given that these icons aren't available to any non-prommies to select by default, I'm assuming this behavior is entirely unintended.
image

Seris02:
fixes #14382

(also actually enables *correct* body type selection for proteans, because I forgot about a shapeshifter check)

### DIFF
--- a/code/modules/client/preference_setup/vore/07_traits.dm
+++ b/code/modules/client/preference_setup/vore/07_traits.dm
@@ -25,6 +25,19 @@
 	var/dirty_synth = 0		//Are you a synth
 	var/gross_meatbag = 0		//Where'd I leave my Voight-Kampff test kit?
 
+/datum/preferences/proc/get_custom_bases_for_species(var/new_species)
+	if (!new_species)
+		new_species = species
+	var/list/choices
+	var/datum/species/spec = GLOB.all_species[new_species]
+	if (spec.selects_bodytype == SELECTS_BODYTYPE_SHAPESHIFTER)
+		choices = spec.get_valid_shapeshifter_forms()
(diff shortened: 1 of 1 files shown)


code/modules/client/preference_setup/vore/07_traits.dm
```diff
@@ -25,6 +25,19 @@
...
 
+/datum/preferences/proc/get_custom_bases_for_species(var/new_species)
+	if (!new_species)
+		new_species = species
+	var/list/choices
+	var/datum/species/spec = GLOB.all_species[new_species]
+	if (spec.selects_bodytype == SELECTS_BODYTYPE_SHAPESHIFTER)
+		choices = spec.get_valid_shapeshifter_forms()
+	else if (spec.selects_bodytype == SELECTS_BODYTYPE_CUSTOM)
+		choices = GLOB.custom_species_bases
```
Heroman3003, line 36: these direct choices = without copying are dangerous as youre modifying the list down the line. Since lists are objects, it will modify wherever you got it from originally too, if you dont make it a copy somewhere along the line.

Seris02: ah
noted

Heroman3003 (approved):

PR #14393, opened 2023-01-28
Makes cooldown between pain messages a pre-set global preference

Runa-Dacino:
~~Significantly increases cooldown between pain messages from once per (100 TICKS - wound severity) to (120 SECONDS - wound severity).~~

~~This means, at sustaining 1 point of brute damage - you get a warning message that your limb hurts. 2 minutes later, you'll also get notified of this.~~

~~Hopefully, natural healing rate will mean you only see it a few times, reminiscent of a dull ache you get from a papercut.~~

~~Also alters the "ignore anti-spam" check. Originally, there's a 10% chance to ignore cooldown and happen more frequently.~~

~~Now, this starts at 0% and stays there until 10 damage is sustained. This can reach 100% probability if you have a wound with 1000 damage but you're already dead. At 100 damage, you get 10% that previously existed.~~

Per discussion in discord, now works as follows:

- Within the preferences tab, a button now appears you can press to switch between Default and Extended
- You can also do this using character set up under the same tab as UI preferences, jukebox etc.
- If default, you get the classic set up.
  - Added a check to default set-up for the preferences to avoid multi-limb spam
- If extended, the following happens:
  - Under 5 damage, you have 0% chance of ignoring cooldowns.
  - Under 20 damage, you have 1% chance to ignore cooldowns. This is rolled every 2 seconds.
~~  - Having injuries on multiple limbs ignores cooldowns. To avoid spam, this is given a 30% probability to occur. On average, this should mean 1 pain message per 10 seconds, with variation in either way depending on RNG's moods. Slightly higher chance to occur more often than once/10 seconds, due to potential for lag making life ticks take more than 2 seconds.~~~
  - Realized a separate var is probably far less resource intensive than checking probability for EACH LIMB. Added a new var, separate from either classic or extended variants. 
  - Under 5 damage, cooldown is 5 minutes
  - Under 20 damage, cooldown varies from 24 to 10
  - Over 20 damage, cooldown is calculated as default (10 seconds, with 1 second reduction for 10 damage)


Runa-Dacino:
Realized that in the original anti spam set up, you could end up with negative values. Which means the anti-spam stops working.

I rather it keeps working, so I clamped it at a 3 second minimum for maximum severity.

Runa-Dacino:
That should be the last of my force pushes.

code/modules/organs/pain.dm
```diff
@@ -59,7 +59,7 @@
 			if(91 to 10000)
 				flash_pain()
 				msg = "<font size=3>OH GOD! Your [damaged_organ.name] is [burning ? "on fire" : "hurting terribly"]!</font>"
-		custom_pain(msg, maxdam, prob(10))
+		custom_pain(msg, maxdam, prob(clamp(maxdam/10,0,100)))
```
TankTheBirb, line 62: This should be a lot lower, I think this is organ damage, so the max health of most limbs is 60. 30 is a bone break in most cases, so you should probably get the most pain messages around that amount of damage on a limb.

Heroman3003:
My general opinion is that this will basically make pain messages even for hardest injuries be no more common than once a minute and thats kinda very bad. The 2 minute timer is absurdly long for a baseline on something like that. But also will want to hear other staff's opinions on this.

code/modules/organs/pain.dm
Runa-Dacino: Broken limbs should cause Shock damage too, which should trigger other pain messages as well (ideally). 

However, could increase this probability.

Do note, this works on top of existing pain messages. Like - you are guaranteed 1 pain message per 90 seconds, plus a 3% chance EVERY LIFE TICK (so like every second or so) to ignore that cooldown.

Even at 3% chance, considering how often this is checked? I think it should triggered frequently enough.

Runa-Dacino:
> My general opinion is that this will basically make pain messages even for hardest injuries be no more common than once a minute and thats kinda very bad. The 2 minute timer is absurdly long for a baseline on something like that. But also will want to hear other staff's opinions on this.

I could wrap this into an if-statement.

Something like

swich(power)
    if(power < 10)
        cooldown = 300 seconds
    if(power < 30)
        cooldown = 100 seconds
    if(power >= 30)
         cooldown = clam((60 - power) seconds, 3, 60)

TankTheBirb:
General notes from testing:
- Deals with minor injuries to single limbs fairly well for being less annoying and spammy.
- The amount of pain messages you get for single limb, big damage is much too low. It should be higher to convey "hey dummy you got hit hard".
- Multiple limb small damage doesn't generate nearly as many pain messages as it should. You're able to get 6 damage per limb, for 66 total damage, while not actually getting more than a pain message ever couple of minutes.

code/modules/organs/pain.dm
Heroman3003: Its called every 2 seconds to be specific. Also its not 3% chance to ignore cooldown, its 3% chance to even trigger at all IF cooldown isnt currently in place. So its 2 minutes of cooldown and then a bunch more time with however it takes to trigger the message on top of that.

Runa-Dacino: You misunderstand.

`if(force || (message != last_pain_message) || (world.time >= next_pain_time))`

IF ANY of the following is true:
- We roll the 3% probability
- Cooldown expired
- It's NOT the last limb that was injured

This is explicitly a "ignore the cooldown."

Heroman3003:
My own thoughts: only damage less or equal 5 should get special treatment in terms of rarer appearances. Anything higher works good with existing pain message commonality.

Runa-Dacino:
Realized that maybe doing it with a var is less intensive.

Replaced probability based anti-spam for multilimb injury with a "global cooldown", as it originally was.

This allows for specific global cooldown based on injuries, too.

TankTheBirb:
This looks like all feedback was adjudicated with functional edits to how the system works by making it a preference that people toggle on. I think this is good to merge if no new critique is brought up.

BigBababooey:
I mean I feel like it's incredibly easy to just make a tab that doesn't display the pain text to flick to when it's ongoing, but some people might want this so good for them. No critique to add really

Heroman3003 (approved):

Heroman3003:
PR is approved, just fix the conflicts with the preference files and its gucci

Runa-Dacino:
> PR is approved, just fix the conflicts with the preference files and its gucci

Done!

I used the web-editor for "resolve conflicts", and moved the conflicting lines to be behind the player_tips lines.

I commited the merge but I cannot see it. I imagine it's hidden to all but maintainers?

PR #14406, opened 2023-01-30
better subtle, with targets including single tile and the normal 1 tile radius

Seris02:
and belly + preds
that was the most complicated logic in this whole thing, jesus

_**this should really, probably get a test merge or something I dunno, I can't test all the cases by myself on local**_

at the very least I know single tile subtle and 1 tile radius subtle both work all the time
and the all in belly + preds target also works, though I don't know for certain if it works for multiple preds, but it really, really should
objects inside bellies, like possessed items, also work for sure

ALSO, not certain if this needs vorestation comments?? because it's in a _vr file

### DIFF
--- a/code/modules/mob/say_vr.dm
+++ b/code/modules/mob/say_vr.dm
@@ -26,6 +26,38 @@
 	else
 		usr.emote_vr(message)
 
+/mob/proc/get_all_in_bellies()
+	var/list/mobs_in_belly = list()
+	for (var/obj/belly/belly in contents) //dogborg sleeper, etc
+		for (var/mob/M in belly.contents)
+			mobs_in_belly |= M.get_all_in_bellies()
+			mobs_in_belly |= M
+	return mobs_in_belly
+
+/atom/proc/get_ultimate_mob()
+	var/mob/ultimate_mob
+	var/atom/to_check = loc
+	var/n = 0
+	while (to_check && !isturf(to_check) && n < 6)
+		if (ismob(to_check))
+			ultimate_mob = to_check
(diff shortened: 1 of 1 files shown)


Seris02:
that wasn't my faaault that it failed

Heroman3003:
tbh with psay and pme im not sure if this is entirely necessary to have but eh?

Seris02:
> tbh with psay and pme im not sure if this is entirely necessary to have but eh?

yeah but what if I want to send it to people in belly who aren't absorbed and don't want to absorb said people

Poojawa:
fun fact

if you log out from byond while still hosting a private server, you can then join your game again as a guest. This way you can test if your code works from multiple perspectives

Seris02:
> fun fact
> 
> if you log out from byond while still hosting a private server, you can then join your game again as a guest. This way you can test if your code works from multiple perspectives

did that to test it but could only get up to one more client running

Very-Soft:
Bringing up a dialogue to ask the range every time anyone sends a subtle is deffo no bueno, that will absolutely make normal subtle use more annoying, and should change if nothing else.

I don't super see the point in 1 tile either though. It sort of feels like, if you're wanting THAT level of privacy, you really ought to just move so you're not adjacent to other people.

Novacat:
If there are only two options the better solution would be to just have a tablesubtle or something like that.

Seris02:
> If there are only two options the better solution would be to just have a tablesubtle or something like that.

there are three, along with having specific targets as well

think I'll just close this and PR it downstream as a separate verb

PR #14414, opened 2023-02-02
Fixes NIF repair tool using up nanopaste when full

Runa-Dacino:
There was an unreported bug that was discovered yesterday while I was playing.

When using nanopaste on an adv. nif repair tool, you should only be able to expend 4 charges to hit the 60 maximum.

Unfortunately, the code was written in a way that used up the nanopaste charge before checking if the repair tool was actually full. This had the consequence of wasting up to 4 whole charges of nanopaste if you click too quickly.

~~Replaced "use(1)" proc with "(get_amount()". Only using "use(1) after confirming we have free space left.~~

Per maintainer advice, dropped get_amount() and just moved the np.use(1) to the right side of the if() check. Works.

### DIFF
--- a/code/game/objects/items/devices/advnifrepair.dm
+++ b/code/game/objects/items/devices/advnifrepair.dm
@@ -24,7 +24,7 @@
 /obj/item/device/nifrepairer/attackby(obj/W, mob/user)
 	if(istype(W,/obj/item/stack/nanopaste))
 		var/obj/item/stack/nanopaste/np = W
-		if(np.use(1) && supply.get_free_space() >= efficiency)
+		if((supply.get_free_space() >= efficiency) && np.use(1))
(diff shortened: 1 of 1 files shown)


code/game/objects/items/devices/advnifrepair.dm
```diff
@@ -24,13 +24,15 @@
 /obj/item/device/nifrepairer/attackby(obj/W, mob/user)
 	if(istype(W,/obj/item/stack/nanopaste))
 		var/obj/item/stack/nanopaste/np = W
-		if(np.use(1) && supply.get_free_space() >= efficiency)
```
Heroman3003, line 27: A better way to achieve that would be to swap the order of this. Put the check for us being empty first, and use() second. That way it will only attempt to use if first check has passed successfully.

Runa-Dacino: Done as requested.

Heroman3003 (approved):

PR #14417, opened 2023-02-03
adds admin ghost privacy override preference

Seris02:
adds a preference so that admins can toggle whether they can hear ghost privacy hidden whispers/subtles anyway (they could already by the way, this just makes it so they can turn it off)

also makes it so that when ghost privacy is overridden it prepends a "(H)" to the message (for "hidden")

### DIFF
--- a/code/modules/admin/admin_verb_lists_vr.dm
+++ b/code/modules/admin/admin_verb_lists_vr.dm
@@ -19,7 +19,8 @@ var/list/admin_verbs_default = list(
 //	/client/proc/deadchat				//toggles deadchat on/off,
 //	/client/proc/toggle_ahelp_sound,
 	/client/proc/toggle_admin_global_looc,
-	/client/proc/toggle_admin_deadchat
+	/client/proc/toggle_admin_deadchat,
+	/client/proc/toggle_admin_ghost_privacy
 	)
 
 var/list/admin_verbs_admin = list(
--- a/code/modules/client/preference_setup/global/setting_datums.dm
+++ b/code/modules/client/preference_setup/global/setting_datums.dm
@@ -420,7 +420,13 @@ var/list/_client_preferences_by_type
 	key = "CHAT_RLOOC"
 	enabled_description = "Show"
 	disabled_description = "Hide"
-
+//VOREStation edit
+/datum/client_preference/holder/show_subtles
+	description ="Admin ghost privacy override"
+	key = "CHAT_HSUBTLES"
+	enabled_description = "Enabled"
+	disabled_description = "Disabled"
+//VOREStation edit end
 /datum/client_preference/holder/show_staff_dsay
 	description ="Staff Deadchat"
 	key = "CHAT_ADSAY"
--- a/code/modules/client/preferences_vr.dm
+++ b/code/modules/client/preferences_vr.dm
@@ -120,3 +120,18 @@
 	SScharacter_setup.queue_preferences_save(prefs)
 
 	feedback_add_details("admin_verb", "TSoundMentorhelps")
+
+//General
+/client/proc/toggle_admin_ghost_privacy()
+	set name = "Toggle Admin Hidden Subtles"
+	set category = "Preferences"
+	set desc = "Toggles seeing subtles/whispers for clients with the ghost privacy option on."
+
+	var/pref_path = /datum/client_preference/holder/show_subtles
+
+	if(holder)
+		toggle_preference(pref_path)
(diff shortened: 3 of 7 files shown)


Heroman3003:
I don't particularly see the point of this, as it kind of bloats the preferences while admins already have a tool to stop monitoring all admin stuff - deadminning. Not really against it, just dont see the 'why' of it.

BigBababooey:
Admittedly I can't really see any benefit this provides us other than potentially making certain staff members feel more comfortable about not violating people's privacy. But, at the same time one of the parts of being a staff is the trust that you won't be doing that anyways so... it's a bit of a moot point. Because those same people probably want to turn off their ghost ears entirely so they don't overhear people's stuff at all.

Idk, it doesn't really add anything I see much benefit in. Staff aren't making a habit of leaping onto people's subtle stuff anyways, I don't think any of us want to see it unless we're actively monitoring a situation.

That said, there's so much verb and preference bloat as a staff member you also just have to learn to tune a lot out, so I'm not really notably opposed to it either.

Seris02:
Mostly it's a difference between not caring about seeing what people don't care if you see, and not caring about seeing what people don't want you to see

BigBababooey:
I'm not sure I really follow what you mean by that. We take privacy of users very seriously and such things are going to be respected universally unless there's something actively in need of enforcement. I cannot say I've ever seen a single staff member delve into people's subtles outside of an active investigation over an incident we know transpired and need to look into whether they were normally hidden or not. And this wouldn't really change that, all it's doing in effect is letting a staff member know that the end user has that preference on to hide their thing that we're still seeing. 

I trust the members of our team to not be violating people's privacy regardless, that's why they were given these tools, because they were trusted with them. I could understand other people not having the same faith and trust but if that's the case, I don't necessarily see how this alleviates any of that either since again, it's the ability to not look at a thing we already have the ability to not look at. Because if I don't wanna be spammed by subtles I'm turning off ghost ears anyways while I look around, I'm not well served by only being able to selectively see the people who are more comfortable with me seeing their posts as an option.

Ultimately I really don't see it giving much tangible benefit from my perspective. But, once again I don't think it's addition is a particular detriment either other than some minor bloat on a very filled already preference tab. There could be staff who would like it as a feature that I'm presently unaware of I suppose!

Seris02:
> I'm not sure I really follow what you mean by that. We take privacy of users very seriously and such things are going to be respected universally unless there's something actively in need of enforcement. I cannot say I've ever seen a single staff member delve into people's subtles outside of an active investigation over an incident we know transpired and need to look into whether they were normally hidden or not. And this wouldn't really change that, all it's doing in effect is letting a staff member know that the end user has that preference on to hide their thing that we're still seeing.
> 
> I trust the members of our team to not be violating people's privacy regardless, that's why they were given these tools, because they were trusted with them. I could understand other people not having the same faith and trust but if that's the case, I don't necessarily see how this alleviates any of that either since again, it's the ability to not look at a thing we already have the ability to not look at. Because if I don't wanna be spammed by subtles I'm turning off ghost ears anyways while I look around, I'm not well served by only being able to selectively see the people who are more comfortable with me seeing their posts as an option.
> 
> Ultimately I really don't see it giving much tangible benefit from my perspective. But, once again I don't think it's addition is a particular detriment either other than some minor bloat on a very filled already preference tab. There could be staff who would like it as a feature that I'm presently unaware of I suppose!

So- I'm not actually doing this for here. This is for my own sake, with my admin perms downstream, so that **I** can turn it on and not see them, I don't care about admins in general seeing what I put out there. Will happily PR it downstream if you guys don't want it.

BigBababooey:
Yeah see there it is. And that's like totally fine, I just was struggling to see what it really did for us, but I fully understand there being individuals, yourself included who would benefit from it, sure.

I don't see an issue with having it, but I'm just a bozo admin. So I'm not vetoing or approving, just had shared thoughts because I like to ramble

TankTheBirb:
Not sure if this is really desired for vorestation. I'm probably the person who would most use it and I don't think I'd ultimately use it. A lot of what this would protect against is already being said in emotes or specialized say messages. Or sometimes in LOOCs. So I could try to filter it down a notch but there are more absolute tools for not seeing content I don't wanna see. I'm not speaking for the team here, but this may be best as a downstream PR for where it's intended and desired.

lm40:
Yea I wouldn't use this, I don't really think it is necessary here.

Seris02:
aight, seems to be fairly unanimous

PR #14424, opened 2023-02-04
Cleaning up DMI/Figuring out what's missing.

Yatiyaya:
Cleaning up DMI files and marking what sprites we currently have missing
(so we/I can figure out what still needs worked on.)

As well as correcting mismatched sprite names.


Yatiyaya:
For now, Uniform.dmi is cleaned up, only thing missing are sprites they don't have, that citadel/vore has.

Yatiyaya:
https://cdn.discordapp.com/attachments/187013248309002240/1071493734023319552/image.png

example of what I mean.... Marked all missing sprites as _MISSING and letting them fallback to the "no name" default

PR #14429, opened 2023-02-04
[MIRROR] Adds a hot pink recolor of the retro shroud

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/8924
--------------------
now both genders of super sentai dude are represented


Heroman3003:
Broken clothing PR due to file system difference

PR #14445, opened 2023-02-05
makes it so you can't walk on small people when you shouldn't

Seris02:
title

previously, you could just place someone on a table and instantly step onto it with the right intent

### DIFF
--- a/code/modules/vore/resizing/resize_vr.dm
+++ b/code/modules/vore/resizing/resize_vr.dm
@@ -242,6 +242,19 @@
 		return TRUE
 	return FALSE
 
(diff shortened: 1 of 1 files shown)


code/modules/vore/resizing/resize_vr.dm
```diff
@@ -242,6 +242,19 @@
 		return TRUE
 	return FALSE
 
+/mob/living/proc/can_pass_into(turf/n, var/prey = null)
```
Heroman3003, line 245: Honestly, this proc is better served just inserted into general handle_bump_other than being its own proc, for better readability, imo

Seris02: fair

PR #14446, opened 2023-02-05
skyrat outfit port continuation (adds to loadout)

Hatterhat:
continues #14418 -
- adds all the outfit bits to loadout
- fixes some grammar here and there
- also does some repathing for inheritance's sake


Heroman3003:
Just to note, but we alerady have items wearable in some clothing slot AND attachable to uniform as accessoty as same time. I suggest trying to use that for maid arm things.

Hatterhat:
@Heroman3003 done

Heroman3003:
Nice. It wasnt really required per se,  just was suggesting since it was a thing.

PR #14455, opened 2023-02-06
fixes nanoform limbs specifically from changing their child limbs

Seris02:
yeah so that PR from earlier in Polaris was a mistake on my part woops
I thought I tested it on Polaris or I woulda realised it was only a problem with nanoform organs
thankfully I got informed when it came downstream so uh
woops
here's the real fix


Seris02:
the integration tests failure wasn't my fault, aaaaaaaa

PR #14456, opened 2023-02-06
Implements periodic tips for (new) players!

Runa-Dacino:
- ~~Using mob/living/life() to call the proc as we already send this only to mobs.~~
- ~~Chose mob/living over carbon to cover silicons and maintpred/other simplemobs.~~
- Uses a subsystem dedicated to periodic tips. Lowest fire priority, fires only during game
- By default, starts ON. Can be turned on/off at any time under preferences tab. The players are informed how to do this on the initial, 10 minute tip.
- ~~Only those mobs that have a client receive this~~
- Tries to ensure every new tip is different from the last.
- Should be easy to add new tips as time goes on, just need to balance the weights when adding a new one
- Could probably add custom per role tips, but I do not know how to check for player job/role so I went for the basics for now.
- Advice list needs significant expansion, but the current size should be good to go onto server already.
- Advice list picks from categories with separate weights to what's inside the category.
- Adds a new verb for players to request tips outside of cooldown. Players can specific tip subtype (none, general, gameplay, roleplay, lore), overriding the category randomization.
- Currently fires after 10 minutes of roundstart, and every rand(45,75) afterwards. Timer is global between players
- Everyone who did not opt out receives the same advice.

This is how it looks like:

image

(while testing, used 10 sec initial, 30-60 second range afterwards)


Heroman3003:
While I am ambivalent towards the feature as a whole, I'd honestly prefer to see it implemented as, perhaps a subsystem or something along those lines, rather than another batch of vars on all mobs that tick on life; and procs tied to it being a mob. Not really necessary, but seems like generally better implementation to me.

BigBababooey:
I like this overall! I think this is a good way to help get a lot of information to people to potentially learn about things.

I do however think some of the tips should be rewritten to be less 'exact' as it were. The lore ones can at places come off as fairly heavily implying/suggesting that what is being said is gospel, and while I love our lore as a loremaster. I feel no urge to, for example, make people feel as if they have to live in the dingy underground colony, and that they would be abnormal for not doing so.

This can easily be solved by providing that information in a simply different manner that's like "hey, this is what's nearby" but, I think it would be good to take a look at some of that stuff. In the same vein with the roleplay tips I would avoid wordings like "This means you _should_ act in X way"

Very-Soft:
I think I agree with heroman, I would prefer if this was something more like a subsystem that just sends tips to people with the pref enabled periodically

it would probably be cheaper on processing if it just has one timer, and fires off the same random tip to everyone now and then

it would be good too, because then everyone with the tips on would see the same thing, and could talk about it with one another in case they have questions

code/modules/player_tips_vr/player_tips_list_vr.dm
```diff
@@ -0,0 +1,69 @@
...
+			return pick(
+				prob(50); "[info] Having difficulty finding scenes? The number one tip that people should take for finding scenes is to be active! Generally speaking, people are more likely to interact with you if you are moving about and doing things. Don't be afraid to talk to people, you're less likely to be approached if you're sat alone at a table silently. People that are looking for scenes generally like to see how you type and RP before they'll start working towards a scene with you.",
+				prob(50); "[info] Please avoid a character that knows everything. Having only a small set of jobs you are capable of doing can help flesh out your character! It's OK for things to break and fail if nobody is around to fix it - you do not need to do others' jobs.",
+				prob(25); "[info] Embrace the limits of your character's skillsets! Seeking out other players to help you with a more challenging task might build friendships, or even lead to a scene!",
+				prob(25); "[info] Participating in an away mission and see something acting strange? Try emoting or talking to it before resolting to fighting. It may be a GM event!",
```
TankTheBirb, line 57: "...to it before resolting to fighting."
I think you meant "...to it before resorting to fighting."

Heroman3003 (approved):

Heroman3003:
Code seems fine, but I haven't looked at tips themselves in depth yet. I'll trust it was tested and others will review tips more properly.

PR #14468, opened 2023-02-08
Greyscaled vox to be more modular (Re-enabled marking whitelist)

Yatiyaya:
A few more edits towards the suit and eyes.dmi for vox
Added in modular greyscale to vox(included removing tail from torso sprite... Voxxy can has tails now)

Enabled marking whitelist, cleansed marking whitelist (Teshari markings only enabled for teshari+custom species, Vox markings for vox... All other markings enabled for all species. Pulls werebeast markings from non-werebeasts)


Runa-Dacino (commented):
I think all restricted markings should include species_custom. 

The only reason we should restrict markings is to ensure it works with a specific icon base. species_custom has the option to pick custom icon bases.

Yatiyaya:
Currently the only restricted markings are teshari, and they include the custom_species.
All others had their restrictions removed

Heroman3003:
Add xenochimera and Alraune to all whitelists that have Custom. Those two species select bodytype too. Protean and Promethean too potentially?

Yatiyaya:
Overhauled the marking system so that it checks for custom species (and shapeshifters) using the correlated sprite/body form.

Yatiyaya:
AKA, using teshari custom base allows teshari markings to be used. hides them when not selected

https://cdn.discordapp.com/attachments/405257702663782400/1074097878173220976/image.png

Heroman3003:
That seems to be general staff opinion. Whitelist part should either be commented out or simply made into a, say, config thing disabled by default or something among those lines. The supposed 'potential taur markings' in future will most likely use a system unrelated to species-based whitelisting anyway, so that one will be reviewed on its own merits.

PR #14470, opened 2023-02-10
Cloaks, undies, the Antediluvian, and some other bits

Burrito-Justice:
Adds a lot of clothing options to the loadout, including:

- The Antediluvian set from CitRP, which is a corset, cape, loincloth, bracers, and legwraps, all of which are separate and interchangeable.
- Many cloaks like the rough half cloak, shoulder capes, mantles, and two chaplain-related cloaks.
- New underwear options like the neko bra/panties, tape, leotard, shibari rope, sarashi bandages, midriff top, and stirrup socks.
- Another colorable qipao, which separates the gold trim to an accessory so that you can color the qipao while preserving the trim.
- A wedding veil, which can be recolored. I'm sort of surprised there wasn't already one.

In addition to these, the halo's sprite has been brightened, from 199,200,200 to 255,255,255. Colors should look more true now.
image
image
image


Burrito-Justice:
Icon conflicts should be fixed? I'm not very familiar with git, sorry

PR #14481, opened 2023-02-12
Adds new machine that mass-produces autoinjectors, maps it in

Runa-Dacino:
**What it does**
Adds a new machine, which Science can make and also starts mapped in on all 3 maps' triage areas.

This machine takes plastic sheets OR autoinjectors and a glass reagent container.

- ~~It uses the plastic sheets to make either a normal or large autoinjector, then fills them with chemicals.~~
- It takes autolathe produced autoinjectors, then fills them with chemicals
- Alternatively, you can make autoinjectors on the spot at 3x the cost of making in an autolathe (500->1500 for large, 10->30 for small)
- It can make multiple autoinjectors.
- It can give the autoinjectors custom names.
- It warns you if the autoinjectors will contain less than maximum chemicals, letting you abort.
- It can recycle spent autoinjectors (prevents doing so while they contain chems)
- It is a wall mounted machine
- It was mapped in to chemistry & triage of each map (triage, to encourage EMTs to make them as they need them without invading chemistry)
- Alt-Clicking removes beakers, for ease of use!

**Why we need this**

- Making autoinjectors, at this time, is a massive chore I would not wish on anyone. You need to manually fill them all with syringes.
- Autoinjectors are useful for crew that are not trained paramedicine, allowing them to self-treat in space without having to break character, or go untreated.
- The reason a more ready access to auto-injectors is required is that the presence of medical staff can make or break whether people can participate in an away mission
- - Medbay staff may only have 1 person and cannot join with clear conscience
- - Medbay staff may not be interested
- - On breathable away missions, this isn't an issue as medical crate/first aid kits do everything you need and are usable by anyone.
- - However, on space-borne missions, lack of injections can almost guarantee death for the intended rag-tag groups.

**Why not just add it to ChemMaster**

TGUI scares me. 

**Mapping Locations:**

Tether:
image

SD:
image

RP:
image

**The Icon**

image


Heroman3003:
I am both uncertain about part where cargo's no longer needed to produce autoinjectors and generally 'a separate machine for ultra specific purpose' aspect tbh.

TankTheBirb:
> I am both uncertain about part where cargo's no longer needed to produce autoinjectors and generally 'a separate machine for ultra specific purpose' aspect tbh.

I think the ultra specific purpose is a valid criticism but I can understand why rolling this in as a chemmaster functionality would be difficult. Given that it was made to fit in the SD’s medbay without being clunky, I think it’s workable.

As to the why for excluding cargo in this case, it’s more that auto injectors are generally an inferior option for players than giving someone a syringe or two and a bunch of bottles. Functionally they act the same but the bottles have 4x the reagent capacity for the same size as a large auto injector.  So people don’t request the empty auto injectors anyways to make or to use. You also can’t rename them without a label maker, which has its own limits. 

Having a machine that makes them and getting a few sheets of plastic from cargo encourages their use as an “unskilled crew” healing option over giving someone a box of medical supplies. And since they don’t start with plastic anywhere but as some tables, they’re going to need to go to cargo anyways. 

I think it’d be fair to start with 3 or so plastic sheets by the machine as well but there’s more than a few ways to get plastic for anyone who really wants it, so they don’t need to have round start plastic really.

Runa-Dacino:
Made it use autoinjectors rather than plastic so cargo can get involved still, making it simply be a quickloader.

Skysteak:
> Made it use autoinjectors rather than plastic so cargo can get involved still, making it simply be a quickloader.

How many does it start with by default?

Heroman3003:
Thats a better solution to that aspect, but I am still somewhat skeptical on the separate machine from chemmaster part. Id like other staff to weigh in on that.

TankTheBirb:
I personally liked the concept more when it used plastic instead of empty auto injectors, however I can see the merit in trying to still involve cargo (or anyone with an autolathe) in the process. I think using plastic sheets cuts down on the tedium of the process of making auto injectors for people since you don't have to load the machine with  them, just the plastic itself. You also don't need to transport the fourty bajillion auto injectors you may want back to medical. On the SD and RP it's a short walk, on Tether it's a hike unless you make science do it (in addition to everything else they do). Also you still have to get the plastic and aside from disassembling tables, cargo is the best source of plastic anyways.

I think the idea's justification for existence is several factors.

-  Auto injectors are a huge PITA to make. You have to make the thing in a bottle, then you have transfer the bottle reagents into the auto injector. and you have to repeat this for 4-12 auto injectors to empty a full bottle. Then you have to label maker it since it doesn't come with a name on it nor an easy way rename it en masse like pills. This solves all those by doing it in fewer clicks and enabling mass production with labels.
- Autoinjectors distribute medical supplies amongst a team with no medic rather than centralizing it to the person who may or may not know how to use it. It also removes the old explo problem of "everyone knows medical" since having a few auto injectors and some trauma kits is a lot more reasonable than everyone and their mother having full medical loadouts with them. Because they're a pain to make, nobody makes them, and people just give away teams a belt of drugs and a syringe.
- This allows someone who is in medical to interact with and prep an explo team without going. It removes the pressure to go from the medic who may not want to, and it removes the anxiety of having no medic from the away team themselves. This is not a new capability, but a good QoL change. Otherwise, it's the "throw drug belt and syringe at team and tell them to go away" strategy.
- Auto injectors have interesting drawbacks and positives. They use instantly and are one use, you just use>drop and and go back to what you were doing or continue first aid. But you can only carry so many, even if you are justified in spreading them across a team instead of one person holding every bottle. Encouraging their use in content like away teams or for supplying miners with custom concoctions is something that should be encouraged. This machine makes this a lot easier.
- They're nice roleplay tools and make some degree of distinction between "character who knows medicine" and "player who knows medicine and is controlling this character". Given that we want to disincentivize "knows everything" characters, this is a good way to do it for something as mechanics heavy as medical is.

Granted, if someone comes along and wants to add this functionality to the ChemMaster, that is probably the better implementation. But what this PR adds is perfectly fine as it works well for QOL and adds in the ability to mass name auto injectors.

Skysteak:
> I personally liked the concept more when it used plastic instead of empty auto injectors, however I can see the merit in trying to still involve cargo (or anyone with an autolathe) in the process. I think using plastic sheets cuts down on the tedium of the process of making auto injectors for people since you don't have to load the machine with them, just the plastic itself. You also don't need to transport the fourty bajillion auto injectors you may want back to medical. On the SD and RP it's a short walk, on Tether it's a hike unless you make science do it (in addition to everything else they do). Also you still have to get the plastic and aside from disassembling tables, cargo is the best source of plastic anyways.

I agree with this post and I specifically wanted to highlight this in particular. I think that be it its own machine or part of the Chemmaster, there should be an aim to have the plastic cost wherein it doesn't rely on interacting with Cargo/Science, but it's encouraged to do so. I don't know how viable that is but I think it's a good way to handle it, and also allows for upgrades to this machine/the Chemmaster to reduce that, which in turn encourages even more Science interaction. 

Now when it comes to collecting plastic I suppose there is going to be the anecdote regarding marble spears, but two things on that; A. If that behavior is disruptive then it can be dealt with ICly or if severe enough OOCly. B. If someone is going to the trouble of deconning stuff instead of wanting to talk to (including a a fax) for a department? They're probably just not going to bother if  they cannot do that, instead of going 'Awww phooey, I guess I have to talk to someone now'. 

So yeah I think plastic cost is way better, with perhaps options for varying sizes, so if you needed to you could scrounge plastic for some quick and easy 5u autoinjectors, but if you wanted something really strong then well you're encouraged to interact. Best of both worlds imo.

Runa-Dacino:
> > I personally liked the concept more when it used plastic instead of empty auto injectors, however I can see the merit in trying to still involve cargo (or anyone with an autolathe) in the process. I think using plastic sheets cuts down on the tedium of the process of making auto injectors for people since you don't have to load the machine with them, just the plastic itself. You also don't need to transport the fourty bajillion auto injectors you may want back to medical. On the SD and RP it's a short walk, on Tether it's a hike unless you make science do it (in addition to everything else they do). Also you still have to get the plastic and aside from disassembling tables, cargo is the best source of plastic anyways.
> 
> I agree with this post and I specifically wanted to highlight this in particular. I think that be it its own machine or part of the Chemmaster, there should be an aim to have the plastic cost wherein it doesn't outright require interacting with Cargo/Science, but the mechanics encourage one to do so. I don't know how viable that is but I think it's a good way to handle it, and also allows for upgrades to this machine/the Chemmaster to reduce that, which in turn encourages even more Science interaction.
> 
> Now when it comes to collecting plastic I suppose there is going to be the anecdote regarding marble spears, but two things on that; A. If that behavior is disruptive then it can be dealt with ICly or if severe enough OOCly. B. If someone is going to the trouble of deconning stuff instead of wanting to talk to (including a a fax) for a department? They're probably just not going to bother to begin with if they cannot do that, rather than going 'Awww phooey, I guess I have to talk to someone now'.
> 
> So yeah I think plastic cost is way better, with perhaps options for varying sizes, so if you needed to you could scrounge plastic for some quick and easy 5u autoinjectors, but if you wanted something really strong then well you're encouraged to interact. Best of both worlds imo.

I could revert to allowing both maybe?

With straight plastic, there's a waste (rather than 500, you need 1000 for big ones or sth).

With injectors, you can save plastic when mass producing.

Skysteak:
> I could revert to allowing both maybe?
> 
> With straight plastic, there's a waste (rather than 500, you need 1000 for big ones or sth).
> 
> With injectors, you can save plastic when mass producing.

That could be an option as long as it remains sane. I suppose a question both to you and any staff reading this is:

Do you want people to specifically be asking Cargo for the autoinjectors, or are you happy if the interaction is akin to Science asking Cargo for materials? 

I say this because that's realistically what's going to happen, as 99.9% are not going to go 'Oh golly gee I'm gonna need to bulk acquire these from Cargo!'. No they'll go: "Cargo? Could I have more plastic please?'. Thus with this assumption, it comes down to if that's desirable. 

In my opinion it probably will probably generate slightly more interaction from Medical to Cargo than it does now, because  most people either don't know that you can get autoinjectors from Cargo (or Science, because Science can makeadasdasdfaaaa-), or know but see it as too much of a pain. I think making plastic part of the process akin to Phoron for Chemistry will keep inter-departmental interaction and have autoinjectors made more overall, thus in turn filling other goals of this PR.

Heroman3003:
Alright, it appears general staff consensus is that machine is fine in the current state.

On personal level however, I do have a suggestion (do not think it a requirement - you dont have to do that, just something i feel would improve it) - to move it into chemistry, so that it is where appropriate to its purpose, and refluff it into being a wall machine rather than a floor one, to occupy less space as to befit its niche nature.

Skysteak:
I still feel the original version was absolutely fine tbh, though yeah I do agree that making it a wall machine or integrating it into TGUI would be lovely.

Skysteak:
> Currently it's a tabletop rather than wall or floor one, but I can do that.
> 
> For plastic vs injectors, how's the "make both work, but plastic more expensive"? It might make stuff more janky, but hey... something to keep myself busy with.

I've said this elsewhere but just so it's on the record: 

I personally would say yes with perhaps a 100% to 200% extra  cost versus the Autolathe.

This permits it people to get the plastic themselves if they desire to do so, or needs must, whilst also encouraging interaction. After all if you're asking Cargo for plastic than mission accomplished. 

Let's be real here, _I think many people don't even know that you can print autoinjectors._

Runa-Dacino:
Fuck you tether

PR #14492, opened 2023-02-15
Make Dogborgs modular

Sharkmare:
Modularizes Dogborgs so you can just enable/disable specific borgs in the DME rather than having to comment them out. Also makes them manually add themselves tot heir respective robot  lists so we dont have to modify Global.dm every time we add a new borg.

This should make borgs easier to add and maintain as well as cherrypick for dowstreams.


Heroman3003:
My personal thought is that Im not really interested in every borg module having its own file. I don't mind the big modules files being split up, but I'd rather see it being split up by the department rather than by individual module, with, say, all four medical modules into medical file, including both dogborgs and non-dogborgs in there.

I also don't really like them being 'manually added' to the global list like that at all. Makes maintaining said lists a mess, imo.

Sharkmare:
> My personal thought is that Im not really interested in every borg module having its own file. I don't mind the big modules files being split up, but I'd rather see it being split up by the department rather than by individual module, with, say, all four medical modules into medical file, including both dogborgs and non-dogborgs in there.
> 
> I also don't really like them being 'manually added' to the global list like that at all. Makes maintaining said lists a mess, imo.

I can see departmental combination files as a nice idea yeah. as long as they are segmented in fille properly

I dont really see the downside of the Global being dynamically written to since its a pretty specific listt that if you want to know where its data comes from you can actually easily code search by its name to find all occurences.

Also means that we dont have conflicts if multiple people are working on a borg since they are no longer editing thte same files

Heroman3003:
New borg moduels in general is a big rarity, so I doubt the conflicts there is problematic in notable way, and in rare case someone will, the conflict resolution would be very simplistic, since its all just line additions, and not, like, icon files. And again, for global lists that are 'static' like that I value ease of readability more regardless. Again, other staff may override me but for now I stand with both assessments - the current file separation is overdone and should be more by-department and with inclusion of 'normal' borgs; and that global list is better off just being defined in its entirety rather than having additions to it spread across half a dozen files.

PR #14495, opened 2023-02-16
[MIRROR] Reworks on-mob overlay icon generation.

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/8920
--------------------
- Ports `/datum/species/proc/get_offset_overlay_image()` and `/proc/check_state_in_icon()` from Nebula.
- Reworks `make_worn_icon()` into `get_worn_overlay()` and folds in species slot adjustments. 
- Removes some uncached icon gen.
- Reworks `get_species()` into `get_species_name()` and `get_species()`.
- Reworks `apply_custom()` and `apply_addblends()`.

So the TL;DR for this system is that it will offset any on-mob icon that has no spritesheet icon_state by a given amount per direction, per species. This allows for Teshari, for example, to offset inhands to line up with their sprite. It also means instead of getting a default state or an invisible state for human gear with no Teshari sprite sheet entry, you instead get a very ill-fitting adjusted human icon. In some cases (like dresses) this can look very bad, in other cases (like rigsuits) it can actually look pretty workable as a placeholder.

Some examples:
image
image
image


Heroman3003:
I would have liked this but at this point its better to just close.

PR #14497, opened 2023-02-16
cap crystal ooc escape fix

Seris02:
title

if cap crystal was inside a storage item and you ooc escaped (which seems quite likely), you get stuck inside the storage item


Seris02:
> I'm not sure I really understand why you went as complicated with this as you did. Why not just do 'drop_loc = get_turf(src)' instead of checking for whether we're in a bag and all that? That way it just always does the same thing, regardless of how nested inside of whatever we might be.

because... I'unno. In belly or something, or in closet, etc. More consistent with the environment. I can change it if you want.

Heroman3003:
Considering its 'ooc escape' I do think it getting you out of places is fine.

Very-Soft:
I just mean, get_turf() always goes and finds the turf you're on or whatever you're in is on, right?

so like, why do the work of even checking to see if you you're in a bag or whatever, just always do get_turf()

PR #14498, opened 2023-02-16
Alien slug ears and tail

JoltDerg:
Adds new markings ported over from Chompstation.
image
image
image


JoltDerg:
oh god why did it do that

PR #14505, opened 2023-02-18
Even more sprites including decent hoodies

Burrito-Justice:
Being the dummy that I am, I neglected to port even more sprites that I somehow forgot to include in my other PR. So consider this part 2 of https://github.com/VOREStation/VOREStation/pull/14470

- New undies: The tape has been whitened up a bit for more color variety, plus there's some swimwear too.
- Even more cloaks: I guess Virgo just needs more cloaks. These weren't included in the first PR because they were GAGS content on another server and therefore hard to work with, but with the power of automation, we now have boat cloaks and shrouds.
- Shoes: The antediluvian wraps have been renamed to something sane, and the flats from Aurora have been ported, available as an alternate option in the loadout.
- Speaking of Aurora: the metal halo and a mouth veil have been ported. Many colorable skirts in a selection are here as well.
- As far as hats are concerned, there's now a conical non la and a bucket hat.
- I'm throwing even more hoodies onto the existing pile. This time, however, their hoods and zippers function independent of each other, so you don't need to have a hoodless hoodie to button your jacket, or suffer having to zip up after putting your hood up.
- Lastly, at a user's request, I've added some of the skimpier swimwear in our code to the loadout, so now they can actually be selected.
image
_New swimwear undies, featuring the new flats_
image
_Alt halo and mouth veil_
image
_New hoodie, featuring zipper and hood independence (pictures slightly outdated, the hood is fixed)_
image
_The new colorable hoodies_
image
_One of the new colorable skirts_
image
_New department boat cloaks. There's a white version you can color as well_
image
_Department shrouds. Admittedly hit or miss_
image
_Non la and bucket hat_


Burrito-Justice:
image
This is the part where I say it compiles fine on my machine and ask you to merge anyways xddd

PR #14530, opened 2023-02-21
prevents persistent dirt (and other filth) from spawning in spess and walls

Seris02:
a) build turf in spess/extend an old one
b) dirt
c) new round has spess dirt


Heroman3003:
add isopenspace() too, or however that check goes. just in case.

Seris02:
> add isopenspace() too, or however that check goes. just in case.

ohhh, good idea

PR #14531, opened 2023-02-21
Adds a new POI for debris field with a new ghostpod type

Runa-Dacino:
The POI (debris field, no duplicated, cost 35):

image

How it looks in game (when RNG decides to load it. So a bit outdated! Removed the Nurse spider near the ghost pods)

image

The Ghost Pods:

Away team can enter the ghost pod area and activate them. This notifies ghosts currently observing of the role. Ghosts cannot activate it themselves until someone touched it.

Spawns a carbon mob of any whitelisted species (meaning: any species the player is whitelisted for. They CANNOT spawn as say, shadekin, if they don't actually have the whitelist). Can pick name on spawn. You are warned NOT to use your station chars in the ghost query.

You receive 3-60 brute damage on spawning.

Player is warned this is not an antag role, and to follow rules. Encourages them to help make the world come alive.

Jobban allows for both ghost role (general) and survivor for bad behaviour.

How it looks IG (screenshot at centcom because area didn't spawn when I got the ghostpod to work):

image

Ghost pod should be easily generalized using the make_antag = MODE_SURVIVOR flag for other POIs and even overmap locations!


Runa-Dacino:
Redesign:
image

Runa-Dacino:
Seeing as the cryo room needs air to not instakill the rescues, added a firelock that starts closed. This should ensure that when players inevitably vent the whole place, they get a warning that "hey, maybe what's inside there is important and we shouldn't open without inflatables."

This is reinforced by a recording loot on the captain's coprse which when printed for transcript should inform that survivors escaped inside.

Heroman3003:
It being able to spawn whitelisted species is something I find very problematic, at the minimum. Werebeasts/Shadekin. I havent looked further, although I also am very ambivalent towards idea of the survivor-in-space role in general.

Runa-Dacino:
> It being able to spawn whitelisted species is something I find very problematic, at the minimum. Werebeasts/Shadekin. I havent looked further, although I also am very ambivalent towards idea of the survivor-in-space role in general.

I have messed up phrasing - what I meant to say that the code automatically checks for whitelisted species (this is done automatically on ghost_pod/human/create_occupant(). 

I tested it as a BYOND guest and only had access to non-whitelisted species. Whereas as host, I could pick every species possible to spawn - therefore, it works! Fixing phrasing in OP.


Also did a small commit where I removed a redundant jobban check - virgo already implemented a check for this way up in /ghost_pod/manual so I don't need to check again.

I also deleted the comments left in the tape recorder code since it was not needed.

KillianKirilenko:
> Seeing as the cryo room needs air to not instakill the rescues, added a firelock that starts closed. This should ensure that when players inevitably vent the whole place, they get a warning that "hey, maybe what's inside there is important and we shouldn't open without inflatables."
> 
> This is reinforced by a recording loot on the captain's coprse which when printed for transcript should inform that survivors escaped inside.

If I were you I'd go the extra mile and add a couple more area defs so you can add some more air alarms. If the door to the cryo cells is seethrough and the air alarms are placed properly, then anyone looking in will see the air alarm on the inside says "OK, Breathable!" and the alarm where they are says "NOT OK!" and then... with any luck they should be able to do the math themselves, yeah? Or at least it ought to drive it home that they need to be careful.

Before you ask: yes, there's some off-grid air alarm subtypes you can use so engineering doesn't get weird air alarm warnings for non-station places, they're used on the Talon and in a few other places.

I'd also suggest going through and making the rear airlock actually function. The airlock helpers make this really easy to do as well! I forget the exact item paths but it doesn't take much work to set up nowadays.

Runa-Dacino:
> > Seeing as the cryo room needs air to not instakill the rescues, added a firelock that starts closed. This should ensure that when players inevitably vent the whole place, they get a warning that "hey, maybe what's inside there is important and we shouldn't open without inflatables."
> > This is reinforced by a recording loot on the captain's coprse which when printed for transcript should inform that survivors escaped inside.
> 
> If I were you I'd go the extra mile and add a couple more area defs so you can add some more air alarms. If the door to the cryo cells is seethrough and the air alarms are placed properly, then anyone looking in will see the air alarm on the inside says "OK, Breathable!" and the alarm where they are says "NOT OK!" and then... with any luck they should be able to do the math themselves, yeah? Or at least it ought to help drive it home that they need to be careful.
> 
> Before you ask: yes, there's some off-grid air alarm subtypes you can use so engineering doesn't get weird air alarm warnings for non-station places, they're used on the Talon and in a few other places.
> 
> I'd also suggest going through and making the rear airlock actually function. The airlock helpers make this really easy to do as well! I forget the exact item paths but it doesn't take much work to set up nowadays.

Good idea on the air alarm thingy. I'll give it a try.

Airlocks do scare me (I need to actually set up atmos rather than make it look like it was set up. smh.)

But might as well try! (altho, will need to make the airlock area bigger then to allow a squad of 5 to enter comfortably).

KillianKirilenko:
> Good idea on the air alarm thingy. I'll give it a try.

I won't pretend it'll be surefire, but hopefully *someone* ought to notice the disparity and point it out before anyone does anything careless.

> Airlocks do scare me (I need to actually set up atmos rather than make it look like it was set up. smh.)
> 
> But might as well try! (altho, will need to make the airlock area bigger then to allow a squad of 5 to enter comfortably).

Like I said, the helpers make it super duper easy to do - take a look at how it's set up on the Talon's port wingtip lock;
image
This is just a quick preview and I suggest looking at the full map to get a better idea of the variables and actual entities involved, but it gets the basics across!

You just set up all those little helper doodads on the same tiles as the doors/pumps/sensors and give the airlock controller a unique ident, and then the helpers handle the rest when the map is loaded.

Runa-Dacino:
Based on its potential use for GMs to create Medical Content (TM), I decided to tear out the ghost pod code from being reliant on debris field location being loaded, and instead gave it its own file.

I decided to make it no longer a child of ghost_pod/human so that I can remove the antaggy parts of create_occupant() that I couldn't find any way to shut up without adding a bunch of new variables already.

Same goes for the injury code. Making it reliant on ghost_pod/human would've required setting start_injured to FALSE and then do a bunch of jury rigging around it.

It feels cleaner this way and won't cause issues downstream by going in to edit /human itself to avoid jank, since it IS a thing action servers want to use as it were.

The Debris field POI STILL relies on ghost_pods heavily as its gimmick.

If I understand git right, there shouldn't be any merge conflicts between the two since the commit exists on both branches and all that.

The ghost-pod PR: https://github.com/VOREStation/VOREStation/pull/14547

Runa-Dacino:
Current iteration (in-game):

https://media.discordapp.net/attachments/743674561576566857/1078736634163503205/image.png?width=778&height=596

in the editor: 

image
(I replaced the air canisters with air tanks at reduced start pressure because the canisters took too long to refill)

I also added a few permanent lights to the outside space area because the hull was too dark without them.

I added an air alarm separate for the cryo room to reinforce "you shouldn't breach this area".

Tried to add functioning powernet like I did for atmos but unfortunately I've no idea how to get the SMES to work so it's running off of needing no power.

KillianKirilenko:
Okay, so you've got air alarms, but the problem now is that nobody can see the cryochamber's alarm from outside, which is the important part of the equation. The door into that room needs to be the kind with windows (as does the emergency 'lock) *or* there needs to be a window in from the big room outside, where the air alarm on the wall there can be visually compared to the air alarm in the cryochamber.

Also make sure the pipes leading to those vents are the same type all along their length or else they'll not hook up properly (if you look closely, it looks like the pipes on the vents themselves are broken; that means something went wrong). Better yet, rig up a system where the scrubbers dump into a hazard canister; the amount scrubbed in a round will be low enough that a single canister should suffice. The vents aren't attached to the hull or supported in any way, which is going to be a problem when it comes to making the thing OM-capable.

Powernet stuff is pretty simple but it's hard to tell without any wiring on the display.

As for making it OM-flight-capable, I'd say take a look at any of the existing ones. Reverse-engineering that stuff is one of the best ways to figure out how stuff works (in my opinion+experience) since you can follow everything and see how it all fits together. If in doubt, ask in one of the dev channels!

Runa-Dacino:
> Okay, so you've got air alarms, but the problem now is that nobody can see the cryochamber's alarm from outside, which is the important part of the equation. The door into that room needs to be the kind with windows (as does the emergency 'lock) _or_ there needs to be a window in from the big room outside, where the air alarm on the wall there can be visually compared to the air alarm in the cryochamber.
> 
> Also make sure the pipes leading to those vents are the same type all along their length or else they'll not hook up properly (if you look closely, it looks like the pipes on the vents themselves are broken; that means something went wrong). Better yet, rig up a system where the scrubbers dump into a hazard canister; the amount scrubbed in a round will be low enough that a single canister should suffice. The vents aren't attached to the hull or supported in any way, which is going to be a problem when it comes to making the thing OM-capable.
> 
> Powernet stuff is pretty simple but it's hard to tell without any wiring on the display.
> 
> As for making it OM-flight-capable, I'd say take a look at any of the existing ones. Reverse-engineering that stuff is one of the best ways to figure out how stuff works (in my opinion+experience) since you can follow everything and see how it all fits together. If in doubt, ask in one of the dev channels!

Vents are intended kind of to reinforce the vibe of "this is a luxury yacht that's wasteful" (with all the glass and lack of proper safety).

Would placing a lattice/airless plating under them resolve the OM convertibility issue?

Will fix the pipes tho! Forgot to update that part I guess.


Power wise, my issue was that it seemed the cable did not connect with the SMES and the SMES did not spawn a terminal either.

KillianKirilenko:
> Would placing a lattice/airless plating under them resolve the OM convertibility issue?

I would err on the side of caution and use airless plating just to be safe if you're trying to keep physical structures like vents or sensors attached.

> Power wise, my issue was that it seemed the cable did not connect with the SMES and the SMES did not spawn a terminal either.

Terminals must be manually placed for SMES, as they can be in any cardinal direction but the SMES itself has no rotation states like an APC (which is how APC terminals get autoplaced). As a reminder, power goes *into* the SMES via the terminal, and *out* via the cable knot immediately underneath, *e.x.* the Von Braun's engineering core;

image

What's probably happening is just that the SMES is confused because it has no input terminal, but you may need to doublecheck how the areas are set up for your map, if they have `requires_power = 0` or anything like that.

Runa-Dacino:
> > Would placing a lattice/airless plating under them resolve the OM convertibility issue?
> 
> I would err on the side of caution and use airless plating just to be safe if you're trying to keep physical structures like vents or sensors attached.
> 
> > Power wise, my issue was that it seemed the cable did not connect with the SMES and the SMES did not spawn a terminal either.
> 
> Terminals must be manually placed for SMES, as they can be in any cardinal direction but the SMES itself has no rotation states like an APC (which is how APC terminals get autoplaced). As a reminder, power goes _into_ the SMES via the terminal, and _out_ via the cable knot immediately underneath, _e.x._ the Von Braun's engineering core;
> 
> image
> 
> What's probably happening is just that the SMES is confused because it has no input terminal, but you may need to doublecheck how the areas are set up for your map, if they have `requires_power = 0` or anything like that.

Got it to work! I was using the wrong SMES, and lack of terminal were the cause after all.

Gonna try to overmapify. It doesn't seem /that/complicated now that I got this far.

Runa-Dacino:
It works!

image


It can land near SD, land on V4.

It was a liiitle too fast with initial 10t weight, doubled to 20t. 

4 out of 6 phoron tanks start empty
fuel port also starts empty

PR #14534, opened 2023-02-22
Adjusts artistic crate orders

Heroman3003:
Adds a full set of art equipment to arts and crafts set

Moves paint buckets from arts and crafts set to station painting set

Reworks how paint buckets spawn their contents to be more reasonable

Adds a few more types of paint buckets  and spawns more paint remover in the crate


lm40:
Yea fine

BigBababooey:
To quote a famous person

Yea fine

PR #14536, opened 2023-02-23
Makes speech bubbles manually selectable, adds a whole bunch of new ones

Heroman3003:
Idk if this will be decided as 'generally desirable' but I felt like doing this so why not

New verb in OOC tab for that. Using 'default' option forces it to select a bubble based on current standards.

Will probably put it in the char setup in separate PR if this one is deemed okay.

image


lm40:
Ok

BigBababooey:
This is great

PR #14537, opened 2023-02-23
dotted orange mushroom cap sprite in ear slot, non-colorable

LiquidFirefly:
image


lm40:
Sure why not

BigBababooey:
I could see some fun characters utilizing this

PR #14546, opened 2023-02-24
Custom Species name in Records

KillianKirilenko:
Changes addition of species to medical and security records to use custom species name if specified. Original species will be displayed in brackets afterwards, *e.g.* "Dragon (Unathi)" or "Chimera (Custom Species)".

Oddly the only places I've seen this actually be relevant are the PDA-based Sec & Med records.


KillianKirilenko:
*Right*. This was a real headscratcher for a while, but now we've also got:
- Species in *static* Medical, Security, and Employment Records Consoles (not just PDA and pAI data)
- Species in Medical, Security, and Employment Records when Examined with AR glasses (tie in to #14069!)
- Species on ID cards (required a tgui bundle recompile)
- Ability to update Species data on relevant terminals (but not on ID cards via ID-editing things, yet)

That took me entirely too long to figure out, but it's done.

PR #14548, opened 2023-02-24
Character Setup Lore Option Tweaks

KillianKirilenko:
I started going through the character setup stuff to add missing corps and entities to the list, and something struck me. *Why are planets listed seperately from systems, and why would citizenship be seperate from faction?* Why would someone have specifically Martian citizenship if their home system is Alpha Centauri? It seemed weird, and more likely to just confuse people.

With these considerations in mind, I've made the following changes;
 - Citizenship is now the *Faction or Political Entity* with which you hold, well, citizenship. This is reserved for either major power players (the Commonwealth, Fyrds, ArCon, etc.) or distinct interstellar entities (KHI, Nomads).
 - Home System is now just Home, and includes planet/system (e.g. "Mars, Sol"). The prompt is now "place of birth or primary residence". The list has been extended (fairly significantly) with a number of new locations.
 - Faction is as it was before, faction/corporation. Government-level entities have been shoved up to Citizenship, and a number of missing factions have been added. A few missing entries have been added.

Other/None options remain available so you can fill in your own or leave them blank as desired.

Marked as draft since I'm sure some elements need adjusting and things may need adding or removing based on lore developments and such.


Very-Soft:
Should add Ironcrest Transport Group~ ITG control is where a lot of the black eyes live and operate out of.

BigBababooey:
I like this overall, I don't see many glaring issues, definitely potential ones to be added like Soft noted.

Only note I have is that it's the Moghes Hegemony not the Unathi Hegemony, it's a holy planet and the seat of political power... yadda yadda.

Runa-Dacino:
I'd rename Nomads as:

Altevian Hegemony
&&
Kosaky Fleets

KillianKirilenko:
> Should add Ironcrest Transport Group~ ITG control is where a lot of the black eyes live and operate out of.

Good point! I probably need to shake down the lore channel and do another traffic update with new factions, honestly.

> I like this overall, I don't see many glaring issues, definitely potential ones to be added like Soft noted.
> 
> Only note I have is that it's the Moghes Hegemony not the Unathi Hegemony, it's a holy planet and the seat of political power... yadda yadda.

Gotcha. Wiki and a few other things probably need a little updating then.

> I'd rename Nomads as:
> 
> Kosaky Fleets

Noted.

> Furthermore, perhaps a separate entry for Origin/Birthplace and Residence might be swell.
> 
> For instance, my main PC would be Origin: W1061c, Wolf1061 Residence: V3b.

I didn't see too much point in seperating birthplace and current residence, but if enough people feel it's important it shouldn't be too hard to set up. It just touches on a few more things than repurposing the existing systems, is all.

KillianKirilenko:
Ok, I went and made birthplace a seperate entry anyhow. I realized I could cheat and have it draw from the same list as home systems without needing to duplicate those, so that saves me a bit of extra effort.

I *also* made it so your known languages appear as part of your employment record, along with all this other data, and all of the above can also be edited via employment record consoles. Language seems like it'd be something useful to have on records, and I meant to add it sooner/along with the last set of records tweaks.

Heroman3003:
If there are any lore-related inconsistencies, can be cleaned up later. Otherwise code seem good.

PR #14579, opened 2023-02-27
Ships!

KillianKirilenko:
Well, two ships to be precise. An SDF Corvette (a design from downstream I did a little overhaul on) and an all-new SDF Cutter. The Cutter is essentially a flying ORB with some minor amenities, whilst the SDF Corvette is a small shuttle/troop transport outfitted with a bunch of PD lasers.

A mildly damaged version of the Corvette can now occasionally be found in the debris field, along with its former occupants. There is no damaged Cutter. Adminspawn/event use only, for now!

Also fixes a silly issue with the intact cargo gecko that would have caused conflicts and docking problems if using it and the intact passenger gecko at the same time, and moves the abductor and lunaship dm loading into common_offmaps.dm with all the others for consistency.

~~Drafted until I remember to go in and add light fixtures.~~

And yes this is me warming up/getting back into the swing of mapping.


Heroman3003:
Sure, variety nice

PR #14585, opened 2023-03-01
Fixes window icons on luxury yacht, nerfs royal spider eggs

Runa-Dacino:
Windows were apparently having a glitchy icon with my attempt to make them transparent. Tried a new attempt, this works.

Also, I was informed that you can use royal spider eggs in your hands - which, given I had 3 of them mapped in... would've caused immense chaos.

Made it so that the POI has a unique subtype of royal spider eggs that has a var-based probability of being crushed when used in your hand [(100-egg_health) probability, so 90% chance on default]. This means there is 3 rolls to spawn a spider princess at 10% chance, which should make it a unique reward to play with BUT avoid the utter chaos it'd cause if it was guaranteed.


Runa-Dacino:
Changes were made per discord!

PR #14594, opened 2023-03-04
Disables Wild West and Arynthi Lake

Very-Soft:
Per prolonged user request


Wulfestwulf:
I, for one, vote no on this change.

Skysteak:
This PR was made either during or in the aftermath of an unpleasant discussion wherein things got quite heated. I wont go into it here but the context of that is important for reasons you can imagine. 

In any case unless there is a compelling reason(s) or the individual in question made the maps themselves, but wants them removed, I naturally cannot support this, and I trust that this is just a 'PR of passion' so to speak.

PR #14598, opened 2023-03-04
[MIRROR] Upkeep on Mech & Cliff code.

VirgoBot:
Original PR: https://github.com/PolarisSS13/Polaris/pull/8946
--------------------
Assembly Mines fixed, they do not spawn with an explosive payload from parent.

Vehicles no longer ignore cliffs.

Exosuits no longer ignore cliffs.

Objects can fall off of cliffs. Objects with a buckled person will hurt the person. (Rollerbeds looking at you.)

Jumpjets added to allow planetary traversal, primarily useful upon the Serenity, Hoverpods, and Marauders (adminspawn). When toggled, they allow movement vertically, and prevent falling through open spaces. When used as the active equipment, it will launch the exosuit toward the target turf, over cliffs, trees, and regular terrain. When not on one of the above mentioned suits, it will cause a small explosion on launch, damaging the exosuit and anything directly nearby.

`throw_at` can now take `arc` as an input, representing maximum pixel height in an arc when the object is moving, giving a guise of depth for a throw animation. Currently, only used by jumpjets and exosuit grenade launchers.


Heroman3003:
This is aparently polaris cliffs which are kinda barely used so like, yeah, nvm, not as important as i thought.

PR #14616, opened 2023-03-06
Creates framework for renaming visitable overmap objs on initialization

Runa-Dacino:
**What it does**
_Currently? Nothing!_

However, it makes it possible to define subtypes of its new landmark that when placed in a POI, will add that POI's unique descriptors list("name","desc","scanner_desc") to a list within the overmap object corresponding to the visitable late-loaded Z level.

The system then checks this visitable overmap object if defined as such within the subsystem, picks a random list() from the list of lists - checks if it won't cause a runtime (checks for this on initializing this landmark as well, just to be safe!) - if it would, it picks a new one and removes the offending element until it finds a working one, or runs out of landmark (in which case it leaves the obj be as it were)

Should a suitable list("name","desc","scanner_desc") be found, it proceeds to rename the overmap object in such a way as to pay attention whether it needs to be scanned first "unknown sector" or not. If it spawns unknown, players must scan it before revealling its new name.

It's also possible to create descriptors independent of the landmarks, for extra variety! To do this, one needs to go where list("default") is and change it to list("default", list("name", "desc", "scanner_desc")) (add a list to the list).

For debugging and also GM's planning, it adds an admin log when the debris field was changed and to what new name. It also records this within the game's initialization messages.

**Why we need this**

Due to Z level constraints, adding a variety of overmap locations is not feasible. While there exists a system for replacing the space whale with a brand new map - the barrier for entry to create such a thing is quite high! I would not feel right to add a new map that is not on par with the gimmicks and opportunities created by a space whale that would deny players a chance to visit it. 

I tried to make it so that multiple randomly chosen maps are loaded from a list (then slap DF, whale, Fuel depot in there) with a "pick 3 from this list" procedure buuut... I learned about Z the tyranny of Z level defines and decided I'm not going to touch that system.

Thus, barring creation of a mapping lateload proc that can pick 3 random maps from a list bigger than 3 maps - this framework affords us the next best thing: Illusion of having multiple overmap locations to visit... without increasing the Z level!

For instance, a player wants to create a fun away mission involving an abandoned space station with a unique story. Before this framework, if they added it to the Debris field, people would have went "we're going to the same debris field again" and remain ignorant of this cool new content by refusing to go there. This way, if the mapper adds the landmark as outlined in the readme (tried to describe it in such a way as to encourage its use by even junior mappers!) - there's a chance that when the SD flies by the unknown sector and scans it - they'll see "Decrepit Space station" with a cool little story blurb for the scanner's description and create excitement for "ooh! I havn't been there before! Let's check it out!"

I also admit I got somewhat irked by people being downers about Debris field and Space Whale and ruining the moment for others by going "I've been here before, stop pretending this is new and exciting!"

But primarily, I want to draw attention to this framework's ability to create an illusion of additional locations to visit and to showcase cool new, more complex POIs!

**Post Scriptum**

My first idea that would use this new framework would be an enormous wreckage that's actually just a tiny part of something implied to be ridiculously big, filled with anomalies/precursor tech! Second idea is a large abandoned mining facility - both for the DF.

Maybe, also adding a landmark to the currently existing Alien Derelict

**Testing**

Tested against landmark with nontext elements
Tested against landmark with 2 elements
Tested against landmark with a null element
Tested against landmark with 4 elements
Tested against landmark with an empty list
Tested against no landmarks, and no list("default") defined (this is why I have the possible_descriptors == list() check. I couldn't find any other way to catch that from making a runtime error than to do that).


Runa-Dacino:
Seeing as my idea of 

"Let's create a global list that we compare strings against to see if Z level loaded to avoid runtimes" failed horribly... I realized I might as well scrap that global var and do the 
for(visitable/V in objlist) -> if(V.unique_identifier = "debris field"
as the only thing that checks if the Z level is loaded. Works like a charm, and 1 less global list and 1 less iterating over a big list - minor performance save, yay!

Inspired by this, I looked at landmarks and saw that each initialization goes on to check a very big list containing shuttles, sectors and stuff... What if we saved the reference from this first check somewhere and just grabbed it for later initializations for the given subtype? Well - this works too, and seems to run more smoothly!

Unfortunately, I could not figure out how to declare a static var in such a way that is unique to each subtype without having to delcare it in each subtype repeatedly... so - when making this system work for a new map, you'll need to declare var/reference for your landmark/renamer/subtype landmark. But! We're now checking a pretty big list only once, which should avoid issues once we might get way more landmarks.

code/__defines/subsystems.dm
```diff
@@ -87,6 +87,7 @@ var/global/list/runlevel_flags = list(RUNLEVEL_LOBBY, RUNLEVEL_SETUP, RUNLEVEL_G
 #define INIT_ORDER_PERSISTENCE	-25
 #define INIT_ORDER_SKYBOX		-30 //Visual only, irrelevant to gameplay, but needs to be late enough to have overmap populated fully
 #define INIT_ORDER_TICKER		-50
+#define INIT_ORDER_MAPRENAME	-51 //Initiating after Ticker to ensure everything is loaded and everything we rely on us working
```
Heroman3003, line 90: I'd suggest setting it to something like -60 so its not as smooth a margin.

code/controllers/subsystems/overmap_renamer_vr.dm
```diff
@@ -0,0 +1,32 @@
+/*
+Readme at code\modules\awaymissions\overmap_renamer\readme.md
+*/
+
+SUBSYSTEM_DEF(overmap_renamer)
+	name = "Rename Overmap Locations"
```
Heroman3003, line 6: Overmap Renamer to match the path name

Heroman3003 (changes requested):

Runa-Dacino:
Apparently there was a merge conflict due to LUNA ship and abductor?
I'm a bit confused since I'd thought they were loaded by offmaps...

vorestation.dme
```diff
@@ -4207,5 +4212,6 @@
 #include "maps\submaps\space_submaps\debrisfield\debrisfield.dm"
 #include "maps\submaps\surface_submaps\wilderness\wilderness.dm"
 #include "maps\submaps\surface_submaps\wilderness\wilderness_areas.dm"
+#include "maps\tether\tether.dm"
```
Novacat, line 4215: This looks to be the issue.

vorestation.dme
```diff
@@ -4191,6 +4194,8 @@
 #include "maps\expedition_vr\beach\submaps\mountains.dm"
 #include "maps\expedition_vr\beach\submaps\mountains_areas.dm"
 #include "maps\gateway_archive_vr\blackmarketpackers.dm"
+#include "maps\offmap_vr\om_ships\abductor.dm"
+#include "maps\offmap_vr\om_ships\lunaship.dm"
```
Novacat, line 4198: These should probably not be automatically loaded as well.

Runa-Dacino: Bit confused how they ended up within my #include since I did not touch those files at all.

c'est la vie. Those caused the merge conflict earlier as well, which I misunderstood and added them.
Tried removing them now.

PR #14618, opened 2023-03-06
Adds a new mech weapon, the GA-X Experimental Gamma Laser

MisterGrimm:
### The GA-X Experimental Gamma Laser

For when you absolutely must keep a corridor protected. The GA-X "Render" is a illegal experimental weapon for combat mechs that excel at high rates of fire, and general suppression of humanoid targets. While it's burn damage is highly subpar by the standard of other laser weapons, it easily makes up for it with it's high fire-rate and ability to give minor radiation to it's target. 

While very effective against infantry, one will quickly learn that it is a terrible weapon against other mechs and structures, and will often perform extremely poorly against them. It should also be noted that the GA-X is a illegal weapon due to it's radioactive properties, and falls under number §212 of Corporate Regulation. Following of the Research Contraband Policy is highly recommended.

https://user-images.githubusercontent.com/26284899/223243113-02c5c59a-341e-4c2c-b207-4209dae3efdb.mp4

### DIFF
--- a/code/game/mecha/equipment/weapons/energy/laser.dm
+++ b/code/game/mecha/equipment/weapons/energy/laser.dm
@@ -68,3 +68,14 @@
 	equip_type = EQUIP_UTILITY
 
 	origin_tech = list(TECH_COMBAT = 4, TECH_MAGNET = 4, TECH_PHORON = 3, TECH_ILLEGAL = 1)
+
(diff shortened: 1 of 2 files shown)


Heroman3003:
The question of 'what use will this ever get' brought up in discord stands, but the mech equipment is currently not that bloated, and this isn't particularly ban-bait-y thing, or too strongly objectionable, so, eh.

PR #14620, opened 2023-03-08
Start With Backup Implant setting

Very-Soft:
Allows one to spawn with a backup implant. Rather than requiring it to be obtained literally every single time you spawn, you can just spawn with one instead, if you don't want to go to medical!

ICly there's no real reason why you would get your implant removed every shift, and OOCly if for some reason we decide to make resleeving not possible one shift, we're going to remove the resleevers, not the implants, so this should only really affect people who don't like getting implants every shift, and should make things a little less stressful for medical.

This option defaults to 'off'. It must be enabled to spawn your character with an implant.


Skysteak:
Sounds solid! Lovely to have the option!

PR #14625, opened 2023-03-08
Implementing tgchat to replace old VChat

ItsSelis:
Tgchat generally seems to have a bunch of neat features that we currently do not have and it also seems to use a better way to save chat history (client-side instead of server-side for every single player). Of course it is still **possible** to export chat logs as well as creating multiple tabs, only selecting specific things.

But additionally with tgchat, one is able to highlight specific words in the chat itself (like the name of a person you are talking with while being in a very populated area), some music/sound player that admins or game masters could use to play specific things for events and some other things like font-size, font-family etc.

Since this is also based on tgui, it would synchronize the dark/light-mode theme a little better, especially with the browser verb panel PR (#13816).

This still is work in progress and the TODO points are down below.
 
Major Changes:
- Replaces VChat with tgchat/tg-panel

Minor Changes:
- Moved some of the assets into separate DM files 
- Added more version compatibility defines for theoretical future use
- Added some more commentary in client_defines

Notes:
- The pref of having oldchat automatically be started is still there. You also dont have to toggle it since it uses the old pref that was there already. But you might see tgchat for a split second before disappearing.

<details>

<summary>Screenshots</summary>

image
image
image
image

</details>

This PR has been atomized into following PRs:
1. #15530
2. #15553

---

TODO:
- [x] Resolve the TODO in [code/_helpers/text.dm](https://github.com/VOREStation/VOREStation/compare/master...ItsSelis:selis-tgchat?expand=1#diff-f668f6ce490bc5ac874e15e9fb9abed0d13f783acc94126c9d97242357b2e421)
- [x] Properly test this (not only using a Windows VM)
- [X] Assets seem to still runtime a bit
- [x] Fixing autoscrolling
- [x] Better history persistency
- [x] Testing all the filters
- [x] Fixing OOC/LOOC/MSAY/ASAY etc. text_tags infront of messages (currently b i g)
- [x] Finalizing the colors
- [x] Automatic change to oldchat via preference

---

<details>
<summary>NOTE: Current Colors:</summary>

image
image
image
image

</details>

---

:cl:Upstream
add: added tgchat
del: removed vchat
fix: most likely fixed some scrolling issues with some of the UIs
/:cl:


Detective-Google:
tgchat can play bloons 10/10 addition

JoanRisu:
Since Aro worked on VChat, I'll have him take a look at this due to his familiarity with the chat systems. (Or was that lesh? I dunno.)

ItsSelis:
> Since Aro worked on VChat, I'll have him take a look at this due to his familiarity with the chat systems. (Or was that lesh? I dunno.)

Sure, though, it's not having any high priority or anything since there are still the open issues on the checklist above (especially about the history of the chatlogs themselves, whjich are a pretty important part, and if course even worse: the currently inability to have autoscrolling working). 

I recently researched a bit more into squeezing a bit more performance out of Vue if possible and found a little library that at least claims to improve the display of big data amounts. That is just very theoretical though so far.

User-U-U:
Does this still have the option to opt out and use regular chat like Vchat does.

ItsSelis:
> Does this still have the option to opt out and use regular chat like Vchat does.

There will be a verb called "Fix chat" that tries to fix the chat and ask the user if it worked or if they rather want to switch to the old one. Though I think the pref would have to be re-integrated (as this was just a pretty rough port).

Heroman3003:
Yolo, will revert if it causes some fatal breakdown in two days, but for now lets just try it

PR #14650, opened 2023-03-09
A bit more hardening of SQL Injection (of stuff using old sql_sanitize_text proc)

ItsSelis:
Note: I do not currently have a TGS instance with a linked discord bot in a testing server myself to test out the communication and registering of that stuff. The departmental hours (at least for fetching the time) system uses the correct proc already, just the inserting of new time is wrong. Though this would have to be observed and tested if there are any issues.

Generally, the logs should be also checked if there are any SQL related errors around (if sufficient logging is actually in place).

Using own function to sanitize SQL is not a good idea.
Or at the very least the sql_sanitize_text is not enough to properly sanitize incoming SQL stuff.

Updates (Queries)/Variables:

- Departmental Hours (Inserting new time)
- SQLite Subsystem
- Banning
- Registering new account (Linking with Discord)
- Some TGS Communication

### DIFF
--- a/code/controllers/subsystems/persist_vr.dm
+++ b/code/controllers/subsystems/persist_vr.dm
@@ -81,8 +81,8 @@ SUBSYSTEM_DEF(persist)
 		dept_hours[department_earning] = min(config.pto_cap, dept_hours[department_earning])
 
 		// Okay we figured it out, lets update database!
-		var/sql_ckey = sql_sanitize_text(C.ckey)
-		var/sql_dpt = sql_sanitize_text(department_earning)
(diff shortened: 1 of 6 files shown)


Heroman3003:
>  ## UNIT_TEST: !! FAILURE !! [SQLITE FEEDBACK: Insert and Retrieve Data]: Data insert and loading failed to have matching information.
> ## UNIT_TEST: User 'Alice' did not receive a cooldown, when they were supposed to.
> ## UNIT_TEST: !! FAILURE !! [SQLITE FEEDBACK: Cooldown]: 1 issues were found.

PR #14651, opened 2023-03-09
better glowing eyes

Seris02:
so they stop showing through glasses n shit, thanks

### DIFF
--- a/code/modules/mob/living/carbon/human/update_icons.dm
+++ b/code/modules/mob/living/carbon/human/update_icons.dm
@@ -575,7 +575,7 @@ var/global/list/damage_icon_parts = list() //see UpdateDamageIcon()
 	eyes_image.plane = PLANE_LIGHTING_ABOVE
 	eyes_image.appearance_flags = appearance_flags
 	*/
-	var/mutable_appearance/eyes_image = emissive_appearance(head_organ.eye_icon_location, head_organ.eye_icon, BODY_LAYER+EYES_LAYER, appearance_flags=appearance_flags) //CHOMPEdit - better glowy eyes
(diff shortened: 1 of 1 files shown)


Very-Soft:
did you actually test this and see to it that it actually works? I went to test it and it doesn't appear to work.
image

Seris02:
> did you actually test this and see to it that it actually works? I went to test it and it doesn't appear to work. image

I did, yeah, lemme see if I can see what's going on

KillianKirilenko:
> so they stop showing through glasses n shit, thanks

What if we *want* them to show through glasses, eh? At the very least it'd be nice if they could still overlay transparent glasses. Sunglasses ok, probably not much sense in glowing behind those, but someone might want that too.

Runa-Dacino:
> > so they stop showing through glasses n shit, thanks
> 
> What if we _want_ them to show through glasses, eh? At the very least it'd be nice if they could still overlay transparent glasses. Sunglasses ok, probably not much sense in glowing behind those, but someone might want that too.

I echo this.

It should be an option either for glowing eyes 
OR
Be sth that checks if the glasses are sunglasses or normal ones.

Not sure how you'd make sure AR glasses accomodate both.

KillianKirilenko:
> I echo this.
> 
> It should be an option either for glowing eyes OR Be sth that checks if the glasses are sunglasses or normal ones.
> 
> Not sure how you'd make sure AR glasses accomodate both.

I think the simpliest way would be to make it a cycle switch like ghost-darkness. `Off->On->Overlay Glasses->Overlay All Eyewear->Off->...`

Seris02:
not ready for merging by the way, haven't gotten around to testing what the issue is that very soft put above

PR #14661, opened 2023-03-11
Changelog Reroute

KillianKirilenko:
In lieu of formalizing the changelog update process so that it actually happens, I've rerouted the "Show Game Updates" button on the front-end menu and the "Changelog" verb to point to the wiki's changelog.

In an ideal world devs would use the proper tagging and changelog addition system to PRs to auto-write to the changelog.html that's stored in the game data. But for now I'd rather have links to the volunteer-maintained changelogs. If staff change their minds and decide to start enforcing the template & tag system, this can always be rolled back.

As an added bonus the current map name will now also be displayed at the top of the front-end menu, so you should never be surprised about which map you load into.

### DIFF
--- a/code/modules/mob/mob.dm
+++ b/code/modules/mob/mob.dm
@@ -433,11 +433,14 @@
 /client/verb/changes()
 	set name = "Changelog"
 	set category = "OOC"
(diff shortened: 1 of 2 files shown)


Heroman3003:
Seems cool.

PR #14662, opened 2023-03-12
Weapon Info on Examine

KillianKirilenko:
As requested, this PR adds rough details of an object's relative power to its description when examined, similar to the rough details provided when examining armour;
image
image
As demonstrated above, it also explains if the weapon can cleave or has reach. This covers everything from `obj/item` down, so you can learn that, hey, the maglight is a mildly effective melee weapon!

The specific values are obfuscated similar to armour values, as giving players the exact damage/etc. numbers isn't really useful without other context. Most players probably know that your average carbon has 100hp before going into softcrit, and 'dies' at -100hp, but simplemob health and armour can be all over the place.

~~I intend to expand this to cover the power of loaded/charged firearms in the future, but that will be a slightly more complicated project.~~ I also intended detail baton stunning power but the procs aren't cooperating.

As a side-effect this PR also makes some changes to e-swords and similar energy weapons, toggling their armour penetration along with their active state. A deactivated e-sword can no longer punch through armour to deal minor cuts (though a deactivated energy-axe is still quite powerful).

~~Some of the numbers/descriptors are not quite final and need adjusting.~~

### DIFF
--- a/code/game/objects/items.dm
+++ b/code/game/objects/items.dm
@@ -25,6 +25,7 @@
 	var/list/origin_tech = null	//Used by R&D to determine what research bonuses it grants.
 	var/list/attack_verb //Used in attackby() to say how something was attacked "[x] has been [z.attack_verb] by [y] with [z]"
 	var/force = 0
+	var/can_cleave = FALSE // If true, a 'cleaving' attack will occur.
 
 	var/heat_protection = 0 //flags which determine which body parts are protected from heat. Use the HEAD, UPPER_TORSO, LOWER_TORSO, etc. flags. See setup.dm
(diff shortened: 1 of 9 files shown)


KillianKirilenko:
Now includes basic coverage for *most* guns, respecting firing mode and loaded state;
image
image
image
(Basic G40 laser rifle set to standard mode, set to suppressive mode, and with depleted/no cell)

There are still a handful of edge-cases I'm trying to iron out like stun revolvers, tool/toy 'guns' such as the medigun, sizerays, and foam-dart guns, but so far it can cover;
- Multi-mode energy guns
- Pump/lever-action ballistics
- Breech/cylinder-loaded ballistics (bartender shotgun, revolvers)
- Magazine-fed ballistics

KillianKirilenko:
Drafted whilst I patch a silly oversight.

KillianKirilenko:
Ok, slapped a bandaid on something that was bothering me. Think I'm done here.

PR #14663, opened 2023-03-12
Fixes item loss when dying as a prommie blob

TheGreatKitsune:
Cause stuff got qdel'd originally for whatever reason.


Heroman3003:
Are the mob_radio and myid ever even set for prommie blobs? Why drop them?

Heroman3003:
Bump

TheGreatKitsune:
Ah, sorry. I need to fix email notifications again...

But to answer, they are not set anywhere here on Virgo that I can see. Downstream on Chomp they get moved over from the carbon's id/radio during transformation.

PR #14664, opened 2023-03-12
Revert "[MIRROR] Replaces Skrell sprites"

peeesh:
Reverts VOREStation/VOREStation#13930

Does LITERALLY nothing, as the sprites were not included by the bot, and the sprites that ARE included are, by default, not used for skrell body sprites, since the code uses 'r_skrell_vr.dmi' instead.


peeesh:
ill probably end up making a separate PR to add more skrell hair from aurora, at some point. since they have similar sort of things, but not with the weird eris/tau-ceti shading

PR #14706, opened 2023-03-17
[WIP-ish] Null and Stolen crate adjustments/changes

Heroman3003:
Pretty much everything in this PR is subject to change, but tl;dr is reducing reliability of getting any specific items from antag contraband orders without removing them entirely. Current iteration comments null crates out and moves most of their unique contents into stolen crates, increases chances of weapons in stolen crates and replaces broken emag there with one-use one (doesnt touch proper full emag).

### DIFF
--- a/code/datums/supplypacks/contraband.dm
+++ b/code/datums/supplypacks/contraband.dm
@@ -53,7 +53,7 @@
  	contraband = 1
  	containertype = /obj/structure/closet/crate/hedberg
  	containername = "Ballistic weapons crate"
-
+/*
 /datum/supply_pack/randomised/misc/telecrate //you get something awesome, a couple of decent things, and a few weak/filler things
 	name = "ERR_NULL_ENTRY" //null crate! also dream maker is hell,
 	num_contained = 1
@@ -105,3 +105,4 @@
 	contraband = 1
 	containertype = /obj/structure/closet/crate/large
 	containername = "Suspicious crate"
+*/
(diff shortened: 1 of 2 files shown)


Novacat:
Null crates have the issue in that thanks to persistent resources + stock market, the 250 cost is no longer a serious obstacle to obtaining them in large quantities and the stuff you get is way too good for that cost.

Skysteak:
> Null crates have the issue in that thanks to persistent resources + stock market, the 250 cost is no longer a serious obstacle to obtaining them in large quantities and the stuff you get is way too good for that cost.

Out of interest, why not just make them away more expensive and call it a day?

TankTheBirb:
> Out of interest, why not just make them away more expensive and call it a day?

That just reduces the quantity at which they can be ordered without resolving the initial problem. Say you make them 1000 points, that just means you can order 4 times less of them. Given that it's a 1/6 chance you get what you want from any given crate you order, it does not take that many crates to actually get what you want. It just means someone plays the stock market a little longer or uses more of the excess (HOPEFULLY the excess) materials made by miners to get this. Eventually, to achieve the desired effect, you'd have to make it so egregiously expensive that it would be effectively removed as content.

Removing the flawed concept and adding some of the more practical and scene items to stolen crates solves the problem while still leaving the desirable things available by certain means.

Heroman3003:
I'll probably get back to this at later juncture.

PR #14718, opened 2023-03-19
Adds a mob to gold extract pool

Heroman3003:


Skysteak:
This is a good change as the added mob is absolutely worth their weight in gold.

PR #14720, opened 2023-03-21
Speed and stuff

Very-Soft:
With the change to movement cooldown, any mob with a positive value in the movement cooldown variable is slow, when previously around 3 was 'normal' speed

3 is now turbo ultra slow, and 0 is 'normal' speed, so I have VERY roughly converted all the movement cooldowns to reflect the new way.

~~This may break REGs since they weren't designed for mobs with negative movement cooldown values but we can address that later if it becomes a problem~~ Actually I did design it with negative values in mind so it will work fine


Heroman3003:
If any mob in specific becomes problematically fast/not fast enough from this they can be dealt with on case-by-case basis, I'd say. Good PR.

PR #14736, opened 2023-03-28
Stops borgs from going unconscious in dropnoms.

foopwo:
No more of this when engaging in a dropnom as a cyborg
_...You almost hear someone talking..._
_...You almost hear someone talking..._
_...You almost hear someone talking..._

In a more perfect world, there would be a function to temporarily disable the actuator, but that's too much to bother for just a vorny feature.

### DIFF
--- a/code/modules/multiz/movement_vr.dm
+++ b/code/modules/multiz/movement_vr.dm
@@ -47,11 +47,13 @@
 		pred.feed_grabbed_to_self_falling_nom(pred,prey)
 		pred.loc = fallloc
 		if(!safe_fall)
-			pred.Weaken(8)
+			if(!isrobot(pred)) //Borgs go totally unconsious when stunned, no fun when they miss reactions.
+				pred.Weaken(8)
 		pred.visible_message("<span class='danger'>\The [pred] falls right onto \the [prey]!</span>")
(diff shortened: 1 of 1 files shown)


Heroman3003:
I will reiterate what I said in discord, that allowing a way for borg to be stunned without full blackscreen thing is a better solution and should probably be done rather than removing sources of borg stun entirely where inconvinient.

foopwo:
This has been a recurring request from borg players for a long time, now. I don't see the utmost importance in preserving a stun effect for a vorny feature.
If you do, that's fine, you can code that.

TankTheBirb:
Having seen this and #14739, I do think the other PR does kinda address the root of the problem rather than one of the symptoms. And that's that borg stuns are aggressively sucky. 

Weaken can happen a lot to carbons by a lot of ways. And to carbons it isn't good, but it isn't catastrophic. For borgs it basically unequips all your tools, blinds you to comms and info, immobilizes you, and basically makes you a casualty. To get back into operational condition, you have to open your inventory, select the tools, then hotkey select them. If they require an activation (like a welder), then it is even more time. 

In comparison, carbons drop something and then pick it up when they get up, or pull another item from another storage and it's usually ready to use or easy to make it ready to use. They also don't lose awareness of everything and can listen to the lamentations of their prey when they land on them.

So basically you highlighted that borg stuns suck (because you miss drop nom reactions), and so this is a good PR that has been eclipsed in scope by #14739.

PR #14737, opened 2023-03-28
Fixes the securiship.

foopwo:
Wrong connector subtype was not allowing gas to get to the engines.


Skysteak:
_Time to get movin'!_

PR #14739, opened 2023-03-28
Makes borgs not stunned when weakened

Heroman3003:
Title says it. Should be a better alternative to #14736. From what I tested it works fine with normal sources of weakening.

### DIFF
--- a/code/modules/mob/living/silicon/robot/life.dm
+++ b/code/modules/mob/living/silicon/robot/life.dm
@@ -87,7 +87,9 @@
 		death()
 
 	if (src.stat != 2) //Alive.
-		if (src.paralysis || src.stunned || src.weakened || !src.has_power) //Stunned etc.
+		if (src.weakened > 0)	// Do not fullstun on weaken
+			AdjustWeakened(-1)
+		if (src.paralysis || src.stunned || !src.has_power) //Stunned etc.
 			src.set_stat(UNCONSCIOUS)
 			if (src.stunned > 0)
 				AdjustStunned(-1)


foopwo:
This has more far-reaching consequences to borg balance.
For the most basic example, a flashed borg can call for help with this change.

Heroman3003:
> This has more far-reaching consequences to borg balance. For the most basic example, a flashed borg can call for help with this change.

In context of our server, that is nigh-irrelevant, I'd say.

TankTheBirb:
While it affects the balance of borgs, it mostly affects PVP balance (non factor for virgo) and negligibly affects PvE balance through QoL (no loss of awareness and no dropping all tools to inventory, but still being immobile. Similar to carbons). Ultimately this is probably the better way to handle the change intended by #14736 while also solving some long standing borg QoL issues. 

I favor this PR more for merging as it solves the intended problem and reverses some questionable balance decisions that don't make sense on virgo. Flashes still stun effectively, and additional sources of stun can be balanced individually. So it should be a net benefit without losing anything other than a source of frustration.

foopwo:
>Plus if something in design SHOULD stun borg completely but doesn't, it can just be changed to paralyze later down the line. Afaik, most current sources of 'weaken' really shouldnt result in complete stunning.

I find it very questionable that it's better to address a specific issue with a deeper change, with little consideration of what it also effects beyond 'we'll fix it later'.

Cyantime:
This is inconsistent with the existing stun variant behaviors. Weaken() is the 'fall on the ground and drop your stuff' one, where it makes sense for borgs to unequip their modules to provide an equivalent. Paralyze() is the 'you are unconscious now' which matches current cyborg stun behavior.

 Stun() is the 'become immobile but still standing' one, which would best represent what's desired here for purely immobilizing a borg. It's also less used as far as I can tell, for borgs specifically at least, so making that the one used for this would have less effect on unrelated things.

Heroman3003:
Ultimately, for borgs, there is no practical difference between most forms of stun due to no horizonaling, and while that could be imitated, I find it unnecesary. However, practically, weaken is one applied in most scenarios which are actually relatively common and affect borg players the most, meaning this is more QoL than anything else. And yes, I mean specifically in reference to 'general' uses of weaken, I haven't particularly looked into what stun types borg-specific interactions do and I don't think it matters that much, when most practically-common stuns come from things such as falls or mob combat, which don't care about whether you're a borg or not, and making borgs suck less in common universal scenario is something that's been requested quite a bit, because, as put in another pr, its just frustration and nothing else for borg to go blackout at any moment of mild inconvinience, and, IMO, tool resetting is just another part of it. In theory, best solution would be to make both stun and weaken not blackout the borg, but keep paralyze and nopower blacking it out. While keeping toolbar reset tied to blackout.

And if I'm allowed a moment of frank honesty, I don't give two shits about the 'pvp' aspect of borg balancing on this server, especially such as "but they can call for help now", which you know what I say to? Good. They should be able to.

> I find it very questionable that it's better to address a specific issue with a deeper change, with little consideration of what it also effects beyond 'we'll fix it later'.

This PR *is* addressing an issue which is a persistent annoyance of borg players with every mild inconvinience resettign toolbar. Again, the fact that the you don't have to reselect your tools after *some* stuns and can call for help/communicate for whatever other intent in said some stuns, I see as improvement, and struggle to see a scenario where it isn't. Grief countering notwithstanding, but that's unnecesary in its own right, considering the many ways one can already easily screw up the borg and the fact that all forms of stun still actually work properly, with same duration, etc, etc, just that this one allows them to still talk and see.

foopwo:
>This is inconsistent with the existing stun variant behaviors. Weaken() is the 'fall on the ground and drop your stuff' one, where it makes sense for borgs to unequip their modules to provide an equivalent. Paralyze() is the 'you are unconscious now' which matches current cyborg stun behavior.

Dropping what's in hands is part of being stunned or knocked down, yes. Historically speaking, borgs were supposed to he immune to most conventional stuns and knockdowns in exchange to their more extreme vulnerabilities to flashes, EMPs, remote lockdown, law subversion, etc. Evidence being the fact that they have never even had visual states for being knocked over, not even given when sprite rotations became possible **way** back, because they simply weren't for getting stunned that way to begin with.

But as is the nature of SS13 codebases, it's hard to tell without trawling through gitblames for hours to figure out if borgs getting stunned by carp and other crap was an intentional change, or a side effect of short-sighted changes affecting more than anticipated.

Tangentially, I still wonder why it's so important to preserve a stun that carbon players can and often take a one-point trait to be immune from, anyway.

PR #14756, opened 2023-04-02
Preferred Language + Autohiss Default Options

Rykka-Stormheart:
Autohiss can now be set to Full/Basic/None from the character menu, and by default, it is assumed Full. It will respect save/load and client connect/disconnect, meaning you no longer have to constantly toggle autohiss.

Preferred Language can now be set. This allows a default spoken language other than Common to be set from spawn, allowing characters to simply join with the language they'd prefer speaking, rather than having to fiddle with "Set Default Language" every time they spawn.
Server rules do still apply, ofc. You must be able to speak Common or at least +understand+ it capably enough to do your job.

See Preferred Language Setting here;

Menu for it here:

See Autohiss Settings here:

Yes, Autohiss is set underneath Egg Type. I wanted to use some of the space there. I can move it up under the Size/etc panels and speech verbs if requested.

Upstream port of https://github.com/CHOMPStation2/CHOMPStation2/pull/5847

### DIFF
--- a/code/modules/client/preference_setup/general/02_language.dm
+++ b/code/modules/client/preference_setup/general/02_language.dm
@@ -1,5 +1,6 @@
 /datum/preferences
 	var/extra_languages = 0
+	var/preferred_language = "common" // VOREStation Edit: Allow selecting a preferred language
 
 /datum/category_item/player_setup_item/general/language
 	name = "Language"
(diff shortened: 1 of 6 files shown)


code/modules/client/preference_setup/general/02_language.dm
```diff
@@ -12,6 +13,10 @@
 	if(islist(pref.alternate_languages))			// Because aparently it may not be?
 		testing("LANGSANI: Loaded from [pref.client]'s character [pref.real_name || "-name not yet loaded-"] savefile: [english_list(pref.alternate_languages || list())]")
 	S["language_prefixes"]	>> pref.language_prefixes
+	//VORE Edit Begin
+	S["species"]			>> pref.species
```
Heroman3003, line 17: Why is species being moved there?

Rykka-Stormheart: Probably a mistaken cherrypick from CHOMP, I can fix that.

Heroman3003 (approved):

PR #14757, opened 2023-04-02
Ear Ringing/Deafened Feedback

Rykka-Stormheart:
Adds deafened ear ring that plays any time you're deafened, and fades back in once you can hear again.

Example of deafened feedback [here](https://streamable.com/8s8vlu). 

Worry not, if you're deafened via mutation or other, you'll simply go deaf, and only hear the fade-in once you regain hearing. 

It will also only play **once** when you are deafened. Further deafening effects while you are already deaf will not force you to hear it again.


TankTheBirb:
Didn't compile for me so I can't test the sound with my personal sound settings but from the video, the sound isn't awful. The ringing is short and pretty quiet. I'd want to test this more when it's able to be compiled.

Rykka-Stormheart:
> Didn't compile for me so I can't test the sound with my personal sound settings but from the video, the sound isn't awful. The ringing is short and pretty quiet. I'd want to test this more when it's able to be compiled.

Yeah, realized there's a missing file. Just re-added the needed bits. Should be working now.

Very-Soft:
I don't think that adding tinnitus to the game is a good idea

Rykka-Stormheart:
> I don't think that adding tinnitus to the game is a good idea

All this is is an audible feedback for if you're deafened, rather than the text-only notification.
I feel it's a useful tool as a notifier that you are presently deaf, and it's pretty cool to hear it when explosions go off. :3

Very-Soft:
> > I don't think that adding tinnitus to the game is a good idea
> 
> All this is is an audible feedback for if you're deafened, rather than the text-only notification. I feel it's a useful tool as a notifier that you are presently deaf, and it's pretty cool to hear it when explosions go off. :3

I think you not hearing anything is a perfectly fine form of audible feedback, personally. This is just annoying.

Heroman3003:
While normally I'm against unnecesary noises being added to every single thing, I will say that I personally think this one is fine. Deafness is rare enough occurence for this to not be a common thing, and the sound is not too loud, and not too long, so you will still experience stillness of silence without the tinnitus noise, but it also does add to the whole 'ow'-feeling part that flashbangs kind alack.

Rykka-Stormheart:
> While normally I'm against unnecesary noises being added to every single thing, I will say that I personally think this one is fine. Deafness is rare enough occurence for this to not be a common thing, and the sound is not too loud, and not too long, so you will still experience stillness of silence without the tinnitus noise, but it also does add to the whole 'ow'-feeling part that flashbangs kind alack.
> 
> EDIT: I realized that it is actually a loop and the short deafening in example made it look like singular short sound. In that case, hard no from me.

The sound doesn't play noise like a traditional loop - you have a silent loop in between.
You only hear intro, once, _ever_, and then outro.

CockatriceXL:
The silent loop doesn't sound very silent. It sounds kind of like an air conditioner. I think it would be better if it was totally silent while you're deaf instead.

Rykka-Stormheart:
I have updated this - ear ringing is now preference-checked and disabled by default.
The middle loop has been replaced with something truly silent.

Heroman3003:
So far the general opinion I see is ranging from 'whatever' to 'nah' so I don't really feel comfortable merging this at the moment.

PR #14758, opened 2023-04-02
Simplemob Injury System

Rykka-Stormheart:
This adds a lightweight simulation of injury for simplemobs. As simplemobs grow more injured, the time between attacks and movement will increase, same as for players.

Formula here: `round(rand(2,6) * damage_fatigue_mult * clamp(((rand(2,5) * (h / getMaxHealth())) - rand(0,2)), 1, 20))`
Threshold for slowdown/etc starts at 60% by default for mobs.
For any coders reading this in the future:tm:, set damage_fatigue_mult to 0 to disable this mechanic for your mobs if you need to for whatever reason.

Upstream port of https://github.com/CHOMPStation2/CHOMPStation2/pull/5840


BigBababooey:
So, off rip I do want to say that I think at it's core this is a nice mechanical idea. And I am happy to have good ideas existing in our code.

However, I would want this to default to all mobs being opted out of the mechanic on that variable. I would like this for a few simple reasons.

- Our codebase isn't presently balanced around the idea that mobs would substantially weaken as damage is taken. And so, this would at present make all of the content (that's already rightfully called less intensive than many of our peers, because that is not our focus as much here) that does exist, become notably easier.
- The endeavor of rebalancing all of our mobs, pois, and the like to account for this would be a hefty undertaking and even if a community member does decide to put in the legwork to do most of that, it does require a lot of oversight from us to fit the vision, and might not fit the vision.
- Keeping it off by default but existing in the code, makes individual rebalancing a lot easier to do. And it allows the design of future mobs with this system in mind. For example, I could see a new Heroman boss mob that still does the "boss phases" it does, but bearing in mind it gets slower, the phases are more intense when they do hit.

Basically I do like this, but I don't think we benefit from the massive shift it would cause if on by default. I could absolutely see myself mass editing certain mobs to have the var flipped on sometimes to make things easier for an away team, but I would prefer it not be an assumed thing for our codebase.

Skysteak:
I'm warm to this PR, and the capabilities it provides are excellent, esp if the inverse gets implemented.

This goes more into future mob balancing more than his PR, but what I will say that is that to an extent, I believe this sort of thing is a 'Either commit (at least to an extent) or don't have it at all,' sort of deal. I fear that if you have it for some, but not others, without any sort of qualifier as what decides who gets it, then you risk player confusion as to why one mob has it why another doesn't etc. It doesn't have to be an ironclad rule, but for example if it applied to 'nature' mobs but not drones, then that'd make intuitive sense.

TankTheBirb:
I think this should probably be enabled by default when it gets merged, mostly because I feel like this is a tool that is going to be overlooked if it isn't, but it may need some tweaks beforehand. From a really short bit of testing (I'll do more later), it shows a lot of promise in delivering something I've wanted for awhile now to make mob fights less binary. 

Feature considerations:
- Is it possible to decouple attack speed and move speed from being the same var? It would be a good aid for balance to be able to set the two independently. I can see wanting to change a ranged mob to be brutally affected by incoming damage, but not affect its movespeed as drastically. Or maybe the opposite if it's a really fast one that I want slowed down after it gets hit once.
- If the decoupling is possible, would it also be possible to add additional variables that can be controlled by injury_level? Things like miss chances.

I can prob say more but im hella tired and need to sleep and also do more testing another time but I very much like this PR.

Skysteak:
Perhaps it could be disabled somehow for played simplemobs? I appreciate that is a tell in of itself, but typically in terms of people playing them, violence is on the more exceptional spectrum of activities, and people usually aren't going 'By jove, the monster mash is less able to graveyard smash, it's a player!' Particularly in the case of maint preds this fits too as alot of them aren't typical critters, which pertains to my earlier point of having _quantifiers_ on what gets them or not.

BigBababooey:
I think the idea that this system would suddenly reintroduce the (bannable) behavior of being a validhunting maint pred hunter that's largely faded from our community (thankfully) is a stretch.

However, since I stand by the belief that this should exist in a default off state anyways, that would not be a concern under my proposed implementation anyways. As for notes of balancing, I didn't really look closely at the numbers, so I appreciate Soft being able to contextualize them further, and it probably is in need of some tweaking for sure. As, it could still have a use case in our codebase for boss mobs created for it, or, mobs designed to have such a mechanic at play.

I also don't think the clarity thing of some mobs having it and others not having it is truly one that would be as big of a deal as one expects though. It would, realistically, just be a scenario where you fight things expecting they wont be, see something exhibiting this mechanic and go "okay, new core memory unlocked, this enemy can be dealt with in this way" and that's it. It's really not that complex, and I don't think we need to insult our player's intelligence.

Skysteak:
> I also don't think the clarity thing of some mobs having it and others not having it is truly one that would be as big of a deal as one expects though. It would, realistically, just be a scenario where you fight things expecting they wont be, see something exhibiting this mechanic and go "okay, new core memory unlocked, this enemy can be dealt with in this way" and that's it. It's really not that complex, and I don't think we need to insult our player's intelligence.

Indeed, you're correct in that it wouldn't cause some massive issue, and people would adapt. To partially reiterate, I just think you'd want to avoid the confusion of it being present for a couple of mobs only, and also avoid it being this forgotten feature, akin to the Mining Drone pacification code (albeit, there is more to that code's disuse tbf). If nothing else, if this does get accepted, then it absolutely should be a consideration for mobs made going forward.

Rykka-Stormheart:
> Perhaps it could be disabled somehow for player controlled simplemobs? I appreciate that is a tell in of itself, but typically in terms of people playing them, violence is on the more exceptional spectrum of activities, and people usually aren't going 'By jove, the monster mash is less able to graveyard smash, it's a player!' Particularly in the case of maint preds this fits too as alot of them aren't typical critters, which pertains to my earlier point of having _quantifiers_ on what gets them or not.

It can be easily disabled for playermobs by editing their `damage_fatigue_mult` var to 0.

Rykka-Stormheart:
> I think this is a sort of poorly conceived idea considering one one of the primary ways in which players encounter simplemobs is when other people are playing as them. If toolboxing someone also makes it harder for them to run away from you, that's gonna suck for basically anyone who plays any simplemob stuff and comes across someone who gets unexpectedly toolboxy.
> 
> Additionally, making the range of move speed slowdown be anything between 1 and 20 is insanity with how move cooldown works now. Based on my calculations, at 50% health at worst based on the randomization, the slowdown can be up to 15. 1 makes you go significantly slower than normal. 2 is a LOT slower, 5 brings you to a CRAWL, at this point it would move so slow that it will no longer be a threat, even if it started out as a pretty fast mob. At 15 the mob would barely be able to move at all, and that's at 50%, and these kinds of penalties start applying at 66%. Most mobs only take a few hits to kill with like, actual weapons. This will make MOST mobs immediately crippled if they were hit with anything more than a feather pillow.
> 
> Further, while I don't think that making some mobs have mechanics like this is a bad idea, blanket applying it to literally all mobs is not a good idea. This feels like the kind of thing that ought to go on tough mobs, boss mobs, or particularly fast and evasive mobs as a gimmick mechanic. As is it looks like it was something made by/for someone who just wants all mob encounters to be easier to overcome at all levels of play, and I'm really not a fan of that. Like, it'd be one thing if you just put this on some bosses and like, mercs. But there's no reason something like this should be on any of the mobs players can play as normally, it will 100% only encourage people to violence. It's not uncommon for someone to go 'woah that's a weird maint bug!' and smack it with the nearest thing not nailed down. It is a good thing for that person to then be able to run away. If getting hit makes you less able to run away, that's only going to encourage those people to try to fight back, which in turn will justify the person who dealt the first blow's actions, and encourage them to do it further. I know that might sound like a little bit of a stretch, but people take the easiest road. If it's easier to run away from a situation people don't want to be in, they'll run, if not, they don't have much choice but to fight. We shouldn't be applying mechanics that encourage people to behave in ways we would prefer they don't.

I'm more than happy to make adjustments to the system accordingly, then.
Easiest solution would be setting damage_fatigue_mult to 0 globally, and then allowing GMs to re-enable this or customize mobs going down the road as needed.
I still _worry_ that blanket disabling this will result in yet another feature that's forgotten about bc it never starts on, but I'll make the adjustment as needed.

As for the calculation - it's hard and fast, and I agree, could be tuned.
I'll make an adjustment that's a little more sane here and push it in a moment.

Rykka-Stormheart:
New formula is `round(rand(1,3) * damage_fatigue_mult * clamp(((rand(2,5) * (h / getMaxHealth())) - rand(0,2)), 1, 5))`
At worst, `3 * 1 * 2` = `6` with worst possible RNG.
I can tweak this further so the initial rand is `rand(1,2)` and that will result in the prior returning 4 instead of 6.

TankTheBirb:
I have a fairly open friday I can put towards testing this, week been busy. Poke me friday night UTC-4 if I forget pls. This PR has good potential and I greatly appreciate that you're willing to adjust it to suit the server.

Heroman3003:
I haven't looked to deep at the formulas, but assuming the system is just a backend thing and no longer applied to any mob by default, I can't really have a good reason to deny.

KillianKirilenko:
Whilst I know Virgo isn't a, let's say, 'pve heavy' server, I do think stuff like this is a good lever to have for balancing new content down the line, whether for POI usage or rare events (since I know Enzo has been intermittently running his salvage crew stuff).

As long as nothing particularly pivots on it or could go horribly wrong if it's left relatively unmaintained, I see no reason not to have something like this. Could be useful someday.

Enzo-Leon:
Having the tools to help adjust and tend to changes with better ease, and tailor to depending the crew that takes part in an event too can make having a matching difficulty for them for engagement, as well as helping ease off the danger if the players seem more focused on RP versus funny unga.

PR #14764, opened 2023-04-07
Selective Markings Port

Rykka-Stormheart:
Ports the following markings upstream, plus an emote:
* Maned Wolf Markings
* Head Paint
* Sect Drone Body
* Sect Drone Eyes
* Sect Drone Tail
* Sect Drone Wings

Adds roarbark emote.


Heroman3003:
I Recommend testing sect body specifically just in case. Beyond that looks fine.

PR #14767, opened 2023-04-08
Concussion Maul

KillianKirilenko:
I wanted to try making a spicy hammer, so I made a spicy hammer. Behold, the concussion maul!

The concussion maul is a slow but powerful melee weapon that must be manually charged up prior to every attack for maximum effect. Whilst it can deal ruinous damage and throw targets away from the wielder when charged, its very slow swing speed, manual charging between attacks, limited number of uses per cell (4 from a standard weapon cell), and sheer size make it difficult to use effectively. Uncharged, it has no knockback or stun effect.

It also gains increased knockback distance when used in disarm mode, with the standard disarm-mode drawback of halved damage.

Finally, it also has limited destructive and industrial applications, as it can easily smash through certain obstacles whilst charged, and can pulverize most rock terrain. It *will* destroy artefacts though, and this still consumes power from the cell.

Low-quality demo GIF:
ezgif-3-be08175925

TODO:
- [x] basic code/functionality. works as intended.
- [x] balance. should probably use 2h wielding mechanics.
- [x] sprites, could maybe do with something more here.
- [x] sfx, functional but could use some spicing up.
- [x] actual mining/destructive functionality?

~~In draft mode so it doesn't get accidentally merged before it's ready.~~

In summary: a slow, powerful melee weapon that allows you to absolutely body single targets in exchange for being weak against groups of enemies. Doesn't have the raw killing power of a fire axe or katana, but has slightly more utility so long as you have enough power cells to spare.


KillianKirilenko:
Updated. Now has;
- Short weaken on charged hits if used two-handed (increased duration w/ disarm stance)
- Wielding mechanics, significantly reduced power if used one-handed
- In-hand sprites

KillianKirilenko:
Updated again! Now has:
- A meatier impact sound.
- Special destruction interacts with glass/windows (same as fireaxe, but requires the hammer be charged & wielded).
- Mining interacts (will mine out a turf instantly if charged and wielded, but destroys artefacts)
- Loot sources (abandoned crates and illegal weapon spawners, with a high cost and low chance)

Also ended up retooling a couple of procs for the special interacts.

And if it wasn't obvious this is being very thoroughly tested, to be 100% sure it's working as intended.

Screemonster:
Can we put an IsAdvancedToolUser() check on the charge feature?
A feral xenochimera (or a monkey) with one of these things will be able to grasp the concept of "hit things with big hammer" but fiddling with the switch to prompt the charge would probably be beyond them.

(also balancewise a feral 'mera with one of these things charged up has the potential to be fucking terrifying)

KillianKirilenko:
Sure thing. Give me a bit to square it away though, I came down with something over the start of the week and still feel like ass.

KillianKirilenko:
Easier than I thought! Made it so only advanced tool users can remove/install cells and charge the hammer.

PR #14778, opened 2023-04-10
Adds PrettyBuild to Tasks for VSCode

Smdytb1:
No more opening up your file directory to run the .bat, just do it from vscode instead.

add the deprecated label to the old tgui build if you are no longer using that in your workflow


Smdytb1:
Proof that I tested it :p 
Code_lotychKIsH

ItsSelis:
Do note, that the prettify process is supposed to run automatically when saving the file. 
But iirc that is being triggered by the Prettier extension if installed, so having a manual button doesn't hurt.

PR #14783, opened 2023-04-11
Fixes some Vox suits from having missing icons in Security

Smdytb1:
regular armor and the HoS armor now have placeholders so they aren't just. **SUIT**
may need to look through to find something more appropriate for the standard vest


AdmiralDragon:
Poor vox, always getting broken more and more with updates.

PR #14785, opened 2023-04-12
Rascal's Pass tweaks

KillianKirilenko:
A couple of minor adjustments to RP:
- Fixed all instances of super-tough reinforced windows (mostly in xenobio I think) and restored them to normal health values like all the other r-windows around them
- Fixed some old/misnamed instances of inset disposal units from the pre-auto-resleeving deposit days
- Moves the QM's office over a few tiles to make room for a small storage office with cargo tech job lockers + job clothing vendor
- Shuffles a few things in the QM's office around to account for the change in door layout
- Made the main table in cargo smaller for ease of moving crates around, shoved some of the package wrappers into a wall cabinet, and threw down some yellow 'put crate here' markings.

Excuse the number of commits, I got a bit slapdash with the testing and and such.


Skysteak:
What is the reasoning behind Xenobio losing those reinforced windows?

KillianKirilenko:
> What is the reasoning behind Xenobio losing those reinforced windows?

They're not "losing" them *per se*, they're just r-windows that inexplicably had `1e+006` health instead of the standard value, and this *only* applies to certain west-facing windows... which are all up against walls, and thus wouldn't actually provide any additional protection from angry slimes (since the cells face north/south).

foopwo:
Do your changes to the cargo department change the overall profile of the building? I would like to know whether or not the nanomap image needs updated.

KillianKirilenko:
> Do your changes to the cargo department change the overall profile of the building? I would like to know whether or not the nanomap image needs updated.

Very slightly. The QM's office has been moved about three tiles west, and a small room extends the former space one tile east. A little extra plating was added on z3 to cover the roof of the ofifce. You can see in the MapDiffBot2 data on the Checks tab, although it's extended out a bit since xenobio is also on z2.

PR #14789, opened 2023-04-13
Lipozine Plus

SatinIsle:
Adds a drug called Lipozine Plus, made with lipozine and diethylamine. Causes both a nutrition drain and a steady drain of actual weight from the person who has drunk it. Drains weight at a rate of 3 lbs per 1u, has an overdose limit of 30u with the normal toxin effect. Does not check for weight loss rate, as it's pretty much a scene tool used to modify weight directly. Stops draining weight when it reaches 50lbs.
Tested locally and seems to work as expected.

I'm sorry that I didn't want to figure out how to make an overdose cause you to spontaneously combust, tost.

### DIFF
--- a/code/modules/reagents/reactions/instant/instant_vr.dm
+++ b/code/modules/reagents/reactions/instant/instant_vr.dm
@@ -163,6 +163,13 @@
 	required_reagents = list("carbon" = 3, "iron" = 1, "nitrogen" = 3)
 	result_amount = 7
 
+/decl/chemical_reaction/instant/lipozineplus
+	name = "Lipozine Plus"
+	id = "Lipozine Plus"
+	result = "lipozineplus"
+	required_reagents = list("lipozine" = 1, "diethylamine" = 1)
+	result_amount = 2
+
 ///////////////////////////////////////////////////////////////////////////////////
 /// Reagent colonies.
 /decl/chemical_reaction/instant/meatcolony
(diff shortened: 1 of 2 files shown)


Heroman3003:
I'm cool with the idea and all that, and code looks fine, although I personally almost want A) a chem that accelerates the weight GAIN and B) a cooler name than 'plus'. Not gonna demand it, just sayin, this can go beyond.

Heroman3003 (approved):

SatinIsle:
Tbh, I think one that does the opposite would be good too, I'll add it in a little bit.
As for the name, I'll have a think on that one. My main thought on Plus was that it's basically lipozine but it does what you'd actually expect, lol.

SatinIsle:
> I'm cool with the idea and all that, and code looks fine, although I personally almost want A) a chem that accelerates the weight GAIN and B) a cooler name than 'plus'. Not gonna demand it, just sayin, this can go beyond.

I have changed the name to Lipozilase and added a new drug called Lipostipo that does the opposite (minus adding nutrition, decided it was better if it just added fat directly in a super unhealthy way).

Heroman3003:
I would suggest making lipostito drain nutrition too, to simulate accelerated fat conversion. But beyond that, yis, looks good.

SatinIsle:
> I would suggest making lipostito drain nutrition too, to simulate accelerated fat conversion. But beyond that, yis, looks good.

I agree with the suggestion and it also crossed my mind that it's a cool alternative to other fattening methods if it actively drains your nutrition. Someone who wanted to avoid nutrition going down could always mix the drug with nutriment. Done!

PR #14797, opened 2023-04-16
RP-Map-Adjustments-Addendum

Enzo-Leon:
Missed a few things that I meant to add for RP!

-Disposals Loop updated to have all blue-bins send equipment to auto-resleevers
-Blue Bins now properly state Auto-Resleeving
-Original Resleever bay disposal exit removed, replaced with a washing machine
-Mens Bathroom on level 1 now updated and contains a blue bin.
-Dorms now has a blue bin in the may lobby of the dorm area
-Firing Range has missing electronics in place, as well as adding extra practice toys and equipment

### DIFF
--- a/maps/groundbase/gb-z1.dmm
+++ b/maps/groundbase/gb-z1.dmm
@@ -437,12 +437,24 @@
 /turf/simulated/floor/tiled/dark,
 /area/groundbase/security/lobby)
 "aZ" = (
-/obj/structure/disposalpipe/segment,
-/obj/machinery/atmospherics/pipe/simple/hidden/black{
-	dir = 10
+/obj/machinery/atmospherics/pipe/simple/hidden/scrubbers{
+	dir = 8
 	},
-/turf/simulated/floor/tiled,
-/area/groundbase/engineering/lobby)
+/obj/machinery/atmospherics/pipe/simple/hidden/supply{
+	dir = 4
+	},
+/obj/structure/cable/yellow{
+	icon_state = "4-8"
+	},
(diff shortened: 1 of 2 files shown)


maps/groundbase/gb-z1.dmm
```diff
@@ -9324,19 +9363,13 @@
...
-	id = "z1stall4";
-	pixel_x = -27;
-	specialfunctions = 4
+/obj/effect/floor_decal/corner/paleblue{
+	dir = 10
 	},
-/obj/effect/floor_decal/rust,
-/obj/effect/decal/cleanable/dirt,
-/turf/simulated/floor/tiled/asteroid_steel,
+/turf/simulated/floor/tiled/white,
```
TankTheBirb, line 9372: As a note, the reason why the men's bathroom looks grimy in contrast to the women's bathroom is a map joke. Men's restrooms tend to look worse than women's restrooms. It was taken to a silly extreme here for the joke, considering the restrooms are ungendered on the other maps. If you're trying to fit a blue bin in, just remove the nanomed since it is rarely (if ever) used and in an emergency, the adjacent bathroom has one. 

Also the men's now has no mirrors. 
image

Heroman3003: I think joke being gone is perfectly fine personally, to note.

Enzo-Leon: I FORGOT THE MIRRORS DAMNIT. I can get that on my next pass by over the next few weeks since we are map rotating soon anyways.

PR #14800, opened 2023-04-16
Adds FerretDook.ogg and the emote dook

Wulfestwulf:
Adds a ferret dooking emote.


Wulfestwulf:
[https://sndup.net/rm82/](url) link to the raw audio

PR #14806, opened 2023-04-18
One-way conveyor switches

ReoDaProtovali:
Makes it possible to make one-way switches that are not limited to being mapper only, inspired by other codebases that do similar things (TG and Yogs I know do this.)

You can make a switch into a one-way (or make a one-way back into a two-way) by wrenching it with it's service panel open.

The original one-way lever is changed to use the new system as well, so maps dont need to remove the levers or anything. Aside one which for some reason was a var-edited one-way switch to be reverse only?? I commented out the var-edit so it wouldn't error.

### DIFF
--- a/code/modules/recycling/conveyor2.dm
+++ b/code/modules/recycling/conveyor2.dm
@@ -332,7 +332,7 @@
 /obj/machinery/conveyor_switch/oneway
 	oneway = 1
 
-/obj/machinery/conveyor_switch/examine
+/obj/machinery/conveyor_switch/examine()
 	.=..()
-	if(oneway)
-	. += " It appears to only go in one direction."
+	if(oneway == 1)
+		. += " It appears to only go in one direction."
--- a/code/modules/recycling/conveyor2.dm
+++ b/code/modules/recycling/conveyor2.dm
@@ -207,6 +207,7 @@
 	var/position = 0			// 0 off, -1 reverse, 1 forward
 	var/last_pos = -1			// last direction setting
 	var/operated = 1			// true if just operated
+	var/oneway = 0				// Voreadd: One way levels mid-round!
 
 	var/id = "" 				// must match conveyor IDs to control them
 
@@ -265,7 +266,7 @@
 		return
 
 	if(position == 0)
-		if(last_pos < 0)
+		if(last_pos < 0 || oneway == 1)
 			position = 1
 			last_pos = 0
 		else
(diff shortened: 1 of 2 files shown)


code/modules/recycling/conveyor2.dm
```diff
@@ -315,22 +316,23 @@
...
-	update()
+/obj/machinery/conveyor_switch/oneway
+	oneway = 1
 
-	// find any switches with same id as this one, and set their positions to match us
-	for(var/obj/machinery/conveyor_switch/S in machines)
-		if(S.id == src.id)
-			S.position = position
-			S.update()
+/obj/machinery/conveyor_switch/examine
```
Runa-Dacino, line 335: this should be examine(), not examine.

If you fix that, the workflow complaints should hopefully disappear.

maps/tether/tether-05-station1.dmm
```diff
@@ -24233,7 +24233,7 @@
 /area/gateway/prep_room)
 "kHf" = (
 /obj/machinery/conveyor_switch/oneway{
-	convdir = -1;
+	//convdir = -1;		//Incompatable with new behavior -Reo
```
Runa-Dacino, line 24236: I feel this change should really, really be tested by hosting the tether map and going to this conveyor.

Runa-Dacino (changes requested):
You forgot to do examine(), and that's breaking the code.

I strongly advise compiling it locally and testing if the local server boots in the first place after making the requested changes. 

Also, you can and probably should replace if(oneway == 1) with just if(oneway) seeing as it's just a value of 0 or 1. 

Could even go and replace 0 and 1 in your var definitions with TRUE/FALSE. Then, rather than
oneway = 1 and oneway = 0, do
oneway = !oneway to invert it from false to true and vice versa.

maps/tether/tether-05-station1.dmm
ReoDaProtovali: I have no idea where this belt is-

ReoDaProtovali: I found it, and it uses belts which could easily be reversed. it's also linked to literally two belts, in an area that i've rarely seen used
image

ReoDaProtovali: the belt setup also seems to be a bit busted, probably because one belt goes in the wrong direction in reverse while the other doesnt. are conveyor corners only available in one direction?

ReoDaProtovali: REGARDLESS- the issue it causes is nonexistent, they can still just switch the switch twice if the singular belt being used is important to be used

Runa-Dacino: That alleviates concerns on this part!

Runa-Dacino (approved):
We'll have to see if the bots complain still. If they don't, we should be good.

maps/tether/tether-05-station1.dmm
KillianKirilenko: Ah yes, the surface cargo shop. Replaced the old surface mining area when that hardware was all moved up to station 1 with cargo.

I'd recommend just deleting the offending line from the .dmm though, those are map files that are basically never read by anything except a map editing program and the commented out line (+comment) will either go completely unseen, or worse, the editor might try to strip the comments out later and accidentally break stuff.

PR #14807, opened 2023-04-19
Chocolate Allergy

SatinIsle:
Adds a neutral trait for an allergy to chocolate.
Adds ALLERGEN_CHOCOLATE tag to a number of food/drink reagents, such as coco, chocolate, chocolate milk.

Tested and it seems to work fine with the foods that I tested (chocolate bar, hot chocolate and a few others).


Heroman3003:
I think cocoa allergy is more 'proper' way to call it but chocolate is also more common so it doesnt matter much either way. Looks good anyway.

Heroman3003 (approved):

Issue #14810, opened 2023-04-20
[GENERAL] Chocolate Allergy


Aerodynamicscop:
#### Steps to reproduce
- (Step 1) select trait
- (Step 2) Load in character
- (Step 3) Eat chocolate item, nothing

 
#### Anything else you may wish to add:

![](https://i.gyazo.com/afdc3a649987e3bfc921d044c1fc0f61.png)

![](https://i.gyazo.com/bcf22b1b6721edda45443bd8425ed084.png)

PR #14811, opened 2023-04-20
Actual Chocolate in donuts

SatinIsle:
Puts the actual chocolate reagent in the chocolate named donuts, mostly so that they trigger chocolate allergies but also taste right.

Partially resolves: https://github.com/VOREStation/VOREStation/issues/14810 But that issue also lists a problem with displaying traits, which I have not looked into.

image


Heroman3003:
Will make some foods SLIGHTLY more nutritious but fine by me

PR #14829, opened 2023-05-01
swap it, don't drop it

TankTheBirb:
SD to tether


KillianKirilenko:
Hang on, weren't we on tether before it rotated to SD? Surely we should be back on RP now.

PR #14835, opened 2023-05-05
[WIP] Motion Sensor

KillianKirilenko:
Decided to dust off a *very* old sprite and put together another little project to see if I could. This time, a lifesign scanner! Draft/WIP to gather feedback whilst I tinker.

image

It scans the surrounding area for lifesigns, and after a short delay, outputs a general headcount of carbon, silicon, and simplemobs within range scanning range. This requires power from a device cell and for the user to remain still whilst operating it, It doesn't indicate direction or distance, just quantity.

Currently there's no means of acquiring them, but I'm thinking they should probably be a science thing? If I can figure it out I might tool around with making them upgradable using components, which could maybe enable other detection functions as well. Not sure.

~~No it's not a svarog detector from STALKER, shush.~~

There's also a secret setting that allows it to detect ghosts, for giggles.


Heroman3003:
I assume this is dead?

KillianKirilenko:
I was hoping for some actual feedback on its acquisition and functionality before doing any further work, to be honest, but that never really went anywhere. Probably because I didn't talk about it much.

If it seems like something folks might want to have though, I might see about dusting it off to add component upgrading, and fix the dmi conflict whilst I'm at it.

PR #14836, opened 2023-05-05
Removes the visible emote from toggling AR hud to bring in-line with …

Runa-Dacino:
…other toggles

The advantage brought by informing people around the user is significantly diminished by it being potentially spammy and distracting. I'd initially added it as a "well, knowing if someone got sec/medhud on you is potentially important", but it's not really relevant for our own usecase. Plus, you still got the examine panel so it's not nearly enough of a big deal to justify the visible emote


Heroman3003:
sure

PR #14845, opened 2023-05-09
More Raptor Borg Modules

SatinIsle:
Adds the following raptor borg modules:
Traumahound
Custodial Hound
Construction Hound
Service Hound (includes a V-4000 version that has a dumb tophat waistcoat and monocle on request.) 
Supply Hound

Tested and seems to work fine!

image


AviiCinders:
YES! YES!

PR #14859, opened 2023-05-14
Mech Nanopaste repairs fix

### LINKED ISSUE #14846: [GENERAL] Mech interaction with nanopaste
Possibly a incorrect or erroneous line of code relating to nanopaste being applied to mechs/exosuits that causes a big problem. 

The mech/exosuit to be fully to partially repaired and using up the nanopaste as an alternative method of repairing with a welding tool.

#### What actually happened
Attempting to repair a heavily damaged exosuit with nanopaste will cause a (what I think to be) endless loop somewhere in the code of it attempting to be used, even if the entire bundle of nanopaste was used, causing the server to heavily chug to the point of unplayable... so its abit URGENT. 

#### Steps to reproduce
- Step 1: Obtain heavily damaged mech.
- Step 2: Apply nano paste to the interior of the mech.
- That's it..
 
- I, myself do not know much about coding talk and such, but I've heard and seen this before.. it 'should' have something to do between the two things interacting, something about the command not ending correctly due to how it might be written.. but I do not know for sure.

KillianKirilenko:
This fixes the repair interact for nanopaste on damaged mechs by overhauling the relevant proc section(s). This isn't a *perfect* fix of the interact as the `do_after` for mechs won't cooperate no matter what, but I'm calling that an acceptable loss in the face of an otherwise server-killing issue.

It also adds additional feedback to the interacts for the repair process and makes advanced nanopaste more effective at repairing damaged mech components.

Tested fairly thoroughly, and fixes #14846

### DIFF
--- a/code/game/mecha/components/_component.dm
+++ b/code/game/mecha/components/_component.dm
@@ -138,11 +138,25 @@
 		var/obj/item/stack/nanopaste/NP = W
 
 		if(integrity < max_integrity)
+			to_chat(user, "<span class='notice'>You start to repair damage to \the [src].</span>")
 			while(integrity < max_integrity && NP)
-				if(do_after(user, 1 SECOND, src) && NP.use(1))
-					adjust_integrity(10)
+				if(do_after(user, 1 SECOND, src))
+					NP.use(1)
+					adjust_integrity(NP.mech_repair)
+
+					if(integrity >= max_integrity)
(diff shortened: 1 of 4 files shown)


SpaceMan-Jim:
I see you found it, yeah that do_after sounds like what was mentioned when a fix for it was talked about. Sorry I couldn't be more help besides report it.

PR #14866, opened 2023-05-19
adds a smattering of clothes sourced from SR & TG

Detective-Google:
the main focus is bringing over a bunch of the jumpskirts, given I and some others I spoke to prefer the way these look over the pencil version.

at the moment I have 0 intention of putting these in loadout because it would expand the scope quite significantly and dealing with dumb clothing icon code took probably 3 years off my lifespan. Some items are intentionally not able to be acquired by normal players, CC outfits, that sort of thing. 
compiles & runs fine, 80-90% of the clothes have been spawned and tested, I'm sure I missed a few testing whilst going down the list but that's what bug reports are for :)
image
here's a preview of some of the security items. generally they are all up to this level of quality. None of this was sprited by me, this is all work by the talented artists at those other servers.


Heroman3003:
conflict in hat file

PR #14872, opened 2023-05-20
Uniform Selectors

KillianKirilenko:
Adds a new category of uniform selectors which include all the valid, role-restricted uniforms for a role under a single option. The intent of this is to provide easier selection of job uniforms in what is now an increasingly bloated loadout list.

This __**will not**__ replace or remove anything at the present time, to avoid disrupting loadouts.

image

- [x] Command
- [x] Security
- [x] Cargo
- [x] Engineering
- [x] Science
- [x] Medical

May eventually be expanded to cover suits and other slots as well, but probably in another PR.


KillianKirilenko:
Uniforms are pretty much done. Doesn't include a couple of jobs simply because they had only one alternative job-locked uniform.

Includes the new outfits from #14866 where appropriate.

PR #14874, opened 2023-05-20
Adds new preference for vore privacy

Runa-Dacino:
* Adds a "global" pref to set default belly privacy behaviour
* "global" pref shows up with prefs like "devourable"
* Adds per-belly pref that, if changed from "default", overrides "global" pref
* per-belly pref in "options" of belly, same column as "save digestion"
* Global options: True/False (subtle/loud)
* Per-belly options: "default", "subtle", "loud"
* Governs attempting to ingest
* Governs successful ingestion
* Governs expelling one/all contents
* does NOT work for slipvore
* Ingestion only works for /mob/living
* Expulsion also works for objects
* Expulsion also has its volume lowered
* Involves touching TGUI

**Why we need it**

When announcing my vampiric subtle pref, I've had multiple players express interest in the same functionality for vore.
Some people are more self-concious/intimate about such, and may feel unable to do vorny things if they feel someone might see the BIG RED TEXT. This should enable them to have their vorny fun now.

This is especially useful for vore involving unaware or significant size difference, potentially allowing people to do AV with micros in the bar without drawing attention for instance while they sit.

It has been raised that you can still see the sprite disappear - but while true, it draws far less attention.

**Possible "issues"**

This may cause issues down-stream where they have security engage in vore-crime punishment, or otherwise have a culture where bystanders are expected to interact in all cases. This could be alleviated by adding a second check for prey to enforce "loudness", but it's not particularly relevant AND logic is that the predator is the one determining resist messages and examine visibility, so it's OK to let the pred decide on their own.

**How it looks:**
image
image

edit: added images


Heroman3003:
Cant review rn, hope this works, but seems good feature-wise

PR #14885, opened 2023-05-22
Item banks for all!

KillianKirilenko:
you get an item bank, you get an item bank! and you get an item bank! everyone gets an item bank!

Edits each centcomm map to have an item bank by both sets of elevators, in case you, like me, had something you wanted to stash but didn't know where to put it and didn't realize there were no banks at Central until it was far too late.


Wickedtemp:
This is pretty rad! I always liked the item bank idea and look forward to any other expansions. It has good potential for some semblance of round persistence without outright ruining or deteriorating certain jobs.

PR #14887, opened 2023-05-22
Adds new human ability: "Explain Custom Traits"

Runa-Dacino:
**What this does:**
Gives all human mobs (ergo: anyone that's not a simple mob or a silicon) a new ability. 
Unless they do not have any traits (at which it returns and notifies them they have an empty trait list),
Pressing this ability brings up a dialogue box listing choices of TGUI, print to chat or cancel.

Unless user cancels, a list of keys is generated from the name vars of each custom trait their species has (confirmed working for custom and non-custom species alike), which is based completely on the character creation window choices. A number of associative lists are likewise populated using the name var of the given trait we are iterating over as the key for easier retrieval. These lists are used for both to_chat and tgui functionalities

**If the user picks the to chat option**, a SPAN_NOTICE formatted message is sent to their chat with line breaks. If the dev did not define tutorial for their trait, the user is recommended to go to #dev-suggestions and offer their own guide.

**If the user picks the TGUI option,** they are presented with a screen similar to the following:

image

Not shown: a scrollbar appears if the text or number of traits are longer than the window.

Furthermore, the text in the right element wraps around automatically, removing need to worry about line lengths for developers.
The only thing developers must keep in mind is to use <br> for their traits over \n. This is reinforced twice - once in the trait_tutorial.dm comment at the top, and once for the var/tutorial itself.

**Why we need this:**

It has often come up before as mentorhelps, #cadet-academy questions and during addition of more complexly behaving traits that just the trait's desc var does not provide sufficient information for proper usage.
While the desc var could be made more descriptive, it is not accessible by users during gameplay easily AND it should be short and and optimized to guide user choices, rather than a detailed description of the respective mechanics. 
While we do have a wiki page for traits, downstreams may lack it causing potential confusion for their players, and our own players often forget about the wiki's own existence. Plus, large walls of text that some traits may require may scare users off of the wiki.

**Commit Contents**
[Adds new human ability: "Explain Custom Traits"]
* New verb of category "Abilities"
* New var for /datum/traits: tutorial
* Prints to_chat: Name, Desc, Category, Tutorial
* var/tutorial defined as a string by default
* Developers are to use <br> over \n when defining subtypes of tutorial
* bloodsucker obligate/freeform have a tutorial defined as example.

[Adds TGUI functionality to Trait Tutorial](https://github.com/VOREStation/VOREStation/pull/14887/commits/cd602eb20e5394f3ad069396b2fedcdb5bd12549) 
[cd602eb](https://github.com/VOREStation/VOREStation/pull/14887/commits/cd602eb20e5394f3ad069396b2fedcdb5bd12549)
* Creates trait_tutorial_tgui.dm to act as backend, adds to vorestation.dme
* Creates TraitTutorial.tsx to act as frontend
* Modifies traits_tutorial.dm for TGUI compatibility
* Adds comment to var/tutorial definition on how to make it pretty in TGUI
* Modifies logic in trait_tutorial to catch empty trait lists.
* Edits bloodsucker obligate/freeform to add another <br>


Runa-Dacino:
Removed [DNM], we're good for merging.

Heroman3003:
I'm personally really unsure about it being a verb at all, much less a verb in Abilities tab. I personally think this should just be a button in char setup, but also thats just my opinion so I'm gonna wait on thoughts of other staff in that regard.

Runa-Dacino:
> I'm personally really unsure about it being a verb at all, much less a verb in Abilities tab. I personally think this should just be a button in char setup, but also thats just my opinion so I'm gonna wait on thoughts of other staff in that regard.

This is specifically NOT for character set-up.

This is for when you are playing the game actively, and want a reference for how your ability works. This is why it was even given a whole TGUI functionality, so you can keep a little window open on the side to see controls if your trait is particularly advanced/complicated.

Furthermore, the reason it's a human ability/verb is simple:
trait list is saved on the instance of species initialized within the human object. This is the cleanest way to extract it without having to figure out how to access the client's prefs during gameplay: human.species.traits

Heroman3003:
Regardless, at the minimum, this verb should ideally be in OOC tab, as well as not show up at all if you don't have traits to begin with.

Runa-Dacino:
> Regardless, at the minimum, this verb should ideally be in OOC tab, as well as not show up at all if you don't have traits to begin with.

The OOC tab is cluttered to all hell and impossible to parse.
This makes sense under the abilities tab, as that is where you look to do your funny special traits.

TankTheBirb:
I agree with Runa in that the abilities tab is probably the better place to put this. It makes sense as a quick reference guide rather than a character setup guide, which a wiki article would probably do easier with (and iirc we have a trait wiki page). If it showing up in the abilities tab proves to be a detriment, it can be moved to IC later.

Runa-Dacino:
I have left the verb in the abilities tab for earlier stated reasons (this verb exists to tell players how to do things if confused. it should therefore be somewhere obvious they can find without guidance. This is the abilities tab. Placing it elsewhere will cause it to never be found & used even if someone desperately needs it.)

I did make it only show up if you got at least 1 trait though.

PR #14895, opened 2023-05-25
Light Lighting Pass (Tether)

KillianKirilenko:
A light pass of lighting on Tether to fix up a couple of spots with poor lighting.

Should also fix the top conveyor on the cargo shuttle going in the same direction as the bottom one.


Heroman3003:
Looks fine

PR #14899, opened 2023-05-26
Altevian Packaged Meals are no longer made of Mylar

Youtubeboy139:
## About The Pull Request

The cool Altevian Packaged Meals use the Packaged code that already exists in the game. When opening it plays a custom soundbite of a chips bag opening (specifically a Lays bag if I remember correctly) but a Lays bag the Altevian Packaged Meals ain't. While I didn't create a custom soundbite for these, I did make it where the Bluespace Upgraded Box makes a 'click' sound as opposed to a plastic/mylar bag opening. Could be changed later down the line, either to encompassed more than the Altevian Packaged Meals simply by changing the name of the check or by adding a custom sfx for the Altevian Packaged Meals opening.

EDIT:
**_This check should only be applied to the MRE like Altevian meals, not every piece of Altevian food!_**

No more mylar cases for the fancy Altevian Packaged Meals means players won't be confused by it. I guess. Honestly I think this just makes better sense than the sound effect currently being used by the Altevian Packaged Meals.

## Changelog

:cl:
code: Added a check for Altevian Packaged Meals
sound: Changed sound effect played by opening Altevian Packaged Meals
/:cl:

### DIFF
--- a/code/modules/food/food/snacks.dm
+++ b/code/modules/food/food/snacks.dm
@@ -37,6 +37,9 @@
 	/// Packaged meals that have opening animation
 	var/package_opening_state
 
+	/// Packaged Altevian Meal Check (for ones with fancy opening animations)
+	var/altevian = FALSE
+
 	/// If this is canned. If true, it will print a message and ask you to open it
 	var/canned = FALSE
 	/// Canned food switch to this state when opened, if set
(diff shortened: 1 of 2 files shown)


Heroman3003:
First, I think the whole 'package opening sound' should just be a var on the food item, rather than having it be binary choice like that, in case more obscure 'packaged foods' are added later.

Second, all altevian stuff is Enzo's purview, so make sure you have his okay to make changes to those.

Youtubeboy139:
> First, I think the whole 'package opening sound' should just be a var on the food item, rather than having it be binary choice like that, in case more obscure 'packaged foods' are added later.
> 
> Second, all altevian stuff is Enzo's purview, so make sure you have his okay to make changes to those.

First issue addressed in `The sound of Packed Food` #14902

Second issue won't be addressed, there's bad blood between Enzo and I that I don't even want to touch

PR #14905, opened 2023-05-27
Caps Downside of Biofuel Processor Traits

VictorZisthus:
## About The Pull Request

If species.synthetic_food_coeff is = 1, then charging is efficiency is zero while still draining power. Likewise values greater than 1 will result in power LOSS while charging. This fixes both issues.

Minimum charge efficiency is now 10%.

Allows someone with a perfect biofuel processor to still get some charge out of a cell charger, and prevents charge draining.

## Changelog
 🆑 qol: Caps Downside of Biofuel Processor Traits /🆑

### DIFF
--- a/code/game/machinery/rechargestation.dm
+++ b/code/game/machinery/rechargestation.dm
@@ -110,7 +110,7 @@
 
 			// Also recharge their internal battery.
 			if(H.isSynthetic() && H.nutrition < 500) //VOREStation Edit
-				H.nutrition = min(H.nutrition+(10*(1-H.species.synthetic_food_coeff)), 500) //VOREStation Edit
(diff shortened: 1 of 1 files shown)


KillianKirilenko:
Seems reasonable enough. A minor oversight on my part I suppose; I didn't expect anyone would ever make traits that gave synths equivalent energy gain to organics.

Heroman3003:
Seems fine but also oh god this template is spreading

Youtubeboy139:
> Seems fine but also oh god this template is spreading

_I take full responsibility, but at the same time I refuse to stop_

VictorZisthus:
> Seems fine but also oh god this template is spreading

Oh, I thought the template was required to do merges here so I just copied the first one I found, lol

PR #14908, opened 2023-05-28
Remove two maw posters and butt from default spawning

lm40:
Staff decided they wanted this, this is a hacky fix but I said I would do it so here it is. The two maw posters and butt poster will not generally show up now, however I have included another hacky fix to the custom posters in cargo so that they can actually be used to select any poster, which I am sure was always intended but did not work; that way people can still order the funky posters and put them up intentionally if desired. If someone wants to fix that properly it is very easy, but very tedious; every poster should have a listing_name but only a small subset do.

I honestly haven't tested this super thoroughly; it compiles, I don't see why it shouldn't work, that's it.


Heroman3003:
Eh, should work for now.

PR #14911, opened 2023-05-28
Adds a new object: Picnic blankets! They can be folded & unfolded

Runa-Dacino:
### Changes
*Adds 2 obj types: one structure (unfolded), one weapon (carried)
*When in carried mode, has a verb to unfold and fill a up to 3x3 area
*Filling is blocked by Dense objects, except for tables.
*Tables follow a special logic: if center turf is a table, acceptable. Otherwise no
*However, if center turf is a table, we no longer accept tableless turfs
*If on table, we set layer just over table. Otherwise, we set it below all objs incl hiding ones
*Also adds a mapper variant that calls fold_out() on unfold() on init
*Mapper variant can be set to be unfoldable
*Creates codersprites for it all. It's "passable"
*To avoid someone trying to unfold the blanket onto a turf that has way too much trash and thus lag the server, we check the len of the turf.contents and only iterate over it for densities if and only if it's smaller than 10 items
*Added it to Bits & Bobs vendor

### Why we need this:

Both Tether and Rascal's Pass offer excellent opportunities to do picnic - as evidenced by the scug picnic!

Unfortunately, the absence of proper tablecloths forced the poor scugs to rob NanoTrasen for proper linens.
This is no longer required.

Players and mappers alike may now create picnic scenarios!
Picnic basket not included.

### How it looks:

**The carry sprite:**
image

**Unfolded completely:**
image

**On a table with limited space**
image

**On a 3x3 table**
image

It is technically possible to have a case where you only got the 4 corners and the center piece and that looks ugly so I didn't post it.

It's also possible to have variants where any number of edges are missing.


Enzo-Leon:
Picnic days on Rascals!

PR #14914, opened 2023-05-29
Changes The Boozemat (It's 5:45am - I can't think of any good jokes)

Youtubeboy139:
## About The Pull Request

Makes the Smartfridge and the Boozemat match in size as God intended. Traditionally the Boozemat was always the same size as the Smartfridge as it could also be mounted into a wall. So this PR reverts it to an older sprite that was once from /TG, but what's this? It still has the fancy screen from before? That's right this is a hybrid Boozemat that combines the current sprite with the "legacy" sprite while making sure the proportions and dimensions match that of the Smartfridge.

Makes the Boozemat look better in scene with the Smartfridge. That was the whole basis of this PR.

## Proof of Testing

<details>
<summary>Screenshots/Videos</summary>
 
New Boozemat in a Test Session on the SD
Screenshot 2023-05-29 041613

</details>

## Changelog

:cl:
image: Added new Boozemat sprites
/:cl:


Enzo-Leon:
Looks a lot better paired together like that! The fridge adjustment was a nice change too.

Heroman3003:
Looks good.

TankTheBirb:
Looks good to me. It may look out of place when the boozemat is alone but that's hardly a problem.

PR #14917, opened 2023-05-29
New Event Site: The Falls

Burrito-Justice:
About a year ago I submitted a map for Duke's mapping competition. For a month's worth of work, I'd say it was ok. Seeing it used in an event inspired me to fix it and add more cool bits to it... before deciding that it'd be better to simply make another map all together with the new stuff I had made in the first quarter of development.

Anyways welcome to **The Falls**, the much prettier and complex cousin of the Atoll, featuring:
- Way more new sprites that probably wouldn't fit in anywhere else, don't you just love content bloat?
- Ambience!
- Godrays???
- A small story that I wrote at like 2AM, which you can discover for yourself using a cataloguer (preferably a non-compact version).
- Waterfalls
Hopefully not-so-spoilery images:
image
image
image
image


KillianKirilenko:
the hyper light drifter called they want the eastern watershelf zone back

looks good though!

Heroman3003:
OUUUGHHH

PR #14928, opened 2023-05-31
The Suits of the Void

Youtubeboy139:
## About The Pull Request

Did you know that there's a `verb` that allows you to retract the helmet on your Voidsuit? I certainly didn't until I made this PR. CEV Eris (where the code _seems_ to have originated from) has made it to where players can click an action button to retract their worn helmet at will instead of having to right click on the suit itself and click on the verb. While I was in there I also made it to where the emergency release for the oxygen tank was more, well, emergent. Like the helmet toggle you had to right click on the suit and then click the verb. I've made it where an `AltClick` will engage the emergency release and drop the tank (or cooler if you're wearing that instead).

CEV Eris has also added an `initialize` to the original code that allows for Voidsuits to spawn with their helmets (or any other gear) pre-attached. This has not been implemented with the existing Voidsuits (there are too many and then I'd have to go into maps and change that around, just an overall headache I don't want to deal with) but I've added it for the possibility of future use.

Makes the Voidsuits more player friendly with regards to overall interaction and fixes a bug I found when coding (the helmet light wouldn't actually turn off when you retracted the helmet so I fixed that).

## Proof of Testing

<details>
<summary>Screenshots/Videos</summary>
 
Voidsuit Showcase in a Test Session

https://github.com/VOREStation/VOREStation/assets/609886/dbc2d676-2784-4257-929b-1318194fcc99

</details>

## Changelog

:cl:
qol: Makes the Voidsuits more player friendly with regards to overall interaction
sound: Added a new click noise that was ported from CEV Eris
/:cl:


KillianKirilenko:
It also appears under the Object tab, fortunately, no need to fuss with the context menu. That said, still a good change!

I would however suggest perhaps a different noise for ejecting the tank/cooler and/or a safety check for trying to eject the tank/cooler whilst they are in use, as a little added user-friendliness.

Youtubeboy139:
> It also appears under the Object tab, fortunately, no need to fuss with the context menu. That said, still a good change!
> 
> I would however suggest perhaps a different noise for ejecting the tank/cooler and/or a safety check for trying to eject the tank/cooler whilst they are in use, as a little added user-friendliness.

I'll change the sound to the original `click` noise as a filler (can't go wrong with that click noise).

I don't think there is a check for internals and even if there were that kinda negates the "emergency release" of the tank. It'd be an issue like "You try to remove the tank, but it is use!" _currently suffocating cause the tank is empty._

KillianKirilenko:
> > It also appears under the Object tab, fortunately, no need to fuss with the context menu. That said, still a good change!
> > I would however suggest perhaps a different noise for ejecting the tank/cooler and/or a safety check for trying to eject the tank/cooler whilst they are in use, as a little added user-friendliness.
> 
> I'll change the sound to the original `click` noise as a filler (can't go wrong with that click noise).
> 
> I don't think there is a check for internals and even if there were that kinda negates the "emergency release" of the tank. It'd be an issue like "You try to remove the tank, but it is use!" _currently suffocating cause the tank is empty._

Check the wearer's `internal` var. It should be referencing the tank if active, or null if it's inactive. Likewise check the suit cooling unit's `on` var.

If low volume is a concern you can also just check the tank's internal pressure and override the check with a failsafe that permits it if the internal pressure is low enough. Would be similarly simple.

Youtubeboy139:
> 

Cooler is easy, there's a var that checks if the cooler is on in the code.

Tank does not, it uses `data` to check if the mask is connected and all attempts to force it were met with errors

While it would prevent accidentally ejecting the tank while it's in use I honestly don't think players are going to kneejerk `AltClick` their suit in the heat of the moment

Enzo-Leon:
UI and bugfixing itself is nice, and I was wondering why we didn't have a helmet toggle this whole time. Other remarks seemed to of been addressed and looked into from the earlier comments!

PR #14929, opened 2023-05-31
TGUI Update

ItsSelis:
Most notable changes:
- More conversion to TypeScript (for proper code completion, in-editor error line thingies, autodocs etc.), mostly for the core files
- Unit tests for some of the Core Files so if some of those are touched in the future, tests can be run to test out if those functions still works or output the expected things on specific input
- The "Kitchen Sink" has had it's pages updated (build tgui with 'yarn tgui:build' to get the development bundle and get a nice button to see it)
- Components synced with TG (due to general improvements and new features like a bit more fixed scrolling, etc., some UIs that are always visible like the vorepanel no longer have that "eye" icon)
- A bunch of new Components like FitText
- Some old unused functions were also removed from some components (literally having had no single use in any interface)
- Updated Yarn to 3.3.1

The interfaces were so far untouched and I have roughly tested this and the UIs seemed to still work fine.

One thing that was strange which should be checked:
The belly overlay preview images in the VorePanel are not "there"?

Another thing that should be checked:
Dropdowns. I have not found a Dropdown yet.
They should still work in the same way but the "placeholder" variable/logic has been changed up.


Heroman3003:
I'll mentally mark this as "revert if anything goes to shit next week" and merge it before conflicts start piling up.

PR #14930, opened 2023-05-31
Buffs harm-injection

Heroman3003:
Now mininmum amount is no longer always 0, but amount of reagents-10, meaning that you are guaranteed to harm-inject *at least* 5 units if you have a full syringe.

### DIFF
--- a/code/modules/reagents/reagent_containers/syringes.dm
+++ b/code/modules/reagents/reagent_containers/syringes.dm
@@ -297,7 +297,7 @@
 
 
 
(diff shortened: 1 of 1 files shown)


Enzo-Leon:
Eyyyyyyyyyyy. Way better handling of this versus the comical rng.

Heroman3003:
Still RNG, just with some level of reliability built in :P

Enzo-Leon:
Yea but.. At least now we shouldn't see empty hits now.

PR #14934, opened 2023-06-01
TGUI Input Checkboxes

ItsSelis:
Instead of asking multiple yes/no questions you can now open a ui and ask all them all at once.
Basically you can give it a list (and some other parameters if wanted) and get a list in return.

Sadly I have not found a good example to replace it with spontaneously, so I only have a screenshot from the TG PR I have ported this from:
image


Heroman3003:
con

f

lick

t

ItsSelis:
> con
> 
> f
> 
> lick
> 
> t

Should be solved now, but needs a moment to run the tests.

PR #14935, opened 2023-06-01
Adds a set of admin verbs for speaking/emoting as objs/mobs/turf without possessing them

Runa-Dacino:
### What it does

- This PR adds four new admin verbs restricted to R_FUN rights in the Fun tab
  - Narrate Entity (takes 3 positional arguments in form of narrate-entity "identifier" "mode" "narration" and makes the entity associated with the identifier say/emote that. This only works for those on personal list
  - Narrate Entity (Interface) does as above, but walks the user through picking each var. It's planned to create a TGUI interface for rapid narration by just selecting buttons that stay that way and on the screen, and the user can press a button to create a input that gets sent.
  - Narrate Entity (Add ref) is a right click verb on all visible atoms & turfs and mob/living in view range  that the user can give a unique identifier that does not show up in-game for easy referencing.
  - Narrate Entity (Remove ref) removes refs to avoid bloat from no-longer-relevant entities at will
- Adding entities to ref list is logged
- Attempting to add mobs with a client, or talk/emote through them is prevented and logged.
- Guard clause prevents adding refs to mob/observer
- Valid "modes" are Speak and Emote (case sensitive)
- When using narrate entity in either mode on a mob, it uses the mob's natural say verbs and accounts for stuttering, language issues and the like.
- When using the above on a obj/turf, there's still audible/visible messages. Speaking thru an obj/turf has no "XYZ says," baked in. User must input message like so: narrate-entity "obj/turf" "speak" "hisses, "Welcome Commander"" to get obj/turf hisses, "Welcome Commander".

### Why we need this

It came to me in a dream.
In all seriousness, I remember Cody complaining how clunky it is to try and work with a group of mobs due to having to specifically possess stuff to speak as them. This should make it much easier. Unfortunately, mob movement is not included in this package.

### Possible weaknesses

I check for clients by going if(mob/living.client)
If the client has logged off, this won't work.
Unfortunately, checking for original occupant only exists apparently for human subtypes from what I saw so, it's hoping that checking for a client being present and log/ghost out logs are enough to catch potential abuse.

### Some demonstrations:
image
image
image
image
image

### Commit details
https://github.com/VOREStation/VOREStation/pull/14935/commits/4ab3c8f8829abeabd1431eedff79aec24c701819

* Makes entity_narrate() call entity_narrate_args() with input gleaned from user interface
* Sanitizes text input. Multiple times: custom IDs are sanitized on creation
* Interface text input sanitized on input
* Sanitizes arguments of argument based function as well
* Enables multi-line text input when making messages using interface.
* Multiline also works with commandline tool using \n

https://github.com/VOREStation/VOREStation/pull/14935/commits/5bcc0a2d5da7c686715e1d2d6c4c8c749b01f001
* Changed type checks to check for /atom for non-mobs
* Changed static typing to use /atom for non-mobs
* Rearranged code logic where relevant so we only check for /atom if mob/living came out invalid
* Added a guard clause against observers on adding a ref. Only added to this as there's hopefully no way to access ghosts with the remaining procs
* Also added "Remove All" as an option for removing refs. It requires confirmation.

https://github.com/VOREStation/VOREStation/pull/14935/commits/6c2ee4a4ce6edb25ba9cb6488caf9ed8ef65ecfa
* Adds verbs to client: Narrate Entity, ... (Add ref), ... (Remove ref), ... (Interface)
* Adds new Client var to hold reference to entity_narrate datum instance *Creates entity_narrate datum to hold list of unique/custom entity names, and an assoc list of name:atomref
* All listed client verbs initialize the datum onto the client var for later use when first used
* User is expected to right click mobs within viewrange to add them to ref list
* User is prompted to create a unique identifier to generate the key:value pair
* User may either request an interface to do the narration (interface)
* User may alternatively go narrate-mob "identifier" "speak/emote" "narration" into command line
* In case of argument call, message may be ommited to bring up non-multiline tgui_input_text
* User may remove entities from their personal list at will
* Users adding entities to their personal list are logged
* Users attempting to add players to their personal list are likewise logged
* Users succeeding despite this are logged if they try to speak for them
* If type is mob/living, it uses .say and .custom_emote() procs
* .say uses the mob's languages, stutters and so forth
* if type is obj/, user must specific speech verb when composing narration.
* User may narrate from any range.
* Each proc checks for R_FUN permission and prevents using if lacking rights

### DIFF
--- a/code/modules/admin/admin_verb_lists_vr.dm
+++ b/code/modules/admin/admin_verb_lists_vr.dm
@@ -166,7 +166,11 @@ var/list/admin_verbs_fun = list(
 	/client/proc/admin_lightning_strike,
 	/client/proc/resize, //VOREStation Add,
 	/client/proc/cmd_admin_droppod_deploy,
-	/client/proc/adminorbit //VOREStation Add,
+	/client/proc/adminorbit, //VOREStation Add
+	/client/proc/add_mob_for_narration,	//VOREStation Add
+	/client/proc/remove_mob_for_narration,	//VOREStation Add
+	/client/proc/narrate_mob,	//VOREStation Add
+	/client/proc/narrate_mob_args //VOREStation Add
 	)
 
 var/list/admin_verbs_spawn = list(
--- a/code/modules/admin/verbs/entity_narrate.dm
+++ b/code/modules/admin/verbs/entity_narrate.dm
@@ -1,0 +1,176 @@
+//Datum that's initialized on first calling the client procs. It's stored in /client
+//We keep a distinct names/refs list in an effort to make things speedy, and for easy checking if something is on our list.
+//We manually add to list element with procs to avoid dealing with clunky global lists that might not be relevant
+//and for ability to narrate from long range.
+/datum/entity_narrate
+	var/list/entity_names = list()
+	var/list/entity_refs = list()
+
+//Appears as a right click verb on any obj and mob within view range.
+//when not right clicking we get a list to pick from in aforementioned view range.
+/client/proc/add_mob_for_narration(E as obj|mob in orange(world.view))
+	set name = "Narrate Entity (Add ref)"
+	set desc = "Saves a reference of target mob to be called when narrating."
+	set category = "Fun"
+
+	if(!check_rights(R_FUN)) return
+
+	//Making sure we got the list datum on our client.
+	if(!entity_narrate_holder)
+		entity_narrate_holder = new /datum/entity_narrate()
+	if(!istype(entity_narrate_holder, /datum/entity_narrate))
+		return
+	var/datum/entity_narrate/holder = entity_narrate_holder
+
+	//Added /obj functionality on top of mob. It works practically the same.
+	if(istype(E, /obj))
+		var/obj/O = E
+		var/unique_name = tgui_input_text(usr, "Please give the entity a unique name to track internally. \
+		This doesn't override how it appears in game", "tracker", O.name)
+		holder.entity_names += unique_name
+		holder.entity_refs[unique_name] = O
+		log_and_message_admins("added [O.name] for their personal list to narrate", usr) //Logging here to avoid spam, while still safeguarding abuse
+
+	//We require a static mob/living type to check for .client and also later on, to use the unique .say mechanics for stuttering and language
+	if(istype(E, /mob/living))
+		var/mob/living/L = E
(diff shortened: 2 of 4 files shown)


TankTheBirb:
> Okay, honest question. How is this any better then the already existing proc make-sound? The make-sound proc already completely covers the emote side of making a object/mob do a emote quickly and easily, and emotes can also be used for 'say' uses if you do not wish to possess a object or mob.

Make sound has several drawbacks. It doesn’t print anything to chat, it is accessed by right click but isn’t available for a multitude of various object types. It also doesn’t work for turfs or mobs at all. Mobs use forcesay, which is in their player panel menu or proc called directly. I’m sure it could be improved but I’d rather make sound not change and just have something new instead.

I haven’t tested this pr yet but this is an improvement over the existing system. Which makes sense because we requested this as a GM tool so we wouldn’t have to crutch on mass Z narrates or possessing objects as ghosts. Which becomes clunky and taxing very quickly.

Runa-Dacino:
> Okay, honest question. How is this any better then the already existing proc make-sound? The make-sound proc already completely covers the emote side of making a object/mob do a emote quickly and easily, and emotes can also be used for 'say' uses if you do not wish to possess a object or mob.

A key advantage, on top of what Tank had said - is that you are able to create your own unique callsigns for your mobs and objs of importance. 

This enables using
narrate-entity "Callsign" "mode" "Message"

which can be done very quickly and easily once set up, allowing quick and easy interaction through multiple mobs and objs as a single GM without the aforementioned player panel jank.

Runa-Dacino:
After discussion with Tank, I extended functionality to include all turfs.

While doing that, I added a Remove All option. This needs confirmation.
Details in commit, OP updated.

TankTheBirb:
I kinda already voiced my opinion earlier but this PR is good and will make GM work a lot easier.

Runa-Dacino:
Although initially I did not feel it was required to sanitize input, seeing as only staff can interact with this piece of code - asking Selis for their opinion changed my mind.

We're sanitizing everything now.

While doing this, I realized I can cut down on code duplication by removing the entire logic in the interface verb for doing the speaky thing, and instead just call the command line verb with the arguments being formed from simple "request input" stuff.

I think I have no further plans to alter this project further. I do talk of TGUI functionality, but I want that as a separate PR and would probably a take a day in itself given my vision.

PR #14936, opened 2023-06-01
Little Weakref Refactor

ItsSelis:
(This mainly is one of the atomic PRs for a MC/Subsystem Update)

This comes with a bigger documentation update for weak-refs and the datums, which is pretty well done and (for the weak-refs) explains it pretty well what they are actually for.

There are of course also a few code improvements in code/datums/weakrefs.dm too (including a little thing for view variables).

Luckily, there don't seem to be too many things that currently actually have a weakref, so there aren't toooo many changes.  During testing nothing seemed to runtime so far and I did test one or two things on functionality and they seemed fine as well.


TankTheBirb:
I cannot speak to whether this should be merged or not but the documentation is clear enough that I now understand what datums and weakrefs are. So I'd say that's at minimum a really solid addition.

Heroman3003:
datumized

PR #14937, opened 2023-06-01
Little Cleanup

ItsSelis:
This only is a little code cleanup in the code/ directory.

To prevent accidental running or compiling into 515, a if-statement with an error message was built into the __byond_version_compat.dm file.

It also offers new defines that become relevant when testing or moving on to version 515 (or maybe for Downstreams if they´d like). 

Additionally, __513_compatibility.dm was removed and the relevant contents put into other files, since the if statement is not reachable as the minimum version is defined as 514.


TankTheBirb:
This seems like good bookkeeping and code maintenance. As far as I know this is a safe merge. And probably a smart one to get out of the way in case we ever move to a more recent version of byond.

PR #14938, opened 2023-06-02
Variety Add-on

Rick4242:
Adds a few extra parts and options to the character setup:

- Heterochromia markings for the _left_ eye
image
- A turkey-styled tail (thanks to Kligor#7935 for helping to make it possible)
       "Would have been a really good thing for Thanksgiving probably but I'm not going to wait that long."
g4HchCMm9I

- Shark-styled ear options as well head-fins, and three new variations of the Akula tail (_**15**_ different variations in total)
image
image
image

Also adds a (semi) new icon for the "modern wrapped coat (pants)". 
dreamseeker_A9ieQj9AKH


TankTheBirb:
More markings is always good! Great addition.

Heroman3003:
While not at all against it, I do question general purpose for second heterochromia

PR #14955, opened 2023-06-03
Adds functionality to buildmode's AI mode to adjust a.i wander behavior

Runa-Dacino:
### What it does

- Adds new functionality to buildmode that, when in AI mode, pressing MMB + shift on anything in the game screen enables all selected mobs to wander freely around their home turf.
- Alternatively, pressing MMB + CTRL turns off their wandering, prompting them to sit still until someone disturbs them.

Mobs chill out if multiple are set to go to the same tile, lining up neatly.

### Why we need this
It has been a recurring complaint that puppeting mobs is a pain in the ass because they refuse to sit still and wander around randomly.

This should put an end to that.

Furthermore, Selis's PR about mob spawning inspired me.

Using the MMB as it was yet unused for AI mode; currently allowing clicking on anything since there's no other functionality tied to that button. If that changes, we can probably wrap this in a check for turfs or something.

### Commit log
https://github.com/VOREStation/VOREStation/commit/d71653c9b693aef4fdfdd22ffc3bbb2130486ef0
* Works by clicking on ANY object/turf/mob
* MMB + Shift = Wander
* MMB + CTRL = NOT wander
* Adds this info to help button

### DIFF
--- a/code/modules/admin/verbs/buildmode.dm
+++ b/code/modules/admin/verbs/buildmode.dm
@@ -158,7 +158,9 @@
 							Left Mouse Button on AI mob            = Select/Deselect mob<br>\
 							Left Mouse Button + alt on AI mob      = Toggle hostility on mob<br>\
 							Left Mouse Button + shift on AI mob    = Toggle AI (also resets)<br>\
-							Left Mouse Button + ctrl on AI mob 	  = Copy mob faction<br>\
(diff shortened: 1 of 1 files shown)


Heroman3003:
I'm not entirely certain about such oddly specific thing occupying one of very few remaining free buttons in that mode, but eh, sure

PR #14959, opened 2023-06-03
Adds verb for players to show event interest, GMs to check this

Runa-Dacino:
### What this does:
 - Adds two new verbs, one for all clients: Show Event Enthusiasm
   - Adds user into a global list where the time they pressed the button is recorded. They can always remove themselves
 - The other verb: Assess Player Interest
   - Prints contents of the aforementioned global list in a new line for each user like in the attached image: 
image
 - Entries inactive for 30 minutes automatically removed, notifying the player that this happened
 - Players leaving the round thru the verb or tram/cryo/portal are likewise removed by cleanup() verb.
 - This list exists on a per-round basis, as this is an active, enthusiastic demonstration of interest.
 - This is also not request-spice. This merely shows enthusiasm, does not request events.
 
### Why we need this:

Months ago I recall some GMs had complained that they had had ideas for the Stellar Delight but were disheartened from pursuing them due to not knowing if people would even bite, show interest.

Hopefully, players who would like such things will press this button to remove anxieties over "Will people even care? Will I get flamed for wanting to make things interesting?"

### Caveats:

Wording may still need refinement Above screenshot is before some commitee work to improve wording. Present wording is in the files changed window.

### Testing:
- Confirmed player quitting the round by verb after dying removed them
- Confirmed player quitting the round by cryo removed them
- Confirmed player dying by digestion or violence did NOT remove them
- Confirmed list shape with 2 entries (1 guest acc, 1 host acc)
- Confirmed player choosing to stop showing enthusiasm is removed from the list
- Confirmed changing the area between calls changes the area var
- Confirmed inactivity check (setting the if comparison to 1 minute temporarily) works for auto removal
- Confirmed it does NOT show up for non-admin.

### Commit details:
https://github.com/VOREStation/VOREStation/pull/14959/commits/431dd7aaf67c46635991ae638576c70382103335
* New global associative list: key player ckey, value world.time
* New client verb under admin tab: "Show Event Enthusiasm"
* If usr.key in global associative list, let them remove themselves from it
* Else, let them add themselves, noting when they did so
* New admin verb under Fun tab:  "Assess player Interest"
* Iterates over player list, comparing ckeys against the new list
* Checks if mob/living or not, incrementing an alive counter
* Prints to chat each hit's name, ckey, time since opt in in minutes, inactivity in minutes and current area.
* If dead, skips area and denotes the possibility
* Removes entries who were inactive for 30 minutes
* Notifies entry player if they were removed
* Has a counter var to avoid iterating over all of player_list if we found our players faster than the whole list.
* cleanup() modified to also remove usrs leaving the round

### DIFF
--- a/code/_helpers/global_lists_vr.dm
+++ b/code/_helpers/global_lists_vr.dm
@@ -12,6 +12,7 @@ var/global/list/everyone_traits_negative = list()	// Neutral traits available to
 var/global/list/traits_costs = list()		// Just path = cost list, saves time in char setup
 var/global/list/all_traits = list()			// All of 'em at once (same instances)
 var/global/list/active_ghost_pods = list()
+var/global/list/event_consent_list = list() // Assoc list of player ckey as key time of entry as value. Used for giving GMs player enthusiasm info.
 
 //Global vars for making the overmap_renamer subsystem.
 //Collects all instances by reference of visitable overmap objects of /obj/effect/overmap/visitable like the debris field.
--- a/code/modules/admin/admin_verb_lists_vr.dm
+++ b/code/modules/admin/admin_verb_lists_vr.dm
@@ -170,7 +170,8 @@ var/list/admin_verbs_fun = list(
 	/client/proc/add_mob_for_narration,	//VOREStation Add
 	/client/proc/remove_mob_for_narration,	//VOREStation Add
 	/client/proc/narrate_mob,	//VOREStation Add
-	/client/proc/narrate_mob_args //VOREStation Add
+	/client/proc/narrate_mob_args, //VOREStation Add
+	/client/proc/assess_player_readiness  //VOREStation Add
 	)
 
 var/list/admin_verbs_spawn = list(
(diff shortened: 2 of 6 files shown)


Very-Soft:
While I appreciate the effort, I do worry that one of two things will happen with this tool:

either no one will ever use it, because events generally happen when someone on staff wants to do them over most other considerations

or

some people will use it, and it will empower players who dislike when things happen to shut down things happening, or complaining that they didn't get a say in whether or not things happened when someone doesn't use it.

I'm not saying don't do this, but more, I worry that it could change the landscape of when GMs are allowed to do events based on the loudest voices, which are almost always the negative ones.

Runa-Dacino:
> While I appreciate the effort, I do worry that one of two things will happen with this tool:
> 
> either no one will ever use it, because events generally happen when someone on staff wants to do them over most other considerations
> 
> or
> 
> some people will use it, and it will empower players who dislike when things happen to shut down things happening, or complaining that they didn't get a say in whether or not things happened when someone doesn't use it.
> 
> I'm not saying don't do this, but more, I worry that it could change the landscape of when GMs are allowed to do events based on the loudest voices, which are almost always the negative ones.

Based on this, how about this PR alternatively?
https://github.com/VOREStation/VOREStation/pull/14987

Runa-Dacino:
Seeing as the alternate implementation (all players who were active in the last 15 minutes get listed with detail) was accepted instead, i am closing this one.

PR #14975, opened 2023-06-05
Extends entity narrate functionality with TGUI

Runa-Dacino:
### What this does:
Adds TGUI functionality for the entity narrate (interface) verb
It's possible to call the TGUI window by first using the verb, then clicking the appropriate button. 
It defaults to non-TGUI mode always to avoid issues from TGUI crashing making it impossible to use.
This isn't a issue tho: the TGUI window stays open!

**Design:**
image

**Final version:**
image

**This allows GMs to:**

- Select One or Multiple entities (any /atom subtype (turf/obj, with special treatment for mobs)) from their pre-set reference list
- When selecting just One, obtain their name and type in a details window
- While selecting multiple, you just get the amount chosen
- You can click a selected entry to unselect it
- When swapping between multi/single, you clear the selection
- You can choose whether to emote or talk
- You can likewise choose to act loudly or subtly
- You may select all atom types at once without issues (altho for non-mobs, the lack of say verb can make "talking" problematic. Normally, for non-mobs, GMs are expected to add the say verb themselves as demonstrated in the video)

### Video showing off the features:

https://www.youtube.com/watch?v=7QM4Yk6-8E4

### Tweaks to base functionality & TGUI:

- Based on #Dev-General discussion with Cody, I am now allowing mobs with attached clients (ergo: players) to be narrated. However, adding and each narration is still logged in case of abuse or well-intentioned use causing hurt feelings.
- Added a protection vs overriding old entries by not using unique names
- Prevents issues from double encoding like apostrophes turning into gibberish while using the interface by prompting a new input.

### Commit Details
https://github.com/VOREStation/VOREStation/pull/14975/commits/fd7e89f0112297ebe3abf028e1665373ecb20b1f
* Removed special protection for mobs with clients
* * Instead, the GM is warned not to abuse this and each use will be logged
* * Added logging everywhere we previously used return or continue to prevent client narration
* * Kept !!!Player!!! typing rather than showing their actual path
* Added prevention of duplicate entity names.
* * Recursively calls the add entity proc until cancelled or a valid name is given
* Only sanitizes the message after it's passed to mob_args() for basic interface, and only for non-mob ones as mob one calls procs that sanitize the input themselves
* Likewise for TGUI, sanitization is only done within the actual narration proc to avoid double sanitization breaking things.
* While at it, I re-introduced [mode] to classic input's instruction string. I had it originally, then removed it while pruning code duplication. I realize I can add it back now.

https://github.com/VOREStation/VOREStation/pull/14975/commits/ae92270cc2c4e8109b02dd0fb6087d9df7287ee2
*Adds all the necessary procs for TGUI functionality as backend *Creates EntityNarrate.tsx
*Extends entity_narrate datum with helper vars for TGUI *TGUI window now allows:
* * Multi Selection
* * Single Selection
* * It shows details if single
* * It shows how many if multi
* * Allows picking between subtle/loud
* * Allows picking between talk/emote
* * Allows sending messages every 0.5 seconds
* * Uses new textbox feature to make editing and tracking narration easy!
* * Stays open.


Runa-Dacino:
A small bug was noticed with the normal interface due to double-sanitization. This will be avoided now; on top of the new TGUI feature.

(the bug was that HTML encoding caused special characters like apostrophes to derp themselves).

Runa-Dacino:
https://github.com/VOREStation/VOREStation/pull/14993

Will resolve merge conflict with the bundle once the above fix gets merged

Runa-Dacino:
Merge conflict resolved

Enzo-Leon:
Yoooo more GM tools yeesss

PR #14976, opened 2023-06-05
More 515 Compatibility

ItsSelis:


Heroman3003:
These changes won't have any effect on pre-515, will they?

ItsSelis:
> These changes won't have any effect on pre-515, will they?

I was able to compile and play it and there seemed to be no runtimes nor anything broken.
Basically, the PROC_REF() is defined in here:
https://github.com/VOREStation/VOREStation/blob/master/code/__byond_version_compat.dm

It decides on the version what kind of thing to apply.

PR #14978, opened 2023-06-06
Replaces security barrier sprite

MisterGrimm:
Proposed change to switch the sprite of the security barrier to something that looks a bit more 3D, to emphasize that it is indeed a barrier, and not something you can walk over.  
_I would have made gifs to compare the barriers, but it's harder then it looks with no tools._

Old Barrier
Captureold

New Barrier
Capturenew

- Replace old security barrier with more 3D barrier
- Places new barrier in objects_vr.dmi , removed old objects.dmi security barrier states. _I can change this if this isn't a good idea._


Youtubeboy139:
I'd keep the old sprites in the 'object.dmi' as opposed to flat out deleting them for the time being, that's just me tho

Enzo-Leon:
> I'd keep the old sprites in the 'object.dmi' as opposed to flat out deleting them for the time being, that's just me tho

I'd Add on this. This may clash or look weird for some of the other maps/areas that use the standard wall sprite and textures. These would still be great to have though yes.

Heroman3003:
I'll agree that ideally, it should be more a subtype with purely visual difference than outright replacement.

MisterGrimm:
> I'll agree that ideally, it should be more a subtype with purely visual difference than outright replacement.

While I personally think it works fine as a replacement, but I can make it a subtype instead. Although, I personally doubt it will be used at all if there are two versions that just look cosmetically different.

AzzyDreemurr13:
> New Barrier
> Capturenew

As for me, it would be more interesting if the barrier was lowered in the off mode. When it is activated with an ID card, it rises up

Heroman3003:
If by end of next week, there's no further response or change, I'll probably close this.

For barrier itself, I'd suggest maybe replacing ones security starts with on SD specifically with this subtype, as that map does use walls and environment that would fit this style of object.

PR #14979, opened 2023-06-06
MC & Subsystem Update

ItsSelis:
The latest version of the MC, Failsafe and Subsystems system.

Currently still a draft since there are a few atoms erroring in the console.
Otherwise everything seems to run just fine.


ItsSelis:
Due to the complexity of the entire system it is a better idea to atomize this into smaller pieces to review.

PR #14987, opened 2023-06-07
Alternate to https://github.com/VOREStation/VOREStation/pull/14959 New Admin verb: Report Player Status

Runa-Dacino:
### What this is:
Based on feedback on https://github.com/VOREStation/VOREStation/pull/14959, 
I asked Tank for what could be salvaged, what would still be useful.

This is what we arrived at:

image

A to_chat() proc that takes which areas players are in, sorts them by said areas and returns their coordinates, activity and name.

### Commit Details
https://github.com/VOREStation/VOREStation/commit/a9cbd8fa4fea0876f6cf56662f97c70a5780c423
* Adds a new verb in the "EventKit" tab
* New verb copies player_list and sorts it into an assoc list of area:(players)
* It then prints out each area's list of players, giving their (x,y) coordinates
* It also notifies of how long since their last action, rounded to nearest minute
* Players who took ABSOLUTELY NO ACTION for 15 minutes are pruned to reduce clutter
* Tracks pruned players, players who are actually in-game as mob/living
* Reports this as well.

### DIFF
--- a/code/modules/admin/admin_verb_lists_vr.dm
+++ b/code/modules/admin/admin_verb_lists_vr.dm
@@ -170,7 +170,9 @@ var/list/admin_verbs_fun = list(
 	/client/proc/add_mob_for_narration,	//VOREStation Add
 	/client/proc/remove_mob_for_narration,	//VOREStation Add
 	/client/proc/narrate_mob,	//VOREStation Add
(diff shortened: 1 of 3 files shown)


Enzo-Leon:
YES. Make use of this option for us to better gauge and prep stuff.

PR #14989, opened 2023-06-07
Implements Climbing

Runa-Dacino:
### What this does
  Creates a brand new mechanic for traversal: Climbing!

  Climbing allows a mob/living to walk up to a special wall (marked by examine) and climb on top of it!
  However, not all can do so:
  Most will fail 80% of the time, unless they use specialized gear or are naturally inclined/trained for climbing.

  Agile people (traceur, tajara) have a lower chance to fall (altho tajara can climb by default and faster)

**As such, 3 new traits are added,**
  Climber, Amateur, costs 1 point for removing fall chance (all species can take)
  Climber, professional that climbs faster, costing 2 pts
  Climber, Natural for 0 points (xenochim and custom only)

  Simple mobs and silicons can try climbing as well. Silicons may later (another PR) receive a climbing module maybe to also let them climb.

  Simple mobs so far only extend to scugs who get a special mechanic! Scugs may reduce their slow climb to a lightning fast one if they drop the spear in their hand and leave it behind.

  Vassilians and Taj are master climbers

### You can climb down too now as of commit https://github.com/VOREStation/VOREStation/pull/14989/commits/11e325cd6f0143401125f0af3c8e9fae81500024

  Same logic as earlier, but we allow trying on unclimbable walls (since you could just walk off the edge anyway).
  In this case, if unclimbable we set a minimum 75% fall chance even if master climber
  You are also warned of trying to descend 2Z at once and prevented from doing so.

### Climbing accounts for nutrition as of https://github.com/VOREStation/VOREStation/pull/14989/commits/d1f8bdf8514d1db8e25ca2de80f73040dcf3fb94

        Climbing up costs 50, climbing down costs 25 nutrition
        You cannot climb if you got less than it'd cost
        Climbing with less than 100 nutrition introduces a minimum 30% fall chance
        Climbing with less than 200 nutrition introduces a 1 second delay.

### Climbing speeds
	 Climb time is 3.75 for scugs with spears (spear is dropped)
	 Climb time is 5 for Vassilians, Taj and best-gear
	 Climb time is 9 for Professional Climbers
	 Climb time is 17.5 Seconds for amateur climbers
	 Climb time is 20 seconds for scugs without a spear
	 Climb time is 30 for gearless untrained people
         Speed increased by 1 second if under 200 Nutrition for all

### Commit Details
https://github.com/VOREStation/VOREStation/pull/14989/commits/d1f8bdf8514d1db8e25ca2de80f73040dcf3fb94

- Traits reorganized
  - No more 0 pt trait for non-custom species, UNLESS: you are a xenochimera
  - Non-custom, non-xenochim must spend 1 point to be able to climb (climber, amateur) in vein of soft fall
  - Custom/Xenochim can take (climber, natural) for 0 pts in vein of winged flight
  - All species may take (climber, master) for 2 pts to halve their climbing speed
- Adjusted Tajara to be as fast as Vassilians to.
- Implemented Nutrition logic in inspiration of winged flight
  - Climbing up costs twice as much nutrition as climbing down (50 vs 25)
  - Climbing while hungry (less than 200 nutrition) introduces a delay of 1 second
  - Climbing while starving (less than 100 nutrition) introduces 30% fall chance (does not override if higher)
  - Cannot climb if nutrition is lower than 50 for up, 25 if down.

https://github.com/VOREStation/VOREStation/pull/14989/commits/11e325cd6f0143401125f0af3c8e9fae81500024
* Same logic as climbing up for fall chance
* Additional fall chance if trying to climb down a unclimbable wall: min 75%
* Works by checking the turf 1 step in front of the mob if empty,
and places them under that turf if it is possible
* Climbing times are same as climbing up
* Grace period somewhat shorter
* Also added tweaks to logic of climbing up to avoid pointless checks
* Also tweaked the traits (expanded tutorial for climbing down, proper trait exclusion)

https://github.com/VOREStation/VOREStation/pull/14989/commits/83f7e90b8e807c0e507a3b26b1c517afa67fa5a2
* Adds new turf/simulated proc: climb_wall
  * Anyone can attempt this by standing next to wall
  * Untrained have chance to fall
  * Trained dont fall unless interrupted
  * Speed varies wildly depending on gear & skills
    * Takes lowest climb_delay, multiplies 5 by it
    * Depending on tresholds, may add 10/5 seconds on top (>1.25, >1.0)
  * Gear can enable even rookies to climb. Simplemobs cant use gear
  * Except scugs. They can specifically use their spears to climb, leaving it behind
* Adds new turf/simulated var: climbable
  * solidrock, /mineral/cave have climbable set to true
* Adds new turf/simulated logic for init, examine() and a yet-unused proc
  * proc toggle_climbability allows turning walls climbable. Useful for GMs!
  * Could also later implement a tool to turn walls climbable when used using it.
* Adds new mob/living vars, can_climb and climbing_delay
  * These are for handling silicons and simple mobs
* Adds new species vars, can_climb and climbing_delay
  * These are for human mobs
* climbing_delay defaults to 1.5
  * Tajara have it at 1.25
  * Vassalian have it at 1
  * Scugs got it at 2.0, reduced to 0.75 when using spear
* New traits: climber; climber, professional; climber, master
  * Cost 0, 1, 2 pts as positive trait respectively
  * Enable safe climbing, reduce delay to 1.25, 1.0 respectively
* moves rock_climber from shoes to items define
* Adds new item var, climbing_delay set at 1

### DIFF
--- a/code/game/objects/items.dm
+++ b/code/game/objects/items.dm
@@ -112,6 +112,9 @@
 
 	var/protean_drop_whitelist = FALSE
 
+	var/rock_climbing = FALSE //If true, allows climbing cliffs using click drag for single Z, walls if multiZ
+	var/climbing_delay = 1 //If rock_climbing, lower better.
+
 /obj/item/New()
 	..()
 	if(embed_chance < 0)
--- a/code/game/turfs/simulated.dm
+++ b/code/game/turfs/simulated.dm
@@ -17,6 +17,7 @@
 	var/dirty_prob = 2	// Chance of being dirty roundstart
 	var/dirt = 0
 	var/special_temperature //Used for turf HE-Pipe interaction
+	var/climbable = FALSE //Adds proc to wall if set to TRUE on its initialization, defined here since not all walls are subtypes of wall
 
 	var/icon_edge = 'icons/turf/outdoors_edge.dmi'	//VOREStation Addition - Allows for alternative edge icon files
 
@@ -63,6 +64,14 @@
 	if(istype(loc, /area/chapel))
 		holy = 1
 	levelupdate()
+	if(climbable)
+		verbs += /turf/simulated/proc/climb_wall
+
+/turf/simulated/examine(mob/user)
+	. = ..()
+	if(climbable)
+		. += "This [src] looks climbable."
+
 
 /turf/simulated/proc/AddTracks(var/typepath,var/bloodDNA,var/comingdir,var/goingdir,var/bloodcolor="#A10808")
 	var/obj/effect/decal/cleanable/blood/tracks/tracks = locate(typepath) in src
--- a/code/game/turfs/simulated/outdoors/atmoscaves_vr.dm
+++ b/code/game/turfs/simulated/outdoors/atmoscaves_vr.dm
@@ -2,11 +2,13 @@
 	oxygen = MOLES_O2STANDARD
 	nitrogen = MOLES_N2STANDARD
 	temperature	= T20C
+	climbable = TRUE
 
 /turf/simulated/mineral/ignore_mapgen/cave
 	oxygen = MOLES_O2STANDARD
 	nitrogen = MOLES_N2STANDARD
 	temperature	= T20C
+	climbable = TRUE
 
 /turf/simulated/mineral/floor/cave
 	oxygen = MOLES_O2STANDARD
--- a/code/game/turfs/simulated/wall_types_vr.dm
+++ b/code/game/turfs/simulated/wall_types_vr.dm
@@ -206,6 +206,7 @@ var/list/flesh_overlay_cache = list()
 /turf/simulated/wall/solidrock
 	icon_state = "solidrock"
 	icon = 'icons/turf/wall_masks_vr.dmi'
+	climbable = TRUE
 
 /turf/simulated/wall/titanium
 	icon_state = "titanium"
--- a/code/game/turfs/simulated_vr.dm
+++ b/code/game/turfs/simulated_vr.dm
@@ -1,5 +1,15 @@
 /turf/simulated
 	can_start_dirty = FALSE	// We have enough premapped dirt where needed
 
+
+
 /turf/simulated/floor/plating
 	can_start_dirty = TRUE	// But let maints and decrepit areas have some randomness
+
+
+/turf/simulated/proc/toggle_climbability() //Again, b
+	if(climbable)
+		verbs -= /turf/simulated/proc/climb_wall
+	else
+		verbs += /turf/simulated/proc/climb_wall
+	climbable = !climbable
--- a/code/modules/clothing/clothing.dm
+++ b/code/modules/clothing/clothing.dm
@@ -604,7 +604,6 @@
 
 	var/water_speed = 0		//Speed boost/decrease in water, lower/negative values mean more speed
 	var/snow_speed = 0		//Speed boost/decrease on snow, lower/negative values mean more speed
-	var/rock_climbing = FALSE // If true, allows climbing cliffs with clickdrag.
 
 	var/step_volume_mod = 1	//How quiet or loud footsteps in this shoe are
 
--- a/code/modules/clothing/shoes/boots.dm
+++ b/code/modules/clothing/shoes/boots.dm
@@ -190,6 +190,7 @@
 	desc = "A pair of winter boots, with metal bracing attached to assist in climbing rocky terrain."
 	icon_state = "climbing_boots"
 	rock_climbing = TRUE
+	//Climbing delay with boots is 1
 
 /obj/item/clothing/shoes/boots/tactical
 	name = "tactical boots"
--- a/code/modules/mob/living/carbon/human/species/species_vr.dm
+++ b/code/modules/mob/living/carbon/human/species/species_vr.dm
@@ -38,6 +38,8 @@
 	var/micro_size_mod = 0		// How different is our size for interactions that involve us being small?
 	var/macro_size_mod = 0		// How different is our size for interactions that involve us being big?
 	var/digestion_nutrition_modifier = 1
+	var/can_climb = FALSE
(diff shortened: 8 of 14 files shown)


Runa-Dacino:
Added ability to climb down walls too

Enzo-Leon:
Interesting concept, could be of use for adding routes and other ways to get around maps and event locations.

Heroman3003:
I think that having 3 traits is a bit excessive in amount. I think 2, one to remove fail chance and one to speed up would be fine. For Vasilissans and Tajara, can just have custom non-trait-replicable numbers, keeping them a bit more unique in that regard as well. But that's just my thoughts.

Runa-Dacino:
> I think that having 3 traits is a bit excessive in amount. I think 2, one to remove fail chance and one to speed up would be fine. For Vasilissans and Tajara, can just have custom non-trait-replicable numbers, keeping them a bit more unique in that regard as well. But that's just my thoughts.
> 
> Also I think that basic climber should cost at least 1 point.

Basic climber was decided to be given 0 points because winged flight, which is faster and has no restrictions like this one (anywhere you can climb, you can fly and even places you cannot climb (like steel walls) you can still fly)

Now winged flight does cost nutrition but it is fairly insignificang for this particular usecase so i think making it cost 0 is fine.

I can add a nutrition cost though since this is a fairly strenous activity after all.

Heroman3003:
Winged flight is also custom_only trait. More comparable one here is Soft Landing in my opinion, which does cost 1 point and serves similar purpose.

Runa-Dacino:
> Winged flight is also custom_only trait. More comparable one here is Soft Landing in my opinion, which does cost 1 point and serves similar purpose.

Soft fall works universally without any chance of failure. If you get bumped and fall off it automatically fires. If you are standing on a 2z high steel wall you can safely descend no matter what instantly.

Climbing at 0 pts takes 17 seconds and moving after 4s (bumped for instance) will make you fall and get hurt unless you got soft fall on top.

Given nutrition cost too, it would also be limited in how often you can use it/attempt it.

My thinking is...

Make it cost half as much nutrition as winged flight ascend/descend (subject to change after i actually see those values), and give it a minimum requirement of 3x that cost.

Climbing while not having at least the minimum amount when starting inflicts a minimum 25% fall chance due to exertion. While easy to avoid, it would allow for mistakes to happen and have a cost.

Furthermore, it only works on rocky cliffs. While you can descend all walls, you got a 75% of failure unless it is rocky itself.

Heroman3003:
Basically, I do not like idea of any sort of completely free non-custom trait with non-scene/fluff benefits. Period. It basically becomes something everyone can and will have. I am not against 0-point positive traits for custom species or if trait doesn't have proper mechanical benefit, but thats neither nor.

Runa-Dacino:
> Basically, I do not like idea of any sort of completely free non-custom trait with non-scene/fluff benefits. Period. It basically becomes something everyone can and will have. I am not against 0-point positive traits for custom species or if trait doesn't have proper mechanical benefit, but thats neither nor.

While making the traits initially, I'd been told the cost was too high (originally it was just standard for 1 pts, professional for 2 pts).

I'll wait with making changes until a consensus is had (seeing Rascal's Pass is 2 weeks away - there's time to refine!)

TankTheBirb:
I’d recommend just making a 1 point version for any species that is identical to the 0 point version for custom species. That way, people can grab the climbing trait to replicate something like mountain climbing experience, without it being basically free for every character that isn’t custom species.

SatinIsle:
I agree with Tank's suggestion. It would be ideal of there were 0, 1 and 2 cost versions of the custom species version. And a single 1 cost basic all species version that is equivalent to the 0 cost. It's a bit of clutter though, but it is more in line with other costs. 

An alternative would be just removing the middle cost one (keeping it as a taj thing only), and having the other two be 1 and 2 cost. Though that does make it inferior to other movement traits for custom species.

Heroman3003:
> I agree with Tank's suggestion. It would be ideal of there were 0, 1 and 2 cost versions of the custom species version. And a single 1 cost basic all species version that is equivalent to the 0 cost. It's a bit of clutter though, but it is more in line with other costs.
> 
> An alternative would be just removing the middle cost one (keeping it as a taj thing only), and having the other two be 1 and 2 cost. Though that does make it inferior to other movement traits for custom species. It'd be less cluttered but a bit less fun too imo.

Well, arguably, Soft Fall is already inferior to winged flight, hence why that was my suggestion as well. Precedent exists as is.

TankTheBirb:
> Well, arguably, Soft Fall is already inferior to winged flight, hence why that was my suggestion as well. Precedent exists as is.
Yes and no. Winged flight needs a verb activation and must be used proactively. Soft fall is purely passive and while you can't go up, you are always safe going down. They have different niches I feel.

Runa-Dacino:
@TankTheBirb @Heroman3003 
Made changes based on discussions in DMs and replies here


https://github.com/VOREStation/VOREStation/pull/14989/commits/d1f8bdf8514d1db8e25ca2de80f73040dcf3fb94

- Traits reorganized
  - No more 0 pt trait for non-custom species, UNLESS: you are a xenochimera
  - Non-custom, non-xenochim must spend 1 point to be able to climb (climber, amateur) in vein of soft fall
  - Custom/Xenochim can take (climber, natural) for 0 pts in vein of winged flight
  - All species may take (climber, master) for 2 pts to halve their climbing speed
- Adjusted Tajara to be as fast as Vassilians to.
- Implemented Nutrition logic in inspiration of winged flight
  - Climbing up costs twice as much nutrition as climbing down (50 vs 25)
  - Climbing while hungry (less than 200 nutrition) introduces a delay of 1 second
  - Climbing while starving (less than 100 nutrition) introduces 30% fall chance (does not override if higher)
  - Cannot climb if nutrition is lower than 50 for up, 25 if down.

Heroman3003:
Seems a lot better.

PR #14990, opened 2023-06-08
The Falls minor improvements

Burrito-Justice:
Some small improvements have been made to the Falls.
- The invisible barriers now have heavy foliage around them so it makes a little more sense why you aren't allowed to go to the edge of the map.
- One of the corner sprites for the overhangs has been fixed.
- Some small grammar and readability improvements.
- The ambience that plays has been quieted down after a couple complaints about it being loud.


Burrito-Justice:
pro tip: don't leave DM open while pulling changes or else new files won't be ticked and you'll be staring at the compiler wondering why seemingly unrelated stuff is erroring like a dummy!!!

PR #14994, opened 2023-06-08
Adds per-belly option to vorepanel for switching between sleeper/belly

Runa-Dacino:
### What this does
This tweak enables silicon players to use the pre-existing vorebelly overlays their dogborg-type (incl drake and raptor) modules get from their "sleepers" (including analyzers, brigs, janiborg thingy and so forth) to use them during normal vore.

This is implemented as a per-belly toggle in the vorepanel (shows for all players for now. Figuring out conditional toggles on vorepanel one day would be neat, but non-critical. It's clearly labeled as "Toggle Silicon belly" and clicking it explains that it only applies to sleeper-possessing entities.

Toggling this between sleeper and vorebelly causes the robot to enable their overlay exclusively using that mode of eating.
~~Using vorebelly option, the belly is triggered at contents >= 1 flat~~ (see As of commit https://github.com/VOREStation/VOREStation/pull/14994/commits/5cb3dc43b6de4538c0f5e97748cddb6b0f714c42)

Upon being triggered, this applies the module's sleeper_r overlay. The sleeper_r overlay seems to be applied in nearly all cases already except for medihound, where there's a legitimate distinction warranting sleeper_r and sleeper_g distinction (I presume: they were the original and the rest just followed convention). As such, I went ahead and used sleeper_r rather than delay the PR by trying to meticiously comb through every dogborg subtype's .dmi for sleeper_r and make a sleeper_o that's amber coloured. Anyone wanting to toy with icons is welcome to do so, or I may do it while bored one day.

Due to how vore works, perform_the_nom is called in all cases. Therefore, I decided to rely only on perform_the_nom to updateicon(). This seems to work using drop and throw vore at least, and so I presume this extends to all other (testing rest on local is pain)

**As of commit** https://github.com/VOREStation/VOREStation/pull/14994/commits/5cb3dc43b6de4538c0f5e97748cddb6b0f714c42 , functionality has been significantly expanded. The following image summarizes the changes:

image

1. The "Dogborg Controls" section is ONLY VISIBLE IF and ONLY IF the belly's owner isrobot() = true and has dogborg = true. It is hidden from all other mob types. Framework is put in place for voremobs (simple_mob/vore), but this is not expanded on yet.
2. User can set minimum amount of ANYTHING their stomach must contain (trash, objects, stripped clothing included) before an overlay may show
3. User may extend this logic by setting min prey size to a value between 25-200 (set to 0, the following is irrelevant)
4. By setting this to a valid number, the code checks if they also enabled the override - which if they did, we check if their belly contents' quantity (any) surpasses the override requirement. If yes - we don't check prey size and set overlay to true.
(e.g.: With this override, a min req of 100% to show is ignored if we eat 10 25% micros, and we set this override to 10.)
5. If they didn't enable the override, or belly contents is less than their personally set treshold, we iterate over all var/content in their belly.contents, type checking for mob/living and then if it's mob/living, we check their size. If their size is over our treshold, we stop iterating and set the belly overlay to be true.

### Why we need this

This has been an oft-repeated request playerside, and seeing as dogborgs already have vorebellies; all this does is expand their use to allow players to use their bellies with custom descriptions and mechanics rather than enforce sleeper use.

### Future considerations

This logic could likely be extended to simplemobs, and the aforementioned sleeper_o. However, I'd rather not delay this PR further than necessary to be "perfect"

### Commit Details
https://github.com/VOREStation/VOREStation/pull/14994/commits/5cb3dc43b6de4538c0f5e97748cddb6b0f714c42
* Adds updateicon() call to belly transfers
* Rebundles tgui.bundle.js
* Adds four new bellyobj vars:
  * visible_belly_minimum_prey
  * overlay_min_prey_size
  * override_min_prey_size
  * override_min_prey_num
* Adds extended overlay logic to robot updateicon() for dogborgs
  * If belly contents lower than minimum_prey, dont show anything
  * Otherwise, if min_prey_size 0, show regardless
  * Otherwise, only show if prey size_multiplier greater than min size
  * However, if override_min_prey_size is true, check LAZYLEN(belly.content)
  * If this list's length is greater than override_min_prey_num, show belly overlay regardless of prey size
* Extends vorepanel_vr, belly_obj_vr and vorepanel.js with the necessary code to enable the above

https://github.com/VOREStation/VOREStation/commit/f2efe267ac3789df14acf77ab9b70f5840c3a5e8
* New option per vorebelly that silicons may use to choose between overlay type
* If picking sleeper, it works as it currently did.
* If picking vorebelly, it gives them sleeper/analyzer/etc belly overlay if their belly's contents are >= 1
* Consequently adds a new belly_obj var to track this
* Adds user.updateicon() call to perform_the_nom() to trigger belly overlay update
* Edits robot_vr's updateicon() to contain logic for sleeper vs vorebelly pref
* For now uses sleeper_r overlay since it exists for ALL icon_states
* Ideally either me or someone I enslave creates sleeper_o overlay using orange colours where relevant. However, rather not delay implementation for this


Enzo-Leon:
Good needed feature.

Heroman3003:
Conflict.

Runa-Dacino:
Merge conflict resolved. Additional functionality added (no longer shows up if not a dogborg, allows fine tuning when you get belly showing IF dogborg)

PR #15003, opened 2023-06-10
Revised cobra hood accessory

CubeJackal:
Spruces up the old cobra hood I added last century. New sprite, dual-colour, and it now lives in ear selection instead of markings and hairstyles. Snakes, rejoice.


SatinIsle:
I feel like this would be better as an addition rather than a replacement of something that people have already built into their sprites.

Heroman3003:
I concur. Ideally, both hoods should remain, call new one something like cobra hood, alt for distinction.

Heroman3003:
I concur. Ideally, both hoods should remain, call new one something like cobra hood, alt for distinction.

CubeJackal:
I'd _like_ it to be a marking, too, but those don't appear to support colour overlays, and I'm loathe to throw more awkward bloat into the markings list by adding the parts as separate markings. Plus, I'm not incredibly keen on leaving what I feel is an outright worse version of the same sprite in the game just for legacy support. I _can_, but only if you feel that it's necessary.

Heroman3003:
While I personally do agree the new one is much better, it also has somewhat distinctly different shape, hence why I believe the old one should stay in some way for people who may have used it not for cobra hood aspect but just for visual appearance.

PR #15016, opened 2023-06-13
tweak (fluff): Removes KHI from character set-up options

Runa-Dacino:
Loremaster discussion concluded that the KHI faction is out-of-scope of our current intended theme and setting. While we do not remove them outright, encouraging new players and characters to originate from KHI homeworld or be part of their faction makes things needlessly complicated and troublesome.

Therefore, KHI remains as something for the ATC system to be referenced and as part of technologies, but player characters ideally do not hold personal relation to them at this time.


KillianKirilenko:
Should the uniforms be renamed as well, then?

Heroman3003:
Uniforms and items will remain. Only the background option is being removed.

PR #15017, opened 2023-06-13
Loadout Overhaul

KillianKirilenko:
This PR seeks to counter some of the *aggressive* bloat the loadout menus (especially the suit and uniform menus) are experiencing by making more use of selector type menus where possible. Full changes follow;
- Casual dress is now under its own "casual dress" category
- The Uniform Selectors tab has been renamed to just uniforms
- Selectors for all uniforms have been reduced to 1pt cost for consistency with other uniform costs (for some reason I had it in my head that certain uniforms cost slightly more because they had integrated armour or the like)
- Uniforms that were previously listed seperately are now *only* available through the relevant selector, but none have been (intentionally) removed
- Several items have been compressed down using selection menus where there are multiple (3+) similar items of a very similar type, for instance basic coloured gloves, softcaps, cowboy hats, leather jackets, and so on
- Departmental hat selectors have been added for Security and Medical since they had enough to warrant them
- A new EMT softcap has been added to mirror the basic Medical softcap
- Loadout no longer inexplicably excludes a particular brand of cigarettes
- In a similar vein, two brands are now no longer premium (that is coin-only) items in cigarette machines

If your favourite uniform has disappeared into the aether, please check the uniform selectors before complaining!


Enzo-Leon:
Cleanup of bloat is always good, and better loadout handling. Will we need to send a notice to players to ensure their loadouts aren't affects from the adjustments?

KillianKirilenko:
You won't "need" to, but I think it would probably be a good idea to warn people, yes.

Enzo-Leon:
Better safe than sorry approach, usually the best bet to go with such matters.

TankTheBirb:
good change, this should help decrease some of the time spent scrolling in them menus for the one thing that you want but don't know where it is. Especially since the categories seem logical to me.

KillianKirilenko:
> good change, this should help decrease some of the time spent scrolling in them menus for the one thing that you want but don't know where it is. Especially since the categories seem logical to me.

One could make a case for it *hiding* items, but I feel that's a small price to pay for greater convenience.

I would love to make a more flexible character designer that isn't subject to the limitations of HTML (such as resetting to the top of the tab every time you make any change whatsoever...) but that is far beyond my skill level, alas. Modest adjustments like this shall have to suffice.

Heroman3003:
Time to break loadout for everyone and their mother for the sake of future simplicity.

Heroman3003 (approved):

PR #15031, opened 2023-06-16
Update Pref Options

KillianKirilenko:
Swaps the positional options to role options. Adds -pref subtypes as well to match the primary options.


KillianKirilenko:
There's already a sanitize function, so worst-case it should just reset people to Unset/No ERP.

TankTheBirb:
Having more specificity in existing categories isn't a bad idea. Although I personally would prefer using dominant/submissive, doesn't this change the option from preferred sexual positions to power dynamic positions?

KillianKirilenko:
> Having more specificity in existing categories isn't a bad idea. Although I personally would prefer using dominant/submissive, doesn't this change the option from preferred partner positions to power dynamic positions?

Yes? That's what the first sentence of the PR says?

I could've added a new column but last I heard staff doesn't want it having too much info "up front".

Heroman3003:
Note to self: include this change in announcement when talking about Loadout Overhaul too

Heroman3003 (approved):

PR #15032, opened 2023-06-16
Adds the ability to move through people who are intentionally pixel shifted

silicons:
Check threshold is 8 to either x or y.
Only other mobs can pass, no wavedashing around projectiles or anything.

### DIFF
--- a/code/modules/mob/living/living_movement.dm
+++ b/code/modules/mob/living/living_movement.dm
@@ -3,6 +3,9 @@
 		var/mob/moving_mob = mover
 		if ((other_mobs && moving_mob.other_mobs))
 			return TRUE
+		if(is_shifted && (abs(pixel_x) >= 8 || abs(pixel_y) >= 8))
+			// they're wallflowering, let 'em through
+			return TRUE
 	if(istype(mover, /obj/item/projectile))
 		var/obj/item/projectile/P = mover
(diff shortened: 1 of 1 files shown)


Heroman3003:
A bit of concern on how it would work with mobs that already use the offset naturally.

TankTheBirb:
> A bit of concern on how it would work with mobs that already use the offset naturally.

The interactions with existing shifted mobs are as follows.
- If you are shifted and they are not, they will run through your space if intending to pass you instead of displacing you. If you are not shifted, the interaction is normal with mobs that are not pre-shifted.
- Being shifted clears on moving, so moving while shifted into a mob's space that is not shifted does nothing.
- Turning on "is_shifted" while they are buckled to a chair allows you to move through them as if they were shifted (because they are), however, there is no way to organically do this so as long as is_shifted is not hard set as always on for something, this shouldn't be a problem.

I don't forsee any real issues as while spawning a xenomorph for testing, I was shifted and forgot to set my faction, so they attacked me as normal and I had no issue being killed despite being shifted and them not being in my tile. So I'd say it's functional and good to go.

Heroman3003:
I mostly meant that in regard to player-controlled mobs, but that does clear up most of my concern on that front regardless.

Heroman3003 (approved):

PR #15033, opened 2023-06-16
Stopgap fix for cyborg surgegical saw missing icon

Heroman3003:
Just a stopgap. Will keep it open until the next release, and close if proper icon is provided before then.


Youtubeboy139:
https://github.com/VOREStation/VOREStation/pull/15034

Issue #14825, opened 2023-04-28
[GENERAL] Protean Blob Speed Issue Recurred


ResidentCody:
#### Brief description of the bug
Protean blobs have somehow reverted to being slower than their carbonmob selves. This is a prior issue that has happened again somehow.
https://github.com/VOREStation/VOREStation/issues/12060

#### What you expected to happen
For the words that say that protean blobs more slightly faster to not be a lie.

#### What actually happened
The words are a lie. You in fact move noticeably slower.

#### Steps to reproduce
- Be a protean.
- Run in a straight line to see how fast you move (Optionally record your screen and count how many frames it takes to run the distance).
- Blob, see the info text that says you are now faster.
- Blob the same straight line and see that you move slower. (Optionally, have someone who's not a blob run parallel and see how they move faster).

#### Code Revision
Server revision: B:release D:
Commit: [db5723a7386f8b0d011d5e9f3eef689335a9b76d](http://127.0.0.1:62406/tmp18496/%3Ca%20href=%22https://github.com/VOREStation/VOREStation/commit/db5723a7386f8b0d011d5e9f3eef689335a9b76d%22%3Ehttps://github.com/VOREStation/VOREStation/commit/db5723a7386f8b0d011d5e9f3eef689335a9b76d%3C/a%3E)
#### Anything else you may wish to add:
This is not a case of the Haste trait or any equipment slowdown effects being involved.
The fix is to either increase protean blob movespeed to match the text, or change the text to say that you move slower.

Runa-Dacino:
This was likely caused by the recent PR that changed mob slowdowns, and likely proten blobs weren't adapted to the new speed rules.

PR #15036, opened 2023-06-17
Player blobform speed adjustments

BonniePandora:
Makes the prommie and protean blobforms no longer move _slower_ than the base humanoid speed.
Protean blob moves fractionally quicker.
Addresses #14825
Images of the new values in testing-
Protean:
image
Prommie:
image


TankTheBirb:
iirc this was the intended behavior before the speed change awhile back. Good catch.

PR #15037, opened 2023-06-17
Adds ability for multi-colored colorable belly overlays to exist, adds one using it

Heroman3003:
Art for the new belly is by VerySoft

The idea is rather simple, adding a new colorable belly is done same way as before, but overlays for it are added to new file, screen_full_colorized_vore_overlays.dmi. The name of icon states are identicial to your 'base' belly overlay, but with _l1, _l2 or _nc added for secondary, trinary or no color option respectively. For example, newly added belly internal uses only _nc overlay, meaning its basically single-color belly but with white shine overlay, making it not lose all its sheen and luster in the colorization process.

In theory, using multiple colors is absolutely possible in the future too.

UI looks a tad jank, but I'm gonna say it 'works enough' and let someone else figure it out later down the line (probably myself)

Simple preview (note that white parts remain white regardless of color choices):
изображение


Heroman3003:
A.

PR #15043, opened 2023-06-18
Spell Fixes

ItsSelis:
It's basically #14776 but only including the player-visible fixes.
I did this because when I glanced at the PR I saw some things that also get printed to the chat or are in the descriptions.

(Yes, I looked through every single file)


code/modules/catalogue/catalogue_data.dm
```diff
@@ -139,7 +139,7 @@ GLOBAL_DATUM_INIT(catalogue_data, /datum/category_collection/catalogue, new)
 	<br><br>\
 	Humanity is the primary driving force for rapid space expansion, owing to their strong, expansionist central \
 	government and opportunistic Trans-Stellar Corporations. The prejudices of the 21st century have mostly \
-	given way to bitter divides on the most important issue of the times� technological expansionism, \
+	given way to bitter divides on the most important issue of the times� technological expansionism, \
```
Heroman3003, line 142: this looks kinda weird. There's more of these in the same file too.

PR #15088, opened 2023-06-27
Four new pAI chassis

scarletflared:
Ah, there's the button for pull requests. If everything works like I think it does, this adds four (three and a half?) new pAI chassis for the five people that actually play pAIs.
paipreview


Very-Soft:
As one of the 5 pAI players, HYPE HYPE HYPE

code/modules/mob/living/silicon/pai/pai.dm
```diff
@@ -36,6 +36,10 @@
 		"Parrot" = "pai-parrot",
 		"Rabbit" = "pai-rabbit",
 		//VOREStation Addition Start
+		"Dire wolf" = "pai-diredog"
+		"Horse (Lune)" = "pai-horse_lune"
```
Very-Soft, line 40: Ah, you need to ass commas at the end of list entries like the others

code/modules/mob/living/silicon/pai/pai_vr.dm
```diff
@@ -48,6 +48,10 @@
 		"typeone",
 		"13",
 		"pai-raptor"
+		"pai-diredog"
```
Very-Soft, line 51: Here too

Very-Soft (changes requested):

code/modules/mob/living/silicon/pai/pai.dm
scarletflared: Ah good catch, I had the feeling I was missing something really obvious like that.

SatinIsle:
These looks excellent

scarletflared:
With any luck, the commas were the only thing I was missing.

PR #15098, opened 2023-07-01
Randomness updates

Very-Soft:
Adds a 'family' var to fake_sun, allowing them to inherit the settings of other fake_suns with the same 'family'

Also adjusts the mob spawning mechanics for areas, moved those mechanics to area, rather than area/awaymission, that way it can be set for any area, should one desire. Additionally adds semirandom mob spawning functionality to the area spawning mechanics, allowing one to feed the system a list of lists of mob types, and having it only pick as many groups of those as is desired.

### DIFF
--- a/code/game/area/Away Mission areas.dm
+++ b/code/game/area/Away Mission areas.dm
@@ -1,4 +1,4 @@
-/area/awaymission
+/area
 	name = "\improper Unknown Location"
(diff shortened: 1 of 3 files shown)


Runa-Dacino:
more dynamic POIs go woo!

I like

PR #15145, opened 2023-07-15
Adjust bow!

Very-Soft:
Now if someone picks up dessa's bow it will use the normal sprite instead of dessa's special sprite


Screemonster:
ohyah avida's hat does a similar thing for people who don't have bnnuy ears

PR #15152, opened 2023-07-16
Autowhisper

Very-Soft:
Adds Autowhisper!

Autowhisper is a pair of verbs and UI buttons!

image
image

Autowhisper, the button on the left is a simple toggle, when enabled, all say/me use will be replaced with whisper/subtle!

Autowhisper Mode, the button on the left allows one to configure the settings that subtle will use, using the new custom subtle modes. It also allows one to automatically use psay/pme if conditions are right. The setting must be set each round, but it will remember your settings for the round, and you can enable and disable that setting with a simple push of the autowhisper button!

Both of these have a verb counterpart of course, 'Toggle Autowhisper' and 'Set Autosubtle Mode', for your macroing pleasure.


code/modules/mob/autowhisper.dm
```diff
@@ -0,0 +1,45 @@
...
+		to_chat(src, "<span class='notice'>Autowhisper has been [autowhisper ? "enabled. You will now automatically whisper/subtle when using say/me" : "disabled"].</span>")
+
+/mob/living/verb/autowhisper_mode()
+	set name = "Set Autosubtle Mode"
+	set desc = "Set the mode your emotes will default to while using Autowhisper"
+	set category = "IC"
+
+
+	var/choice = tgui_input_list(src, "Select Custom Subtle Mode", "Custom Subtle Mode", list("Adjacent Turfs (Default)", "My Turf", "My Table", "Current Belly (Prey)", "Specific Belly (Pred)", "Specific Person", "Psay/Pme"))
+	if(!choice || choice == "Adjacent Turfs (Default)")
```
Heroman3003, line 31: I think the list should have a separate option for 'null' from "Adjacent Turfs (Default)". It can be called something more obvious, just that it seems like if someone wants to reset back to "ask me every time" mode, its not obvious that you select default mode or use cancel in the window to do that.

Very-Soft: > I think the list should have a separate option for 'null' from "Adjacent Turfs (Default)". It can be called something more obvious, just that it seems like if someone wants to reset back to "ask me every time" mode, its not obvious that you select default mode or use cancel in the window to do that.

While autowhisper is on, it doesn't ask you every time unless you select an option that needs a target. Adjacent Turfs (Default) behaves the same way normal subtle does!

PR #15157, opened 2023-07-18
Moar buttan

Very-Soft:
Adds more buttons

image

On the right side there are four new buttons! 
The left most one, L, opens the check known languages panel!
The right most one, P, opens set pose!
The arrows below these two are 'move up' and 'move down' respectively, for Z level travel.

The U above the hands in the bottom center allows one to use the item in their active hand on themself! This allows one to use items on themself when they are not visible on screen, such as if one is in a belly, a locker, a hand, or otherwise obscured! Mostly meant for the 'I'm in a belly and I want to eat this cheeseburger in my pocket' situations that come up. People could probably use it for other reasons too though.


BonniePandora:
No preview of what they look like for a player?

Curt5:
I like most of these, but the "click self" button is... really weird and dumb seeming to me. In what scenario is it easier to click a tiny button on the side of your screen rather than click your character right in the middle?
I think removing that would be good, to reduce clutter on the UI. The others seem nice additions.

Very-Soft:
> I like most of these, but the "click self" button is... really weird and dumb seeming to me. In what scenario is it easier to click a tiny button on the side of your screen rather than click your character right in the middle? I think removing that would be good, to reduce clutter on the UI. The others seem nice additions.

when your character is not visible, such as when you are in a belly, or a locker, or otherwise obscured and unclickable.

PR #15160, opened 2023-07-19
ports tallborgs from skyrat

PastelPrinceDan:
i've been wanting to port these for a while, and some people have voiced interest in the Discord about having these here too, but didn't know how to do it. UniquaSa over on CHOMP ported them, so i'm shamelessly taking the way they ported it and adding them here too.
atm they're a dogborg subtype; i'm aware it's not intuitive, and there's likely better ways to do this, but i'm not a coder by any means. if anyone wants to do it the 'right' way, **please** do so. for now, though, it'll do.

moving on to the actual content: **tallborgs from skyrat!**
in order, from left to right:
- original MEKANIKA sprites by theOOZ/ghostsheep (included because some people (me) prefer some of the old sprites)
- MEKANIKA resprites by Ebin-Halcyon
- NIKA & NIKO sprites by Motho & TheDragMeme
- K4T sprites by Zydras & Juniper

not pictured here but included in the spritesheet (as per some spriters' request to port the whole skinlines): cargo modules, standard modules, syndicate & ninja modules, clown module for K4T.

Untitled

belly sprites for MEKA v2, NIKA and NIKO modules included, all by Kcin from CHOMP

here's a preview
check the [original PR on CHOMP](https://github.com/CHOMPStation2/CHOMPStation2/pull/6611) for more

image


Very-Soft:
I love them

code/modules/mob/living/silicon/robot/robot_vr.dm
Aroliacue, line 1: I would recommend adding vis_height = 32 at the start of /mob/living/silicon/robot/proc/vr_sprite_check().

Then in the else if for the new tallborgs, add:
vis_height = 64
update_transform

This will fix tile scaling issues that Chomp had when we first added them should your size be larger than 200%.

PastelPrinceDan: done, thank you!

Heroman3003:
Only concern I have is artists' permissions, otherwise it looks good.

Aroliacue:
Meka has a credit file, check the additions. It should be fine to use!

PastelPrinceDan:
Credits were given (icons/txt file) and the original skinline artist is fine with porting, given the txt file that came with these. It should be fine!

PR #15162, opened 2023-07-19
Recolourable flannel

SatinIsle:
Added a recolourable flannel jacket. I have managed to make the whites light enough to show any colour whilst keeping the gradiants and the blacks.

It is a separate option to the other recolourable jackets, as I couldn't figure out the way to add it as a weak to the list without making the others recolourable too. I think that's just a limitation of the loadout system, as all other recolourable things seem to work the same way.

image


Very-Soft (changes requested):
Conflicts

SatinIsle:
> 

I believe that I have resolved the conflicts now.

PR #15166, opened 2023-07-22
Simplified Coin Press

KillianKirilenko:
Rewrites the coin press to a much simpler method of operation: it is now a basic hand-operated press that turns one ingot into six coins (loss of 500 material units in shavings/etc.) after two seconds, placing them at the user's feet. It will continue to process as long as your in-hand stack has sheets on it, and if you move away during the process it aborts.

Valid materials are silver, gold, diamond, phoron, uranium, and iron (previously steel). More could be added pretty easily too, if coin sprites and objects were made and coded.

Also adds a description to the coin press, since it hasn't had one this entire time.

Tested and works quite nicely. Now the Talon's coin press will actually be usable!

Previously the coin press required mapped input and output spots for some reason. Typical overcomplicated project that the original developer seems to have forgotten about, no surprises there.

### DIFF
--- a/code/modules/economy/mint.dm
+++ b/code/modules/economy/mint.dm
@@ -1,210 +1,49 @@
 /**********************Mint**************************/
 /obj/machinery/mineral/mint
 	name = "Coin press"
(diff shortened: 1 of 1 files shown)


Heroman3003:
I put reviewing this off, and while my only thing to say on this is very minor and practically ineffectual, ideally the names of mats should've used defines rather than plaintext. Nothing big or even needing instant fixing or anything, just for the future.

PR #15169, opened 2023-07-23
Changed the new player notification sound to be less intrusive.

SatinIsle:
This changes the sound that plays when a player connects for the first time or has a byond account made within the last two days. The new sound is a more subtle and pleasant ping, far less annoying and intrusive.

The sound is: https://freesound.org/people/bluetshirt/sounds/570265/
It is a creative commons 0 license, meaning the author has waived all rights to the work.


Very-Soft:
tested it, seems fine

PR #15172, opened 2023-07-24
Adds(feature): Triggerable Landmarks for Event-help

Runa-Dacino:
### What This does

Adds a new tool for staff to use when either preparing or actively running events.
By pressing the "Manage Event Triggers" button under the EventKit tab, a set of dialogue boxes appear instructing the user in creating either a notification trigger or a narration trigger.
Furthermore, it permits managing (deleting or teleporting to) the trigger at will if you own it,
and deleting other staff's triggers if they're DC/inactive (30 mins). You may teleport at any time.

Other staff's triggers may be accessed by inputting their ckey in the appropriate dialogue box.

The **Notify event Trigger** creates a special turf interaction (like mousetraps) where if a mob that isn't an observer AND has a ckey, it sends a message to the staffmember that created the landmark UNLESS they marked it as a "teamwork trigger" OR they've DC'd. In which case, it sends to adminlog.

The sent message includes whether the trigger is a narrate or notify type, the player's mobname and ckey that triggered it, who placed the trigger (if teamwork mode/DC'd) and the coordinates of the trigger. Furthermore, for direct notifications a "BWOINK!" may be requested to be played.

Afterwards, the trigger either deletes itself or goes on cooldown, based on inputted options.

The **Narrate event trigger** has all the functionalities of the Notify trigger, PLUS it can do automated narration on being triggered.

It may then either to_chat() the player that triggered it, either normally or with big scary red text OR create a visible/audible message of specified hearing/viewing range. 

The message it displays supports normal chat formatting (+bold+, _underscore_, |italics|)

### Why We Need this

Like Entity Narrate, this came to me in a dream.

I foresee four usecases:

A staffmember prepares a little surprise ahead of time, then alt tabs or goes on to do some low-investment player-side activities rather than actively stare for someone who bites. Combined with the bwoink and the message, they can tap out/tab back in and head over to make the surprise interactive for the player.

Alternatively, in an event that has a risk of the participants splitting up - the staffmember may set up special loot, points of interest, key chokeholds for "Hey, I should check over there because someone is touching something important."

Furthermore, the narrate functionality can ensure the player won't dash past the surprise - either due to the weird visible/audible message or the direct to_chat

Finally, in a larger event with some "Things are not exactly what the in-game sprites/mechanics say" scenarios, the staffmember may place strategic notifications to ensure players are properly informed of such divergences.

### Testing

- Tested for stepping onto my own triggers of both types (self-delete)
- Tested for repeating triggers as well
- Tested with a guest account for both of the above
- Tested to_chat, visible and audible narrates
- Tested self-management 
- Partially tested managing others' triggers. Mainly could not test the inactivity pass.

### Commit Details
https://github.com/VOREStation/VOREStation/pull/15172/commits/694a5b617b82aca343481e3602b5811798682163
 - Adds a new verb to staff with FUN permissions under the eventkit tab
     - name: "Manage Event Triggers"
     - Permits creating "Notification" and "Narration" landmarks
     - Permits deleting/teleporting to self-created landmarks at any time
     - Permits deleting other staff's landmarks if DC'd/Inactive
     - Permits teleporting to other staff's landmarks at any time
(the landmark management buttons I couldn't implement without code repetition due to managing others' landmarks needing (imo) admin logging)
 - Adds new landmark subtype: event_trigger
     - Handles notification of staff/team depending on options
     - May autodelete if requested
     - Makes a loud bwoink at the creator if requested
     - Is triggered by the base DM proc /Crossed(var/atom/movable/AM)
     - Ghosts and NPCs don't trigger it.
 - Adds a subtype of event_trigger, auto_narrate
     - Adds functionality to automatically print a normal/red message to player entering the turf with the landmark.
     - Alternatively, allows creating a visible/audible message when the turf is triggered
     - Inherits all behaviours of event_trigger otherwise
  -Adds new global list, "event_triggers"
      - It is an associative list of key:list(element1,element2,element3) form
      - key is the staffmember's ckey, elements are references to event_trigger landmarks they created


Heroman3003:
Can't see any immedeate problems on first skim

code/game/objects/effects/landmarks_events.dm
```diff
@@ -0,0 +1,129 @@
...
+	var/coordinates = ""
+	var/cooldown = 0 //in seconds
+	var/last_trigger = 0
+	var/isLoud = FALSE
+	var/isNarrate = FALSE
+
+/obj/effect/landmark/event_trigger/New()
+	..()
+	coordinates = "(X:[loc.x];Y:[loc.y];Z:[loc.z])"
+	invisibility = 100
```
Heroman3003, line 22: Why is invisibility set in new and not in the main definition itself, if its never touched anyway?

code/game/objects/effects/landmarks_events.dm
```diff
@@ -0,0 +1,129 @@
...
+
+/obj/effect/landmark/event_trigger/proc/set_vars(mob/M)
+	name = sanitize(tgui_input_text(M, "Input Name for the trigger", "Naming", "Event Trigger"))
```
Heroman3003, line 25: Will cancelling input not result in a blank name here?

code/game/objects/effects/landmarks_events.dm
```diff
@@ -0,0 +1,129 @@
...
+		delete_me = TRUE
+	log_admin("[M.ckey] has created a [isNarrate ? "Narrtion" : "Notification"] landmark trigger at [coordinates]")
+
+/obj/effect/landmark/event_trigger/Destroy()
+	if(event_triggers[creator_ckey])
+		event_triggers[creator_ckey] -= src
+	..()
+
+/obj/effect/landmark/event_trigger/Crossed(var/atom/movable/AM)
+	if(isobserver(AM) || !isliving(AM))
```
Heroman3003, line 47: isobserver is included in !isliving naturally.

code/game/objects/effects/landmarks_events.dm
```diff
@@ -0,0 +1,129 @@
...
+Admin verb is called by code\modules\admin\verbs\event_triggers.dm
+*/
+/obj/effect/landmark/event_trigger
+	name = "ONLY FOR EVENTS. DO NOT USE WHEN MAPPING"
+	desc = "Please notify staff if you see this!"
+	var/creator_ckey = ""
+	var/isTeamwork = FALSE //Whether to notify creator or whole team.
+	var/isRepeating = FALSE
+	var/coordinates = ""
+	var/cooldown = 0 //in seconds
```
Heroman3003, line 14: In world ticks actually (aka miliseconds)

code/game/objects/effects/landmarks_events.dm
```diff
@@ -0,0 +1,129 @@
...
+	if(!L.ckey) return FALSE
+
+	if(world.time < (last_trigger + cooldown))
+		return FALSE
+	if(!isRepeating && last_trigger) //Used to avoid spam if qdel(src) fires too slowly
+		return FALSE
+	last_trigger = world.time
+
+	if(!creator_ckey)	//For some reason, the user didn't have a ckey. Let's clean up
+		qdel(src)
```
Heroman3003, line 59: Are you sure you want even narration landmarks to be deleted before actually narrating just because creator key is missing? Shouldn't happen normally, but still.

code/game/objects/effects/landmarks_events.dm
```diff
@@ -0,0 +1,129 @@
...
+		if(!isRepeating)
+			qdel(src)
+		return FALSE
+	else
+		return L
+
+
+/obj/effect/landmark/event_trigger/auto_narrate
+	var/message
+	var/isVis_or_isAud = 0	//0 for vis, 1 for aud
```
Heroman3003, line 85: honestly, isVis_or_isAud and isPersonal could all be rolled up into single var with 3 values, 0, 1 and 2 and comparing it to those specific numbers.

code/modules/admin/verbs/event_triggers.dm
```diff
@@ -0,0 +1,121 @@
...
+
+	var/mode = tgui_input_list(src, "What do you wish to do?", "Manage Event Triggers", \
+	list(
+		"Create Notification Trigger",
+		"Create Narration Trigger",
+		"Manage Personal Triggers",
+		"Manage Other's Triggers",
+		"Cancel"
+	), "Cancel" )
+	if(mode == "Cancel") return
```
Heroman3003, line 21: Technicaly, nothing happens if you just cancel out of it, but still, having !mode check there is good practice I think.

code/modules/admin/verbs/event_triggers.dm
```diff
@@ -0,0 +1,121 @@
...
+						if(confirm_teleport == "Cancel")
+							return
+					M.forceMove(get_turf(ET))
+				if(decision == "Delete")
+					if(other_ckey && GLOB.directory[other_ckey])
+						var/client/C = GLOB.directory[other_ckey]
+						var/mob/M = C.statobj
+						if(M.client && M.client.inactivity < 30 MINUTES)
+							to_chat(src, SPAN_WARNING("[M] has only been inactive for [M.client.inactivity / (1 MINUTE)] minutes.\n \
+							If you want to delete their event triggers, ask them in asay or discord to do it themselves or wait 30 minutes"))
```
Heroman3003, line 113: This is up for staff debate tbh, but I do think having some authority override on ability to delete those should be present.

code/game/objects/effects/landmarks_events.dm
Runa-Dacino: The reason it's set in New() because the type it inherits from (landmark) sets it at 101, and I want it at 100 for ability to see it  with see invisibility

code/game/objects/effects/landmarks_events.dm
Runa-Dacino: Good point, I'll make cancelling return FALSE

code/game/objects/effects/landmarks_events.dm
Runa-Dacino: Yes, because you cannot delete them using the management screen and it could be spammy for event chat if badly set.

code/modules/admin/verbs/event_triggers.dm
Runa-Dacino: Wouldn't kicking a non-compliant staff creating abusive landmarks handle this?

code/modules/admin/verbs/event_triggers.dm
Runa-Dacino: I'll make it if(!mode || mode == "Cancel") return then

code/game/objects/effects/landmarks_events.dm
Heroman3003: Wait, landmark sets it in new? Why does THAT do that? Why not change that?

code/game/objects/effects/landmarks_events.dm
Heroman3003: Just have a temporary var for name and check like if(!var) return or something

code/modules/admin/verbs/event_triggers.dm
Heroman3003: On general level yes, but again, this is something for staff to discuss as a whole. I just think that, code-side, these things shouldnt be hardcoded like that, with 0 regard for hiearchy or potential abuse.

code/game/objects/effects/landmarks_events.dm
```diff
@@ -82,9 +84,8 @@ Admin verb is called by code\modules\admin\verbs\event_triggers.dm
 
 /obj/effect/landmark/event_trigger/auto_narrate
 	var/message
-	var/isVis_or_isAud = 0	//0 for vis, 1 for aud
+	var/isPersonal_orVis_orAud = 0	//0 for personal, 1 for vis, 2 for aud
```
Heroman3003, line 87: The only thing I'll say is that this name is way too wordy. That, again, is subjective/personal opinion, but something far simpler, like narrate_mode would be easier to understand. But one var is still better.

Runa-Dacino: It's wordy, but the code it's used in is fairly short and it stays within like 68 colums or whatever the standard is (it looks fine in my code editor on desktop)

As it is not a var that'll ever really be reused outside of this feature - I think it should be fine to stay as is.

I did think of setting up some preprocessor #defines but I felt it may be too much for sth like this.

PR #15186, opened 2023-07-28
Underwear? Under there!

KillianKirilenko:
Recolours some of the greyscale underwear so that they'll take colours better and more closely match other items of underwear without too much fiddling. Applies to boxers, both masc. tank tops, the undershirt, and the alt fem. tank top.

Also adds two new underwear options; HYPER briefs and boxers. They have no mechanical effects, they just share the aesthetic of the jumpsuit.


Heroman3003:
uwu hyper underwear....

PR #15188, opened 2023-07-29
Staves

KillianKirilenko:
Just a little experiment in making a new material weapon type; staves! ~~Or staffs, if you prefer being *wrong*.~~

Whilst not very powerful as weapons go, the humble staff has a couple of perks;
- First, using it two-handed allows you to occasionally block/parry incoming attacks
- Second, being on disarm stance increases the chance to block/parry and gives a decent chance to briefly weaken any opponent you hit (the tradeoff being that you still need to be wielding it and if you're in disarm stance your attacks deal half damage)

They can be crafted from most material sheets.

I slapped these together in maybe an hour on a whim. The sprites are pretty simplistic. The balance is probably awful. But hey, funny stick go *bap*.


CMDRYouth:
I thought Virgo didn't do Combat?

PR #15226, opened 2023-08-09
fixes equip/dequip noises from not playing

Seris02:
image


Heroman3003:
Don't forget the preference check for equip/drop sounds.

Heroman3003:
This PR appears to not involve the sound preferences in the check. While I am all for this check being fixed, I'd like the sound pref checks to also be added there.

PR #15230, opened 2023-08-11
Custom Link!

Very-Soft:
Allows one to set a custom link to show up in their examine text!

One thing people used the OOC notes for was to post image/flist links in the chat and get clickable links!

Since the OOC panel is a thing now, I thought it would be good to enshrine that function into something that's specifically for it.

This allows you to enter an extra 100 character string underneath your flavortext in your examine text, which can technically be anything, but it's intended to be space for links to show up and be clickable!

image


Heroman3003:
Seems fine

Heroman3003 (approved):

Issue #14817, opened 2023-04-23
If you lose all your blood, you lose the ability to have blood [GENERAL] 


TankTheBirb:
#### Brief description of the bug
When someone (somehow) reaches 0 blood via blood loss by bleeding, they cannot have blood in their vessel anymore.

#### What you expected to happen
Well I kinda expect blood to be like, something you can put back in people.

#### What actually happened
reagents_list variable in the carbon mob's vessel gets cleared. (see pictures)
Person who lost all their blood
![image](https://user-images.githubusercontent.com/96669534/233864754-496b28bf-c4ec-403d-aa0c-7244a7718687.png)
Person who has not lost any blood
![image](https://user-images.githubusercontent.com/96669534/233864805-4db5f774-72ea-4c98-8d0c-23a95f8e67eb.png)

#### Steps to reproduce
1. Cause a carbon mob to bleed enough that they bleed out to 0% blood
2. If they're a besk, they will somehow be alive throughout this unless you deliberately kill them. However this should work on all carbons with blood to lose.
3. Once they hit 0% blood, the reagent_list list clears its entry of the blood reagent datum
4. They can no longer receive blood via natural regeneration, injection, IV, or any other method I tried. 
5. Make bug report

#### Code Revision
Server revision: B:release D:
Commit: [efd9bfe5a34db70990f40a78b2ecb975ad26a87b](http://127.0.0.1:58539/tmp15472/%3Ca%20href=%22https://github.com/VOREStation/VOREStation/commit/efd9bfe5a34db70990f40a78b2ecb975ad26a87b%22%3Ehttps://github.com/VOREStation/VOREStation/commit/efd9bfe5a34db70990f40a78b2ecb975ad26a87b%3C/a%3E)
 

#### Anything else you may wish to add:
Please do not lose all your blood. We have to aheal it back.

PR #15250, opened 2023-08-13
Fixes human mobs ending up with their blood datums deleted, creates fallback if that still happens

Runa-Dacino:
### What This does

In multiple commits, it tries to avert the blood datum getting deleted by hitting 0 blood.

1. Updates the code to use remove_blood() rather than remove_reagent() where sensible. Remove_reagent straight up deletes blood, whereas remove_blood() has checking to ensure at least 1 u blood remains
2. Updates remove_blood to scale its blood loss by creature size BEFORE it does sanity checks
3. Updates remove_blood and take_blood to use a preprocessor #define to outright stop processing/handling blood removal if we're below critical treshold (defined as 2.1, and we increased sanity check to be at 2u. 1u is too dangerous due to floating point weirdness (assuming)).
4. Implements a fall-back if the blood datum still gets deleted based on staff-side fix for such issues
### Why we need this

Fixes https://github.com/VOREStation/VOREStation/issues/14817

More generally, remove_reagent("blood", amt) is too dangerous to use, we should use the appropriate proc for handling blood which is remove_blood. In some cases direct touching is justified - such as during set_species as O_HEART might not apply to all species but we still want them to have blood and sometimes we're working with an organ outside a mob and other wackiness.

We also want a fallback because although I did my best to do testing - organ code is cursed and I rather want to make it a seamless experience even if stuff breaks. We make sure to create debug logs if it still happens so that we can track down such issues.

### Commit details

https://github.com/VOREStation/VOREStation/commit/09acdfd92225de78f764f99057ba01d2063f221e

Fixes https://github.com/VOREStation/VOREStation/issues/14817

remove_reagent has no sanitization for making sure the person has at least 1 u of blood remaining, and is therefore unsafe to use.

This commit changes all human remove_reagent("blood", amt) calls with remove_blood(amt).

This should prevent blood disappearing from internal bleeding or dragging someone while they're down or from drawing blood from them and so forth.

Not all cases of remove_reagent("blood", amt) were changed, as some act on organs or other reagent containers not part of a human mob or because it's set_species

https://github.com/VOREStation/VOREStation/commit/12bb4f00bf8f8fd557b9dbdb7854fd01fffa33c0

Mob size affects how much blood you should lose, by default 1 (20/20). However, in cases this is not true (non-medium mobs), this can lead to overflow of blood taken as this scaling happens AFTER we ensure amt is not greater than whatever it takes to get 1 blood.

This should safeguard against that.

https://github.com/VOREStation/VOREStation/commit/f1b5e812ce3ccc7d0285cf77dcbf6ce2286b00a8
Also stops processing blood loss at 2.1. This allows the mob to regenerate some blood without losing any more, but I doubt anyone would farm a human mob for blood at a rate of 0.1 per multiple ticks. 2.1 is defined as a preprocessor #define to make it easier to sync stuff and avoid magic numbers.

The fallback essentially follows the same procedures as a staff member would do to fix this issue: creates blood, sets their blood volume at species amount if wrong, fixes datums around blood. It logs failures and occurances.

### DIFF
--- a/code/modules/organs/blood.dm
+++ b/code/modules/organs/blood.dm
@@ -219,8 +219,6 @@ var/const/CE_STABLE_THRESHOLD = 0.5
 	if(!amt)
 		return 0
 
-	amt = amt * (src.mob_size/MOB_MEDIUM)
-
 	var/current_blood = vessel.get_reagent_amount("blood")
 	if(current_blood < BLOOD_MINIMUM_STOP_PROCESS)
 		return 0 //We stop processing under 3 units of blood because apparently weird shit can make it overflowrandomly.
@@ -271,7 +269,7 @@ var/const/CE_STABLE_THRESHOLD = 0.5
 	if(!should_have_organ(O_HEART))
 		return null
 
-	if(vessel.get_reagent_amount("blood") < amount || vessel.get_reagent_amount("blood") < BLOOD_MINIMUM_STOP_PROCESS)
+	if(vessel.get_reagent_amount("blood") < max(amount, BLOOD_MINIMUM_STOP_PROCESS))
 		return null
 
 	. = ..()
--- a/code/modules/organs/blood.dm
+++ b/code/modules/organs/blood.dm
@@ -210,10 +210,12 @@ var/const/CE_STABLE_THRESHOLD = 0.5
 	if(!amt)
 		return 0
 
+	amt = amt * (src.mob_size/MOB_MEDIUM)
+
 	if(amt > vessel.get_reagent_amount("blood"))
 		amt = vessel.get_reagent_amount("blood") - 1	// Bit of a safety net; it's impossible to add blood if there's not blood already in the vessel.
 
-	return vessel.remove_reagent("blood",amt * (src.mob_size/MOB_MEDIUM))
+	return vessel.remove_reagent("blood",amt)
 
 /****************************************************
 				BLOOD TRANSFERS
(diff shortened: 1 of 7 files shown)


code/modules/organs/blood.dm
```diff
@@ -256,11 +271,11 @@ var/const/CE_STABLE_THRESHOLD = 0.5
 	if(!should_have_organ(O_HEART))
 		return null
 
-	if(vessel.get_reagent_amount("blood") < amount)
+	if(vessel.get_reagent_amount("blood") < amount || vessel.get_reagent_amount("blood") < BLOOD_MINIMUM_STOP_PROCESS)
```
Heroman3003, line 274: vessel.get_reagent_amount("blood") < amount || vessel.get_reagent_amount("blood") < BLOOD_MINIMUM_STOP_PROCESS

Could be replaced with

vessel.get_reagent_amount("blood") < max(amount, BLOOD_MINIMUM_STOP_PROCESS)

code/modules/organs/blood.dm
```diff
@@ -210,10 +219,16 @@ var/const/CE_STABLE_THRESHOLD = 0.5
 	if(!amt)
 		return 0
 
-	if(amt > vessel.get_reagent_amount("blood"))
-		amt = vessel.get_reagent_amount("blood") - 1	// Bit of a safety net; it's impossible to add blood if there's not blood already in the vessel.
+	amt = amt * (src.mob_size/MOB_MEDIUM)
```
Heroman3003, line 222: I think taking out that division from under remove_reagent was all that was needed to fix it in the first place because it was probably causign rounding errors, but rest of it seems fine either way.

Heroman3003:
Unrelated note, but I'd remove mob size from bloodloss consideration entirely, as something to consider. The only species we have thats not normal size are monkey species and werebeast, and its not some big point of balancing concern either way.

PR #15252, opened 2023-08-13
Robot species

Very-Soft:
Robots don't have a species but now it will tell people what sprite you picked in your examine text, similar to how species is done for carbon mobs!

image
image

If you don't pick a sprite, or are a maint drone, it doesn't set a name like this and will look as it normally does!

### DIFF
--- a/code/modules/mob/living/silicon/robot/robot.dm
+++ b/code/modules/mob/living/silicon/robot/robot.dm
@@ -1125,7 +1125,7 @@
 		var/selection = tgui_input_list(src, "Select an icon! [triesleft ? "You have [triesleft] more chance\s." : "This is your last try."]", "Robot Icon", module_sprites)
 		sprite_datum = selection
 		if(!istype(src,/mob/living/silicon/robot/drone))
-			robot_species = selection
+			robot_species = sprite_datum.name
 		if(notransform)
 			to_chat(src, "Your current transformation has not finished yet!")
 			choose_icon(icon_selection_tries)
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/silicon/robot/robot.dm
```diff
@@ -1122,6 +1124,8 @@
 	else
 		var/selection = tgui_input_list(src, "Select an icon! [triesleft ? "You have [triesleft] more chance\s." : "This is your last try."]", "Robot Icon", module_sprites)
 		sprite_datum = selection
+		if(!istype(src,/mob/living/silicon/robot/drone))
+			robot_species = selection
```
Heroman3003, line 1128: I'd suggest going for 'sprite_datum.name' instead of 'selection' here

code/modules/mob/living/silicon/robot/robot.dm
```diff
@@ -1147,6 +1151,7 @@
 
 	icon_selected = 1
 	icon_selection_tries = 0
+	sprite_type = " [robot_species]"
```
Heroman3003, line 1154: I'd suggest not putting a space in the sprite_type, but instead up where name as a whole is formed.

Very-Soft: If I do that then drones and people who don't pick a sprite will have double spaces in the name. u wu

code/modules/mob/living/silicon/robot/robot.dm
Very-Soft: Why, the selection is the name of the sprite you picked as it appears in the list.

code/modules/mob/living/silicon/robot/robot.dm
Heroman3003: Something like '[sprite_type ? "[sprite_type] " : ""]' maybe?

code/modules/mob/living/silicon/robot/robot.dm
Heroman3003: Selection is actually sprite itself, and it might be byond jank but it automatically applies name var to the list/converts it when doing "[selection]" *despite* the fact that name is not a native datum var and its weird like that and this is mostly pedantics.

PR #15274, opened 2023-08-17
Char Setup: Reset Buttons

KillianKirilenko:
Quick and dirty addition of some reset buttons for nickname and all three record groups, as I mentioned maybe adding a little while ago.

The html for flavour text is a bit more complex and I'm not sure how best to approach that, but I would like to do a similar function for that, to ensure that when an entry is wiped the values are properly nulled and not just left as `""` or `" "`.

Also corrects the capitalization of Custom **H**eat Discomfort message under the misc. tab since that didn't match the rest of that lot and it was vaguely bothering me.


Heroman3003:
"" is functionally identical to null when it comes to if(x) checks so that's fine for flavour text, for the future

PR #15290, opened 2023-08-21
Being outdoors doesn't mean you can be dark forever

Very-Soft:
Makes it so that outdoors being true doesn't automatically return false on set_luminosity, and instead has it check to see if the turf is included in a planet or a fake_sun's area (if the fake_sun is actually doing the light part). If it's not, then it will set_luminosity as one would expect.

This means that outdoor turfs that DON'T have light can be lit up by lights!

It might be somewhat cool to make it so that planet and fake sun will store the planet lighting information on the turf, so that you can light it up, and have it return to those values if your light goes out. But that's a future pondering.

### DIFF
--- a/code/modules/lighting/lighting_fake_sun_vr.dm
+++ b/code/modules/lighting/lighting_fake_sun_vr.dm
@@ -1,3 +1,5 @@
+var/static/list/fake_sunlight_zs = list()
+
 /obj/effect/fake_sun
 	name = "fake sun"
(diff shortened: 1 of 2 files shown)


Heroman3003:
This looks like a decent fix, although it seems like outdoors_adjacent previously overrode even force so I'm curious what that's about; and I do think logic in that if proc could be redone to not have luminousity set twice, but beyond those things, fix as a whole looks good.

PR #15291, opened 2023-08-22
3 edited outfits

AmatsuDF:
This PR contains three outfits, listed below:

A shortsleeve hoodie variant, a greyscaled version of the Flame Dress, and a greyscaled version of the Yellow Dress. All three can be recolored in the loadout or by the recolor machines (can't recall their name right now) on all maps. The only significant name change from the source outfit is renaming the greyscaled Yellow Dress to 'Fancy Dress, colorable' as it no longer was yellow. The original icons are untouched and still available.


code/modules/clothing/under/miscellaneous_vr.dm
```diff
@@ -596,3 +596,23 @@
...
+
+/obj/item/clothing/under/dress/cdress_fire
+	name = "flame dress"
+	desc = "A small black dress with a flames print on it. Perfect for recoloring!"
+	icon = 'icons/inventory/uniform/item_vr.dmi'
+	default_worn_icon = 'icons/inventory/uniform/mob_vr.dmi'
+	icon_state = "cdress_fire"
+
+/obj/item/clothing/under/dress/cbridesmaid
+	name = "fancy dress, recolorable"
```
Heroman3003, line 608: I suggest not putting 'recolorable' or 'colorable' on actual item names, only loadout names

PR #15320, opened 2023-08-27
515 Compat

ItsSelis:
This PR is basically a big ToDo List of things that are still not working correctly and for people to test out.

---

ToDo:
- [x] Browser Button always automatically pressed (Verbpanel) (Usually happens when a byond window was closed)
- [x] Fixing "code/modules/mob/living/carbon/human/human_modular_limbs.dm,76: wrong type of value for list"
- [x] Fixing "code/modules/mob/living/carbon/human/human_modular_limbs.dm,80: wrong type of value for list"
- [x] Fixing "code/modules/vore/resizing/sizegun_vr.dm,28: wrong type of value for list"
- [x] Fixing "code/datums/callback.dm,89: bad proc"
- [x] Comment out (or create config) to disable prof_init() in code/game/world.dm by default
- [x] Fixing errors in https://github.com/VOREStation/VOREStation/actions/runs/5991530920/job/16249968123?pr=15320
- [x] vchat is currently making default byond html windows unusable after an unknown amount of time. It is also currently unknown why exactly this happens (will be fixed in #14625)

The (out of scope) checklist items were moved to #15771 as these fixes/changes also apply to 514 without causing the same issues/bugs that only appear on 515.

Notes:
1) Some switch statements (647d4f4d496ece560125a09f8794439ab5225e51) had to be turned into if statements due to the switch cases only being able to have constant vars inside of them.
2) Tracy Profiler has been added to monitor performance (83ffa14ae7639d7dc3c2d4fe0bea392e88129684)
3) The workflow has been configured to use 515.1610 for the BYOND version (8878874f9053707772953a1bd66d7fca2184b360)

### DIFF
--- a/_build_dependencies.sh
+++ b/_build_dependencies.sh
@@ -4,8 +4,8 @@ export SPACEMAN_DMM_VERSION=suite-1.7
 # For NanoUI + TGUI
 export NODE_VERSION=16
 # Byond Major
-export BYOND_MAJOR=515
+export BYOND_MAJOR=514
 # Byond Minor
-export BYOND_MINOR=1620
+export BYOND_MINOR=1589
 # Macro Count
 export MACRO_COUNT=4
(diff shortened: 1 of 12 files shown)


ItsSelis:
Contents of this PR have been moved/applied to #15739

PR #15323, opened 2023-08-29
Makes it so that tesla's zaps no longer respect LOS

Heroman3003:
Title. DOES NOT apply to regular lightning strikes, only tesla zaps.

Makes it more destructive when loose, and also reduces effectiveness of certain engine setups.

### DIFF
--- a/code/game/objects/structures/flora/trees.dm
+++ b/code/game/objects/structures/flora/trees.dm
@@ -124,7 +124,7 @@
 	if(Proj.get_structure_damage())
 		adjust_health(-Proj.get_structure_damage(), TRUE)
 
-/obj/structure/flora/tree/tesla_act(power, explosive)
+/obj/structure/flora/tree/tesla_act(power, ignore_los, explosive)
 	adjust_health(-power / 100, TRUE) // Kills most trees in one lightning strike.
 	..()
 
@@ -295,4 +295,3 @@
 		set_light(bulbs, 1, "#33ccff")	// 5 variants, missing bulbs. 5th has no bulbs, so no glow.
 		add_overlay(mutable_appearance(icon, "[base_state][bulbs]_glow"))
 		add_overlay(emissive_appearance(icon, "[base_state][bulbs]_glow"))
-	
--- a/code/modules/admin/verbs/lightning_strike.dm
+++ b/code/modules/admin/verbs/lightning_strike.dm
@@ -72,14 +72,14 @@
 		return
 
 	if(ground) // All is well.
-		ground.tesla_act(LIGHTNING_POWER, FALSE)
+		ground.tesla_act(LIGHTNING_POWER, FALSE, FALSE)
 		return
 
 	else if(coil) // Otherwise lets bounce off the tesla coil.
-		coil.tesla_act(LIGHTNING_POWER, TRUE)
+		coil.tesla_act(LIGHTNING_POWER, FALSE, TRUE)
 
 	else // Striking the turf directly.
-		tesla_zap(T, zap_range = LIGHTNING_ZAP_RANGE, power = LIGHTNING_POWER, explosive = FALSE, stun_mobs = TRUE)
+		tesla_zap(T, zap_range = LIGHTNING_ZAP_RANGE, power = LIGHTNING_POWER, explosive = FALSE, stun_mobs = TRUE, ignore_los = FALSE)
 
 	// Some extra effects.
 	// Some apply to those within zap range, others if they were a bit farther away.
--- a/code/modules/power/gravitygenerator_vr.dm
+++ b/code/modules/power/gravitygenerator_vr.dm
@@ -38,7 +38,7 @@ GLOBAL_LIST_EMPTY(gravity_generators)
 	if(prob(20))
 		set_broken()
 
-/obj/machinery/gravity_generator/tesla_act(power, tesla_flags)
+/obj/machinery/gravity_generator/tesla_act(power, ignore_los, tesla_flags)
 	..()
 	qdel(src)//like the singulo, tesla deletes it. stops it from exploding over and over
 
@@ -96,7 +96,7 @@ GLOBAL_LIST_EMPTY(gravity_generators)
 /obj/machinery/gravity_generator/main/station/Initialize()
 	. = ..()
 	setup_parts()
-	middle.add_overlay("activated")	
+	middle.add_overlay("activated")
 
 //
 // Generator an admin can spawn
@@ -408,7 +408,7 @@ GLOBAL_LIST_EMPTY(gravity_generators)
 /obj/machinery/gravity_generator/main/proc/update_list()
 	levels.Cut()
 	var/my_z = get_z(src)
-	
+
 	//Actually doing it special this time instead of letting using_map decide
 	if(using_map.use_overmap)
 		var/obj/effect/overmap/visitable/S = get_overmap_sector(my_z)
@@ -418,7 +418,7 @@ GLOBAL_LIST_EMPTY(gravity_generators)
 			levels = GetConnectedZlevels(my_z)
 	else
 		levels = GetConnectedZlevels(my_z)
-		
+
 	for(var/z in levels)
 		if(!GLOB.gravity_generators["[z]"])
 			GLOB.gravity_generators["[z]"] = list()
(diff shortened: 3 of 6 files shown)


Skysteak:
For the sake of having it put into words, how do you feel this will improve the server experience?

I have no dog in this fight,  but it wasn't described anywhere, so colour me curious.

TankTheBirb:
image
PR not verified, as it is not functioning as intended. Assuming this is what kind of build you were trying to stop. Verified that this is the PR build and it has the ignore_los changes in it.

Regardless of functionality, I'm not sure the intention behind this change does anything productive for the engineering gameplay experience. For one, removing the LOS requirement for zaps is unintuitive. People could rightly assume a solid wall would protect them and removing the LOS requirement means the only guaranteed protection is strapping yourself to a grounding rod. Also, with the significant toning down of random events, engineering is a "find your own fun" kind of department. A lot of that ends up being engine designs. Reducing or limiting creativity in engine designs is going to make people less inclined to do anything fun.

Also the tesla getting loose with these changes is likely to be a shuttle call most of the time rather than just the majority of the time. I don't see the value in that.

KillianKirilenko:
> For the sake of having it put into words, how do you feel this will improve the server experience?
> 
> I have no dog in this fight, but it wasn't described anywhere, so colour me curious.

The tl;dr is probably best shown by Tank's screencap, but to explain: the tesla obeys line of sight on its discharges and a couple of months back some engineers discovered this meant it could be corralled very effectively by using curtains to influence the pattern of discharges to yield anywhere upwards of 3 or 4x the power of a setup that *doesn't* use this mechanic. It requires very little effort compared to actually upgrading the components or otherwise increasing the yield as well.

It's basically become the new meta for tesla core setups, and any meta has the dangerous potential for players to start being assholes to each other.

I've been expecting this to get the bat for a while. It's a silly, and imo counter-intuitive means of minmaxing tesla power output.

Very-Soft:
AS SOMEONE WHO NEVER DOES THE CURTAIN META IN THEIR SETUPS

This seems like a very anti fun change specifically targetting experimentation and exploration of the mechanics of this engine.

I'm all for making the engine more destructive when it gets out, but that is 100% not why this change is being made. Do I think the curtain thing is stupid? Sure. But the tesla as an engine concept in general is stupid, so why would any customization of it be any different?

I don't think the curtain thing being a thing actually hurts anything by existing, and it only really adds fun for people who want to set it up.

pontifexminimus2:
I'm in full support of this, btw. Always stroke me as odd that the Tesla needs LOS when it can just phase through walls easily.

Heroman3003:
Frankly, yes, the curtains setup is the problem I personally see there. Not because it's possible to do or because it's unrealistic, but because it's become so seemingly common and simple to do, that now we have engineers who just passed the time requirements to play department building those and seemingly only knowing how to set engine that way. It turned from cool build that produces some extra power into a monkey-see-monkey-do level meta that HAS to be done every time, sometimes used by people with little understanding of why it works and how it works. To me, it's a sign that this is no longer some fun experimental thing, but is THE CORRECT META that is being actively taught to people as such. And I absolutely do not want that to be THE CORRECT META. It doesn't even matter much to me how much extra power it produces, it's the very fact that what was the meme/big build has taken that spot over the default setup.

RE: not working - PR's not yet stress tested, and I plan to do debugging in regards to it over next few days, I've mostly put it up as draft to start the inevitable discussion around it early. I just forgot to mark it as so, so my bad there.

RE: special insulator object - even if I were to, say, make tesla zaps target and melt curtains, there are already half a dozen potential replacements that are just as easy to use in that scenario. And making just one-specific-object capable of blocking lgihtning would mean either officially canonizing the strat by making it easily buildable roundstart, failing to solve the specific issue I am percieving, or effectively lock it out by NOT making it easily buildable roundstart. Not even mentioning the inevitable issue slew of actually making something like that actually work with lightning strikes.

NautilaChai:
Good change, the curtains were lame. Seeing those every shift wasn't creative or experimental. Looking forward to not seeing weird curtains in the engine room every single shift.

Runa-Dacino:
I am not a fan of curtains.

If the curtains were replaced with some sci-fi mcguffin, even if it literally just takes using plastic to make? I'd feel that'd be decent improvement for the current situation.

"Tesla Shielding" or something like that.

I am not a fan of curtains.

TankTheBirb:
> I'm glad someone's done the numbers at least. I want to say you can generate a fair bit more than 3.3MW if you fully upgrade the coils and set up the two freebie extras the SD starts with, but that does at least require cooperation from science to get bigger caps and similarly upgraing the SMES and/or building a second one to fully exploit, so that's not quite so bad.
> 
> Perhaps a better tack would be to sort of "canonize" it, with some kind of "tesla inhibitor" buildable? That way it's less absurd than just `ha ha funny showercurtain stop lightning`.

I assumed no upgrades because any upgrade from science immediately overshadows all other designs. Hell just knowing you can sometimes get up to a super capacitor in tech storage and that boosts your output by like 200% for some setups.

Wulfestwulf:
No more funni shower curtain able to stop MegaWatts of energy from passing through it...

Really though, this is a good change. The curtain meta is kinda insane and weird and has bred some elitism. Also making tesla loose's more dangerous is only good as it adds an additional level of urgency to the already bad situation.

SatinIsle:
I do not have strong feelings regarding the actual setup of the tesla, but I do not personally like meta of making it use comical props. I don't really mind whether this change goes through, or if there is a change to require something more sensible than curtains, but I support a change in general.

Naessa:
The best solution to elitism is to swat the prats who practice it. Yes, the curtain meta is silly (and metas in general exist in _every_ department, not just engineering) but SS13 is inherently a very silly game where eating enough carrots has a chance of healing ocular damage. If the curtains themselves are an issue, make some form of plastic shielding or option instead of the curtains, but tesloose is devastating as it is on smaller maps such as the Stellar Delight. If people want to optimize 10 MW setups then let them.

Skysteak:
Thank you for the earlier explanation on the reason why for this!

Given multiple people expressing a desire for an object to replace the curtains, and given the PR's aim breaking a meta that's supposedly become accepted with some not knowing the principles behind it, I almost want to say the solution should be to give this setup distinct advantages/disadvantages.  I fully admit I am not sure _how_ one goes about this, as it's beyond mechanics I really engage in, but having some sort of trade off keeps the experimentation aspect, whilst not just making this 'curtains method' a straight upgrade. I fully respect that this is not an easy task however, though it feels more 'correct' than straight up than removing a facet of an already engagement starved department.

Beyond that, making the Tesla more deadly is something I am not against, and I also think should be applied to the SM too. Just do be aware that you're increasing the likelihood of an engine loosing causing a round evac however.  

Also if this does go through, do announce it please, so we don't have another 'Belt of Holding change' style mess-up.

PastelPrinceDan:
i said this in the engi channel already but i'll put it here too so it's not lost

i personally don't care about the change
i used to use curtains but got bored of it, and went back to another set-up instead so it doesn't affect me at all
if bigger teslooses happen, might affect me, don't care either way. it's a game, round-ending stuff should be expected

however, i have to suggest just making a general announcement to remind players that elitism is shit behavior and to consider just jobbanning people who shit on others for not using their meta set-ups
rather than remove something some people actually have fun building and setting up
because people will fall back on another meta set-up, and elitism will breed from that from the bad apples, and etc etc. x-strike set-ups come to mind because they were the meta pick before curtains (and what i still use today)

on a smaller note, i do think >curtains are very silly to use for this. walls were used first and that was less silly, but still pretty bad. the suggestions to make items buildable/constructable from plastic (or another material readily available in engineering) should be kept in mind and considered, totally agree with that

KillianKirilenko:
> RE: special insulator object - even if I were to, say, make tesla zaps target and melt curtains, there are already half a dozen potential replacements that are just as easy to use in that scenario. And making just one-specific-object capable of blocking lgihtning would mean either officially canonizing the strat by making it easily buildable roundstart, failing to solve the specific issue I am percieving, or effectively lock it out by NOT making it easily buildable roundstart. Not even mentioning the inevitable issue slew of actually making something like that actually work with lightning strikes.

I am under no illusions that it would be easy, but I would rather attempt to seek some manner of compromise on the matter rather than an all-or-nothing approach.

That said, I also don't see how you're "locking it out" by not making it buildable at roundstart. The tesla can be safely depowered in about 20-30 minutes, so as long as you have cargo to order a replacement core generator from you can start up a new one as long as the SMES blocks are fully charged. If the substations are fully charged too, you should have _more_ than enough time to make any modifications you need to the engine, though upgrading the main SMES itself won't be so easy.

PR #15328, opened 2023-08-30
New redgate Islands map

SatinIsle:
Added a new redgate islands map. This consists of two Z-levels: An ocean with a few islands to visit, many with vore mobs on them. An underwater Z-level with lots of things to find hidden around the ocean floor, a little more sparse on vore mobs.

Added a new underwater turf. Standing on this turf feeds you CO2 instead of air, suffocating you. It is a subset of water turfs, and therefore interacts with aquatic, which also allows you to breathe on these turfs. This does mean that you can use the "dive" ability of aquatic under the water too, but it can be treated as though you are swimming up above the sea floor.

Added a "diving" subset of the ocean turf, which allows you to dive down to the turf beneath it, or to swim up from below. Uses normal move up/down verbs.

Added a bunch of underwater flora.

Added a collapsed ships mast prop.

Added a new variation of Fake_Sun that applies a new underwater weather effect and has only one option for the colour and brightness.

Added a new "vorny" variation of the great white carp. This one is very hard to escape from and digests quickly, but is stunned if you manage to do so to give you a chance to flee.

Added a new water resistant and friendly hiveless hivebot called Harry.

Added a couple of creative commons 0 sound effects for the ocean and underwater areas.


code/modules/lighting/lighting_fake_sun_vr.dm
```diff
@@ -206,3 +206,14 @@ var/static/list/fake_sunlight_zs = list()
 			"color" = "#27024B"
 		)
 	)
+
+/obj/effect/fake_sun/underwater
+	do_weather = TRUE
+	weather_visuals_icon = 'icons/effects/weather.dmi'
+	weather_visuals_icon_state = "underwater"
```
Heroman3003, line 213: does the underwater fake sun actually work on the map? It does not seem to have any outdoor turfs in underwater z.

SatinIsle: Yeah, it works absolutely fine! All of the new underwater turfs are outdoors.

Screemonster: FAKE SUN
FAKE SUN UNDER PARADISE

PR #15341, opened 2023-09-01
Makes energy ball explode upon fizzling out

Heroman3003:
This is more of an alternative proposal to previous tesla-related PR, that changes it in ways that is more in-line with staff stance on engine situation.

Makes tesla release a moderate-sized explosion upon fizzling out due to running out of energy. Only applies to 'main' ball and does not explode at all if tesla is destroyed for any reason other than running out of energy.

Tested explosion sizes, they dont appear to be too devastating, but enough to be pretty damaging around the area tesla itself is in. If it explodes that way when in containment, it will not reach engine observation room, but the setup itself is getting wrecked reliably. Of course, there's RNG involved, but I doubt the few tests I did, I always hit the lowball.

PR #15344, opened 2023-09-02
Adds a variety of new tesla coils with unique properties and upgrades

Heroman3003:
Unrelated to #15341, but inspired by an off-handed comment during discussion of it. Please judge the two PRs on independant merits, as neither relies on another, and either one of two, both or neither can end up merged.

This PR adds some actual properly-intended optimization and customization options for Tesla Engine in form of new coil types.

To obtain new coils, just grab a regular coil board and slap it with a multitool to reconfigure into another type of coil. Process is reversible, so dont worry about misclicking wrong entry in the list. Those coils require same things as regular coil to build, with only a single capacitor. Their upgrades are also exact same, just upgraded capacitor, although they do get all special effects of their own from that.

The following coil types are available:

- Standard - our good old red coil. It's only been slightly bit nerfed, to split functionale elsewhere - chains jumping off of it now lose power even if coil isn't connected to a powernet. Previously it allowed infinite chains as long as you used unconnected coils. Otherwise it functions same in terms of power production and the like.

- Relay - orange coil. As name implies, it takes away the feature of relaying lightning across from the standard coil. It does NOT produce power and only serves as chaining point. It also does have slight loss of power of 10%, but at upgrade level 3 it no longer loses power, and going beyond actually makes it amplify lightning coming through it a bit!

- Recaster - yellow coil. It serves as more long range lightning coil. It produces exactly half the power a normal coil would, and relays same amount of power to next coil in the jump chain as normal coil would. However, it has much higher range of zap jumps! From default 5, its actually...6. But it also increases by 1 per upgrade level, with 8 at level 3 and maximum of 10 at level 5! Power it collects is also amplified by capacitor upgrade level same way it would be for regular coil.

- Collector - green coil. It is a good finishing point for the chain. It collects energy much more efficiently, collecting all power in the lightning zap it recieves, and not bouncing it beyond anymore. Efficiency is 80% rate, over default coil's 50%, but this specific coil also increases with upgrades, to 90% at tier 3 and 100% full power conversion at level 5. Power it collects is also amplified by capacitor upgrade level same way it would be for regular coil.

- Prism - blue coil. The one for the books - this coil... by default is regular coil that produces half the power, like recaster, but without extra range. But as you upgrade it, you'll find it's true power. It splits one bolt into multiple with even power split among them (and one portion of power split goes towards Prism coil's own power production). Amount of bolts it outputs is equal to its upgrade level, so be careful to not upgrade this one too much. Power it collects is also amplified by capacitor upgrade level same way it would be for regular coil.

- Amplifier - purple coil. The weirdest one, for it does not produce power. It, in fact, consumes it! When a bolt jumps to it, this coil looks at its connected powernet's power supply. If there's enough power in the powernet, it consumes that amount and increases power of lightning chain by 50% of its current amount, before passing it through to next one. If there's not enough power, it passes it through still, but with loss of 25% power instead. So, how much power do we need to amplify a strike of lightning? By default, exactly 50% of its power... Which sounds terrible, but with upgrades that amount goes down to 25% at tier 3 and ~16% at tier 5. And that's a lot more value than 0-gain default, if you can make the power flow work for you!

Now, this is a lot of words, but tl;dr new coils that do weird stuff and maybe-maybe-not enable some proper engine customization options, if you manage to actually find space to fit all that.

Just dont forget, default setup is absolutely unchanged in either power production or setup complexity, and is just as viable and functional, so just because new options exist, doesnt mean you need to throw shade on those uninteresetd when flexing the extent of new things.


Screemonster:
I would like it if relays specifically _didn't_ go kaboom or attract lightning bolts when not completed, because this would allow a whole bunch of coils to be constructed off to the side of an existing engine and then for a relay to be built to connect the two networks without the risk of being exploded because the tesla decided "hm time to strike oh look there's a coil whoops it isn't finished cowabunga it is"

Runa-Dacino:
I love this one.

I also like the relay idea. It should allow engineering to quickly do a reliable engine, then iterate on it indefinitely as a way to keep themselves busy while bored without keeping the rest of the station powerless or the boat unable to move.

KillianKirilenko:
Hm. I can definitely see this putting out some *silly* amounts of power just with prisms and collectors, even before thinking about the other ones.

I think my only other comment at this point is that it could probably do with a bit more artistic variance on the coils if you can find someone to rope into doing that. But I know that ain't always easy.

Heroman3003:
> I think my only other comment at this point is that it could probably do with a bit more artistic variance on the coils if you can find someone to rope into doing that. But I know that ain't always easy.

Like with all my PRs that add icons, its devsprites and I 100% welcome someone replacing those with something more distinct. Until then, devsprite awayyyy

FartMaster69420:
> I would like it if relays specifically _didn't_ go kaboom or attract lightning bolts when not completed, because this would allow a whole bunch of coils to be constructed off to the side of an existing engine and then for a relay to be built to connect the two networks without the risk of being exploded because the tesla decided "hm time to strike oh look there's a coil whoops it isn't finished cowabunga it is"

I feel like that would make things a bit too easy, but what if the PA had a setting that would stop tesla zaps, allowing for maintenance?

Also would add a step engineers could forget and cause teslooses.

FartMaster69420:
Also would be cool to see coils that produce heat or radiation instead of power, allowing cross over with other engine types.

PR #15347, opened 2023-09-03
Decaftea

Very-Soft:
decafteapowder is cringe, all my friends eat decaftea

Fixes this
image

I didn't test this at all, someone test it please


Heroman3003:
Okay, I looked a bit mroe in-depth. It seems issue isnt with powder, but with tea itself. I suggest replacing all the recipies' decaftea with teadecaf instead and leaving powder as is.

PR #15356, opened 2023-09-05
Wall/Table Icon_Base Split (+hull plating creation)

KillianKirilenko:
Material-based constructions currently draw their wall and table states from the same filename. This has obvious issues if you use something for the former that has no matching sprites for the latter. Case in point; concrete!

So, this resolves that by adding a new `table_icon_base` var, redefining it where needed, and using it in the tables file. Practical examples of this include using it for gemstones, concrete (now you can make concrete tables!) and hull plating (tables made using hull plating also use the shiny stone appearance). In the long run this will enable more versatility with materials and such, especially if more wall/table sprites are created.

It also makes it possible to create new/replacement hull sheets at a cost of two base material sheets; 2 steel sheets = 1 steel hull sheet, 2 durasteel sheets = 1 durasteel hull sheet, and so on. Can be done up to 5 at a time.

*Also* fixes the fifty-spawner and any manual spawns of durasteel hull plating silently returning standard durasteel sheets instead.

Moderately tested; I might have missed a couple of materials that need different table states.

I mostly just did this so it'd be possible to have big solid metal counter-style tables again, useful for places like security or other worksurfaces.


Heroman3003:
One thing I'll note is that iirc you could already obtain hull sheets using autolathe

KillianKirilenko:
Ah, so you can. I knew it was acquirable *somehow*.

Heroman3003:
Eh, I'm more of a lean towards "If its emergency/field situation, you can slap a regular wall there and replace with proper hull a bit later", but that's just my opinion. Just thought to let you know that hose materials are obtainable already.

KillianKirilenko:
Might have broken a couple things with this. Drafted for now.

PR #15358, opened 2023-09-05
Make Fennec Vorny

SatinIsle:
Turns off the auto-gib on the fennec mapped to rain city. Increases it's chance to perform vore.


Very-Soft:
VOOOOOOOOOOOOOORRRRRREEEEEEEEEEEE

PR #15361, opened 2023-09-07
Makes Zorren blood actually based on copper: Iron no longer regens their blood, copper is needed

Runa-Dacino:
### What this does

Changes zorren to require copper to regain blood rather than iron supplements.
This change is documented in the health analyzer guide, and will be sent to wiki if merged

### Why we need this

Zorren are a haemocyanic species like the skrell and therefore need copper to make blood cells.
This change was requested by head loremaster PontifexMinimus

### Commit Details

[tweak(species/chemistry): Makes zorren need copper for blood](https://github.com/VOREStation/VOREStation/commit/eca94f7854869495d194d0c29136e965a0ad89d8)
- Changes iron reagent to no longer regenerate copper for zorren
- Changes copper reagent to regenerate blood for zorren
- Adds IS_ZORREN reagent_tag define and applies it to species
- Updates medical scanner guide to recommend using copper for zorren and skrell

PR #15377, opened 2023-09-10
Blood Reagents Refactor

KillianKirilenko:
Expands on #15361 by bumping the blood-regen handling up a level and comparing the reagent id to a species-level blood_reagent variable for regen purposes.

You can set your blood regen reagent under the VORE tab alongside blood colour, from a list of pre-approved types. Your chosen reagent will be shown on health analyzers and in your medical records.

Also makes slime jelly the default blood regen reagent for prometheans because why not?

### DIFF
--- a/code/datums/datacore.dm
+++ b/code/datums/datacore.dm
@@ -369,6 +369,7 @@ var/global/list/PDA_Manifest = list()
 		var/datum/data/record/M = CreateMedicalRecord(H.real_name, id, hidden)
 		M.fields["species"]		= "[H.custom_species ? "[H.custom_species] ([H.species.name])" : H.species.name]" //VOREStation Edit
 		M.fields["b_type"]		= H.b_type
+		M.fields["blood_reagent"]	= H.species.blood_reagents
 		M.fields["b_dna"]		= H.dna.unique_enzymes
 		M.fields["id_gender"]	= gender2text(H.identifying_gender)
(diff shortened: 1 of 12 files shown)


Heroman3003:
There's per-trait customization now, no? I think stuff like winged flight having toggle on default behaviour and such. Could that be used for blood type selection?

KillianKirilenko:
Mmmmmaybe. It looks like I'd need to extend the functionality of that... well, function, since it only supports booleans and colours right now.

KillianKirilenko:
I think I figured out a fix, I just need to offload the handling to the VORE tab instead, so it'll be with blood colour rather than blood type. Will test in a little bit.

KillianKirilenko:
Now it works! Just needs review for the actual allowed reagents for blood recovery.

SatinIsle:
I support this addition, it's really cool customisation, but I'd like this information to be presented on medical devices.
Mostly: Health analysers and body scanners. Possibly in medical records too.

Heroman3003:
Noble-metal-based blood is a meme but an okay meme. Sure, lesgo.

KillianKirilenko:
> Noble-metal-based blood is a meme but an okay meme. Sure, lesgo.

They were mostly just placeholders whilst I was testing the system. I was expecting a little more feedback on allowed reagents before merging!

PR #15384, opened 2023-09-11
Fruit spawners

SatinIsle:
Added fruit spawners that can be mapped in, each one spawns one of the desired fruit/veg/mushroom/whatever on initialising.

image


Very-Soft:
fwoot

PR #15405, opened 2023-09-19
Throw Notices

KillianKirilenko:
Adds visible_message prompts when you go to activate throw mode, showing nearby characters whether you're about to throw something or if you're ready to catch something.

image

Could be gamed slightly by preparing to throw with an empty hand, then switching, but I feel like that's better than no notice at all.

### DIFF
--- a/code/modules/client/preference_setup/global/01_ui.dm
+++ b/code/modules/client/preference_setup/global/01_ui.dm
@@ -18,6 +18,7 @@
 	S["tgui_large_buttons"]		>> pref.tgui_large_buttons
 	S["tgui_swapped_buttons"]	>> pref.tgui_swapped_buttons
 	S["chat_timestamp"]			>> pref.chat_timestamp
+	S["throwmode_loud"]			>> pref.throwmode_loud
 
 /datum/category_item/player_setup_item/player_global/ui/save_preferences(var/savefile/S)
 	S["UI_style"]				<< pref.UI_style
@@ -35,6 +36,7 @@
 	S["tgui_large_buttons"]		<< pref.tgui_large_buttons
 	S["tgui_swapped_buttons"]	<< pref.tgui_swapped_buttons
 	S["chat_timestamp"]			<< pref.chat_timestamp
+	S["throwmode_loud"]			<< pref.throwmode_loud
 
 /datum/category_item/player_setup_item/player_global/ui/sanitize_preferences()
 	pref.UI_style			= sanitize_inlist(pref.UI_style, all_ui_styles, initial(pref.UI_style))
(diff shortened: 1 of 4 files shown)


Screemonster:
I'unno about this one
I have a feeling that it'll get really spammy outside of combat (like throwing stuff on tables)
plus... eh, I like being able to throw stuff without warning (and having catch mode on to surprise someone who I suspect is gonna throw stuff!)
like I don't really see what "telling everyone in the room you're ready to catch stuff thrown at you" adds

KillianKirilenko:
There have been times I've wanted to throw something to someone and having confirmation that they're ready to catch it without them necessarily having to say so (and thus slowing things down) would be nice. It might also help people avoid accidentally throwing things when in hotkey mode, as the throwmode button change can be quite subtle and R is right next to T.

I could potentially redefine/extend the procs down a couple levels at the `/carbon/human/` layer and make some tie-in neutral traits that all species can take, perhaps? A 'fast reflexes' trait which hides the 'ready to catch' part, and 'fastball' which suppresses the 'preparing to throw' notice?

TankTheBirb:
> I could potentially redefine/extend the procs down a couple levels at the /carbon/human/ layer and make some tie-in neutral traits that all species can take, perhaps? A 'fast reflexes' trait which hides the 'ready to catch' part, and 'fastball' which suppresses the 'preparing to throw' notice?

I think that'd solve that concern well. Otherwise I'm ambivalent to the concept. I could do with or without it.

Skysteak:
Yeah I am in basically the same boat as Tank is on this.

Screemonster:
I'd rather they just be preferences options tbh. Really though if you want to throw something to someone say catch, and I'd rather not blow trait slots on a thing we can all do by default right now.

foopwo:
I'd rather have it so alt-clicking the throw button would show the message. No need to add to the trait bloat.

KillianKirilenko:
> I'd rather have it so alt-clicking the throw button would show the message. No need to add to the trait bloat.

Now this is an interesting suggestion. I'm not sure how feasible it is, but I might do a little digging and see what I can come up with.

I'll convert the PR to a draft for now whilst thoughts and ideas percolate.

KillianKirilenko:
> > I'd rather have it so alt-clicking the throw button would show the message. No need to add to the trait bloat.
> 
> Now this is an interesting suggestion. I'm not sure how feasible it is, but I might do a little digging and see what I can come up with.
> 
> I'll convert the PR to a draft for now whilst thoughts and ideas percolate.

Ok, I've taken a couple passes at this, but the UI click code wasn't cooperating. Rather than bang my head against that and dig deeper into UI code, I've opted to take the easier route and make it pref-based.

Screemonster:
Another option could be to make it intent-based - help intent tells people you're ready to throw/catch, any other intent does not, with the pref toggle overriding it (similar to the safe firing intent pref toggle that stops you shooting on help intent if it's on but has no effect on any other intent)

KillianKirilenko:
> Another option could be to make it intent-based - help intent tells people you're ready to throw/catch, any other intent does not, with the pref toggle overriding it (similar to the safe firing intent pref toggle that stops you shooting on help intent if it's on but has no effect on any other intent)

This is also something I considered at the time and is simple enough as changes go, but I'm going to await staff consensus/judgement before making any further changes.

JoanRisu:
Make it pref based and I'll sign off on it. It's harmless as it is and making it pref-based will address the message bloat concerns.

KillianKirilenko:
It was already pref-based, and took maybe fifteen minutes to rig up and test intent-based checking as well.

ItsSelis (approved):

PR #15406, opened 2023-09-20
Attempts implementation of char setup vore panel

Heroman3003:
Done in most disgustingly bruteforce way of moving all the related variables and such from /living level to /mob level and giving new_player the vore panel verb. This shouldnt have any effect on anything other than allowing new_player's to have bellies and such, and be able to edit them.

Saving and loading is exactly same slot-based as it is on when alive, so Save Warning ALWAYS pops up when you're editing in character setup.

The one single thing I couldn't figure out how to make work is the preview of vorebelly visuals, but that's minor enough to be an 'adjust this when alive' kinda thing and otherwise I am satisfied. All setting editing was tested to work; and all of that applied properly after I spawned in, so it works, at least.


Rick4242:
Aye yo? This is a surprise

Heroman3003:
yolo

PR #15415, opened 2023-09-23
[WIP]Buffs food, nerfs food

Very-Soft:
Once upon a time, someone thought it was a good idea to make most food just give 1/3rd as much nutrition, because protein shakes were too good, which made most food that wasn't protein shakes suck

I un-nerfed food, and just nerfed protein shakes so they give about what people expect them to now. I also undid what appears to have been an attempt at buffing food that doubled nutriment amount when it is set on the var, as with the un-nerfed nutrition conversion, that was making common food give bookoo nutrition.

At their default values, a lot of foods are much more viable now in comparison to protein shakes. Protein shakes are still one of the best vendor foods you can get, but 2 bags of jerky gives about the same amount of nutrition as 1 container of protein shake, for reference.

On top of this, I added a bunch more positive traits for food preference! These new traits are in the positive field because they give a big boost to nutrition gained when ingesting certain kinds of allergens. Herbivores for example gain extra nutrition by eating vegitables, and fruits. Carnivores gain extra nutrition by eating meat, and Omnivores gain extra nutrition from eating meat or vegitables. These traits all have 0 cost, but you can only pick one at a time, because they're pretty powerful. The hope is that when using one of these traits, you can make your character's favorite kind of food be a better choice than something like a protein shake. You can even gain nutrition from drinking coffee now!

Ultimately, I just think it would be cool if there were more different kinds of food that is actually worth seeking out, and I think it would be extra cool if, what that food IS, is different from person to person!

PR #15417, opened 2023-09-23
Soup's On!

KillianKirilenko:
Something that was vaguely bothering me in that way that small things do is that soup is a liquid, but we can't bring a thermos of soup to work!

So I fixed that. I've;
- Added eight drink-type soup reagents (to go with existing snack-type soups) plus a base generic soup. These soups are all selectable for the vacuum flask reagents list.
- Converted most of the snack-type soups to have the new soup reagents added during creation.
- Also cleaned up a duplicate recipe for creating chicken noodle soup in the microwave.

Might need some tweaking in light of #15415, possibly. I'm not sure yet.

### DIFF
--- a/code/modules/food/food/snacks.dm
+++ b/code/modules/food/food/snacks.dm
@@ -2476,13 +2476,12 @@
 	trash = /obj/item/trash/snack_bowl
 	filling_color = "#AFC4B5"
 	center_of_mass = list("x"=16, "y"=8)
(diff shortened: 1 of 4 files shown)


Heroman3003:
Since it shouldn't cause any immedeate issues with relation to the other PR, gonna merge this now while its a WIP to avoid further conflicts down the road.

PR #15418, opened 2023-09-24
Removes unecessary tooling and re-adds necessary tooling to research borgs

TankTheBirb:
Removes the Surgical Scalpel and Circular Saw. Adds the Hemostat and Surgical Drill.

Why? Because the research borg should not be surgically removing brains and they don't even have all they need to do it. But what they should be doing is synth repairs and NIF installs. And that requires just two surgical tools. They used to be able to do this but I figure this was an oversight in the refactor.

Also they can't start surgeries without a scalpel so organics are safe from malpractice. If these two tools were needed for something else, do let me know and I'll change the commit to be just the addition.


Heroman3003:
I am 99% sure they cannot install NIF even with these tools. I recommend testing that but iirc any implant procedure is currently incompatible with grippers.

TankTheBirb:
> I am 99% sure they cannot install NIF even with these tools. I recommend testing that but iirc any implant procedure is currently incompatible with grippers.

I fixed this! Unless it re-broke again. I'll test it.

TankTheBirb:
image
Yep! Successfully installed with the robotics organ gripper.

PR #15419, opened 2023-09-24
Armour Examine Tweaks

KillianKirilenko:
I had a thought the other day of adding slowdown/mobility adjustment to armour examines, and ended up sliding down a (small) rabbithole of refactoring some other bits in the process. Oops.

This PR:
- Adds clothing slowdown to examine.
- Adjusts some existing examine data to cut down on the number of lines, and provide more accurate/useful values (for example, atmos voidsuits weren't showing as protecting from extremely high temperatures when examined, as their temp protection *exceeds* 30,000).

### DIFF
--- a/code/modules/examine/descriptions/armor.dm
+++ b/code/modules/examine/descriptions/armor.dm
@@ -22,7 +22,7 @@
 
 /obj/item/clothing/proc/describe_slowdown()
 	switch(slowdown)
-		if(-1)
+		if(-INFINITY to -0.1)
 			return "It looks like it might actually make you faster!"
 		if(0 || null)
 			return "It doesn't look like it'll impede your mobility."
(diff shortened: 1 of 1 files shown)


code/modules/examine/descriptions/armor.dm
```diff
@@ -20,6 +20,28 @@
 			if(100)
 				return "You would be immune to [descriptive_attack_type] if you wore this."
 
+/obj/item/clothing/proc/describe_slowdown()
+	switch(slowdown)
+		if(-1)
+			return "It looks like it might actually make you faster!"
```
Heroman3003, line 26: Couldnt you do something like (-INFINITY to 0) after the check for 0, in order to have it cover all speed up levels, not just specifically -1?

KillianKirilenko: Done, although there's only one other case where armour speeds you up and that's the commonwealth medical hardsuit (-0.5).

Heroman3003: Future proofing is always beneficial.

PR #15425, opened 2023-09-28
Borg belly overlays can be supported by both sleepers and vore panel bellies

ReoDaProtovali:
### What This does
On the surface, this allows cyborgs to have their vorebelly overlay be controlled by both the sleeper and vore panel vorgans.
Additionally, this system allows extension of preexisting systems by allowing cyborgs to have multi-stage bellies that could have multiple size stages of "fullness" (such as a small, medium and large belly that changes automatically based on how full the borg is)

However, the system requires a very noteworthy change to the way the borg's icon-states are defined. Previously, borg's belly icon states were named in the fashion of "`[base icon state]-sleeper`", or "`[base icon state]-sleeper_[r/g]`" for medical borgs who had colored lights on their vore belly overlays. The new system uses the naming scheme of "`[base icon state]-sleeper-[size]`" or "`[base icon state]-sleeper-[size]-[r/g]`" for borgs who have changing-light bellies (more on this in the next section). 

As an example, the old naming scheme would look something like this:
`"miningBorg-sleeper"` or `"medicalBorg-sleeper_g"`
and the new system will look like
`"miningBorg-sleeper-1"` or `"medicalBorg-sleeper-1-g"`

The system has also been changed so that any borg can support light-colored overlays (as they were instead given a snowflake proc to determine if they should instead use a completely different system), To use this system, in the borg's sprite_datum definition, set the variable "**has_sleeper_light_indicator**" to true. Furthermore to have a borg use more than 1 belly size sprite, set the "**max_belly_size"** variable to the number of belly sprites you have.

### Why we need this

Honestly, the new system is to help support a downstream change which is even more complex than this system (yet somehow also simpler) without causing a large rift in how the borg systems work and require manual conversion of newly added borg sprites upstream that end up downstream as well. Ultimately, this system both helps downstream while also giving upstream some benefit from the naming change that would otherwise not make sense to add to this codebase


Heroman3003:
Linter is not likey some things.

code/modules/mob/living/silicon/robot/sprites/_sprite_datum.dm
```diff
@@ -88,7 +112,7 @@
 	has_dead_sprite_overlay = TRUE
 	has_custom_equipment_sprites = TRUE
 	pixel_x = -16
-
+/* //Does not need to be dogborg-only, letting all borgs use these -Reo
```
Heroman3003, line 115: All borgs don't actually HAVE resting sprite styles for alternative options, is why this specific selection is in there, they don't even have default sprite-rest sprites.

code/modules/mob/living/silicon/robot/sprites/_sprite_datum.dm
```diff
@@ -12,6 +12,8 @@
 	var/has_custom_open_sprites = FALSE
 	var/has_vore_belly_sprites = FALSE
 	var/has_vore_belly_resting_sprites = FALSE
+	var/has_sleeper_light_indicator = FALSE //Moved here because there's no reason lights should be limited to just medical borgs. Or redefined every time they ARE used.
```
Heroman3003, line 15: Initially I wanted to get rid of sleeper light indicators entirely. Of borgs that use them, it's only medical ones and even then only SOME of them, not all. I am not against this, but I am questioning how many borgs will actually use it.

ReoDaProtovali: I was considering making the light indicator more open, and possibly also allow other colors for other modes, but I feel leaving it as an option is a good thing, and making it be handled by snowflake procs just makes it ick and harder to debug (it certainly was when initially trying to make this PR for chomp) I definitely want to make lights vore-belly based, however

code/modules/mob/living/silicon/robot/sprites/_sprite_datum.dm
ReoDaProtovali: On this... They might not have resting sprites, but if the intent was to merg dogborgs with normal borgs, I think requiring all forward going borgs to have to be dogborg subtypes to use resting sprites is kinda ick and against the point of merging the two, thus this will allow borgs going forward to use what dogborgs have and not have to be dogborg subtypes. Arguably this could also be used to make dogborgs not a subtype outside of their specialized, dog-like differences (like having jaws for crowbars)

Heroman3003: Basically, dogborg subtype was made entirely for purpose of "this is set of features that old dogborg modules had and is common among all those sprites". New sprites can have that individually. Regardless, my concern for this portion is that it will break sprites for regular borgs if they rest for whatever reason, hence why I say this should stay as is. Plus you can see with tallborgs and such that you can just make 'individualized', per-sprite-type procs for selecting the resting sprite so it wont be necessary to be subtype of dogborg to do that.

ReoDaProtovali: the rest code checks for if they have resting sprites, which by default is False and skips that check. Unless I missed something, but I'm pretty sure I had this issue and fixed it before. I'll test later

Heroman3003:
One more thing I have uncertainty about is that in current state, 'max belly size' is pretty much unused feature since no sprite actually uses more than 1 size, but that's not too significant.

PR #15428, opened 2023-09-30
Refactors tool typechecks, refactors transforming tools, makes Altevian wrench into one

Heroman3003:
Title. God this shouldn't have been that much work, but there it is. Should fix a lot of jank regarding jaws of life and power drill longer term, but also make adding tools like that more.

Also rat wrench


Heroman3003:
Quickmerging primarily to prevent conflicts. This is WAY too many diffs to avoid them.

PR #15451, opened 2023-10-11
They're out there man, I seen em'

Very-Soft:
Updates alien whitelist

https://forum.vore-station.net/viewtopic.php?f=46&t=2375


config/alienwhitelist.txt
```diff
@@ -40,6 +40,7 @@ digi5 - Black-Eyed Shadekin
 digitalsquirrel95 - Black-Eyed Shadekin
 digitalsquirrel95 - Protean
 draycu - Vox
+durandal - Teppi
```
ItsSelis, line 43: The ckey seems to be incorrect (at least from their modmail post) 

```suggestion
durandaal - Teppi
```

Issue #15262, opened 2023-08-15
[GENERAL] Subtles with underlines don't appear if all non-localchat and non-NIF messages are filtered out


KillianKirilenko:
#### Brief description of the bug
As subject. Any subtle that includes underlines won't appear in a tab that has all messages but Local Chat and NIF marked to be filtered out. Underlines also seem to be a bit inconsistent in subtles in general.

#### What you expected to happen
Subtles with underlines to appear on my IC tab along with normal subtles.

#### What actually happened
They didn't.

#### Steps to reproduce
- Make a tab where everything *except* Local Chat and NIF Messages are filtered out
- Send or receive a subtle that includes underlines
- Marvel at how it does not appear in your filtered out tab

KillianKirilenko:
As an extension, formatting with underlines may just be kind of wonky in general. Some more testing might be needed.

PR #15454, opened 2023-10-11
Fixes timestamp sorting incompatibility.

Kashargul:
-> Timestamps have been simple added as prefix in front of the entire message, having the classifying `<span class="...">` behind it what prevented message filtering. So instead of adding it in front, we are inserting it with a regex replace after the first `<span class="...">`

-> Underlining breaks formatting. This issue happens as _ is used as a character in class labels. It never should be used in labels as it's a marking character being parsed. Using it in a label causes the label to be partially parsed and therefore the message breaks. I've removed the underline in "emote_subtle" to at least have that working again. But likely all labels should be reworked to no longer use underlines.

! Known issue: Absorb messages are not having timestamps. The way absorb messages are implemented prevents the current "client.prefs.chat_timestamp" from being used as it would rely to the speaking client to have it enabled. I haven't looked enough into to code, but if there's a way, it might be best to move the entire timestamping to code/modules/vchat/vchat_client.dm using a regex to insert them behind the classifying span.


KillianKirilenko:
Nice work getting to the bottom of these two.

As a sidenote I filed the latter as an actual report (#15262 specifically) so you could add "Fixes #15262" to your message somewhere to have it autoclose the issue when merged.

PR #15456, opened 2023-10-13
Vore Panel Health Check

SatinIsle:
Added a new option to humanoid mobs in vore belly contents: Health Check.

Using this on a mob in your belly returns a very simple readout of their health as a percentage. It also gives notice when they are weakened, paralysed, blinded or sleeping, so that the pred can know whether or not they'll be able to emote/hear emotes.

image
image

### DIFF
--- a/code/modules/vore/eating/vorepanel_vr.dm
+++ b/code/modules/vore/eating/vorepanel_vr.dm
@@ -876,17 +876,26 @@ var/global/list/belly_colorable_only_fullscreens = list("a_synth_flesh_mono",
 		if("Health Check")
 			var/mob/living/carbon/human/H = target
 			var/target_health = round((H.health/H.getMaxHealth())*100)
+			var/condition
+			var/condition_consequences
 			to_chat(usr, "<span class= 'warning'>\The [target] is at [target_health]% health.</span>")
-			if(H.weakened)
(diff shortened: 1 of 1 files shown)


code/modules/vore/eating/vorepanel_vr.dm
```diff
@@ -872,12 +873,27 @@ var/global/list/belly_colorable_only_fullscreens = list("a_synth_flesh_mono",
 					b.absorb_living(ourtarget)
 				if("Cancel")
 					return
+		if("Health Check")
+			var/mob/living/carbon/human/H = target
+			var/target_health = round((H.health/H.getMaxHealth())*100)
+			to_chat(usr, "<span class= 'warning'>\The [target] is at [target_health]% health.</span>")
+			if(H.weakened)
+				to_chat(usr, "<span class= 'warning'>\The [target] is currently weakened, this does not stop them from hearing emotes or emoting themselves.</span>")
+			if(H.blinded && H.paralysis)
```
Heroman3003, line 882: A suggestion I have is to append the messages in something like.
The [target] is currently [list of effects]. This stops them from [list of consequences].

SatinIsle: Had a go at doing this. Got rid of the weakened one seeing as it doesn't affect their ability to emote anyway.

PR #15457, opened 2023-10-13
Jaffa cakes

SatinIsle:
Added Jaffa Cakes. They can be either bought from Sol Snacks vendors in boxes of 12 or ordered from cargo for 10 boxes. They can also be made in the oven, 6 at a time, with 1 orange, 10 cocoa and 15 cake batter.


KillianKirilenko:
But are they cakes or biscuits?

Heroman3003:
That much cocoa might dillute the flavor of cooked version a lot. I suggest you test that out.

SatinIsle:
> That much cocoa might dillute the flavor of cooked version a lot. I suggest you test that out.

Nah, just checked it out and there is no difference between the two
image

PR #15464, opened 2023-10-19
Digitigrade legs port

KillianKirilenko:
Upstreaming digitigrade legs from chomp, finally.
image

I apologize in advance to downstream for any issues this causes.


code/modules/clothing/clothing.dm
```diff
@@ -1059,3 +1062,61 @@
...
+		var/mob/living/carbon/human/H = user
+
+		// if digitigrade-use flag is set
+		if(H.digitigrade)
+
+			// figure out what slot we care about
+			var/update_icon_define_digi = null
+
+			if(istype(src, /obj/item/clothing/shoes))
+				update_icon_define_digi = "icons/inventory/feet/mob_digi.dmi"
```
Heroman3003, line 1077: Ideally, that could be a variable on the clothing obj itself, rather than determined in the proc tbh.

KillianKirilenko: It does strike me as odd that they didn't do it that way originally now that I look at it, but unfortunately untangling it myself might take a while. I'm an amateurish coder at best.

Heroman3003: Well that specific part shouldn't be hard to fix, I imagine, so I'll leave it up for a bit.

KillianKirilenko: Took a shot at it and it seems to compile and run with no runtimes, and changing clothes between planti/digi characters has no issues in my local server either.

Heroman3003 (approved):

PR #15468, opened 2023-10-22
Avii's Borg Event Map

Wulfestwulf:
As the title says. I discussed it with them, they said the map works perfectly for their event and wanted me to add it to our code that way they and other gamemasters can use it in their events!

### DIFF
--- a/maps/submaps/admin_use_vr/avii_eventMapv2.dmm
+++ b/maps/submaps/admin_use_vr/avii_eventMapv2.dmm
@@ -1,0 +1,29497 @@
+//MAP CONVERTED BY dmm2tgm.py THIS HEADER COMMENT PREVENTS RECONVERSION, DO NOT REMOVE
+"ab" = (
+/obj/machinery/atmospherics/pipe/simple/hidden/supply{
+	dir = 6
+	},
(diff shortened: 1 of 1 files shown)


Runa-Dacino (changes requested):
Mapdiffbot is throwing an error. I do not know what it means, but I'm making this review so you are notified of it:

```
Caused by:
0: Loading added maps
1: 8814:9: error: cannot reference variable "s" in this context
- 8812:2: within key: "VC"
```

KillianKirilenko:
Ah, I see you've refurbished the minitest map.

As a heads up, the lead-wall backstop will only stop PA waves if it's on a normal setting. If someone hacks the PA and overcharges it, they'll have enough energy to pass through and thus wash over the area on the lower left.

AviiCinders:
> Ah, I see you've refurbished the minitest map.
> 
> As a heads up, the lead-wall backstop will only stop PA waves if it's on a normal setting. If someone hacks the PA and overcharges it, they'll have enough energy to pass through and thus wash over the area on the lower left.

At which point I will obliterate them. ANYWAYS THANK YOU HYPO YOUA RE GOOD

PR #15491, opened 2023-11-02
Adds digestion sound to the process option

SatinIsle:
Added a digestion sound that plays when a predator uses the "process > digest" option on their prey.

Ported from: [CHOMPStation2/CHOMPStation2#7158](https://github.com/CHOMPStation2/CHOMPStation2/pull/7158)


Heroman3003:
This should probably first check if belly has fancy sounds or not before sending the sound, similar to how regular digestion does it

PR #15514, opened 2023-11-12
Fixes telecomms having too much access

Heroman3003:
Sets main telecomm entrance door on tether access to match that needed to disable turrets. For obvious reasons.


Skysteak:
Does this prevent borgs from being able to open up Tcomms? I assume not, and it's just simply making it match the permissions of SDs TCs?

PR #15521, opened 2023-11-17
Jobfuscate!

KillianKirilenko:
For those concerned about privacy/ckey-stalking; adds a pref toggle to hide your chosen job during ready-up; if it's enabled, all that'll be visible is "Ready!".

Make sure to enable it *before* readying up, though!


SatinIsle:
Good idea imo.

Heroman3003:
This might be me, but I think something the other way around would be better for privacy. Where it doesn't tell you who readied up or didnt, just how many people readied upo for each job. But thats just my opinion.

Screemonster:
I like Hero's idea. Some people flick their medium/high priority jobs around depending on what other people have readied up as.

KillianKirilenko:
Now with added ability to have it both ways. Obfuscate your ckey! Or your job! Or both! Or neither! Likely a cleaner way of doing this but I'm lazy.

The obfuscation only applies in the ready-up list, it doesn't hide your key from staff or in any other instance.

ItsSelis:
If it is the theoretical possibility of having more privacy options, it might be neat to use a kind of bitflag instead of multiple variables instead.

Storage wise it would be a bit more efficient and would allow something like this:
FLAG_PRIVACY_NONE = 0
FLAG_PRIVACY_CKEY = 1
FLAG_PRIVACY_JOB = 2
[...]
FLAG_PRIVACY_M = n
[...]
FLAG_PRIVACY_EVERYTHING = ...

The tgui checkbox input can be probably nicely utilized to check/uncheck multiple of these at once.

KillianKirilenko:
TGUI checkboxes would be great, if that particular interface was TGUI. It's not; it's just HTML with very similar stylings. Making it a TGUI popup *just* for the privacy checkboxes also seems wasteful, and unfriendly for our users who are unable or unwilling to use TGUI for whatever reason since it would lock them out of said privacy features.

ItsSelis:
Nono, it literally is a premade thing like tgui_text_input or tgui_number_input

KillianKirilenko:
Eh, maybe, but there's really only two options to hide right now, so it might be a bit overkill/out of scope?

Perhaps if someone was writing a more comprehensive privacy management interface that included settings like the non-persistent hiding your ckey in dchat (so that it just says "Ghost of <charname>") and other details, I could see a case for using it. That person will not be me though!

As far as this PR goes really what I'd rather do is have four buttons on one row going; None, Key, Job, Both. Might try and figure that out at some point maybe, if I can wrap my head around the clunky html stuff this UI has going.

ItsSelis:
Yeah this is exactly what such a bitflag would be for; if someone decided to add another privacy thing to it. Its probably more about doing work now or the other person doing the work later on.

Generally its rather also about reducing the amount of "wasted space" by reducing the amount of unnecessary vars that is pasted into the player savefile thingy. And well, if the variable name of the savefile gets deleted or reduced to "privacy_flags" or something settings might more or less reset in a way (aka the player has to turn it back on).

I mean, I am not going to stop you if getting this merged or anything. I do like the feature!
Just wanted to give a recommendation about the bitflag thingies and that the (unfortunately still unused iirc) tgui_input_checkboxes exist.

KillianKirilenko:
Well, if you have any examples of how to implement it in this case I'd be all ears. I can handle the bitflag and vars side of things, but without a practical example to reverse-engineer for the actual checkboxes in the menu things are a bit trickier.

Again, the key consideration here for the moment is "no popups that 100% rely on TGUI", because I *do not* want these options locked away from people who can't use TGUI. And as far as I can tell, `tgui_input_checkboxes` creates a popup that does exactly that judging by the in-file comments for it.

ItsSelis:
afaik you can give it a list of the things that could be selected. I am not sure though how exactly it handles the list. My current guess is that it would work like this:

```dm
[...]
// Preparing list items
my_list["Test Selection"] = FLAG_PRIVACY_TEST
[...]
// Show list & get result
var/list/result = tgui_input_checkboxes(usr, "Please select your privacy flags", "Privacy Flags", my_list, min_checked = 0)

// Handles result
[...]
[...]
```

And the result returned is the original list but **only** with the items that are actually selected.



Additionally, yes, the checkboxes do support default byond like any other input thingy:
image

PR #15527, opened 2023-11-19
Makes absorbs escapable (and other tweaks)

Screemonster:
Firstly, escape timers have been cut in half across the board so you don't have to wait for ages after a successful escape roll.

Secondly, adds a chance of escaping if absorbed into a belly that is possible to escape from.
By default this chance is 25% but it is checked *after* waiting out an otherwise successful escape attempt - in effect, this changes absorption from "game over" to "75% chance of the predator dragging you back in on the spot when you'd otherwise escape" - still means that if you get pudged you've got a good chance of being stuck there for a while.

Thirdly, a minor tweak to the messages when resisting in a belly - it now only gives you the fail message if you actually fail, rather than showing the struggle message (most of which imply failure) *then* the escape message on a successful roll.

Fourthly, messes about with gaslamps to give them some somewhat different probabilities since they're the only dedicated absorby mob we actually have.

### DIFF
--- a/code/modules/mob/living/simple_mob/subtypes/animal/space/gaslamp_vr.dm
+++ b/code/modules/mob/living/simple_mob/subtypes/animal/space/gaslamp_vr.dm
@@ -81,7 +81,7 @@ TODO: Make them light up and heat the air when exposed to oxygen.
 	vore_ignores_undigestable = 0 // they absorb rather than digest, you're going in either way
 	vore_default_mode = DM_HOLD
 	vore_digest_chance = 0			// Chance to switch to digest mode if resisted
-	vore_absorb_chance = 20			// BECOME A PART OF ME.
+	vore_absorb_chance = 100		// Will always start to absorb if the prey fails its generous escape chance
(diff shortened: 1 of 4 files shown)


Screemonster:
Out-of-scope things that could be added but are beyond my ability: adding custom _success_ messages to a struggle (instead of just having "you start to climb out") and adding those as options on the vore panel, and likewise allowing the vore panel to edit the escapechance_absorbed var (do any players actually have escapable turned on?)

Heroman3003:
One thing I notice is that players cant actually edit the absorbed escapechance on their bellies.

Screemonster:
> One thing I notice is that players cant actually edit the absorbed escapechance on their bellies.

yeah, that's tgui stuff and someone else is working on that one for when this is merged

Kashargul:
Already working on a tgui overhaul of the vorepanel to allow the chance to be edited as well as the entire struggle message section to be more customizable.

Kashargul:
Maybe change it to 0% by default for now to have the functionality disabled, once it's merged in, I'll make a draft for the tgui parts and get to proper testing on it to finalize it soon.

PR #15530, opened 2023-11-20
tgchat part 1

ItsSelis:
See #14625 for details

This PR adds the javascript base and various defines.
This should have no effect in-game.

This **does not** touch any chat code.


Heroman3003:
Doesn't touch anything actually active, so seems fine.

.gitignore
```diff
@@ -11,10 +11,12 @@ vchat.db*
...
 *.pyc
 *.pid
-data
-data/
 cfg/
 
+#Ignore everything in datafolder and subdirectories
+/data/**/*
+/tmp/**/*
+
```
Leshana, line 19: Removing the `data` line will cause the `data` symlink to be deleted.   The lines being added in as replacements only ignore the _contents_ of data, not data itself.

PR #15532, opened 2023-11-22
Paralysis Refactor

KillianKirilenko:
This started as a "hey, there's no drugs that *just* apply paralyze, I should change that" and became "wait, why the fuck is Paralyze just Sleeping 2". So I set out on a QUEST to refactor paralysis into a more unique condition; being paralyzed means you can't move or do anything, but you can still see and hear the world around you! So this could be used either for simplemobs (esp. if they have some 'drag back to nest' AI), maintpreds, or whatever else!

A fairly extensive work in progress as there are still a few things I need to iron out and test to be sure there's no weird interactions.

- [x] Make paralysis, well, paralyze you. No movement or say/emote allowed.
- [ ] Make sure you can still hear/see messages. Partially done; hearing works, but visible_messages are being parsed as if blind for some reason.
- [ ] Make sure paralysis properly stops buckling and most item usage.
- [ ] Fix any other lingering oddities.

Would greatly appreciate a couple more sets of eyes on this to help me track down and fix some of the leftover issues.

### DIFF
--- a/code/ZAS/Airflow.dm
+++ b/code/ZAS/Airflow.dm
@@ -4,7 +4,7 @@ Contains helper procs for airflow, handled in /connection_group.
 
 /mob/var/tmp/last_airflow_stun = 0
 /mob/proc/airflow_stun()
-	if(stat == 2)
+	if(stat == DEAD)
 		return 0
 	if(last_airflow_stun > world.time - vsc.airflow_stun_cooldown)	return 0
 
--- a/code/__defines/mobs.dm
+++ b/code/__defines/mobs.dm
@@ -1,7 +1,8 @@
 // /mob/var/stat things.
 #define CONSCIOUS   0
-#define UNCONSCIOUS 1
-#define DEAD        2
+#define PARALYZED	1
+#define UNCONSCIOUS 2
+#define DEAD        3
 
 // Bitflags defining which status effects could be or are inflicted on a mob.
 #define CANSTUN     0x1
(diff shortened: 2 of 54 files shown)


Heroman3003:
I am fairly sure condition doing exactly what you describe already exists. Paralyze, weaken, stun and sleep are four different conditions and while I will agree that names are not at all intuitive compared to one another, I do believe they cover the basics when all put together.

KillianKirilenko:
> I am fairly sure condition doing exactly what you describe already exists. Paralyze, weaken, stun and sleep are four different conditions and while I will agree that names are not at all intuitive compared to one another, I do believe they cover the basics when all put together.

I can assure you this is not the case from my initial testing. I made the reagent first (note the seperate "paralyzant" pull) and was baffled when it put me to sleep as if I'd just been dosed with chloral.

ReoDaProtovali:
just gonna pitch in my thoughts and say it's probably not a good thing to rework a mechanic like this... it'll definitely fuck up somewhere down the line, be it missed stuff, or simply downstream

Heroman3003:
> > I am fairly sure condition doing exactly what you describe already exists. Paralyze, weaken, stun and sleep are four different conditions and while I will agree that names are not at all intuitive compared to one another, I do believe they cover the basics when all put together.
> 
> I can assure you this is not the case from my initial testing. I made the reagent first (note the seperate "paralyzant" pull) and was baffled when it put me to sleep as if I'd just been dosed with chloral.

What I meant is, that your proposed function of paralyze, I believe, exists under another name already. I might be mistaken, as the variety of different 'stuns' in the game and their un-intuitiveness is staggering (surprised 'stagger' isnt one of them tbh), but just mentioning in context of rework for something to check,

PR #15533, opened 2023-11-22
ignore devoured in virus selection lists

Kashargul:
-> Devoured are ignored in greater and lesser virus selection
-> Devoured can no longer be event & admin infected

No, you can not cure diseases by devouring / absorbing people. The virus will not magically disappear to cheat virology.


Heroman3003:
Events are unused, but honestly, Id prefer it ignore anyone in bellies, absorbed or otherwise.

Kashargul:
That can be adjusted, will look into it tomorrow then.

PR #15536, opened 2023-11-24
Removes sleep toxin from tofurkey

SatinIsle:
Removed sleep toxin from tofurkey.


Heroman3003:
Vegans DESERVE it

AviiCinders:
Can we replace it with regular toxin?

PR #15537, opened 2023-11-25
Remove Player Polls Button

KillianKirilenko:
BEGONE, TH--

_uhhhh_, I mean, removes the "Show Player Polls" button, because this feature hasn't been used in years, possibly not even since it was ever added to the codebase originally.


Heroman3003:
BEGONE THOT

PR #15538, opened 2023-11-26
Admin rejuvenate now restores robot modules as well

Kashargul:
-> Rejuvenate now restores modules entirely
-> If the cell is destroyed, the new type can be selected during rejuvenation
-> rejuvenate restores max charge


Kashargul:
Will get integrated into #15539 as it requires some overhaul and changes from there if battery selection will be dynamic

PR #15539, opened 2023-11-26
Admin add/remove borg modules, radio channels and components verb + rejuvenate fully restores borgs

Kashargul:
New admin verb "Modify Robot Module", accessible by right clicking living. Only works on robots who have already selected a module.

-> Adds the option for admins to quickly add or remove modules, radio channels and components from robots. This should make event handling with borgs much easier.
-> Minor sprite fix.
-> Memory handling. We also should delete the emag item list explicitly in case a module was not emagged and therefore had the objects not linked to the main module list.
-> Rejuvenate now fully restores borgs


Kashargul:
Will see to add the option to add radio channels / remove them as well

Kashargul:
Todo: Additional options
for some more specific admin handling.... really the last option to add now.

- [ ] rework borg batteries to use path instead of hard coding (also affects the rejuvenate change. Move this to this PR and rework. #15538 to allow easier access
- [ ] Add option to remove modules
- [ ] Add option to add/replace modules

PR #15544, opened 2023-11-28
nif duplication prevention

Kashargul:
When we are creating organs through the create_organ call, we should first do the organ part, then check if there is already a nif installed and in case there is, we do not want to add another one.

If someone rather wants to have the old one deleted, it's just to remove it from both lists in case it exists and qdel it. Then continue with adding a new one.

H.contents["head"].implants
H.contents["head"].contents

Someone with more experience with NIFs should test this as well. I mostly did a few quick tests.


Kashargul:
Not going to work like this, will have to do some more troubleshooting.

PR #15545, opened 2023-11-28
[WIP] Protean Reconstitutor

KillianKirilenko:
Finally, a crude process by which exploded proteans can be revived in-round! Take the brain, the orchestrator, a load of nanopaste, and optionally the refactory, then shove them all in a vat! ~~Stir vigorously! Wait, what do you mean it's stirring itself?~~

If you need to you can fish parts back out of the muck using a wrench.

TODO:
- [x] Get the basic functionality down (mostly).
- [ ] Tighten up and test the actual copy/spawn proc more thoroughly. This needs a lot of work.
- [ ] Replace the sprites with something that isn't 5-second coder art.
- [ ] Add reconstructors to robotics on the various maps.


Heroman3003:
I'd suggest rather than adding them to maps, make them buildable using default materials and minimal tech levels, is the only suggestion; while also making them single-use. It seems to just kinda make sense to me, and it'd make them not an item that's possible to lose.

KillianKirilenko:
> I'd suggest rather than adding them to maps, make them buildable using default materials and minimal tech levels, is the only suggestion; while also making them single-use. It seems to just kinda make sense to me, and it'd make them not an item that's possible to lose.

Since they're a subtype of machinery they won't be too hard to make constructable; once I have the core respawning functionality dialled in and working properly, I'll be adding upgradability and construction steps, to make them comparable to the other resleeving/respawning equipment.

KillianKirilenko:
Steadily making progress. This is messy as hell and there's probably cleaner ways to do some of this stuff, but it's getting there.

Last thing I need to check is autoloading prefs and a couple other things. Plus some more thorough tests.

PR #15548, opened 2023-11-29
Nif reju fix attempt

### LINKED ISSUE #15547: [GENERAL]  Rejuvenate breaks NIFs
When Rejuvenate is used on people with NIFs, the item gets duplicated and no longer initializes. This issue is to keep track of the current findings for now.

The NIF shouldn't get duplicated and become unusable.

#### What actually happened
The nif ends up duplicated within the contents and implants of the brain list.

#### Steps to reproduce
- Spawn in with a NIF
- Use Rejuvenate (admin ability)
- The NIF duplicates and stops working

Locations of the NIF inside the mob:
```
H.contents["head"].implants
H.contents["head"].contents
```

#### Current progress.
The duplicating happens within the execution of the `revive()` within `\code\modules\mob\living\carbon\human\human.dm`
Going through the list stack, the first NIF instance is added on the `client.prefs.copy_to` proc

The following section adds another one to the content / implant list:
```
	// Vorestation Addition - reapply markings/appearance from prefs for player mobs
	if(client) //just to be sure
		client.prefs.copy_to(src)
		if(dna)
			dna.ResetUIFrom(src)
			sync_organ_dna()
	// end vorestation addition
```
The duplicated adding happens within the `client.prefs.copy_to(src)` call

Neither seems to get initiated properly

=>  A new head is generated
=> There's two times the nif added
=> It does not get initialized

A working workaround consists out of removing the NIF reference from within the head and readding it back in later. Changes are marked with //Workaround.

```
/mob/living/carbon/human/revive()
//Workaround Start
	if(nif)
		var/obj/item/organ/external/head/H = get_organ(BP_HEAD)
		for(var/find_nif in H.contents)
			if(istype(find_nif, /obj/item/device/nif))
				H.contents.Remove(find_nif)
				H.implants.Remove(find_nif)
//Workaround End
	if(should_have_organ(O_HEART))
		vessel.add_reagent("blood",species.blood_volume-vessel.total_volume)
		fixblood()

	species.create_organs(src) // Reset our organs/limbs.
	restore_all_organs()       // Reapply robotics/amputated status from preferences.

	if(!client || !key) //Don't boot out anyone already in the mob.
		for (var/obj/item/organ/internal/brain/H in all_brain_organs)
			if(H.brainmob)
				if(H.brainmob.real_name == src.real_name)
					if(H.brainmob.mind)
						H.brainmob.mind.transfer_to(src)
						qdel(H)

	// Vorestation Addition - reapply markings/appearance from prefs for player mobs
	if(client) //just to be sure
		client.prefs.copy_to(src)
		if(dna)
			dna.ResetUIFrom(src)
			sync_organ_dna()
	// end vorestation addition

	for (var/ID in virus2)
		var/datum/disease2/disease/V = virus2[ID]
		V.cure(src)

	losebreath = 0

	..()
//Workaround Start
/mob/living/carbon/human/rejuvenate()
	..()
	if(nif)
		var/obj/item/organ/external/head/H = get_organ(BP_HEAD)
		for(var/find_nif in H.contents)
			if(istype(find_nif, /obj/item/device/nif))
				H.contents.Remove(find_nif)
				H.implants.Remove(find_nif)
		H.contents.Add(nif)
		H.implants.Add(nif)
//Workaround End
```

Kashargul:
This should prevent the duplication of NIFs during rejuvenation as well as allow it to properly re-add and reinitialize. Needs to be tested though if it doesn't cause any other issues.

For more details:
Fixes  #15547

### DIFF
--- a/code/modules/mob/living/carbon/human/species/species_vr.dm
+++ b/code/modules/mob/living/carbon/human/species/species_vr.dm
@@ -51,17 +51,17 @@
 
 /datum/species/create_organs(var/mob/living/carbon/human/H)
 	if(H.nif)
-		var/type = H.nif.type
+		/*var/type = H.nif.type
 		var/durability = H.nif.durability
 		var/list/nifsofts = H.nif.nifsofts
-		var/list/nif_savedata = H.nif.save_data.Copy()
+		var/list/nif_savedata = H.nif.save_data.Copy()*/
 		..()
-
-		H.nif = null
-		var/obj/item/device/nif/nif = new type(H,durability,nif_savedata)
-		nif.nifsofts = nifsofts
+		H.nif = null //A previous call during the rejuvenation path deleted it, so we no longer should have it here
+		/*var/obj/item/device/nif/nif = new type(H,durability,nif_savedata)
+		nif.nifsofts = nifsofts*/
 	else
 		..()
+
 /datum/species/proc/produceCopy(var/list/traits, var/mob/living/carbon/human/H, var/custom_base)
 	ASSERT(src)
 	ASSERT(istype(H))
--- a/code/modules/mob/living/carbon/human/species/species_vr.dm
+++ b/code/modules/mob/living/carbon/human/species/species_vr.dm
@@ -57,6 +57,7 @@
 		var/list/nif_savedata = H.nif.save_data.Copy()
 		..()
 
+		H.nif = null
 		var/obj/item/device/nif/nif = new type(H,durability,nif_savedata)
 		nif.nifsofts = nifsofts
 	else


Heroman3003:
I am somewhat worried about it having adverse effects in non-rejuvination uses of create_organs

Kashargul:
That's also my worry, but I've only seen this function call used on shapeshift and changeling heal. Or did I miss any other calls of it?

But there's one more thing I want to test, the case if one has a nif and hasn't saved yet and then gets rejuvenated. (That might be an edge case that needs to be handled in create organs, checking if nif and not nif in save file, the add it there)

Kashargul:
Due to the concerns I've created the following test case in the original code:

Code:

```
/datum/species/create_organs(var/mob/living/carbon/human/H)
	if(H.nif)
		var/type = H.nif.type
		var/durability = H.nif.durability
		var/list/nifsofts = H.nif.nifsofts
		var/list/nif_savedata = H.nif.save_data.Copy()
		..()
		var/obj/item/device/nif/nif = new type(H,durability,nif_savedata)
		nif.nifsofts = nifsofts
	else

```

Testproc:

```
/client/proc/create_organs_now(var/mob/living/carbon/human/H in player_list)
	set name = "create_organs_now"
	set desc = "Testing."
	set category = "Admin"
	H.species.create_organs(H)
```

I've then used an character with a saved nif and here's the result upon using the proc on them:
grafik

As soon as we call the proc, the old head along with the installed NIF is deleted.
grafik

The NIF remains lingering until the GC removes it.
grafik

We lose our nif.
grafik

If at any time we would've used the create organs proc, we should have noticed it as it would've lead to broken NIFs.

PR #15563, opened 2023-12-04
Crop Jacket Sprite Tweaks

KillianKirilenko:
Looking at these (#14734) again, the little orange dots bothered me. So I went through and purged them all.


Heroman3003:
conflict appeared, sorry. I should have merged this first, since it was open longer, my bad.

PR #15586, opened 2023-12-09
fix swapped messages in UI interaction, added missing filter in tg-chat, hopeuflly the last follow up for  #15559

Kashargul:
Fixes: the accidental message swapping pred/prey between the vorebelly and the tgui panel
Adds: The filter for the vorgan messages to tg-chat


ItsSelis:
Will be integrated into #15585

PR #15613, opened 2023-12-16
TWEAK: Enables roller beds to cross redgate

Runa-Dacino:


SatinIsle:
Tested with a simple mob attached to the roller, works fine by kicking the simple mob off whilst sending the roller through, so it doesn't allow anything to bypass that normally couldn't.

PR #15615, opened 2023-12-17
allows pulling of whitelisted items through redgate

Kashargul:
requires #15614

🆑 Upstream
Adds: The ability to pull whitelisted items through the redgate
Adds: Advanced roller bed to the whitelist as well
/🆑


code/modules/awaymissions/redgate.dm
```diff
@@ -12,7 +12,8 @@
 	var/secret = FALSE	//If either end of the redgate has this enabled, ghosts will not be able to click to teleport
 	var/list/exceptions = list(
 		/obj/structure/ore_box,
-		/obj/structure/bed/roller
+		/obj/structure/bed/roller,
+		/obj/structure/bed/roller/adv
```
Heroman3003, line 16: I suggest removign roller/adv from the list and using is_type_in_list (or however proc name correctly goes) in the check for type.

code/modules/awaymissions/redgate.dm
```diff
@@ -54,9 +55,24 @@
...
+		if(M.pulling)
+			var/atom/movable/pulled = M.pulling
+			M.stop_pulling()
+			playsound(src,'sound/effects/ominous-hum-2.ogg', 100,1)
+			M.forceMove(ourturf)
+			if(pulled.type in exceptions)
+				if(istype(pulled,/obj/structure/bed/roller))
+					var/obj/structure/bed/roller/R = pulled
+					for(var/mob/on_roller in R.buckled_mobs)
+						if(!on_roller.key || (on_roller.type in restrictions))
```
Heroman3003, line 67: Same for mob restrictions as I said above, but also, I'd suggest not making this check exclusive to roller beds, and instead all objects, since its possible more will be added later.

Heroman3003 (changes requested):

code/modules/awaymissions/redgate.dm
Kashargul: my mistake, missed that buckled mobs applies to all movable, gonna change that up

code/modules/awaymissions/redgate.dm
```diff
@@ -55,17 +54,15 @@
...
 			M.forceMove(ourturf)
-			if(pulled.type in exceptions)
-				if(istype(pulled,/obj/structure/bed/roller))
-					var/obj/structure/bed/roller/R = pulled
-					for(var/mob/on_roller in R.buckled_mobs)
-						if(!on_roller.key || (on_roller.type in restrictions))
-							R.unbuckle_mob(on_roller, TRUE)
+			if(is_type_in_list(pulled, exceptions))
+				for(var/mob/living/buckled_on in pulled.buckled_mobs)
+					if(!buckled_on.key || (buckled_on.type in restrictions))
```
Heroman3003, line 64: generalize buckled_on typecheck too, please.

PR #15621, opened 2023-12-18
Fixes an easy server crash.

BlackMajor:
>While loop with negative value


Heroman3003:
Is that caused by negative meat amount or something? Lol

Zeldazackman:
This sounds like the inverse of the endless chicken nugget fridge incident from a while back.
Balance after so many years has finally been achieved, nature is healing.

PR #15634, opened 2023-12-21
Replaces the RnG of electric shocks with a more sane calculation

BlackMajor:
TL;DR
Shock damage being an RnG value hard capped at 1MW with "jumps" in damage at certain breakpoints was weird so I smoothed out the math. Overall, damage is uncapped but does a little less damage.

Extended answer:
The old method returns burn damage to the victim in a series of five "tiers" based on the available power at the time.
These tiers roll a random with disadvantage, the thresholds in question were as follows:
Note that these values are rolled with disadvantage.
1kW to 50kw = 10-20
50kW to 100kW = 15-40
100kW to 200kW = 20-60
200kW to 1MW = 25-80
1MW+ = 50-160

I've replaced the math with a logarithmic that looks like this
burn damage = log(1.1, power) - log(1.1, (log(1.1, power)*1.5))
which results in a smoothed out damage curve with a plateau around 1MW, as follows:
1kW = 5
10kW = 24
100kW = 45
250kW = 53
1MW = 66
10MW = 88
100MW = 110
1GW = 132

### DIFF
--- a/code/modules/power/cell.dm
+++ b/code/modules/power/cell.dm
@@ -233,30 +233,17 @@
 	return
 
 /obj/item/weapon/cell/proc/get_electrocute_damage()
-	switch (charge)
-/*		if (9000 to INFINITY)
-			return min(rand(90,150),rand(90,150))
-		if (2500 to 9000-1)
-			return min(rand(70,145),rand(70,145))
-		if (1750 to 2500-1)
-			return min(rand(35,110),rand(35,110))
-		if (1500 to 1750-1)
-			return min(rand(30,100),rand(30,100))
-		if (750 to 1500-1)
-			return min(rand(25,90),rand(25,90))
-		if (250 to 750-1)
-			return min(rand(20,80),rand(20,80))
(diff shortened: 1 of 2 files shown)


Runa-Dacino (approved):
At first, hearing of this PR I was apprehensive, worried that we will end up with nerfed consequences and safer electricity.

This worry had been appeased, and even reversed with the uncapping! It gives reason to be careful with RCON set-up (shocks can be made harmless if you don't set output blindly to 250!) and to be smarter about how you power shields on SD rather than just consequence-free hotwiring things into the main.

I like these changes, and they have my support.

BlackMajor:
I can tweak the math if you want more fine tuning to a specific curve.
The one implemented is just kinda a rough guess of fitting the curve to around where the old values lied, just without the whole.. break points thing.

Whole inspiration for doing this was realising there was effectively only three choices for engineers setting the RCON. Those being 99kW, 199kW and 250kW, at each respective breakpoint.

Heroman3003 (approved):

PR #15639, opened 2023-12-23
fixes unusable grill recipes

Kashargul:
adds the special ingredients to the grill and cleans up duplications which are already on the default list.

🆑 Upstream
fix: Fixes the unusable grill recipes by adding the special ingredients to the insertable list
/🆑


Heroman3003:
The brain is still duplicated; and why move it to initialization anyway?

Kashargul:
I do not think the fryer can use them, the brain is used in the grill and oven though. Hence why I kept them to where they are actually usable there and only removed those which were already in the general one.

PR #15644, opened 2023-12-26
allows to disable the vorerpanel textlimit warnings

Kashargul:
🆑 Upstream
qol: Adds an option to disable the text limit warning texts in the vorepanel
/🆑


Heroman3003:
I think its better if it was more a "dont show again after showing once" thing. Maybe a var on the mob or something among those lines. Separate pref to already-rather bloated pref list just for that is a bit unnecesary.

Kashargul:
Mostly went with it as we have that section planned downstream anyway to add a few more options, like disable spot prey / spont pred confirmations temporarily for the round. But if you want it differently upstream, I can see about that in a few days or a weeks, depending how much time the current TG-Chat improvements consume.

Kashargul:
Changed up to be on the warning itself and temporarily in #15656

PR #15658, opened 2024-01-02
borg privacy in droms

Kashargul:
It's been something that had me bothered for a while. Why can borgs be tracked in OOC protected areas by security and ai?

Well this PR aims to remove both options within dorms. If a borg is inside a dorm, the AI as well as Security will now only see a static screen instead of the view of the borg.

For this, a new area flag has been added to the dorms `block_tracking`. This flag allows to create areas where borg cameras can not be remote accessed.

🆑 Upstream
qol: Adds privacy for borgs in dorms
/🆑


Heroman3003:
While Id say this could apply to some other camera types not just borgs, thats out of scope. Good PR.

PR #15667, opened 2024-01-08
Fix GLOB Edit button crashing the server

Repede:
Fixes GLOB Edit for global view variables so it does not crash the server instantly - Code by [@Willburd](https://github.com/Willburd)

Does this by removing a sleep call and replacing with spawn(0)


Heroman3003:
Tempmerge.

PR #15668, opened 2024-01-08
Absorb Escape default changed to 0

Fracshun:
Small quality of life change. Before absorb escape was introduced it was impossible to escape while absorbed. To keep in line with old functionality this should be 0 as a default, and can be increased if desired. I've had to change this to 0 in so many of my existing absorb bellies already to prevent struggles from triggering the escape chance.

### DIFF
--- a/code/modules/mob/living/simple_mob/simple_mob_vr.dm
+++ b/code/modules/mob/living/simple_mob/simple_mob_vr.dm
@@ -31,7 +31,6 @@
 	var/vore_digest_chance = 25			// Chance to switch to digest mode if resisted
 	var/vore_absorb_chance = 0			// Chance to switch to absorb mode if resisted
 	var/vore_escape_chance = 25			// Chance of resisting out of mob
-	var/vore_escape_chance_absorbed = 20// Chance of absorbed prey finishing an escape. Requires a successful escape roll against the above as well.
 
 	var/vore_stomach_name				// The name for the first belly if not "stomach"
 	var/vore_stomach_flavor				// The flavortext for the first belly if not the default
@@ -237,7 +236,6 @@
 	B.contamination_color = vore_default_contamination_color
 	B.escapable = vore_escape_chance > 0
 	B.escapechance = vore_escape_chance
-	B.escapechance_absorbed = vore_escape_chance_absorbed
 	B.digestchance = vore_digest_chance
 	B.absorbchance = vore_absorb_chance
 	B.human_prey_swallow_time = swallowTime
--- a/code/modules/mob/living/simple_mob/simple_mob_vr.dm
+++ b/code/modules/mob/living/simple_mob/simple_mob_vr.dm
@@ -31,6 +31,7 @@
 	var/vore_digest_chance = 25			// Chance to switch to digest mode if resisted
 	var/vore_absorb_chance = 0			// Chance to switch to absorb mode if resisted
 	var/vore_escape_chance = 25			// Chance of resisting out of mob
+	var/vore_escape_chance_absorbed = 20// Chance of absorbed prey finishing an escape. Requires a successful escape roll against the above as well.
 
 	var/vore_stomach_name				// The name for the first belly if not "stomach"
 	var/vore_stomach_flavor				// The flavortext for the first belly if not the default
@@ -236,6 +237,7 @@
 	B.contamination_color = vore_default_contamination_color
 	B.escapable = vore_escape_chance > 0
 	B.escapechance = vore_escape_chance
+	B.escapechance_absorbed = vore_escape_chance_absorbed
 	B.digestchance = vore_digest_chance
 	B.absorbchance = vore_absorb_chance
 	B.human_prey_swallow_time = swallowTime
--- a/code/modules/mob/living/simple_mob/subtypes/animal/alien animals/catslug.dm
+++ b/code/modules/mob/living/simple_mob/subtypes/animal/alien animals/catslug.dm
@@ -124,6 +124,7 @@
 	B.digestchance = 10
 	B.absorbchance = 1
 	B.escapechance = 15
+	B.escapechance_absorbed = 20
 
 /datum/ai_holder/simple_mob/melee/evasive/catslug
 	hostile = FALSE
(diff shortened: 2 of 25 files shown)


Heroman3003:
Absorbs are still not escapable unless somebody actually enables escapability in their bellies, which is still off by default. This 'change' (which wasnt really a change because there was no default functionale to the feature before) only affects simplemobs. The way its being changed will make absorby simplemobs inescapable again and is generally undesirable in my eyes. Currently against.

Screemonster:
The 20% chance will only be rolled if the prey _already_ succeeds at a regular escape attempt against the normal escape chance of your stomach. You don't need to move it off default unless you're in the habit of making escapable bellies.

Unless you want to go down the list of simplemobs and add a nonzero absorb-escape chance, this will break escaping from absorby mobs.

Fracshun:
This is not currently the case. The 'escapable' value needs to be turned on for any interaction chance, and not just being escapable. So the problem lies that bellies which use struggle for transfers do in fact become escapable for those absorbed in them with the default value. This has been tested and it has happened to myself randomly in game which is rather frustrating. The normal escape chance which defaults to 0% does not play a role in if absorbed creatures can escape. See line 1524 in belly_obj_vr.dm.
`if(prob(escapechance_absorbed) || owner.stat) //Let's have it check to see if the prey's escape attempt starts.`
So if the intention of only escaping if the normal escape chance succeeds is how it's intended, then that's not how it's currently written.

If the goal is to make simplemobs that absorb to be escapable then this could be done another way, or they can use OOC Escape which is how it was previously done. Right now the default value has been an inconvenience to a number of other players I've spoken with who use the absorb mechanic a lot. Existing bellies which used struggle transfers all updated to use the 20% value so the damage is done.

Let me know how you would like this to be handled, but right now it is an issue. A ckey check for the belly host on struggle attempts could be used instead of this default if you want simplemobs to be escapable.

Screemonster:
Had another dive into the current code and I see the problem.

https://github.com/VOREStation/VOREStation/blob/73f613400d83422a4917dfae629699687fa7f5b6/code/modules/vore/eating/belly_obj_vr.dm#L1524

>                if(prob(escapechance_absorbed) || owner.stat) //Let's have it check to see if the prey's escape attempt starts.

This should be checking escapechance, not escapechance_absorbed - the actual absorb check takes place on line 1552.

This means there's currently a bug where the absorbchance is being checked twice, rather than the struggle chance _then_ the absorb chance. That bug should be fixed instead.

Fracshun:
How about the above? This fixes the bug that was mentioned, keeps the default 0, but adds a condition to allow non-player controlled mobs a default absorb escape chance of 20 if one is not already given.

Screemonster:
This would break the couple of mobs that have a non-20% absorb-escapechance. It doesn't need to be hardcoded as a thing mobs do, it just needs to be set on mobs that don't have it.

Heroman3003:
I'm not sure about this hardcoded solution. I think editing it so that wherever its defined that default mob bellies are escapable, it also automatically gives them 20 absorb escapechance is better, rather than this, as it prevents having varied chances entirely.

Screemonster:
This is a good enough stopgap for until someone wants to go down the list of mobs and give them all defined escape chances (though doing that would be preferable)

Fracshun:
If a non-zero value is supplied it will use that first. The order of the condition is not hardcoding all simplemobs to have a 20% chance. The only caveat is as Scree pointed out; Defining an absorb mob to be inescapable wouldn't work. Though a player controlled one could do it if they wanted.

Fracshun:
Needing to sleep, but tomorrow I'll look at adding the variable to existing mobs to remove the additional condition and see how we go from there.

Screemonster:
> Looking into this, if it's meant to prob both, that's an awfully low chance, might be better to go the deep dive and make it actually respect the % chance so that players can have it independently from normal escape.
> 
> 10% proc on 10% proc is like a 1% chance currently for example... Might be confusing for players. If Fracshun doesn't want to do the mob dive, I could do it later and add the var to every mob.

That's kinda the point - it _is_ supposed to be a fairly low chance, such that you have to succeed a roll to _start_ struggling and have a good chance of the mob going "nope" and dragging you back in - which is why I set it as high as 20% in the first place considering that once you're absorbed there's really not a lot else the mob can do to you and you can just keep spamming resist until your escape attempt succeeds.

To give an example of one of the exceptions to this, gaslamps are a specifically absorby mob and they have a high base escapechance and a considerably lower absorbed escapechance - the intent is that they're fairly easy to get out of if you just get swallowed, but if you stick around in them too long then you'll have a way harder time.

Kashargul:
> > Looking into this, if it's meant to prob both, that's an awfully low chance, might be better to go the deep dive and make it actually respect the % chance so that players can have it independently from normal escape.
> > 10% proc on 10% proc is like a 1% chance currently for example... Might be confusing for players. If Fracshun doesn't want to do the mob dive, I could do it later and add the var to every mob.
> 
> That's kinda the point - it _is_ supposed to be a fairly low chance, such that you have to succeed a roll to _start_ struggling and have a good chance of the mob going "nope" and dragging you back in - which is why I set it as high as 20% in the first place considering that once you're absorbed there's really not a lot else the mob can do to you and you can just keep spamming resist until your escape attempt succeeds.
> 
> To give an example of one of the exceptions to this, gaslamps are a specifically absorby mob and they have a high base escapechance and a considerably lower absorbed escapechance - the intent is that they're fairly easy to get out of if you just get swallowed, but if you stick around in them too long then you'll have a way harder time.

Understandable for MOBs, just seeing it out of the player view looking at the Panel where they might think that those settings are not linked. That's also how that mix up with the variable happened. Having talked with Fracshun, it might be more convenient for players to have it as:

Escape init chance
Escape chance
Escapce chance absorbed

So that those settings are clearly defined and 10 % mean 10%. But that's more the view of how to interpret those settings from player perspective. But when going to touch all mobs anyway, could be worth to consider.

Fracshun:
The absorb escape chance has now been added to the bellies of all simplemobs I could find that had bellies. With the exception of the gaslamp which already had it. The failsafe condition has been removed as a result.

Heroman3003:
Okay, thats exactly what I meant with one last change to request. Putting the escape chance on *every individual mob* is unnecesary unless its meant to be different, since its already set with default mob define.

Fracshun:
Sure. If it would be preferred to not add it on all simplemobs then this should be the only change required.

Heroman3003 (approved):

PR #15669, opened 2024-01-08
Metamorphosis Ray design for RnD

Fracshun:
A fun scene tool now more available, similar to how RnD can print others like the Body Snatcher. Tech and Material costs based off the Recombobulation Ray design which is effectively the same gun but only reverts. Added requirement for Illegal level 5 and placed under the energy weapon category.

### DIFF
--- a/code/modules/research/designs/weapons_vr.dm
+++ b/code/modules/research/designs/weapons_vr.dm
@@ -33,6 +33,14 @@
 	build_path = /obj/item/weapon/gun/energy/netgun
 	sort_string = "MAAVC"
 
+/datum/design/item/weapon/energy/metamorphosisray
+	name = "metamorphosis ray"
+	id = "metamorphosisray"
+	req_tech = list(TECH_MATERIAL = 3, TECH_ENGINEERING = 3, TECH_POWER = 4, TECH_BIO = 5, TECH_BLUESPACE = 4, TECH_ILLEGAL = 5)
+	materials = list(MAT_STEEL = 1000, MAT_GLASS = 2000, MAT_URANIUM = 500, MAT_PHORON = 1500)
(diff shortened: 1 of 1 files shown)


Heroman3003:
I am very iffy on a tool with such big potential for memery and stupidity being accessible too easily. Currently against it myself. There's a reason staff of change which once was actually in protolathe in server's history got removed, and this is literally a better version of it being put there.

Fracshun:
This feels like a moderation issue. The device still needs to be requested from science unless it's a science member procuring it for themselves. If someone is memeing and breaking OOC prefs with it then the problem more lies in the one breaking prefs. Those with spontaneous TF preference off will not be affected by the ray at all. Note that this device can already be found in trash piles, and there are a number of tools that can be used in similarly disruptive ways. The idea is to open up the tool for those who want to use it properly. With it now being months after the device was introduced the initial wave of those wanting to mess around with the new tool has died down with it being more widely known about.

Runa-Dacino:
I am with Heroman.

Since you specifically argue for it as a scene tool - if you wish to use Meta ray for a scene - if you Ahelp, we will happily provide it for you given some caveats around frequency, staff availability. This has the benefit of ensuring it will be actually used for a scene rather than disruptive behaviours.

I have had to deal with numerous people using meta ray to turn into catslugs and run around and be highly disruptive without any real counter by command or security. Playerside, I'm not looking forward to certain people getting to do that more often.

Screemonster:
I remember when the staff of change was in the protolathe and all the utter bullshit issues that went with it and I'm not eager to relive those days.

Fracshun:
No worries. I'll close this one early then as I would like to submit it downstream. Let me know if anything changes.

PR #15670, opened 2024-01-08
SleeveMate Emag action

Fracshun:
Another use for an Emag. When used on a SleeveMate it will hack and modify it to become the illegal Body Snatcher version. Comes with harmless sparks! More ways to get scene tools and just more fun uses of the green card.


Heroman3003:
I don't have any strong thoughts on this. Seems fine, but with how rare emags have been made, its unironically much easier to find or make a bodysnatcher than get an emag for that, so I don't entirely get the point.

PR #15671, opened 2024-01-09
Spider

Very-Soft:
Makes it so that nurse spiders can't spawn from any random sources, such as spider egg clusters, spider spawners, and capture crystals.

GMs and map makers can still put nurse spiders out if they want to, but this should stop any UNEXPECTED nurse spiders from appearing, and stop spiders from snowballing quite so hard when the conditions are right.


code/modules/mob/living/simple_mob/subtypes/animal/giant_spider/broodmother.dm
```diff
@@ -31,7 +31,6 @@
 		/mob/living/simple_mob/animal/giant_spider/electric/broodling,
 		/mob/living/simple_mob/animal/giant_spider/hunter/broodling,
 		/mob/living/simple_mob/animal/giant_spider/lurker/broodling,
-		/mob/living/simple_mob/animal/giant_spider/nurse/broodling,
```
Heroman3003, line 34: Broodmother is itself adminspawn only, and it seems excessive to nerf its sub-spawns there

code/game/objects/effects/spiders.dm
```diff
@@ -154,7 +154,7 @@
 	var/amount_grown = 0
 	var/obj/machinery/atmospherics/unary/vent_pump/entry_vent
 	var/travelling_in_vent = 0
-	var/list/grow_as = list(/mob/living/simple_mob/animal/giant_spider, /mob/living/simple_mob/animal/giant_spider/nurse, /mob/living/simple_mob/animal/giant_spider/hunter)
```
Heroman3003, line 157: Those eggs aren't used normally, so them spawning a nurse is fine I believe. If they are mapped in or used somewhere, I think it's better to make a subtype that spawns only hunters and regulars instead.

code/game/objects/effects/spiders.dm
```diff
@@ -163,7 +163,7 @@
 	grow_as = list(/mob/living/simple_mob/animal/giant_spider/frost)
 
 /obj/effect/spider/spiderling/varied
-	grow_as = list(/mob/living/simple_mob/animal/giant_spider, /mob/living/simple_mob/animal/giant_spider/nurse, /mob/living/simple_mob/animal/giant_spider/hunter,
```
Heroman3003, line 166: Ditto

code/game/gamemodes/technomancer/spells/summon/summon_creature.dm
```diff
@@ -30,7 +30,6 @@
 		"BAT"			=	/mob/living/simple_mob/animal/space/bats,
 		"SPIDER"		=	/mob/living/simple_mob/animal/giant_spider,
 		"SPIDER HUNTER"	=	/mob/living/simple_mob/animal/giant_spider/hunter,
-		"SPIDER NURSE"	=	/mob/living/simple_mob/animal/giant_spider/nurse,
```
Heroman3003, line 33: This could just be replaced with the /eggless subtype

code/_helpers/global_lists_vr.dm
```diff
@@ -664,9 +664,7 @@ var/global/list/xenobio_gold_mobs_hostile = list(
 										/mob/living/simple_mob/animal/giant_spider/webslinger,
 										/mob/living/simple_mob/animal/giant_spider/phorogenic,
 										/mob/living/simple_mob/animal/giant_spider/carrier,
-										/mob/living/simple_mob/animal/giant_spider/nurse,
```
Heroman3003, line 667: Also a good place to use /eggless is both spiders in this file

Heroman3003 (commented):
At moments it seems like a bit of an overcorrection. Some of those sources are tied to adminspawn/pre-mapping already, and not player obtainable, while others would be fine with being replaced with /eggless. I agree with reducing ability for breedable spiders to be randomly obtainable, but completely eliminating nurse spiders from every pool is more eh.

Novacat:
Nurse spiders should still show up outside of station. I would say that the proper course of action is to just remove them from capture crystals, since those can be readily obtained on station.

PR #15678, opened 2024-01-15
Bad Swimmer

SatinIsle:
Added a bad swimmer trait that slows you down in water and causes you to drown in deep water (depths of 2 or above).

One thing I would like to improve before merging and want advice on: Currently the splashing/"you are drowning" message spams pretty frequently and I would like to slow that down.


Heroman3003:
Easiest way to do that would probably be to make message appearing chance based (like 5% chance per time)

SatinIsle:
> Easiest way to do that would probably be to make message appearing chance based (like 5% chance per time)

I tried it with 5% but it seemed rare for a message to pop up at all before the character was almost dead, 10% feels like a good balance to me now.

PR #15679, opened 2024-01-15
ends multiproccing on butchering

Kashargul:
🆑 
fix: no more multiproccs on butchering
/🆑


Screemonster:
so what's the deal here? Does this get rid of the thing where you slice meat off a fillet at a time and then gib it at the end?

'cause I kinda liked that, I _almost_ did a thing with puffercarp where you could butcher them by slicing meat off and you'd be safe as long as you stepped away and aborted the final step that gibs

PR #15684, opened 2024-01-18
Many bug fixes, QoL, and PDA notes port

Darlantanis:
Fixes many old bugs, nerfs a foam greatsword exploit, lets players OOC escape from microwaves, and ports an upgrade to the PDA notes app. Not sure why github is detecting entire file replacements; no files have been completely replaced. I guess I'll detail all the code changes below:

- Microwaves are now OOC escapable, just adds to the OOC if() checks.
- Adds "can_sharpen" bool to material datums with TRUE by default, sets it to FALSE for foam, and adds it to the sharpen() check. This prevents players from creating the foam durasteel greatsword, a HALLOS weapon which can knock out most species in one hit.
- Tweaks codexes to generate a single codex_tree per book type instead of a tree per book, so 300 books can use 1 massive datum instead of 300 identical massive datums which also patches a memory leak when codexes are deleted. Also fixes a bug with the "go back" button permanently breaking due to a bad math.
- Makes a "rocks" decl and assigns it to the outdoors/rocks turf, reintroducing correct footstep sounds.
- (Hopefully) fixes a bizarre edge case with tg/map_template where submaps which do not allow duplicates repeatedly spawns on top of itself by checking "allows_duplicates" in the while() loops
- Adds extra note sections to the PDA notes app, allows the notes to copy text from papers, and allows notes to print onto blank papers
- Fixes a chem ID for zombie powder so it appears to stop hearts again.
- Tweaks highlander mode contents deletion loop to no longer delete organs, which seems to have fixed a crash issue. I doubt anybody uses this, but having less server crash buttons is probably a good thing.
- Adds check so conveyor belts can no longer move obj/effect/abstract, so it can't steal your flashlight light anymore.
- Adds a missing flag to particle accelerator particles and spawns the particle on the device itself, so now it bumps and irradiates mobs in front of the particle accelerator again.
- Adds missing surgery names
- Tweaks photographs and dogborg sleepers to account for MaxHealth so species with lower max health no longer appear hurt while healthy.
- Adds loops to organ spawning to properly run meat initializing, making organs butcherable again. Also changes a forceMove which could leave organs stuck to player HUD's.

Everything has been tested locally. Big thanks to [@Willburd](https://github.com/Willburd) for the bulk of this code.

Downstream changelog
:cl:
-Microwaves are now OOC escapable
-Nerfs foam durasteel greatsword
-Fixes memory leak with codexes
-Fixes bug which bricks codex "go back" button
-Reimplements footstep sounds for rock turfs
-Fixes edge-case which sometimes repeatedly spawned POI's on top of each other
-Adds more PDA notes, adds copy and print function to PDA note app
-Zombie powder makes hearts appear stopped again
-Highlander mode no longer crashes the server, and also doesn't delete your organs anymore
-Conveyor belts no longer move flashlight light cones
-Particle accelerator properly applies radiation to bumped mobs again
-Adds missing surgery names
-Healthy tesh no longer appear hurt on photographs and dogborg sleepers
-Organs no longer get stuck to the HUD
-Organ butchering works again
/:cl:


Heroman3003:
What in the world are those diffs

Kashargul:
With it only being 23 files, I heavily recommend to recommit them to a new branch. Having entire files shown as replaced even though there's only been a few changed makes any later bug hunting a hell.

PR #15703, opened 2024-01-27
Fixes disposal overpressurization

ReoDaProtovali:
Changes disposal bin's target pressure to be the same as Yogstation's. No real reason for yogstation in specific, I just had the file to compare to, and I was astonished by the incredibly large difference in pressure target. Anyhow, this should significantly help with the mailroom being 20 billion KPA under the plastic flaps


Heroman3003:
Would this not cause the pressurization of the bins to become near-instant?

ReoDaProtovali:
It might, but bins draining rooms of air and overpressurizing their destination is a bigger issue. if you want, I could also adjust the fill rate to reflect yogs as well

Heroman3003:
Well, here we don't have pressurization problem in part because the delivery rooms were mapped to have special vents that stabilize it, so this doesn't really fix any issues for us, but if it does make the bins instant, its definitely not wanted.

Heroman3003:
Alright. Since there is general appearance of people mostly not having strong opinions now that recharge has been adjusted and only one against for the change of atmos, I'll make a call and say its fine. That change is ultimately irrelevant on our end.

PR #15706, opened 2024-01-27
we never want half or empty stacks

Kashargul:
We neither want half stacks as they can break things, nor do we want empty, unusable stacks, so let's just round and test if the stacks actually contain something before creation.

🆑  Upstream
fix: dropping empty stacks during protolathe / mech fab deconstruction
/🆑


ReoDaProtovali:
epic material return fail

PR #15707, opened 2024-01-28
more law board datums

Kashargul:
🆑 Upstream
add: allows the circuit printer to produce most of the lawboards now except for the event lawboards
/🆑


code/modules/research/designs/circuits/ai_modules.dm
```diff
@@ -104,4 +104,141 @@
...
+
+/datum/design/aimodule/illegal/corp
+	name = "Corporate"
+	id = "corp"
+	req_tech = list(TECH_DATA = 2, TECH_ILLEGAL = 4, TECH_MATERIAL = 2, TECH_COMBAT = 2)
+	build_path = /obj/item/weapon/aiModule/corp
+	sort_string = "XADAB"
+
+/datum/design/aimodule/illegal/drone
+	name = "Drone"
```
Heroman3003, line 184: I'd say this is a 'technical' one and shouldnt really be in the list

Kashargul: Hmmh, true, might be better to keep it to the maint drones / admin poking.

PR #15719, opened 2024-02-01
Quick fix for some instances of erroneous planetary lighting

BlackMajor:
tl;dr
Outdoors turfs refer to the planetary subsystem to calculate lighting.
/turf/simulated/floor has a check for outdoors turfs and adds them to planets.
/turf/simulated does not. Thereby /turf/simulated/open gets funky looking when outdoors. Maybe some others too, but it's mostly that. Tether gets around this by defining an explicit subtype that does this very same check.
Figure it'd be easier for map developers in the future to not have to make exceptions and instead just include the check in the base /simulated type path.

A forewarning that this ///might/// produce an instance or two of weather effects in unexpected locations that I can't predict.


Heroman3003:
this won't result double addition of the turf, considering, as you said, the proc also exists on the children of that?

Heroman3003:
I will reiterate my concern with this change potentially adding turf twice to planetary turfs list. addTurf proc does NOT check whether its already present in the list before adding it and it would be a very unnecesary additional processing for planets subsystem.

Kashargul:
If it's added to the parent, why keep it on the child?

Heroman3003:
@BlackMajor

BlackMajor:
> @BlackMajor

Oh my bad, I've been mad busy and outright forgot about this PR.

Works in testing, but I should probably scrub the redundant calls in the children procs.
Existing multi-Z planet maps would need to be updated though. Y'want me to just do that right now?

BlackMajor:
Ah blood hell, I'll just remake this properly

PR #15721, opened 2024-02-01
dismantle takes optional var

Kashargul:
🆑 Upstream
qol: change the place_dismantled_product to accept an amount variable to allow to drop stacks larger than 1, defaulted to 1
/🆑


Heroman3003:
I actually prefer walls not being perfect 100% resource refund on deconstruction myself.

Kashargul:
Fully fair, removed that section.

PR #15727, opened 2024-02-04
removes all \icon

Kashargul:
Seemingly the \icon code is fully deprecated. It does nothing on old chat and causes a massive error log in Tg-Chat, inflating server log files into the GB area within merely a few rounds.

🆑Upstream
del: removes all \icon tags
/🆑


Heroman3003:
Are you certain it does nothing on oldchat? IIRC it was added specifically because bicon didnt work for oldchat.

Kashargul:
> Are you certain it does nothing on oldchat? IIRC it was added specifically because bicon didnt work for oldchat.

I didn't see any icons using oldchat, maybe that needs to be tested again. I only tested it for a short time and all icons in oldchat looked broken / didn't load during my tests.

Kashargul:
Seems it's actually removing the images from oldchat. Going to test the icon2html proc later to see how that ends up.

Kashargul:
I've not found a way for it to work with old chat and TG-Chat without erroring on TG-Chat or removing the icons on old chat.

Kashargul:
Part of the 515 PR to rid of those.

PR #15748, opened 2024-02-11
Whitespace Standardization [MDB IGNORE]

Drulikar:
This PR updates the editorconfig and gitattributes to standardize more whitespace configurations (primarily LF line endings). The values are just what TG has been using as their standard for some time: https://github.com/tgstation/tgstation/blob/master/.editorconfig https://github.com/tgstation/tgstation/blob/master/.gitattributes

I originally did this for a downstream so I could update their dmm and dmi merger hooks, but the inconsistent line endings were causing problems with the hooks. I am now PRing the whitespace changes at their request upstream because mirror bot has just been always conflicting changes now.

I would have to remove various maps from the repository that are corrupt in order to update the dmm/dmi hooks here (many are committed with unresolved merge conflicts that have been compounding - but others I just could not find any way to resolve and had to opt to just remove). See https://github.com/CHOMPStation2/CHOMPStation2/pull/7620 if you're curious.

Should you want to ignore this PR from git blame, see how TG has done so for other mass whitespace changing PRs like this one by entering the commit hash in a .git-blame-ignore-revs file e.g. https://github.com/tgstation/tgstation/blob/master/.git-blame-ignore-revs.

I recommend viewing the files changed with whitespace ignored to see the actual (lack of) changes. 
image


Drulikar:
Not sure if I can satisfy the lint https://github.com/VOREStation/VOREStation/actions/runs/7859122317/job/21444866267?pr=15748#step:6:10

Locally I tried reverting the commit where it was altered then added an override gitattributes and editorconfig to the html folder to allow yml to be crlf instead, but even after `git rm --cached -r html/changelogs/example.yml` it still is considered altered. So if I just redact this from the PR, its very likely it will just become a file everyone will start to see as changed; so it would need to then become a gitignored file.

So I guess let me know if you'd like me to revert that commit, perform the whitespace changes again without that example, and gitignore the example yaml.

ItsSelis:
> Not sure if I can satisfy the lint https://github.com/VOREStation/VOREStation/actions/runs/7859122317/job/21444866267?pr=15748#step:6:10
> 
> Locally I tried reverting the commit where it was altered then added an override gitattributes and editorconfig to the html folder to allow yml to be crlf instead, but even after `git rm --cached -r html/changelogs/example.yml` it still is considered altered. So if I just redact this from the PR, its very likely it will just become a file everyone will start to see as changed; so it would need to then become a gitignored file.
> 
> So I guess let me know if you'd like me to revert that commit, perform the whitespace changes again without that example, and gitignore the example yaml.

The linter fails because after the change the md5 hash of that file changes.
You can make the error go away if you regenerate the MD5 after the change and set the hard coded value accordingly:
https://github.com/VOREStation/VOREStation/blob/bb1343e63d221c629b5623a948d2f106d95c34be/tools/ci/validate_files.sh#L50C1-L50C80

Drulikar:
Since the line endings locally are different than what will exist on github I don't think the SHA is going to match up (assuming you're using `core.autocrlf=true`); also if you're linting against it being changed then might as well just ignore changes made to it via gitignore I think. Does make it harder to update it - but if that was a problem there wouldn't be a lint against changing it :P

But if you don't want it gitignored then I can see about getting the hash for it with LF endings.

ItsSelis:
Ye it'd be neat to have the CI have the change too xP
I think this should be the correct hash with the LF:
ea467b7b75774b41ecdf35e07091d96f

ItsSelis (approved):

PR #15774, opened 2024-02-17
Smart Storage Tweaks

KillianKirilenko:
Some adjustments to Sheet Storage based on general feedback discussion. Implements the following;
- First, reduces the loss rate from 35\~45% to 20\~25%
- Second, increases the sheet maxcap from 250 to ~~400~~ 500
- Third, adds a minimum threshold of 20 to sheet culling

Numbers are of course up for debate, but as long as the storage is stocked *reasonably well* at some point in the prior 3~4 shifts, then even the deadest of deadshifts should have *some* materials to call upon with these numbers.

Tested locally, seems to work OK.

Input at end of test round was 800 steel, 300(?) silver, 100(?) gold, 15 graphite, 10 diamond. Stock at the start of the next round:
image

### DIFF
--- a/code/modules/persistence/storage/smartfridge.dm
+++ b/code/modules/persistence/storage/smartfridge.dm
@@ -33,7 +33,7 @@
 /datum/persistent/storage/smartfridge/sheet_storage/lossy
 	name = "sheet storage lossy"
 	min_storage = 20	//if the amount is at or below this, don't cull
-	max_storage = 400	//if the amount is above this, cull it to this amount THEN do math
+	max_storage = 500	//if the amount is above this, cull it to this amount THEN do math
 	stacks_go_missing = TRUE
 	minimum_storage_reserve = TRUE
 
--- a/code/modules/persistence/storage/smartfridge.dm
+++ b/code/modules/persistence/storage/smartfridge.dm
@@ -25,12 +25,17 @@
 	store_per_type = TRUE
 	target_type = /obj/machinery/smartfridge/sheets
 
+	var/min_retained = 75	//minimum percentage of current stock for retention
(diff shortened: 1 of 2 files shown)


KillianKirilenko:
Did a little bit of napkin math. Assuming roughly x0.77 each time, decay for a full storage should look like;

[first stocked round] 400 -> ~300 -> ~235 -> ~180 -> ~139 -> ~100 -> ~82 -> ~63 -> ~48 -> ~37 -> ~29 -> ~22 -> ~17 STOP (no further decay due to minimum cap).

Assuming cargo and science are frugal/efficient with resources (cargo only handing science as much as they need, and science returning all unused materials to cargo by round-end), then a full stock of a given material should last around two-and-a-half to three real-time days.

Heroman3003:
The only request that seems to have agreement on is upper cap of 500 (for rounder number). I am still very HMMMM about particle smasher materials being preservable indefinitely using this though.

KillianKirilenko:
Now, admittedly, I don't play often, but I can't remember the last time I saw R&D go to the trouble of putting together a particle smasher and then actually turning over any surplus for storage. I really wouldn't worry about it myself.

If it does actually turn out to be a problem down the line then it can be fixed.

TheToaster98:
The number of miners has been rather low lately, at least from my point of view of the Eurohours.
And maybe I am biased from playing sci for a week straight without any mats at all... but it could be at least tested for a week or two and see if people raise complains or not.
People complain or it is too OP? Revert it, takes 5 minutes.

Leaving it in the air doesn't do much, specially considering you can just drop a google Forms asking players for input and get a sizeable amount of data.

PR #15777, opened 2024-02-18
Adjusts a few SD things

Very-Soft:
Stuff people have been asking for over the years:

removes star treaders and space ghosts from space rocks spawn pool, as well as the more dangerous carp variants. In exchange adds space rays and space gnats, and readjusts the spawn rates taking these things into account. This should make things more forgiving for miners.

Mining also now has an additional material fridge. This one is exposed to the hallway though, where science can get at it. This should be useful for allowing science to return materials, or for cargo to make materials available, if they want to.

Additionally, the SD starts with a Qpad in front of science. (Science still needs to set it)

Oh also, caps the hallucination damage that space ghosts can do if you stand around and let them hit you

### DIFF
--- a/code/modules/mob/living/simple_mob/subtypes/animal/alien animals/spookyghost.dm
+++ b/code/modules/mob/living/simple_mob/subtypes/animal/alien animals/spookyghost.dm
@@ -97,7 +97,8 @@
 
 /mob/living/simple_mob/vore/alienanimals/space_ghost/apply_melee_effects(var/atom/A)
 	var/mob/living/L = A
-	L.hallucination += 50
+	if(L.hallucination <= 100)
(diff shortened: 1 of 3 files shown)


Heroman3003:
The only thing that should probably be adjusted is that the fridge should have access restrictions on it for cargo/engi/sci if it is to be THAT public. On tether its technically available to sci, but thats moreso due to access through backdoors, rather than it being public, so ideally that should be replicated. Also I do think plan of 'moving' it by having two fridges but only for a week before removing old one (while making sure all materials from it get moved to new one) is worth considering.

Heroman3003:
Oh, it also appears new fridge is not actually lossy subtype?

PR #15778, opened 2024-02-18
Tgui ui fix

Kashargul:
Fixes most of our TGUI UIs. Also, a few UIs are more bandaided right now. Overhauling them at some point might be better.

Merge #15770 first. Then update this branch and add the bundle.

🆑 Upstream
fix: fixes most of the TGUI Uis
/🆑


Kashargul:
No bundle on this as it would conflict with the previous PR.

Kashargul:
Adding bundle later today.

ItsSelis (approved):

PR #15791, opened 2024-02-29
Adds new mode for mining resonators: Resonance Cascade!

Runa-Dacino:
### What this does
By activating your held object while holding a resonator, you can set it to "Cascade" mode. By doing so, on clicking a tile it proceeds to spread in a pattern of opposite the origin, then relative left and relative right.

This repeats recursively until the field limit is reached. This recursive test was done up to field limit 9 where it produces up to 9 fields, but currently only field limit 5 tools are available without admin bus.

### Why we need this
@ResidentCody requested their addition as resonators currently are rather useless.

### Potential Problems

It appears, depending on how much the qdel() is delayed due to lag, the cascade MAY produce fewer fields than intended. It was not a problem on local unless I spammed the resonator without waiting a second from the time it finished resonating. 

### Commits
[Add: Cascade mode for resonators](https://github.com/VOREStation/VOREStation/commit/4d0e8ebb987801d230080c1fc052827ed6ce4043) 

:cl: Upstream
add: Added Cascade mode for resonators.
/:cl:

### DIFF
--- a/code/modules/mining/resonator_vr.dm
+++ b/code/modules/mining/resonator_vr.dm
@@ -18,6 +18,8 @@
 	var/fieldsactive = 0
 	var/burst_time = 50
 	var/fieldlimit = 3
+	var/spreadmode = 0
+	var/cascading  = 0
 
 /obj/item/resonator/upgraded
 	name = "upgraded resonator"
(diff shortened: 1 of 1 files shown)


Noobetski:
neat

PR #15808, opened 2024-03-05
Closet vore and Tents

SatinIsle:
Added a new verb to the ability tab when interacting with closets called "Vore Occupants". This allows you to eat another person when both of you are inside of the same closet.

Added tents as a subtype of closets.


Heroman3003:
Re: hidden vore

I recommend a typecheck on usr. Currently it seems like it might be causing runtimes if a ghost were to attempt doing it.

SatinIsle:
> Re: hidden vore
> 
> I recommend a typecheck on usr. Currently it seems like it might be causing runtimes if a ghost were to attempt doing it.

Done!

Heroman3003:
It has also been tested yes?

SatinIsle:
> It has also been tested yes?

Yep

PR #15813, opened 2024-03-08
Makes mentorhelps check if there's anyone available to answer, offer alternative if not

Runa-Dacino:
### What this does

Adds logic to mentorhelps to check the global mentorlist and the admin list (filtered for "Present" admins, as some admins like to leave their PC on when going to bed and they can't exactly answer either).

If nobody is available to answer, the player is prompted to make an admin help - choosing which immediately forwards/escalates their mhelp to an ahelp instead which then notifies discord so staff can step up and help.

Alternatively, they are suggested to go straight to discord themselves and ask in cadet academy. 

### Why we need this

There is nothing currently that informs players that there aren't any mentors online, and if they make an mhelp hoping to get help they'll likely not know to check staffwho in the first place and therefore, think mentors are ignoring them. Furthermore, even if they wouldn't think this - they still need that help and this will increase the probability of them receiving it.

### Commit details

@Runa-Dacino
[add(mentorhelp): Makes mhelps check for active mentors and admins](https://github.com/VOREStation/VOREStation/pull/15813/commits/24bf14bc0e361f25d46664b6e7c42345de6a32d4) 
[24bf14b](https://github.com/VOREStation/VOREStation/pull/15813/commits/24bf14bc0e361f25d46664b6e7c42345de6a32d4)
If no admin/mentor, it prompts sending ahelp instead or recommends going to discord personally.

### DIFF
--- a/code/modules/mentor/mentorhelp.dm
+++ b/code/modules/mentor/mentorhelp.dm
@@ -429,6 +429,25 @@ GLOBAL_DATUM_INIT(mhelp_tickets, /datum/mentor_help_tickets, new)
 	if(!msg)
 		return
 
+	// Making sure there's actually a mentor or admin who can respond.
+	var/list/admins = get_admin_counts()
+	var/list/activeAdmins = admins["present"]
+	var/list/mentors = GLOB.mentors
+	if(!mentors.len && !activeAdmins.len)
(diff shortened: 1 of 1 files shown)


Runa-Dacino:
I'll reopen this but I fucked up my git hard lmfao

PR #15816, opened 2024-03-10
Revert "Revert "Garbage collection, asset delivery, icon2html revolution, and general fixes (515)""

Heroman3003:
Reverts VOREStation/VOREStation#15815

Merge this when 515 update on box goes through

we're going deep down revert hole


Raeschen:
I'll continue porting runtimes ect after this is solved if you want me to, 's just more of them kinda require this now.

PR #15817, opened 2024-03-10
Enforces eol=lf for .swrc files thru .gitattributes

Runa-Dacino:
Also rebuilds tgui bundle just to be safe

.swrc was reported as not of the right format preventing building tgui. This should fix that.


Runa-Dacino:
I want selis to double-check if this works tho

PR #15823, opened 2024-03-10
hands over the round ID to TGchat

Kashargul:
Requires round IDs and DB subsystem

Downstream: Uncomment the round ID handover, we already can use it

🆑 
fix: handing over the round ID to the chat properly
code: optimizes chat export verb
/🆑


Kashargul:
Adding some other optimizations in here

ItsSelis (changes requested):
Requires tgui-panel bundle rebuild

Kashargul:
Bundle added.

PR #15844, opened 2024-03-18
Ports cooker changes

SatinIsle:
Ports cooker changes from RogueStar by VerySoft: https://github.com/TS-Rogue-Star/Rogue-Star/pull/173 Does not include everything (retains oven doors).

Added a "Toggle Safety" option to cooking appliances. This is off by default, can be toggled with a verb on the appliance. When on, instead of food burning when it reaches that point, it is instead ejected from the machine safe to eat.

Significantly sped up the speed that machines reach cooking temperature at. Generally reducing the wait time from about 10 minutes to 2 minutes.

Reduced cooking time of most food significantly.

Main goal of this to make it less frustrating to produce large spreads of food.

I want feedback from players who play chef or enjoy kitchen related RP in general, these numbers are easily tweaked.

### DIFF
--- a/code/modules/food/kitchen/cooking_machines/_appliance.dm
+++ b/code/modules/food/kitchen/cooking_machines/_appliance.dm
@@ -41,6 +41,7 @@
 	var/container_type = null
 
 	var/combine_first = FALSE // If TRUE, this appliance will do combination cooking before checking recipes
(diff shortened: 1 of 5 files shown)


Heroman3003:
If there are any problems with the new balancing, they can be brought up after this is live. For now let's just call it experimental.

PR #15849, opened 2024-03-19
Adds MultiZ chat filter

Heroman3003:
Adds separate chat filter that blocks out all says, emotes, whispers and subtles that come from z-level that isn't your own. Doesn't affect stuff like radio and such.

Currently overrides regular say filters, but its obscure enough to not cause too much issues and easy enough to just toggle off where unneeded.


Runa-Dacino (commented):
Tested on my end on SD with 2 chars. Works as expected.

I presume the override means it overrides stuff like npcsay and stuff right? Can you give some examples of what the obscure cases would be? Likely as you say, not an issue but prolly good to have it explicitly documented somewhere.

Otherwise me like.

Heroman3003:
> Tested on my end on SD with 2 chars. Works as expected.
> 
> I presume the override means it overrides stuff like npcsay and stuff right? Can you give some examples of what the obscure cases would be? Likely as you say, not an issue but prolly good to have it explicitly documented somewhere.
> 
> Otherwise me like.

Yeah, that's the exact thing. Again, don't think it's problematic. Although worth noting, npcsay WILL take priority over it, and then all 'normal' says and emotes will be after it.

PR #15858, opened 2024-03-22
Added a Kelpie mob

SatinIsle:
Added a new Kelpie mob, based on the large horse sprite. This mob is significantly more likely to eat people and harder to escape if you have digestion enabled. It has a unique trait of diving underwater with it's prey (if on an appropriate tile), or surfacing when it has no prey.

image


Heroman3003:
👀

PR #15859, opened 2024-03-22
Gives simple mobs thermals

SatinIsle:
Added a verb to all simple mobs controlled by players that lets them see mobs through walls for one minute, with a five minute cooldown. Intent is to make it so that maint preds can easily track down prey outside of maintenance without having to expose themselves.

### DIFF
--- a/code/modules/mob/living/simple_mob/simple_mob.dm
+++ b/code/modules/mob/living/simple_mob/simple_mob.dm
@@ -175,6 +175,8 @@
 	// VOREStation Add End
 
 	var/has_recoloured = FALSE
+	var/hunting_cooldown = 0
+	var/hasthermals = TRUE
 
 /mob/living/simple_mob/Initialize()
 	verbs -= /mob/verb/observe
@@ -222,7 +224,8 @@
 /mob/living/simple_mob/Login()
 	. = ..()
 	to_chat(src,"<b>You are \the [src].</b> [player_msg]")
-	verbs |= /mob/living/simple_mob/proc/hunting_vision //So that maint preds can see prey through walls, to make it easier to find them.
+	if(hasthermals)
+		verbs |= /mob/living/simple_mob/proc/hunting_vision //So that maint preds can see prey through walls, to make it easier to find them.
 
 /mob/living/simple_mob/SelfMove(turf/n, direct, movetime)
 	var/turf/old_turf = get_turf(src)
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/simple_mob/simple_mob.dm
```diff
@@ -354,3 +355,24 @@
...
+	if(hunting_cooldown + 5 MINUTES < world.time)
+		to_chat(usr, "You can sense other creatures by focusing carefully on your surroundings.")
+		sight |= SEE_MOBS
+		hunting_cooldown = world.time
+		spawn(600)
+			to_chat(usr, "Your concentration wears off.")
+			sight -= SEE_MOBS
+	else if(hunting_cooldown + 5 MINUTES > world.time)
+		to_chat(usr, "You must wait for a while before using this again.")
+
```
Runa-Dacino, line 378: While the |= verb addition I believe should be defined on the subtype, I think it's fine to define the proc itself higher in the types on simple_mobs proper as it allows adding them to mobs on a case by case basis by GMs. I just don't think it should be on non-scene mobs.

code/modules/mob/living/simple_mob/simple_mob.dm
```diff
@@ -222,6 +222,7 @@
 /mob/living/simple_mob/Login()
 	. = ..()
 	to_chat(src,"<b>You are \the [src].</b> [player_msg]")
+	verbs |= /mob/living/simple_mob/proc/hunting_vision //So that maint preds can see prey through walls, to make it easier to find them.
```
Runa-Dacino, line 225: As I've later written, I believe making this verb addition happen on /mob/living/simple_mob/vore might be a better idea as we're intending for this to be a scene tool. This likely will not cause us issues, but I believe it's best to restrict it to the scene mob subtypes by default.

This is more or less for sake of downstreams.

Alternatively, you could add a var/ShouldHaveThermals = FALSE on simple_mob then change this to

`
if(shouldHaveThermals)
   verbs |= /mob/living/simple_mob/proc/hunting_vision
`

Then modify /mob/living/simple_mob/vore to have this var as true.

I think this might be the best approach to this, as we can fine tune down the line whether we want a mob to have it or not. Downstream can also make their own edits.

Runa-Dacino (changes requested):
Hello!

I like the concept of this scene tool but I have some comments, see below.

- One of them is to make it modular, as detailed later. This is a **Suggestion**
- The other is to please don't re-define a type and add a var to it orphanned from the main definition. This is a **Change Request**

Runa-Dacino (approved):
Looks fine by me now, has my kudos.

Only thing left is maybe discussing internally what we definitely want to have as FALSE by default and waiting for that before merging.

PR #15876, opened 2024-03-28
Added Post-Drain belly modes

SatinIsle:
Added "Drain Finishing Modes" that kick into effect once a drain belly has extracted all of the nutrition that it can out of a prey character. These are found in the options tab of the vore panel bellies.

These modes are currently:

- Normal: Current behaviour, just stops when it runs out of nutrition to extract. 
- Sleep: Slowly increases the prey's tiredness until the fall asleep. Be warned that they won't be able to post or see posts when they fall asleep. 
- False Sleep: Slowly increases the prey's tiredness until just before they fall asleep, darkening their screen but NOT preventing them from posting and reading posts. 
- Weight Drain: You're done draining their nutrition? Well, time to start taking their weight directly and adding it to your own body. Is dependent on the prey's weight loss % chosen in character setup.

This adds a tiredness value to mobs that is only used for these drain modes. It naturally decreases over time as part of the mobs life cycle, so it will take a little time to clear if the drain mode is deactivated or they leave the belly.
These modes only work on carbon mobs, they do not work on simple mobs.

Part 2:

Added a "Knockout" process option to prey in vore bellies. It adds 500000 sleeping to the prey, with the usual confirmations.
Changed ejecting prey to reduce their sleeping to 20 so that they will wake up shortly after.
Changed OOC Escape to instantly remove sleeping.


Heroman3003:
Needs to be updated from master and have bundle recompiled

SatinIsle:
> Needs to be updated from master and have bundle recompiled

Hopefully fixed now!

PR #15885, opened 2024-03-29
Quick Radio Fix for Eggnog Town

Rick4242:
~~This probably should have been done sooner~~

Adds an extra copy of the communications relay on the above-ground level of the eggnog town map to the underground level. Radio and PDAs _(should)_ work on both levels now _(I hope)_

image
image

_~~Now if only I could figure out how to make commlinks and NIFs work on redgate maps too~~_


Rick4242:
Quite frankly not sure how that other commit for the hotsprings got thrown into this PR, but not sure how to remove it... Some help would be nice.

Rick4242:
Going to close this, and then remake it in a moment, apologies

PR #15893, opened 2024-03-31
Ports "No Shoes" from RogueStar

SatinIsle:
Added the ability to choose to spawn with no shoes in character setup. This was ported from RogueStar: https://github.com/TS-Rogue-Star/Rogue-Star/pull/196

### DIFF
--- a/code/datums/outfits/outfit.dm
+++ b/code/datums/outfits/outfit.dm
@@ -127,6 +127,8 @@ var/list/outfits_decls_by_type_
 	if(gloves)
 		H.equip_to_slot_or_del(new gloves(H),slot_gloves)
 	if(shoes)
+		if(H.client?.prefs?.shoe_hater)	//RS ADD
+			return
 		H.equip_to_slot_or_del(new shoes(H),slot_shoes)
 	if(mask)
 		H.equip_to_slot_or_del(new mask(H),slot_wear_mask)
--- a/code/datums/outfits/outfit.dm
+++ b/code/datums/outfits/outfit.dm
@@ -127,9 +127,8 @@ var/list/outfits_decls_by_type_
 	if(gloves)
 		H.equip_to_slot_or_del(new gloves(H),slot_gloves)
 	if(shoes)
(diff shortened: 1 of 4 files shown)


Heroman3003:
//RS add tags arent really necessary here, since they seem to be for upstream compatibility, here they just create a bit of a mess.

SatinIsle:
> //RS add tags arent really necessary here, since they seem to be for upstream compatibility, here they just create a bit of a mess.

This was something I was told by Lira that we need to keep for the conditions of our license, to note changes to the code and where they come from. I don't know any better to say otherwise, which is why I keep them in.

Raeschen:
I tested this upstream and it doesn't appear to work. Does it work for you? 

Tested: roundstart spawn, latejoin spawn, also ghost/mannequin doesn't show the change.

SatinIsle:
Worked on spawning via Spawn Character too

SatinIsle:
And observer seems to work too: 
image

I don't know if it could be a minitest thing that's letting it work here, I'll have to compile with the tether or something.

Raeschen:
For me i tested specifically the job gear (engineer), not loadout shoes. Also, of a teshari that was being equipped with sandals.

SatinIsle:
> For me i tested specifically the job gear (engineer), not loadout shoes. Also, of a teshari that was being equipped with sandals.

Yeah, I also tested it with job gear (bartender for me), I just happened to also have a corset on from another test here, lol

SatinIsle:
Mhm, just tested a tesh too and they start with sandals still, probably something in teshari code due to the fact they can't spawn with normal shoes.

SatinIsle:
Yeah, found it, will PR that now

Raeschen:
What about observers/previews?

SatinIsle:
> What about observers/previews?

Observers didn't work, I think in my test it's just because I'd left the round without shoes before testing that. But I don't really care about this working.
Previews it doesn't work on. It's not important to me, but I'd like it to work, though how previews work is a bit beyond my skillset atm.

PR #15902, opened 2024-04-07
cell div by 0

Kashargul:
We have some random spawn weapons in xenoarch that spawn with a 0 cell. Really shouldn't cause divisions by 0 with it because the cell procs are unsafe.

🆑 Upstream
fix: division by 0 on powercells
/🆑


Kashargul:
Active edges -> Checks, Rerun failed jobs

PR #15913, opened 2024-04-09
Ports : ability to dominate mobs and ghosted player bodies via dominate predator trait

Delams-The-SM:
Ports my PR from chomp station that I feel like might be interesting with the new redgate jungle.

https://github.com/CHOMPStation2/CHOMPStation2/pull/7793

Important to note, due to differences in maps, mobs they are on and so on I haven't ported over the mind binder whitelist for dominate-able mobs, should such whitelist be required obviously I will work on it.
It should be noted due to the way the dominate predator trait works it is only possible to dominate mobs with a belly.

Interactions :

player vs mob : domination, still asks if prefrences allign

player vs player (in body) : asks both parties
player vs player (client closed) : domination is declined
player vs player (ghosted or in some other way not assigned to body) : domination, unique log to admins "X is taking control over Y while they are out of their body."

player vs player mob (Maintpred spawn etc) : same as above for pvp

### DIFF
--- a/code/modules/mob/living/simple_mob/subtypes/vore/dominated_brain.dm
+++ b/code/modules/mob/living/simple_mob/subtypes/vore/dominated_brain.dm
@@ -293,7 +293,7 @@
 			if(db.ckey == db.pred_ckey)
 				to_chat(src, "<span class='notice'>You ease off of your control, releasing \the [db].</span>")
 				to_chat(db, "<span class='notice'>You feel the alien presence fade, and restore control of your body to you of their own will...</span>")
-				if(db.was_mob==1) //CHOMPEdit start - clean up if the dominated body was a playerless mob
+				if(db.was_mob==1) //clean up if the dominated body was a playerless mob
 					db.pred_ckey=null
 					db.ckey=null
 					db.restore_control()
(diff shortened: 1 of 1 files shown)


Kashargul:
Do not port CHOMPEdit, CHOMPAdd or CHOMOPRemove comments to upstream.

Delams-The-SM:
sorry must have slipped through, removed now

Heroman3003:
I don't like either the mob or ghosted player domination.
The former looks like its just asking for trouble with somebody rushing to force feed themselves to OP mob to puppeteer it, I definitely dont want to see that happening.
The latter is literally prefbreak ability, why is it even being proposed.

Heroman3003:
Yeah, staff opinion of this is overwhelmingly negative, so I'll just be closing this. Just. No.

Novacat:
To further explain things. The idea of dominating ghosted characters is something that was rejected outright because of the potential for prefbreak.

The idea of dominating mobs was largely rejected because we already have mechanisms in place to join as specific mobs, the previous mechanisms already has mechanics in place so we can choose what mobs you can and cannot join as. This would remove control of that, allowing potentially any mob to be puppeteered and brought to the station. We do not really want that. My personal experience is that if a mechanic is in place that could be abused by the players, it absolutely will be abused.

Raeschen:
for context, we enforced addition of a whitelist upstream. this seems to of skipped that.

PR #15924, opened 2024-04-14
some borg upgrade overhaul, power use function, bluespace pounce and some fixes

Kashargul:
The sight modules are mostly meant for events to give borgs different view options in a quick modify robot. They still need to equip the module though. So it's a selective option. (The anomalous one is similar to thermal for seeing invisible level 2 mobs, but without seeing through walls. Also anomalous sight gives borgs the bluespace pounce, including a glowy pounce icon..., higher range, leaping through objects at a higher power cost, instant pounce without miss chance. For downstream, it will dephase shadekin)

🆑 Upstream
add: adds some more upgrade modules to give different vision flag modules to borgs. currently admin only modules
add: bluespace pounce, pounce can be upgraded with the anomalous sight, increasing range (from 3 to 5 fields), allowing to pounce through objects, but at a higher power cost (from 750 base to 1000), bluespace pounce will be instant and won't miss.
add: vorespan defines to the chat spans
fix: icon for borg x-ray vision
fix: a rare case in which the store HUD icon could be overwritten by throw mode
refactor: new function (use_direct_power(amount, lower_limit)) to directly use power for borgs with a lower power limit to prevent full self-discharge. Power is now used before the action.
code: ports some defines upstream, just to prevent compatibility issues in the future, so that no one uses them for something else
code: reworks borg upgrades again to clean them up some more
balance: pouncing phased shadekin now costs an additional 1000 charge
balance: robot shields now overload when the power cell has 200 charge remaining instead of fully depleting the robot
/🆑 

Downstream: Don't merge right away. It needs some things added for us before merging.


Kashargul:
Back to draft, adding full functionality even if not used as well as refactoring power usage throughout all robot items, the current way is really urg...

Heroman3003:
While the designs are no_prod, I will note that the shadekin-vision being in any way player accessible without admin magic is 100% not something desirable for the future.

Kashargul:
> While the designs are no_prod, I will note that the shadekin-vision being in any way player accessible without admin magic is 100% not something desirable for the future.

This is same for downstream, it's an admin only item as var edit does not work for borgs due to how vision flags work it gets instantly overwritten.

PR #15925, opened 2024-04-15
Upport Teshari Synth Face mask from Chompstation

ReoDaProtovali:
I gotteded the code and it was kinda mid so I also made it better and this will hopefully trickle down to all of the downstreams :DDDD

Anyhow for real it upports the teshari synthface. Maybe I'll do more with the bit of refactors I did

Isnt this part for chompstation maybe I shouldnt do this part
:cl:
add: Teshari Face Mask from chompstation
add: Facial protection to the mask, since it cant be removed
fix: Fixed an old `to_chat` and also added framework for adding more face states to the mask
:/cl:


Heroman3003:
actually seems like its broken and is not included as was reported. derp.

PR #15928, opened 2024-04-16
Opens more positive and negative traits to non-custom species.

SatinIsle:
Added access to various negative and positive traits for non-custom species, where it makes sense that there might be variations due to natural differences and body modifications. I have excluded specific species from accessing most of these where they are already meet those levels or would be buffed too much and lose their balanced flavour.
Whilst I realise there is a system in place to stop negatives making things stronger, I have excluded these species anyway so people don't accidentally waste points on them.

Specifically:
### **Positive traits:**
- Hardy and Hardy+ (excluding Tesh, Prommie and Protean to avoid balance. Alraune, Unathi and diona already have this).
- High Endurance (excluding Tesh and BESK to avoid buffs, and Unathi who already have this).
- Darksight and Darksight+ (excluding all those who already get to see in the dark)
- Brute resist minor only (exlcuding Tesh for the buffs and those that already have this level of resistance).
- Grit

### **Negatives:**
- Slow and Slow+ (excluding those that already move so slow)
- Weakling and weakling+ (excluding BESK and tesh who are already this weak)
- Low endurance and + (excluding besk for the first and tesh for both as they're already that weak)
- Brute weakness minor, normal and major (excluding those that are weak to specific levels)
- Neural Hypersensitivity

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/traits_vr/negative.dm
+++ b/code/modules/mob/living/carbon/human/species/station/traits_vr/negative.dm
@@ -6,30 +6,40 @@
 	desc = "Allows you to move slower on average than baseline."
 	cost = -2
 	var_changes = list("slowdown" = 0.5)
+	banned_species = list(SPECIES_ALRAUNE, SPECIES_SHADEKIN_CREW, SPECIES_DIONA, SPECIES_UNATHI) //These are already this slow.
+	custom_only = FALSE
 
 /datum/trait/negative/speed_slow_plus
 	name = "Slowdown, Major"
 	desc = "Allows you to move MUCH slower on average than baseline."
 	cost = -3
 	var_changes = list("slowdown" = 1.0)
+	custom_only = FALSE
+	banned_species = list(SPECIES_DIONA) //Diona are even slower than this
 
 /datum/trait/negative/weakling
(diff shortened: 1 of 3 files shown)


Heroman3003:
I'll be honest I'm not all too sure bout those, mostly haste. Also ideally, rather than manual blacklist, there'd be a way to detect if a stat deviates from default and prevents species from taking all traits related to that stat separate from general blacklist, since balance changes can happen and traits would be forgotten to be adjusted, but I'm not sure how in-scope that would be. Thats the main reason I chose the traits I chose when adding all-species traits tbh, to avoid need for that.

SatinIsle:
> I'll be honest I'm not all too sure bout those, mostly haste. Also ideally, rather than manual blacklist, there'd be a way to detect if a stat deviates from default and prevents species from taking all traits related to that stat separate from general blacklist, since balance changes can happen and traits would be forgotten to be adjusted, but I'm not sure how in-scope that would be. Thats the main reason I chose the traits I chose when adding all-species traits tbh, to avoid need for that.

Removed haste from the options, that was the main one I wasn't too sure whether to include or not anyway.
The other portion, I agree it's a bit out of scope but may be a useful feature to add in another PR later.

PR #15949, opened 2024-04-28
Expanded size trait options

SatinIsle:
Added new "Tall, Major" and "Short, Major" traits that add and reduce the sprite's height by 15% respectively. As opposed to tall and short, which are about 9%. The short, Major trait does not make someone who is 25% scale any smaller than the short trait, so no problems with balance there.

Added new "Heavy Frame, Major" and "Light Frame, Major" traits that increase and reduce your threshold for size interaction respectively.


Heroman3003:
Seems fine.

PR #15950, opened 2024-04-28
Changes wall damage from being (usually) 25 brute to 12 and doesnt add a stun to it

ReoDaProtovali:
25 damage for a wallthrow is pretty extremely high. That's higher than some low calibre bullets! It also turns bones into powder because your bones are made of glass. This makes the damage lower but it has a stun when you hit wall because oof ouchy. I dont think "speed" is ever passed in so the default value of 5 was always used. Consistency is also nice

### DIFF
--- a/code/game/turfs/turf.dm
+++ b/code/game/turfs/turf.dm
@@ -301,9 +301,10 @@
 
 // Called when turf is hit by a thrown object
 /turf/hitby(atom/movable/AM as mob|obj, var/speed)
-	if(src.density)
-		spawn(2)
-			step(AM, turn(AM.last_move, 180))
+	if(density)
+		if(!has_gravity(AM)) //Checked a different codebase for reference. Turns out it's only supposed to happen in no-gravity
+			spawn(2)
+				step(AM, turn(AM.last_move, 180)) //This makes it float away after hitting a wall in 0G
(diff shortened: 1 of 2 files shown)


Heroman3003:
Damage reduction is fine, but I don't think stun is a good idea.

ReoDaProtovali:
it's an extremely short stun for being THROWN into the wall, you get a stun for stumblewalking into walls, windows and pretty much everything else that isnt a maw. I think a stun is absolutely fine

Heroman3003:
General staff consensus is that rather than full stun this should at most be a short slow or confusion.

ReoDaProtovali:
should that be changed for stumbling too, then?

Heroman3003:
No, specifically for the throws. Otherwise throws can result in complete easy stunlock from externals sources, while stumbling is purely self-inflicted. And if someone gets confused from being slammed into wall and then also smacks into it, then its comedy.

ReoDaProtovali:
I could decrease the amount of time, but I think being launched into a wall, and just getting dizzy from it both doesnt make sense, and could result in running back into the wall immediately, which is a pretty similar issue, I think

Also plenty of things cause dizziness that isnt in control

ReoDaProtovali:
stun is gone

PR #15963, opened 2024-05-05
Ducking silly

SatinIsle:
Added 3 more admin spawn rubber duckies (grey, green and pink).

Added a variable to rubber duckies called "honk text" that allows it to put an audible message in chat when honked.


Heroman3003:
I love the pink one

PR #15964, opened 2024-05-05
Adds inhand sprites for rubber duckies

TheToaster98:
What says on the title, adds a sprite for each duck, yes, the normals AND the seven admemes.
This is my first time touching sprites at all, so please let me know if it could be improved and I will get to that.


TheToaster98:
I had the feeling it was redundant but forgor to actually check it, will get to that.

PR #15982, opened 2024-05-16
AFK Status/Toggle

KillianKirilenko:
Adds a verb under the IC tab that toggles off an overhead 'AFK' indicator, so you can easily mark yourself as such if you need to hop up and run to the bathroom or something!

I probably could've rigged this into the existing system for inactivity as well, but I wanted a manual option for it first and foremost.


SatinIsle:
Great idea. I think it is good to have started as a manual toggle for now as you have, I'll definitely be using it myself.

PR #15984, opened 2024-05-18
Event signs

SatinIsle:
Added some simple plaque-like signs to be mapped in for events. The main thing I have in mind for them is to be used for OOC notes or something similar for event characters outside of the area that they are actually in.


code/defines/obj.dm
```diff
@@ -20,6 +20,23 @@
...
+	density = FALSE
+
+/obj/structure/signevent/blue
+	icon_state = "sign_event_blue"
+
+/obj/structure/signevent/red
+	icon_state = "sign_event_red"
+
+/obj/structure/signevent/yellow
+	icon_state = "sign_event_yellow
```
Heroman3003, line 38: forgor "

SatinIsle: Thank you!

PR #15997, opened 2024-05-22
Auto-AFK

KillianKirilenko:
Automatically sets the AFK status after ten minutes (same as the standard inactivity timer), and automatically *un*sets it on activity *if* you didn't manually mark yourself as AFK. New Automatic AFK pref that defaults to on, but if set to off won't mark you with the AFK bubble.

Examining someone will also show "away from keyboard for [time]" if manual AFK was set, or "inactive for [time]" if not. Having the auto-afk pref disabled *does not* prevent you from being flagged as inactive on examine.


SatinIsle:
I'd personally like the automatic AFK flag to be a toggle, because I personally don't really want to have the indicator whilst I am doing something else in another window but actually still listening out for interactions, because it might discourage those interactions happening in the first place.

Heroman3003:
> I'd personally like the automatic AFK flag to be a toggle, because I personally don't really want to have the indicator whilst I am doing something else in another window but actually still listening out for interactions, because it might discourage those interactions happening in the first place.

Agreed, that should be a pref IMO

PR #16013, opened 2024-05-29
allow to only show suitable belly messages

Kashargul:
🆑 Upstream
qol: allows to hide non suitable message settings in the Description section
/🆑 

@ Downstream, do not merge, needs to be implemented manually!


Kashargul:
Active edge in integration. Re-run test.

PR #16018, opened 2024-05-31
fix lavender seed icon

Kashargul:
Also, Up ports some seed icons for tea as I've currently not the time to edit the subsystem to properly work with our _ch files.

🆑 Upstream
fix: lavender using an non-existing seed icon
/🆑


Kashargul:
Active edge in integration. Re-run test.

PR #16020, opened 2024-06-02
allows to use more newline breaks for longer posts

Kashargul:
Allows for up to 9 paragraphs. We strip leading and trailing new lines

Tg chat will not display more than 3 new lines in succession. Everything more will get filtered out before displaying the message. So

\n\n\n\n....

Will always only be shown as \n\n\n

This can be adjusted if needed, but otherwise, it prevents to just spam newlines.

Also, fixes a small oversight where the idle options where shown even if disabled in the vorepanel.

🆑 Upstream
qol: increases the amount of newlines allowed in 1 post from 6 to 18
qol: TG chat will limit displayed successive new lines to a maximum of 3
/🆑


Kashargul:
active edge...

Kashargul:
Gonna look into the regex again, as I might not plan to extend it to spaces / tabs anytime soon, might as well simplify it.

PR #16030, opened 2024-06-08
Persistent papers should not replace the chat window

### LINKED ISSUE #15956: [UI] Persistent text items breaking the chat window
Examining a sticky note, or a paper that is on a notice board, from a previous shift causes the text on the item to take over the chat window without any way to back out.
These items can not be found in the right click list of items on a turf.

NOTE: This only applies to items persistent from the previous shift and NOT to ones created on the same shift.

For a pop out window to appear just like examining the same items created on the same shift.

#### What actually happened
The text window was replaced with a white box containing the text from the item that was examined. There is no way to remove this box except for using the Fix Chat verb.

#### Steps to reproduce
- Examine a sticky note or piece of paper that is persistent from a previous shift.
- Right click on the turf containg the item to see that it is not displayed there.

#### Code Revision
Commit: [9219da9919fdccbb61484a6ae55b50ac56abef6f](https://github.com/VOREStation/VOREStation/commit/9219da9919fdccbb61484a6ae55b50ac56abef6f)
 

image
image

Kashargul:
the windows opened on paper examination are relying on the paper name. IF NULL is given, the chat window is replaced.

fixes #15956

🆑 Upstream
fix: persistent papers no longe replace the chat on examine
/🆑

### DIFF
--- a/code/modules/persistence/effects/paper.dm
+++ b/code/modules/persistence/effects/paper.dm
@@ -18,7 +18,7 @@
 	paper.info = token["message"]
 	paper.name = token["title"]
 	if(!paper.name)
-		paper.name = "Unknown"
+		paper.name = "No Title"
 	paper.last_modified_ckey = token["author"]
 	paper.age = token["age"]+1
 	if(requires_noticeboard)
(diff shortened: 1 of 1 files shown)


Kashargul:
> "No Title" might work better than "Unknown", but good to know this one's finally being tackled.

Ye, I've even found a better way which ensures ingame sanity, pushing a commit in a few

PR #16036, opened 2024-06-10
assign account to nifs

Kashargul:
🆑 Upstream
fix: nif duplication exploit through sharing, as well as an issue to override your own nif with other's
/🆑 

Not tested yet... and someone knowing nifs should give it a look, not that I forgot some setup


Heroman3003:
One potential issue I see is "What if someone is disconnected when NIF is being calibrated and assigned ckey" but looks good otherwise.

Kashargul:
> One potential issue I see is "What if someone is disconnected when NIF is being calibrated and assigned ckey" but looks good otherwise.

You mean the device getting installed in round and disconnecting before the initialization finished?

Heroman3003:
> You mean the device getting installed in round and disconnecting before the initialization finished?

Yes, that case is what I mean.

PR #16037, opened 2024-06-10
New event species - Part 1

SatinIsle:
(Well, so much for hoping the bot would cut off the below text on discord, lmao)

Added a new species called Lleill, a heavily fae inspired species found in redspace. Heavy focus on transformations. They are shapeshifters with the ability to transform objects into something they call "glamour". Added a handful of new glamour objects, some of which are just resprited existing items, but with a few new ones too.

This so far has all been tested and appears to work okay on my test server. The species functions perfectly fine as it is now and this should be merged this week ideally ahead of events planned from next week onwards.
This species RIGHT NOW, is only intended to be played as event characters by staff and therefore has not been balanced. My plan now is to create an ability resource for these creatures and limit the types of items that can be transmuted, so that once they have been encountered in events, players could ask to be able to play one for a shift here and there without being stupidly powerful.

Also, extra mention to Kashargul for helping me fix some runtimes with the shapeshifting, so this PR also includes a fix for a runtime in get_tail()

With further testing, I'm not running into any issues. I'd like this merged this week if possible so that I can run the relevant event next week.


Heroman3003:
Event only stuff so if something is broke nbd

PR #16042, opened 2024-06-14
Some tgui stuff, lavender again...

Kashargul:
seemingly the product and seed name need to match... so just putting lavender back to the original flower6 and renaming the seed not to cause further issues...

🆑 Upstream
fix: lavender products should show again
qol: For string formatting, toFixed should be used and not round!
/🆑


Kashargul:
active edge

Kashargul:
I forgot a UI that can be funky due to backend interactions to suddenly show extremely small decimals... 

Todo me: use toFixed(val, 2) for colorMate

Kashargul:
> weird conflict

due to mapswap, moment

PR #16048, opened 2024-06-17
Make Patches Useful(?)

KillianKirilenko:
Patches are a really cool idea as a non-invasive treatment alternative to syringes and pills, but they're locked behind the requirement of using the distillery which means going through cargo and doing the whole process of fucking around with that.

So instead I've just made them transfer reagents in the patch into the bloodstream. This is already possible IRL using compounds like [DMSO](https://en.wikipedia.org/wiki/Dimethyl_sulfoxide#Medicine).

Might mess around with them later and see if I can make it a progressive transfer thing rather than just slap-on-and-done, since otherwise they *are* kind of objectively better than syringes and pills alike. Possibly cap the max reagent amount to be lower as well?

This does sort of make all the -aze reagents redundant, but was anyone really using those anyway? It doesn't help there's only five of them, covering inaprovaline, bicaridine, dermaline, tricordrazine, and a custom antibiotic/painkiller combo.

:cl:
tweak: made patches transfer their reagents straight into blood, so you can use them for anything
/:cl:

### DIFF
--- a/code/modules/reagents/reagent_containers/patch.dm
+++ b/code/modules/reagents/reagent_containers/patch.dm
@@ -40,7 +40,7 @@
 			to_chat(H, "<span class='notice'>\The [src] is placed on your [affecting].</span>")
 			M.drop_from_inventory(src) //icon update
 			if(reagents.total_volume)
-				reagents.trans_to_mob(M, reagents.total_volume, CHEM_TOUCH)
+				reagents.trans_to_mob(M, reagents.total_volume, CHEM_BLOOD) //CHEM_TOUCH
 			qdel(src)
 			return 1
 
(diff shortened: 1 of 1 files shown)


ReoDaProtovali:
What happens to chems that are supposed to be touch-applied opposed to bloodstream applied? Not that we have a strong chemistry system that actually encourages different application types, but if we were to ever overhaul chems, this would be an extremely unforseeing change that makes touch-apply chemicals much harder to use in a sensible way without just splashing yourself with a beaker

Also, I dont know how the distillery works, but why is it a requirement to using patches here? Are the only touch-chems we have here locked behind it or something?

Heroman3003:
I'd like something done about the -azes so they aren't a noob trap, but maybe as separate PR.

Heroman3003:
> > I'd like something done about the -azes so they aren't a noob trap, but maybe as separate PR.
> 
> What's the vibe then? Remove? Repurpose?

Maybe specifically make them "work better than regular when applied in patches rather than injected" somehow, maybe made unobtainable. Just not the current result of "this says its a medicine but when you try using it, you poison people"

PR #16062, opened 2024-06-19
Bluespace Crackers

SatinIsle:
Added a bluespace cracker item. It works similarly to a normal cracker, in that one person offers it to another to pull the cracker together. When pulled, the winner is rewarded with a toy, a joke, a cheap paper hat and a powerful bluespace effect!

The effects include:
Shrinking to 25% scale
Growing to 200% scale
Having a drug like effect colour their screen.
Making them invisible for one minute.
Knocking them over.
Teleporting them into the losers active belly (if prefs allow!) Or giving them a bunch of money.

They can be found in the bluespace random item pool and rarely as an item in maint!
There are also variants with names revealing specifically which effect it will have.


Heroman3003:
Suggestion: ability to rig crackers using varedit, so that user offering them always wins/loses

SatinIsle:
> Suggestion: ability to rig crackers using varedit, so that user offering them always wins/loses

Done!

PR #16090, opened 2024-07-05
Choose to rest left or right

SatinIsle:
Added new hotkey mode combinations for resting. Ctrl+U now always rests to the right, whilst Shift+U always rests to the left. These can also be used with the Rest-Left and Rest-Right verbs in the command bar.


Novacat (approved):

Heroman3003:
A way to 'reset' it to random would be good.

PR #16118, opened 2024-07-21
Char Setup Background Switch

KillianKirilenko:
Removes the dark tiles, empty plating, and reinforced plating from character setup backgrounds, adding desert, grass, and snow as alternatives, for colour/contrast test options besides shades of grey.

:cl:
tweak - switched up the character setup preview backgrounds a little
/:cl:


SatinIsle:
I'm not too bothered what the defaults are, but I'd still quite like to have the featureless voids as options. I find them far better for actually putting together the sprite to start with without the distracting textures outside of it.

Heroman3003:
Yeah, I know voids/plating are not ALWAYS useful, but they can be in rare cases and I'd rather not see them go.

PR #16127, opened 2024-07-24
Lleill Species - Part 2

SatinIsle:
This PR is primarily to bring the lleill species to a point where players could reasonably request to play as them, having some semblance. This one is ready now! There will likely be a part 3 if I decide to give it proper HUD icons and make the abilities and energy generic.

The following tasks need to be completed:
- [x] Added an energy system for some Lleill abilities (invisibility, transmutation and teleportating to rings).
- [x] Added ways to restore that energy. There are now two ways, either 75 energy every 10 minutes from rings, or by taking energy from another creature that consents to it.
- [x] Changed glamour rings to allow them to be destroyed by either the Lleill or another character. The Lleill is notified who has destroyed their rings (for fae antics).
- [x] Added a glamour gasmask/helmet option for Lleill to use for breathing outside of the tether.
- [x] Added a variant of the nudity permit of Lleill to use for storing items.
- [x] Added an alchemy ability to allow Lleill to do alchemy without using an alembic. Not very useful but very flavorful.
- [x] Added two new Lleill ears and two new Lleill tails.
- [x] Fixed the formatting of transmuting item text.
- [x] Fixed the formatting of the glamour shard item name.
- [x] Fixed cancelling the glamour face ability spawning faceless homunculus.
- [x] Fixed invisible Lleill having the AR HUD show above them.
- [x] Fixed speaking glamour to translate from glamour speak.


SatinIsle:
> > Add an energy system for some Lleill abilities (invisibility, transmutation and teleportating to rings).
> 
> At this point, between shadekin and the lleill both having energy systems, I would suggest moving the basic energy system to the human level and then modifying the specifics on a per-species basis.
> 
> This would come with the advantage of potentially allowing staff -- and maybe even rare items -- to grant access to powers on a temporary basis for any given species that happens to be part of an event, via verb-granting commands and varediting.
> 
> Might be out of scope for this PR in particular, but food for thought either way.

I agree on the whole!
For the Lleill abilities, I may just do exactly that. Though I won't personally touch anything for the shadekin side of things, as I don't want to step on toes.

PR #16128, opened 2024-07-24
tgui update

Kashargul:
🆑 Upstream
code: Ports the latest changes on most of  the main elements from TG (NumberInput only had some slight changes, TG's current version is not user friendly)
code: adjusts our uis to be compatible 
/🆑 

@ Downstream, do not merge right away, needs edits on our UIs first!


Kashargul:
Active edge.

PR #16131, opened 2024-07-26
edits some no_vore usage

Kashargul:
makes no sense to init on no_vore or to re init on bots on logout when they are not meant to have vore without clients.

🆑 Upstream
code: removes some unnecessary lines and returns the init_vore proc on no_vore
/🆑


Kashargul:
active edge.

PR #16133, opened 2024-07-28
allows to disable sliders again

Kashargul:
No bundle needed as we only touch the props.

Allows to disable the slider element again.

can be closed downstream, already fixed for us.


Kashargul:
active edge...

PR #16148, opened 2024-08-02
Add bapi-dmm to reduce startup time by... a lot

ShadowLarkens:
This adds my project [bapi-DMM](https://github.com/tigercat2000/bapi-dmm) to replace the dmm_suite maploader with a dll written in rust. 

bapi-dmm is hilariously fast compared to the unoptimized version of dmm_suite currently used here, taking ~1 second to load all of the maps where the current suite takes ~30 seconds. 

Additionally, I added a little extra thing to replace /datum/random_map/automata/proc/generate_map, taking 65 *milliseconds* to do what softcode takes 16.46 *seconds* to do.

Note that bapi-dmm is *required* to run the server in this PR - maploading and automata generation will completely fail if it is not present. It should be considered just as necessary as rust_g. This will require a lot of testing to make sure the server and all devs are happy and it works consistently!

Additional note: This removes rotation from the maploader. It already doesn't work with tgm files and is therefore almost completely useless.

On my machine, these changes sums up to going from 200-144 second to ~112 second inits, though it's inconsistent to measure that way (as you might be able to tell by the fact that my before times are between *200* and *144* seconds...)

TODO:
- [x] Change all map templates to use string paths instead of file paths (`"` vs `'`) - bapidmm cannot load maps from byond file hooks and has to dump them out into the data directory, which takes a while.
- [x] Test on other devs machines
- [x] Test on live

### Startup Profile Before
before

### Startup Profile After
after

:cl:
tweak: Add bapi-dmm, a rust maploader, to reduce init time by 30-60 seconds.
tweak: Moved tracy init to before-world-New.
del: No more pretending to support map rotation 
/:cl:


ShadowLarkens:
@KillianKirilenko Fixed the tether bug - parser wasn't accounting for variables starting with underscores

ShadowLarkens:
[This has now been merged downstream at CHOMP](https://github.com/CHOMPStation2/CHOMPStation2/pull/9201)

Kashargul:
Should be updated to latest version (windows, linux and rolled out on virgo, too)

PR #16160, opened 2024-08-05
TGUI Say

ShadowLarkens:
This adds the say modal from /tg/, affectionately labeled "tgui say". It is on by default, turning it off will result in going back to the way it was. 

The main features:
- Allows switching between different modes with tab/shift.
- It's pretty.
- Typing indicators have a new mode: thinking. This is automatically switched to when you are not actively typing.
- Command bar spying - Typing indicators will now come up (if turned on as normal) for typing in the command bar at the bottom too, only with relevant verbs.
- Shows you what radio channel you're going to speak into ahead of time.

Additional features that I added just for here:
- Shift-enter lets you do multiline input (later stripped for anything other than say and me)
- Shift-tab lets you cycle backwards through the modes.

:cl:
add: Advanced TGUI Say modal
tweak: Verb panel say/me/whisper/subtle have been renamed to "Say verb"/etc - and pop open the tgui say panel/tgui input/normal input instead of always using the native dialog.
imageadd: Every speech bubble now has an extra spinning/"thinking" state.
tweak: Added /tg/'s trait system, currently just used for indicating that you're in the typing state.
fix: TGUI dev server works again (keep_local_name = TRUE is really important, both for that and reliability)
tweak: Subtle/Whisper now has typing indicators by default - you can turn indicators off for these channels specifically with a new pref.
tweak: Typing indicators switched fully to /tg/'s system for them with traits and client integration and... stuff.

/:cl:


PastelPrinceDan:
i really like this, just got one question:
is the window extendable for ease of writing long Says/Mes/whatever? that's my main gripe from tgcode servers with that system, you pretty much have to click the verb rather than use hotkeys for those

ShadowLarkens:
> i really like this, just got one question: is the window extendable for ease of writing long Says/Mes/whatever? that's my main gripe from tgcode servers with that system, you pretty much have to click the verb rather than use hotkeys for those

yeah this is extra wide already and expands vertically when you put in enough text, plus you can hit shift+enter to do multiline input and scroll up and down with the arrow keys

I can make the heights even bigger if needed, but this felt good to me in brief testing

ItsSelis (approved):

Novacat (approved):
Going to give this one more day for other staff to review this, and if no serious concerns pop up I will be merging this.

PR #16161, opened 2024-08-05
Supply System Updates

KillianKirilenko:
This started as a simple random crate addition plan, but quickly became... *bigger*. But only slightly.

Initially I just wanted to add funny random corporate supply crate to the system so you could order Some Corporate Stuff In An Appropriate Box, but this necessitated a refactor of the actual delivery system to avoid daft matryoshka crate situations or other weirdness, amongst other things. So, it is now possible for things to be delivered *without coming in a crate at all*. As a proof-of-concept, this has been applied to the cargo tug + trolley and all large atmospherics canisters.

So far as I can tell this works flawlessly and has not disrupted any of the existing options in the slightest, though I should probably test it a bit more to be sure.

Minimal testing aside this is mergeable in its current state, but I would like some staff input on what should (and shouldn't) be available from these random corpocrates, and whether certain orders should spawn without being crated (and if, say, the canisters should be put back in crates for safety reasons).

This also throws in a small fix for a pizza order that I noticed was wonky.

:cl:
rscadd: added "random corporate supply" and contraband "special corporate supply" crates to cargo terminals; the former mostly has harmless stuff, whilst the latter can spawn all sorts of stuff from the random pool but can also include weapons (rarely).
tweak: refactored the supply spawning behaviour slightly, so crates are no longer mandatory. imagine the manifest comes stickytaped to the item, or whatever
tweak: supply consoles start tracking orders at 0 (so the first order is order 1, etc.) rather than a random number between 1 and 9000
fix: the random pizza pack crate now comes with the pizza delivery outfit and visor, deleting a duplicate pizza supply pack that was inaccessible anyway due to not having a defined name or container
rscdel: merged all _vr supplypacks into the main supplypack files
rscadd: new "full lid" 'decal' option for crates
tweak: added desatti catering, nukies energy drinks, and coyote salvage corp suit crates to random spawners
/:cl:


Heroman3003:
I assume its still a WIP, so not touching for now.

vorestation.dme
```diff
@@ -4341,7 +4326,7 @@
 #include "maps\southern_cross\items\clothing\sc_accessory.dm"
 #include "maps\southern_cross\items\clothing\sc_suit.dm"
 #include "maps\southern_cross\items\clothing\sc_under.dm"
-#include "maps\tether\tether.dm"
```
Novacat, line 4344: This should probably not be changed.

Novacat:
What can the Special Corporate Supply have?

PR #16173, opened 2024-08-08
Add recipe macro recording to reagent dispenser

ShadowLarkens:
As title states, this adds a /tg/ inspired system to all reagent dispensers (including the bar) that allows you to record a series of dispenser button presses and play them back.

:cl:
add: Reagent Dispensers now allow you to save a series of button presses as a recipe/macro and replay it. This is per-dispenser per-shift.
/:cl:


tgui/packages/tgui/interfaces/ChemDispenser/types.ts
```diff
@@ -8,6 +13,8 @@ export type Data = {
   beakerCurrentVolume: number | null;
   beakerMaxVolume: number | null;
   chemicals: reagent[];
+  recipes: { [key: string]: Recipe[] };
```
Kashargul, line 16: Wouldn't Record<string, Recipe[]> be easier?

ShadowLarkens: I guess it would be, I'm just used to doing it this way because it works

PR #16174, opened 2024-08-08
Fixes species blood color

Darlantanis:
Fixes species blood color, which is currently overridden by the default red blood. Selecting a species will now copy that species blood color, while the "reset" blood button now resets to the species' default.

This will not reset any blood colors currently in player save data, which must be manually reset if desired.

Downstream changelog:
:cl:
fix: Species blood color is used again
qol: Blood color reset button now uses the species' default
qol: Blood color button now has a colored dot
/:cl:


Darlantanis:
Done.

PR #16177, opened 2024-08-09
Feather Speech bubble!

crabcube:
A basic, if a liiiittle too big icon for speech bubbles. Just my first go, going to try and make a bunch more that are a bit smaller and better aligned! 
image
image


Heroman3003:
This looks way too huge as  a bubble. It will cover half the average character's head. If others think its fine, sure but holding off for now.

crabcube:
All good! Was my first effort and learning how to do! I'll cook up a few smaller ones as soon as I am restored to motivation.

Novacat (changes requested):
A redesign in general will be required, because it has come to our attention that this uses the same sprite as the feather item from Super Mario World.

PR #16178, opened 2024-08-10
Away Team/Pilot Supply Crates

Novacat:
- Adds some recently introduced stuff to the pilot/away team supply crates
- Adds an Away Team Leader supply crate
- Removes a duplicated bomber jacket in the pilot crate


Novacat:
I considered it, but I did not want to intermingle PRs and there is already an emergency beacon supply pack so an emergency beacon being excluded is not too bad.

PR #16193, opened 2024-08-15
Fix Electric Guitar

Lilly-Lira:
Incorrectly capitalized sound files prevent the electric guitar from working.

Draft until I test on Linux since Linux and windows servers treat sound files differently.


Lilly-Lira:
Confirmed this works on a linux server as intended!

PR #16214, opened 2024-08-25
Removed negative trait limit

SatinIsle:
This PR was made for staff discussion and should not be merged until staff agree on it.

Changed negative traits to no longer reduce the "traits left" count, allowing you to take as many as you would like. This allows people to potentially take up to five positive traits, provided that they take enough negative traits to gain the points.

Fixed non-custom species being able to completely ignore trait count and trait point limits.


Heroman3003:
GDI i was going through the PRs and didn't notice that one line. Pls mark DO NOT MERGE or somthing in the title xD

PR #16215, opened 2024-08-25
fix belly text sanity

Kashargul:
Do not merge currently, as the limits never worked before, a lot of people likely have longer messages currently!

🆑 Upstream
fix: fixes the belly examine / idle / etc message length sanity
fix: special characters like ' or % breaking for entries without delim or if exceeding the sanity limit
fix: emergency shutters will no longer act when the input is closed
code: examine messages had their length increased from 510 to 2048 characters
code: all other messages had their individual length increased to 1024 characters and now have a maximum length of 6144 characters
/🆑 

Previously, it was intended that examine messages had a length of 510 characters, idle messages 260 characters and all others 160 characters. Due to multiple bugs, it was possible to fabricate messages up to:

20480 characters!
Players easily could pass their custom messages up to 4096 characters as well without any hassle.
Even if the sanity would've worked, almost all messages would've been cut down to 160 characters!


Kashargul:
> Looks good codewise.
> 
> I feel like 2048 limit for stomach examine text is high and I generally see people with 1-2 line stomach examine text, but like if someone DOES want to do 2048 character then all the more power to them.
> 
> > code: all other messages had their individual length increased to 1024 characters and now have a maximum length of 6144 characters
> 
> This seems good to me. 1024 is a good amount and should probably cover 99% of messages people currently have (idle, struggle, etc). I can't imagine too many people have over that.
> 
> It would be ideal (if possible) to allow people to be grandfather'd in if they have text messages over this (as I can't imagine this is many people)

I've seen some downstream with a bit over 1k, but ye, 2k would be a lot, I doubt we'll hit that.

Need to test something later, I have a bad feeling about the way I took the iterating over...

Kashargul:
> Looks good codewise.
> 
> I feel like 2048 limit for stomach examine text is high and I generally see people with 1-2 line stomach examine text, but like if someone DOES want to do 2048 character then all the more power to them.
> 
> > code: all other messages had their individual length increased to 1024 characters and now have a maximum length of 6144 characters
> 
> This seems good to me. 1024 is a good amount and should probably cover 99% of messages people currently have (idle, struggle, etc). I can't imagine too many people have over that.
> 
> It would be ideal (if possible) to allow people to be grandfather'd in if they have text messages over this (as I can't imagine this is many people)

if someone really hits that cap and they get the message reflected, one could adminbus set it directly on the belly. I would leave such requests more to admin decision. Var edit, belly, edit idle messages, replace the string

PR #16219, opened 2024-08-26
/tg/ pref datums (part 1)

ShadowLarkens:
Highlights:
- Entire ass pref system from /tg/
- Fully replaces /datum/client_preferences with /datum/preference/toggle
- `Game Options` TGUI panel replaces all of the verbs and switches under the Global tab
  - Complete with searching functionality that /tg/ doesn't even have!
- JSON savefiles! Savefiles are automatically migrated from BYOND .sav to JSON.
  - This just makes it a lot easier to work with prefs at a code level, because savefiles are magic and ugly and eughhh, y'know?

I'm sorry in advance, this is a *huge* diff because of the client_preferences conversion.


Heroman3003:
com flicmt

PR #16221, opened 2024-08-27
Spice Smites

KillianKirilenko:
I was reminded of something I said I'd do the other day by #16220 having a Panic smite, so here we are. Spice smites!

"Harmless" mode spawns a bottle of space spice at the target's feet and gently sasses them.

"Extra Spicy" mode spawns and immediately detonates a teargas grenade at their feet instead. This does mean it is something of an "AOE smite" in that the spread of peppergas will hit anyone else standing too close!

This may gently conflict with 16220 due to the way the smite list is loaded/handled but it'll be easy to resolve.

The original concept was to make a single peppergas cloud spawn on top of the victim, but smoke reactions are fucky so I decided not to bother with that.

:cl:
rscadd: added two spice smites; one harmless, one extra spicy
/:cl:


Heroman3003:
Con flict

PR #16223, opened 2024-08-28
Adds Robotic Digitigrade Legs

Essiuun:
There's probably a better way of going about this, but here we are.
Adds digitigrade robolegs, which needs digitigrade actually enabled on a character for it to not look weird. Still manages to look jank with some cybernetic bodies, but it isn't horrible and can be hidden by simply wearing pants, so here it is.


Heroman3003:
Seems to be missing an actual path for them? Or icon?

Essiuun:
ah yeah hold on I coded this as like... midnight so. Lemme fix that.

Kashargul:
https://github.com/CHOMPStation2/CHOMPStation2/pull/5565

How it's done downstream

PR #16226, opened 2024-08-29
Load Prefs From Slot

SatinIsle:
Added a new button to the vore panel preferences tab, "Load Prefs From Slot". This button allows you to simply load another bellies slots onto your current character, but does not automatically save them.

My implementation is just a couple of new procs to load a new slot, save the number of the slot you started on, apply the bellies, then load the original slot again without applying the bellies. Probably needs a keen eye to look at it, but it didn't seem to cause any problems or runtimes in testing.


Heroman3003:
Co nflicty

SatinIsle:
> Co nflicty

Should be resolved now!

PR #16229, opened 2024-08-30
Implements Env Message system

Heroman3003:
Adds a system of Env Messages (Short for environmental message). Any living player can make up to one, or remove at will from wherever they are. Env Messages are floating messages that display a message when hovered over, allowing an RP tool for notifying people of various things easily. They intentionally appear over most things like doors or mobs, making it possible to leave one in, say, a doorway.

Currently if multiple people make a message on same tile, they combine, separated by paragraphs.

When removing a message you can either remove your own message from anywhere, or remove another person's message by standing on top of it.

Making a message when you already have one will remove your old one automatically.

Screenshots:
image
image


code/modules/env_message/env_message.dm
```diff
@@ -0,0 +1,111 @@
...
+
+	if(!istype(src) || !src.ckey)
+		return
+
+	var/ourturf = get_turf(src)
+
+	var/obj/effect/env_message/EM = locate(/obj/effect/env_message) in ourturf
+
+	if(EM)
+		var/answer = tgui_alert(src, "Do you want to remove this env message? (Note: Selecting 'Yes' will remove other players' messages on this tyle too. Please don't remove other players' messages for no reason. Use 'Only My Message' to remove yours only.)", "Env Message", list("Yes", "Only My Message", "No"))
```
SatinIsle, line 103: players' messages on this tyle too.
Should be
players' messages on this tile too.

Heroman3003:
> What happens to someone's old `env_message` when they leave the round?

It sticks around. If they rejoin, it still counts as theirs, as they're remembered by ckey. This is mostly to prevent them disappearing when somebody disconnected momentarily.

I do not see it as too problematic mostly because anyone can delete it if necessary

SatinIsle:
One more suggestion. I would quite like the contents and the person who placed (also maybe the person who deleted) environmental messages to be logged so that they can be reviewed in the logs, for cases where someone writes something offensive and either they delete before we can see, or the person who is offended instinctually deletes it before ahelping.

Heroman3003:
> One more suggestion. I would quite like the contents and the person who placed (also maybe the person who deleted) environmental messages to be logged so that they can be reviewed in the logs, for cases where someone writes something offensive and either they delete before we can see, or the person who is offended instinctually deletes it before ahelping.

Done. Added logging for when one is created and when one you dont own is removed.

PR #16235, opened 2024-08-31
atmos grief...

Kashargul:
We likely have more inputs handled this badly... might be something to go through at some point

🆑 Upstream
fix: closing the fix atmos grief window will no longer trigger it
/🆑


Kashargul:
> I believe the fix-chat and/or fix-tgui prompts also do this, off the top of my head.

a quick mobile search returned multiple pages of hits ... I might have to do a full refactor once I have a few minutes on a pc to write a replace script

PR #16245, opened 2024-09-01
textbox not respecting value changes

### LINKED ISSUE #16239: [GENERAL] TGChat Export Doesn't Preserve Linebreaks

When exporting the log to html via the tgchat export function, linebreaks in posts are not preserved.

*All* formatting to be preserved, including linebreaks.

#### What actually happened

Posts including linebreaks were scrunched into a single line, which messes with (intended) readability.

#### Steps to reproduce
- Make a post that includes linebreaks (shift+enter)
- Export chatlog to html via the tgchat export function
- Witness lack of linebreaks in exported log

Kashargul:
🆑 Upstream
fix: textinput not updating on value changes
qol: archive now preserves newlines
/🆑 

Brings back the editing state and rerenders the textbox once it's no longer edited. This once more allows the value to update on state changes.

fixes #16239

### DIFF
--- a/tgui/packages/tgui-panel/chat/model.js
+++ b/tgui/packages/tgui-panel/chat/model.js
@@ -79,14 +79,21 @@ export const createMessage = (payload) => ({
   ...payload,
 });
 
-export const serializeMessage = (message, archive = false) => ({
-  type: message.type,
-  text: message.text,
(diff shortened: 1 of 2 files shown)


Kashargul:
Would need an editing state again

PR #16257, opened 2024-09-03
September 2024 various fixes

Darlantanis:
Ports more fixes, QoL, and additions from a private downstream:

-Allows vox to use the shaved beard type.
-Allows prometheans to use teshari and vox icon bases
-Adds a new vox marking.
-Tweaks default AI laws to use the map's default instead of hardcoded NT laws. Default is NT laws.
-Prevents vehicle/train from running over the user when pulled up stairs.
-Buffs space cleaner to the same cleaning power as soap, mops, and janiborgs.
-Fixes incorrectly formatted mode_flags on the greatwolf (this should never use a list!).
-Adds vital medical equipment to the medical belt.
-Allows climbing closed creates with mouse drag.
-Allows using airlock doorbells to check for electrification.
-Allows health and body scanners to detect husking.
-Adds spawning with a mind scan. This is a one-time scan, not a backup implant.

DOWNSTREAM NOTE
-DNM in current state; this requires a modular tweak to prevent VR avatars from becoming resleevable.
-Greatwolf fix should be mirrored in the modular file

Downstream changelog
:cl:
fix: Vox can take shaved beards
add: Vox dino muzzle
add: Allows prometheans to use teshari and vox icon bases
fix: AI default laws use the mapset default laws
qol: You can't run yourself over by pulling a vehicle
qol: Buffs space cleaner to match other cleaning methods
fix: Formats greatwolf mode_flags correctly
add: Allows more medical equipment to be stored in the medical toolbelt
qol: Climb crates with mouse drag
qol: Check airlock electrification with the doorbell
qol: Med scanners can detect husking
add: Players can spawn with a one-time mindscan
/:cl:

EDIT: forgot the line about prometheans


code/game/objects/items/weapons/storage/belt.dm
```diff
@@ -244,7 +244,12 @@
 		/obj/item/device/flashlight,
 		/obj/item/weapon/cell/device,
 		/obj/item/weapon/extinguisher/mini,
-		/obj/item/weapon/storage/quickdraw/syringe_case
+		/obj/item/weapon/storage/quickdraw/syringe_case,
```
Novacat, line 247: Odd, this is actually a duplicate and should probably be deleted.

code/game/objects/items/weapons/AI_modules.dm
```diff
@@ -269,6 +269,11 @@ AI MODULES
 	desc = "A 'reset' AI module: 'Clears all, except the inherent, laws.'"
 	origin_tech = list(TECH_DATA = 3, TECH_MATERIAL = 4)
 
+// VOREstation edit: use map default laws
+/obj/item/weapon/aiModule/reset/Initialize()
+	. = ..()
+	laws = new global.using_map.default_law_type // Pull from loaded map
```
Novacat, line 275: What happens if there is no default law defined on the map?

Novacat (commented):
This particular batch of changes seem mostly fine, but massive omnibus PRs should probably be broken up into separate PRs for future reference, in case there is a change that may require discussion/further alteration.

code/game/objects/items/weapons/storage/belt.dm
Darlantanis: Good catch

code/game/objects/items/weapons/AI_modules.dm
Darlantanis: Default is var/default_law_type in /maps/~map_system/maps.dm. This will break if that's changed to null for some reason or a child map datum overrides it incorrectly.

Darlantanis:
Will do if that helps. Still have a lot of commits to sort through and see what's appropriate to send up here. 

> This particular batch of changes seem mostly fine, but massive omnibus PRs should probably be broken up into separate PRs for future reference, in case there is a change that may require discussion/further alteration.

Darlantanis:
oop-

Novacat (approved):
Seems to be mostly fine, though I would like a second set of eyes to double check

PR #16269, opened 2024-09-06
[Ready] Hanner Species

SatinIsle:
Added a new whitelist species, the Hanner. These are the half-offspring of Lleill and share a small number of their abilities. Notably, they are shapeshifters, can drain energy from consenting people, can do basic alchemy and can transform into beasts.

Added the beast form ability to Lleill and Hanner, allowing them to turn into a small selection of simple mobs.

Changed lleill_energy and lleill_energy_max to be a base species variable rather than lleill species, allowing those abilities to be granted to other characters by GMs.

Added a small chance of finding potion materials in maint loot (pretty rare), to allow a little more consistent use of the alchemy ability for lleill, rather than relying on redgates alone.


Novacat (approved):
This will have to be kept an eye on because the simplemob transforms, but otherwise this is fine.

SatinIsle:
> This will have to be kept an eye on because the simplemob transforms, but otherwise this is fine.

Don't worry, this is my baby and I wanna keep a good eye on if it causes problems and fix them as it comes up. I think starting on this basis of trusting people (who we allow onto the whitelist) is a good start, and then changing it from there if problems arise.

PR #16293, opened 2024-09-08
Add examine_block to mob examine

ShadowLarkens:
It looks like this:

:cl:
add: Examining a mob now puts their description in a box. It's pretty.
/:cl:


Kashargul:
This could break sorting, info was used as classifier for the sorting tabs. Best to test

Kashargul (approved):

PR #16305, opened 2024-09-12
Player Effects Panel

SatinIsle:
Added a new Player Effects TGUI panel that replaces the smite verb for staff. This new window contains all smites (plus three new ones), a bunch of medical effects that are a pain to apply via procs, a bunch of useful abilities that are a pain to grant through the give verb proc, inventory controls that are normally a pain to apply manually, a centralised location for some admin procs, and a few a small number of bug-fix related ones.

Here is a full list of buttons by category:

### Smites
- Break Legs
- Bluespace Artillery
- Spontaneous Combustion
- Lightning Bolt
- Shadekin Attack
- Shadekin Vore
- Redspace Abduction
- Autosave
- Autosave AOE
- Ad Spam
- Peppernade
- Spawn Spice
- Terrify
- Terrify AOE (also terrifies all mobs in the affected players view range)
- Spin (applies SpinAnimation)
- Squish (applies the squish hammer effect)

### Medical
- Appendicitis
- Repair Organ (via amount of damage or rejuvenate)
- Damage Organ/Limb (via damage amount, bruising or killing the organ)
- Break Bone
- Drop Organ/Limb
- Make Organ Assisted
- Make Organ Robotic
- Rejuvenate
- Toggle Stasis (Useful for heavily damaging a mob and then saving them in a safe condition until players approach)

### Abilities
- Vent Crawl
- Set Darksight (to value)
- Give Cocoon Transformation
- Give TF verbs
- Give Set Size
- Set Lleill Energy Levels (both current and max amounts)
- Give Lleill Invisibility
- Give Leill Beast Form
- Give Leill Transmutation
- Give Leill Alchemy
- Give Lleill Drain
- Give Brutal Predation
- Give Trash Eater

### Inventory
- Drop Everything
- Drop Specific Item
- Drop Held Items
- List All Items
- Add Marked Item To Hands
- Equip Marked Item To Inventory

### Admin
- Quick NIF
- Resize
- Teleport
- Gib
- Dust
- Paralyse
- Subtle Message
- Direct Narrate
- Open Player Panel
- Open View Variables
- Enable/Modify AI
- Make Marked Datum Orbit

### Fixes
- Rejuvenate
- Send Message Box (designed for those times where a player needs help but can not view chat)
- Clear All Orbiters


Kashargul:
Due to size, it would be preferred to split the UI into multiple tab files

SatinIsle:
> Will need tgui bundle recompie

Rebundled it

SatinIsle:
> Due to size, it would be preferred to split the UI into multiple tab files

You might be right, but I've not quite been able to wrap my head around how that works, even from following the example of the colormate. The current file is still very simple and less than 300 lines, so I'm personally okay with keeping it as it is until I can work it out, then I might split it later as I learn more about TGUI (unless someone wants to do it first).

SatinIsle:
> > > Due to size, it would be preferred to split the UI into multiple tab files
> > 
> > 
> > You might be right, but I've not quite been able to wrap my head around how that works, even from following the example of the colormate. The current file is still very simple and less than 300 lines, so I'm personally okay with keeping it as it is until I can work it out, then I might split it later as I learn more about TGUI (unless someone wants to do it first).
> 
> > > Due to size, it would be preferred to split the UI into multiple tab files
> > 
> > 
> > You might be right, but I've not quite been able to wrap my head around how that works, even from following the example of the colormate. The current file is still very simple and less than 300 lines, so I'm personally okay with keeping it as it is until I can work it out, then I might split it later as I learn more about TGUI (unless someone wants to do it first).
> 
> I'll likely split it then, the way is quite easy
> 
> The UI name becomes a folder name.
> 
> So the folder is called `PlayerEffects`
> 
> the current UI file will be renamed to `index.tsx`
> 
> the type Data definition goes to `types.ts` and will be imported from there import {Data} from './types'
> 
> Each tab goes to a new file, called for example `PlayerEffectsSmites.tsx' for the smites
> 
> This sub file will habe a single export element, for example called `PlayerEffectsSmites` that then gets reimported into the main UI
> 
> import {PlayerEffectsSmites} from './PlayerEffectsSmites'
> 
> And will be used like normal elements
> 
> Some general things, don't use data and the Data prop if you element has no data.
> 
> If a sub element has props, make sure to type the props of that new element

Thanks again for the advice! If you are happy to split it then, I will look at how you've done it afterwards, I think that'll be a good way for me to learn exactly how it's done when it's something I was familiar with.

SatinIsle:
Kashargul has split up the TGUI files for me, and I've now rebundled, tested and merged them!

PR #16319, opened 2024-09-16
Hostile mob friends!

SatinIsle:
Added 1 new hostile mob, the homunculus. This mob will transform into the first person that it sees, and will gain potential attack and armour depending on what that mob is wielding without copying it exactly.

Added 3 new variants of the corrupt hound, fire breather, fast and laser. They have slightly modified sprites to make them easy to identify. I have been using these mobs a lot in events lately and it'd be nice to give them some variety.


code/modules/mob/living/simple_mob/subtypes/glamour/homunculus.dm
```diff
@@ -33,3 +33,92 @@
...
 
+/mob/living/simple_mob/homunculus/evil
+	var/seen_melee
+	var/seen_gun
+	var/transformed
+	var/seen_armour
+	maxHealth = 100
+	health = 100
+
+	faction = "glamour"
```
Kashargul, line 44: Do not bring back local factions. Define them for an overview

SatinIsle: Done! (Probably, if you wanna double check, lol)

PR #16323, opened 2024-09-17
A bunch of new clothes

SatinIsle:
Added 6 underclothing variants of feminine T-shirts, including midriffs, text and shoulder colours.

Added 2 new very short skirts, one with a split exposing one leg.

Added 4 recolourable belt accessories. Thin, thick, strap (no buckle) and studded.


Heroman3003:
Con f lict

SatinIsle:
> Con f lict

Should all be resolved now.

PR #16348, opened 2024-09-22
Replicant Species

SatinIsle:
This PR makes the replicant outsider species a whitelist species after an application was made to play one as a crew member.

Added the Gamma Replicant station species, accessible via whitelist only. This species has additional weaknesses (radiation and shock). They have a few unique organs: The crew version of the rage heart can only activate once every 60 minutes (rather than every minute) but lasts for 40 seconds. The mending lungs repair a few organs very very slowly, instead of extremely quickly. The plasma sac generates plasma a little faster without added phoron.

Changed ventcrawl to specifically allow replicants to take underclothes and pocket sized items with them. I've balanced this for replicants only, but could remove the species check if wanted.

Changed alien and resin structures so that they can also be destroyed by replicants (though more slowly than aliens).

Added the user to the whitelist as approved here: https://forum.vore-station.net/viewtopic.php?f=45&t=2448&p=14179#p14179


Novacat:
I am not really too thrilled with these because they essentially give mechanical advantages for almost zero cost at all.

SatinIsle:
> I am not really too thrilled with these because they essentially give mechanical advantages for almost zero cost at all.

Specifics like this are still being discussed in staff channels, it isn't going to be merged just yet.

PR #16357, opened 2024-09-23
allows to order more deuterium if needed

Kashargul:
I've no clue why we can only order tritium and not the much safer variant to build RUSTs

🆑 Upstream
add: deuterium crates to cargo
/🆑


Kashargul:
active edge.

PR #16359, opened 2024-09-23
Check for 516 byond:// hrefs

ShadowLarkens:
Requested by @Kashargul 

Draft PR to track these until we fix them all.


Kashargul:
Those are a lot... that will be worse than the define rework.... time to also add one that forbids span usage

PR #16360, opened 2024-09-24
Make map selection marginally better and more CI friendly

ShadowLarkens:
Instead of hard defining the map in the dme and then having to mess with the dme to do map swaps, there's now a file with three defines to select between each of our maps: `maps/~map_system/_map_selection.dm`

@Very-Soft If this PR goes through, map update procedure will change from ticking things to editing the file `maps/~map_system/_map_selection.dm` and commenting/uncommenting defines: For example, switching from Stellar Delight to Tether would be

From
```
// #define USE_MAP_TETHER
#define USE_MAP_STELLARDELIGHT
// #define USE_MAP_GROUNDBASE
```
To
```
#define USE_MAP_TETHER
// #define USE_MAP_STELLARDELIGHT
// #define USE_MAP_GROUNDBASE
```


SatinIsle:
We should wait until minitest is included in this before merging

ShadowLarkens:
> We should wait until minitest is included in this before merging

fixed

PR #16363, opened 2024-09-25
Print your own healthbar as prey

SatinIsle:
Added the functionality to print your own healthbar via "Print Prey Healthbars" when you yourself are the prey.


code/modules/vore/chat_healthbars.dm
```diff
@@ -117,6 +117,16 @@
 	set category = "Abilities"
 
 	var/nuffin = TRUE
+
+	var/obj/belly/amprey = src.loc
+
```
Kashargul, line 122: what happens with a switch having prey in a pred on this?

PR #16373, opened 2024-09-27
Mob tf smite

SatinIsle:
Added a new smite called Mob Transformation that allows staff to turn anyone into another mob using the same methods as mouse rays/polymorph/beastform, except with no limits.

Added a player effects panel "Fix" option to revert mob TF that will work on the aforementioned smite and any other methods that use the same type of mob TF.


Kashargul:
No more bundles on repo.

SatinIsle:
> No more bundles on repo.

Do I just delete them from this then?

Kashargul:
> > No more bundles on repo.
> 
> Do I just delete them from this then?

Yes, compilation is handled server side now

SatinIsle:
> Yes, compilation is handled server side now

Alrighty, think I did it properly!

PR #16377, opened 2024-09-27
Remove tgui bundle from repo

ShadowLarkens:
I believe we're now switched over fully to CBT on the server, so this is safe to remove.


Heroman3003:
Conflict

PR #16403, opened 2024-10-01
Give the AI back most of their shell options (not research)

ShadowLarkens:
Changes AI shell selection from `Standard`, `Service`, `Clerical` to 
`Standard`, `Engineering`, `Surgeon`, `Crisis`, `Miner`, `Janitor`, `Service`, `Clerical`, and `Research`.

### DIFF
--- a/code/global.dm
+++ b/code/global.dm
@@ -139,7 +139,8 @@ var/list/emergency_module_types = list(
 )
 // List of modules available to AI shells
 var/list/shell_module_types = list(
-	"Standard", "Service", "Clerical"
+	"Standard", "Engineering", "Surgeon", "Crisis",
+	"Miner", "Janitor", "Service", "Clerical", "Research"
 )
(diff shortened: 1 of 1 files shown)


Heroman3003:
Research is probably much more problematic than Security could ever be, but this is up for staff discussion

Very-Soft:
I deffo agree that the research shell is significantly more problematic than any of the other shells, as it allows the AI to produce more shells without help, which in the past, lead to AI players who would start with a research shell, so that they could make more shells, so they could do everything all on their own!

Heroman3003:
The consensus among staff is that Security module is fine, but Research is not. There were concerns raised over Engineering and Miner too, but those were not really concrete. Will update when specific decision is reached.

Heroman3003:
Alright, overall consensus is:

Security module is fine to have unforbidden
Research module should be forbidden instead

Do that and it's good to go.

PR #16417, opened 2024-10-03
Vore Messages

ShadowLarkens:
:cl:
refactor: Belly message formatting has been unified into one proc.
add: All formatted belly messages can now use a number of new %replacements, which choose from a list of preset words. You can see all of the words that will be used in a new help menu next to the belly description, but this PR adds the following: %goo, %happybelly, %fat, %dick, %boob, %pussy, %grip, %cozy, %angry, %acid, %snack, %hot
add: For example, %goo will be replaced randomly with "muck", or "goo", or "sludge", or "slime", or another 10 different options - this random replacement happens every time a message is printed, so it's really helpful for idle messages!
/:cl:

### DIFF
--- a/code/modules/vore/eating/belly_messages.dm
+++ b/code/modules/vore/eating/belly_messages.dm
@@ -1,0 +1,507 @@
+/obj/belly
+	// Don't forget to watch your commas at the end of each line if you change these.
+	var/list/struggle_messages_outside = list(
+		"%pred's %belly wobbles with a squirming meal.",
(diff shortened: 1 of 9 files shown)


Heroman3003:
I'm not sure if our policies changed, but direct references to dicks, boobs and vaginas might be a big nono.

ShadowLarkens:
> I'm not sure if our policies changed, but direct references to dicks, boobs and vaginas might be a big nono.

Can easily get rid of them if it's a problem

PR #16420, opened 2024-10-04
entertainment cameras now update on move and some cleanup

Kashargul:
cleans up some spans as well

🆑 Upstream
fix: Entertainment cameras now update on move
/🆑


Kashargul:
Do not merge currently, also using this to add some things I find during deconfliction of the span pr

PR #16428, opened 2024-10-04
Enable research module / prevent AI shells from building more AI shells

lm40:
I believe the staff conclusion was that we trust the playerbase to handle this short of the self-replication which was considered to be a rules trap, so this disables self-replication.


ShadowLarkens:
I wouldn't have thought of this, great way to do it. 👏

PR #16433, opened 2024-10-05
Add two new alt titles: Software Engineer and Geneticist

ShadowLarkens:
:cl:
add: Station Engineer has a new alt title: Software Engineer - "A Software Engineer fulfills similar duties to other engineers, but specializes in working with software and firmware. They also often deal with integrated circuits."
add: Medical Doctor has a new alt title: Geneticist - "A Geneticist is a specialist in reforming technology and genetic editing with tools such as the body design console."
/:cl:

### DIFF
--- a/code/__defines/jobs.dm
+++ b/code/__defines/jobs.dm
@@ -200,6 +200,7 @@
 	#define JOB_ALT_ELECTRICIAN "Electrician"
 	#define JOB_ALT_CONSTRUCTION_ENGINEER "Construction Engineer"
 	#define JOB_ALT_ENGINEERING_CONTRACTOR "Engineering Contractor"
+	#define JOB_ALT_SOFTWARE_ENGINEER "Software Engineer"
 
 #define JOB_ATMOSPHERIC_TECHNICIAN "Atmospheric Technician"
 	// Atmospheric Technician alt titles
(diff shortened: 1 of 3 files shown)


Heroman3003:
Tbh both titles make more sense for Scientists than they do for Doctors or Engineers, considering how doctors have nothing to do with genetics, and engineers have extremely limited and partial access to circuits compared to scientists.

ShadowLarkens:
my thought is that it fits reforming and body editing which are medical duties

> Tbh both titles make more sense for Scientists than they do for Doctors or Engineers, considering how doctors have nothing to do with genetics, and engineers have extremely limited and partial access to circuits compared to scientists.

my thoughts are that it fits medical because of resleeving/reforming and body design and modification

SatinIsle:
Geneticist IRL can either be a scientific geneticist (studying genes) or medical geneticist (treating genetic disorders), so both work. Seeing as anything mechanical that really hints at geneticist work here is in medical, it makes more sense for it to be a medical title.

I do feel software engineer makes more sense for science though, mostly for the same reasons.

Heroman3003:
Main reason I say that for Geneticist is because genetics is TECHNICALLY still in the game, buildable and doable, and is as a whole considered science's purview. That specific title has pretty specific association in context of the game.

PR #16434, opened 2024-10-05
next set of spans

Kashargul:
That should be all spans themselves that can be changed. Leaves us with `<b>, <i> and <u>` tags to go at a later time.

Please... give your messages sorting tags... see spans... default is `span_infoplain` which just sorts messages under info.


Kashargul:
active edge.

PR #16442, opened 2024-10-07
TGUI Shock Collar

ShadowLarkens:
:cl:
add: Shock collar now has a pretty TGUI interface.
/:cl:


SatinIsle:
Have you tested that this doesn't cause problems with the bluespace collars (which are a subtype of shock collars)?

ShadowLarkens:
> Have you tested that this doesn't cause problems with the bluespace collars (which are a subtype of shock collars)?
> 
> (Keep in mind that there are 3 variants of bluespace collars that have different input options if it does)

They'll just use the old UI

PR #16449, opened 2024-10-08
Convert preferences to /tg/ preferences

ShadowLarkens:
Draft pull to work on with Karshagul. One by one, we'll convert each preference to a tg pref, and move loading/saving out of the legacy system, while maintaining the legacy system for UI control alone.


Kashargul:
> Draft pull to work on with Karshagul. One by one, we'll convert each preference to a tg pref, and move loading/saving out of the legacy system, while maintaining the legacy system for UI control alone.

Open issues:
- [ ] TG prefs are not saved right after character creation, save

Kashargul:
Splitting up the prefs between living, silicon and human. This should allow us to call the procs dependings. Apply to mob ob all mobs where needed. Apply to mob and silicon on all silicons and apply to mob and human on humans.

PR #16456, opened 2024-10-10
Glamour mob pack

SatinIsle:
Added four new glamour themed mobs, primarily for event areas: The ddraig (a large dragon with a lot of powers), the blaidd (a wolf creature that behaves similarly to but more aggressively than the cave stalker), the fluffball (a shy little critter that runs away unless you give them food) and the unicorn (shoot deadly rainbows).


Screemonster:
I love the fluffballs but they should probably be vore_ignores_undigestable = FALSE if they're primarily absorby otherwise they'll just ignore people who are otherwise entirely suited towards being claimed as fluff

SatinIsle:
> I love the fluffballs but they should probably be vore_ignores_undigestable = FALSE if they're primarily absorby otherwise they'll just ignore people who are otherwise entirely suited towards being claimed as fluff

Good shout! I fixed it.

PR #16462, opened 2024-10-11
adds a verb to print playtime

Kashargul:
🆑 Upstream
add: a client verb to print out department hour playtime
/🆑

### DIFF
--- a/code/modules/client/client procs.dm
+++ b/code/modules/client/client procs.dm
@@ -640,7 +640,7 @@
 	set name = "Active Playtime"
 	set category = "IC"
 
-	if(!play_hours.len)
+	if(!play_hours?.len)
 		to_chat(src, span_warning("Persistent playtime disabled!"))
 		return
 
--- a/code/modules/client/client procs.dm
+++ b/code/modules/client/client procs.dm
@@ -640,7 +640,7 @@
 	set name = "Active Playtime"
 	set category = "IC"
 
-	if(!play_hours?.len)
+	if(!play_hours.len)
 		to_chat(src, span_warning("Persistent playtime disabled!"))
 		return
 
--- a/code/modules/client/client procs.dm
+++ b/code/modules/client/client procs.dm
@@ -641,11 +641,14 @@
 	set category = "IC"
 
 	if(!play_hours.len)
-		to_chat(src, span_warning("No recorded playtime found!"))
+		to_chat(src, span_warning("Persistent playtime disabled!"))
 		return
 
-	var/department_hours = "Your department hours:"
+	var/department_hours = ""
 	for(var/play_hour in play_hours)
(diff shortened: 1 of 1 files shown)


Kashargul:
> I'm not convinced the values this is returning are accurate. I've played plenty of rounds in cargo, but I'm pretty sure I haven't logged just over 88 full rounds. And whilst I've done a fair bit of engine setup, I almost _never_ log full engineering rounds-- definitely not enough to have gotten almost 73 hours of engi playtime according to this verb.
> 
> Are you sure this isn't counting post-PTO-adjustment hours? IIRC every hour played counts for 3 hours of off-duty time, and if I divide all the numbers by 3 they seem rather more believable.

Those are the values directly from the Database on your client. There are multiple reasons they are off.


Firstly, they only update when you actively play and don't idle / afk

Secondly, the subsystem is on a 15 minutes timer and running as background. This is a horrible combination with how bad our code is optimized, it might only fire once every hour, so you only get 15 minutes for 1 hour playing. The subsystem needs to have the background flag removed.

Kashargul:
> > Those are the values directly from the Database on your client. There are multiple reasons they are off.
> > Firstly, they only update when you actively play and don't idle / afk. So they are lower than the real playtime even.
> > Secondly, the subsystem is on a 15 minutes timer and running as background. This is a horrible combination with how bad our code is optimized, it might only fire once every hour, so you only get 15 minutes for 1 hour playing. The subsystem needs to have the background flag removed. Virgo likely has less issues with ticks than we have downstream, I don't know your live tick budged.
> 
> Well that would explain a lot, but it still doesn't explain why they seem _too high_.
> 
> It also means this verb is functionally kind of useless or at the very least highly misleading if it can't perform its intended function and give an accurate return of playtimes.

It gives the active hours in relation to what the game checks for the playhours on job requirements. That is what it's for.

Kashargul:
> Well I dunno what to tell you but it's telling me I have 72.75 hours in Engineering which works out to 12 full rounds give or take, and I'm pretty dang sure I haven't logged that much time in that department.

I have no access to your database. Check it, that entry is in there. How it got there, only serverlogs would show. But in general the subsystem writes 15 minutes every 15 minutes of active gameplay.

Kashargul:
> > I have no access to your database. Check it, that entry is in there. How it got there, only serverlogs would show. But in general the subsystem writes 15 minutes every 15 minutes of active gameplay.
> 
> I don't have access to that database either because I'm checking the live server and I'm not staff. :|

I can only say it works fine on my private instances as well as on downstream since we moved the subsystem from background to run normally. But I still dislike that we have no tick budged left.

A big issue we had downstream was that the database was formatted wrongly, so we never had play time recorded until we fixed that recently. All observations were that the records are if then slower than real time and never faster.

PR #16463, opened 2024-10-11
Port tg statpanel

ShadowLarkens:
Mostly an up-port from CHOMP, CHOMPStation2/CHOMPStation2#8007

Known issues (won't fix in this PR for diff size sake):
- Verbs aren't sorted yet
- Ping isn't implemented and shows -- Not Available --
- Round ID still isn't implemented (we really gotta do this sooner rather than later, I'm tempted to just shove it in a json file at this point)

Screenshots:

:cl:
add: The entire upper right quadrant of the game window has been replaced with a new stat/verb panel that doesn't lag the client to death. 
/:cl:


ItsSelis (approved):
Tested it, also looks good to me. Haven't found any issues myself either and if it's straight from downstream then it was already tested a bunch by similar code.

PR #16475, opened 2024-10-14
Add support for deploying space helmets over top of other helmets

ShadowLarkens:
Currently, spacesuit helmets can go over any other helmet or head clothing item except other spacesuit helmets. May need to tweak and do stuff like whatever helmets security wears? 

:cl:
add: You can now deploy hardsuit helms over top of your normal headwear, the same as magboots.
/:cl:

### DIFF
--- a/code/modules/clothing/clothing.dm
+++ b/code/modules/clothing/clothing.dm
@@ -460,6 +460,7 @@
 
 	var/light_overlay = "helmet_light"
 	var/image/helmet_light
+	var/overhead = 0			//Used by spacesuit helmets
 
 	sprite_sheets = list(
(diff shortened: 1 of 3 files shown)


Novacat:
It will be a bit silly for space suit helmets to be able to go over armored helmets, and iirc the armor values may stack as well?

ShadowLarkens:
> It will be a bit silly for space suit helmets to be able to go over armored helmets, and iirc the armor values may stack as well?

armor values definitely won't stack, this whole thing is a hack that basically just automatically unequips it and stuffs it inside the other helmets contents

But I need a full list of things that shouldn't be able to be put over each other

ShadowLarkens:
Okay I blacklisted /obj/item/clothing/head/helmet which seems to cover all armored helmets

PR #16484, opened 2024-10-16
Ports over configuration controller

ItsSelis:
# Attention downstreams
You **will** have to edit your configuration files to make stuff still work.
Changes that are required were made in the examples, so its best to check the diffs.

The configuration examples have been updated to how the configuration keys (for example PROBABILITY changed to PROBABILITIES) should look like.

The game will write debug logs in case it does not recognize a specific configuration string when booting up.


code/game/objects/items/weapons/wiki_manuals.dm
```diff
@@ -0,0 +1,145 @@
...
+/obj/item/book/manual/wiki
+	dat = "Nanotrasen presently does not have any resources on this topic. If you would like to know more, contact your local Central Command representative." // safety
+	/// The ending URL of the page that we link to.
+	var/page_link = ""
+
+/obj/item/book/manual/wiki/display_content(mob/living/user)
+	var/wiki_url = CONFIG_GET(string/wikiurl)
+	if(!wiki_url)
+		//user.balloon_alert(user, "this book is empty!")
+		to_chat(user, "this book is empty!")
```
Kashargul, line 39: should span_info

PR #16486, opened 2024-10-16
Experimental rust-g update

ItsSelis:
The latest code from tg (some of the changes made in our custom repo were already added in the official tg one already).

So far no runtimes while running the code, logs also get saved at the same location with the same name and pretty much the same format too.

Note: This will cause the time in the logs to be using UTC instead of the local server time (which is generally recommended to be used for logs for daylight savings time alone).


Kashargul:
If we swap to TG rust, also make sure to edit the dockerfile

PR #16492, opened 2024-10-18
Contraband Scanner

KillianKirilenko:
Adds a machinery object that scans the contents of mobs that walk over it, and if it detects they're carrying an item matching its internal blocklist, sounds an alarm. The alarm has a short, area-shared cooldown (10 seconds) so you can't spam the sound/text message by walking over the same (or multiple) scanners in rapid succession.

Currently TODO for the following;
- [x] need to figure out looser type checking on the list (currently requires exact matches)
- [x] needs to have a setting to deep-scan (iterate through bags/etc.)
- [x] replace the placeholder alarm sound with a proper one
- [x] adjust area lockdown procedure
- [x] customizable alert text
- [x] test more thoroughly in general

:cl:
rscadd: added contraband scanner machine, which can sound an alarm if you walk over it whilst carrying an illegal item
/:cl:


code/game/machinery/contraband_scanner.dm
```diff
@@ -0,0 +1,67 @@
...
+				contraband_count++
+			if(deep_scan)
+				for(var/obj/O2 in O.contents)	//one layer deep is fine for now I think
+					if(O2.type in contraband)
+						contraband_count++
+		if(contraband_count && last_trigger < world.time - cooldown)
+			L.visible_message(
+				span_danger("\The [src] buzzes loudly, \"CONTRABAND DETECTED.\""),
+				span_danger("\The [src] buzzes loudly, \"CONTRABAND DETECTED.\""),
+				"<b>You hear a loud metallic buzz as a mechanical voice announces \"CONTRABAND DETECTED.\"!</b>"
```
Kashargul, line 43: should use span_hear(span_bold(""))

Kashargul: More than possible, I didn't finish with the full rework yet, but I would prefer if all new commits already follow the style and fix things where they are sighted on the way

SpadesNeil:
If I want to add this to my map, I read something about a lockdown procedure. How does that work? Door bolting? Shutters? Etc?

PR #16522, opened 2024-10-25
Experimental tweak to make chat backlog rendering smoother

ShadowLarkens:
Basic theory is to chunk up updates into small batches over time instead of a huge batch all at once.

Current settings: 250 messages per 100ms. Subject to further experimentation to find ideal value.

:cl:
fix: Chat loads smoother now and shouldn't freeze the client while it does so.
/:cl:


Kashargul:
Should be a customizable number, my 50k archive from ramdisk just turns to horror with that. Or % based on hhe archive size like 20-25%.

ShadowLarkens:
Yeah I think this will just have to wait for 516

PR #16533, opened 2024-10-28
sanity checks some things in case we really have no bellies

Kashargul:
sanity checks the bellies, just to prevent crashes in case we really have none and use the vorepanel


Kashargul:
hmm nope, this can't be null.

PR #16537, opened 2024-10-28
Airlock Construction

ShadowLarkens:
Makes it possible to construct all the parts of a cycling airlock.

 

Instructions:
- Frame Parts -> Click on wall -> "Button" frame
- Use "Cycling Airlock Button" board (there are 20 in the engivend and research can print them), hit with multitool to change to different parts

- Construct like electrochromatic button, wrench - board - screwdriver - wires - screwdriver
- Hit sensors, buttons, and vent pumps with multi-tools to change frequency and ID (Master Tag means the airlock controller's ID, otherwise a unique ID of your choice)
- Screw open door, look at wires to set frequency and ID
- Screw open controller to see a configuration interface.
- Once all the IDs are set (minimum is exterior and interior door, air pump, and chamber sensor), it works!
- Beware: Because high volume vent pumps aren't obtainable (YET), your airlock is destined to be slowww cycling out.
- Examine every part when you're within one tile to see their tags.

:cl:
add: You can now de and reconstruct all parts of an airlock.
add: Ability to set tags on different airlock parts (button, sensor, vent pumps) with a multitool.
/:cl:


Heroman3003:
Conflict

PR #16539, opened 2024-10-29
Ring/Watch Accessory Transfer

KillianKirilenko:
Refactors Rings, Watches, Bracelets, and Knuckledusters into Accessories that can be attached to most Gloves. They can still be worn standalone if desired. Rings and knuckledusters will give your gloves a small bump in force/punchforce whilst attached as well.

Note this will make it slightly trickier to access watch features, but I'll see if I can figure out a clean solution to this at some point.

The traitor and toy buzzer rings have been removed for now, since I'm not sure how best to preserve their functionality whilst attached.

All accessory files have also been moved up from being with uniforms as plenty of them are no longer uniform-based.

Long-term wishlist is converting the wrist PDA into an attachable/accessory whilst still allowing ID checks and all other PDA functions to work normally and to be easily accessible, but I do not think I will be tackling that in this PR.

Known issue in that icons are a bit weird for attached watches on gloves due to the way it handles the states. Not sure how to fix them just yet.

:cl:
tweak: refactored rings, knuckledusters, bracelets, and watches into glove-attached accessories (they can still be worn by themselves!)
tweak: bracelets are now worn on your hands/wrists rather than being in the neck slot
rscdel: removed buzzer rings
/:cl:


code/modules/clothing/accessories/rings.dm
```diff
@@ -0,0 +1,201 @@
...
+
+//Sleepy Ring
+/obj/item/clothing/accessory/ring/reagent/sleepy
+	name = "silver ring"
+	desc = "A ring made from what appears to be silver."
+	icon_state = "material"
+	origin_tech = list(TECH_MATERIAL = 2, TECH_ILLEGAL = 5)
+
+/obj/item/clothing/accessory/ring/reagent/sleepy/New()
+	..()
```
Kashargul, line 72: This likely can use Initialize without issues

Initialize()
. = ..()

then the add reagents. We want to get away from NEW usages

code/modules/clothing/accessories/rings.dm
```diff
@@ -0,0 +1,201 @@
...
+
+/////////////////////////////////////////
+//Reagent Rings
+
+/obj/item/clothing/accessory/ring/reagent
+	flags = OPENCONTAINER
+	origin_tech = list(TECH_MATERIAL = 2, TECH_ILLEGAL = 4)
+
+/obj/item/clothing/accessory/ring/reagent/New()
+	..()
```
Kashargul, line 49: Like with the other

PR #16542, opened 2024-10-30
Metal Vore Solid

SatinIsle:
Added a new AI function called guard_limit. This limits the "find target" portion of the AI to only acquire targets that are in front of them, allowing people to potentially sneak behind them.

Added a few guard_limit variants of existing AI holder types, particularly for mercs.

Tested and it seems to work exactly as intended. If their friends are engaged in combat, they will still rush in to help them even if they were looking the other way.


code/modules/ai/ai_holder_targeting.dm
```diff
@@ -49,7 +50,10 @@
 	. = list()
 	if(!has_targets_list)
 		possible_targets = list_targets()
-	for(var/possible_target in possible_targets)
+	for(var/atom/possible_target in possible_targets)
```
ShadowLarkens, line 53:
```suggestion
	for(var/atom/possible_target as anything in possible_targets)
```

PR #16577, opened 2024-11-07
Fixed virology themed redgate maps

SatinIsle:
Added 4 new props based on the old virology machines for decoration.

Fixed the missing virology equipment in the city and facility redgate maps.


Kashargul:
Hmmm are redgate maps not unit tested?

SatinIsle:
> Hmmm are redgate maps not unit tested?

Unfortunately they only seem to get tested if one of the main maps is tested and that redgate happens to load for it.

PR #16588, opened 2024-11-10
Completely merk all traces of the skill system

ShadowLarkens:
Completely removes the skills tab from character setup and rips out the entire system and everything that interacted with it.

Why?
1. This is one less thing that needs to be ported to /tg/ prefs (#16449) if we just do away with it now.
2. Literally the only place this was used was for overmap ship controls, and even then, 90% of that was stubbed out to just use SKILL_EXPERT anyways! I've just finished making the last 10% use SKILL_EXPERT, and then went through and simplified all of the now-constant math to just apply bonuses permanently, without calculating 3 - 1 first. 
3. If we ever want a comprehensive skill system, we should just take /tg/'s, not continue using this *ancient* mess. 

:cl:
del: The skill system is OUTTA HERE
/:cl:


ShadowLarkens:
> Whilst I'm thinking about it: since this is gonna be nuking the entire skill tab/button, I'd like to propose that we move the allergy/food preference traits into its place, so they're no longer taking up tons of space in the neutral traits list.

I'd rather avoid this for now because we're working towards completely rewriting the menu and this would cause worse conflicts instead of making it easier

PR #16593, opened 2024-11-11
Removes ability for research cyborgs to self-upgrade

DigitalSquirrel95:
Port of https://github.com/TS-Rogue-Star/Rogue-Star/pull/335

### DIFF
--- a/code/modules/mob/living/silicon/robot/robot.dm
+++ b/code/modules/mob/living/silicon/robot/robot.dm
@@ -747,7 +747,7 @@
 		else if(!src.module && U.require_module)
 			to_chat(usr, span_filter_notice("The borg must choose a module before it can be upgraded!"))
 		else if(user == src) //RS edit: No self-upgrading.
-			to_chat(user, span_filter_notice("You lack the reach to be able to upgrade yourself."))
+			to_chat(usr, span_warning("You lack the reach to be able to upgrade yourself."))
 		else if(U.locked)
 			to_chat(usr, span_filter_notice("The upgrade is locked and cannot be used yet!"))
 		else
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/silicon/robot/robot.dm
```diff
@@ -737,13 +737,17 @@
 				update_icon()
 			else
 				to_chat(user, span_filter_notice("[span_red("Access denied.")]"))
+				if(user == src) //RS edit: No self-unlocking.
+					to_chat(user, "<span class='filter_notice'>You are not allowed to manipulate your own lock.</span>")
```
Kashargul, line 741: use of old spans is forbiden

code/modules/mob/living/silicon/robot/robot.dm
```diff
@@ -737,13 +737,17 @@
...
 
 	else if(istype(W, /obj/item/borg/upgrade/))
 		var/obj/item/borg/upgrade/U = W
 		if(!opened)
 			to_chat(usr, span_filter_notice("You must access the borgs internals!"))
 		else if(!src.module && U.require_module)
 			to_chat(usr, span_filter_notice("The borg must choose a module before it can be upgraded!"))
+		else if(user == src) //RS edit: No self-upgrading.
+			to_chat(user, "<span class='warning'>You lack the reach to be able to upgrade yourself.</span>")
```
Kashargul, line 750: use of old spans is forbidden

Kashargul (changes requested):

liache:
This seems like an unnecessary nerf. Do we really want this?

Mattjpark89:
On other servers where people are trying to fulfill antag objectives and whatnot, this change makes sense as it increases the difficulty of hacking or modifying a neutral borg (as they can no longer unlock themselves simply by being told to).
On this server though? I don't see a reason for it. Unless it's being abused in some way I can't think of.

SpaceMan-Jim:
I certainly don't believe this is a very valid issue to justify removing abilities. If there IS a case off a sci unit swapping their module around after upgrading. I'd personally say its a moderation issue than anything else, feels more spite filled considering said individual also wanted to remove the self repair.

PR #16595, opened 2024-11-11
usr to user up to player effects

Kashargul:
🆑 
code: converts some usr to user
/🆑 

up to code\modules\admin\player_effects.dm


SatinIsle:
Appearance changer mostly works fine, except that all the ear options are missing
image

Kashargul:
> Appearance changer mostly works fine, except that all the ear options are missing image

Ye, it's broken since the second ear adding.
Another bug I see in the code is the registering of the move component... I highly doubt that the mirror works properly for multiple people... But I need to check the entire vore usage before axing more from it

Heroman3003:
conflict

PR #16597, opened 2024-11-12
up ports misc gripper

Kashargul:
Up ports the misc gripper from chomp, best to look over the list of scene tools, it's currently 1:1 ours from downstream

### DIFF
--- a/code/modules/mob/living/silicon/robot/drone/drone_items.dm
+++ b/code/modules/mob/living/silicon/robot/drone/drone_items.dm
@@ -177,6 +177,23 @@
 		/obj/item/material/gravemarker
 		)
 
+/obj/item/gripper/scene
+	name = "misc gripper"
+	desc = "A simple grasping tool that can hold a variety of 'general' objects..."
+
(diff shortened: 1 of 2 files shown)


Kashargul:
add pertri

Kashargul:
> So, testing:
> 
> * Petrification mostly seems to work okay, EXCEPT that the petrified player can not escape it. The petrification ability as a gargoyle does nothing and OOC escape just returns "You are not inside anything". Though it seems to work fine if you have the gargoyle trait.
> * The gargoyle trait is missing the fields to change the material and adjective, though the colour still works.
> * Gripper seems to work as expected.
> * I don't understand what body layer shifting is supposed to be so I didn't test that.

gonna check those later, the layer change should have no player facing changes, so if nothing looked weird that's good

SatinIsle:
The OOC escape works fine now. The "Can Unpetrify" option on the petrification machine gives the "gargoyle - petrification" verb to the statue, but it doesn't appear to be able to be used.

PR #16599, opened 2024-11-12
Fluff item

SatinIsle:
Approved here: https://forum.vore-station.net/viewtopic.php?f=26&t=2462

A second one approved here: https://forum.vore-station.net/viewtopic.php?f=26&t=2461


code/modules/games/tarot.dm
```diff
@@ -41,3 +41,59 @@
...
+	for(var/suit in list("wands","pentacles","cups","swords"))
+		for(var/number in list("ace","two","three","four","five","six","seven","eight","nine","ten","page","knight","queen","king"))
+			P = new()
+			P.name = "[number] of [suit]"
+			P.card_icon = "dark_[suit]"
+			P.back_icon = "dark_back_tarot"
+			cards += P
+
+/obj/item/deck/dark_tarot/shuffle()
+	var/mob/living/user = usr
```
Kashargul, line 84: horrible xS, the shuffle proc should get mob/user 

and in card, the verb should call shuffle with usr

SatinIsle: I agree! Though that's a bit out of the scope of this one maybe, because this is just a variant of the normal tarot deck above it (and playing cards in general). After this one is merged, I can go through and fix up all those rather than doing it as part of a fluff item.

SatinIsle: Or maybe I'll do it all in one once it's been approved actually.

SatinIsle: Can you check I've done it as you meant?

Kashargul (approved):

PR #16607, opened 2024-11-17
Potential partial revert to borg self-upgrading restrictions

SatinIsle:
Pending staff discussion.

This reverts the restriction of the borgs from upgrading themselves, for the most part. However, it keeps in place a restriction from them resetting their own module.

### DIFF
--- a/code/modules/mob/living/silicon/robot/robot.dm
+++ b/code/modules/mob/living/silicon/robot/robot.dm
@@ -707,8 +707,6 @@
 				update_icon()
 			else
 				to_chat(user, span_filter_notice("[span_red("Access denied.")]"))
-				if(user == src) //RS edit: No self-unlocking.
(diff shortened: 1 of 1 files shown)


Heroman3003:
Approved

Wulfestwulf:
This was FULLY reverted on Rogue star... I feel like we should also fully revert it. Its a moderation issue, not a mechanical issue.

SatinIsle:
> This was FULLY reverted on Rogue star... I feel like we should also fully revert it. Its a moderation issue, not a mechanical issue.

Please use the discord channels (perhaps general-feedback or one of the dev channels) to discuss this. We're open to the discussion, and if you can present genuine use cases for keeping borgs able to change their module freely, then we may revert it/remove it to an admin notification. Otherwise, this PR probably isn't going to be monitored for discussion.

PR #16626, opened 2024-11-20
Flippable beds and sheets

SatinIsle:
Added a verb to beds, bedsheets and sub-objects to allow them to be flipped to face the opposite direction (if they have directional sprites available). This applies to chairs too, because why not. If it's possible for me to just not display the verb for non-flippable subtypes, let me know how! Characters will still be buckled down to the right regardless, but that's not the largest use-case for beds over here and I think people would prefer the option.

Changed some beds on the Stellar Delight to face the other direction so that their pillows are against the wall.

image


code/game/objects/structures/stool_bed_chair_nest/bed.dm
```diff
@@ -168,6 +169,30 @@
 	if(padding_material)
 		padding_material.place_sheet(get_turf(src), 1)
 
+/obj/structure/bed/verb/turn_around()
+	set name = "Turn Around"
+	set category = "Object"
+	set src in oview(1)
+
+	if(!flippable)
+		to_chat(usr,"\The [src] can't face the other direction.")
```
Kashargul, line 178: no  sorting span used

Kashargul (changes requested):

code/game/objects/structures/stool_bed_chair_nest/bed.dm
SatinIsle: Fixed!

Kashargul (approved):

PR #16639, opened 2024-11-24
"Maint Lurker" Hole

KillianKirilenko:
Very simple PR in function. It adds a new "strange maintenance hole" that causes the activating ghost to spawn as their currently loaded character slot using a special "maintenance lurker" role (with no datacore entries, accounts, or announcement). Ideal for more monster-adjacent 'carbonmob' characters who might be a little odd as straight-up visitors or crewmembers. These "maintenance lurkers" can spawn with any loadout item that is not role-restricted, but otherwise receive no special benefits or perks beyond the usual scene protections.

Core functionality works as-designed so far, save that checking for flavour text on the spawning mob isn't handled right now.

Todo:
- [x] ensure players can't use it to spawn as whitelisted species they aren't permitted to play
- [x] add a hidden 'job'/role and outfit instead of just using Visitor
- [x] find a better name for the role

:cl:
rscadd: adds "maint lurker" hole, a spin on the classic maint pred that lets you spawn as your current character slot
/:cl:


code/game/objects/structures/ghost_pods/event_vr.dm
```diff
@@ -172,3 +172,82 @@
...
+
+/obj/structure/ghost_pod/ghost_activated/maint_lurker/create_occupant(var/mob/M)
+	..()
+
+	var/picked_ckey = M.ckey
+	var/picked_slot = M.client.prefs.default_slot
+
+	var/mob/living/carbon/human/new_character = new(src.loc)
+	if(!new_character)
+		to_chat(src, "Something went wrong and spawning failed. Please check your character slot doesn't have any obvious errors, then either try again or send an adminhelp!")
```
Kashargul, line 219: Should get a sorting span, example span_warning

PR #16686, opened 2024-12-09
fix air alarm temperature controls

Kashargul:
Having the TLV check match >= and <= prevented values from e.g. 0 and 40 to be used on the manual input. Furthermore, checking the difference of the target temperature led to starts heating, stops cooling if the target temperature was changed.

🆑 
fix: limit temperatures not reachable and old state being displayed wrong
/🆑


Heroman3003:
I think some RIG bugfixes snuck on there, is that intended?

Kashargul:
> I think some RIG bugfixes snuck on there, is that intended?

as it's an MC crash chance, I copied it over yes

PR #16696, opened 2024-12-10
Adds a new ghost vis blocking area flag

Kashargul:
Also using this to add the missed the defines while deconflicting downstream

🆑 
add: a new flag AREA_BLOCK_GHOST_SIGHT to prevent ghosts from seeing certain areas
/🆑 

grafik

todo:
- [x] add verb to add areas for admins
- [x] add verb to remove areas for admins
~~implement a shadekin option to allow event managers to exclude them from events~~ maybe later...
- [x] null the nets in Destroy...


Kashargul:
> Does this also stop ghosts from entering those 'blocked' spaces, be it by movement or using orbit/follow prompts?

No, for that there is another flag, this only blocks their sight

PR #16702, opened 2024-12-12
fix multizmovement

Kashargul:
Need some testing, but this should resolve our issues and easily allow us to add new turfs in the future which might allow free passing into either direction.

🆑 
fix: multiZmovement in space in cases where a non space turf exists below
/:cl:


TankTheBirb:
Ran around and checked turfs on the SD, change looks all good to me.

PR #16705, opened 2024-12-12
Borg Icon Optimization and Weapon Modulation

Cameron-The-Raven:
Cross-port of Rogue Star's [PR 708 here](https://github.com/TS-Rogue-Star/Rogue-Star/pull/708)

## Credit: [Spotteh](https://github.com/amenitykitten) & [Cameron The Raven](https://github.com/Cameron653)

- Makes borgs NOT update_icon() every tick.
- Makes borgs use a flag system for overlays INSTEAD of having a thousand variables spread throughout the various borg types.
- Consolidates all the various handle_extra_icon_updates into a singular handle_extra_icon_updates that can be modified for TRULY unique borgs (Ex: foopwotch has two unique sprites: dagger & disabler. It calls the parent and adds its custom, unique overlays.)
- Adds has_active_type_list (Allows for a list to be used to search for a list of modules. Want a red eyes overlay if you have an 'evil' module selected and a green eyes overlay if you have a 'good' module selected? You can do it with this!) This should be used sparingly however, as searching lists CAN be resource intensive if the list is too long.
- Converts the borg gun system to a more modular one so instead of random guns being slapped with the 'can be used flag' they're all children of '/obj/item/gun/energy/robotic'.
- Converts the borg melee system to a more modular one

### The process it'd do it is listed below

=======

1. cut_overlays() (Removes all overlays)
2. add_overlay (for tummy)
2.5 If they are resting, we then do cut_overlays AGAIN (meaning the add_overlay above is USELESS and generated just to be removed)
3. add_overlay (tummy resting sprite)
4. add_overlay (eyes)
5. add_overlay (lights on)
5.5 (If we have DIED we do cut_overlays and then an add overlay, but this will pretty much never happen / only happen once so whatever)
6. add_overlay (if our panel is open)

=======

So every tick we were, FOR MOST BORGS, doing a minimum of 1 cut overlay and 1-4 new add_overlay() per tick...Which while not being MASSIVELY resource intensive, is a waste of resources.

### This streamlines the process and makes it so update_icon only updates when it's REQUIRED. (Equipping items into your hotbar that would cause an overlay to appear, eating people, dying, etc)

Additionally, adds in a new var for borgs that is "Has_shield_and_speed" for if you have a sprite for when you have both speed AND shield equipped. Previously, it'd only update your sprite if you had a specific thing selected (like swapping to the shield, even though the shield was passive, or taser if you had it selected, but did not show if it was just on your hotbar) but now it shows as long as it's on your hotbar.
It has a priority system (Due to speed sprites changing the icon entirely for most if not all borgs):
1. Shield+Speed (for sprites that are compatible with this)
2. Speed
3. Shield
4. Gun (Compatible with shield)

### Things tested with each other (Unless otherwise noted, they have same functionality as before this PR):
- Vore Sprites
- Shield Sprites
- Gun Sprites
- Shield + Gun Sprites
- Shield + Gun + Panel sprites
- Shield + Gun + Panel + Vore + Flashlight Sprites (Everything barring the speed sprites)
- Speed Sprites
- Speed + Panel Sprites (This looks weird but it's same functionality as before this PR)
- Vore + Panel Sprites (Looks weird on some borgs, but again, same functionality as prior to this PR)
- Death (Removes all modules from your hotbar and gives you the dead icon, as it worked prior.)

This /SHOULD/ cover all bases, but if there's another edge case that pops up, tossing it up in Issues on the github and referencing this PR would give a good idea of where to start looking for what we missed. We looked through every borg (including the fluff borgs) for issues, but things can always slip through.

Edits:

- Made the overlays use flags to decide if they can get the overlay or not
- Made it so the code isn't in 5 bajillion places.
~~- Made it so the guns are contained in a nice little list that borgs currently use so none of them lose their overlays~~
- Makes guns and melee weapons modular


code/modules/mob/living/silicon/robot/sprites/combat.dm
```diff
@@ -6,10 +6,16 @@
 
 	var/has_speed_sprite = FALSE
 	var/has_shield_sprite = FALSE
+	var/has_shield_and_speed = FALSE //RS Edit
```
Kashargul, line 9: I don't think we should add even more booleans.... rather convert those has_xxx _sprite into a flag

Kashargul:
> * Made the overlays use flags to decide if they can get the overlay or not
> * Made it so the code isn't in 5 bajillion places.
> * Made it so the guns are contained in a nice little list that borgs currently use so none of them lose their overlays
> 
> Added a big note that "HEY, SOMEONE SHOULD MAKE THE BORG GUN SYSTEM MODULAR". It's WAY outside the scope of this PR but needs to be done eventually.

Making it modular is actually really easy

#define ROBOT_SPRITE_LASER
#define ROBOT_SPRITE_TASER
#define ROBOT_SPRITE_BLADE

/obj/item
   robot_sprite_flag = 0

/obj/item/.../laserx
    robot_sprite_flag = ROBOT_SPRITE_LASER

/obj/item/proc/get_robot_overlay(flag)
return robot_sprite_overlay & flag

code/modules/mob/living/silicon/robot/sprites/_sprite_datum.dm
```diff
@@ -1,7 +1,28 @@
+#define ROBOT_HAS_SPEED_SPRITE 0x1	//Ex:	/obj/item/borg/combat/mobility Replaces old has_speed_sprite
```
Kashargul, line 1: defines need to go to the _defines folder or need to get undefined

move those best to the defines section

code/modules/projectiles/guns/energy/cyborg.dm
```diff
@@ -0,0 +1,126 @@
...
+	name = "Cybernetic Taser"
+	desc = "An integrated taser that is used to neutralize foes."
+	borg_flags = COUNTS_AS_ROBOT_GUN | COUNTS_AS_ROBOT_TASER
+	icon_state = "taser"
+	item_state = null	//so the human update icon uses the icon_state instead.
+	projectile_type = /obj/item/projectile/beam/stun
+	charge_cost = 400
+	recharge_time = 7 //Time it takes for shots to recharge (in ticks)
+
+/obj/item/gun/energy/robotic/disabler
```
Kashargul, line 27: should be removed from the dog_modules file as it's now here and the original no longer in use

code/modules/projectiles/guns/energy/cyborg.dm
```diff
@@ -0,0 +1,126 @@
...
+	self_recharge = 0 //Ditto
+	use_external_power = 0 //Ditto
+	accuracy = 75 // Suppressive weapons don't work too well if there's no risk of being hit.
+	burst_delay = 1 // Burst faster than average.
+
+	firemodes = list(
+		list(mode_name="single shot", burst = 1, burst_accuracy = list(75), dispersion = list(0), charge_cost = 24),
+		list(mode_name="five shot burst", burst = 5, burst_accuracy = list(75,75,75,75,75), dispersion = list(1,1,1,1,1)),
+		list(mode_name="ten shot burst", burst = 10, burst_accuracy = list(75,75,75,75,75,75,75,75,75,75), dispersion = list(2,2,2,2,2,2,2,2,2,2)),
+		)
```
Kashargul, line 126: With all weapons being here, we also should move the borgblade. The ERT borgs use it even though virgo has no sprite state for it yet.

code/modules/mob/living/silicon/robot/sprites/_sprite_datum.dm
```diff
@@ -27,7 +28,41 @@
...
+			if(!ourborg.has_active_type(gun)) //The gun we're checking is NOT selected. Pass over it.
+				continue
+			if(sprite_flag_check(ROBOT_HAS_GUN_SPRITE) && gun.gun_flag_check(COUNTS_AS_ROBOT_GUN))
+				ourborg.add_overlay("[sprite_icon_state]-gun")
+			if(sprite_flag_check(ROBOT_HAS_LASER_SPRITE) && gun.gun_flag_check(COUNTS_AS_ROBOT_LASER))
+				ourborg.add_overlay("[sprite_icon_state]-laser")
+			if(sprite_flag_check(ROBOT_HAS_TASER_SPRITE) && gun.gun_flag_check(COUNTS_AS_ROBOT_TASER))
+				ourborg.add_overlay("[sprite_icon_state]-taser")
+			if(sprite_flag_check(ROBOT_HAS_DISABLER_SPRITE) && gun.gun_flag_check(COUNTS_AS_ROBOT_DISABLER))
+				ourborg.add_overlay("[sprite_icon_state]-disabler")
```
Kashargul, line 65: if(sprite_flag_check(ROBOT_HAS_BLADE_SPRITE) && gun.gun_flag_check(COUNTS_AS_ROBOT_BLADE))
				ourborg.add_overlay("[sprite_icon_state]-blade")

used for the /obj/item/melee/combat_borgblade

code/modules/projectiles/guns/energy/cyborg.dm
Kashargul: ok ye that might be an issue... I have no clue if anyone uses those elsewhere

PR #16740, opened 2024-12-20
More area flag option fixes

Kashargul:
🆑 
admin: fixed an issue de-phasing admins in no phase zones
admin: de and re admin should now properly update the visibility of the ghost blocking sight
admin: fixed an issue with aghosting not working in ghost blocked zones
add: a switch to allow the ghost vision toggling if placed in an area
/🆑


Kashargul:
Todo: adjust the update handling some

PR #16741, opened 2024-12-20
Experimental: Compile TGUI for Edge and IE separately

ShadowLarkens:
This functions but I'm not sure if it's a great idea or not.

:cl:
tweak: 516 clients are served an Edge-optimized bundle.
/:cl:


AyIong:
image

PR #16749, opened 2024-12-22
prevents falling in divable water

### LINKED ISSUE #16748: [MAPPING] Gravity on the Islands is overactive.
In the Islands Redgate (The one with the beach, the ocean, and the underwater z-level), one has a bad habit of instantly plummeting to the bottom of the ocean upon entering deep water. Violently.

When I enter deep water, I expect to keep wading slowly, or be able to dive if I have the aquatic trait. At any point I can move down a z-level if I'm over an open space in the ocean.

#### What actually happened
I plummet to the underwater z-level, and take damage as if falling. There's a thud and I enter stun from falling. Pain follows. Incidentally this also affects objects similarly, so boats pushed into the water violently sink.

Ow. ): 
Satin believes this is a funny consequence of recent fixes to z-level transitions.

Kashargul:
🆑 
fix: you can no longer fall through water
/🆑 

fixes #16748

### DIFF
--- a/code/__defines/is_helpers.dm
+++ b/code/__defines/is_helpers.dm
@@ -68,3 +68,6 @@
 
 #define istaurtail(A)	istype(A, /datum/sprite_accessory/tail/taur)
 #define islongtail(A)	istype(A, /datum/sprite_accessory/tail/longtail)
+
+// Diveable water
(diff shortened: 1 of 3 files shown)


SatinIsle:
Tested on request and works fine!

PR #16753, opened 2024-12-22
Allows borg tasers to properly charge when in charging ports

Cameron-The-Raven:
See title.

### Fixes:
Allows borg tasers to properly charge when in charging ports

### DIFF
--- a/code/modules/mob/living/silicon/robot/robot_modules/station.dm
+++ b/code/modules/mob/living/silicon/robot/robot_modules/station.dm
@@ -497,7 +497,7 @@ var/global/list/robot_modules = list(
 		F.icon_state = "flash"
 	else if(F.times_used)
 		F.times_used--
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/silicon/robot/robot_modules/station.dm
```diff
@@ -497,7 +497,7 @@ var/global/list/robot_modules = list(
 		F.icon_state = "flash"
 	else if(F.times_used)
 		F.times_used--
-	var/obj/item/gun/energy/taser/mounted/cyborg/T = locate() in src.modules
+	var/obj/item/gun/energy/robotic/taser/T = locate() in src.modules
```
Kashargul, line 500: As tasers can be removed, I heavily recommend a sanity check below as well in case no taser is found.

Kashargul: thanks^^

PR #16771, opened 2024-12-28
Vision bugfixes

Cameron-The-Raven:
Fixes a few vision related things and the 'feysight' modifier to allow for more redspace spookiness things for if admins want to do things like make a ghost visible and whatnot.

- Fixes see_invisible from being overwritten if your see_invisible_default is > SEE_INVISIBLE_LEVEL_ONE.

- Fixes a runtime of when manifest() was called on a ghost with no user. Allows admins to call manifest() without it runtiming.

- Due to ghost 'toggle visibility's last_toggled' it being a variable on the proc itself, it wouldn't properly call and store the time.

- Makes it so 'feysight' modifier actually applies the proper plane_holder and removes it as well. (You could 'see' ghosts with feysight, but only on rightclick due to the layer not being applied properly. You could point, try to (and fail to) pull them, and throw things at them, but you couldn't physically 'see' them which was unintended.)

- Allows ghosts to ACTUALLY USE the ghost whisper and become visible proc if their is_manifest is enabled via admins.

- Allows admins to give any /mob verbs and procs instead of just  /mob/living via admin add_verb topic

### DIFF
--- a/code/modules/mob/dead/observer/observer.dm
+++ b/code/modules/mob/dead/observer/observer.dm
@@ -883,7 +883,7 @@ This is the proc mobs get to turn into a ghost. Forked from ghostize due to comp
 	set category = "Ghost.Settings"
 	toggle_ghost_visibility()
 
-/mob/observer/dead/proc/toggle_ghost_visibility(var/forced = 0)
+/mob/observer/dead/proc/toggle_ghost_visibility(var/forced = FALSE)
 	if(!is_manifest)
 		to_chat(src, span_filter_notice("You are not strong enough to pierce the veil..."))
 		return


code/modules/mob/dead/observer/observer.dm
```diff
@@ -844,6 +845,10 @@ This is the proc mobs get to turn into a ghost. Forked from ghostize due to comp
 	add_verb(src, /mob/observer/dead/proc/toggle_visibility)
 	add_verb(src, /mob/observer/dead/proc/ghost_whisper)
 	to_chat(src, span_filter_notice(span_purple("As you are now in the realm of the living, you can whisper to the living with the " + span_bold("Spectral Whisper") + " verb, inside the IC tab.")))
+	if(!user)
+		visible_message(span_purple("The ghost of \the [src] is dragged back in to our plane of reality!"))
```
Kashargul, line 849: span_deadsay

code/modules/mob/dead/observer/observer.dm
```diff
@@ -876,11 +876,17 @@ This is the proc mobs get to turn into a ghost. Forked from ghostize due to comp
...
-/mob/observer/dead/proc/toggle_visibility(var/forced = 0)
-	set category = "Ghost.Settings"
+/mob/observer/dead/verb/toggle_visibility()
+
 	set name = "Toggle Visibility"
 	set desc = "Allows you to turn (in)visible (almost) at will."
+	set category = "Ghost.Settings"
+	toggle_ghost_visibility()
 
+/mob/observer/dead/proc/toggle_ghost_visibility(var/forced = 0)
```
Kashargul, line 886: FALSE please

PR #16806, opened 2024-12-31
update DNA up port

Kashargul:
Quick git diff with strip edit comments from mobile.


Kashargul:
That's likely all chomp additions, will need to check when not on mobile

PR #16818, opened 2025-01-02
fix use of nonexisteing icon state

Kashargul:
🆑 
fix: invisible ore chunks
/🆑 

the strange state only exists in the xenoarch file, not in the ore file. So those turn invisible if they are not overridden by a single specific resource in the dynamic one. Using the ore2 state for now to at least show them.


Kashargul:
domirror 16821

PR #16863, opened 2025-01-07
Fighter Update 2

Aroliacue:
Next round of changes to exosuit fighters. Fighters have now been given their own armor and hull type, allowing them a significantly higher top speed right out the gates, but at the cost of durability. New custom soundfiles have also been added to replace old ones, and the Pinnace has been reflavoured into an entry-level fighter craft.

:cl:
add: Exosuit Fighters now have exclusive hull and armor types, with a faster top speed at the cost of reduced durability.
add: Exosuit Fighters now have collision mechanics. Fly responsibly.
sfx: Exosuit Fighters have new enter sound effects and a special collision effect.
bal: Pinnace has been rebalanced into a weak, entry-level fighter. Has reduced health compared to other fighters, but can mount weapons and fly in atmosphere.
:/cl:


Aroliacue:
The voices are from Mechwarrior 4, the effects are a mix of ones I've made and existing ingame sounds :)

PR #16868, opened 2025-01-08
Conversion many Globals to Managed Globals

ItsSelis:
## About The Pull Request

It generally is better/recommended to use the managed global defines for the variables to be properly tracked and even be edited later in-game if needed.

This PR already converts a bunch of these global vars to those managed ones.

#### Todo:

- [ ] Issue with computer files in ntos stuff not being able to read the GLOB.ntnet. It will most likely require New() => Initialize()
- [ ] Undefining converted defines


ItsSelis:
Already covered by future PRs

PR #16899, opened 2025-01-13
don't qdel turfs

Kashargul:
## About The Pull Request
This needs some testing as virgo has no TM, but in general, we don't delete turfs, we change them. In case of cleaning a level, we revert them back to space turfs.

## Changelog

:cl:
fix: runtime flood
/:cl:

### DIFF
--- a/code/modules/rogueminer_vr/zonemaster.dm
+++ b/code/modules/rogueminer_vr/zonemaster.dm
@@ -396,6 +396,10 @@
 		if(I.type == /turf/space)
 			I.cut_overlays()
 			continue
+		if(isturf(I))
+			var/turf/T = I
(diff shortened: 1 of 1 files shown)


SatinIsle:
> I'll see about doing this on a test server on the SD (since SD has the rogue miner) sometime tomorrow or the day after.
> 
> If you've already tested this locally, let me know and I can merge it.
> 
> Looking at the code, I don't see any reason this would cause issues and it looks good to merge for me.

I have tested this on Kash's request just now (Hadn't seen your comment yet). Scanned for asteroids, went to them, blew loads of stuff up, went home and re-scanned, went out to a new set of asteroids. Everything worked as expected and it produced no runtimes.

PR #16922, opened 2025-01-18
more 516 compat

Kashargul:
## Changelog

:cl:
fix: more 516 browser windows 
/:cl:


SatinIsle:
Seems to work fine on my test server for both exporting and importing vrdb files on 515. Also tested an arcade machine and that worked fine too.

PR #16924, opened 2025-01-18
Simultaneous map definitions

Drulikar:
## About The Pull Request

This PR is a follow up to #16861 making it so tether, ground base and stellar delight can have their map paths all simultaneously defined so all can be rendered in actions like nanomaps, MDB, and can be worked on in strongdmm without use of the map switch. This should bring the option of a map vote closer. However, the map switch is still needed to handle overrides and actually determines what dmm files get included.

This ought to be TM'd before just outright merged. But I cannot find anything obviously broken by the change.

## Changelog
:cl: Drathek
refactor: Reworked map includes for ground base, stellar delight, and tether to have all paths included simultaneously
/:cl:


Cameron-The-Raven:
Just wanted to say this hasn't been forgotten, just a LOT going on at the moment!

Cameron-The-Raven:
Going ahead and testing. If all goes well, will merge.

Cameron-The-Raven:
Did some more testing on an unrelated PR and stumbled across this 
dreamseeker_2025-03-07_22-16-51
dreamseeker_2025-03-07_22-14-45
dreamseeker_2025-03-07_21-52-21


I then checked to make sure it wasn't the PR and, nope, it is the master branch this is happening on.

All I can pinpoint it to is possibly being this PR, but I'm not certain what could have broken here...Any ideas? It seems like the overmap Z level isn't properly loading? But I have no clue.

PR #16940, opened 2025-01-21
up to modify robot

Kashargul:
Migrates our UIs to tgui core. This will need proper testing...

TODO:
- [ ] Flex is deprecated, remove them

currently: up to modifyRobot

## Changelog

:cl:
code: moves our UIs to tgui core
/:cl:

### DIFF
--- a/tgui/packages/tgui/interfaces/AccountsTerminal.tsx
+++ b/tgui/packages/tgui/interfaces/AccountsTerminal.tsx
@@ -1,6 +1,6 @@
 import { BooleanLike } from 'common/react';
-
-import { useBackend, useSharedState } from '../backend';
+import { useBackend, useSharedState } from 'tgui/backend';
+import { Window } from 'tgui/layouts';
 import {
   Box,
   Button,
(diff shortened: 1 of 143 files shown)


ShadowLarkens:
@Cameron-The-Raven 

AdminTicketPanel - works fine
BodyZoneSelector is unused, not sure what/if it was ever used by.
ComplexModal - Easiest way to check is ChemMaster - works fine
FullscreenNotice - APC grid check - works fine
EmbeddedController - TODO
GeneralAtmoControl  - Not used, so whatever
Stories - You hit F12 while in UI debug mode - They're busted, but not from tgui-core, the whole thing is just unmaintained
Layouts - If any UI works, layout is working

Body Designer - existing bug
Cloning Console - we only use resleeving anyways and the records are broken, so whatever
Matrix recolor - I think this is an existing bug? this PR doesn't touch character setup since it's not TGUI yet
Laptop Vendor/Skills - existing bug

Appearance changer shown sprite is TINY - known 516 issue
Body Designer too - same, known 516 issue, out of our control
When opening the body designer tgui window it does this: - known 516 issue, same as tooltips flickering
ComputerVendor is large - intentional, actually, I designed the UI to be really beeg. 
Exosuit Fabricator UI - asset issue, confirmed

ShadowLarkens:
Fixed copy to clipboard in PDA messenger

Research Lock screen: Someone please double check on 515 that it doesn't runtime

PR #16978, opened 2025-01-27
Old event material cleanup

KillianKirilenko:
## About The Pull Request

Cleans up a slew of old event material, including some 'questionable' material and sprites that were not fully cleaned up and longer obtainable (or not obtainable in the first place) and thus had no reason to remain as assets in our codebase.

## Changelog

:cl:
del: Removed old clothing and weapon assets
/:cl:


Kashargul:
A really bad idea to just delete icons / objects just because they are currently not in use this will mess with any downstream using them. At least keep the dmis in else they will just get back ported again at some point.

PR #17021, opened 2025-02-01
ports the skyrat snow leopard tail

Kraseo:
## About The Pull Request

what it says on the tin, ports the skyrat snow leopard tail because it looks better imo. 

snep

## Changelog
:cl:
image: The snow leopard/"snep" tail has a new sprite, courtesy of Skyrat.
/:cl:


Cameron-The-Raven:
Let me see about adding both versions, so people that want (and like) the old version can still use it, while this newer version also exists

Cameron-The-Raven:
Done in #17032 instead to allow both the old and new sneptails. Will merge it shortly.

PR #17056, opened 2025-02-04
Gets rid of sleep in Fire and Fire_userless

Cameron-The-Raven:
## About The Pull Request

Gets rid of the sleep that is in the Fire() proc and the Fire_userless() proc.

## Changelog

:cl:
code: Adjusts the way guns fire on the backend
/:cl:


Cameron-The-Raven:
DNM until Kash can examine this

PR #17067, opened 2025-02-05
Adds back diamond pickaxe & Kinetic Gauntlets to vendor

Cameron-The-Raven:
## About The Pull Request

This was done 6 years ago and miners have way better tools that the diamond pickaxe now. Nobody is going to science asking for a diamond pick, so there's no harm in letting them have it through the vendor as well. If they want a diamond drill, then go ask science.
Additionally, the kinetic gauntlets have been added back to the vendor. They're a slightly higher damage but more bulky version of the kinetic machete. You give up 2 hands for 3 more damage increase, so I'm making the same price as the kinetic machete. If you want to look cool smashing rocks with your fist, go ahead.

## Changelog

:cl:
add: Mining vendor now contains the Kinetic Gauntlets
add: Mining vendor now contains the Diamond Pickaxe
/:cl:


Cameron-The-Raven:
> Idle thoughts for possible (future) revisions: make the drills upgradable using mats and some kind of kit or pack from the vending machine.

I definitely agree. Mining is in need of some gameplay adjustments to make it so it isn't just 'I get X tool I am done with my progression tree.'

In the future I'd like to approach mining and get feedback to see  'what does mining feel like it's missing', but outside the scope of this PR alone.

PR #17086, opened 2025-02-07
Toggleable Recoil

Cameron-The-Raven:
## About The Pull Request
Makes a toggleable recoil for micros.
Disabled on Virgo, but used downstream.
## Changelog
:cl:
add: Up-ports recoil. Disabled here, but used downstream.
/:cl:

### DIFF
--- a/code/modules/projectiles/gun.dm
+++ b/code/modules/projectiles/gun.dm
@@ -94,6 +94,7 @@
 	var/obj/item/dnalockingchip/attached_lock
 
 	var/last_shot = 0			//records the last shot fired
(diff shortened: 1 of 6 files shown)


TeleTubby2:
Not that we're using it here, but a short-distance-throw based on the size diff would be comedic

Cameron-The-Raven:
> Not that we're using it here, but a short-distance-throw based on the size diff would be comedic

Honestly yeah, it'd be funny.

Something that needs to be edited:
1. weaken() is unreliable to make someone 'drop' an item and needs to be changed to force drop
2. It needs to check if you're buckled to a chair and unbuckle you if so.

And I could add another var to it that does knockback if wanted for bigger guns as well.

But those are for another PR, as this one is setup to be 1:1 with how downstream had it.

PR #17177, opened 2025-02-17
Larger TGUI say messages

SatinIsle:
## About The Pull Request

Changes TGUI say to allow double the previous max characters, to match emotes that don't use TGUI say.

## Changelog
:cl:
code: Changes TGUI say to allow double the previous max characters, to match emotes that don't use TGUI say.
/:cl:


SatinIsle:
Ah, unfortunately MAX_HUGE_MESSAGE_LEN is defined in a weird spot
Will have to fix the PR later.

PR #17194, opened 2025-02-21
Fixes some runtimes [IDB IGNORE]

Cameron-The-Raven:
## About The Pull Request
Fixes a collection of runtimes.
Fixes a memory leak.
Fixes a bug where pipes would lose all their gas under certain circumstances
## Changelog
:cl: Diana
add: You can now unwrench pipes that are over-pressurized, at the cost of being flung back.
fix: Atmos analyzers will no longer fail on portable connectors
fix: No more attempting to blendRGB if we have nothing to blend.
fix: Diseases will not try to copy non-existent vars over.
fix: Recursive explosions no longer runtime if the turf it tries to step into is invalid.
qol: Health analyzer guide now tells users what to do, no longer requiring medical access. Also some wording changes.
fix: Health analyzer guide no longer runtimes if someone has no viruses.
fix: Cards no longer runtime if you try to pull a card without being a human.
fix: Anything-turned-protean will alert you if it has no humanform
fix: Tunneler spiders will no longer runtime if the turf they try to tunnel into doesn't exist
fix: Mannequins now properly get deleted when you observe.
fix: When placing a pipe on a pipeline, the pipeline will no longer lose all the gas within the pipeline. This occurred when you placed a new pipe on a pipeline that had not had not previously had a pipe unwrenched on the pipeline, resulting in the pipes on the pipeline not saving the gas, resulting in all the gas within the pipes on that pipeline being voided. 
/:cl:

### DIFF
--- a/code/_helpers/atmospherics.dm
+++ b/code/_helpers/atmospherics.dm
@@ -42,10 +42,16 @@
 	return atmosanalyzer_scan(src, src.air_contents, user)
 
 /obj/machinery/atmospherics/pipe/atmosanalyze(var/mob/user)
-	return atmosanalyzer_scan(src, src.parent.air, user)
+	if(parent && parent.air) //Sometimes we may have a pipe that has no parent. This seems to happen if you add a pipe onto a pipeline, causing it to delete the parent for every pipe on that pipeline...Yeah. It's complicated and a bug.
+		return atmosanalyzer_scan(src, src.parent.air, user)
 
+// This one is strange. The connector is not guaranteed to have a network (if you placed it down by itself)
+// 'gases' is also a list. But the atmos analyzer wants you to give it a gas mixture.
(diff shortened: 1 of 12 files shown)


code/_helpers/atmospherics.dm
```diff
@@ -45,7 +45,7 @@
 	return atmosanalyzer_scan(src, src.parent.air, user)
 
 /obj/machinery/atmospherics/portables_connector/atmosanalyze(var/mob/user)
-	return atmosanalyzer_scan(src, src.network.gases, user)
+	return atmosanalyzer_scan(src, src.network.gases[1], user) //This is weird, but src.network.gases was giving atmosanalyze a list previously.
```
Kashargul, line 48: Intended? It's more than 1 gas? Fix the analyzer

PR #17206, opened 2025-02-23
Chat history prototype

ItsSelis:
## About The Pull Request

The goal of this project is to clear up most, if not all, of the lag that appears after a while of the message logs inside of the chat getting too large.

## Changelog
:cl:
qol: using a database to store the round logs to reduce client-side lag
qol: using same database to load the most recent messages into the chat (instead of having to fetch from client browser)
/:cl:


Kashargul:
Todo:
- [ ] Fix export decline crash
- [ ] Implement the initial line loading depending on pref or 1000
- [ ] allow export for more than thencurrent round
- [ ] have a list with all stored rounds?

tgui/packages/tgui-panel/chat/middleware.ts
```diff
@@ -136,8 +141,76 @@ const loadChatFromStorage = async (store: Store<number, Action<string>>) => {
...
+    Byond.sendMessage('databaseExportLines', {
+      length: settings.visibleMessageLimit,
+      json: true,
+    });
+  });
+};
+
 export const chatMiddleware = (store) => {
   let initialized = false;
+  let dbLoading = false;
```
Kashargul, line 213: ok we don't need that, initialized ensures it can only be called once

tgui/packages/tgui-panel/chat/middleware.ts
```diff
@@ -170,14 +243,24 @@ export const chatMiddleware = (store) => {
       settings.hideImportantInAdminTab,
       settings.interleave,
       settings.interleaveColor,
+      game.databaseBackendEnabled,
     );
     // Load the chat once settings are loaded
-    if (!initialized && (settings.initialized || settings.firstLoad)) {
+    if (
+      !initialized &&
+      game.dataReceived &&
```
Kashargul, line 251: This casues chat loading to fail... we need to find a better way

tgui/packages/tgui-panel/chat/constants.ts
```diff
@@ -4,6 +4,10 @@
  * @license MIT
  */
 
+import { createAction } from '../../common/redux';
+
+export const loadFromDb = createAction('chat/loadFromDb');
```
Kashargul, line 9: should be under actions

tgui/packages/tgui-panel/chat/middleware.ts
```diff
@@ -37,6 +37,7 @@ import { createMessage, serializeMessage } from './model';
 import { chatRenderer } from './renderer';
 import { selectChat, selectCurrentChatPage } from './selectors';
 import { message } from './types';
+import { loadFromDb } from './constants';
```
Kashargul, line 40: import from actions

tgui/packages/tgui-panel/chat/middleware.ts
```diff
@@ -250,14 +311,16 @@ export const chatMiddleware = (store) => {
         // This needs to be handled on a timeout of the websocket...
         loadChatFromStorage(store);
```
Kashargul, line 312: this needs to be moved away from here and into the invoke

tgui/packages/tgui-panel/chat/middleware.ts
```diff
@@ -380,6 +443,11 @@ export const chatMiddleware = (store) => {
       const event = new Event('chatexportplaced');
       document.dispatchEvent(event);
     }
+    if (type === loadFromDb.type) {
+      const user_payload = payload;
+      loadChatFromDBStorage(store, user_payload);
```
Kashargul, line 448: if !payload.token 

then we do the old chat loading

tgui/packages/tgui-panel/telemetry.ts
```diff
@@ -6,22 +6,30 @@
 
 import { storage } from 'common/storage';
 import { createLogger } from 'tgui/logging';
+import { loadFromDb } from './chat/constants';
```
Kashargul, line 9: chat/actions.

Kashargul (changes requested):

Kashargul (approved):

PR #17218, opened 2025-02-26
fake coins and coin/mint refactors

KillianKirilenko:
## About The Pull Request

Adds "fake" coins to loadouts, and seperates coin icons into their own dmi. These fake coins can be flipped and are almost indistinguishable from the real thing, except that they can't be slotted into vending machines to get access to the premium items.

Also refactored the mint to not use an `if/else if` chain and instead check if the sheet has a coin_type set - if it does, print those coins. Otherwise it prints an error message (to your chat log, not as a physical item, as funny as that might be).

Also also added 7 new coin types for flavour.

## Changelog
:cl:
add: added fake coins to loadouts, can be flipped, can't be put in vendors
add: added seven new coin types (plus fakes): steel, titanium, copper, durasteel, plasteel, graphite, and lead. some of these are in the random coin spawner.
image: moved coin icons into their own dmi
refactor: a sheet's coin type is now set on the stack obj definition, tidied up the mint code to not be a string of if/else if, and to not require a list of valid materials on the machine, so now it's easier to add new coin types. coin names also use material def names now.
/:cl:


code/game/objects/items/toys/fake_coin.dm
```diff
@@ -0,0 +1,81 @@
...
+	randpixel = 8
+	force = 0.0
+	throwforce = 0.0
+	w_class = ITEMSIZE_TINY
+	slot_flags = SLOT_EARS
+	var/sides = 2
+	drop_sound = 'sound/items/drop/ring.ogg'
+	pickup_sound = 'sound/items/pickup/ring.ogg'
+
+/obj/item/fake_coin/New()
```
Kashargul, line 20: Initialize(mapload()
    . = ..()



We no longer use NEW on atoms.

Kashargul (changes requested):

code/game/objects/items/toys/fake_coin.dm
```diff
@@ -17,47 +17,57 @@
 	drop_sound = 'sound/items/drop/ring.ogg'
 	pickup_sound = 'sound/items/pickup/ring.ogg'
 
-/obj/item/fake_coin/New()
+/obj/item/fake_coin/Initialize(mapload)
```
Kashargul, line 20: Init MUST call parent
. = ..()

Kashargul (approved):

PR #17236, opened 2025-02-28
Flamethrower TGUI + Tweaks

Cameron-The-Raven:
## About The Pull Request

Flamethrower moved to tgui. Projectiles changed to a modifier dragonflame, rebalancing pending. Extremely dangerous as is and needs tweaks. All other functionality should be the same, code cleaned up to remove sleep() and old item movement that forced loc instead of using forceMove().

## Changelog

:cl: Will
fix: The flamethrower UI now works properly.
add: Flamethrower now is more useful than spawning a few fire tiles. It's now an actual threat!
/:cl:


Cameron-The-Raven:
The flamethrower projectiles theirselves are _laughably_ weak...Whhich is fine, the main focus of the flamethrower is setting things on fire, it doesn't _need_ to do good damage.

Additionally, it doesn't really heat up a room to an INSANE degree. It heats it up a good amount, but gives plenty of time to get out if someone is using one. You have to have a very small room to get the temp up a significant amount.

When lingering fires are upported, it'll make this more of a legitimate tool (and less dumb with it's 'dump liquid fuel on whatever tiles we touch') for lighting things on fire good.

PR #17241, opened 2025-02-28
OOC escape text adjustment

Cameron-The-Raven:
Let's not assume abuse if it's being used.

## About The Pull Request
Adjusts OOC escape text to be less active-aggressive tone to a more passive tone.

## Changelog

:cl:
spellcheck: Adjusted the OOC escape text. Instead of starting strong with 'don't use this', let's be more passive tone.
/:cl:

### DIFF
--- a/code/modules/vore/eating/living_vr.dm
+++ b/code/modules/vore/eating/living_vr.dm
@@ -543,7 +543,7 @@
 			s.undo_prey_takeover(TRUE)
 			return
 		var/obj/belly/B = loc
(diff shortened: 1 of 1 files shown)


code/modules/client/preference_setup/vore/02_size.dm
```diff
@@ -111,9 +111,9 @@
 
 	else if(href_list["weight"])
 		var/new_weight = tgui_input_number(user, "Choose your character's relative body weight.\n\
-			This measurement should be set relative to a normal 5'10'' person's body and not the actual size of your character.\n\
-			If you set your weight to 500 because you're a naga or have metal implants then complain that you're a blob I\n\
-			swear to god I will find you and I will punch you for not reading these directions!\n\
+			Note: Scifi characters come in all shapes and sizes in this game, and not all follow the traditional shape of a human. Like a naga or a taur or a giant will weigh a\n\
```
SpadesNeil, line 114: At last, I correct this mistake from all those years ago.

PR #17244, opened 2025-03-01
code/global.dm => code/_global_vars/

ItsSelis:
## About The Pull Request

This PR moves the globals from the old code/global.dm file into the code/_global_vars/ directory, so that we can create them as managed globals.

## Changelog
:cl:
refactor: Moved all global vars from the code/global.dm to code/_global_vars/
/:cl:


Cameron-The-Raven:
tested, lgtm.

should be test merged but we don't have test merge here 😔

PR #17256, opened 2025-03-03
40 fps example config

Kashargul:
No more need to run the codebase at 20 FPS... it works fine on 40, so time to adjust the example


Kashargul:
> Probably worth updating the defaults in `code/controllers/configuration/entries/general.dm` too for FPS & Ticklag

I would keep the defaults, as those are long standing conditions, in case people want to fall back they see the original default and just lower the config.

PR #17266, opened 2025-03-04
load_ext update for the chatlogging library

ItsSelis:
## About The Pull Request

It is possible to preload specific external library calls for quicker later use (as of byond 516).
Since the chatlogging calls its external calls a lot (at least for the writes), this would make sense to be set.

Unfortunately two of the calls (that return something) had been converted into a proc for now.

Todo:

- [ ] Rebuild library binaries and add them to the PR

## Changelog
:cl:
code: external chatlogging library calls are even faster now
/:cl:


ItsSelis:
This will be done in a later PR once it is compatible. Parts of this PR were also superseeded by the 1664 patch.

PR #17287, opened 2025-03-08
Better sheet snatchers

SatinIsle:
## About The Pull Request

Added a new bluespace sheet snatcher of holding that can be created by science (cheaper and easier than other "of holding" items). This allows it to carry up to 5000 sheets. It can also be bought from the mining vendor for 1000 points.

Changed the normal sheet snatcher to hold 500 (previously 300), and the borg version to hold 700 (previously 500).

## Changelog
:cl:
add: Added a new bluespace sheet snatcher of holding that can be created by science (cheaper and easier than other "of holding" items). This allows it to carry up to 5000 sheets. It can also be bought from the mining vendor for 1000 points.
balance: Changed the normal sheet snatcher to hold 500 (previously 300), and the borg version to hold 700 (previously 500).
/:cl:


Kashargul:
Hmm can we get that one as borg upgrade, too?

code/game/objects/items/weapons/storage/bags.dm
```diff
@@ -276,7 +276,7 @@
 	icon_state = "sheetsnatcher"
 	desc = "A patented storage system designed for any kind of mineral sheet."
 
-	var/capacity = 300; //the number of sheets it can carry.
+	var/capacity = 500; //the number of sheets it can carry.
```
Kashargul, line 279: same here

PR #17312, opened 2025-03-11
Fixes swarm enemies behaving like walls when dead

TheCaramelion:
## About The Pull Request

Fixes an oversight that made the corpses have density. Additionally, projectiles will now check if the target is lying down instead of checking for density, allowing to shoot at swarm enemies without the need to click on them individually.

## Changelog
:cl: Guti
fix: Fixed swarm mobs becoming walls on death
code: Slightly changed projectile code to check for lying down mobs instead of their density
/:cl:

### DIFF
--- a/code/modules/projectiles/projectile.dm
+++ b/code/modules/projectiles/projectile.dm
@@ -554,7 +554,11 @@
 	else
 		var/mob/living/L = target
 		if(!direct_target)
-			if(L.lying)
+			// Swarms are special scuffed critters. They must have density FALSE to swarm, but then they don't get hit.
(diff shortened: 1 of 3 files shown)


code/modules/projectiles/projectile.dm
```diff
@@ -554,7 +554,7 @@
 	else
 		var/mob/living/L = target
 		if(!direct_target)
-			if(!L.density)
+			if(L.lying)
```
Kashargul, line 557: Does this mean we'll return true for all the non density mobs again?

PR #17321, opened 2025-03-12
More xenoarch artifact tweaks

Cameron-The-Raven:
## About The Pull Request
Makes destroying artifacts less punishing (if hit with long range)
Makes artifacts have a coinflip's chance to spawn active (to mitigate the 'sterility' that was noted due to xenoarch artifacts not starting active anymore) This EXCLUDES gas artifacts, to prevent hellfires.
Makes artifact deposits always visible, to mitigate people accidentally destroying them without intent of doing such.

## Changelog

:cl:
add: Artifacts now have a coinflip's chance of spawning active when unearthed (EXCLUDING gas artifacts)
qol: Artifacts are always visible at any depth now.
qol: Artifacts no longer apply a Z wide spawn, it will only apply effects for those within a small range.
/:cl:

### DIFF
--- a/code/modules/mining/mine_turfs.dm
+++ b/code/modules/mining/mine_turfs.dm
@@ -644,14 +644,14 @@ var/list/mining_overlay_cache = list()
 		var/pain = 0
 		if(prob(50))
 			pain = 1
-		for(var/mob/living/M in range(src, 200))
+		for(var/mob/living/M in range(src, 5)) //Let's only hit people nearby us.
 			to_chat(M, span_danger("[pick("A high-pitched [pick("keening","wailing","whistle")]","A rumbling noise like [pick("thunder","heavy machinery")]")] somehow penetrates your mind before fading away!"))
 			if(pain)
 				flick("pain",M.pain)
 			M.flash_eyes()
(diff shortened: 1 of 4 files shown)


Cameron-The-Raven:
so one problem with the randomness
/obj/machinery/artifact/Initialize(mapload)
    . = ..()

    if(ispath(artifact_master))
        AddComponent(artifact_master)

        artifact_master = GetComponent(artifact_master)

so artifact is made -> it gets to Init -> It makes the artifact_master -> artifact_master/New does do_setup() -> adds new artifact_effect(s) -> artifact_effect/New happens -> triggers the ToggleActivate -> does target.update_icon (target is the /obj/machinery/artifact here) -> /obj/machinery/artifact/update_icon() does the if(LAZYLEN(artifact_master.get_active_effects())) -> that returns active_effects (this is where it runtimes)

Cameron-The-Raven:
I conceded and forwent the update icon, unfucking that proc chain would be such hell for such a simple feature it's simply not worth it

Cameron-The-Raven:
made it do the update icon, just very cursedly

PR #17340, opened 2025-03-14
Removes baymiss. Mostly.

Cameron-The-Raven:
Goodbye Baymiss

## About The Pull Request
More or less removes baymiss and adjusts the logic behind the code

Details:

- Evasion is now calculated in get_zone_with_miss_chance. This means if you have a bunch of evasion, you are able to dodge blows more easily. Previously, evasion was only calculated for projectiles (but did a horrible job of that) but this opens it up so evasion actually has a purpose

- PvE combat feels less random. No longer will you miss 20 shots that are almost point blank. If the mob you're targeting doesn't  have a client, you'll hit (barring other factors like it having evasion, but in that case, it's /INTENDED/ to be dodging shots. I don't think we have any mobs that have evasion, but it's able to be calculated now.)

- PvE combat feels better on the PLAYER side. Before, even if you had 100000000 evasion, projectiles had a 30% chance of going 'nuh uh' and hitting you anyways. Now, evasion PROPERLY calculates. No more '30% chance of being hit by a projectile even if you're Neo.'

- Evasion now extends to melee attacks...This will probably need some tweaks if common sources of evasion is found and stacked together, but given evasion was a pretty rare modifier, this shouldn't be too much of an issue. If needed, evasion checks can be removed from the melee portion of it.

- Projectiles now have a 10% chance to target a random limb when attacking if they hit you through your evasion. This means they might target a harder to hit area (your hands) vs an easier to hit area (your chest) and have a chance to still miss their attack.

These numbers probably will need some tweaking and adjustment, but this should solve the big problem of baymiss

## Changelog

:cl:
qol: Baymiss removed/replaced.
qol: Evasion now works in melee.
qol: Projectiles will no longer ignore evasion 30% of the time
/:cl:

### DIFF
--- a/code/modules/mob/mob_helpers.dm
+++ b/code/modules/mob/mob_helpers.dm
@@ -175,24 +175,26 @@
 	//However, if a mob IS attacking a player, let's throw in some RNG into the mix to make it feel better for players.
 	//If a mob eats hits and dies, people are happy.
 	//If you shoot a mob point blank 10 times and every hit misses, people are upset (and rightfully so)
-	if(mob_misses)
-		var/randomization_chance = 10 //This can also be set to 0 to ensure mobs ALWAYS target the limb they're originally targeting.
 
-		/// First, we roll to see if we're going to target a random limb.
-		if(randomization_chance) //We got a 10% chance! Randomize where we're targeting!
-			zone = pick(base_miss_chance)
+	if(!mob_misses) //If mob_misses is disabled, they land their blow on the zone they're targeting.
+		return zone
 
-		// Second, we make sure to see if the place we are attacking is a valid area.
-		if(zone in base_miss_chance)
-			randomization_chance = base_miss_chance[zone]
+	var/randomization_chance = 10 //This can also be set to 0 to ensure mobs ALWAYS target the limb they're originally targeting.
+	/// First, we roll to see if we're going to target a random limb.
+	if(randomization_chance) //We got a 10% chance! Randomize where we're targeting!
+		zone = pick(base_miss_chance)
 
-		// Eyes and mouth can be targeted (although typically not by mobs) so we set it to the head.
-		else if (zone == "eyes" || zone == "mouth")
-			randomization_chance = base_miss_chance["head"]
+	// Second, we make sure to see if the place we are attacking is a valid area.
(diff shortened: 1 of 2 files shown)


Cameron-The-Raven:
> Bold move.
> 
> What does this mean for the accuracy traits though? Guessing those are redundant under this.

I don't even know if they were even calculated or used anywhere, I had completely forgotten that they exist.

I'll have to take a look at what var they change when I get home to see if it was being calculated, or if I can make it incorporated in some way with this new system.

They would only be useful for PvP ultimately, which might apply to some downstream but wouldn't really effect us here.

Cameron-The-Raven:
> > I don't even know if they were even calculated or used anywhere, I had completely forgotten that they exist.
> 
> I know how they were supposed to work, roughly, and did actually take the traits on a couple of characters. They don't come up very often but it'd still be nice to have something similar as a character flavour thing.

Yeee, I can see about making it take player accuracy into account by adding accuracy back but simply making it a negative modifier for the dodge-chance (so the attacker doesn't have to be passed in every time).

Cameron-The-Raven:
Alright I'm going to have to modify the description because this grew even larger than I expected.

- Fixed scoped weapons not properly retraining their scoped accuracy while scoped
- Added a toggle for 'always_hit' (Serverside toggle. Able to turn it so players will ALWAYS hit each other or not. Off by default)
- Added a toggle for PvP RNG missing (Players missing players. off by default)
- Added a toggle for PvE RNG missing (Mobs missing players. On by default)
- Accuracy is back, but ONLY affects things if the thing you're attacking has evasion chance. (This means accurate weapons are good against high evasion individuals)
- Bad Shot / Eagle Eye act as a 1:1 detractor/additive to evasion. If someone has no evasion, doesn't affect anything. If they DO have evasion, it either subtracts or adds to their evasion chance.
- Point-blank shots will always hit player characters, evasion or not.

Cameron-The-Raven:
as an aside, this all stemmed from the fact I wanted to build a melee only PSG for people that want to do melee instead of ranged (because ranged is literally 10000% more viable than melee at all times) and noticed that evasion had a 30% chance of going 'lolno'

so it gets modernized instead

PR #17348, opened 2025-03-15
no more self callbacks on update fullness

Kashargul:
This needs testing, but we should no longer need this
## Changelog
:cl:
code: removes self callbacks on update fullness
fix: belly color resets / bad loads
/:cl:


Kashargul:
Can be integrated into the large git diff

Kashargul:
included in #17350

PR #17362, opened 2025-03-16
fix some issues

Kashargul:
## Changelog
:cl:
fix: get_space_zlevels returning strings instead of z levels in some cases
/:cl:

### DIFF
--- a/maps/tether/tether_defines.dm
+++ b/maps/tether/tether_defines.dm
@@ -287,7 +287,7 @@
 	announce_atc(AM,going = TRUE)
 
 /obj/effect/overmap/visitable/sector/virgo3b/get_space_zlevels()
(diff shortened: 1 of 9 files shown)


maps/tether/tether_defines.dm
```diff
@@ -287,7 +287,7 @@
 	announce_atc(AM,going = TRUE)
 
 /obj/effect/overmap/visitable/sector/virgo3b/get_space_zlevels()
-	return list(Z_LEVEL_TETHER_SPACE_LOW)
+	return list(GLOB.map_templates_loaded[Z_LEVEL_TETHER_SPACE_LOW])
```
Drulikar, line 290: I don't think this change is what you want.

You need to parse any Z_NAME defines, but not the Z_LEVEL defines. They are not late loaded and are already a Z number.

PR #17384, opened 2025-03-20
Resizing & better runechat

Cameron-The-Raven:
## About The Pull Request
Sprites no longer break the resize limit when uncapped = FALSE
Sprite decals also adjusted. Now get generated BEFORE tummies, so they don't overlay over stomachs.
Runechat is now calculated better.
Code by Kash
## Changelog
:cl:
code: Sprites no longer break the resize cap when resized, unless uncapped is set to TRUE
code: Robot sprite_decals are now generated before stomachs
fix: Runechat will properly appear above sprites.
/:cl:


Cameron-The-Raven:
made it so RESIZE_NORMAL is always selectable

PR #17388, opened 2025-03-20
you can grippy the floor tiles now as borg

TankTheBirb:
## About The Pull Request

I had this cool idea for a like, storage box of tiles that you could kinda just slurp up and store so you could redistribute them later. But then I realized that was more complex than I wanted to spend time on and all borgs needed was literally any method of interacting with stacks of floor tiles beyond visual contempt. so this is that.
## Changelog
:cl:
qol: added the ability to pick up floor tiles to the sheet loader gripper that borgs have

/:cl:

### DIFF
--- a/code/modules/mob/living/silicon/robot/drone/drone_items.dm
+++ b/code/modules/mob/living/silicon/robot/drone/drone_items.dm
@@ -253,11 +253,12 @@
 
 /obj/item/gripper/no_use/loader //This is used to disallow building with metal.
 	name = "sheet loader"
-	desc = "A specialized loading device, designed to pick up and insert sheets of materials inside machines."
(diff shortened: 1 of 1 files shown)


TankTheBirb:
Not sure what testing was requested, I might've missed the request. But I've verified that:
- sheet loader can now pick up floor tiles
- sheet loader can only pick up one type of floor tiles at a time
- sheet loader cannot pick up a stack of other sheets while holding tiles
- sheet loader can only pick up floor tile amount to its maximum stack (60)
- sheet loader can put floor tiles into a machine object (such as an autolathe)
- sheet loader can place floor tiles back on the ground via dropping OR placing them onto tiles that can take floor tiles

Lemme know if I missed something, it didn't generate any runtimes from doing this so I assume it fine.

PR #17414, opened 2025-03-23
Fixed private notes panel

SatinIsle:
## About The Pull Request

Fixed a bug with the private notes panel that showed it in right click menus of other mobs (forcing them to see your notes).

## Changelog
:cl:
fix: Fixed a bug with the private notes panel that showed it in right click menus of other mobs (forcing them to see your notes).
/:cl:

### DIFF
--- a/code/modules/mob/living/private_notes.dm
+++ b/code/modules/mob/living/private_notes.dm
@@ -1,4 +1,6 @@
 /mob/living/proc/private_notes_window(mob/user)
+	if(user != src)
+		return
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/private_notes.dm
```diff
@@ -1,4 +1,6 @@
 /mob/living/proc/private_notes_window(mob/user)
+	if(user != src)
```
Kashargul, line 2: all the user == src checks below are no longer needed

PR #17418, opened 2025-03-25
Removes sleeping from the code in more places

Cameron-The-Raven:
## About The Pull Request

Mostly putting this here so that it can be looked at and examined what should be done where.
This is probably butchering the code in some places, but hey, that's what it's here for.
All of these (minus the robot one and hitscan one because I checked that myself) need to be examined further to ensure it didn't break anything.

1. Removes stoplag in hitscan. This was 100% unused and seems to have been added in 6 years ago and then just...axed at some point. I don't really care to dig into it's future too much. It wasn't used, so whatever. (Doesn't need testing...Unused var.)

2. Anything involving ghost queries had their ghost query calls happen in LateInit (Tested and works https://i.imgur.com/MVWbrBr.png..Also bugfixed borers)

3. pick_module was moved from /init on robot_module to /LateInit on /robot. Triple checked to make sure it still works and yeah it does https://i.imgur.com/5rGbqIR.png (Tested and works)

4. ~~AI controlled humans had their death proc moved into LateInit  (NOT tested)~~ Gets rid of their death proc...It was unused and broken from the start. Added 6 years ago and never used. Just spawn a corpsespawn and NOT an AI mob.

5. Scrubbers only have their TGUI occur when a user is interacting with it. Splits the proc into two. One that is user controlled and one that is not.  (TESTED: https://i.imgur.com/6yOEW1o.png)

6. Corpse landmarks now do the actual death() on lateload, since that was what was causing massive problems with corpses. (Tested and works)

7. Mob chasing moved to parent instead of being on ian and the snake (Tested)

8. The TGUI input for exit_vr has been fixed by splitting the proc up. (Tested and works)

9. Promethean and protean blob sleep when they don't have humanforms upon death has been removed. (They will ALWAYS have a humanform)

10. Gets rid of the 'machine all suspend when the mapload is beginning' and swaps it to only do the atmos machinery (since those were the ones that could created, but then would have problems as it could process() before it could be init'd) (Tested and works https://i.imgur.com/VkzciPt.png)

11. More blob fixes.

This leaves ONE LAST SLEEP IN INIT and that is /obj/effect/landmark/poi_loader/Initialize

Which... InitializeAtoms and setup_atmos_machinery both have CHECK_TICK in there. Can't fix that for now.

## Changelog
:cl: Diana
del: bad code
add: good code (questionable)
fix: VR Sleepers properly exit you if EMP'd, depowered, or someone tries to pull you out of it
fix: Ghostspawns no longer multiply SECONDS by SECONDS
/:cl:


Kashargul:
@Cameron-The-Raven can we merge that on monday, that I can give it a full week TM on chomp?

Cameron-The-Raven:
Yeh sounds good to me
Everything seems to work on paper and in testing but who knows what the test merge could unearth.

Kashargul:
> Yeh sounds good to me Everything seems to work on paper and in testing but who knows what the test merge could unearth.

Ye, I really would like to have that off the tracker before continuing with new to init

PR #17428, opened 2025-03-26
adds a trait that lets you chase people down the halls at 2am 

TankTheBirb:
## About The Pull Request

**Unusual Gait** is a trait that modifies your movement delay by -0.5 (aka: one whole haste or taj base speed) so long as your hands are empty. Other slowdowns still apply so heavy items and armor will cut your speed but not deactivate the trait. Currently, it is not locked to custom species only because of two thinkies:
- I bothered to check who gets their slowdown changed at the species level. And if you do? no trait for you
- Movement as a balance concern is most key when it comes to combat/damage because of kiting strategies. However, it is pretty damn hard to fight with only your unarmed attacks and thus the majority of the mechanical advantage around this trait is removed simply by revoking hand privileges (then you spent 2 points for nothing). Theoretically you can get around this by using a hardsuit integrated weapon but those kinda suck if imma be honest. 

So it's an RP trait, that does mechanics. The initial vision was to allow people who play species that might have a faster "natural gait" that doesn't work well with holding things (ie, running on all fours) to get a lil speed boost when they aren't holding stuff. But if you want to roleplay this as your character naruto running around then by all means, go wild lol.

Also it's mutually exclusive with haste. I felt that was kind of an obvious thing to do, but it's worth mentioning.

https://github.com/user-attachments/assets/25c28188-27d4-4c4c-b7e6-6613434d0232

## Changelog
:cl:
add: Unusual gait positive trait for 2 points. This trait makes you move faster when your hands are empty.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human_movement.dm
+++ b/code/modules/mob/living/carbon/human/human_movement.dm
@@ -115,6 +115,14 @@
 			. *= 0.5
 		. -= chem_effects[CE_SPEEDBOOST]	// give 'em a buff on top.
 
+	if(ishuman(src)) //r u human
+		var/mob/living/carbon/human/H = src //wowie you are, take this badge
+		if(species.unusual_running == 1) // do you have the trait
+			var/obj/item/I = H.get_active_hand() // checking their hand
+			var/obj/item/OH = H.get_inactive_hand() // and their other hand
+			if(!istype(I,/obj/item) && !istype(OH,/obj/item)) //better not have any items on you mfer
+				. += -0.5 // ok vibe check passed, take this small movement buff and leave
+
 	. = max(HUMAN_LOWEST_SLOWDOWN, . + CONFIG_GET(number/human_delay))	// Minimum return should be the same as force_max_speed
 	. += ..()
 
--- a/code/modules/mob/living/carbon/human/human_movement.dm
+++ b/code/modules/mob/living/carbon/human/human_movement.dm
@@ -121,7 +121,7 @@
 			var/obj/item/I = H.get_active_hand() // checking their hand
 			var/obj/item/OH = H.get_inactive_hand() // and their other hand
 			if(!istype(I,/obj/item) && !istype(OH,/obj/item)) //better not have any items on you mfer
-				. += -0.5 // ok vibe check passed, take this small movement buff and leave
+				. -= 0.5 // ok vibe check passed, take this small movement buff and leave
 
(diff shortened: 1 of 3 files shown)


TankTheBirb:
`[2025-03-26T15:50:37] Runtime in code/modules/admin/verbs/lightning_strike.dm,31: Cannot read null.z
  proc name: lightning strike (/proc/lightning_strike)
  src: null
  call stack:
  lightning strike(null, 1)
  Electric Field (/datum/artifact_effect/electric_field): DoEffectAura()
  Electric Field (/datum/artifact_effect/electric_field): process()
  /datum/component/artifact_mast... (/datum/component/artifact_master): process(20)
  Objects (/datum/controller/subsystem/processing/obj): fire(0)
  Objects (/datum/controller/subsystem/processing/obj): fire(0)
  Objects (/datum/controller/subsystem/processing/obj): ignite(0)
  Master (/datum/controller/master): RunQueue()
  Master (/datum/controller/master): Loop()
  Master (/datum/controller/master): StartProcessing(0)`
  unit test found this in a file I didn't touch so uh, restarting this lol

code/modules/mob/living/carbon/human/human_movement.dm
```diff
@@ -115,6 +115,14 @@
 			. *= 0.5
 		. -= chem_effects[CE_SPEEDBOOST]	// give 'em a buff on top.
 
+	if(ishuman(src)) //r u human
+		var/mob/living/carbon/human/H = src //wowie you are, take this badge
+		if(species.unusual_running == 1) // do you have the trait
+			var/obj/item/I = H.get_active_hand() // checking their hand
+			var/obj/item/OH = H.get_inactive_hand() // and their other hand
+			if(!istype(I,/obj/item) && !istype(OH,/obj/item)) //better not have any items on you mfer
+				. += -0.5 // ok vibe check passed, take this small movement buff and leave
```
Kashargul, line 124: I feel like those all can be stacked?

also -= please

TankTheBirb: what do you mean by stacked? I'm not familiar with that unfortunately. And yeah -= not being  used was a derp on my part,  I'll fixy

Kashargul: If you use chems and the trait, how far will it let you speed up as tesh?

TankTheBirb: made the -= change. Also tested with hyperzine and it seems to work as expected so it should be fine

PR #17435, opened 2025-03-27
Update mob_vr.dm nutrition_messages

SpadesNeil:
## About The Pull Request

This is something I have wanted to do for _years._

Ever since Aronai and Leshana added the persistent weight gain mechanic, these ancient pre-Polaris and pre-Baystation descriptions have been made hugely obsolete. They also massively pre-date the addition of actual visible bellies.

As a consequences, for the last 6 years, these descriptions have felt horribly out of place.

This is an attempt to rewrite them not to describe how big a person's belly is, but instead to only describe just how stuffed and glorpy that tummy is, big or small. The goal is to create something that serves as a good baseline for the average character. I tried to keep it vague enough to work with most typical digestive organs a player might have in their body. I can't account for edge cases like slime people or whatever, but this isn't _for_ them. They can write their own or blank it out. I can't cover every scenario, but I can cover most of them better than the old text.

My goal was to write lines on a scale of 1 to 10 between "absolutely starving" and "**_so fucking engorged with calories that your body literally can't take any more and you're just wasting food at this point because your body can't even absorb all of it_**."

For reference, you have to be at nearly a whopping 6,000 nutrition to hit that last level. This is usually only accomplished by digesting multiple players in a shift.

I strongly suggest you come back and re-adjust these descriptions based on actual player behavior and play testing against normal food intake for reference. This is supposed to serve as a new baseline. It's already better than the old one, but I somewhat struggled making it vague but appropriate.

Anyone who uses the old texts on their character should be unchanged.

## Changelog
:cl: Ace
qol: New default nutrition examine texts.
/:cl:

### DIFF
--- a/code/modules/vore/eating/mob_vr.dm
+++ b/code/modules/vore/eating/mob_vr.dm
@@ -30,17 +30,18 @@
 	var/allow_mimicry = TRUE 	// Allows mimicking their character
 	var/allow_mind_transfer = FALSE			//Allows ones mind to be taken over or swapped
 	var/nutrition_message_visible = TRUE
-	var/list/nutrition_messages = list(
-							"They are starving! You can hear their stomach snarling from across the room!",
-							"They are extremely hungry. A deep growl occasionally rumbles from their empty stomach.",
+	var/list/nutrition_messages = list( // Someday I would like for it to be possible to name the character or refer to specific pronouns... but that is out of scope for me. -Ace
+							"You can hear their empty stomach snarling from across the room. They must be starving!",
+							"You notice a faint growling occasionally rumble from their hungry gut.",
 							"",
-							"They have a stuffed belly, bloated fat and round from eating too much.",
-							"They have a rotund, thick gut. It bulges from their body obscenely, close to sagging under its own weight.",
-							"They are sporting a large, round, sagging stomach. It contains at least their body weight worth of glorping slush.",
-							"They are engorged with a huge stomach that sags and wobbles as they move. They must have consumed at least twice their body weight. It looks incredibly soft.",
-							"Their stomach is firmly packed with digesting slop. They must have eaten at least a few times worth their body weight! It looks hard for them to stand, and their gut jiggles when they move.",
-							"They are so absolutely stuffed that you aren't sure how it's possible for them to move. They can't seem to swell any bigger. The surface of their belly looks sorely strained!",
-							"They are utterly filled to the point where it's hard to even imagine them moving, much less comprehend it when they do. Their gut is swollen to monumental sizes and amount of food they consumed must be insane.")
+							"They seem quite content, if perhaps a little stuffed. Their stomach gurgles with a little more food than is necessary at the moment.",
+							"They look very satisfied, subconsciously licking their lips as their digestive system churns along an excessively rich meal.",
+							"Their face is pleasantly flushed, breathing a bit more heavily than usual as their belly works some mighty feast into soupy nutrients. There's enough sloshing around in there to leave them feeling bloated for hours.",
+							"Their face looks a little woozy while their breathing is somewhat panting. If you listen closely, you can hear a series of lurid wet blorps as their body works down what would have been enough calories to feed a small party.",
+							"Their expression looks dazed and maybe a little uncomfortable while their breathing consists of deep, long, queasy-sounding gasps. If you pay attention, you can make out the constant gushing from their insides as digestion struggles to cope with a whole buffet worth of mush.",
+							"Their expression looks downright hedonistic while their breathing is slow and lethargic. You can plainly hear a series of thick, lurid glorps as their gut strains to process nearly its limit worth of densely packed chyme.",
+							"Their drooling expression looks pale and clammy, their breathing is weak and heaving, and you vividly overhear the emulsion of calorie-dense foodstuff working slowly through their guts with awful, bubbling GROANS... Maybe they don't consciously realize it, but they are definitely at their limit. Eating more won't make a difference at this point; their body couldn't soak up any more nutrients even if they want it to."
+							)
 	var/weight_message_visible = TRUE
 	var/list/weight_messages = list(
 							"They are terribly lithe and frail!",


TheToaster98:
Question.
Isn't this largely irrelevant, since we have a system to customize your nutrition levels per character?

Also reading these I can see a clear bias for stuffing and weight and stuff that I don't think people (who like me) may not be into it, and probably don't want to read blatant kink posting against their prefs.

Kinks aside the last ones read like they are dying. It is... Deeply uncomfortable.

SatinIsle:
Strongly prefer the originals to the ones that you have written, as the originals are more neutral and focus specifically on the stuffing itself. I do not like the assumptions that yours make about the character and they're clearly biased by your own personal preferences.

TeleTubby2:
Yeah, given these are things meant to be overridden on a per-character basis, I think it's better to leave them fairly neutral in terms of content as a result (and tbh, probably more neutral than they are now). Especially as the new additions dictate the character in question's reaction to it as opposed to just diagnosing them with fat fuck, which is ehhhh even if the content was super mundane.

Kashargul:
> Seconding all of the above, mostly on the grounds of enforced character reaction.
> 
> What the systems _actually_ need is a better explanation and outline of what the steps are, so that people that they can, and how to, customize them to their tastes.
> 
> If people don't bother to do that? Well, that's not something that needs fixing.

That is more of a future todo for the new TGUI. Check bingle once merged for a few options. Currently waiting on some major overhauls on tgui core before starting the vore panel rewrite.

SpadesNeil:
[TheToaster98](https://github.com/TheToaster98):
> Question. Isn't this largely irrelevant, since we have a system to customize your nutrition level messages per character?

Yesn't.
These are the defaults for if you don't know you can set those or choose not to.\

> Also reading these I can see a clear bias for stuffing and weight and stuff that I don't think people (who like me) may not be into it, and probably don't want to read blatant kink posting against their prefs.

Again, the goal of this is to try to set these up so _most people_ will be fine with what it says. The people who are eating to extreme levels are going to be doing that anyway. I don't know a lot of people inside the vore community who are actively put off by stuffing.

What I **don't** want is to create something that goes to such an extreme that it's difficult to ignore. Like I don't like herms. Reading "shi" and "hir" bothers me. But it doesn't bother me so much that I think that shouldn't have been added, and in fact, it was added _for_ those players with the understanding that other people wouldn't like it and my response was simply, _"Not to be mean but dude get over it."_ The benefit outweighs the tiny handful of people who will be badly squicked out by it. I feel the same way about this change.

If you're concerned about _other people's_ prefs who are observing, then the default examines should be removed entirely.

_That all said..._

> Kinks aside the last ones read like they are dying. It is... Deeply uncomfortable.

This is an example of what I **don't** want.

So I should revise those lines, assuming that this idea isn't already dead in the water. I am open to suggestions on how to better make these lines neutral while still explaining how full the person is.

Do bare in mind those are levels where someone has to have eaten _multiple_ other mobs to get to. Do also bare in mind that the over-eating text alerts, which cannot be changed by players by the way, also have varying levels of extreme discomfort. That's in another part of the code. I made those changes years ago.

[SatinIsle](https://github.com/SatinIsle):
> Strongly prefer the originals to the ones that you have written, as the originals are more neutral and focus specifically on the stuffing itself, which is how I feel it should be for the default options. I do not like the assumptions that yours make about the character and **they're clearly biased by your own personal preferences.**

It's weird you say that last part because that's exactly the problem I feel the original text actually have. The goal was to try to avoid that but I seriously need better feedback. The old texts, I feel, did not accomplish the goal that they should have especially once you start getting into the more crazy levels of eating.

The goal with these new texts was to _try to_ replace them with something that isn't referencing the actual size of their belly and to somehow explain other common physical signs of someone who had over-eaten to various degrees.

I honestly wasn't satisfied with the original, but I wasn't sure about my suggested replacement. I **still** think the originals should be replaced because they **are** a reflection of **my own** prefs, and that's explicitly why they aren't good. However, because I can only frame it with my own prefs, it's hard to make a series of replacements.

I want suggestions if you think these versions are no good. I'd also be open to just deleting them entirely, but then I fear as time goes on players won't edit their own descriptions because it'll be such an afterthought that they won't bother to do so or won't even know what the customization option is for!

[TeleTubby2](https://github.com/TeleTubby2):
> Yeah, given these are things meant to be overridden on a per-character basis, I think it's better to leave them fairly neutral in terms of content as a result (and tbh, probably more neutral than they are now). Especially as the new additions dictate the character in question's reaction to it as opposed to just diagnosing them with fat fuck, which is ehhhh even if the content was super mundane.

Again, I'm surprised you guys actually preferred the originals, because they're the ones that **are** written in language I feel wasn't very neutral about people's preferences or fetishes.

[KillianKirilenko](https://github.com/KillianKirilenko):
> Seconding all of the above, mostly on the grounds of enforced character reaction.
> What the systems _actually_ need is a better explanation and outline of what the steps are, so that people that they can, and how to, customize them to their tastes.

This is true too, but while I considered just removing the default examines, I was concerned people would slowly forget this whole thing even exists, or will misunderstand how it works.

code/modules/vore/eating/mob_vr.dm
```diff
@@ -30,17 +30,18 @@
...
+							"Their face looks a little woozy while their breathing is somewhat panting. If you listen closely, you can hear a series of lurid wet blorps as their body works down what would have been enough calories to feed a small party.",
+							"Their expression looks dazed and maybe a little uncomfortable while their breathing consists of deep, long, queasy-sounding gasps. If you pay attention, you can make out the constant gushing from their insides as digestion struggles to cope with a whole buffet worth of mush.",
+							"Their expression looks downright hedonistic while their breathing is slow and lethargic. You can plainly hear a series of thick, lurid glorps as their gut strains to process nearly its limit worth of densely packed chyme.",
+							"Their drooling expression looks pale and clammy, their breathing is weak and heaving, and you vividly overhear the emulsion of calorie-dense foodstuff working slowly through their guts with awful, bubbling GROANS... Maybe they don't consciously realize it, but they are definitely at their limit. Eating more won't make a difference at this point; their body couldn't soak up any more nutrients even if they want it to."
```
SpadesNeil, line 43: Assuming the idea isn't dead in the water, these last few lines absolutely need to be changed, and I would request someone to help rewording it to something else.

For context this is the upper **max limit** of how much a player can eat without doing vore. The limit exists because certain players have abused just eating the entire kitchen (and still do) and this proved extremely hard to stop through moderating because it often happens when others aren't looking. The gulp sounds of item vore are more noticeable than the normal munching sounds which get ignored. That's the only reason the max limit exists at all. And also because I can only write so many descriptions before the heat death of the universe.

But what I **don't** want is for people to compare this extreme level of eating with thinking the person is literally gonna drop dead.

SpadesNeil:
Let's try a different direction.

Let's pretend I didn't make a PR about this at all.

**What problems, if any, do you have with the original lines?** How would you change them if there are problems?

SpadesNeil:
Closing this since I am not confident enough to want it merged, but I still wanna discuss it.

PR #17440, opened 2025-03-29
New nutrition texts, attempt #2

SpadesNeil:
## About The Pull Request

Same goal as https://github.com/VOREStation/VOREStation/pull/17435 but hopefully way better.

## Changelog
:cl:
code: Changes the default nutrition texts for new characters.
/:cl:


SpadesNeil:
That should be enough changes for now.

Kashargul:
Why does every line of code has an arbitrary comment?

Comment important things... not every single line

SpadesNeil:
> Why does every line of code has an arbitrary comment?

Because _apparently_ there was some amount of confusion on what these lines did, and I had to explain myself.

I don't really care. If you don't want so many comments, just delete them. I frankly don't want them there either.

ShadowLarkens:
Sorry for the long wait! 

After discussion with staff, we've decided that we're fine with the current belly texts and would rather not see them changed.

PR #17466, opened 2025-04-02
DNA Clone Fix

Willburd:
## About The Pull Request
In a prior change dna was cloned per organ to avoid the GC locking up from multiple hardrefs to the same dna datum. However, cloning the full dna datum is very costly, and caused substantial lag on character spawn.

## Changelog
Organs have been altered to use an organ_data datum, with all the dna entries that they actually use, instead of a full dna or species datum copy. While I cannot fully achieve what I wanted to with species, the current code is still functional. I would require a full recode of transforming species to achieve the function I want. 

Organ references to original species have been moved to weakrefs, and cache their data if the species datum is ever deleted. This is to prevent any future work on deleting hanging species datums, though they are currently not deleted at all, and are a memory leak. Deleting them is currently too dangerous due to mixed global and local species datums.

This also makes tracking bugs slightly less confusing, as there is no longer a dna datum in every organ. So finding errors where "they're entire dna was deleted" only to learn it was their hand's dna that was deleted... Should be easier from the outset, as the datums are entirely different.

:cl:
qol: spawning carbon mobs is less laggy
refactor: dna datum cloning per organ removed in favor of a organ_data datum
/:cl:

### DIFF
--- a/code/__defines/species_languages.dm
+++ b/code/__defines/species_languages.dm
@@ -12,6 +12,7 @@
 #define NO_INFECT		0x400	// Don't allow infections in limbs or organs, similar to IS_PLANT, without other strings.
 #define NO_DEFIB		0x800	// Don't allow them to be defibbed
 #define NO_DNA          0x1000	// Cannot have mutations or have their dna changed by genetics/radiation/genome-stolen.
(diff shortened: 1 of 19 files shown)


Kashargul:
Todo me: tracy

Kashargul:
grafik

PR #17476, opened 2025-04-04
split tgui html

Kashargul:
just a repackage

Also porting https://github.com/tgstation/tgstation/pull/90340

🆑 
fix: some badly handled HasProximity
code: splits the tgui html
/🆑


Cameron-The-Raven:
Opened a TGUI panel with it to make sure it works and yeah, looks good 


if nothing else specific needs to be tested, this has my approval

PR #17492, opened 2025-04-08
Fixes stumble & flight vore

Cameron-The-Raven:
## About The Pull Request
Fixes an issue where stumble and flight vore where sometimes the prey would come back out from the stomach if the person stumbling into them was the pred.

In depth:
Stumble vore happens during Crossed which is in the middle of Move() but before Moved() is called. This means that if the prey was the one moving, no problem, they already did their move and get moved into the gut.
But if the PRED was the one moving, the prey would sometimes be moved due to how the crossed interactions work, so the prey would get eaten and then before the tick ends they'd be properly 'moved', causing them to move out from the pred's stomach and onto the tile.

By invoking ASYNC, we allow the moved proc to properly go through and THEN have the prey get eaten. As far as you can see in the game this is instant.

This appears to have been ORIGINALLY doing its own invoke_async by doing a delay of 1 when stumble vore or dropnoms happened which wasn't working for some reason.
## Changelog
:cl: Diana
fix: Stumblevore and flight vore should now work properly.
/:cl:


code/modules/mob/living/simple_mob/subtypes/animal/alien animals/space_mouse.dm
```diff
@@ -84,7 +84,7 @@
 	if(vore_fullness == 0 && movement_cooldown == 10)
 		movement_cooldown = initial(movement_cooldown)
 
-/mob/living/simple_mob/vore/alienanimals/dustjumper/perform_the_nom(mob/living/user, mob/living/prey, mob/living/pred, obj/belly/belly, delay)
+/mob/living/simple_mob/vore/alienanimals/dustjumper/perform_the_nom(mob/living/user, mob/living/prey, mob/living/pred, obj/belly/belly, delay_time, no_delay)
```
Kashargul, line 87: Why do we need 2 vars here???

delay_time of 0 implies no delay....

PR #17493, opened 2025-04-08
Makes FBP not go blind from welders

Cameron-The-Raven:
## About The Pull Request
See title.
Also makes it so the eye color verb is actually attached to the eyes.

## Changelog

:cl:
qol: FBPs will no longer go blind when a welder is used.
/:cl:

### DIFF
--- a/code/modules/organs/internal/eyes.dm
+++ b/code/modules/organs/internal/eyes.dm
@@ -10,7 +10,10 @@
 /obj/item/organ/internal/eyes/robotize()
 	..()
 	name = "optical sensor"
+	innate_flash_protection = FLASH_PROTECTION_MAJOR
 	verbs |= /obj/item/organ/internal/eyes/proc/change_eye_color
+	organ_verbs = list(/obj/item/organ/internal/eyes/proc/change_eye_color)
+	handle_organ_mod_special()
 
 /obj/item/organ/internal/eyes/robot
 	name = "optical sensor"


Cameron-The-Raven:
Hmm, yeah that's a problematic side effect from it. This was an upport of the fix to FBPs going blind when they do the equivalent of putting a bruise pack on theirselves.

Let me find a different way to go about this.

Cameron-The-Raven:
Changed it to a different way that ONLY will effect when FBPS use a welder to heal so they don't get the flash or flashbang protection.

TeleTubby2:
tbh I'm not sure if this is something we should encourage to begin with. If they're qualified, they either have the NIFSoft or the eye protection already (Engineering/Robotics)

If welding was a simple thing, that'd probably be fine, but it's a lot more complicated than putting a bandage on - we have something adjacent to that anyway via nanopaste.

Cameron-The-Raven:
> tbh I'm not sure if this is something we should encourage to begin with. If they're qualified, they either have the NIFSoft or the eye protection already (Engineering/Robotics)

The fact that NIFs are able to turn this mechanic off entirely is making the point that synths _should_ have this even stronger. It's extremely bad gameplay design.

If I have a NIF, I can go up to the nifsoft at round start and **immediately invalidate the welding mechanic entirely** (even as an organic) for a small 225 thalers. The only difference is that if you're playing a character that doesn't want a NIF, you'll be permanently blinded for the rest of the round when you do need to heal yourself.

Meanwhile, proteans don't even need a welder, they can just wrench a single chair, table, weld a wall (they ALSO don't get blinded by welders) and then blobform for a little bit and go back to 100% health.

Organics can just spam-click their limbs with a bruise pack and they're healed.

Meanwhile, as a FBP, if I get damaged and heal myself (the equivalent of using a bruise pack) I get completely blinded, barring me from seeing emotes for the rest of the game until I get eye surgery.

I've looked at other RP and combat oriented servers and as far as I can see, it's pretty universal that FBPs can heal theirselves with a welder.

If we started to include nanopaste everywhere that bruise packs can be gradded from, then having them not be able to use welders to heal up would be reasonable (they would have an alternative) but currently, they're massively punished for healing and you **will** go blind when you decide to heal.


image

Cameron-The-Raven:
I'd argue that FBPs should be be qualified to do basic self-repairs on theirself, much like how organics can do basic self-healing w/ bruise packs. It's basic self repair. 

And, again, this is looking past the fact that it's extremely poor gameplay design.

The welder _already_ has a limit for healing FBPs. A FBP that is severely damaged _can't_ heal theirself and has to go to robotics to get healed. It's the equivalent of someone breaking a bone, except in FBP's case, they can't even temporarily heal the damage, as it requires robotics to intervene to fix it.

**Everyone** (that has a NIF) already has access to protection to welder damage. There's 0 qualifications required. The NIFsoft is not access locked, nor are NIFs gated in any way from the general population. You can simply go up to the vendor at round start (if you have a NIF) and buy the UV protection.

Because at the moment, if I have a NIF, I can just buy the nifsoft at round start and heal myself with a welder no problem. 
If I don't have a NIF, I just have to get a NIF and I can heal my limbs forever for the rest of time with no penalty and not get perma-blinded.

TeleTubby2:
TBH that reads to me as more a fundamental issue of "someone forgot to put access on the NIFsoft/left it unlocked so people can build vibe hangouts" rather than "This was an intentional decision so robots can go heal themselves scott free if they have a NIF".

This is even before the consideration that repairing a synthetic is always going to be (ICly) far more complicated than healing an organic, up to the point of needing surgical intervention, mostly due to how the healing items are represented.

Bruise packs are as the name implies, slap one on. possibly put some ointment. Smaller cuts can be bandaged, because the body does the hard part of healing on its own. The baseline for medical complexity in regards to doing synth repair (at least ICLY) is probably around fixing a phone, solely because it can't fix itself like an organic body. 

Synths don't get that luxury, which means the (IC) task of actually fixing one is far more complicated. obviously this is abstracted away, and varies based on character or whatever, but it's the difference between "Eh, that bruise will go away on it's own eventually", and trying to un-dent some part of your car. it requires finesse and an actual IC skillset, of which I don't think it's healthy to imply that literally every robot should have (ae, otherwise - why would the medical android have that hardware?)

Cameron-The-Raven:
> TBH that reads to me as more a fundamental issue of "someone forgot to put access on the NIFsoft/left it unlocked so people can build vibe hangouts" rather than "This was an intentional decision so robots can go heal themselves scott free if they have a NIF".

But that's the thing. The solution has been _easily_ available to the problem synths have had for years - the only downside is you _have_ to get a NIF. Which, if you want to heal your own damage as a synth, you're going to do. 

Realistically, yes, repairing a synth would be more difficult than putting a bandaid on a wound, but that is still missing the point that it is a _very_ unfun game mechanic that forces interaction. But we're not trying to get a perfectly realistic station. Realism and game mechanics often don't mix well.

If a character _wants_ to be able to say 'I don't want to know how to repair my synthetic body' then that is their decision to say and do that, much like if someone wants to not be medically trained to treat their own minor injuries they sustain and go to medical instead. The fact that synths won't get permanently blinded and be unable to interact with the game for the rest of the round until they get eye surgery doesn't change that.

However, blocking characters _off_ from being able to heal theirself (or be forced to do a risk-reward calculation of 'Alright, how many times have I healed myself so far? Am I reaching the blind threshold yet?') is a very unfun game mechanic, which is the issue here.

If it wasn't for the NIFsoft, FBPs would be forced to go to robotics for every injury they sustain instead of healing. Which is just forcing engagement at that point. It's the same exact reason we disabled the degrading genetics perk, as it occasionally forcing people to go to medical by giving them a damage type they couldn't heal theirself.

The realism aspect isn't the point of this PR. The main point of this PR is that FBPs are either in their current stated either forced to get a NIF to heal theirself even if they don't _actually_ want a NIF (which will continue to happen), get a welder mask/welder goggles from somewhere, or go to robotics every time they take minor damage so they can get their health back to 100%.

The first of those two have no RP aspect whatsoever, and thee third brings nothing to the table and generates no actual engaging RP. Just clicking body parts with a welder and then saying bye to the roboticist or whoever healed you. (Much like how the degrading genetics trait required you to go to medical to be given a genetic restoration pill to heal the 2 damage you incured, which would on average only happen ~1 or 2 times in a full shift)

FBPs already are forced - much like organics are forced - to go to to their respective 'doctors' when they sustain over a certain amount of damage on a limb. 

Meanwhile, again, proteans can just completely avoid this mechanic entirely by eating a sheet of steel. Proteans are already the 'superior synthetic' species and have massive QoL built into them.

This is entirely a PR to give **SOME** QoL to FBPs. It's a quality of life to make game mechanics _less unfun_.

SatinIsle:
Just coming in to share my own opinion. Synths getting welding eye protection by default is not going to significantly impact player behavior, in my opinion. I understand that it takes some of the work from roboticists, but only to a similar extent to what is expected to be serviced by medical. I also understand concerns that it sets a precedent that synths might be expected to know how to repair themselves, but I don't believe that's really going to be a thing. At worst it's that they can do what they already have the ability to do, repair minor cosmetic damage and fix dents. On the other hand, they can already repair damaged wires with a wire coil and that's for more involved than what's implied with the welder.

At the end of the day, I see this being a very small buff to synths that will have little effect in terms of player behaviour. It simply makes what they can already do more convenient. I don't really care that much either way, but I think it is just fine (especially with flashing not being protected). I think people are getting a bit too worked up about nothing much at all.

PR #17498, opened 2025-04-08
Reagent Addictions

Willburd:
## About The Pull Request
Addictions were coded downstream, porting up here for other servers to use if desired. All addictions can be configured, and addictions during rounds can be disabled, and are disabled in this pr by default to respect virgo game/roleplay style. More severe drug addiction traits can be commented out or hidden if desired by staff.

This may also see some fun future use in xenoarcheology, with artifacts that make you addicted to bizarre reagents like copper or sugar, but no code for that is included in this PR. Only arbitrary addiction support. This PR's goal is primarily for standard chem addictions and their withdrawls.

## Changelog
Addictions have been added for various chems, from alcohol to hardcore drugs. as well as traits to begin the round addicted to them. Addictions cause no harm to a player unless not fed or treated with inaprovaline during standard withdrawls. Reagents can override the basic handle_addiction proc to give a reagent unique logic for addictions. 

Examples of unique handle_addiction procs include coffee, and the newly added artificial sustenance. Coffee only gives mild headaches and irritation, and artificial sustenance is lethal to drink(unless already addicted) and kills you if you fail to feed its addiction (meant for vat-grown creatures that want to eat a substance to stabilize their body in both roleplay AND gameplay.) 

:cl:
add: Addictions and consequences for not feeding or treating them. 
add: Various reagents are now addictive, servers can decide if reagents can be addicted to during a round, or if only with a trait.
add: Artificial sustenance, a vat-clone body stabilizer using the addiction system, enabled with a trait
add: Traits to be addicted to a reagent from round start
add: inaprovaline treats addiction withdrawls in most cases 
/:cl:


Cameron-The-Raven:
Note:
Hard addictions have been disabled.
Mid-round addictions are disabled

Alcohol and Nicotine have special effects for each.

code/modules/mob/living/carbon/carbon_defines.dm
```diff
@@ -23,6 +23,9 @@
 
 	var/does_not_breathe = 0 //Used for specific mobs that can't take advantage of the species flags (changelings)
 
+	VAR_PROTECTED/list/addictions = list() // contains currently addicted chem reagent IDs
```
ShadowLarkens, line 26: both of these should be lazylists to reduce memory usage

ShadowLarkens (changes requested):

code/modules/mob/living/carbon/carbon_defines.dm
Willburd: Modified to use lazylen, disabled alcohol as a trait addiction

Willburd:
Should be ready, I retested to make sure. Lists null on all addictions being cleared as well.

ShadowLarkens (approved):

PR #17499, opened 2025-04-08
Explosion subsystem

Willburd:
## About The Pull Request
This change moves explosions into a subsystem so that they may be quickly processed over several ticks, reducing lag, and preparing for a future MC that is not tolerant of full lockups. Has been tested for multiple sessions on outpost21 without any issues, we have multiple large explosions regularly through the round due to liberal use of landmines (coded to actually explode as well.)

Explosions now behave as the old "recursive explosions" except now without the recursion and possibility of exploding the MC with a stack overflow. Over a certain size the "recursive" logic stops checking explosion resistance at the center of the explosion, and only toward the edges, as a way to boost the speed of explosions beyond max cap. Something only possible with unique explosions or admemes. There should be minimal player facing changes for servers with phoronics/toxins tank transfer bombs.

## Changelog
Explosions were originally giant MC blocking loops that were not chunked in any way. Resulting in explosions needing to fully resolve before any other processing could happen. While this does not matter much on MOST explosions. Anything larger than max-cap could easily lock the server up for up to 10 minutes.

:cl:Will
refactor: Explosions moved to a subsystem to substantially reduce lag of massive explosions.
/:cl:


Willburd:
This is the best I can offer without redoing the entire powernet regen myself, and that's pointless if the plan is to migrate it to rust entirely. I locked full regen behind larger explosions. This shouldn't be an issue for most blasts that happen on station anyway. 

Includes admin notice that nets are deferred if something bad happens and it can't finish the regen. Calling makepowernets() on machines will manually fix it if something breaks during a round. Hopefully none of this info will be needed.

I'd also advise giving this a full test as well, preferably with multiple eyes looking it over and with any situations I might be unaware of that could break everything horribly.

Willburd:
This is pending merge. Needs tracy profiling.

Cameron-The-Raven:
debugged this shit for 7 hours straight to find multiple server-crashing bugs (Not due to the PR but from what the PR unearthed)

fucking hell

Kashargul:
> debugged this shit for 7 hours straight to find multiple server-crashing bugs (Not due to the PR but from what the PR unearthed)
> 
> fucking hell

Warned you about that and how bad it is

PR #17545, opened 2025-04-16
AllergyPen and autoinjector costs

SatinIsle:
## About The Pull Request

Added a new AllergyPen autoinjector, available via the loadout (1 point) and medical vendors. Basically the same as the inaprovaline injectors that everyone starts with anyway, but with a new sprite and text to theme them around characters with allergies.

Reduced the plastic cost of large autoinjectors in the injector maker from 1500 to 250. The small ones cost 30 already, and the large ones are definitely worth way more but getting more than one per sheet of plastic would be really nice! Hopefully will make the injector maker more used.

## Changelog
:cl:
add: Added a new AllergyPen autoinjector, available via the loadout (1 point) and medical vendors.
balance: Reduced the plastic cost of large autoinjectors in the injector maker from 1500 to 250, and smaller ones from 30 to 25.
/:cl:


SatinIsle:
> I'm sorry, they were _1500_? Wild.
> 
> 30 & 300 are a little awkward, though. Might be worth just making them some evenly-divisible values like 25 & 250, or 40 & 400.

Yeah, fair, will change that later.

PR #17546, opened 2025-04-16
NIF Commlink auto-install

TeleTubby2:
## About The Pull Request
NIF commlink auto-install 

## Changelog
:cl:
add: NIF commlinks auto-initialize if the relevant preference is set in character creation
/:cl:


code/modules/nifsoft/software/14_commlink.dm
```diff
@@ -12,6 +12,8 @@
 /datum/nifsoft/commlink/install()
 	if((. = ..()))
 		nif.comm = new(nif,src)
+		if(nif?.human?.client?.prefs?.communicator_visibility)
```
Kashargul, line 15: why are we null checking the nif? after just having something assingned in the line above?

TeleTubby2: That's a good question with the correct answer being "yeah probably shouldn't bother with that I wrote this in 20 seconds while downing a soder"

PR #17547, opened 2025-04-16
Fix mob holders not being able to be thrown into disposals

TeleTubby2:
## About The Pull Request
Mob holders now play nice with disposals

CanPass code has been removed wholesale - you can still throw things into both the wall mounted and normal disposal units, the wall ones are a bit more precise. 

On the upside, it fixes the unintended spam when something bounces off the rim.

Potential side effect of this being a balance issue for combat, but if people are throwing eachother into disposals it'll tell an admin anyway.
## Changelog
:cl:
fix: mob holders play nice with disposals again
/:cl:


TeleTubby2:
Integration test hates be 😭

PR #17548, opened 2025-04-16
Fixes material accessories having slowdown applied

### LINKED ISSUE #17537: [GENERAL] rings cause slowdown
### Brief description of the bug

wearing a ring, be it on the hand slot (under gloves or not) or as an accessory attached to your uniform, causes slowdown.
tested with:
- painite ring crafted in game.
- silver ring from loadout.
- plastic ring from loadout.

### What you expected to happen

_No response_

### What actually happened

_No response_

### Reproduction steps

```bash
1. either pick a ring from the ring selection in loadout, or craft one in game.
2. wear it.
3. walk and weep.
```

### Code Revision

Server revision: B:release D: Commit: 337ed1454605879e4ba3880519936579bc982c9d

### Anything else you may wish to add

_No response_

TeleTubby2:
## About The Pull Request
fixes #17537

The ol' ball and chain may still weigh you down spiritually, but the ring won't.

## Changelog
:cl:
fix: Rings and Bracelets won't slow you down anymore
/:cl:


code/modules/clothing/accessories/rings.dm
```diff
@@ -136,6 +137,8 @@
 /obj/item/clothing/accessory/ring/material
 	icon = 'icons/inventory/hands/item.dmi'
 	icon_state = "material"
+	material_slowdown_multiplier = 0 //it's a ring, it's never gonna be heavy enough to matter
+	slowdown = 0
```
Kashargul, line 141: isn't that inherited?

TeleTubby2: Yep, the default is 0.5. No idea why the slowdown itself is a recent thing, but it is because of the material multiplier. Slowdown is defined explicitly so if it changes up the chain the ring/bracelet will stay as it is.

PR #17550, opened 2025-04-17
Removing the last sleeps in Init

Cameron-The-Raven:
## About The Pull Request
Makes protean and promethean blobs handle their init better to prevent sleeping in init.

Changes POI landmark generation to a subsystem.


code/game/objects/effects/landmarks_poi_vr.dm
```diff
@@ -12,7 +12,7 @@ var/global/list/global_used_pois = list()
 INITIALIZE_IMMEDIATE(/obj/effect/landmark/poi_loader)
 
 /obj/effect/landmark/poi_loader/Initialize(mapload)
-	src.load_poi()
+	SSpoints_of_interest.poi_queue += src
 	return ..()
 
 /obj/effect/landmark/poi_loader/proc/get_turfs_to_clean()
```
Kashargul, line 18: those procs really should be moved to the subsystem and the effect should be deleted... we just query the location and init a POI there.

code/controllers/subsystems/pois.dm
```diff
@@ -1,3 +1,4 @@
+var/global/list/global_used_pois = list()
```
Kashargul, line 1: please not GLOBAL_LIST_EMPTY(global_used_pois )

Cameron-The-Raven: fixed

code/controllers/subsystems/pois.dm
```diff
@@ -82,4 +82,4 @@ SUBSYSTEM_DEF(points_of_interest)
 	annihilate_bounds(poi_to_load)
 	// Actually load it
 	template_to_use.load(T)
-	qdel(src)
+	qdel(poi_to_load)
```
Kashargul, line 85: no...

PR #17558, opened 2025-04-19
Projectile fix

Cameron-The-Raven:
## About The Pull Request
Fixes a bug where turrets, claymores, mines, etc would miss 100% of the time
Projectile changes also fixed a bug where mines had a 100% miss chance if you stepped on them...Only, it made it have a 0% dodge chance, making you 20 projectiles to the fact multiple times. (death). This PR fixes that and gives the miss chance back, without leaving it at 100%
## Changelog
:cl: Diana
fix: turrets no longer have 100% baymiss chance
fix: Fragmines will no longer do 400 damage when you step on them.
/:cl:

### DIFF
--- a/code/modules/mob/mob_helpers.dm
+++ b/code/modules/mob/mob_helpers.dm
@@ -190,8 +190,11 @@
 	if(istype(attacker, living_attacker))
 		living_attacker = attacker
 
-	else if(non_living_always_hits)
+	else if(non_living_always_hits) //Warning: This will make things like frag mines ANNIHILATE people without evasion.
 		return zone
+	else
+		if(!has_evasion_chance && prob(miss_chance_mod)) //Only take miss chance into account when we have no evasion IF the attacker is non-living (turret/mine/claymore).
+			return null //They missed.
 
 
 	/// Toggle for servers that desire to have players able to miss each other.
--- a/code/modules/mob/mob_helpers.dm
+++ b/code/modules/mob/mob_helpers.dm
@@ -137,7 +137,7 @@
 /// always_hit: Miss chance is not calculateed. Always hit the zone it's targeting. (Default: FALSE)
 /// user_misses: If Player Characters are subjected to RNG hitchance on other Player Characters (Default: FALSE)
 /// mob_misses: If mobs are subjected to RNG hitchance on Player Characters (Default: TRUE)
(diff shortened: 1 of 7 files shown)


code/modules/mob/mob_helpers.dm
```diff
@@ -182,20 +182,35 @@
 	if(!target.client) //If the target is an NPC, we will always hit (barring extreme circumstances like mobs having modified evasion, handled above). Removes baymiss against mobs.
 		return zone
 
+
+	var/mob/living/living_attacker
```
Kashargul, line 186: local scope

code/modules/mob/mob_helpers.dm
```diff
@@ -182,20 +182,35 @@
...
+	/// Toggle if you desire to have it so things like claymores, mines, and turrets ALWAYS will hit their selected zone & are not subject to RNG miss chance.
+	/// By default, this is set to FALSE. If toggled to TRUE, non living entities will ALWAYS hit 100% of the time.
+	var/non_living_always_hits = FALSE
+	if(istype(attacker, living_attacker))
```
Kashargul, line 190: if(isliving()

code/modules/mob/mob_helpers.dm
```diff
@@ -182,20 +182,35 @@
...
+		living_attacker = attacker
```
Kashargul, line 191: var goes here....

code/modules/mob/mob_helpers.dm
```diff
@@ -182,20 +182,35 @@
...
+			return null //They missed.
+
+
 	/// Toggle for servers that desire to have players able to miss each other.
 	/// This is if users are subjected to the same RNG hitchance against other players as mobs are.
 	/// By default, this is set to off, as evasion being calculated is (in my eyes) enough for PvP combat.
 	/// However, if you wish to enable it so there's miss chance, flip this FALSE to TRUE
 	var/user_misses = FALSE
-	if(!user_misses && attacker.client)
+	if(!user_misses && living_attacker && living_attacker.client)
```
Kashargul, line 205: why do those all check for the living and aren't moved into the only living scope?

Kashargul (changes requested):

code/modules/mob/mob_helpers.dm
Cameron-The-Raven: Moved it all around to the living scope

Kashargul (approved):

PR #17604, opened 2025-04-26
Properly paths sound files and folders

Cameron-The-Raven:
## About The Pull Request
Gives most sound files and folders proper pathing. I can't be bothered to find every single one right now, but this should knock out a good amount of them.

> [!WARNING]
> IF YOU USE A WINDOWS OPERATING SYSTEM, THERE IS A RISK THAT YOUR GITHUB WILL NOT LIKE THE PATHS BEING REWRITTEN
> TO FIX THIS, GO TO YOUR GIT COMMAND AND RUN THE FOLLOWING COMMAND:
> git reset --hard && git clean -df
> THIS WILL FIX THE PATH CHANGES


Kashargul:
You better add a proper git hard reset for windows users as command hint here, some people new to git will suffer heavily.

PR #17620, opened 2025-04-27
Recycler hotfix 1

TeleTubby2:
## About The Pull Request
whoopsies. few missing entries for the recycler entries, added 2 layers of sanity check. 

## Changelog
:cl:
/:cl:


code/modules/maint_recycler/code/unit_tests.dm
Kashargul, line 1: path code/unit_tests

TeleTubby2: Less modular than I'd like, but your call 🧎

PR #17649, opened 2025-05-01
puts firefighter into global_lists_vr.dm and moves the icon to the right folder

crabcube:
## About The Pull Request

does what it says on the tin 

My entire test environment dissolved like sand trying to confirm if this even worked or not

two changelogs popped up so i don't know if it'll mess with this too


Kashargul:
You've worked on the master branch once again...

Kashargul:
Eh unfucking that is way too annoying with the workflows etc on... another file I would need to unfuck... closing it. Please redo it from a new branch that isn't master

PR #17650, opened 2025-05-01
Makes Big Leggies not have taur mask

Cameron-The-Raven:
## About The Pull Request
Fixes an error where the taur mask was being applied to Big Leggies

## Changelog

:cl: Arrhythmia_V
fix: Clothing now correctly renders over big leggies
/:cl:


Verkister:
kuva
This was not an error btw.

PR #17654, opened 2025-05-02
Update alienwhitelist.txt

DigitalSquirrel95:
Adds theXDkidoflol to Xenochimera whitelist

Forum post here: https://forum.vore-station.net/viewtopic.php?f=42&t=2486&p=14263#p14263


theXDkidoflol:
why do they call it vorestation when you vore in the station and not vore out the in the station

PR #17658, opened 2025-05-03
Changes knight armour slowdown

SatinIsle:
Changed the knight armours to have a slowdown of 0.5 instead of 0, and the robin version to have a slowdown of 0 rather than -1. On this server these are mostly used just as fluff items for the fantasy areas, and people have been rushing to the robin armour in the fantasy redgate due to its buffs. Rather than removing the armour from the map, making it not a straight up movement buff and letting people still find it is better, as it's not used elsewhere.

## Changelog
:cl:
balance: Changed the knight armours to have a slowdown of 0.5 instead of 0, and the robin version to have a slowdown of 0 rather than -1.
/:cl:


TeleTubby2:
Might be worth tuning it down so a medieval themed armor isn't better than riot gear at stopping bullets too.

SatinIsle:
Cancelling this one, I have a better idea

PR #17661, opened 2025-05-04
Hallucination Refactor

Willburd:
## About The Pull Request
Hallucination code is very old, and full of old memory leaks due to how much less was understood during its creation. As well, most of the code uses sleep and spawn extensively. This is a complete rework of the code to bring it up to modern standards while retaining most of the original features.

## Changelog
A new standardized object for sending images to specific clients safely has been added /image/client_only, motion tracker echos have been updated to use this, (they were a prototype for it.) Using << to send images to clients is unsafe and results in hard-dels. 

Xenochimera hallucination code was entirely copypasted and extended. It has been rolled into hallucination code, and given a subtype. It is still handled slightly differently, but uses the new component system as well. 

All events the hallucination component can cause have been moved to procs for ease of reading and future extension of the system. New effects can be added trivially with a new proc, and an extension to the switch that activates them. This also isolates event logic from the timer logic that controls the hallucination component itself. No more confusion while adding new stuff!

Hallucination attackers have been entirely rebuilt from the ground up and spawn rarely again. Recoded to allow for future additions to be done much easier. Including an example of a fleeing hallucination that is not spawned by standard hallucination code. Some minor features have been lost, such as the attacker hitting you with specific weapons, but these we not spawned at all before, and teleported to move, making them feel very unfinished and old.

:cl:
refactor: complete rebuild of hallucination code to use components and more safely handled client images
/:cl:


code/game/objects/effects/client_only_image.dm
```diff
@@ -0,0 +1,26 @@
+/*
+* Sending images to clients can cause memory leaks if not handled safely.
+* This is a wrapper for handling it safely. Mostly used by self deleting effects.
+*/
+/image/client_only
+	var/list/clients = list()
+
+/image/client_only/New(icon, loc, icon_state, layer, dir)
+	. = ..()
```
Kashargul, line 9: why do we have an empty scope?

Willburd: Removed empty scope

Willburd:
xenochi code was merged and has been properly handled by the new hallucination refactor now.

Cameron-The-Raven (approved):
LGTM

PR #17670, opened 2025-05-06
Fixes some various runtimes

Cameron-The-Raven:
## About The Pull Request
Fixing some various runtimes while examining a live server.
- Makes step triggers not assume things passing it is a mob
- Makes Demons not runtime if qdeleted during their phase. Gets rid of their phase code using Sleep and Spawn and replaces with timers. Additionally makes it respect phase blocked areas.
- Makes Mech Code only send a message or sound to the occupant if there IS an occupant.

These are all things that I saw pop up while on a liver server and just fixed as I saw them.
The demon phase needs testinng.

### DIFF
--- a/code/game/mecha/equipment/mecha_equipment.dm
+++ b/code/game/mecha/equipment/mecha_equipment.dm
@@ -68,38 +68,37 @@
 			if(equip_type == EQUIP_SPECIAL)
 				chassis.special_equipment -= src
 				listclearnulls(chassis.special_equipment)
-			//VOREStation Addition begin: MICROMECHS
 			if(equip_type == EQUIP_MICRO_UTILITY)
 				chassis.micro_utility_equipment -= src
 				listclearnulls(chassis.micro_utility_equipment)
 			if(equip_type == EQUIP_MICRO_WEAPON)
 				chassis.micro_weapon_equipment -= src
 				listclearnulls(chassis.micro_weapon_equipment)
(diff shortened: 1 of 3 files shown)


code/modules/surgery/implant.dm
```diff
@@ -282,7 +282,8 @@
...
 		if (prob(fail_prob))
 			var/obj/item/implant/imp = affected.implants[1]
-			user.visible_message(span_danger(" Something beeps inside [target]'s [affected.name]!"))
-			playsound(imp, 'sound/items/countdown.ogg', 75, 1, -3)
-			spawn(25)
-				imp.activate()
+			if(istype(imp))
+				user.visible_message(span_danger(" Something beeps inside [target]'s [affected.name]!"))
+				playsound(imp, 'sound/items/countdown.ogg', 75, 1, -3)
+				spawn(25)
```
Cameron-The-Raven, line 288: needs to be replaced with a proccall timer

code/game/gamemodes/changeling/powers/armor.dm
```diff
@@ -50,7 +50,8 @@
 
 /obj/item/clothing/suit/space/changeling/dropped(mob/user)
 	..()
-	qdel(src)
+	if(!QDELETED(src))
```
Cameron-The-Raven, line 53: Can be replaced with var that tells it to delete on drop

PR #17671, opened 2025-05-06
Airlock Controller Spawn Removal

Willburd:
## About The Pull Request
The virology access console, and related buttons run an airlock control program that can sometimes fail to properly put doors into a correct state on map init, due to their logic still using spawn(). This PR implements a missing "update" radio command for airlocks that make them report their current status to their airlock controller.

This change only affects the virology access airlock in any meaningful way, as other airlock programs do not care about the initial state of the doors, and will forcibly set themselves to the proper docking or airlock cycle state. Viro access airlocks will instead lockup and become unable to cycle in or out until a borg or AI forces the door to open (thus triggering the same send_status() proc we needed). This fixes that init order dependent lockup.

## Changelog
Removed spawn(10)s from the airlock control program news. The proc signalDoor( X,"update") is now responsible for initial door state being updated. Using the already present "delayed signal" logic used in airlock signal processing. Instead of using a hardcoded 10tick spawn that can be called before the doors get their radio controller, it will now properly await the end of initialization to process the timers it uses. 

:cl:
fix: Virology access buttons no longer depend on init lag to setup their state correctly.
code: Removed spawn() behavior from airlock program. Added airlock update radio command to setup doors on airlock init.
/:cl:

### DIFF
--- a/code/game/machinery/doors/airlock_control.dm
+++ b/code/game/machinery/doors/airlock_control.dm
@@ -73,6 +73,9 @@
 			close()
 			addtimer(CALLBACK(src, PROC_REF(check_completion), TRUE, 0.2 SECONDS), anim_length_before_density + anim_length_before_finalize)
 
+		if("update")
+			check_completion(delayed_status = TRUE)
+
 /obj/machinery/door/airlock/proc/do_secure_open()
 	PRIVATE_PROC(TRUE)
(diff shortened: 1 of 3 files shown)


ShadowLarkens (approved):
Looks good to me, surprised that this update signal was just completely missing.

PR #17673, opened 2025-05-06
Startup Time Optimization

ShadowLarkens:
## About The Pull Request
Reduce tether init time by 37 seconds (on my machine).
Should also impact other maps, to lesser extents.

Set of optimizations to bring init time down from an average (over ten trials) of 90.75 seconds to 53.36 seconds

## Changelog

:cl:
server: Server starts faster now
/:cl:


Cameron-The-Raven (approved):
made my tether almost 20% faster

PR #17674, opened 2025-05-06
Moved firefighter pr to a branch and did , sorrry

crabcube:
## About The Pull Request
Same as last pr, moving firefighter Hud icon into correct folder, and adds the job to the global list


crabcube:
Apologies for the mess this has been, SURELY this will work
if it doesn't, i will combust and turn to ash

PR #17675, opened 2025-05-07
Waddle / Animated Movement component

Cameron-The-Raven:
## About The Pull Request
Adds a waddle component.
Players can toggle this component on at will as any mob/living.
Players can adjust anything from the back-lean, forward-lean, z-height, along with the time of this component. This allows for variation between more 'realistic' movement and more 'comical' movement.
Admins can give this component to anything that can move.
## Changelog
:cl: Diana
add: New Waddle mechanic that can be turned on via 'Character' under Preferences. Allows you to have animated movement.
/:cl:


code/datums/components/traits/waddle.dm
```diff
@@ -0,0 +1,103 @@
...
+	if (!isobj(parent) && !ismob(parent))
+		return COMPONENT_INCOMPATIBLE
+	if(isliving(parent))
+		living_owner = parent
+		add_verb(living_owner, /mob/living/proc/waddle_adjust)
+		//add_verb(living_owner, /mob/living/proc/waddle_debug)
+	our_atom = parent
+	RegisterSignal(our_atom, COMSIG_MOVABLE_MOVED, PROC_REF(process))
+
+/datum/component/waddle_trait/process()
```
ShadowLarkens, line 22: this should probably not be process since it's not actually being processed, but rather a proc

Cameron-The-Raven: ye, I had thought about renaming all the traits (all three of the trait PRs currently use process) from using process() to another proc name...I think I'll just rename them to 'comp_activation' for simplicity's sake.

PR #17680, opened 2025-05-07
Mobvore fix

Cameron-The-Raven:
## About The Pull Request
Fixes a bug where mobs wouldn't eat downed people

## Changelog

:cl: Diana
fix: Mobs can now properly eat you when you're down!
/:cl:


code/modules/mob/living/simple_mob/simple_mob_vr.dm
```diff
@@ -173,7 +173,7 @@
 		M.visible_message(span_danger("\The [src] attempts to pounce \the [M] but misses!"))
```
Kashargul, line 173: I think we should add a 2 second stun to prevent initial failures if you had a key pressed.

Cameron-The-Raven: Added the stun

PR #17720, opened 2025-05-17
Faltered & Dormant diseases update

TheCaramelion:
## About The Pull Request

A small one, but with fancy stuff.
Allows virologist to falter diseases, rendering them non-spreadable (unless it's with an injection).
Dormant diseases are diseases present in crew members; These diseases are **not** spreadable, their only purpose is for a virologist to poke at them.

Some more fancy chems to make Virologists and Chemists suffer alike.

## Changelog
:cl: Guti
add: Added the ability for virologists to falter their diseases
add: Added dormant diseases, that some crewmembers might carry.
add: Added three new chemicals for higher level symptoms.
add: Old syringes and old syringe boxes, which may include include diseases.
balance: Buffed the chance of mice having diseases on them.

/:cl:


Cameron-The-Raven:
Would be best to have a trait to enable spawning with a dormant disease over random spawn.

TheCaramelion:
> Would be best to have a trait to enable spawning with a dormant disease over random spawn.

Adding it as a trait would mean it will either go into the depths of the many trait list we've got currently or a must pick choice for virologists to get strong symptoms roundstart, which would end up losing the gist of it. Having it being randomly given to people once they join the shift would mean that Virologists could go and interact with late joiners, or poke at more people across the station, which would lead to more interaction and maybe some scene opportunities; For example, locking up someone in virology to "quarantine" them since they've got a virus (just not telling them the dormant part...).

As they are currently, for a large portion of the playerbase, they won't be affected at all- As they don't affect their health, diseases don't get processed, and the symptoms won't activate.

I could see about additionally adding a trait to give people a round-start disease, but not the strong symptoms that these give.

PR #17724, opened 2025-05-18
Global chem blacklist + SM addition to blacklist

Cameron-The-Raven:
## About The Pull Request
Moves the chem blacklist for obtainable sources (Xenobotany & Xenoarcheology) to a global list instead of the same list in 4 different spots.
Adds REAGENT_ID_SUPERMATTER to the list of blacklisted reagents...It currently has no use in game other than instakilling people. (If there was a reason to obtain it such as a recipe, a reaction, or ANYTHING then it could make sense to be in there, but ATM it has 0 use other than "instakill everyone" which...Yeah.)
## Changelog
:cl: Diana
del: Supermatter is no longer obtainable via xenoarcheology & xenobotany
/:cl:

### DIFF
--- a/code/_helpers/global_lists.dm
+++ b/code/_helpers/global_lists.dm
@@ -414,7 +414,7 @@ var/global/list/item_digestion_blacklist = list(
 		/obj/item/rig/protean)
 
 ///A list of chemicals that are banned from being obtainable through means that generate chemicals. These chemicals are either lame, annoying, pref-breaking, or OP (This list does NOT include reactions)
-var/global/list/obtainable_chemical_blacklist = list(
+GLOBAL_LIST_INIT(obtainable_chemical_blacklist, list(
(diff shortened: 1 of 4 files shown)


code/_helpers/global_lists.dm
```diff
@@ -413,6 +413,17 @@ var/global/list/item_digestion_blacklist = list(
 		/obj/item/mmi/digital/robot,
 		/obj/item/rig/protean)
 
+///A list of chemicals that are banned from being obtainable through means that generate chemicals. These chemicals are either lame, annoying, pref-breaking, or OP (This list does NOT include reactions)
+var/global/list/obtainable_chemical_blacklist = list(
```
Kashargul, line 417: Wtf? new list and not using GLOBAL_LIST_INIT?

Cameron-The-Raven: i copy pasted from above ::(

PR #17731, opened 2025-05-19
Replace render nanomaps workflow with Holomap piggyback

ShadowLarkens:
## Changelog

:cl:
qol: Crew Monitor & Alarm Map UIs now use a derivative of the Holomap instead of a weird custom mix of handmade icons.
/:cl:


Kashargul:
Don't think we should remove it. We rather should fix the area color markings on SD.

PR #17732, opened 2025-05-20
Dna, Bodyrecord, Xenochi Revive Refactor

Willburd:
## About The Pull Request
This ones a doozy! Original I only set out to rework xenochimera respawns to no longer use ahealing. This expanded considerably as I got more and more frustrated looking at how much copypasta was in dna code and all of resleeving. Every machine that spawns a human mob plays the same game and dance, but in a different dumb way and order. 

This unifies it all into the body record's responsibility. The machines just ask the record to give them an acceptable body. Then do any remaining things. This also includes the xenochimera component storing an internal body record and using it to respawn.

Future plans include changeling migrating to this system instead of a different record format, and promethean reform also using this, once they gain a species component for their abilities. This greatly simplifies how human mobs need to respawn from a body record. As the spawner only needs to supply a location, the body record, and if it can produce robotic limbs.

## Changelog

Added a signal 'COMSIG_HUMAN_DNA_FINALIZED', it is called when a human mob is spawned and has fully completed all dna related activities. This could not just be shoved into UpdateAppearance() as it needs to be ONLY on the creation of a mob. Not just updating its UI/SE/icons.

Moved dna related logic in mob/UpdateAppearance() into dna/ApplyToMob() this proc directly follows dna/ResetUIFrom() in the file, and makes it much easier to compare and make sure that you have assigned all dna variables to and from the mob correctly. Having these split into two separate locations made that much more difficult.

Move hair gradient style and color to use UIs instead of raw vars. There's no player facing changes here. Just data side this matches the rest of the vars doing the same kind of things.

Fully refactored how body_records, dna2/records and dna datums perform their copy() and Clone() procs. Due to near constant neglect. These copy procs almost always fail to have every var properly applied. On top of the programmer ALSO needing to apply it to the mob and read it from the mob correctly. This has been reworked into a fully automated vars[] clone proc for each datum. To prevent extreme lag, and illegally copying several vars. I have also...

Added the BLACKLISTED_COPY_VARS define. Any code that loops through the vars of an object should use this during copy to ignore several list and vars that should not be copied in most cases. This list was put together from already present examples of this in the code, compiling a list of things that shouldn't be messed with. It has been placed in those same locations, as the new entries should not provide any interference to the original goal of those loops anyway. Though many of these procs are either never called or called only by admin procs.

Fully updated the ancient code of the cloning pod to respect weakrefs from the dnascanner, and use bodyrecords instead of dna2/records. This means that if you are somehow using the old cloning pod setup, it'll properly spawn you without half your vars unset, and none of the ooc notes transferred. I've not touched the mind related code, so that may still be incorrect. This PR is focused on bodyrecords and xenochimera so out of scope for now.

Clonepod and dnascanners no longer use loc = src.loc when releasing occupants inside it. This would ignore movement code, and not update recursive listener components. Meaning you would hear from any radios around the pod if they are broadcasting, regardless of your location or if you are wearing a radio. They have also been moved entirely to using weakrefs to avoid any hardrefs bogging up qdel.

Fully recoded xenochimera revival to use an internal bodyrecord from the mob's creation using the new signal. This record is entirely contained in the xenochi component, and is used to restore the mob to healthy condition on hatching. Xenochimera no longer use admin procs to heal themselves. This logic will be used in the future to also handle changeling transforms, and promethean revivals. Both of those require component reworks, and are future plans. 

Xenochimera are now given a choice between reviving from their current appearance, or using the appearance of their slot. Traits will not be changed, and only appearance will be applied from the slot. This allows a form of slot changing while not breaking character swapping rules, and also does not admin spawn a new character into the round as the name will remain the same as well. FBP xenochimera will regenerate prosthetic limbs, and organic xenochimera will drop their prosthetics before regenerating. prosthetics can be put on without assistance by medical or robotics.

Added missing ooc_note entries to body_record data. Another example of how often these records are not maintained, and often are missing data for long periods of time without any bug reports. These we also partially missing their copy() operations, but that has been resolve now, and into the future by the copy/Clone() rework.

Almost fully refactored human resleeving code when using body records. Records are now directly responsible for forming the body and applying the dna to them. All of the dna related machines have massive copypasted walls of "mostly" the same code. These machines are no longer responsible for the body, meaning bodyrecord spawns should be completely consistent when making a body, regardless of the method used. 

Resolved a very small issue where the old cloning console computer would rotate itself to face east every time it was clicked.

Resolved issue where body redesign UI preview would appear to be extremely small, offset in the wrong direction, and make it difficult to design taurs. 

Fixed an extremely laggy loop in species/copy_vars that was checking variable validity for vars guaranteed to exist, but also resulting in it checking all vars in the base datum type + species datum type + byond builtin vars for every var it was checking.

:cl: Diana, Will, Kash
qol: Prometheans now use an appearance changer instead of 100 specialty verbs!
qol: Cocoon trait can now change more things about you!
qol: TF methods using the apperance changer allow you to load slots.
fix: missing ooc notes added to body record for transfer
fix: clonepods do not break listener code when releasing occupants
fix: clonepod computer does not spin right around baby when interacted with
fix: body redesign UI preview aligned correctly and no longer tiny
fix: species/copy_vars no longer takes centuries validating known good data
qol: anyone adding new dna vars no longer needs to worry about the dna cloning proc.
balance: FBP Xenochimera regenerate their robolimbs back
balance: Organic Xenochimera drop their robolimbs, allowing them to slap them back on post-revival
code: added BLACKLISTED_COPY_VARS macro, for safely iterating or filtering atom.vars 
code: added COMSIG_HUMAN_DNA_FINALIZED signal. Sent when a human mob is fully init with all dna and species vars applied.
code: hair gradients use dnaUIs instead of separate vars
refactor: massive rework to dna, bodyrecords and dna2record datums to handle copying vars, and applying them to their owner mob more gracefully
refactor: all dna related machines that resleeve humans, except the autosleever, use a unified proc to ensure no loss of data.
refactor: Xenochimera no longer use admin procs to heal, and use an internal body record datum.
/:cl:


Screemonster:
Does this keep the ability for xenochimera to rebake their appearance? Because that part is an intended happy accident of using the copied aheal.

Willburd:
> Does this keep the ability for xenochimera to rebake their appearance? Because that part is an intended happy accident of using the copied aheal.

This cannot be done by using bodyrecord. If editing the appearance of a xenochimera during regeneration is desired, the cocoon style appearance designer should be used instead. I was requested to do this by kash,as xenochi should not be abusing admin procs.

Willburd:
> Does this keep the ability for xenochimera to rebake their appearance? Because that part is an intended happy accident of using the copied aheal.

To clarify as well, I'm not strongly against or for it. It's just the thing I was asked to resolve. If staff decision is that the behavior should remain on virgo. I can add either a config or ifdef to retain previous behavior. I'm just not in any place to be making those choices without staff discussion.

Willburd:
All dna machines tested as properly detecting and releasing occupant with weakref changes.

Kashargul:
> Does this keep the ability for xenochimera to rebake their appearance? Because that part is an intended happy accident of using the copied aheal.

Swapping character slots while being logged in is a horrible concept that needs to go. It causes too many issues. If that is desired, it needs an explicit copy appearance from slot function.

Willburd:
> > Does this keep the ability for xenochimera to rebake their appearance? Because that part is an intended happy accident of using the copied aheal.
> 
> Swapping character slots while being logged in is a horrible concept that needs to go. It causes too many issues. If that is desired, it needs an explicit copy appearance from slot function.

I'm considering a copy-slot button in the appearance editor tgui, used in cocoon/voretf and the body design console. it would be a good way to produce sleeves of your characters, or for tfing someone into one of your characters? It would just mean xenochi need to have access to a body design menu while regenerating. Thankfully, both the menu and xenochimera both use bodyrecords for data. So this has become considerably easier to add in the future.

Willburd:
Work is currently done, PR should be as is, but requires extensive testing before I'm comfortable taking this out of draft. Will get on that soon with cameron.

Screemonster:
> > Does this keep the ability for xenochimera to rebake their appearance? Because that part is an intended happy accident of using the copied aheal.
> 
> Swapping character slots while being logged in is a horrible concept that needs to go. It causes too many issues. If that is desired, it needs an explicit copy appearance from slot function.

I don't care what the implementation is, it's retaining the functionality that I'm concerned about.

Willburd:
> > > Does this keep the ability for xenochimera to rebake their appearance? Because that part is an intended happy accident of using the copied aheal.
> > 
> > 
> > Swapping character slots while being logged in is a horrible concept that needs to go. It causes too many issues. If that is desired, it needs an explicit copy appearance from slot function.
> 
> I don't care what the implementation is, it's retaining the functionality that I'm concerned about.

Further discussion with kash is also that TG prefs, which is currently half done will not allow this functionality anyway? The best I can offer is allowing xenochi to have body design access in cocoon and a way to copy your current slot's appearance to the character that way. It seems like prefs and ingame chars are being more isolated?

Cameron-The-Raven:
Made it so xenochi can change form during regeneration

Cameron-The-Raven (approved):
Worked good in testing.

Cameron-The-Raven:
This PR fixes a fuckup of the year. Milestone attached.

Willburd:
> Does this keep the ability for xenochimera to rebake their appearance? Because that part is an intended happy accident of using the copied aheal.

I am replying to this, as this feature has been met with some compromise. You may load appearance from slot, size, markings, base_icon, hair, etc. Not counting traits will be copied to the respawning mob. You now have the choice between using your current appearance, or a cosmetic copy of your current character slot. It will also not change your name. 

Otherwise the ability to change appearance by slot is now possible. A lot of the issues before were solved with vanity_copy being fixed by cameron.

PR #17754, opened 2025-05-27
Fixes Nulls getting into the player list

Cameron-The-Raven:
## About The Pull Request
Fixes nulls getting into the player list
## Changelog
nothin player facing


Kashargul:
Assumption:
The tgui panel load could pause the proc, allowing Logout to trigger before the mob was added to the list.

PR #17773, opened 2025-05-31
Lights tubes respect lightswitch state on insertion

Willburd:
## About The Pull Request
Because lights ignore light switches, sabotaged light tubes will instantly detonate as soon as they are inserted. This makes it so that lights will only be turned on if the light switch of the room is also on when they are.

## Changelog
Lights check if a light switch is on before turning on when a bulb is inserted

:cl:
fix: lights do not ignore light switch state when a bulb is inserted
/:cl:


Willburd:
CI still failing due to ddos

PR #17782, opened 2025-06-02
Spider tweaks and Mob pulling

SatinIsle:
## About The Pull Request

Added a small amount of weaken to spider warning bites to knock people over. 
Added a new variable to mobs that can prevent them from being pulled when alive, currently only applies to the spiders. (Idea is that it could be applied to any mob that we don't want to be dragged around the station).
Fixed an issue where spiders would sometimes try to inject poison where they couldn't and runtime.

## Changelog
:cl:
add: Added a small amount of weaken to spider warning bites to knock people over. 
add: Added a new variable to mobs that can prevent them from being pulled when alive, currently only applies to the spiders. 
fix: Fixed an issue where spiders would sometimes try to inject poison where they couldn't and runtime.
/:cl:

### DIFF
--- a/code/modules/mob/mob_defines.dm
+++ b/code/modules/mob/mob_defines.dm
@@ -120,6 +120,7 @@
 	var/m_intent = I_RUN//Living
 	var/lastKnownIP = null
 	var/obj/buckled = null//Living
+	var/pulled_living = TRUE //Test for if it can be pulled when alive
 
 	var/seer = 0 //for cult//Carbon, probably Human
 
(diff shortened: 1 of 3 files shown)


code/modules/mob/mob_defines.dm
```diff
@@ -120,6 +120,7 @@
 	var/m_intent = I_RUN//Living
 	var/lastKnownIP = null
 	var/obj/buckled = null//Living
+	var/pulled_living = TRUE //Test for if it can be pulled when alive
```
ShadowLarkens: this variable name is kinda unclear, can we make it `no_pull_when_living` and invert it so FALSE is default?

ShadowLarkens (approved):

PR #17787, opened 2025-06-02
New Janicart

Willburd:
## About The Pull Request
Ports the updated ride-able janicart floor washer from outpost21. The original janicart is still just a reskined bed you drive around. This moves it to be an actual vehicle with toggle-able floor scrubbers and a reagent tank.

## Changelog
Changes the janicart into a vehicle like the cargo tug. Add reagent tank, and scrubbers that can be enabled and disabled. Also fixes another issue with stairs and vehicles where you'd run yourself over going down them.

repaths include:
/obj/structure/bed/chair/janicart -> /obj/vehicle/train/engine/janicart
/obj/item/key -> /obj/item/key/janicart

:cl:
fix: cult ritual runes missing wash proc
fix: vehicles no longer run you over going down stairs if you are driving them
refactor: janicart updated to use vehicle code
/:cl:


code/modules/vehicles/janicart.dm
```diff
@@ -0,0 +1,148 @@
...
+	turn_off()	//so engine verbs are correctly set
+	create_reagents(600)
+	update_icon()
+	verbs -= /obj/vehicle/train/verb/unlatch_v // Nothing to unlatch
+
+/obj/vehicle/train/engine/janicart/attackby(obj/item/W as obj, mob/user as mob)
+	if(istype(W, /obj/item/mop))
+		if(reagents.total_volume > 1)
+			reagents.trans_to_obj(W, 2)
+			to_chat(user, "<span class='notice'>You wet [W] in the [callme].</span>")
```
TheCaramelion: span_notice()

Willburd: ty

code/game/gamemodes/cult/ritual.dm
```diff
@@ -608,3 +608,10 @@ GLOBAL_LIST_INIT(rnwords, list("ire","ego","nahlizet","certum","veri","jatkaa","
 				R.word3=GLOB.cultwords["technology"]
 				R.loc = user.loc
 				R.check_icon()
+
+/obj/effect/rune/wash(clean_types)
+	. = ..()
+	if (. || (clean_types & CLEAN_TYPE_RUNES))
+		qdel(src)
+		return TRUE
+	return .
```
Kashargul: no... . = ..() already sets a return type. DO not return .

Willburd: This mirrors code\game\objects\effects\decals\cleanable.dm:29 should I update that too?

Kashargul: yes please. . already sets a default return.

Willburd: Fixed

code/game/objects/structures/janicart.dm
```diff
@@ -370,7 +370,7 @@ GLOBAL_LIST_BOILERPLATE(all_janitorial_carts, /obj/structure/janitorialcart)
 
 
 
-
+/* farewell sweet prince, replaced by code\modules\vehicles\janicart.dm
```
ShadowLarkens: can we murder this entirely instead of commenting it out?

code/modules/mob/living/living_movement.dm
```diff
@@ -229,8 +229,10 @@ default behaviour is:
 		MB.runOver(src)
 
 	if(istype(AM, /obj/vehicle))
-		var/obj/vehicle/V = AM
-		V.RunOver(src)
+		if(!istype(buckled, /obj/vehicle)) // Don't run ourselves over, needed for going down stairs in vehicles!
```
ShadowLarkens, line 232: should this maybe be `if(buckled != AM)` so that you can.. run over people on vehicles with another vehicle however unlikely that is?

Willburd: This behavior is intentional, as pulling your own carts can end up with them landing on you as you pull causing harm.

code/game/objects/structures/janicart.dm
Willburd: Farewell sweet prince, cast to the abyss.

PR #17796, opened 2025-06-04
Forensics Datum

Willburd:
## About The Pull Request
The forensics variables have been merged into a singular datum, avoiding more variable bloat on /atom. As this datum is only added to objects if anything transfers forensics information to it. Hopefully a nice handful of mbs reclaimed while also collecting everything forensic into a single place.

## Changelog
moved fingerprints, fingerprintshidden, suit_fibres, blood_DNA, fingerprintslast, and gunshot_residue into a single datum named forensics_crime. This is accessed with the forensic_data var on an atom, usually with a nullcheck. init_forensic_data() is used if the action being performed will create the forensics data.

This also removes a copy of the b_type var on humans, human mobs are never spawned without a dna datum, so they should always be using their dna datum's b_type anyway. This var is confusing and leads to constant bodyrecord and spawning issues with both becoming out of sync. This has been resolved by removing it and falling back to a default blood type if dna somehow doesn't exist.... if you don't have dna as a human you have bigger problems though.

:cl:
code: removed duplicate b_type var from human mobs, they use their dna one only now
fix: bloodtype not being assigned to organs and limbs
add: Fingerprints can now be washed off in washing machines and showers.
refactor: Reworked all bloodsplatter, fingerprint, gunshot residue, and clothing fiber code into a single datum
/:cl:

### DIFF
--- a/code/controllers/subsystems/lobby_monitor.dm
+++ b/code/controllers/subsystems/lobby_monitor.dm
@@ -21,7 +21,7 @@ SUBSYSTEM_DEF(lobby_monitor)
 			continue
 
 		log_tgui(player, "Reinitialized [player.client.ckey]'s lobby window: [ui ? "ui" : "no ui"], status: [player.lobby_window?.status].", "lobby_monitor/Fire")
-		INVOKE_ASYNC(player, TYPE_PROC_REF(/mob/new_player, initialize_lobby_screen))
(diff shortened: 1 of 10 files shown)


Willburd:
Further testing before merge by anyone more knowledgeable in the detective role would be helpful. DNA forensics appears consistent, and all the basic tools worked.

PR #17809, opened 2025-06-06
Improved AI Vore

ShadowLarkens:
## About The Pull Request

- AI bellies are moved into their holograms when formed
- Moved to their core when hologram dissipates for any reason
- AIs can holonom anyone next to their hologram into their selected vore belly

This really needs a lot of testing before I'm confident in it

## Changelog

:cl:
add: Improved AI holonoms, they use real bellies now
/:cl:


RandoGuy999:
It is possible to have someone (like a borg) eat the AI core and then have the AI hologram eat that someone. Nothing immediately breaks but some unintended situations can occur:
- If the someone expels the AI core, then you end up inside your own belly.
- If the hologram disappears for any reason (like turning off the hologram emitter), everyone inside the AI (including the AI) dies and their bodies gets deleted. (except borgs who leave an MMI on the floor.)

Best solution I see is to implement a check inside holo_nom making sure you aren't eating someone who's eaten you.

PR #17828, opened 2025-06-08
Add output prediction to cooking machines

ShadowLarkens:
## About The Pull Request

This makes it so the cooking machines will attempt to predict what they're going to output and display it in the UI. As usual with the cooking system, this prediction is best-effort and may not always reflect what will actually come out, because you can't know for sure unless you actually do the cooking operations.

## Changelog

:cl:
qol: The cooking machines like the grill and deep fryer now attempt to predict what they will produce.
/:cl:


ShadowLarkens:
I'm gonna do more work on kitchen machines in this PR

PR #17835, opened 2025-06-10
Add a very simple bluespace RPED

ShadowLarkens:
## About The Pull Request

Same stats as the prototype brped, but works at range. (read: in view() of mob)

Costs the same as the PBRPED, but requires 1 level of anomalous/precursor tech. Open to changing the balance of all of these, I have no idea how research should be balanced.

## Changelog

:cl:
add: Real Bluespace RPED that works at range. Locked behind prototype brped requirements + anomalous 1.
/:cl:


Cameron-The-Raven:
Precursor requires people to go and do xenoarch, so this seems good to me.

PR #17844, opened 2025-06-11
Primitive Colorgrading

Willburd:
## About The Pull Request
Adds support for primitive color grading to areas. Though right now it's just applying slight tints to client color when you enter an area or update your client color in it. Includes some preset example tints. No areas or maps modified for it, but the var only needs a color set to it (or one of the preset macros). modifiers that affect client color take full priority over this and disabled the  effect entirely(colorblindness for example.) 

I'm mostly laying groundwork for any future improvements, but i doubt we'll ever get shaders to do efficient LUTs. Also includes weather being able to apply its own tint if you are on an outdoor turf on a planet. Otherwise weather falls back to area, and if no area blending is set, it will use no blending at all.

## Changelog
Adds the ability to tint an area's color slightly for whatever artistic reason you would want.
Adds the ability for weather to apply color tints as well.

:cl:
code: added primitive color grading for areas and weather
/:cl:


TeleTubby2:
Finally, we can have a mexico redgate.

Willburd:
> Finally, we can have a mexico redgate.

Unironically this is the joke that made me code it.

PR #17858, opened 2025-06-14
Forensics Merge Data Fix

Willburd:
## About The Pull Request
Forensics data will fail to merge if the current item has no data, but is merging with an item that does. The reverse of this is already handled correctly.

## Changelog
Made forensics data merge properly in both conditions

:cl:
fix: Forensics data merging on objects without data but with data on the merging object, no longer crash on merge
/:cl:

### DIFF
--- a/code/datums/forensics_crime.dm
+++ b/code/datums/forensics_crime.dm
@@ -82,6 +82,8 @@
 /datum/forensics_crime/proc/merge_prints(var/datum/forensics_crime/origin)
 	if(!islist(origin?.fingerprints))
 		return
+	if(!fingerprints)
+		fingerprints = list()
 	if(fingerprints)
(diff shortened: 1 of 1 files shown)


code/datums/forensics_crime.dm
```diff
@@ -201,6 +205,8 @@
 /datum/forensics_crime/proc/merge_fibres(var/datum/forensics_crime/origin)
 	if(!islist(origin?.suit_fibres))
 		return
+	if(!suit_fibres)
```
Kashargul, line 208: LAZYOR(L, I) if(!L) { L = list(); } L |= I;



Goes for all those changes.

Willburd: Resolved in this PR https://github.com/VOREStation/VOREStation/pull/17880

PR #17868, opened 2025-06-15
Unglogs Soap and makes it work similar to how it used to

ReoDaProtovali:
## About The Pull Request

Soap got hit pretty hard recently. Even if it's functional, it was functionally nerfed into the ground. It both was given a usetime and a limited number of uses. Yikes. This PR removes the durability on soap, but retains the usetime on them, with the exception of being able to wet soap in order to instantly clean floors (requested feature). When not wet, it scrubs as normal. Wetness has no effect on cleaning non-turf atoms.
## Changelog
:cl: Reo
qol: soap can be wetted to instantly clean tiles as it used to work.
balance: Balances soap so that it's not terrible again. (removed the durability mechanic)
/:cl:


TheCaramelion:
It would be nice to not lose the soap name/descriptions

ReoDaProtovali:
It replaces the name with enhanced descriptions, reduces soap-to-soap clutter, but if it's truly desired I can readd the names

Willburd:
> It replaces the name with enhanced descriptions, reduces soap-to-soap clutter, but if it's truly desired I can readd the names

This would be desired downstream for our cargo vendor. We had the name edit down there prior.

ReoDaProtovali:
Got it. Definitely dont like the name clutter personally, but I'll readd them if they're actually used elsewhere. ...Since Im touching this PR again, I'll also bother to add the chemistry recipe that I forgot to add the first time. ...And also clean up the afterattack() proc.

Willburd:
Adding inhibitor = list(REAGENT_ID_SODIUM = 1) to the water recipe(The actual hydogen+oxygen recipie, not the lye reaction using water) should prevent it from fighting with lye's, adding an excess of sodium will make it always take priority, as inhibitors prevent the a reaction from happening. That or you could just keep the water focused one. Having doubled up reactions to prevent one being weird might make unit-tests unhappy for weird reasons if future reactions are made that conflict with the lye one in a new way.

ReoDaProtovali:
Never knew about inhibitors. I'll do that tomorrow when I wake up

ReoDaProtovali:
also, just going to mention that the water chem recipe is REALLY weird, and a thing not seen in other codebases, which is why the conflict is an issue in the first place

Water is an extremely abundant chemical already offered in chem dispensers as well so it having a recipe like that just seems like something that was done by polaris or something a long time ago without thought

ReoDaProtovali:
open PR purgatory

ShadowLarkens:
Sorry about that, didn't realize you did the inhibitor thing.

PR #17896, opened 2025-06-20
Looting element for trash piles and more

Willburd:
## About The Pull Request
Loot tables were tied to each instance of a trash pile. This turns looting objects into an element that can simply be attached to the object. Each loot table is a subtype, and while they currently are only implemented for trashpiles and lootpiles, this can be expanded for handling mob drops or unique rewards from arcade games. 

Also puts most groups of tables into their own files. Instead of a giant mixed of objects and loot tables like it was before. Theoretically, this could also be used to add monster hunter style carving drop on mobs where each player only gets one drop. Can I mortgage my house on good boy points yet Kash?

## Changelog
Moved loot and trash pile drop tables into an element
Updated trash piles to have similar behavior to loot piles, such as single user digging.
Trashpile drop table altered due to changed format to a list. expect rng changes. Contents are the same.

:cl:
balance: rng of trashpiles altered
code: trash piles and loot piles now use a unified looting element for handling drops and rarity. Not limited to just loot piles.
/:cl:


Kashargul:
@Cameron-The-Raven also this

Willburd:
Retesting

Willburd:
tested

PR #17907, opened 2025-06-24
1664 patch

ItsSelis:
## About The Pull Request

Library patches for byond 1664


Kashargul:
Not sure if we also should update icon forge with the rust update

PR #17909, opened 2025-06-24
simple ghost pod find

Kashargul:
## About The Pull Request
People often complained since mice were added to ghost pods downstream that it's a mess. This does allow for some basic sorting of ghost pods, for now, we merely allow to differ between structures, mice and mobs. But it can be done much more fine grained.
## Changelog
:cl:
qol: moves find ghost pod, vorespawn and some other join options from tgui list to a tgui datum
add: additional ghost spawn options (2 slots), requires multiple /obj/effect/landmark/start/ghost_roles to be mapped in
/:cl:


Kashargul:
and this @Cameron-The-Raven

Cameron-The-Raven:
tested and looks good from what I can see. Only suggestion is that pods you can not use be marked red on the list

PR #17938, opened 2025-06-30
Replace the alt click menu with the RPG Lootpanel

ShadowLarkens:
## About The Pull Request

Port of https://github.com/tgstation/tgstation/pull/82533

Note: Does not automatically refresh by *design*, so you get a consistent list to pick from if you're working with a very active turf.
Second note: This checks that you are in viewers() of the turf, so you can't continue to look at turf contents unless you can physically see it.

## Changelog

:cl:
qol: Replaced the funky alt-click turf examine menu with a separate TGUI window called the Lootpanel
/:cl:


Kashargul:
I know there was some fuckiness of abuse, is this here the case too by not having it refreshed if you keep a list of items forever with you and then use them with telekinesis etc?

Exploitability confirmed.

grafik

code/controllers/subsystems/statpanel.dm
```diff
@@ -153,13 +146,6 @@ SUBSYSTEM_DEF(statpanels)
 	var/description_holders = target.description_holders
 	var/list/examine_update = list()
 
-	if(!target.obj_window)
-		target.obj_window = new(target)
-	if(!target.examine_icon && !target.obj_window.examine_target && target.stat_tab == "Examine")
```
Kashargul, line 158: Tearing out normal examines?

tgui/packages/tgui/interfaces/LootPanel/LootBox.tsx
```diff
@@ -0,0 +1,78 @@
+import { Button, Stack } from 'tgui-core/components';
+import { type BooleanLike } from 'tgui-core/react';
+import { capitalizeFirst } from 'tgui-core/string';
+
+import { useBackend } from '../../backend';
```
Kashargul: tgui/backend

tgui/packages/tgui/interfaces/LootPanel/index.tsx
```diff
@@ -0,0 +1,104 @@
+import { useState } from 'react';
+import { useMemo } from 'react';
+import { Button, Input, Section, Stack } from 'tgui-core/components';
+import { isEscape } from 'tgui-core/keys';
+import { clamp } from 'tgui-core/math';
+import { type BooleanLike } from 'tgui-core/react';
+
+import { useBackend } from '../../backend';
```
Kashargul: tgui/

same the line below

code/controllers/subsystems/statpanel.dm
TeleTubby2: If you're ditching the examine tab, at minimum go through & tweak the TG pref for it

Kashargul: I don't think the normal examine should be removed at all. The turf one replaced yes, normal, no.

grafik


Did some testing, that PR kills the examine icons for the normal examine tab.

Kashargul:
Materials miss their atom color

grafik

Kashargul:
> I know there was some fuckiness of abuse, is this here the case too by not having it refreshed if you keep a list of items forever with you and then use them with telekinesis etc?
> 
> Exploitability confirmed.
> 
> grafik

It's funny but impractical as player. More likely to be an issue with admins. So just to mark it as admibuse if it actually happens.

PR #17955, opened 2025-07-05
Reagent Refinery machines

Willburd:
## About The Pull Request
refinery

Contains machinery for the reagent refinery. Intended for use with the old ores enabled and it's own area with engineering and possibly cargo and chemistry access. This is an upport from outpost 21, and currently lacks a lot of balancing due to much less population and feedback. This PR seeks to upport the machines so that they are not lost downstream. As well as allowing upstream stations to rebalance them in a more sane way. 

This adds:

1. Industrial Reagent Grinder: Allows grinding ores, plants, sheets, and anything containing reagents. It then pumps it out from its output.
2. Filter: Allows you to select a reagent currently inside it to be filtered. Filtered reagents will be diverted to a secondary output on the filter.
3. Sintering Furnace: Allows some reagents to be melted back together into sheets. Does not allow the production of alloys, such as steel or plasteel. So frostoil still has a valid use.
4. Pipe: A passive pumping structure used between machines, transfers reagents.
5. Pump: similar to a pipe, but has inputs for hoses to be attached.
6. Vat: A large open top vat for reagents to be stored and mixed. Beakers can be dumped in like sinks, or scooped from. Can be buckled to as well to use as a hot tub.
7. Reactor: Can attach to an gas network using a connection pad. Will use the gas in that network to provide a specific atmosphere for distillation reactions to take place.
8. Waste Disposal: Periodically burns any reagents dumped into it. Used as a quick way to destroy unwanted reagents.
9. Hub: Allows a cargo train tanker to park on the pad, and either be drained or filled depending on the direction of the machine it is attached to.

This also adds:
1. Cargo Train Tankers: A large tank designed for transport huge quantities of reagents. This is used to sell reagents on the cargo shuttle. 
2. Smart centrifuge: A simple machine that can take large amounts of reagents, and transfer them into either bottles of chemical canisters
3. Chemical canisters. Large tanks designed to refill chemical dispensers. Cannot be refilled, only drained.
4. Pump Relay: A simple machine with two pump inputs and an output, for use with reagent hoses. Meant to allow fluid pumps to connect over long distances to the refinery itself, or to a pump and hub for easy collection.

## Changelog
Adds reagent refinery machines, circuit boards, designs etc.
Extends distillation reactions to require specific atmospheric conditions to react. Such as rejecting phoron or oxygen, or being performed under nitrogen. Original reactions have not been altered, and may be done under any atmosphere.

:cl: Will
add: Reagent refinery machines
refactor: Distillation reactions are now required to be done under specific atmospheres. The original distillation reactions can be done under normal atmosphere.
code: Extended unit-test for reagents to respect new distillation
/:cl:


Willburd:
As far as I can see this is ready for merging whenever its reviewed. Should not be any more additions till after merge.

PR #17990, opened 2025-07-10
fix some spare pixels on the veymed chassis

TankTheBirb:
## About The Pull Request

my apologies for the like 2 people who use veymed female chassis and are entirely naked, enjoying their giant butt. unfortunately it clips outside of every pant based clothing item in the game so I have fixed it. You will notice because two flesh colored pixels will no longer clip out of nearly every piece of clothing in the game.
## Changelog
:cl:
fix: removed 2 pixels on the veymed female torso sprite to make it actually fit into any clothes at all that aren't skirts.
/:cl:

### DIFF
--- a/html/changelogs/AutoChangeLog-pr-17989.yml
+++ b/html/changelogs/AutoChangeLog-pr-17989.yml
@@ -1,0 +1,4 @@
+author: "Kashargul"
+delete-after: True
+changes:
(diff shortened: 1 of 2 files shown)


KillianKirilenko:
can't believe tank took the synth's butts in the divorce

PR #18012, opened 2025-07-16
Admin Duckie Pack 3

SatinIsle:
## About The Pull Request

Adds three more admin duckies. A gold, viking and galaxy variant. Purely for admin abuse.

## Changelog
:cl:
add: Adds three more admin duckies.
/:cl:

### DIFF
--- a/code/game/objects/structures/watercloset.dm
+++ b/code/game/objects/structures/watercloset.dm
@@ -314,7 +314,7 @@
 	honk_sound = 'sound/voice/quack.ogg' //VOREStation edit
 	var/honk_text = 0
 
-/obj/item/bikehorn/rubberducky/attack_self(mob/user as mob)
+/obj/item/bikehorn/rubberducky/attack_self(mob/user)
 	if(spam_flag == 0)
 		spam_flag = 1
 		playsound(src, honk_sound, 50, 1)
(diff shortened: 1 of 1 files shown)


code/game/objects/structures/watercloset.dm
```diff
@@ -492,6 +494,73 @@
...
 
+/obj/item/bikehorn/rubberducky/gold
+	name = "rubber ducky"
+	desc = "You could give your very life for this duck."
+	icon = 'icons/obj/watercloset.dmi'
+	icon_state = "rubberducky_gold"
+	item_state = "rubberducky_gold"
+	honk_sound = 'sound/voice/quack_reverb.ogg'
+
+/obj/item/bikehorn/rubberducky/gold/attack_self(mob/user as mob)
```
Kashargul: just mob/user

code/game/objects/structures/watercloset.dm
```diff
@@ -492,6 +494,73 @@
...
+/obj/item/bikehorn/rubberducky/viking
+	name = "rubber ducky"
+	desc = "Honking is a duckie exclusive power."
+	icon = 'icons/obj/watercloset.dmi'
+	icon_state = "rubberducky_viking"
+	item_state = "rubberducky_viking"
+	honk_sound = 'sound/voice/scream_jelly_m1.ogg'
+	honk_text = "DUK ROH DAH!"
+
+/obj/item/bikehorn/rubberducky/viking/attack_self(mob/user as mob)
```
Kashargul: ^ likely following as well

SatinIsle: Does this apply to all of the rubber duckies? Because I can just change them all at once.

Kashargul: ves, mob/user is explicitly a mob 

var/something as mob implies that it is one. We shouldn't use that for single types anymore and even for multiple types the parent type might be better in use.

SatinIsle: Done!

PR #18036, opened 2025-07-22
simple mob adnmin spawn

Kashargul:
## About The Pull Request
After making sure that init_vore is called instantly, we should recall it just to be safe that that admin spawned simple mobs get their verbs.

This works as mobs with bellies loaded the verbs already before the player has logged in. Passive mobs without bellies will only call the verb adding in the re init.

## Changelog

:cl:
fix: admin spawned simple mobs not getting their vore verbs if they weren't vore mobs
/:cl:


Kashargul:
Draft, maint preds and such should be tested.

PR #18046, opened 2025-07-23
Feature/circuity overhaul

NovaDusklight:
## About The Pull Request

Some major improvements, and fixed to circuitry, as well as minor additions.
Reagent consuming circuits can now use the reagents in reagent storage circuitry, as was likely intended.
EPV2 circuit can send messages to communicators more intuitively.
Toggle circuit was added, allowing circuits to be disabled if it is toggled off.
Improved the afterattack() for circuitry assemblies.
Added reference grabber circuit, able to grab references of things within sight and 7 range.
Fixed text constraint issue on Text Pad when using TGUI.
Added circuitry bug assembly. This assembly appears beneath some sprites, like tables. It still appears when the tile is inspected, so it is weaker than the spy bug. Also has very low complexity for balancing reasons.
Added an electronic earpiece exclusive circuit that cannot be removed. Allows the wearer to hear messages, but not anyone sharing the tile.

## Changelog

:cl: Zizzi
add: Added a new circuitry assembly.
add: Added two new circuits
add: Added unique speaker to electronic earpiece.
fix: Fixed reagent using circuitry having no reagent input.
/:cl:

### DIFF
--- a/code/controllers/subsystems/circuits.dm
+++ b/code/controllers/subsystems/circuits.dm
@@ -84,7 +84,8 @@ SUBSYSTEM_DEF(circuit)
 		/obj/item/clothing/shoes/circuitry,
 		/obj/item/clothing/head/circuitry,
 		/obj/item/clothing/ears/circuitry,
-		/obj/item/clothing/suit/circuitry
+		/obj/item/clothing/suit/circuitry,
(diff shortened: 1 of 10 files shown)


TeleTubby2:
I kneel

PR #18048, opened 2025-07-23
Pathfinding circuitry and ID lockable assemblies

NovaDusklight:
## About The Pull Request

This pull request adds two new pathfinding circuits. Now with the ability to avoid obstacles. Rebalanced complexity so that they can still fit within the circuitry droids.
An ID circuit was added to allow an assembly to open airlocks.
Circuits can now "bump" airlocks.
Circuit assemblies can now be locked
dreamseeker_br78mZu6AF

## Changelog

:cl: Zizzi
add: Targeted pathfinding circuit
add: Pathfinding locomotion circuit
add: Ability to ID lock circuit assemblies
add: ID access circuit
/:cl:


NovaDusklight:
Just in case this comes into question. This PR is compatible both with, and without the other circuitry PR I created.

code/modules/integrated_electronics/core/assemblies.dm
```diff
@@ -203,6 +209,8 @@
 
 /obj/item/electronic_assembly/GetAccess()
 	. = list()
+	if(access_card)
+		. |= access_card.access
```
Kashargul: why are we doing this on an empty list? +=

NovaDusklight: > why are we doing this on an empty list? +=

Thanks for the catch, Kash. I looked over my pathfinding, and ID card support for the circuitry.
I was unhappy with the ID/access handling, so I've removed those parts, including the mentioned list from the PR. This is tested after the removal.

PR #18057, opened 2025-07-24
Turf slip component [NEEDS TESTING]

Willburd:
## About The Pull Request
Moves slip processing code from turf into a component applied to a mob that has slipped on a turf

Current needs more extensive testing. I've done some basic tests, but I don't know every situation I should be testing.

## Changelog
Moves slip code to component. Makes slip code easier to maintain, M O D E R N I Z A T I O N.

:cl: Will
code: Moves turf slip code to a component
/:cl:


Cameron-The-Raven (approved):
tested.
lube: works
dirtslip: outdoors and dirty  tiles work
water: works
ice: works

good for merge after ci

PR #18083, opened 2025-07-27
Fix multiple pref issues

Kashargul:
## About The Pull Request
Of notice: To fix those issues, slot swapping is no longer possible while being mid round.

## Changelog
:cl:
fix: tail layering pref breaking the entire character
fix: several issues with accidental pref saves
/:cl:


Kashargul:
For some reason we can't use choiced there currently... Something to look into

PR #18119, opened 2025-08-01
admin tickets

Kashargul:
Needs testing but we might have oversanitized that

:cl:
fix: admin ticket replies not being relayed to the chat
/:cl:


SatinIsle:
When testing this, it no longer allows the player to reply to their ticket via the UI, returns:
"Error: Admin-PM: Non-admin to non-admin PM communication is forbidden."
Other methods of replying from both sides seem to work fine. Note though that both accounts were my own connected on the same PC, if that could potentially cause an issue.

Kashargul:
> When testing this, it no longer allows the player to reply to their ticket via the UI, returns: "Error: Admin-PM: Non-admin to non-admin PM communication is forbidden." Other methods of replying from both sides seem to work fine. Note though that both accounts were my own connected on the same PC, if that could potentially cause an issue.

ye... that's likely the main issue as we got no "ref" unless a ticket is handled by someone. I think the tickets on live never had a handler assigned

PR #18131, opened 2025-08-03
Allergic Reaction Element (And direct_say proc)

Willburd:
## About The Pull Request
Move allergy handling code to an element using a signal so that less processing per human mob is needed per life tick. Also consolidates all the allergic reaction code into a single place where more special stuff can be added in the future. Also upports the outpost 21 allergies and reactions. Gibbing hidden on virgo for obvious reasons.

## Changelog
Moved allergic reaction code to an element and signal
Added salt allergy
Added pollen allergy
Fixed mushroom allergy not working on most mushroom foods
Added salt to food that makes sense to have it.

:cl: Will
add: Pollen allergy
add: Salt allergy
add: coughing allergic reaction
add: sneezing allergic reaction
add: gibbing allergic  reaction (disabled on virgo)
fix: Mushroom allergy reacts to more foods
code: allergic reaction handling code moved to an element
code: added missing SIGNAL_HANDLER macros to other elements
code: split say() into a second nonblocking half, direct_say()
code: telecomms uses timers instead of sleep for transmission delay
/:cl:


Cameron-The-Raven (approved):

Cameron-The-Raven:
Thoroughly tested. Seems fine.

PR #18136, opened 2025-08-03
Vampire Hanners get the right hunger icons

TheToaster98:
## About The Pull Request
Changes it so if you are a hanner AND an obligate bloodsucker, you get the blood droplet alerts instead of the burger ones. Simply because it annoyed me >:/
## Changelog
:cl:Tost
fix: Bloodsucker hanners now get the right hunger alerts.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/life.dm
+++ b/code/modules/mob/living/carbon/human/life.dm
@@ -1549,7 +1549,7 @@
 			fat_alert = /obj/screen/alert/fat/synth
 			hungry_alert = /obj/screen/alert/hungry/synth
 			starving_alert = /obj/screen/alert/starving/synth
-		else if(get_species() == SPECIES_CUSTOM || SPECIES_HANNER)
+		else if(get_species() in list(SPECIES_CUSTOM, SPECIES_HANNER))
 			var/datum/species/custom/C = species
 			if(/datum/trait/neutral/bloodsucker in C.traits)
 				fat_alert = /obj/screen/alert/fat/vampire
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/human/life.dm
```diff
@@ -1549,7 +1549,7 @@
 			fat_alert = /obj/screen/alert/fat/synth
 			hungry_alert = /obj/screen/alert/hungry/synth
 			starving_alert = /obj/screen/alert/starving/synth
-		else if(get_species() == SPECIES_CUSTOM)
+		else if(get_species() == SPECIES_CUSTOM || SPECIES_HANNER)
```
Kashargul: No. Always truthy? wtf

Kashargul (changes requested):

code/modules/mob/living/carbon/human/life.dm
ItsSelis: The current change would most likely always return true as SPECIES_HANNER is a valid string:
get_species() is equal to species_custom OR species_hanner is a valid string

```suggestion
		else if(get_species() in list(SPECIES_CUSTOM, SPECIES_HANNER))
```

TheToaster98: ooooh, I see. I had no idea how get_species works, thanks for the notes.
Responding here to kash too while I am at it, I assumed this would be just "this OR this", I see it isn't at all how that works, woops.

PR #18137, opened 2025-08-03
Makes pain simulation more "realistic"

zeskorion:
## About The Pull Request
Makes pain simulation actually simulate the effects of traumatic shock. This means synthetics that have pain simulation on will be able to be stunned, batoned, etc. 
This will also mean that if a xenochimera has an fbp and pain simulation, they will be able to go feral from pain
## Changelog
:cl:
fix: Pain simulation now simulates traumatic shock
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/shock.dm
+++ b/code/modules/mob/living/carbon/shock.dm
@@ -3,7 +3,7 @@
 
 // proc to find out in how much pain the mob is at the moment
 /mob/living/carbon/proc/updateshock()
-	if (!can_feel_pain())
+	if (!can_feel_pain() && !synth_cosmetic_pain)
 		src.traumatic_shock = 0
 		return 0
 


ShadowLarkens:
I feel like this should be a separate trait like "Shock Simulator" so you aren't required to have all of these downsides if you just want to receive vore pain messages.

zeskorion:
> I feel like this should be a separate trait like "Shock Simulator" so you aren't required to have all of these downsides if you just want to receive vore pain messages.

pain simulation can be toggled on and off at will. Due to how this code works, turning your pain simulation off (which you can do whilst in shock) will clear all traumatic shock on the next life tick. By no means are you inhibited by this being the case.

(that, and allowing for shock is more useful for RP-related reasons)

code/modules/mob/living/carbon/shock.dm
```diff
@@ -3,7 +3,7 @@
 
 // proc to find out in how much pain the mob is at the moment
 /mob/living/carbon/proc/updateshock()
-	if (!can_feel_pain())
+	if (!can_feel_pain() && !synth_cosmetic_pain)
```
Kashargul, line 6: Is can_feel_pain used more widely as of why we aren't doing the synth check in the function itself?

zeskorion: > Is can_feel_pain used more widely as of why we aren't doing the synth check in the function itself?

frankly, I'm not sure.  Likely for future-proofing reasons. This is how this is handled wherever synthetic cosmetic pain is used

zeskorion: update: I misunderstood the question. Can_feel_pain() Is used in plenty of cases. Updating it here fixes what I had wanted to fix/works as desired, but I can move the proc up the chain if you want pain simulation to have greater effects.

zeskorion: [discussed on discord]

PR #18158, opened 2025-08-04
Whitelist Update

SatinIsle:
Approved here: https://forum.vore-station.net/viewtopic.php?p=14333#p14333


ItsSelis:
the whitelist has been added

PR #18172, opened 2025-08-05
Fixes a server-crashing issue from woofs

Cameron-The-Raven:
## About The Pull Request
Makes it so the killswtich is static
Swapped the if()if()if() to a for() loop
Makes it so there's a global ticker that ensures that woofs won't get to levels that will crash thee server
## Changelog
:cl: Diana
fix: Wolves will no longer crash the server
/:cl:

### DIFF
--- a/code/__defines/dcs/signals.dm
+++ b/code/__defines/dcs/signals.dm
@@ -197,6 +197,11 @@
 ///from base of atom/MouseDrop_T: (/atom/from, /mob/user)
 #define COMSIG_MOUSEDROPPED_ONTO "mousedropped_onto"
 
+///from base of atom/MouseDrop_T: do_after(mob/user, delay, atom/target, needhand, progress, incapacitation_flags, ignore_movement, max_distance, exclusive)
+#define COMSIG_DO_AFTER_BEGAN "do_after_began"
+///from base of atom/MouseDrop_T: do_after(mob/user, delay, atom/target, needhand, progress, incapacitation_flags, ignore_movement, max_distance, exclusive)
+#define COMSIG_DO_AFTER_ENDED "do_after_ended"
+
 // /area signals
 
 ///from base of area/Entered(): (atom/movable/M)
--- a/code/_helpers/mobs.dm
+++ b/code/_helpers/mobs.dm
@@ -279,6 +279,8 @@ Proc for attack log creation, because really why not
 	if (progress)
 		progbar = new(user, delay, target)
 
+	SEND_SIGNAL(user, COMSIG_DO_AFTER_BEGAN)
+
 	var/endtime = world.time + delay
 	var/starttime = world.time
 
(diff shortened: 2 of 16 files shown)


code/modules/mob/living/simple_mob/subtypes/vore/softdog.dm
```diff
@@ -265,38 +265,30 @@
 			M.visible_message("The petting was interrupted!!!", runemessage = "The petting was interrupted")
 	return
 
+var/global/wolf_killswitch = FALSE
```
Kashargul: NO.

code/modules/mob/living/simple_mob/subtypes/vore/softdog.dm
```diff
@@ -265,11 +265,11 @@
 			M.visible_message("The petting was interrupted!!!", runemessage = "The petting was interrupted")
 	return
 
+GLOBAL_VAR_INIT(woof_maximum, 0)
```
Kashargul, line 268: Where is the var counted down again.

Cameron-The-Raven: I assumed we wanted them to only duplicate an X amount of times total and then not be able to duplicate ever again (as you could get in a loop of 'kill 1, it duplicates all up to 20, kill 1, it duplicates back up to 20, kill 1, it duplicates back up to 20' and you have the infinite loop issue still, but to a lesser degree.

Kashargul: > I assumed we wanted them to only duplicate an X amount of times total and then not be able to duplicate ever again (as you could get in a loop of 'kill 1, it duplicates all up to 20, kill 1, it duplicates back up to 20, kill 1, it duplicates back up to 20' and you have the infinite loop issue still, but to a lesser degree.

I see no issue resetting it when there are 0 left. so can duplicate gets set to false on 20 then enabled again on 0

code/modules/mob/living/simple_mob/subtypes/vore/softdog.dm
```diff
@@ -275,9 +276,13 @@ GLOBAL_VAR_INIT(woof_maximum, 0)
 	. = ..()
 	var/thismany = (rand(25,500)) / 100
 	resize(thismany, animate = FALSE, uncapped = TRUE, ignore_prefs = TRUE)
+	if(!GLOB.woof_current) //all our previous brothers have gone to valhallah
+		GLOB.woof_maximum = 0 //Let's start duplicating again
+	GLOB.woof_current++
 
 /mob/living/simple_mob/vore/woof/hostile/aweful/death()
 	. = ..()
+	GLOB.woof_current--
```
Kashargul: please call this on destroy, not on death.

Kashargul: No, even worse to have it on both

PR #18178, opened 2025-08-06
Reagent cooled atmo machines

Willburd:
## About The Pull Request
coolant

Upgrades the cooler and heater atmo machines to use reagents as coolant. Reagents now have a coolant factor that stacks with other reagents. Machines can check the cooling factor of their reagents datum and calculate how much better they should operate.

## Changelog
Added cooling factor to reagents
Added wikicomputer and chemical analyzer entries for cooling factor
cooler and heater atmo machines now have a coolant tank filled by reagent hoses
cooler and heater will increase their temperature change based on the cooling factor of the reagents inside them. This stacks with part upgrades. These machines were awful to begin with, so any buff seemed fine.
Heaters now can be installed with micro lasers, which allow them to heat faster!

:cl: Will
add: heater and cooler atmo machines now have hose connectors
add: heater and cooler atmo machines can now use coolants to increase their performance.
add: Coolant tanks can now be ordered from Cargo
code: reagents now have a cooling factor for machine efficiency calculations
qol: Heaters can have micro-lasers installed in them, allowing them to heat faster.
/:cl:


Cameron-The-Raven:
Fuckup of the year reason will be added post-merge
reason: Exploding a coolant tank would set the entire world's temperature to absolute 0.
Blow it up on tether? The world is now absolute 0.

PR #14821, opened 2023-04-28
2023 EnerGen Power Cells

Youtubeboy139:
## About The Pull Request

Adds in significantly better looking power cell sprites than what was previously added in. I never fully liked those cells, they felt cheap in the end, left a bad taste in my mouth. These are based on a combination of TG and Skyrat power cells in terms of color pallet and overall design, however, they are completely original designs for the most part (there are some minor exceptions such as the Lead Acid battery).

Also adds in an alternative random power cell spawner that is specific to power cells which might be useful to map makers further down the line. With this addition there is one new device power cell that has a [600] charge value which makes it only slightly better than the base device cell which has a [480] charge value. It costs the same in base raw materials as the EMP shielded device cell which is meant to give players the choice between better cell capacity or having the cell shielded from EMPs.

**_EDIT:_**
I decided to not be so lazy and actually give an update to the Cell Charger as well other than just barely changing the sprite. It now resembles the modern TG station cell charger with some small differences (such as a different 'fully charged' light animation). Cells placed in the charger will have their own wired overlay sprite, the only thing that was not pulled from TG was the updating cell charge light on the cells themselves as just getting these new overlays to work in Bay Code was a pain. An example video of the Cell Charger icon_update feature is included below.

I also moved the sprite for the 'crap' tier power cell to the device cells adding a new device cell that holds half the charge of a standard device cell (similar to the old 'crap' tier power cell). A new 'crap' tier power cell has been added as well to fill the space made by moving the old sprite over to the device cells. The description of both cells was also changed from the copy-paste that was pulled from TG (they were described as AA batteries despite being the wrong size). The normal 'crap' tier power cell is described as a DD sized battery while the smaller 'crap' tier device cell is described as a D sized battery.

Removed the Secborg powercell that was sitting in the code and moved its sprite to the Backup Power Cell that is carried by synths. The Secborg cell was not used in the code and was actually just a copy-paste that was pulled from TG station.

## How This Contributes To The VOREStation Roleplay Experience

Removes the cheaper looking cell sprites and adds in these newer, fresh looking cell sprites.

**_EDIT:_**
Adds in a fresh new and fancy cell charger that has a custom icon_update system that allows for more dynamic cell charging (no more having to guess what cell is on the charger, a player can now see exactly what cell is charging at a glance)

## Proof of Testing

<details>
<summary>Screenshots/Videos</summary>
 
Power Cells in a Test Session
Screenshot 2023-04-27 203539

Power Cell Chargers with Power Cells in a Test Session
[...]

PR #18193, opened 2025-08-07
New batteries

Cameron-The-Raven:
## About The Pull Request
Ports over cell & device cell sprites. (I have been informed these are from Eris, but I got them from [this PR ](https://github.com/CHOMPStation2/CHOMPStation2/pull/4264))
dreamseeker_2025-08-07_04-37-17

dreamseeker_2025-08-07_04-45-00
dreamseeker_2025-08-07_04-38-13

Adjustments:
/obj/item/cell/device/hyper moved from 600 charge to 4800 charge.
(For reference, it is not a weapon battery, and a normal battery has 480 charge.)

/obj/item/cell/device/weapon/recharge/alien charge bumped from 2400 to 5000  (Also charges 10 charge per more, from 120->130)

obj/item/cell/void charge bumped from 4800 to 10000. Now charges 500 instead of 1200.

Yellow slime extract bumped from 10000 charge to 20000 charge. Self charges 500 as well instead of 25.

Giga cell added to the code. Has 40000 charge. Not yet obtainable. Will be a high-tier techweb unlock.

Adds Omni alien cell. Unobtainable.

Old file updated to be the same icon_states (and have new sprites for the sprites it lacked for the new additions) as the new file, meaning if a revert is wished, all that has to be done is add _old at /obj/item/cell
## Changelog
:cl: Diana 
image: Battery icons able to be swapped to battery variants
balance: Hyper device cell now has 2400 charge, up from 600
balance: Void cell and recharging alien device cell adjusted.
balance: Yellow slime extract now  has 20000 charge. Self charges 500 instead of 25 now. 
/:cl:


Cyantime:
Why is this being timelooped. Eris cells were already ported four years ago in #10824 and widely disliked for being super generic grey sci-fi rectangles that dropped all of the established SS13 imagery for no good reason.
Brand new, unique sprites were made specifically for here in #14821 which maintained the older style with a fresh coat of paint to fit the rest of Virgo's aesthetic. What's the point of trying to homogenize everything? What makes Eris sprites such an irresistible lure when they only work when in their own context of a unified art style which any non-Eris server isn't?

Cameron-The-Raven:
Personally, they look much nicer than the cells we currently have. (And as mentioned in that original PR in #10824 these have a more 'sci-fi' look to them than the current standard-day batteries.)

They're much easier to tell at a glance what their charge is with the large charge indicators (which are helpful for those that are colorblind. Not that that's a big factor, but it is at least an added bonus.)

In terms of aesthetic, yes, there's a lot of sprites that could generally use newer sprites to make them look more in line with the game, especially given we're still running Polaris sprites with minimal changes there, but that costs a lot of money to get an artist to completely revamp all walls/floors/windows/table/etc sprites to get it to all look adequate.

Cyantime:
The current cells look like batteries, you can tell they have terminals, are meant to be put into something, and match with every existing sprite that has a standard power cell-shaped hole in them like APCs or space heaters or whatever. The Eris cells look like nothing. Having a "sci-fi" look isn't a benefit when they look like every single other sci-fi widget that's been ported from some other server.

Once again, there were brand-new sprites specifically made for this server to replace hastily ported Eris sprites because they were disliked. Why are they coming back?

Cameron-The-Raven:
I was upporting some battery tweaks (giga batteries, hyper device cell being made useful) and the battery sprites. Initially, I was thinking of just leaving it be its own file and not enabled here, but after looking at them they struck me as the nicer looking sprites and I decided to enable it over the current ones.

(It should be noted as well, I updated the _old .dmi to have the same naming structure, so it's a one line swap to between sprites)

Cameron-The-Raven:
eh, doing sprite cleanup, going to leave these as old for now

PR #18200, opened 2025-08-07
Allows survival watch users to set its GPS tag name

Tenryumi:
## About The Pull Request

See title. Using the GPS watch inhand allows you to set the name of its tag now! This will no doubt make it more useful as a GPS now that people actually know who the watch belongs to. However it still cannot do anything else a normal GPS can do.

## Changelog

:cl:
qol: Survival watch users can now set the tag of the watch's inbuilt GPS by using it in-hand.
/:cl:

### DIFF
--- a/code/game/objects/items/devices/gps.dm
+++ b/code/game/objects/items/devices/gps.dm
@@ -278,6 +278,15 @@ GLOBAL_LIST_EMPTY(GPS_list)
 	popup.set_content(dat.Join(null))
 	popup.open()
 
+/obj/item/gps/proc/set_tag_prompt(mob/user)
+	var/a = tgui_input_text(user, "Please enter desired tag.", name, gps_tag, 10)
+	a = uppertext(copytext(a, 1, 11))
+	if(in_range(src, user))
+		gps_tag = a
+		name = "global positioning system ([gps_tag])"
+		to_chat(user, "You set your GPS's tag to '[gps_tag]'.")
+		. = TRUE
+
 /obj/item/gps/Topic(var/href, var/list/href_list)
 	if(..())
 		return TRUE
@@ -327,13 +336,7 @@ GLOBAL_LIST_EMPTY(GPS_list)
 				. = TRUE
 
 	if(href_list["tag"])
-		var/a = tgui_input_text(usr, "Please enter desired tag.", name, gps_tag, 10)
-		a = uppertext(copytext(a, 1, 11))
-		if(in_range(src, usr))
-			gps_tag = a
-			name = "global positioning system ([gps_tag])"
-			to_chat(usr, "You set your GPS's tag to '[gps_tag]'.")
-			. = TRUE
+		set_tag_prompt(usr)
 
 	if(href_list["range"])
 		local_mode = !local_mode
--- a/code/modules/clothing/accessories/watches.dm
+++ b/code/modules/clothing/accessories/watches.dm
@@ -71,3 +71,7 @@
 	if(Adjacent(user))
 		gps.tracking = !gps.tracking
 		to_chat(user,span_notice("You turn the micro beacon [gps.tracking ? "on" : "off"]."))
+
+/obj/item/clothing/accessory/watch/survival/attack_self(mob/user)
+	. = ..()
+	gps.set_tag_prompt(user)


code/game/objects/items/devices/gps.dm
```diff
@@ -278,6 +278,15 @@ GLOBAL_LIST_EMPTY(GPS_list)
 	popup.set_content(dat.Join(null))
 	popup.open()
 
+/obj/item/gps/proc/set_tag_prompt(mob/user)
+	var/a = tgui_input_text(user, "Please enter desired tag.", name, gps_tag, 10)
+	a = uppertext(copytext(a, 1, 11))
+	if(in_range(src, user))
+		gps_tag = a
+		name = "global positioning system ([gps_tag])"
+		to_chat(user, "You set your GPS's tag to '[gps_tag]'.")
```
Kashargul: to_chat without sorting span

KillianKirilenko:
FWIW, I didn't add tag-changing to these when I made them because that's kind of meant to be a little bit of the friction, so to speak. They're an emergency option, and one that will (ideally) be used so sparingly that whenever you *are* looking for one, there should be no other conflicting signals to cause confusion.

Kashargul:
> FWIW, I didn't add tag-changing to these when I made them because that's kind of meant to be a little bit of the friction, so to speak. They're an emergency option, and one that will (ideally) be used so sparingly that whenever you _are_ looking for one, there should be no other conflicting signals to cause confusion.

In general there never should be untagged messages anymore as those go to unsorted, a category we want to axe.

Kashargul:
> FWIW, I didn't add tag-changing to these when I made them because that's kind of meant to be a little bit of the friction, so to speak. They're an emergency option, and one that will (ideally) be used so sparingly that whenever you _are_ looking for one, there should be no other conflicting signals to cause confusion.

Hmm... Are the tags unique or is it the same for everyone, so we could 20 times the same tag?

KillianKirilenko:
> > FWIW, I didn't add tag-changing to these when I made them because that's kind of meant to be a little bit of the friction, so to speak. They're an emergency option, and one that will (ideally) be used so sparingly that whenever you _are_ looking for one, there should be no other conflicting signals to cause confusion.
> 
> Hmm... Are the tags unique or is it the same for everyone, so we could 20 times the same tag?

It's always `SRV-WATCH` or something like that, IIRC. You could probably give them a unique tag in `New()` using random number gen or some other method.

Kashargul:
> > > FWIW, I didn't add tag-changing to these when I made them because that's kind of meant to be a little bit of the friction, so to speak. They're an emergency option, and one that will (ideally) be used so sparingly that whenever you _are_ looking for one, there should be no other conflicting signals to cause confusion.
> > 
> > 
> > Hmm... Are the tags unique or is it the same for everyone, so we could 20 times the same tag?
> 
> It's always `SRV-WATCH` or something like that, IIRC. You could probably give them a unique tag in `New()` using random number gen or some other method.

New is illegal... but ye Initialize(mapload) would work for that

KillianKirilenko:
> > It's always `SRV-WATCH` or something like that, IIRC. You could probably give them a unique tag in `New()` using random number gen or some other method.
> 
> New is illegal... but ye Initialize(mapload) would work for that

Excuse my *momentary lapse of memory*, you know what I meant. :p

Tenryumi:
> FWIW, I didn't add tag-changing to these when I made them because that's kind of meant to be a little bit of the friction, so to speak. They're an emergency option, and one that will (ideally) be used so sparingly that whenever you _are_ looking for one, there should be no other conflicting signals to cause confusion.

That's cool! And I understand the thought process; it makes sense.

However, I do not think that GPS devices are difficult enough to acquire to warrant such great limitations being placed on this part of the watch's functionality. I think this would help give the watch just a little bit more of its own niche.
There are publicly available GPS's that do everything that the watch does but better, and with added functionality that it lacks. Additionally, the GPS is already a very tiny item that takes up hardly any space at all. This kinda makes the watch's GPS feature redundant, sadly. Even for pilots, who get these watches in their lockers, in those very same lockers they also get well. A GPS, which does what the watch's GPS does but better. As it stands, there isn't really a reason to use the things at all beyond fluff simply because its superior counterpart is so broadly accessible.
What I'm saying is, this change allows the watch's inbuilt locator to do its job just a bit more conveniently to give it a bit more of a niche. It still can't locate other devices, but with this it will be easier to tell who is who in case something goes wrong for its wearer.

> It's always SRV-WATCH or something like that, IIRC. You could probably give them a unique tag in New() using random number gen or some other method.

I feel like letting people name them is important for them to fill their niche as a last-ditch measure to make sure its wearer is locatable. Regardless of whether they have randomized names or the same names, without an identifiable label in the GPS list, you still won't really be able to tell who the watch belongs to. This change simply makes it easier to identify who is who!

... However all of what I said is pointless if the intent of this item was for it to be useless by comparison, in which case, uuhhhhhhhh... I guess nevermind to all this??

KillianKirilenko:
> ... However all of what I said is pointless if the intent of this item was for it to be useless, in which case, uuhhhhhhhh... I guess nevermind to all this??

It is a *little* bit of a joke item in that it *riffs on/pokes fun at* overdesigned watches that include all sorts of useless 'features', but it *is* still something that can theoretically help save your ass in an unusual situation.

At the end of the day I don't have that much of a dog in this fight, but I did at least want to explain that it's in no way meant to be a stand-in for a regular GPS. Keep in mind, *anyone* can get one of these; they're available in character setup. They're also just kind of an afterthought I added to the watches PR for giggles.

I just figured that, since there probably wouldn't be that many floating around, it was entirely safe to leave them with an unchangeable tag since there shouldn't be so many *active* survival watches around that you can't easily tell who it belongs to from context/situation. Mind you I have no idea how many people use/wear any of the watches these days.

Tenryumi:
Alright! Fair enough then. Didn't know about the intent of the item's creation when I made this; just thought I was adding some QoL and wasn't aware I'd be overriding a specific vision you had for the item.
In that case nevermind to all this!

PR #18201, opened 2025-08-07
Toggle for bloopers (multiple say sounds)

SatinIsle:
## About The Pull Request

Added a preference for "Multiple Say Sounds", which is enabled by default. If you have it enabled it plays multiple say sounds for long sentences as normal. If disabled it will play only one say sound, if say sounds are enabled. It does mean there's an extra bloop at the beginning now, but it doesn't change the effect much in my testing.

A few people prefer the single say sound, myself included, and find the multiple bloopings to be a bit overwhelming, especially when there's a lot of people around. We'd prefer this option to disabling say sounds completely.

## Changelog
:cl:
qol: Added a preference for "Multiple Say Sounds", which is enabled by default. If you have it enabled it plays multiple say sounds for long sentences as normal. If disabled it will play only one say sound, if say sounds are enabled. It does mean there's an extra bloop at the beginning now, but it doesn't change the effect much in my testing.
/:cl:

### DIFF
--- a/code/modules/client/preferences/types/game/sound.dm
+++ b/code/modules/client/preferences/types/game/sound.dm
@@ -63,6 +63,12 @@
 	default_value = TRUE
 	savefile_identifier = PREFERENCE_PLAYER
 
+/datum/preference/toggle/bloop_sounds
+	category = PREFERENCE_CATEGORY_GAME_PREFERENCES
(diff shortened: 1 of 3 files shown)


Cameron-The-Raven:
The first sound respects say/whisper pref, while the sounds afterwords are based solely on Blooper pref...This feels like a fair compromise to me.

PR #18256, opened 2025-08-14
fixes spritesheet runtime on load

Cameron-The-Raven:
## About The Pull Request
Fixes runtime on spritesheet load.

Only found because my CPU is from the 80's and spritesheets take forever for me to make. Didn't happen on newer CPUs

## Changelog

:cl: Kash
code: Chat spritesheet now instantly loads, preventing a runtime on slower CPUs
/:cl:


Kashargul:
Before people wonder, this is fine to generate immediately as we need it in assure ready on the first player connection anyway

PR #18258, opened 2025-08-14
Clothing fallback

Cameron-The-Raven:
## About The Pull Request
Ports /tg/'s icon_exist improvements and **ICON_EXISTS_OR_SCREAM**
Makes clothing have a fallback in the event of a custom species file not having the appropriate icon state.

Example of when this might happen:
New space suit has been created, has icon_state = new_suit
Only one sprite was made for this suit.

Instead of having to copy and paste this new suit sprite into EVERY species .dmi, it instead will check 'Do we have this custom sprite present in our custom species sprite .dmi?'
If no, it defaults back to the original non-species specific icon. 
If yes, it uses the species specific icon like normal

## Changelog
:cl: Diana
code: Suit icons are now smart and will automatically fallback to the original icon if a species does not possess the icon state.
refactor: Makes all uses of X in cached_icon_states(icon) into the much faster RUST based icon_exists() proc
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human_movement.dm
+++ b/code/modules/mob/living/carbon/human/human_movement.dm
@@ -241,14 +241,22 @@
 */
 #undef HUMAN_LOWEST_SLOWDOWN
 
+///Gets whatever jetpack we may have and returns it. Checks back -> rig -> suit storage -> suit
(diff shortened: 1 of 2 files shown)


Cameron-The-Raven:
All clothing sems to pass, hmm

PR #18283, opened 2025-08-17
Alt click reagent containers to set transfer volume

SatinIsle:
## About The Pull Request

Made it so that you can Alt click reagent containers to set transfer volume.
Changed all transfer amounts on reagent containers to be set by number input, between a min and max value, instead of using the lists.

## Changelog
:cl:
qol: Made it so that you can Alt click reagent containers to set transfer volume.
refactor: Changed transfer amounts to be set by number input rather than lists.
/:cl:


code/modules/reagents/reagent_containers/_reagent_containers.dm
```diff
@@ -171,3 +171,11 @@
 		if(B)
 			balloon_alert(user, "\The [W] burns the blood in \the [src].")
 			B.changling_blood_test(reagents)
+
+/obj/item/reagent_containers/AltClick(mob/user)
+	. = ..()
+	if(!Adjacent(user))
+		return
+	var/N = tgui_input_list(user, "Amount per transfer from this:","[src]", possible_transfer_amounts)
```
Kashargul: Please pass the current value as default

SatinIsle: > Please pass the current value as default

I tested this and it doesn't seem to work. I suppose because "possible_transfer_amounts" is a list of values, and setting the default to the current amount_per_transfer_from_this doesn't know what place in the list this is?

SatinIsle: Did a major change to all reagent containers to now use max and min values for number input instead.

PR #18290, opened 2025-08-18
Makes loyalty potions work better on slimes

Cameron-The-Raven:
## About The Pull Request
Makes loyalty potions work better on slimes

Makes it so when the loyalty potion is used on a slime, it updates their mood and makes them dropvore eligible.

Additionally, makes it so when a loyalty potion is used on a slime, when it splits all the slimes it creates will be dropvore eligible, as their faction will be 'neutral' instead of the default "slime"
## Changelog
:cl: Diana
fix: Loyalty potions now allow slime spawn to be eaten easily
fix: Loyalty potion now immediately triggers a slime's mood changing, making them able to be eaten.
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_mob/subtypes/slime/xenobio/xenobio.dm
+++ b/code/modules/mob/living/simple_mob/subtypes/slime/xenobio/xenobio.dm
@@ -160,7 +160,7 @@
 		pacified = FALSE //We are not obedient enough to be considered pacified.
 
 	if(!client) //Only update if we don't have a client.
-		if(faction != "slime") //We have had a loyalty potion used on us.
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/simple_mob/subtypes/slime/xenobio/xenobio.dm
```diff
@@ -160,7 +160,7 @@
 		pacified = FALSE //We are not obedient enough to be considered pacified.
 
 	if(!client) //Only update if we don't have a client.
-		if(faction != initial(faction)) //We have had a loyalty potion used on us.
+		if(faction != "slime") //We have had a loyalty potion used on us.
```
Kashargul: Use the define!

PR #18311, opened 2025-08-20
Mass Part Installation For Frames From Storage Boxes/Bags

Willburd:
## About The Pull Request
Installing large numbers of stock parts on multiple frames is very slow, tedious, and incredibly painful on borgs due to needing to swap gripper around for it. This PR makes it so storage containers can be dumped into machines to install all parts at once. This still requires the machine to be constructed, boarded, secured, and wired however. It does not instantly construct the machine either. It just removes the tedium of manually parting machines.

Also borg qol with a special pouch just for parts instead of one by one part gripping.

## Changelog
Made frames accept bags as input, and scans their inventory for parts to install.
RPED is already a bag, bluespace one can install parts remotely, but still not finish the machine.
Parts pouch can now pickup all stockparts on a turf, and drop all at once
Borgs given a very big special parts pouch to give them some qol with this pr

:cl: Will
add: Frames now accept parts stored in boxes or bags and load them all at once
add: RPED can now install parts in a frame, bluespace RPED can do so at range
add: Engineering borgs given an expanded parts pouch
qol: Engineering parts pouch can now pick up all stock parts on a turf, and drop them all at once, can be used to mass install parts in machines like any other bag
/:cl:

### DIFF
--- a/code/game/machinery/frame.dm
+++ b/code/game/machinery/frame.dm
@@ -631,33 +631,63 @@ GLOBAL_LIST(construction_frame_floor)
 	else if(istype(P, /obj/item))
 		if(state == FRAME_WIRED)
 			if(frame_type.frame_class == FRAME_CLASS_MACHINE)
-				for(var/I in req_components)
-					if(istype(P, I) && (req_components[I] > 0))
-						playsound(src, 'sound/items/Deconstruct.ogg', 50, 1)
-						if(istype(P, /obj/item/stack))
-							var/obj/item/stack/ST = P
-							if(ST.get_amount() > 1)
-								var/camt = min(ST.get_amount(), req_components[I]) // amount of stack to take, idealy amount required, but limited by amount provided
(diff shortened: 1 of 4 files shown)


code/game/machinery/frame.dm
```diff
@@ -631,34 +631,64 @@ GLOBAL_LIST(construction_frame_floor)
...
+		req_components[I]--
+		break
+
+	if(defer_feedback)
+		return installed_part
+
+	if(installed_part)
+		playsound(src, 'sound/items/Deconstruct.ogg', 50, 1)
+		update_desc()
+		to_chat(user, desc)
```
Kashargul, line 675: Are we sure that desc always is sorted under info?

If not, infoplain both times to be sure

Willburd: This matches the original behavior which was my goal. I've changed nothing else except how parts are loaded.

PR #18326, opened 2025-08-23
Belly temperature

SatinIsle:
## About The Pull Request

Added new belly temperature settings to the vore panel.
Vore panel options now include:

- Belly Temperature: [Set to 20C by default] Prey feel the temperature in your belly, will an icon if it's uncomfortably warm/cold. 
Based on belly temperature and not the body temperature math of the prey, so easier to trigger. Does not paste any messages in chat.
- Temperature Damage: [Set off by default] If on, can deal damage to the prey, ONLY if they have digestion turned on.
- Preference for Affected By Temperature: [Set on by default] You will only have the above apply to you at all as prey if this is enabled. You will still only take damage if you have digestion turned on and the pred has damage turned on.

To enable this, calculations in human/life.dm now only apply temperature warnings and damage effects when the characters location is not a belly, this skips these sections of code when the prey is in a belly only. Bellies are otherwise perfect environments, so this does not actually affect anything about gameplay otherwise.

## Changelog
:cl:
add: Belly Temperature: [Set to 20C by default] Prey feel the temperature in your belly, will an icon if it's uncomfortably warm/cold. 
Based on belly temperature and not the body temperature math of the prey, so easier to trigger. Does not paste any messages in chat. Has a range of -100C to +200C. 
add: Temperature Damage: [Set off by default] If on, can deal damage to the prey, ONLY if they have digestion turned on.
add: Preference for Affected By Temperature: [Set on by default] You will only have the above apply to you at all as prey if this is enabled. You will still only take damage if you have digestion turned on and the pred has damage turned on.
refactor: calculations in human/life.dm now only apply temperature warnings and damage effects when the characters location is not a belly, this skips these sections of code when the prey is in a belly only.
fix: Fixes a bug where heat and cold discomfort alerts were cleared instantly outside of bellies.
fix: Fixes a SmartVend UI alphabetical ordering to use modern code.
/:cl:

### DIFF
--- a/tgui/packages/tgui/interfaces/VorePanelExport/VorePanelExportBellyString.tsx
+++ b/tgui/packages/tgui/interfaces/VorePanelExport/VorePanelExportBellyString.tsx
@@ -221,7 +221,7 @@ export const generateBellyString = (belly: Belly, index: number) => {
   let result = '';
   result += `<div class="accordion-item"><h2 class="accordion-header" id="heading${index}">`;
   result += `<button class="accordion-button collapsed" type="button" data-bs-toggle="collapse" data-bs-target="#collapse${index}" aria-expanded="false" aria-controls="collapse${index}">`;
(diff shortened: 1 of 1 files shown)


tgui/packages/tgui/interfaces/VorePanelExport/types.ts
```diff
@@ -52,6 +53,7 @@ export type Belly = {
   storing_nutrition: BooleanLike;
   entrance_logs: BooleanLike;
   item_digest_logs: BooleanLike;
+  temperature_damage: BooleanLike;
```
Kashargul, line 56: Missing implementation in export

PR #18332, opened 2025-08-23
grepchecking

Kashargul:
When people want to use papers, they should just make prefabs and not write html in map files. Reviewing HTML inside a map is a lot more annoying than to do it in code.

🆑 
code: bans htlm inside dmm files
/:cl:


Cameron-The-Raven:
Should be good for merge now.

PR #18334, opened 2025-08-24
Makes uses of do_after sane

Cameron-The-Raven:
## About The Pull Request
Makes all uses of do_after(...) sane.
Makes it so ALL uses of do_after have a target noted (this prevents spamming an action 100s of times in one second)
Does some code quality adjustments and changes things like 40, 30, etc to "4 SECONDS" "3 SECONDS" etc.
Fixes a typo.
~~Clamps do_after so you can't get instant actions (You can have from 25% to 300% action time, depending on your encumberance...Might need to be adjusted/toned down/modified by traits/etc).~~

Encumberance removed.

Standardizes do_after() to follow the format of:
do_after(X, duration, target)
X can be anything. 
duration is X SECOND(S) (if a full second amount) or whatever the number is, or whatever it's special formula is.
Target is either 'target' or 'target = X'

This should hopefully make it so it can be regex'd in the future instead of done by hand like I did...As every one of these were things like do_after(src,duration,dude) do_after(src, duration,dude) do_after(src,duration, dude) etc etc, making it impossible to regex.

This also fixed some completely illegal/broken do_afters that would have never worked

Some things will need to be adjusted to be exempt encumbrance slowdown, but that's for a later PR. 

## Changelog
:cl: Diana
fix: Gets rid of the encumberance factor entirely
fix: do_after can not be spammed on one thing, resulting in loss of materials and such
code: Converted some base numbers to proper SECONDS define
refactor: Refactors do_after to only allow one use per thing.
/:cl:

### DIFF
--- a/code/_helpers/mobs.dm
+++ b/code/_helpers/mobs.dm
@@ -300,13 +300,13 @@ Proc for attack log creation, because really why not
 		drifting = TRUE*/
 
 	var/holding = user.get_active_hand()
-
+/* //Disabling for now. Meant to be used for modifier slowdowns.
 	if(!(timed_action_flags & IGNORE_SLOWDOWNS))
 		var/slowdown = user.calculate_item_encumbrance()
 		if(slowdown)
 			//Let's not be TOO evil. You can be up to 4x faster, but never more than 3x slower.
 			delay *= CLAMP(slowdown, 0.25, 3)
-
+*/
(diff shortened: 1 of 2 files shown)


KillianKirilenko:
This has broken climbing over tables from what I can tell.

Cameron-The-Raven:
Am finally back home and can work on this.

PR #18357, opened 2025-08-27
Neck Grab Pain Fix

SpringSkipper:
### What it does
Removes the occasional pain gasps from neck grabbing.

### More details
Our pain code means that any amount of damage, no matter how small, has a nonzero chance of making you gasp/yell/scream in pain. Previously, neck grabbing (distinct from choking) did 1 oxy damage per tick; an inconsequential amount, since neck grabbing didn't actually make you unable to breathe so you'd just instantly heal it. However, it could cause you to make automatic pain emotes.

This is annoying for many scenes; while a neck grab would realistically be uncomfortable, plenty of people use neck grabs for things other than literally grabbing someone by the neck, like holding them in a very firm hug or pinning them to immobility. Further, in scenes where someone _is_ actually intended to be in pain, I feel like it's better to let the victim act that out through emotes and /me messages rather than the game occasionally forcing them to. At a glance, "yeah a neck grab should be painful" makes sense. But in practice, anyone who does rough unwilling scenes with any sort of frequency can tell you that it's a headache.

That small amount of oxy damage has instead been moved to when you are actively choking someone (a 'kill grab'), so balance-wise the damage remains completely unchanged if you're actually trying to kill someone, or roleplaying out strangling someone. The only balance difference this would make if someone being neck-grabbed also happened to be suffocating from an unrelated cause, in which case they'll now suffocate _slightly_ slower.

## Changelog

:cl:
qol: removed neck grab pain gasps
/:cl:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,3 +50,5 @@ prof.dll
 
 # From /tools/define_sanity/check.py - potential output file that we load onto the user's machine that we don't want to have committed.
 define_sanity_output.txt
+config/alienwhitelist.txt
+/config
--- a/code/modules/mob/mob_grab.dm
+++ b/code/modules/mob/mob_grab.dm
@@ -145,15 +145,15 @@
 
 	if(state >= GRAB_NECK)
 		affecting.Stun(3)
-		if(isliving(affecting))
-			var/mob/living/L = affecting
-			L.adjustOxyLoss(1)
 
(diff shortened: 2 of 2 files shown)


SpringSkipper:
fucked up something with the git ignore file, not sure how to fix it without a whole new PR

PR #18361, opened 2025-08-28
Bunsen Burner

Willburd:
bunsenburner

## About The Pull Request
Ports outpost 21's refactor of the bunsen burner from yawnwider's old phorochem. It now functions as a very basic form of distillation reactor. Allowing for distilling reactions from the turf's current temp up to 600c as well as using its current atmosphere. Also fixes some other small things. 

This shouldn't change much for distilling reactions, as more complex reactions require more complex atmospheres, or temperatures not easily obtainable with the bunsen burner. Maybe players will finally make the patch chems if this is mapped into chemistry! A pipe dream I know~

> [!NOTE]
> These should probably get mapped into chemistry, along with a chemical analyzer? Be sure to anchor it!

## Changelog
Adds bunsen burners, allowing easier basic distilling
Bunsen burners can be constructed from a steel sheet, from the sheet construction menu
Chemical analyzer now has a circuit board and can be deconstructed properly
Chemical analyzer and Smart Centrifuge are now their own techweb node "chemical isolation", 
Chemical analyzer board can now be printed
Reagents datum handle_reaction proc now returns TRUE when successful (it was already returning actually defined FALSEs, but nothing actually checked it's return until now)
Reduce tablesalt distilling reaction no longer has an atmospheric requirement
Reduce tablesalt distilling reaction no longer produces twice as much sodium as it should
Dexalin reaction can use tin as an inhibitor as well as water to fix an issue when trying to mass produce it in the chemical refinery
Vitals monitor circuit techweb entry is not overridden by RTGs
Improved bunsenburner icon, as it was hard to click before

:cl: Will
add: Bunsen burner, a table machine for simple distilling reactions in chemistry
add:chemical analyzer circuitboard
balance: chemical analyzer and smart centrifuge moved to their own techweb entry
fix: chemical analyzer can be properly deconstructed
fix: tin added to dexalin reaction as a possible inhibitor, to fix an issue in the chemical refinery
fix: reduce sodium distillation reaction no longer produces twice as much sodium as it should, it also does not require a specific atmosphere anymore
fix: vitals monitor techweb entry no longer accidentally overwritten by rtg techweb entry
code: handle_reaction() properly returns true or false if the reaction was successful or not
/:cl:


code/modules/reagents/machinery/bunsen_burner.dm
```diff
@@ -0,0 +1,161 @@
...
+	held_container = W
+	held_container.forceMove(src)
+	reagents.maximum_volume = held_container.reagents.maximum_volume // Update internal reagent distilling volume
+	to_chat(user,span_notice("You put \the [held_container] onto \the [src]."))
+	if(held_container.reagents.total_volume > 0)
+		start_boiling()
+	else
+		update_icon()
+
+/obj/machinery/bunsen_burner/attack_hand(mob/user as mob)
```
Kashargul: no as mob

code/modules/reagents/machinery/bunsen_burner.dm
```diff
@@ -0,0 +1,161 @@
...
+	if(..())
+		return
+	add_fingerprint(user)
+	if(!held_container)
+		to_chat(user,span_notice("There is nothing on \the [src]."))
```
Kashargul: can we please keep a space between the , and the next param?

code/modules/reagents/machinery/bunsen_burner.dm
```diff
@@ -0,0 +1,161 @@
...
+/obj/machinery/bunsen_burner/process()
+	if(!heating)
+		return
+
+	if(held_container && !anchored)
+		drop_held_container()
+		end_boil()
+		return
+
+	if(!held_container?.reagents?.reagent_list?.len)
```
Kashargul: ok... what? LAZYLEN instead of ?.len

Kashargul (approved):

PR #18369, opened 2025-08-29
selective belly import

Kashargul:
## About The Pull Request
Finally makes tgui input checkboxes usable and uses it to allow selective belly import

Also porting this https://github.com/tgstation/tgstation/pull/90227/files

## Changelog

:cl:
qol: one can now skip the lobby timer on test servers
qol: multiple bellies can be imported selectively if tgui input framework is enabled
fix: tgui input checkboxes now consistently returns null or a list of strings
/:cl:


TankTheBirb:
> qol: one can now skip the lobby timer on test servers

what’s the difference in use case between this and the start-now admin verb? Automation at the map level?

Kashargul:
> > qol: one can now skip the lobby timer on test servers
> 
> what’s the difference in use case between this and the start-now admin verb? Automation at the map level?

You have a button directly in the game ui

Issue #18366, opened 2025-08-29
[UI] Pod storage UI doesn't actually sort alphabetically


Tenryumi:
### Brief description of the bug

Smartfridges, (or at least survival pod storage,) when set to sort alphabetically, don't actually do so. Instead, they sort them by the order of the items in their list of contents.

<img width="641" height="798" alt="Image" src="https://github.com/user-attachments/assets/4769500a-12b1-4ebb-9d4f-717f4bb069fa" />

### Reproduction steps

```bash
1. Put anything in any pod storage
2. Look at the alphabetical sorted list
3. But it isn't actually :(
```

### Server Revision Info

Round ID: 809
BYOND Version: 516.1667
Server revision compiled on: 2025-08-24T19:40:35+00:00
Master commit: [ef42ca696d15804f98c3bda3cf82bb1d3a65ac81](https://github.com/VOREStation/VOREStation/commit/ef42ca696d15804f98c3bda3cf82bb1d3a65ac81)
The following pull requests are currently test merged:
[#17936: 'GITHUB API ERROR: BAD CREDENTIALS' by GITHUB API ERROR: BAD CREDENTIALS at commit c8bfb52ece](https://github.com/VOREStation/VOREStation/pull/17936)

Local commit: 4b1fcd4dec3f75ce63da3aa87f07e6054a01fdc9
TGS version: 6.18.0
DMAPI version: 7.3.3

Tenryumi:
After further investigating... this ONLY happens with pod storage on initialization. On testing, I removed the the pod's contents then put them back in, and they would then be sorted correctly. This seems to happen because the UI doesn't actually sort the items within on init.

Tenryumi:
After testing again, https://github.com/VOREStation/VOREStation/pull/18326 has fixed this! Thank you @SatinIsle !!

PR #18370, opened 2025-08-29
Fixes pod storages not initializing with their inventory sorted

Tenryumi:
## About The Pull Request

Fixes https://github.com/VOREStation/VOREStation/issues/18366 .
Turns out this issue was caused by them not sorting each item being put into them on initialize!

## Changelog
:cl:
fix: fixed capsule pod storages not initializing with their inventories sorted.
/:cl:

### DIFF
--- a/code/modules/mining/shelter_atoms_vr.dm
+++ b/code/modules/mining/shelter_atoms_vr.dm
@@ -329,6 +329,7 @@ GLOBAL_LIST_EMPTY(unique_deployable)
 	for(var/obj/item/O in loc)
 		if(accept_check(O))
 			stock(O)
+			sortTim(item_records, GLOBAL_PROC_REF(cmp_stored_item_name))
 
 /obj/machinery/smartfridge/survival_pod/accept_check(obj/item/O)
 	return isitem(O)


Kashargul:
It's an UI side issue.

Tenryumi:
> It's an UI side issue.

After testing smart fridges, I realized that this only affects pod storages. When I took out all the items in the storage and put them back in, they would be sorted correctly. After checking, I saw that this sorting proc wasn't called when items were stored on initialize. What would be a better fix for this?

Kashargul:
> > It's an UI side issue.
> 
> After testing smart fridges, I realized that this only affects pod storages. When I took out all the items in the storage and put them back in, they would be sorted correctly. After checking, I saw that this sorting proc wasn't called when items were stored on initialize. What would be a better fix for this?

It's already fixed and should be live next round. Old code on Internet explorer allowed to do this

`Alphabetical: (a: content, b: content) => a.name > b.name,`

This is not a proper native sorting function comparison though. To sort properly, we need -1, 0, 1 as returns.

`Alphabetical: (a: content, b: content) => a.name.localeCompare(b.name),`

Is the native solution. That UI slipped during the 516 updates.

Tenryumi:
Ooohh, okay, great! Yeah, I'm sadly not very knowledgeable about UI code. Thank you, this definitely sounds like a cleaner solution than my implementation lol.

PR #18381, opened 2025-08-30
some sleep cleanup

Kashargul:
@Cameron-The-Raven for you to test...

Mass drivers seem to never have set their active state back


Cameron-The-Raven:
Made the timers unique and made it so if a button is active you can't click it to restart the timer until the button is finally un-active.

Other than that, this PR seems good to go.

PR #18385, opened 2025-08-31
Better Tourette's

MeepleMuncher:
## About The Pull Request

Readds Tourette's as a neutral, less obtrusive trait. It's no longer funny haha 2006 Tourette's guy humor, but actually makes it so characters have a chance to perform various tics that change every shift, which become more frequent during times of stress.

## Changelog

:cl:
add: Readds Tourette's as a neutral, less obtrusive trait.
/:cl:

### DIFF
--- a/code/datums/components/disabilities/coprolalia.dm
+++ b/code/datums/components/disabilities/coprolalia.dm
@@ -1,0 +1,36 @@
+/datum/component/coprolalia_disability
+	var/mob/living/owner
+
+/datum/component/coprolalia_disability/Initialize()
+	if (!ishuman(parent))
+		return COMPONENT_INCOMPATIBLE
+
+	owner = parent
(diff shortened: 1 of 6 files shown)


MeepleMuncher:
I tried to portray it in the most respectful way I could... I hope is okay

SatinIsle:
I'm sorry, but we don't want tourettes to be mechanically represented here at all. I appreciate you are trying to make it less offensive (it is 100x better than old one), but it still isn't an accurate representation of the condition and may be offensive to people who actually have the condition. We also don't want to encourage people to play with serious disabilities that they do not understand. People can play a character suffering from the condition if they have a good understanding of it, but they're better doing so via actual roleplay.

Edit: If it needs to be added for downstream compability, that's not so much of an issue, but it should be commented out here.

MeepleMuncher:
completely understandable! I just saw the old tourette's in the code and felt the desire to make it less horrible, I'll comment it out in case any downstream wants to use it

PR #18407, opened 2025-09-02
Techweb SMES Coils (& small tweaks)

KillianKirilenko:
## About The Pull Request

Adds two tiers of upgraded capacity and I/O coils for SMES use, unlocked by Tier 3 and Tier 5 parts.

New coils double, then triple, their primary value, and double their secondary value twice;
\- Super Capacitance: +1000kwh cap / +50kw flow (current)
\- Ultra Capacitance: +2000kwh cap / +100kw flow (new)
\- Hyper Capacitance: +6000kwh cap / +200kw flow (new)

\- Super Transmission: +10kwh cap / +1000kw flow (current)
\- Ultra Transmission: +20kwh cap / +2000kw flow (new)
\- Hyper Transmission: +40kwh cap / +6000kw flow (new)

This is ready to merge but the values are just kinda arbitrary and I have *no idea* what good material costs would be for these.

Also moves holosign projectors and a couple of other items over into the Tools category, since they are definitely not Machine Boards.

## Changelog

:cl:
add: Added two tiers of high-tech smes coils acquirable through research nodes
qol: moved holosign projectors and a couple other tools out of machine boards and into tools where they belong
/:cl:


code/modules/research/tg/designs/stock_part_designs.dm
```diff
@@ -578,3 +578,48 @@
 		RND_CATEGORY_STOCK_PARTS + RND_SUBCATEGORY_STOCK_PARTS_5
 	)
 	departmental_flags = DEPARTMENT_BITFLAG_SCIENCE | DEPARTMENT_BITFLAG_ENGINEERING
+
```
TankTheBirb, line 581: May be worth popping in the three existing SMES coils in here. The basic one being just steel and glass, the super ones needing silver and gold as well as a bit of research. Basically as a non-cargo way of getting these things.

code/modules/research/tg/designs/stock_part_designs.dm
```diff
@@ -578,3 +578,48 @@
...
+//Super SMES Coils
+/datum/design_techweb/smes_super_capacity_ultra
+	name = "Superconductive Capacity Coil, Ultra"
+	build_type = PROTOLATHE
+	id = "smes_cap_ultra"
+	materials = list(MAT_STEEL = 1700, MAT_GLASS = 550, MAT_DURASTEEL = 230, MAT_MORPHIUM = 320, MAT_METALHYDROGEN = 600, MAT_URANIUM = 60, MAT_VERDANTIUM = 150, MAT_PHORON = 900)
```
TankTheBirb: For materials, consider using whole sheets or near whole sheet amounts since I can't see a reason to use more than a few of these at a time (except pure blissful overkill in a compact package). Each part tier reduces the cost by an extra 20%. SMES coils also won't be used that heavily, maybe only six at most a shift if someone is really doin' wacky projects. An example cost for the ultras could be something like:

`materials = list(MAT_STEEL = 10000, MAT_GLASS = 6000, MAT_GOLD = 5000, MAT_SILVER = 5000, MAT_DURASTEEL = 2000,  MAT_METALHYDROGEN = 2000, MAT_VERDANTIUM = 2000)`

For the hypers, I would increase mat costs by 50% and also add a sheet of morphium to that on top of it. But that's just me perhaps.

KillianKirilenko: The current prices are copy-pasted from one of the omni parts I think, so all input is welcome.

code/modules/research/tg/designs/stock_part_designs.dm
KillianKirilenko: I opted to add copper to the basic coil, but only went with silver for the standard super- cap/trans coils.

Nothing else uses copper at this time (as there was backlash when the extra materials were initially introduced) but there's been plenty of time for miners to become familiar with the new materials and build up a stock on both maps, and it makes more sense than just steel and glass.

Silver's a better electrical conductor than gold and copper.

code/modules/research/tg/designs/stock_part_designs.dm
KillianKirilenko: I opened to use your example values but adjusted them as follows;
\-Capacity coils have a higher Steel/Glass/Durasteel cost relative to Transmission coils
\-Transmission coils have a higher Gold/Silver cost relative to Capacity coils

Cameron-The-Raven (approved):
Looks good. More engineering stuff is always good. Will merge after CI.

Number tweaks can be made after engineers mess around with the parts and can point out if the values are too high/low.

PR #18420, opened 2025-09-07
Fluff icon fixes [IDB IGNORE]

Cameron-The-Raven:
## About The Pull Request
Fixes a lot of fluff icons.

Makes it so the on-mob and the item sprites aren't combined in the same file together.

Fixes a few oddities and bugs found with various fluff items while testing (Ex: RGB cloak did not properly update the color when swapped until you moved direction, it now works immediately when swapped)

Makes things use the more-proper default_worn_icon instead of the sledgehammer that is icon_override. icon_override is 'We don't want there to be ANY species specific sprites EVER' while default_worn_icon is 'if we don't have any slot specific or species specific icon states, we will use this icon instead'

Cleans up a lot of unneeded item_state = XXX
## Changelog
:cl: Diana
image: Splits the _mob and icon files for fluff items into two files
fix: Fixes a few broken fluff sprites
/:cl:


Cameron-The-Raven:
for reference, I spawned every single fluff item, put it on, made sure it worked, and  did any special toggles along with it.

So this  should be good to go, barring any style edits requested

PR #18421, opened 2025-09-07
Make the dud implants in the loadout free

lbnesquik:
## About The Pull Request
In the loadout, you can have some useless implants that do nothing but take space. There's also some tracking implant that costs no money you can stick in your neck, if you fancy that.

Personally, I like the idea of having some useless implanted cybernetics in my characters, so this PR reduces the price to zero, so I don't have to spend 3 whole points on things that only a doctor will see. Maybe. 
## Changelog
:cl:
balance: Made the dud implants free in the loadout
/:cl:


lbnesquik:
I do want to add some more 'useless' implants in the future, when I get off my fat ass.

PR #18427, opened 2025-09-08
Added missing complexities and added bugfixes

NovaDusklight:
## About The Pull Request

This PR focuses entirely on fixing the circuitry complexities, adding some fixes, and two very simple new circuits.
The changes will be listed concisely here for sanity sake.

* Fixed that many circuits did not have a complexity variable, defaulting tiny, or large circuits alike to a value of 3.
* Changed some complexities on circuits, such as how a basic microphone would fill most of the capacity of an assembly.
* Rebalanced some circuits. The weapon manipulation circuit, which can hold weaponry is now 50% larger, due to the circuit being exploited.
* Added a circuit gate, which will not let data pass through it if disabled by boolean.
* Added a boolean toggle circuit, which alternates a boolean each time it is pulsed.
* Fixed that the bump proc of circuits could open any door without access restriction, which meant it could open shutters.
* Fixed the pathfinder circuits seeing their targets globally. They now attempt to follow the player, or their last known location.

## Changelog

:cl: Zizzi
add: added circuits for boolean handling
fix: fixed pathfinding circuit seeing targets out of sight range.
balance: improved pathfinding circuit
balance: changed many circuit complexities
/:cl:

### DIFF
--- a/code/modules/integrated_electronics/core/assemblies.dm
+++ b/code/modules/integrated_electronics/core/assemblies.dm
@@ -475,10 +475,18 @@
 // Bump functionality, for pathfinding circuits. (Droid circuit assembly types)
 /obj/item/electronic_assembly/Bump(atom/AM)
 	..()
-	if(istype(AM, /obj/machinery/door) && can_move())
-		var/obj/machinery/door/D = AM
-		if(D.check_access(src))
-			D.open()
+	if(can_move())
+		// Check if it's an airlock or windoor. (Prevents opening blast doors and shutters)
+		if(istype(AM, /obj/machinery/door/airlock) || istype(AM, /obj/machinery/door/window))
+			var/obj/machinery/door/D = AM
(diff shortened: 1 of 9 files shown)


NovaDusklight:
> This looks all fine but I don't know enough about circuits to comment on complexity values.
> 
> Going to let someone else take a look over this and if it all seems good, can be merged.

I could be mistaken, but I don't think any current maintainers know the circuitry balance. Most of the code is about a decade old, with only very minor additions in the past.
There is another piece of balance to circuitry, which is the "space in circuit" which I have not changed in this PR. You are hard-limited in how many circuits can go into each type of assembly, regardless of complexity. (Depends on the size of the circuits inserted)

Cameron-The-Raven:
Eh, this should be fine.

If values end up needing to be tweaked they can be further adjusted in another PR.

PR #18431, opened 2025-09-09
Disposal Connector Component

Willburd:
## About The Pull Request
Moves disposal flushing and receiving code to a component, so that any other object can interact with the disposal system by just having the component. This PR also grew in scope a bit as I realized how ancient disposal code was. Fixes many old code style issues to reach at least basic review standards. Please review this as if it was a new addition and not just some patches to old code. I'd rather this be properly modernized now instead of sitting for another 15 years. If anything jank stands out let me know.

## Changelog
Adds disposal_system_connection component
Adds COMSIG_DISPOSAL_FLUSH and COMSIG_DISPOSAL_RECEIVING component
Moves the massively bloated disposals.dm into multiple smaller files for each object
Moves disposal flushing and trunk to machine transfer procs to signals
Fixes an issue where being ejected from disposals when the turf has flooring above the pipe, would result in an invisible null flooring stack being made. Properly breaks the turf to stack now.
Removes a lot of sleep()s from very old code, uses do_after() for tool actions, and timers for animations/movement
Uses get_turf() instead of src.loc as well as forceMove(). Also fixes some checks using null loc to detect if stuff is qdeleted.
Collapsed the disposals_vr.dmi into the main one, and renamed intake-closing to intake-flush so that a single line change in an entirely duped proc didn't need to exist anymore.
Added "stattracking" var to disposals, for if a disposal machine should count for the round-end stats when flushed.

:cl: Will
add: disposal_system_connection component
code: COMSIG_DISPOSAL_FLUSH  signal
code: COMSIG_DISPOSAL_RECEIVING  signal
code: disposal code cleaned up, moved to their own files
refactor: Sleep() removed from disposal packet movement, now uses timers
refactor: disposal code no longer uses sleeps for actions, and properly uses do_after
refactor: disposal code uses get_turf() instead of src.loc for moving things to its turf
fix: Being ejected from disposals while under floor paneling now properly breaks and drops the correct floor tile stack, instead of an invisible null flooring stack
/:cl:


Willburd:
Everything is tested working with disposals. Unfortunately I expect some things from the old code that will raise a few eyebrows in review. Especially with a bunch of code being split into other files.

Willburd:
Did more cleanup, modernized a lot of stuff like src.loc to get_turf(). Fixed sleep() being used as a do_after() in trunk disassembly code. A lot of this code is ancient and was in dire need of modernization. As well, I split up all of the disposal stuff into their own files. I know it messes with downstreams, but a 2k long file of different objects is awful to look through.

code/modules/recycling/sortingmachinery.dm
```diff
@@ -443,17 +441,13 @@
 /obj/machinery/disposal/deliveryChute/flush()
 	flushing = 1
 	flick("intake-closing", src)
-	var/obj/structure/disposalholder/H = new()	// virtual holder object which actually
-												// travels through the pipes.
-	air_contents = new()		// new empty gas resv.
-
 	sleep(10)
```
Kashargul: not sure we want to keep those sleeps around

code/modules/recycling/disposal_holder.dm
```diff
@@ -0,0 +1,153 @@
...
+	dir = 0
+
+	// initialize a holder from the contents of a disposal unit
+/obj/structure/disposalholder/proc/init(list/flush_list, datum/gas_mixture/flush_gas)
+	gas = flush_gas// transfer gas resv. into holder object -- let's be explicit about the data this proc consumes, please.
+
+	//Check for any living mobs trigger hasmob.
+	//hasmob effects whether the package goes to cargo or its tagged destination.
+	for(var/mob/living/M in flush_list)
+		if(M.stat != 2 && !istype(M,/mob/living/silicon/robot/drone))
```
Cameron-The-Raven:
```suggestion
		if(M.stat != DEAD && !istype(M,/mob/living/silicon/robot/drone))
```

code/modules/recycling/disposal_holder.dm
```diff
@@ -0,0 +1,153 @@
...
+			M.client.perspective = EYE_PERSPECTIVE
+			M.client.eye = src
+
+	//Checks 1 contents level deep. This means that players can be sent through disposals...
+	//...but it should require a second person to open the package. (i.e. person inside a wrapped locker)
+	for(var/obj/O in flush_list)
+		if(!O.contents)
+			continue
+		for(var/mob/living/M in O.contents)
+			if(M && M.stat != 2 && !istype(M,/mob/living/silicon/robot/drone))
```
Cameron-The-Raven:
```suggestion
			if(M && M.stat != DEAD && !istype(M,/mob/living/silicon/robot/drone))
```

code/modules/recycling/disposal_holder.dm
```diff
@@ -0,0 +1,153 @@
...
+// movement process, persists while holder is moving through pipes
+/obj/structure/disposalholder/proc/move()
+	if(!active)
+		return
+
+	// Clonk n' bonk
+	if(hasmob && prob(3))
+		for(var/mob/living/H in src)
+			if(!istype(H,/mob/living/silicon/robot/drone)) //Drones use the mailing code to move through the disposal system,
+				H.take_overall_damage(20, 0, "Blunt Trauma")//horribly maim any living creature jumping down disposals.  c'est la vie
```
Cameron-The-Raven, line 66: Just leaving a note here so this can be commented out on Chomp, since they don't have lethal disposals

code/modules/recycling/disposal_machines.dm
```diff
@@ -0,0 +1,495 @@
...
+					C.update()
+					qdel(src)
+				return
+			else
+				to_chat(user, "You need more welding fuel to complete this task.")
+				return
+
+	if(istype(I, /obj/item/melee/energy/blade))
+		to_chat(user, "You can't place that item inside the disposal unit.")
+		return
```
Cameron-The-Raven:
```suggestion
```

Really no reason for this to even exist...This seems like some kind of bandaid fix for 'placing eblades in a disposal bin and not having it go into it's un-equipped state' which if that does turn out to be the case, should just be fixed directly.

PR #18440, opened 2025-09-10
Rotatable Atom Element

Willburd:
## About The Pull Request
Moves all rotation and flipping object code to a unified element and set of verbs given by addelement(). Instead of 10 different copypastes of the same code in different objects.

## Changelog
Refactors rotation code for modern sensibilities.
rotatable element gives atoms verbs for rotations.
base procs on atoms for rotation handling, but not directly exposed as verbs unless given by the element
many refactors throughout the code to remove the now defunct rotation verbs

:cl: Will
code: rotatable element, gives atoms the rotation verbs in a safe and clean way 
refactor: removed all defunct rotation verbs to use the rotatable element instead
/:cl:


Willburd:
Tested multiple times on each object and redone my initial plans. This should be honed to a pretty sharp edge now, and not just more spaghetti on top.

PR #18443, opened 2025-09-10
Swaps fighter landing gear raise/lower command wordings

Tenryumi:
## About The Pull Request

Noticed that fighters' landing gear equipment said it would raise the landing gear when it was already raised and lower it when it was already lowered. This is a simple swap fix for that.

## Changelog

:cl: Ryumi
fix: Fighters will now correctly say whether or not you'll raise or lower their landing gear.
/:cl:

### DIFF
--- a/code/game/mecha/combat/fighter.dm
+++ b/code/game/mecha/combat/fighter.dm
@@ -139,7 +139,7 @@
 	..()
 	if (href_list["toggle_landing_gear"])
 		landing_gear_raised = !landing_gear_raised
-		send_byjax(src.occupant,"exosuit.browser","landing_gear_command","[landing_gear_raised?"Raise":"Lower"] landing gear")
+		send_byjax(src.occupant,"exosuit.browser","landing_gear_command","[landing_gear_raised?"Lower":"Raise"] landing gear")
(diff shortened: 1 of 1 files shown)


code/game/mecha/combat/fighter.dm
```diff
@@ -139,7 +139,7 @@
 	..()
 	if (href_list["toggle_landing_gear"])
 		landing_gear_raised = !landing_gear_raised
-		send_byjax(src.occupant,"exosuit.browser","landing_gear_command","[landing_gear_raised?"Raise":"Lower"] landing gear")
+		send_byjax(src.occupant,"exosuit.browser","landing_gear_command","[landing_gear_raised?"Lower":"Raise"] landing gear")
 		src.occupant_message(span_notice("Landing gear [landing_gear_raised? "raised" : "lowered"]."))
```
Kashargul, line 143: why is it still inverted down here then?

Tenryumi: Because these are shown after the landing gear has been toggled, so the message displays correctly here. For the other one though, it's shown before it's raised/lowered, so it gets it backwards.

PR #18465, opened 2025-09-12
Add an implantable health scanner

lbnesquik:
## About The Pull Request
This PR adds a health scanner implants. I've essentially mashed it as a copy of the sprint implant, adding them to pretty much all the same spot, including traitor uplinks. It is much cheaper, though. 

Also, there were some names that weren't correct, and some paths that were a little silly, so I've changed those as I saw them. 

## Changelog
:cl:
add: Added an implantable health scanner
fix: fixed incorrect names
/:cl:


Cameron-The-Raven:
Going to make it a hand augment instead of a groin augment.

As funny as having the health analyzer come out of your groin is - let's not

Edit: Just realized it only scans the owner...Hm. Not too sure, now.

Cameron-The-Raven:
Eh, seems fine. Made it so it can be made via R&D (like the rest of the augments, minus sprinter) and that it has an advanced scanner instead of a basic one (Since it takes an entire augment slot)

PR #18467, opened 2025-09-12
Improved Smes Map Upgrades

Willburd:
## About The Pull Request
Old code needed cleaning up, and kash wants to lint away the sins of map-set smes vars.

> [!WARNING]
> THIS WILL REQUIRE MAP EDITS. Most engine smes setups were coded with magic numbers guessing the max charge and output of a smes. many of them do not match their coil upgrade levels at all, so do not deconstruct legally. You should place the correct coils on them in the map (auto upgrades) and use the prefab that fits most from this PR for setting the charge state and output levels/settings.  

## Changelog
Moves apply_mapped_upgrades() to a base machine empty proc
Fixes smes calling mapped part upgrade when being built (only on map init now)
added apply_mapped_settings() to smes late init, so mapped subtypes can set configs like charge and output on map load.
ported rtg mapped upgrades from outpost 21, same feature as upgrading the smes

:cl: Will
refactor: Improves smes mapped part upgrading
refactor: Adds more smes subtypes so manually mapping smes vars is no longer needed
server: Mapped smes should be updated to use prefabs, and a coil place on the smes to meet the original charge capacity.  
/:cl:


code/modules/power/port_gen_vr.dm
```diff
@@ -97,6 +97,48 @@
...
+			C.forceMove(src.loc)
+			C.Destroy()
+	if(locate(/obj/item/stock_parts/micro_laser) in parts_found)
+		while(TRUE)
+			var/obj/item/stock_parts/micro_laser/M = locate(/obj/item/stock_parts/micro_laser) in component_parts
+			if(isnull(M))
+				break
+			component_parts.Remove(M)
+			M.forceMove(src.loc)
+			M.Destroy()
```
Kashargul: Destroy is never allowed to be called by hand.

USE qdel!

code/modules/power/port_gen_vr.dm
```diff
@@ -97,6 +97,48 @@
...
+	if(parts_found.len == 0)
+		return
+	if(locate(/obj/item/stock_parts/capacitor) in parts_found)
+		while(TRUE)
+			var/obj/item/stock_parts/capacitor/C = locate(/obj/item/stock_parts/capacitor) in component_parts
+			if(isnull(C))
+				break
+			component_parts.Remove(C)
+			C.forceMove(src.loc)
+			C.Destroy()
```
Kashargul: Destroy is never allowed to be called by hand.

USE qdel!

Willburd: Corrected, this was old shitcode from before i understood how qdel even works, and habits from gml coding where Destroy() is the qdel() lol.

PR #18480, opened 2025-09-13
Implements Autowiki

ItsSelis:
## About The Pull Request
With this feature we can get wiki templates automatically created from the latest code that we can use in our wiki. This will reduce a lot of manual work in things like recipes.

Currently autowiki has been enabled for science techwebs. There is also code to generate it for vending machine contents, but due to an issue it will be enabled later on.

## Changelog
:cl:
add: Added autowiki functionality to automatically create wiki templates from code
/:cl:


SatinIsle:
Yooo, this is extremely cool. How easy it for me to make it apply to more things?

PR #18494, opened 2025-09-14
Stop removing edge whitespace from IC string input

### LINKED ISSUE #18462: [GENERAL] Inputting a space into circuitry auto-converts it to NULL
### Brief description of the bug

Any text that consists of an empty space is auto-converted. this is most notable when trying to set up a string exploder to split with spaces.

This occurs with any sort of input method I've tried, Keypad, Debugger, Old inputs, TGUI input framework, et cetera. it just sets it to NULL 

Image

### What you expected to happen

the data to be set to " "

### What actually happened

the data is set to NULL

### Reproduction steps

```bash
1. use circuit debugger in hand.
2. select "string".
3. input an empty space.
4. get NULL'd, idiot.
...
```

### Server Revision Info

Round ID: 863
BYOND Version: 516.1667
Server revision compiled on: 2025-09-11T15:28:38+00:00
Master commit: [b0f0f4685ff28c75600fa20f58eeb5f6875987a0](https://github.com/VOREStation/VOREStation/commit/b0f0f4685ff28c75600fa20f58eeb5f6875987a0)
The following pull requests are currently test merged:
[#17936: 'Toggle on Tether' by ItsSelis at commit c8bfb52ece](https://github.com/VOREStation/VOREStation/pull/17936)

Local commit: 59f0bb843ef76933dacb24d35829f558ff101702
TGS version: 6.18.0
DMAPI version: 7.3.3

### Anything else you may wish to add

:(

NickBelmont:
## About The Pull Request
Fixes #18462 and an associated issue where leading/trailing spaces would also be deleted. Textpads and debuggers were encoding text, which trims edge whitespace, and also seems like it's not needed, since other methods in IC code explicitly set encode to 0 or False.

## Changelog

:cl:
fix: IC Textpad and debugger no longer trim whitespace at edges.
/:cl:

### DIFF
--- a/code/modules/integrated_electronics/core/special_pins/char_pin.dm
+++ b/code/modules/integrated_electronics/core/special_pins/char_pin.dm
@@ -3,7 +3,7 @@
 	name = "char pin"
 
 /datum/integrated_io/char/ask_for_pin_data(mob/user)
-	var/new_data = tgui_input_text(user, "Please type in one character.","[src] char writing")
+	var/new_data = tgui_input_text(user, "Please type in one character.","[src] char writing", encode = FALSE)
 	if(holder.check_interactivity(user) )
 		to_chat(user, span_notice("You input [new_data ? "new_data" : "NULL"] into the pin."))
 		write_data_to_pin(new_data)
--- a/code/modules/integrated_electronics/subtypes/input.dm
+++ b/code/modules/integrated_electronics/subtypes/input.dm
@@ -103,7 +103,7 @@
 	power_draw_per_use = 4
 
 /obj/item/integrated_circuit/input/textpad/ask_for_input(mob/user)
-	var/new_input = tgui_input_text(user, "Enter some words, please.", "Text pad", get_pin_data(IC_OUTPUT, 1), MAX_KEYPAD_INPUT_LEN, FALSE, FALSE)
+	var/new_input = tgui_input_text(user, "Enter some words, please.", "Text pad", get_pin_data(IC_OUTPUT, 1), MAX_KEYPAD_INPUT_LEN, encode = FALSE)
 	if(istext(new_input) && CanInteract(user, GLOB.tgui_physical_state))
 		set_pin_data(IC_OUTPUT, 1, new_input)
 		push_data()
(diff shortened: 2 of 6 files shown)


Kashargul (changes requested):
No. Sanitize inputs. Removing encode is not the way to go if there is no sanitize.

code/modules/integrated_electronics/subtypes/input.dm
```diff
@@ -103,7 +103,7 @@
 	power_draw_per_use = 4
 
 /obj/item/integrated_circuit/input/textpad/ask_for_input(mob/user)
-	var/new_input = tgui_input_text(user, "Enter some words, please.", "Text pad", get_pin_data(IC_OUTPUT, 1), MAX_KEYPAD_INPUT_LEN)
+	var/new_input = tgui_input_text(user, "Enter some words, please.", "Text pad", get_pin_data(IC_OUTPUT, 1), MAX_KEYPAD_INPUT_LEN, encode = FALSE)
```
Kashargul: missing sanitize

code/modules/integrated_electronics/subtypes/memory.dm
```diff
@@ -100,7 +100,7 @@
 	switch(type_to_use)
 		if("string")
 			accepting_refs = 0
-			new_data = tgui_input_text(user, "Now type in a string.","[src] string writing","", MAX_NAME_LEN)
+			new_data = tgui_input_text(user, "Now type in a string.","[src] string writing","", MAX_NAME_LEN, encode = FALSE)
```
Kashargul: missing sanitize

NickBelmont:
There we go, inputs are now sanitized, and by doing so, input can also still be encoded without trimming.

code/modules/integrated_electronics/core/special_pins/char_pin.dm
```diff
@@ -3,7 +3,7 @@
 	name = "char pin"
 
 /datum/integrated_io/char/ask_for_pin_data(mob/user)
-	var/new_data = tgui_input_text(user, "Please type in one character.","[src] char writing")
+	var/new_data = sanitizeSafe(tgui_input_text(user, "Please type in one character.","[src] char writing", encode = FALSE), encode = 1, trim = 0)
```
Kashargul: 1 character max length should be 1, right?

NickBelmont: Don't see why it shouldn't be, the max length is enforced later when writing the data to the pin, but certainly doesn't hurt to also do it on input.

Kashargul (approved):

Issue #17197, opened 2025-02-21
[GENERAL] Resin doors are unbreakable


TeleTubby2:
### Brief description of the bug

Resin doors never get breakable set, so someone (Xeno, Resin-inclined jorfing it replicant enjoyer) can place a door and it's effectively impossible to remove.

Probably should set it as either part of the subtype, or in /mob/living/carbon/human/proc/resin().

### Reproduction steps

```bash
1.
2.
3.
...
```

PR #18495, opened 2025-09-14
Fix resin doors not being breakable

lbnesquik:
## About The Pull Request
Fixes [#17197](https://github.com/VOREStation/VOREStation/issues/17197)
## Changelog
:cl:
fix: fixed simple doors not being breakable
/:cl:


code/game/objects/structures/simple_doors.dm
```diff
@@ -4,6 +4,7 @@
 	density = TRUE
 	anchored = TRUE
 	can_atmos_pass = ATMOS_PASS_DENSITY
+	breakable = 1
```
Kashargul: TRUE

lbnesquik: image

Kashargul: see above

anchored = TRUE

we don't want 1/0 for boolean values anymore

lbnesquik: I was going to fix it, I'm just outside ATM!

PR #18499, opened 2025-09-15
Tethered Item Component

Willburd:
## About The Pull Request
Tethered items are rare, but use mostly copypasted code. Moved to a component to share code and reduce possible bad cases where one item works correctly, but all others do not. Unfortunately requires some specific structure to make proper use of it. 

Component must be added before parent is called in Initialize() and attack_hand() in the host item must send a signal for COMSIG_ITEM_ATTACK_SELF to check if we are able to remove the sub item, or if we are picking it up instead. It is unfortunately not possible to block the pickup action on a mob, seemingly even in TG code, while still retaining the ability for mobs to pick it up while not worn. My only option was to add a check to attack_hand, and handle it there like how it was done originally. Despite this downside, the code is now unified, and any future objects only need to follow this pattern to safely use the component.

I kinda expect the review for this to be a bit grungly. Any better options are welcome.

## Changelog
Added tethered item component
Moved defibs to tethered item component
Moved bluespace radios to tethered item component
Unified tethered item ui action button
Unified tethered item toggle verb

:cl: Will
code: Adds tethered item component for items that have a handheld subitem that can be removed from it while it is worn
/:cl:


Willburd:
Needs significant testing from someone other than me. I've verified basic functionality, but this really needs to be broken to make sure I got everything possible handled.

PR #18510, opened 2025-09-16
various map related features

zeskorion:
## About The Pull Request
Various changes, mostly adding map-related objects for a map I'm making downstream. 
Player facing changes:
- Maint blood and vomit is old and dried. This means maint vomit that spawns randomly can be used as virus samples, and the blood wont get your feet bloody. Blood will also have different colors. Does not change oil
-  catwalks prevent turfs beneath from slowing the player
- water turfs can break falls like the soft landing perk can
- adds a version of the aquatic trait that makes you swim noticeably faster
- emergency lights now stay red during nightshift

Mapperbus changes:
- Adds a micro hole that takes the user to a random valid destination
- adds caution stripe ledges
- adds false cosmetic stairs as an object
- adds a maint turf initializer that does not spawn junk, but still makes floors dusty
- water turfs can now have custom overlay colors. Digestive enzymes now show green water over the player sprite, instead of blue water.
- a new version of the glitch NIF which is not bioadaptive nor higher durability
- adds the nanite-infested floor. Available to mappers

Nanite infested floor is intended for use in AI chambers. It should be surrounded by  /obj/structure/railing/overhang/hazard/nanite, and entrances should use /obj/structure/dummystairs/hazardledge. Patches of nanite goop are activated based on a common ID var
It acts like normal floors when inactive. 
If there is an SMES in an area, all nanite goo in the area will link to the SMES. SMES linked nanite goo can only be controlled by an AI and ai shells. Otherwise, the ai, borgs, and carbons with a NIF can control the goo. Nutrients from digested creatures/things will be given to the controller (or an SMES in the area). AI always have priority over goop, on a first come first serve basis, unless their core is adjacent, in which case they always have priority. If a non-ai controller leaves the area the goop is in, it will become inactive when interacted with
When activated, the goop becomes a thick liquid that imparts very high slowdown. It can digest mobs inside of it, adding painkiller to carbons. The controller decides if it digests, and if borgs and/or synthetics should be ignored. The controller is always ignored, unless they are an organic with an NIF, in which case, they cannot be ignored.
https://cdn.discordapp.com/attachments/404660884531707906/1416699457700888637/CHOMPStation_2025-09-14_01-15-06.mp4?ex=68cb176f&is=68c9c5ef&hm=03bd432e2d5cc8debe0b813a45481b830a6c887ed436e5b58ac8083394f1ee23&

## Changelog

:cl:
add: Added several new mapping features for your local mapper
add: digestive nanite turfs. currently mapperbus only
add: faster swimming aquatic trait
del: Removed old things
qol: catwalks now prevent turfs from slowing people
qol: water turfs now break falls
qol: Maint blood and vomit is old and dried.
balance:  water turfs now break falls
fix: emergency lights stay red during nightshift
/:cl:


code/game/turfs/simulated/nanogoop.dm
```diff
@@ -0,0 +1,338 @@
...
+
+/turf/simulated/floor/water/digestive_enzymes/nanites/Initialize(mapload)
+	. = ..()
+	for(var/obj/machinery/power/smes/M in GLOB.smeses)
+		if(!M)
+			continue
+		if(!get_area(M))
+			continue
+		if(get_area(M) == get_area(src))
+			smes = M
```
Kashargul: you are settin references, where do you clean them UP?

code/game/turfs/simulated/nanogoop.dm
```diff
@@ -0,0 +1,338 @@
...
+	reagent_type = REAGENT_ID_LIQUIDPROTEAN
+	oxygen		= MOLES_O2STANDARD
+	nitrogen	= MOLES_N2STANDARD
+	temperature = T20C
+	var/digesting = FALSE
+	var/digest_synth = FALSE
+	var/digest_robot = FALSE
+	var/active = FALSE
+	var/usesmes = TRUE
+	var/mob/living/nutrienttarget
```
Kashargul: same for this refernece, what if the turf gets changed, where do you clear it? Destroy usually!

zeskorion: switched this all to weakrefs

code/game/turfs/simulated/nanogoop.dm
zeskorion: switched this all to weakrefs

code/game/turfs/simulated/nanogoop.dm
```diff
@@ -0,0 +1,352 @@
...
+		active = FALSE
+		for(var/obj/structure/railing/overhang/hazard/nanite/R in src)
+			R.icon_modifier = "inactive_"
+			R.icon_state = "inactive_railing0"
+		for(var/obj/structure/railing/overhang/hazard/nanite/R in range(src, 1))
+			R.update_icon()
+		for(var/obj/structure/dummystairs/hazardledge/stairs in src)
+			stairs.icon_state = "stair_hazard"
+			stairs.update_icon()
+	else
```
Kashargul: early return is to be used over if / else

code/game/turfs/simulated/nanogoop.dm
```diff
@@ -0,0 +1,352 @@
...
+			var/mob/living/silicon/robot/R = nutrienttarget
+			if(R.cell)
+				R.cell.give(amt * 20)
+				return
+
+
+/turf/simulated/floor/water/digestive_enzymes/nanites/return_air_for_internal_lifeform(var/mob/living/L)
+	if(!can_digest(L))
+		return return_air() //Nanites should always be nonlethal until the AI turns on digestion
+	else
```
Kashargul: this else can never be reached, delete it

code/game/turfs/simulated/nanogoop.dm
```diff
@@ -0,0 +1,352 @@
...
+		return ..()
```
Kashargul: move one indent back

code/game/turfs/simulated/nanogoop.dm
```diff
@@ -0,0 +1,352 @@
...
+			if(H.isSynthetic())
+				if(!digest_synth)
+					return FALSE
+				if(L == nutrienttarget)
+					return FALSE
+			if(!H.pl_suit_protected())
+				return TRUE
+			if(H.resting && !H.pl_head_protected())
+				return TRUE
+		else return TRUE
```
Kashargul: why is there an else? nut just a return TRUE

zeskorion:
tested after fixes....

zeskorion:
oh heck
i included the map changes oops, didnt wanna do that

zeskorion:
that ought to fix it

Cameron-The-Raven:
Will merge after CI

PR #18511, opened 2025-09-17
Modernizing doors

Willburd:
## About The Pull Request
Refactors doors to remove _vr overrides and attempt_vr() stuff. Implements signal. Should not have any player facing changes. God I hope it doesn't. This code was a rat nest, even on TG it looks flaky. This has very slight differences from TG, in that items used on airlocks sends a signal before trying to open the door, so the signal handlers can prevent door use. 

Unfortunately like TG code, all preceding item checks like crowbars take priority over generic items. So you'll probably just want to hook open and close in most cases. Even TG has hardcoded checks calling the parent proc for it's signals when barricading doors. Door code is a mess.

attempt_vr() and hook_vr() are also removed. It makes this PR a bit loaded, but these need to go someday. What's better than today. DCS has long replaced this method of code modularization, and only a few calls remained. This pr clears them up.

## Changelog
Moves all subtypes in airlock.dm to their own file.
Heavily refactors door code to be closer to TG
Implements COMSIG_DOOR_CAN_OPEN
Implements COMSIG_DOOR_CAN_CLOSE
Implements COMSIG_DOOR_OPEN 
Implements COMSIG_DOOR_CLOSE 
Implements COMSIG_AIRLOCK_SET_BOLT 
Implements COMSIG_MOB_BUMPED_DOOR_OPEN 
Removes all attempt_vr() calls
Removes all hook_vr() calls
Removes all MinutesToTicks in favor of MINUTES define
Removes all SecondsToTicks in favor of SECONDS define

:cl:
refactor: Huge door code cleanup
refactor: attempt_vr() removal
refactor: hook_vr() removal
refactor: few remaining MinutesToTicks and SecondsToTicks updated to use defines instead
code: Implements COMSIG_DOOR_CAN_OPEN
code: Implements COMSIG_DOOR_CAN_CLOSE
code: Implements COMSIG_DOOR_OPEN 
code: Implements COMSIG_DOOR_CLOSE 
code: Implements COMSIG_AIRLOCK_SET_BOLT 
code: Implements COMSIG_MOB_BUMPED_DOOR_OPEN 
/:cl:


Willburd:
It all seems to be functional, but review is needed for sure.

code/_onclick/item_attack.dm
```diff
@@ -65,7 +65,8 @@ avoid code duplication. This includes items that may sometimes act as a standard
 	if(can_operate(src, user) && I.do_surgery(src,user))
 		return TRUE
 
-	if(attempt_vr(src,"vore_attackby",args)) return //VOREStation Add - The vore, of course.
+	if(vore_attackby( I, user)) // The vore, of course.
```
Cameron-The-Raven:
```suggestion
	if(vore_attackby(I, user)) // The vore, of course.
```

Cameron-The-Raven (commented):
Looks good. Will give it a test when I get a bit.

Cameron-The-Raven:
Tested and everything seems good.

PR #18519, opened 2025-09-18
Adds an alt service MEKA V2 borg sprite without the tray

Tenryumi:
## About The Pull Request

A simple edited version of the pre-existing MEKA V2 service borg sprite that doesn't have the tray, for players like me who love the sprite but would prefer it without it.

dreamseeker_2025-09-18_02-54-52

## Changelog
:cl: Ryumi
image: Added an alternate service MEKA V2 borg sprite that doesn't have the drink/cigarette tray.
/:cl:


Tenryumi:
oops- I forgot to add the lights and panel parts-

Kashargul (changes requested):
No. Use the decal system. We won't start that mess again.

Tenryumi:
> No. Use the decal system. We won't start that mess again.

Oh, alright! Sorry, I'm kinda unfamiliar with borg code/mechanics and didn't know that was an option.

Tenryumi:
After trying to figure out what I was gonna need to do, I'd need to:
- Break off the arms of both this and the original version of the borg sprites and make them into separate sprites
- Make a new type of overlay that makes the "eyes" change icon state in specific scenarios to fit the alt versions
- Edit the module choice UI to make it so it responds to the presence of these alts and makes you choose between the two (using a normal "Yes/no" prompt wouldn't work because the user could potentially close it and break the whole process)

I lack the knowledge and experience for this, and it's too far outside the scope of what I originally wanted to do with this. Someone else more willing/capable is free to do something like that, but I give up on this.

PR #18539, opened 2025-09-19
Fix material stacks runtime when stacked entirely into another stack

Willburd:
## About The Pull Request
Stacking one stack entirely into another results in a runtime. This happens now because the material datum reference is now cleaned on Destroy() but the old code still expected it to be present.

## Changelog
Fixes stacking materials by checking if the stack was qdeleted before updating string data. Previously it only checked if the reference vars used were null. Which will not protect it from qdeleted stacks with nulled material vars.

:cl: Will
fix: Runtime when combining one stack entirely into another
/:cl:


Willburd:
Resolved in another PR

PR #18545, opened 2025-09-20
Boneless Trait

Cameron-The-Raven:
## About The Pull Request
Boneless trait for people that lack bones (Synx), wouldn't logically have bones (plant-people), goopy people that aren't quite prometheans but are still gooey, or otherwise!

Adds two variants: Boneless and Boneless Major

Boneless: Gives -3 trait points. Limb health remains untouched.
Boneless, Major: Gives -6 trait points. Limb health is reduced by half.

Copy-pasting my coder comment here to explain why this is a _very_ harsh negative, even though it sounds like an upside on the surface:

So, I know what you're thinking.
Why is this a negative trait? It sounds like a positive one!"
In theory, yes. Being unable to have your limbs break DOES sound like a good thing!
However, that is where limb code comes into play.
Normally, with highly damaged limbs, you have a RNG(X)% chance for it to be sliced off if it takes a lot of damage and the damage from the blow is over a certain threshold.
Additionally, limbs have a 'maximum damage' it can reach in which it can't reach anymore. Your arm, for example, is 80 damage maximum. Head 75.
Trying to hit it more won't do any more damage, just constantly roll chances (if the hitting blow is strong enough) to cut it off/gib it.
With this trait, however, as soon as you reach that maximum damage, your limb IMMEDIATELY gibs. No RNG chance rolls. 'attack must be X strong to hit it off'. Nothing.
Additionally, normal limbloss code has one-shot protection and prevents bullets from gibbing limbs. This makes you forfeit that protection.

For example, with the major version, you can hit someone in the head 3 times with a lit welder and they will die. Twice if they have burn weakness. Likewise, having a brain (with no head) makes surgery to bring someone back alive harder.

## Changelog

:cl: Diana
add: Adds boneless trait
fix: Brains no longer qdel() if the head is gibbed.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/traits/negative.dm
+++ b/code/modules/mob/living/carbon/human/species/station/traits/negative.dm
@@ -853,7 +853,7 @@
 
 /datum/trait/negative/boneless/major
 	name = "Boneless, Major"
-	desc = "You have no bones! Though your limbs are also much, much easier to gib in exchange. (Seriously this can result in one shot deaths and similar)"
+	desc = "You have no bones! Your limbs are also much, much easier to gib in exchange. (Seriously this can result in one shot deaths and similar)"
 	cost = -6 //For reference, getting hit in the head with a welder 3 times kills you. Head has 37.5 HP. Reaching that cap = instant death
 	limb_health = 0.5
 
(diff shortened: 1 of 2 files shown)


Cameron-The-Raven:
Merging after CI. Have mentioned this twice in adminchat and no complaints/suggested changes were noted and it's been sitting here for a month.

PR #18552, opened 2025-09-22
Increases cost of virus immunity

Cameron-The-Raven:
## About The Pull Request
Virus immunity goes from 1 -> 3 points

An **IMMUNITY** trait should not be 1 point. Resistance, sure. 
But outright immunity is absurd.
## Changelog
:cl: Diana
balance: Makes Virus Immunity trait 3 points
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/traits/positive.dm
+++ b/code/modules/mob/living/carbon/human/species/station/traits/positive.dm
@@ -535,7 +535,7 @@
 /datum/trait/positive/virus_immune
 	name = "Virus Immune"
 	desc = "You are immune to viruses."
-	cost = 1
+	cost = 3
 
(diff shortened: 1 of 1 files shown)


zeskorion:
Virus immunity is 1 point rn, i believe, as an option to disengage from a system that's currently pretty eork in progress, and sucks to engage with in the rare time an outtbreak does happen

TheCaramelion:
> Virus immunity is 1 point rn, i believe, as an option to disengage from a system that's currently pretty eork in progress, and sucks to engage with in the rare time an outtbreak does happen

I only upported it without any point changes couple months ago when things were more unstable. No one blew up (again) due to viruses, so it should be fine

lbnesquik:
We very rarely have any sort of virus going around.

PR #18557, opened 2025-09-24
a few more changes related to map

zeskorion:
## About The Pull Request

a few changes and fixes. 
- Adds a var to turfs to prevent them from counting as outdoors for shields
- upports part of the random micro tunnel code i forgot
- fixes a nanogoop runtime due to mob sometimes having a client when deleted

## Changelog

:cl:
fix: few runtimes and new vars
/:cl:

### DIFF
--- a/code/game/turfs/turf.dm
+++ b/code/game/turfs/turf.dm
@@ -35,7 +35,6 @@
 	var/can_build_into_floor = FALSE // Used for things like RCDs (and maybe lattices/floor tiles in the future), to see if a floor should replace it.
 	var/list/dangerous_objects // List of 'dangerous' objs that the turf holds that can cause something bad to happen when stepped on, used for AI mobs.
 	var/tmp/changing_turf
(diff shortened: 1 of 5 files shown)


code/game/turfs/turf.dm
```diff
@@ -35,6 +35,7 @@
 	var/can_build_into_floor = FALSE // Used for things like RCDs (and maybe lattices/floor tiles in the future), to see if a floor should replace it.
 	var/list/dangerous_objects // List of 'dangerous' objs that the turf holds that can cause something bad to happen when stepped on, used for AI mobs.
 	var/tmp/changing_turf
+	var/shieldable = TRUE //can set this to FALSE to prevent this turf from having shields over it
```
Kashargul: why aren't we using flags here? TURF_CAN_SHIELD

code\__defines\turfs.dm

zeskorion: fixed, should work but testing now to be sure

zeskorion: got it working with a flag

PR #18559, opened 2025-09-25
Oil and blood with no dna can be streaked again

Willburd:
## About The Pull Request
Oil and blood spawned from maintenance have no dna, this prevent streaking them along the ground. However this check was only done on shoes. 

## Changelog
Fixes shoes being unable to spread bloody footsteps if the blood was mapspawned

:cl: Will
fix: Wearing shoes no longer prevents map spawned blood from spreading with footsteps
/:cl:

### DIFF
--- a/code/game/turfs/simulated.dm
+++ b/code/game/turfs/simulated.dm
@@ -118,9 +118,9 @@
 				var/obj/item/clothing/shoes/S = H.shoes
 				if(istype(S))
 					S.handle_movement(src,(H.m_intent == I_RUN ? 1 : 0), H) // handle_movement now needs to know who is moving, for inshoe steppies
-					if(S.track_blood && S.forensic_data?.has_blooddna())
+					if(S.track_blood)
 						bloodDNA = S.forensic_data.get_blooddna()
-						bloodcolor=S.blood_color
+						bloodcolor = S.blood_color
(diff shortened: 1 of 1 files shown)


code/game/turfs/simulated.dm
```diff
@@ -118,9 +118,9 @@
 				var/obj/item/clothing/shoes/S = H.shoes
 				if(istype(S))
 					S.handle_movement(src,(H.m_intent == I_RUN ? 1 : 0), H) // handle_movement now needs to know who is moving, for inshoe steppies
-					if(S.track_blood && S.forensic_data?.has_blooddna())
+					if(S.track_blood)
 						bloodDNA = S.forensic_data.get_blooddna()
```
Kashargul: This was null checked above, is it truly safe to just access it now?

Willburd: get_blooddna() returns an empty list if the forsensic's data is null. in this condition it means the blood contains null dna. which will always be seen as an empty list as well by any of the machines. This case is mostly for oil as well. No runtimes during testing, but i can be more extensive with my test if desired. footsteps with barefeet worked the same before, and was changed to spread null dna blood puddles much earlier. I'm more surprised that i somehow only fixed bare feet before.

PR #18561, opened 2025-09-25
Luckier 7 event map content + Card deck init improvement/bugfixing

Tenryumi:
## About The Pull Request

These are the code changes necessary to make @TheToaster98 's event map (for the event TOMORROW!!) actually work!
Worth noting is that the audio I used for the casino ambience uses Creative Commons Zero licensing, but I have credited the audio source anyway in the metadata of said audio.

:cl: Ryumi
code: Greatly cleaned up the initialization code of playing card decks.
fix: Fixed triple card decks generating at quadruple size instead of triple, as well as casino card decks spawning with jokers.
/:cl:


code/__defines/sound.dm
```diff
@@ -225,3 +225,9 @@
 	'sound/ambience/engineering/engineering3.ogg',\
 	'sound/ambience/atmospherics/atmospherics1.ogg'\
 	)
+
+#define AMBIENCE_CASINO list(\
+	'sound/ambience/casino/casino_1.ogg',\
```
Kashargul, line 232: source to license on each

code/modules/casino/casino_cards.dm
```diff
@@ -47,3 +47,33 @@
...
 	icon = 'icons/obj/playing_cards.dmi'
 	icon_state = "casino_deck"
+
+/obj/item/deck/cards/casino/triple
+	name = "big deck of casino cards"
+	desc = "A triply-sized deck of playing cards from the golden goose casino, comes without a joker card!"
+
+/obj/item/deck/cards/casino/triple/Initialize(mapload)
+	. = ..()
+	var/datum/playingcard/casino/P
```
Kashargul: no more single letter vars

code/modules/casino/casino_cards.dm
```diff
@@ -47,3 +47,33 @@
...
+	for(var/a = 0, a<3, a++)
+		for(var/suit in list("spades","clubs","diamonds","hearts"))
+
+			var/colour
+			if(suit == "spades" || suit == "clubs")
```
Kashargul: switch case

switch(suit)
     if("clubs", "spades")
...

Kashargul (changes requested):

PR #18568, opened 2025-09-26
Makes weaver TGUI

Cameron-The-Raven:
This all works, just needs to be made nicer looking

## About The Pull Request
Makes the Weaver component TGUI instead of having 5 different commands.
dreamseeker_2025-09-26_03-41-45

## Changelog

:cl: Diana
qol: Weaver trait now has a TGUI panel
code: Cleans up some signals pertaining to Weaver.
/:cl:

### DIFF
--- a/code/datums/components/traits/weaver.dm
+++ b/code/datums/components/traits/weaver.dm
@@ -17,22 +17,10 @@
 		return COMPONENT_INCOMPATIBLE
 
 	owner = parent
-	add_verb(owner, /mob/living/proc/check_silk_amount)
-	add_verb(owner, /mob/living/proc/toggle_silk_production)
-	add_verb(owner, /mob/living/proc/weave_structure)
-	add_verb(owner, /mob/living/proc/weave_item)
-	add_verb(owner, /mob/living/proc/set_silk_color)
+	add_verb(owner, /mob/living/proc/weaver_control_panel)
 
 	//Processing
 	RegisterSignal(owner, COMSIG_LIVING_LIFE, PROC_REF(process_component))
-
-	//When procs are used
-	RegisterSignal(owner, COMSIG_CHECK_SILK_AMOUNT, PROC_REF(check_silk_amount))
-	RegisterSignal(owner, COMSIG_WEAVE_STRUCTURE, PROC_REF(weave_structure))
-	RegisterSignal(owner, COMSIG_TOGGLE_SILK_PRODUCTION, PROC_REF(toggle_silk_production))
-	RegisterSignal(owner, COMSIG_WEAVE_ITEM, PROC_REF(weave_item))
-	RegisterSignal(owner, COMSIG_SET_SILK_COLOR, PROC_REF(set_silk_color))
-
 /datum/component/weaver/proc/process_component()
 	if (QDELETED(parent))
 		return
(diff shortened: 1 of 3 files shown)


tgui/packages/tgui/interfaces/WeaverConfig.tsx
```diff
@@ -0,0 +1,152 @@
...
+              <Section fill title="Create Objects">
+                <LabeledList>
+                  <LabeledList.Item label="Items">
+                    <Stack align="center">
+                      <Stack.Item>
+                        <Button
+                          onClick={() => act('weave_item')}
+                          tooltip="Weave an item!"
+                        >
+                          Weave Item
```
Kashargul: could we have the costs directly visible?

name (cost)

code/datums/components/traits/weaver.dm
```diff
@@ -58,105 +41,101 @@
...
+			weave_check(50, /obj/item/clothing/suit/weaversilk_bindings)
+		if("weave_floor")
+			weave_check(25, /obj/effect/weaversilk/floor)
+		if("weave_wall")
+			weave_check(100, /obj/effect/weaversilk/wall)
+		if("weave_nest")
+			weave_check(100, /obj/structure/bed/double/weaversilk_nest)
+		if("weave_trap")
+			weave_check(250, /obj/effect/weaversilk/trap)
+/*
```
Kashargul, line 143: every action needs a return TTRUE on success to rerender the UI or FALSE on failure which don't need rerender.

Kashargul (changes requested):

code/datums/components/traits/weaver.dm
```diff
@@ -58,105 +41,110 @@
...
+				return FALSE
+			silk_color = set_new_color
+			return TRUE
+		if("toggle_silk_production")
+			silk_production = !(silk_production)
+			to_chat(owner, span_info("You are [silk_production ? "now" : "no longer"] producing silk."))
+			return TRUE
+		if("check_silk_amount")
+			to_chat(owner, span_info("Your silk reserves are at [silk_reserve]/[silk_max_reserve]."))
+			return TRUE
```
Kashargul: FALSE ? we only do a chat print, why should we rerender

Cameron-The-Raven: oh, lemme fix that. I unfortunately didn't know enough about the tgui states, pensive

Cameron-The-Raven:
```suggestion
			return FALSE
```

code/datums/components/traits/weaver.dm
```diff
@@ -58,105 +41,110 @@
...
+		if("new_silk_color")
+			var/set_new_color = tgui_color_picker(ui.user, "Select a color you wish your silk to be!", "Color Selector", silk_color)
+			if(!set_new_color)
+				return FALSE
+			silk_color = set_new_color
+			return TRUE
+		if("toggle_silk_production")
+			silk_production = !(silk_production)
+			to_chat(owner, span_info("You are [silk_production ? "now" : "no longer"] producing silk."))
+			return TRUE
```
Cameron-The-Raven:
```suggestion
			return FALSE
```

Kashargul (approved):

PR #18577, opened 2025-09-27
Fixing client eye

Willburd:
## About The Pull Request
Client eye code was god awful, and 15 years of techdebt. Most of it had become completely nonfunctional if it even functioned to begin with. All the new code is either done by adding a remote_view component if you need a SAFE temporary one, or reset_perspective() if you need a view that is not cleared by moving.

Subtypes include:

/datum/component/remote_view is a component designed to avoid all the original pitfalls of client eye code. Moving at all, being deleted, dying, or having something else change your view will delete the component and end the view. Almost everything should be using this. This will be given automatically if the object is tagged with REMOTEVIEW_ON_ENTER, unless it is an item. that uses a much more aggress subtype automatically regardless of the REMOTEVIEW_ON_ENTER flag being used or not.

/datum/component/remote_view/item_zoom is designed to allow an item to control the remote view. This item being moved/dropped/destroyed will also end the view, along with all the conditions of the base component. Used on things like the binoculars or spybug. In the binocular case. it targets the mob itself and offsets the view, but this is not required.

/datum/component/remote_view/mremote_mutation functions exactly the same as the base type, but if the mob it's focused on dies, or you mutate and love mremote, the remote view will also end.

/datum/component/remote_view/viewer_managed is a special case subtype where an object you interact with holds a managed list of viewers, as well as handled look() and unlook() behaviors. like unique machines, overmap shuttles, or anything else downstream where looking through a console changes your view. the list the component and coordinating object share MUST exist before being passed to the component.

/datum/component/remote_view/mob_holding_item is a VERY aggressive version of the component, designed to work around a byond issue. Where mobs that drop objects holding a mob, cause the held mob's view to become stuck to the mob that dropped the item. This component also handles deeply nested objects, such as a pai that could be inside it's card, inside a pda, inside a small box, inside a backpack and dropped by a mob. This unfortunately is require any time a mob enters a held item. Even if the item is not flagged with REMOTEVIEW_ON_ENTER

Testing:
- [x] Examine should not be possible through remote view(port signal flags?)
- [x] TK should not be possible through remote view
- [x] being flushed through disposal should follow the packet and release the view on exit
- [x] spykit pda should allow selecting cameras and viewing them, doing anything or using the pda should release view
- [x] mutating to lose the remote view mutation should end remote view gene power
- [x] entering and leaving /obj/machinery/dna_scannernew
- [x] entering and leaving /obj/machinery/optable
- [x] entering and leaving /obj/machinery/sleeper
- [x] entering and leaving /obj/machinery/bodyscanner
- [x] entering and leaving /obj/machinery/clonepod
- [x] entering and leaving /obj/machinery/atmospherics/unary/cryo_cell
- [x] entering and leaving /obj/machinery/cryopod
- [x] entering and leaving /obj/machinery/recharge_station
- [x] stumbling into and leaving /obj/machinery/suit_cycler
- [x] stumbling into and leaving /obj/machinery/suit_storage_unit
- [x] entering and leaving /obj/machinery/transportpod
- [x] entering and leaving /obj/machinery/vr_sleeper
- [x] entering and leaving /obj/item/mecha_parts/mecha_equipment/tool/passenger
- [x] entering and leaving /obj/item/mecha_parts/mecha_equipment/tool/sleeper
- [x] putting brains into /obj/mecha as MMIs
- [x] item zoom() procs, like binoculars or rifle scopes
- [x] check whatever this is even doing /obj/item/chameleon
- [x] PAI mob inhabit, movement, pickup, etc
- [x] entering and leaving /obj/machinery/implantchair
- [x] entering and leaving /obj/structure/closet
- [x] getting cursed and leaving /obj/structure/closet/statue
- [x] AI cannot flush /obj/structure/toilet
- [x] check disposals stumble into /obj/machinery/disposal
- [x] stuffing mobs into and falling into /obj/machinery/gibber
- [x] EVERYTHING to do with /mob/observer/eye/aiEye and AI gameplay
- [x] EVERYTHING to do with /obj/item/holder , throwing, pickup, passing, eating, etc
- [x] morph superpower view code
- [x] remoteobserve superpower view code
- [x] protean nano_rig_transform()
- [x] entering and leaving /obj/item/dogborg/sleeper
- [x] /mob/Login() sets view correctly
- [x] /obj/machinery/computer/ship/tgui_act behaviors and overmap views
- [x] entering and leaving /obj/machinery/transhuman/resleever
- [x] ethereal_jaunt spell
- [x] mining UAV for the modular laptops
- [x] ventcrawling behaviors
- [x] vore bellies
- [x] entering and leaving /obj/machinery/slime/extractor
- [x] entering and leaving /obj/machinery/xenobio/editor
- [x] entering and leaving /obj/machinery/xenobio2/manualinjector

## Changelog
Fixes client eye behaviors
Implements COMSIG_MOB_RESET_PERSPECTIVE (when a client's mob changes viewing location)
Implements COMSIG_CLIENT_SET_EYE (when the client changes its viewing location, sent from client)
Implements COMSIG_MOB_DNA_MUTATION (when a mob mutates its genes)
Implements COMSIG_MOB_DROPPED_ITEM (when a mob drops an item)
Implements COMSIG_REMOTE_VIEW_CLEAR (clears all mobs remote viewing an object, or using the object to remote view)
reset_view() has been replaced by reset_perspective()
REMOTEVIEW_ON_ENTER has been added for non-item objects that a mob can enter, it will automatically set the client's view to the object, and release it on exit.
many remote views focused on objects now use a component with self contained and safe logic, instead of pinky promising that the object in question will handle the view correctly
remote view gene no longer uses a hardref
remote view items use component to safely drop view
spybug pda is less aggressive about making you choose a camera
Fixes various AI hud issues, multicam buttons were not functional.

Removes set_machine() calls from pda, as it never releases the mob's machine and is a constant issue with remote views. This code should not be needed on tgui objects like the pda anyway, and was is left behind for unclear reasons. If issues appear then the issues should b e resolved without using set_machine. That entire family of procs and the machine var needs to go.

:cl: Will, Diana
code: implements signal COMSIG_MOB_RESET_PERSPECTIVE
code: implements signal COMSIG_CLIENT_SET_EYE
code: implements signal COMSIG_MOB_DNA_MUTATION
code: implements signal COMSIG_MOB_DROPPED_ITEM
code: implements signal COMSIG_REMOTE_VIEW_CLEAR
refactor: rebuilds client eye transfer code entirely
refactor: cleaned up holder drop and destroy code
add: several small pai QOL, ejecting from pda, and escaping vore eggs
code: REMOTEVIEW_ON_ENTER flag for non-item objects
fix: black screen or locked game view when dropped or moved between objects or mob holders
fix: pai hud being non functional
fix: Some AI hud buttons being non functional
fix: You can no longer use binoculars/scopes from inside a vore belly or sleeper
fix: Forcewall spellbook recoil no longer makes you an invisible statue, it just makes you a statue
fix: dropping someone from a holder within a vore belly no longer breaks their screen
fix: dropping pais from a holder no longer breaks their screen
fix: dropping mice from a holder no longer breaks their screen
fix: rare situation where having remote view gene and being put on a surgery table would lock your view to the table
fix: pda breaking most types of remote views by never releasing it's mob machine state
/:cl:


Willburd:
This needs extensive testing, I've not looked full at AI yet either...

Willburd:
mob holder code is cursed

Cameron-The-Raven:
Needs a final review code. Everything tested works.

code/game/objects/items.dm
```diff
@@ -820,91 +819,42 @@ GLOBAL_LIST_EMPTY(blood_overlays_by_type)
...
-/obj/item/var/ignore_visor_zoom_restriction = FALSE
-
+/// For zooming with scope or binoculars. Uses remote_view/item component for disabling when you move or drop the item
 /obj/item/proc/zoom(var/mob/living/M, var/tileoffset = 14,var/viewsize = 9) //tileoffset is client view offset in the direction the user is facing. viewsize is how far out this thing zooms. 7 is normal view
 	SIGNAL_HANDLER
 	if(isliving(usr)) //Always prefer usr if set
 		M = usr
-
+	if(!M.client)
+		return 0
```
Cameron-The-Raven:
```suggestion
		return FALSE
```

code/game/objects/items.dm
```diff
@@ -820,91 +819,42 @@ GLOBAL_LIST_EMPTY(blood_overlays_by_type)
...
 	if(!isliving(M))
 		return 0
```
Cameron-The-Raven:
```suggestion
		return FALSE
```

code/game/objects/items.dm
```diff
@@ -820,91 +819,42 @@ GLOBAL_LIST_EMPTY(blood_overlays_by_type)
...
-
+	if(isbelly(M.loc) || istype(M.loc,/obj/item/dogborg/sleeper))
+		return 0
```
Cameron-The-Raven:
```suggestion
		return FALSE
```

code/game/objects/items.dm
```diff
@@ -820,91 +819,42 @@ GLOBAL_LIST_EMPTY(blood_overlays_by_type)
...
+	if(M.is_remote_viewing())
+		to_chat(M, span_warning("You are too distracted to do that."))
+		return 0
```
Cameron-The-Raven:
```suggestion
		return FALSE
```

PR #18601, opened 2025-10-07
Don't slip phased shadekin

Willburd:
## About The Pull Request
Mar!

## Changelog
Prevents phased shadekin from slipping on wet or icey turfs.

:cl: Will
fix: shadekin no longer slip on wet or icey turfs while phased
/:cl:

### DIFF
--- a/code/datums/components/turfslip.dm
+++ b/code/datums/components/turfslip.dm
@@ -86,6 +86,8 @@
 /turf/simulated/check_slipping(var/mob/living/M,var/dirtslip)
 	if(M.buckled)
 		return FALSE
+	if(M.is_incorporeal()) // Mar!
+		return FALSE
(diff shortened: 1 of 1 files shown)


code/datums/components/turfslip.dm
```diff
@@ -86,6 +86,8 @@
 /turf/simulated/check_slipping(var/mob/living/M,var/dirtslip)
 	if(M.buckled)
 		return FALSE
+	if(M.is_incorporeal()) // Mar!
```
Kashargul, line 89: And there I thought that was intentionally missing

Willburd: While it is very funny to slip phased kin, it is unintended. The mars are spared the clown's wrath.

PR #18612, opened 2025-10-08
methane in gas sensor console

Willburd:
## About The Pull Request
This console is really old and seems to mostly be used in phoronics on some maps, and thus was missed in the methane PR.

## Changelog
Adds methane gas levels to gas sensor console.
Renames phoron from TX(toxin) to PH like all other gas consoles.

:cl: Will
fix: methane missing from gas scanner console in phoronics/toxins
/:cl:


tgui/packages/tgui/interfaces/GeneralAtmoControl/Sensors.tsx
```diff
@@ -56,7 +63,8 @@ const AtmoSensor = (props: { sensor: sensor }) => {
           {carbon_dioxide ? (
             <Stack.Item>({carbon_dioxide}% CO²)</Stack.Item>
           ) : null}
-          {phoron ? <Stack.Item>({phoron}% TX)</Stack.Item> : null}
+          {phoron ? <Stack.Item>({phoron}% PH)</Stack.Item> : null}
+          {methane ? <Stack.Item>({methane}% CH₄)</Stack.Item> : null}
```
Kashargul, line 67: When doing scientific notation on methane, can you also fix CO2, N2, O2 above? I really prefer consitency

Willburd: fixed to make all entries use subscript instead of superscript

Willburd: These really need an element for gasses, these are spread out over multiple sensors and displayed in different ways.

PR #18621, opened 2025-10-12
Drag Fixes

Cameron-The-Raven:
## About The Pull Request
Fixes some aspects about dragging:
If there is a risk of worsening damage via dragging, it will now ALWAYS inform you instead of having a 70% chance of not informing you (pull_can_damage)
Dragging now properly checks its parent proc first instead of doing the same check twice. This resulted in some cases where one would trigger (bleeding) but then the other wouldn't (the message that 'the person is being dragged' and the actual damage).

Adds it so there is a runechat message when someone is dragged, just to provide visual feedback without reading the chat (Since in combat, you might not even realize/have time to look at chat)
## Changelog
:cl: Diana
fix: Dragging now will always warn you if it will damage the person you're dragging
fix: Dragging now properly tell you if the person being dragged is damaged.
fix: Fixes issue where you would bleed via drag when you had no bleeding wounds / broken bones
qol: Dragging will provide a visual message on screen if the person is being injured, as opposed to just chat.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human.dm
+++ b/code/modules/mob/living/carbon/human/human.dm
@@ -1822,19 +1822,16 @@
 // Drag damage is handled in a parent
 /mob/living/carbon/human/dragged(var/mob/living/dragger, var/oldloc, trigged_bleeding)
 	if(..())
-		var/bloodtrail = TRUE
 		if(species?.flags & NO_BLOOD)
-			bloodtrail = FALSE
-		else
-			var/blood_volume = vessel.get_reagent_amount(REAGENT_ID_BLOOD)
-			if(blood_volume < species?.blood_volume*species?.blood_level_fatal)
-				bloodtrail = FALSE	//Most of it's gone already, just leave it be
-			else
-				remove_blood(1)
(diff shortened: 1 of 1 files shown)


KillianKirilenko:
I would suggest using a larger message size for the `pull_can_damage` as well, similar to how getting held up at gunpoint gives you a massive warning. Whilst a single warning line is pretty obvious, it might still be easy to miss if a lot is going on at once.

Otherwise, very good.

Cameron-The-Raven:
> I would suggest using a larger message size for the `pull_can_damage` as well, similar to how getting held up at gunpoint gives you a massive warning. Whilst a single warning line is pretty obvious, it might still be easy to miss if a lot is going on at once.
> 
> Otherwise, very good.

Went ahead and upped the size, it's more noticeable in chat now!

dreamseeker_2025-10-12_03-34-12

KillianKirilenko:
> Went ahead and upped the size, it's more noticeable in chat now!
> dreamseeker_2025-10-12_03-34-12

That ought to do nicely, aye.

PR #18628, opened 2025-10-13
Removes Gender

TheCaramelion:
## About The Pull Request

Removes /datum/gender and replaces them with helper procs

## Changelog
:cl: Guti
refactor: Refactored pronouns handling
del: Removed /datum/gender
code: Added pronoun helper procs
/:cl:

### DIFF
--- a/code/_helpers/gender.dm
+++ b/code/_helpers/gender.dm
@@ -82,7 +82,7 @@
 	/// otherwise we go off the gender of our object
 	var/gender
 	if(targeted_gender)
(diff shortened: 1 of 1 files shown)


ShadowLarkens:
I'm not sure this is an improvement especially given that it removes the herm gender. This will also break anyone using this gender as it is available in character setup.

TheCaramelion:
> I'm not sure this is an improvement especially given that it removes the herm gender. This will also break anyone using this gender as it is available in character setup.

I forgor :(

It's back now ✨

ShadowLarkens:
👍 

I think this is a good solution for now as we can always change the datum procs for custom gender identity easily in the future.

PR #18647, opened 2025-10-18
Removes Armorsoak

ReoDaProtovali:
## About The Pull Request

Originally, this was going to be part of a larger pull request, but I've decided it'd be easier to do stuff if I did it in sections, so im starting with the lowest, easiest to start section. That, and this system is barely used except to make stuff much, much more tankier than it ever should be. (mostly simplemobs)
Also, a lot of proccalls kind of just had to manually declare params past armorsoak because like half of them didnt use it but did use parameters after it lol

Next up, assuming I dont get burnt out, or this gets denied will be porting armor datums, and possibly doing an overview of a lot of armor values while im converting stuff
Bye_Armorsoak

## Changelog
:cl:
del: Removed Armorsoak. no more flat damage reduction after percentage damage reduction
qol: made a lot of basic simplemobs less tanky as a side effect, I guess? A lot of mobs had armorsoak when they really shouldn't have.
balance: probably affects combat a bit, atleast for exploration, some mobs will be easier to kill, while padded exploration suits wont protect as much.
/:cl:


ReoDaProtovali:
I would like to mention one of the worst parts of making this was saying "soak" so many times in my head that it started to sound like an abstract sound

Willburd:
As a downstream opinion from a more combat oriented server. The only major change here is that bosses will now receive chip damage. Raising armor percents will make boss mobs tankier to correct for it. The current code implementation of armor soak is really just for adding a minimum damage to incoming weapons against huge mobs. This can be done in a better way that is limited to unique mobs and their AI state (such as rage/defense mode/etc). 

I have a potential component/element in mind for those. Otherwise this PR shouldn't break things too much for explo mobs, but I will begin working on that after this PR. I am not requesting changes to this one.

ReoDaProtovali:
Does not actually permit chip damage as it already permits it. Armorsoak can only stop at most 80% of damage. It's just a really shitty system that makes tanky mobs even tankier at worst, and weak mobs feel like they're ignoring a disproportionate amount of damage unless you're using an overpowered weapon. Arguably, bosses generally shouldnt be using an absurd amount of armor anyways

PR #18608, opened 2025-10-07
Unified Ghost Hole

KillianKirilenko:
## About The Pull Request
A long overdue refactor to the Maintenance Spawner Holes that replaces the three different options with a singular interactable object, so that the three events no longer need to fight each other for their spawn chances.

The new Unified Hole spawner event has a chance equal to the combined chances of the maintenance predator and morphspawner pre-lurker addition. This means they might be slightly less common overall, but this can always be tweaked if desired.

Some phrasing on critter/morph spawning has been revised so that it's not so aggressively "you are a pred", for people who just wanna play li'l goobers or even be prey.

The original holes still exist in code and are otherwise untouched, so mappers can continue to use them (esp. Redgate Inhabitant spawners) as much as they like. The original events also still exist so they can be manually queued up in the event manager (I think) if admins want.

- [x] basic functionality
- [x] replace events
- [x] spelling/grammar pass
- [x] set up ghost query
- [x] new spawner icon

## Changelog

:cl:
refactor: unified the mob, morph, and lurker spawns into a single multipurpose hole that can be used to spawn any one of them based on the interacting ghost's choice. the associated seperate events have been replaced with a single event that uses the new unified spawner, so that the events are no longer competing for spawn chances. the new spawner also has a high-vis icon so that it won't be difficult for ghosts to spot if they're just scrolling around with darkness enabled.
/:cl:

PR #18652, opened 2025-10-18
Revert "Unified Ghost Hole"

Kashargul:
Reverts VOREStation/VOREStation#18608

Please don't merge when tgui actions are altered without making sure the UI is fixed and adjusted as well.


KillianKirilenko:
Why go to all the trouble of reverting when it'd be less of a hassle to just PR a fix to the UI? That way the whole thing doesn't need to be re-PR'd later.

PR #18674, opened 2025-10-21
various downstream fixes for maps

zeskorion:
## About The Pull Request
-Adds a var to overhangs that allows them to be deconstructed like railings
-adds a dug_icon to the outdoors turf file to allow mineral turfs to use icons therein without icon errors
## Changelog
Nothing player facing


maps/redgate/falls/atoll_objs.dm
```diff
@@ -19,10 +19,11 @@
 	icon = 'maps/redgate/falls/icons/objs/bronze_overhang.dmi'
 	icon_modifier = "bronze_"
 	icon_state = "bronze_railing0"
+	var/interactable = FALSE
 
 //Escape most railing interactions besides tackling people over them
 /obj/structure/railing/overhang/attackby(obj/item/W)
-	if(!istype(W, /obj/item/grab))
+	if(!istype(W, /obj/item/grab) && !interactable)
```
Cameron-The-Raven, line 26: Just had to check the logic here:

Interactable:
Grab does  ..()
Item does : ..()

Not-interactable:
Grab does ..()
Item does return

Seems good.

zeskorion: Yep! Pretty much, and was tested

PR #18685, opened 2025-10-24
Fix hypo injector being unable to inject common injectables.

NovaDusklight:
## About The Pull Request

Fixes the injector in circuitry (the hypo injector) being unable to inject common injectable items. Being restricted to living targets.

Injecting living targets was already the most powerful thing it could inject, so this is not too drastic a chance. It is being given the same rules as a syringe, and only allowed to inject the common syringe targetable objects.

This is usually better done with a normal syringe, since circuits are slow and need to be coded to inject a single item by ref, but can be helpful for rare cases, like working with proto-organs, which need immense amounts of injected reagents. or working in xenobio, where you might want to make a machine to speed up your workflow. 

## Changelog

:cl: Zizzi
fix: Circuitry injector can now inject common injectables.
/:cl:

### DIFF
--- a/code/modules/reagents/reagents/_helpers.dm
+++ b/code/modules/reagents/reagents/_helpers.dm
@@ -8,14 +8,8 @@
 	return TRUE
 
 /obj/can_be_injected_by(var/atom/injector)
-	// Check basic injection requirements first
-	if(!Adjacent(get_turf(injector)))
+	if(!..())
 		return FALSE
-	if(!reagents)
-		return FALSE
-	if(!reagents.get_free_space())
-		return FALSE
-
 	// Then check if this is a type of container that can be injected
 	return is_injectable_container()
 


code/modules/integrated_electronics/subtypes/reagents.dm
```diff
@@ -134,22 +134,40 @@
...
+			// Safety checks for valid injectables.
+			if(!AM.reagents)
+				activate_pin(3)
+				return
+			if(!AM.reagents.get_free_space())
+				activate_pin(3)
+				return
+
+			// Allow injection only for basic syringe-compatible objects.
+			if(!AM.is_open_container() && !istype(AM, /obj/item/reagent_containers/food) && !istype(AM, /obj/item/slime_extract) && !istype(AM, /obj/item/clothing/mask/smokable/cigarette))
```
Kashargul: This heavily looks like copy pasted from syringes in parts? Should be generalised in a define and not copied about

NovaDusklight: It isn't copy pasted, but I totally agree with you that this should be handled by an injection helper. 

I'll work on making a helper handle this. Will just take some testing, since it will effect the syringe as well.

NovaDusklight: @Kashargul I've set up an injection helper, and tested that it works reliable on containers, mobs, and common injectables for both the hypo injector, and syringe using the helper. 
Is this looking better to you? I think this would be much nicer to maintain, and use if something else injects in future.

code/modules/reagents/reagents/_helpers.dm
```diff
@@ -8,7 +8,24 @@
 	return TRUE
 
 /obj/can_be_injected_by(var/atom/injector)
-	return is_open_container() && ..()
+	// Check basic injection requirements first
```
Kashargul: what's with the parent call here?

NovaDusklight: At first, I thought this wasn't being used, but I see now it usually handles checking the checks I added here.
This wasn't necessary for the hypo injector, it just needs a helper for if an object is a valid injectable, which other things can use. I'll change this right after testing.

Kashargul: Please don't duplicate code and call the parent instead

Kashargul (approved):

PR #18692, opened 2025-10-26
Gets rid of hasorgans() helper proc

Cameron-The-Raven:
## About The Pull Request
This was just ishuman() but with a different coat of paint.
This made it confusing because it implies it's checking for organs of some sort when, no, it's just ishuman()
## Changelog
:cl: Diana
code: removes deprecated organ check 
/:cl:


Willburd:
Should also look into the isOrgan() helper, that only checks for organ/external

PR #18706, opened 2025-10-28
unified conflict check for traits and genes

Willburd:
## About The Pull Request
Fixing an old copypasta problem I created back when making traitgenes. This unifies the conflict check between traits and genetraits... No player facing changes as there is no actual issues with this yet. I'm just making this into a non-issue in the future.

## Changelog
Traitgenes and traits use unified code when checking for conflicts.

:cl: Will
refactor: traitgenes and traits share conflict checking code
/:cl:


Willburd:
CI needs a kick because tether is active edging.

PR #18707, opened 2025-10-28
Disposals delete remains

SatinIsle:
## About The Pull Request

Disposal chutes now delete any digestion remains when they flush, making them a safe place to throw them away.

As for why we want this option: New players often do not know how to handle digestion remains, not knowing you can destroy them in-hand. Understandably, they just bin them, but this gets them into trouble with exposing others to messy skulls. Alternatively, it also allows players to spice up their scenes a bit by just dumping skulls casually without the risk of squicking anyone!

## Changelog
:cl:
qol: Disposal chutes now delete any digestion remains when they flush, making them a safe place to throw them away.
/:cl:

### DIFF
--- a/code/modules/recycling/disposal_machines.dm
+++ b/code/modules/recycling/disposal_machines.dm
@@ -382,9 +382,8 @@
 
 	// We don't ever want digestion remains going through disposals, but people understandably thing they're doing right by trashing them
 	// So let's just delete them instead!
-	for(var/obj/item/flushed_item in src)
-		if(istype(flushed_item, /obj/item/digestion_remains))
-			qdel(flushed_item)
+	for(var/obj/item/digestion_remains/flushed_item in src)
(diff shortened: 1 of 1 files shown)


SatinIsle:
> You currently have a negative Fix/Feature pull request delta of -17. Maintainers may close this PR at will. Fixing issues or improving the codebase will improve this score.

You hate me

code/modules/recycling/disposal_machines.dm
```diff
@@ -380,6 +380,12 @@
 	if(flushing)
 		return
 
+	// We don't ever want digestion remains going through disposals, but people understandably thing they're doing right by trashing them
+	// So let's just delete them instead!
+	for(var/obj/item/flushed_item in src)
```
Kashargul: can we do a typed loop right away? instead of an istype check below?

PR #18710, opened 2025-10-29
Smes Prefab Alterations

Willburd:
## About The Pull Request
Upports various smes prefabs used downstream, and a requested engineering tweak.

## Changelog
Adds multiple poi and shuttle smes prefabs
Sets engine default smes to have 50% charge instead of full charge

:cl: Will
code: Smes prefab tweaks
/:cl:

### DIFF
--- a/code/modules/power/smes_prefabs.dm
+++ b/code/modules/power/smes_prefabs.dm
@@ -6,7 +6,7 @@
 /// Defaults to 100% input and output settings, starts with maximum charge by default
 /obj/machinery/power/smes/buildable/engine_default/apply_mapped_settings()
 	// Set charge
-	charge = capacity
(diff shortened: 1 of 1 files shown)


TankTheBirb:
Engine SMES being at full charge was intentional for Virgo iirc. Basically just gives people some margin for error when setting up stuff like Tesla containment after no engineer shows up for a deadshift or similar, and without having to wait for a pacman to give it any amount of charge.

Willburd:
> Engine SMES being at full charge was intentional for Virgo iirc. Basically just gives people some margin for error when setting up stuff like Tesla containment after no engineer shows up for a deadshift or similar, and without having to wait for a pacman to give it any amount of charge.

Was requested downstream to half it to allow for discharge and coil upgrade. I can just make that change an edit down there

PR #18713, opened 2025-10-30
Fix Synth Instruments

Willburd:
## About The Pull Request
Non-legacy instruments were not playing because of a single comma

## Changelog
Corrected sound arguments for synthed instruments, the sound file is actually passed to the sound argument, instead of if the sound is affected by pressure argument.

:cl: Will
fix: non-legacy instruments should now play music
/:cl:


SaintLuther:
Also tested and functionality is restored :)

PR #18714, opened 2025-10-31
Relaymove Moved To Remote Views

Willburd:
## About The Pull Request
Moves relaymove() interaction from machine code to remote views, now that machine var code is being removed. This intends to cleanup and streamline relayed movement for things like the uav and overmap shuttle. It also makes it extremely easy to do remote views with custom overlays like the uav does. Also fixes a personal pet peeve about custom overlays taking an entire life tick to appear on screen.

Most objects that use relaymove() are not directly affected, as being inside an object and moving inherently calls relaymove(). This only applies to objects that are remote controlled through another object, uav and overmap for direct examples. However there is nothing stopping you from using direct views to relay moves too.

I moved all remote view settings into overridable datums that are passed to the remote view during AddComponent(). This gives the remote view various options about what signals to subscribe to, as well as allowing bespoke logic to link relaymove() and vision code updates to the object you are controlling... Once again, uav and overmap shuttle for examples. Also cameras, which need to set vision flags if they have xray upgrades.

Also moves the hud health update into it's own proc that's existed for years but was never used, and attached it to a signal so components will be able to control the health display if they need to somehow. Added a signal to darksight hud code cause it was in about the same place. Allows components to change the darksight values before the rest of the code runs. Both of these have been added as option signals for remote view.

## Changelog
Adds a configuration datum for remote views
Adds relaymove support to remote views
Implements COMSIG_MOB_RELAY_MOVEMENT
Implements COMSIG_LIVING_HANDLE_HUD_HEALTH_ICON
Implements COMSIG_LIVING_HANDLE_HUD_DARKSIGHT
Updates remote views to use the new code if they need to forward inputs
Removed set_machine() calls on fixed objects
Fixes MMIs removed from borgs having their view desync
Fixes a runtime between remoteviews and machine var components fighting over view ending. Now only remoteviews handle that code entirely.
Fixed assert fail if your ghost qdels fast enough after vore, before the remote view from belly can decouple.
Restored REMOTEVIEW_ON_ENTER on several medical machines, as prey in bellies would have their views desync when their pred leaves them.

:cl: Will
fix: remote views with decorations and sight changing flags apply instantly
fix: weird view behavior when cancelling remote view while using the spy-kit pda with only a single camera linked
fix: MMIs removed from borgs and fbps no longer have their view desync
fix: remote view runtime from interrupted views in the machine var shim
fix: prey views becoming locked to several medical machines if their pred left them
refactor: Moved relaymove() from machine var to remote_view component
code: Implements COMSIG_MOB_RELAY_MOVEMENT
code: COMSIG_LIVING_HANDLE_HUD_HEALTH_ICON
code: COMSIG_LIVING_HANDLE_HUD_DARKSIGHT
/:cl:

### DIFF
--- a/code/modules/tgs/README.md
+++ b/code/modules/tgs/README.md
@@ -5,9 +5,9 @@ This folder should be placed on its own inside a codebase that wishes to use the
 - [includes.dm](./includes.dm) is the file that should be included by DM code, it handles including the rest.
 - The [core](./core) folder includes all code not directly part of any API version.
 - The other versioned folders contain code for the different DMAPI versions.
-  - [v3210](./v3210) contains the final TGS3 API.
-  - [v4](./v4) is the legacy DMAPI 4 (Used in TGS 4.0.X versions).
-  - [v5](./v5) is the current DMAPI version used by TGS >=4.1.
+    - [v3210](./v3210) contains the final TGS3 API.
+    - [v4](./v4) is the legacy DMAPI 4 (Used in TGS 4.0.X versions).
+    - [v5](./v5) is the current DMAPI version used by TGS >=4.1.
 - [LICENSE](./LICENSE) is the MIT license for the DMAPI.
 
 APIs communicate with TGS in two ways. All versions implement TGS -> DM communication using /world/Topic. DM -> TGS communication, called the bridge method, is different for each version.
--- a/code/modules/tgs/core/README.md
+++ b/code/modules/tgs/core/README.md
@@ -2,7 +2,7 @@
 
 This folder contains all DMAPI code not directly involved in an API.
 
-- [\_definitions.dm](./definitions.dm) contains defines needed across DMAPI internals.
+- [_definitions.dm](./definitions.dm) contains defines needed across DMAPI internals.
 - [byond_world_export.dm](./byond_world_export.dm) contains the default `/datum/tgs_http_handler` implementation which uses `world.Export()`.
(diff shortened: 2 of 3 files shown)


Willburd:
This should be all handled now. A second look over is always welcome, but it looks like it should all be functioning with no runtimes. I also fixed the runtime issue caused by the machine shim component's interaction with remote views.

Willburd:
Reo gave me some additional reports about machines that mess with prey views if you enter them. Originally i removed these machines remoteview flags, but it appears it will be needed to handle prey also disconnecting their views from stuff like the bodyscanner when their pred leaves it. Fixed in above commits.

code/datums/components/remote_view.dm
```diff
@@ -151,14 +209,56 @@
...
+	SIGNAL_HANDLER
+	SHOULD_NOT_OVERRIDE(TRUE)
+	RETURN_TYPE(null)
+	PRIVATE_PROC(TRUE)
+	settings.handle_hud_darkvision(src, host_mob)
+
+/datum/component/remote_view/proc/handle_mob_vision_update(datum/source)
+	SHOULD_NOT_OVERRIDE(TRUE)
+	PRIVATE_PROC(TRUE)
+	SIGNAL_HANDLER
```
Kashargul: move to top please

Willburd: corrected

PR #18716, opened 2025-10-31
Fusion Coils

KillianKirilenko:
## About The Pull Request

A little project I've had cooking in the back of my head for a few weeks now, these single-shot/disposable battery units can be plugged into an SMES to recharge it. They're a one-shot, all-or-nothing affair, incapable of overcharging an SMES past its capacity, and once they're spent there's no way to recharge them. A single fusion coil is enough to juice the Engine SMES up by 80%, or the Main SMES up by 20%.

Two fusion coils can be found in each Engineering Workshop. Additional coils can be acquired from Cargo for 150 points per 2 coils.

Icons are far from final - I'd like to revisit them at some point, maybe make it a bit glowy in its charged state. But they'll suffice for now. I might make them rechargeable later using a special machine, I'm not sure.

https://github.com/user-attachments/assets/f5269e5f-0355-46ef-83b6-4eeabad68166

## Changelog

:cl:
add: Added fusion coils, special single-use cells which can be used to kickstart depleted SMES units
/:cl:


Kashargul:
Considering that this allows to bypass the need for any engine setup... I'm not too sure those shouldn't require to be charged beforehand, printed / ordered empty and only 2 charged emergency coils exist.

=> Requires discussion

KillianKirilenko:
In the interest of being scientific: assuming Main is left with a standard 1 MW output (its maximum), Tether burns 1 coil every 4 minutes and 50 seconds, give or take. That's roughly 12.5 coils per hour, at a cost of 937.5-ish supply points per hour. It can only hold 5 coils of charge at a time without modification, requiring that it be checked roughly every 25 minutes.

If you know how to modify the SMES and/or are accessing engineering that often, just set up the damn engine.

Some quick napkin math suggests something on the order of ~5,250 supply points to power an entire shift w/ coils if we account for initial charge and give some wiggle room.

Issue #18698, opened 2025-10-27
[GENERAL] Capsules lose gravity when the station (Or at least Stellar Delight) does.


NickBelmont:
### Brief description of the bug

When the gravity generator fails, capsules also lose gravity (Presumably all, have only seen it happen with the sauna capsule so far)

### What you expected to happen

Capsules not on the Stellar Delight to not lose gravity

### What actually happened

Capsule had no gravity when the Stellar Delight lost gravity, despite being in a redgate.

### Reproduction steps

```bash
1. Deploy a capsule in a redgate or other offsite map
2. Gravity generator anomaly event fires
3. Capsule loses gravity for the event duration
...
```

PR #18742, opened 2025-11-06
Fixes gravity surge event affecting capsule shelters on maps without gravity generators

Tenryumi:
## About The Pull Request
Fixes https://github.com/VOREStation/VOREStation/issues/18698. 
Areas now have a flag that tells the game it should always have gravity, which prevents them from being affected by gravity surge events. This flag has also been given to all capsule areas!

## Changelog
:cl:
fix: Fixed shelter capsules being affected by gravity surge events on maps that don't have gravity generators.
/:cl:

### DIFF
--- a/code/game/area/areas.dm
+++ b/code/game/area/areas.dm
@@ -40,7 +40,6 @@ GLOBAL_LIST_EMPTY(areas_by_type)
 
 	var/music = null
 	var/has_gravity = TRUE // Don't check this var directly; use get_gravity() instead
-	var/ignores_gravity_events = FALSE // Whether gravity-altering events can set the gravity of this area.
 	var/obj/machinery/power/apc/apc = null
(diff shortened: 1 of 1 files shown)


Tenryumi:
Why did the bot remove the fix label...

Kashargul:
What happens if a capsule is removed? How does the effect ship capsules?

Tenryumi:
> What happens if a capsule is removed?

That depends on whether or not the capsule's area is removed with it. As it stands, there is no way to conveniently remove capsules. Because they use shuttle walls, they can't even be deconstructed. It'd require admin intervention.

>How does the effect ship capsules?

I believe those are admin-spawn only? But this won't affect them since they use a different area path altogether. Also, after trying to spawn one in, the server crashed. I tested this on the main branch as well, and it still crashed.

PR #18743, opened 2025-11-07
Revamps flash code

Cameron-The-Raven:
## About The Pull Request
Makes flash code less crungo by making it a proc that returns true/false instead of a giant if() chain.

Makes flashes be able to be repaired by default. It seems like this feature was disabled due to a 'compile error' (????) but it doesn't happen anymore, so we can let it back.
(Note: This was apparently the case, BUT the disable was bugged due to a faulty do_after check in the past...My do_after rework fixed it, but flashes were able to be repaired until then...So it was 'disabled' due to a compile error, but not really disabled, but my TG do_after port disabled it...Yeah, confusing.)

Also makes it so the variables in the flash are autodoc'd in VSC.

Fixes a bug where species with a flash mod of 2x was instead having a flash mod of 4x for flashes.
## Changelog
:cl: Diana
fix: Fixing flashes no longer cause compile errors, meaning they can be fixed again.
fix: Flashes no longer do 2x the amount of halloss when affecting species that are weak to flashes. Instead multiplies the halloss by the appropriate amount.
code: Makes the flash code a little higher quality.
/:cl:


Cameron-The-Raven:
Tested and works

code/game/objects/items/devices/flash.dm
```diff
@@ -169,50 +171,6 @@
...
-				var/obj/item/borg/combat/shield/shield = locate() in R
-				if(shield)
-					if(shield.active)
-						shield.adjust_flash_count(R, 1)
-						flashfail = 1
-	else
-		flashfail = 1
 
 	if(isrobot(user))
 		spawn(0)
```
Kashargul: why do we still need this?

PR #18760, opened 2025-11-09
Cortical Borer Fixes

Willburd:
## About The Pull Request
Upports a giant pile of cortical borer bug and gameplay fixes from outpost21. The intent for this is to eventually rework borers into something more pref friendly for upstream use. Basically rebranding borers as "head friends" for rp servers, while still having antag options downstream where they are used as hostile creatures.

This pr is primarily upported fixes, and does not correct a major issue in their code. They still use very old mind hostage code, as well as manual IP changing. This needs to be updated to use vorecode's captured minds. It is not included in this PR as that is a significant change that will likely need to be it's own pr for organization and review reasons.

Also fixes a minor motion tracker subscription bug.

## Changelog
There are many changes to borers in this pr, and many were done at least a year ago on outpost.
-Cortical borer target selection will auto-select the first target, if a single target is the only one present. Instead of showing a target list with a single entry.
-Cortical borer psychic attack range raised from 3 to 5.
-Chemical list has been significantly updated, it now shows the chemical name and what it will do.
-Borers can now jumpstart their host's heart similar to ling revival. However this does not rejuvenate the body, and instead injects a mess of healing chems and clear some damage. Bodys with extreme damage may not be possible to revive.
-Headwear no longer blocks borer entry. This was done due to meta-abuse of the headwear mechanic basically making it impossible for a borer to jump hosts once they are discovered among the crew, and also the crew wearing any funny helmet they could find the moment a borer is known.
-Borer chemical weakness mechanic actually functions. Due to reagent code changes, it was looking at the mob's skin instead of bloodstream or stomach for it's secret weakness.
-Borers assigned through the traitor panel no longer cause a ghost spawn notice, delaying the player's change into the borer, and likely resulting in other strange issues if a ghost accepts the invite.
-There's also too many other logic fixes to list. Mostly in regards to tgui menus needing additional checks after closing them, due to their async nature. Many old !src checks updated to QDELTED(src), spawn()s updated to addtimer().
-Added motion sensing to cortical borers
-Fixes an accidental subscribe, instead of unsubscribe when a motion tracker is destroyed while in a mob's hands.
-Borer paralyses attack turned into a knockout, as players would scream over radio instantly, not letting the borer have any time to communicate intentions or threats if antag

:cl: Will
qol: Cortical borer target selection will instantly select single targets instead of always showing the target list
qol: Cortical borers use changeling chemical hud
balance: Cortical borer psychic attack range raised from 3 to 5
balance: Cortical borer chemicals updated,and made more clear to the player selecting them
balance: Cortical borers will only show on upgraded health analyzers, does not apply to bodyscanner
balance: Cortical borers are faster at entering vents
balance: Headwear no longer prevents borer infestation entirely, and instead significantly delays their entry
balance: Cortical borers can now spend a large sum of chemicals to jumpstart their host's heart like a defib, and slowly heal them.
balance: Cortical borers have 80 chems max, lowered ability costs to compensate
balance: Cortical borers require 5 chems to torture a captured host
balance: Cortical borer brain damage chance lowered from 5 to 2 while the borer is in control
balance: Cortical borers have motion tracking like teshari
balance: Cortical borer paralyze replaced with knockout 
fix: Cortical borers assigned by admins in the traitor panel no longer attempt to ghost invite when changing a player into a borer
fix: Cortical borer weakness mechanics respect modern reagent code
fix: Motion tracker properly unsubs the mob that holds it if qdeleted in hand
fix: Admin logging added to borer attacks and abilities
refactor: Many logic and delay code issues in cortical borers, moved spawn() to timers.
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_mob/subtypes/animal/borer/borer.dm
+++ b/code/modules/mob/living/simple_mob/subtypes/animal/borer/borer.dm
@@ -123,8 +123,9 @@
 		chemicals += 0.25
 	else
 		chemicals += 0.1
-	if(FLOOR(chemicals/10,1) > FLOOR(chem_before/10,1))
-		to_chat(host, span_alien("Your chemicals have increased to [FLOOR(chemicals/10,1) * 10]"))
+	var/new_chem = FLOOR(chemicals/10,1)
+	if(new_chem > FLOOR(chem_before/10,1))
+		to_chat(host, span_alien("Your chemicals have increased to [new_chem * 10]"))
 
 /mob/living/simple_mob/animal/borer/proc/handle_docile()
 	if(stat == DEAD)


code/modules/mob/living/simple_mob/subtypes/animal/borer/borer_powers.dm
```diff
@@ -82,29 +89,33 @@
...
-	to_chat(M, "Something slimy begins probing at the opening of your ear canal...")
+	if(!protected)
+		to_chat(M, span_vdanger("Something slimy begins probing at the opening of your ear canal..."))
+	else
+		to_chat(M, span_vdanger("Something slimy begins trying to find a way past your helmet..."))
 	to_chat(src, span_warning("You slither up [M] and begin probing at their ear canal..."))
 
-	if(!do_after(src, 3 SECONDS, target = M))
-		to_chat(src, span_warning("As [M] moves away, you are dislodged and fall to the ground."))
+	if(!do_after(src,entering_timer, target = M))
```
Cameron-The-Raven:
```suggestion
	if(!do_after(src, entering_timer, target = M))
```

Willburd:
Functional for initial PR, but these are probably still not usable upstream.

code/modules/mob/living/simple_mob/subtypes/animal/borer/borer.dm
```diff
@@ -53,78 +75,172 @@
...
+		return
+	if(chemicals >= BORER_MAX_CHEMS || docile)
+		return
+
+	var/chem_before = chemicals
+	if(controlling)
+		chemicals += 0.25
+	else
+		chemicals += 0.1
+	if(FLOOR(chemicals/10,1) > FLOOR(chem_before/10,1))
```
Kashargul: why do we floor twice and don't cache?

Willburd: This is for announcing to the borer every 10 chems gained (you cannot see the borer hud while in control of a host). This code stores the chems before the chem generation is applied, then compares both chems before and after, divided by 10, and floored. If the newly genned chems / 10 is above the previous chems amount / 10, both floored. it will show a message showing your current chems. tl;dr if the old chems is 19, and new chems is 20, it will become 2 > 1 and show message.

PR #18773, opened 2025-11-11
topmost turf crossed component

Willburd:
## About The Pull Request
It is very difficult to forward crossed() actions, for example if you want something in a bag or vorebelly to react to something crossing the turf it is ultimately in. A blunt example would be a bikehorn honking when you cross it, even if the horn is inside a backpack, or 12 vore bellies deep. This component will forward all Cross() calls to the Crossed() proc of the atom this component is applied to.

The bikehorn of testing is in the component's file as both an example and test code.

The intent of this component is to have the object that adds the component to handle all it's logic in Crossed() itself, and solving what should affect it or not. Probably needing to recursively check upward for parent chain stuff if it needs to do any weird exclusions... It's a lot better than needing objects to recursively search downward into the contents of everything in a turf to do this the other direction.

## Changelog
Adds a component that allows forwarding the Cross() signal of a turf to an object, regardless of how deep in nested storage the item is.

Fixes a fundamental issue with the recursive move component. If the recursive tagged object is inside a bag on a turf, and a mob picks up the bag, the hierarchy will not be updated. This is because the signal the component registers is only applied to atom/movables in it's location chain. It does not subscribe the turf it is on to it. For obvious reasons.

This means when the bag exits the turf it is on, it will not inform the recursive move component that the bag was moved, and therefor the item trying to maintain recursive move information will no longer have any updates beyond the bag, meaning it will never fire if the mob moves.

I added a COMSIG_ITEM_EQUIPPED signal to the loc signals in the recursive move component. This means that even though the exit signal fails from the turf, you will still get a rebuild when it finishes entering a mob's inventory by hand pickups. Which is the only condition that this is realistically an issue. 

I cannot use COMSIG_ITEM_PICKUP as it is fired before the loc is set... or forceMove() even happens. COMSIG_ITEM_EQUIPPED is called upon entering hands surprisingly, after the forceMove() completes and the loc has long been set. I need the loc set as recursive move component relies on it to rebuild the loc parent chain.

How horridly cursed, thank you for listening to my tedtalk.

:cl: Will
code: topturfcrossed component for when you need to check for Crossed() actions from within nested items
fix: recursive move component properly rebuild it's hierarchy when picked up by a mob from a turf
/:cl:

### DIFF
--- a/html/changelogs/AutoChangeLog-pr-18768.yml
+++ b/html/changelogs/AutoChangeLog-pr-18768.yml
@@ -1,4 +1,0 @@
-author: "Guti"
-delete-after: True
-changes:
-  - bugfix: "Fixes sparks recursively teleporting through shadekin portals"
--- a/html/changelogs/AutoChangeLog-pr-18769.yml
+++ b/html/changelogs/AutoChangeLog-pr-18769.yml
@@ -1,4 +1,0 @@
-author: "Will"
-delete-after: True
-changes:
-  - bugfix: "Fix runtime in machine var shim, removed balloon alert when using old byond uis handled by it."
--- a/html/changelogs/AutoChangeLog-pr-18770.yml
+++ b/html/changelogs/AutoChangeLog-pr-18770.yml
@@ -1,4 +1,0 @@
-author: "Kashargul"
-delete-after: True
-changes:
-  - bugfix: "runtimes due to multiple signal registers on opacity changes"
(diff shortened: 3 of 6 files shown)


Cameron-The-Raven:
> COMSIG_ITEM_EQUIPPED is called upon entering hands surprisingly

It's so fucking cursed but yeah I ran into the same realization some time ago...I genuinely hate we don't have a specific 'inhand' signal.

PR #18796, opened 2025-11-17
Fixes fulton beacons to be useable for their intended purpose

ReoDaProtovali:
## About The Pull Request

fulton beacons were accidentally re-nerfed when a section of code was reverted on accident at one point or something
beacons are able to be placed anywhere due to bluespace, and this fixes that again. Extraction packs must be used outdoors, though
## Changelog
:cl
fix: Fixed fultons being useless, recovery beacons can be placed "indoors" again!
code: Removed a doubled definition that was there for some reason
/:cl:

### DIFF
--- a/code/modules/mining/fulton.dm
+++ b/code/modules/mining/fulton.dm
@@ -143,7 +143,7 @@
 
 /obj/item/fulton_core/attack_self(mob/user)
 	var/turf/T = get_turf(user)
-	if(do_after(user, 15, target = user) && !QDELETED(src))
+	if(do_after(user, 1.5 SECONDS, target = user) && !QDELETED(src))
 		new /obj/structure/extraction_point(get_turf(user))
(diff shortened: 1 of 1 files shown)


Cameron-The-Raven (approved):

code/modules/mining/fulton.dm
```diff
@@ -143,8 +143,7 @@
 
 /obj/item/fulton_core/attack_self(mob/user)
 	var/turf/T = get_turf(user)
-	var/outdoors = T.is_outdoors()
-	if(do_after(user, 15, target = user) && !QDELETED(src) && outdoors)
+	if(do_after(user, 15, target = user) && !QDELETED(src))
```
Cameron-The-Raven:
```suggestion
	if(do_after(user, 1.5 SECONDS, target = user) && !QDELETED(src))
```

PR #18800, opened 2025-11-18
add more whitelist checks

Kashargul:
## Changelog
:cl:
qol: adds input validation for whitelist changes
/:cl:

### DIFF
--- a/code/modules/tgs_commands/vorestation.dm
+++ b/code/modules/tgs_commands/vorestation.dm
@@ -459,7 +459,7 @@ GLOBAL_LIST_EMPTY(pending_discord_registrations)
 					message.text = "Error, invalid species entered. Check spelling and capitalization."
 					return message
 				if(!(role in GLOB.whitelisted_species))
-				    message.text = "Error, species \"[role]\" is not a whitelist species."
+					message.text = "Error, species \"[role]\" is not a whitelist species."
 					return message
 
 			var/datum/db_query/command_add = SSdbcore.NewQuery(


code/modules/tgs_commands/vorestation.dm
```diff
@@ -438,7 +438,30 @@ GLOBAL_LIST_EMPTY(pending_discord_registrations)
...
+					return message
+				if(!job.whitelist_only)
+					message.text = "Error, job \"[role]\" is not a whitelist job."
+					return message
+			if(kind == "species")
+				if(role in GLOB.playable_species)
+					message.text = "Error, species \"[role]\" is not a whitelist species."
+					return message
+				if(!(role in GLOB.whitelisted_species))
+					message.text = "Error, invalid species entered. Check spelling and capitalization."
```
ItsSelis: This should be the other way around, otherwise everything is fine.

```suggestion
				if(role in GLOB.playable_species)
					message.text = "Error, invalid species entered. Check spelling and capitalization."
					return message
				if(!(role in GLOB.whitelisted_species))
				    message.text = "Error, species \"[role]\" is not a whitelist species."
```

Kashargul:
```
	//Split up the rest
	for(var/speciesname in GLOB.all_species)
		var/datum/species/S = GLOB.all_species[speciesname]
		if(!(S.spawn_flags & SPECIES_IS_RESTRICTED))
			GLOB.playable_species += S.name
		if(S.spawn_flags & SPECIES_IS_WHITELISTED)
			GLOB.whitelisted_species += S.name
```

It's not a subset. It's either or, so we first need to check if it's a playable, then reject bad inputs.

PR #18813, opened 2025-11-21
Teslacoil Buff

Willburd:
## About The Pull Request
Fixes teslacoil parts affecting the coil's performance. The code for it was present, but it was setting local vars to the proc instead of vars on the object, so it discarded the intended effects at the end of the proc by setting only the temp var instead.

## Changelog
Makes teslacoil's upgraded parts actually affect a coils performance.

:cl: Will
fix: Teslacoil parts actually upgrade performance
/:cl:


Willburd:
Combined into https://github.com/VOREStation/VOREStation/pull/19339 and closing.

PR #18819, opened 2025-11-22
Fix shadekin voices being revealed

NovaDusklight:
## About The Pull Request

Fixes allows shadekin, when in phase, to hide their voice using the shadekin control panel.
This updates the shadekin control panel with a simple tickbox boolean.
This enables/disables their voice being "Something" when in phase, which is the default behavior for emotes.
This only applies when the shadekin is in phase. When they are visible, their name is shown, as was always default.

## Changelog

:cl: Nova
fix: Fix shadekin voices always being known in phase.
qol: Add ability to reveal/hide voice in shadekin control panel.
/:cl: Nova

### DIFF
--- a/code/__defines/dcs/signals/signals_mob/signals_mob_main.dm
+++ b/code/__defines/dcs/signals/signals_mob/signals_mob_main.dm
@@ -314,9 +314,6 @@
 ///from end of revival_healing_action(): ()
 #define COMSIG_LIVING_AHEAL "living_post_aheal"
 
-///from /datum/species/handle_fire. Called when the human is set on fire and burning clothes and stuff
-#define COMSIG_HUMAN_BURNING "human_burning"
-
 ///from /mob/living/carbon/human/GetVoice(): (list/voice_data) - voice_data[1] contains the voice name
 #define COMSIG_HUMAN_GET_VOICE "human_get_voice"
 	#define COMPONENT_VOICE_CHANGED (1<<0)


code/modules/mob/living/carbon/human/say.dm
```diff
@@ -88,6 +88,11 @@
 	return ..()
 
 /mob/living/carbon/human/GetVoice()
+	// Check if shadekin is phased and has voice hiding enabled
+	var/datum/component/shadekin/SK = get_shadekin_component()
```
Kashargul: we don't use single or two letter vars anylonger

tgui/packages/tgui/interfaces/ShadekinConfig.tsx
```diff
@@ -38,14 +39,15 @@ export const ShadekinConfig = (props) => {
...
+    hide_voice_in_phase,
   } = data;
 
   const isSubtle =
     flicker_time < 5 || flicker_break_chance < 5 || flicker_distance < 5;
 
   const windowHeight =
     (isSubtle ? 220 : 190) +
-    (extended_kin ? 95 : 0) +
+    (extended_kin ? 105 : 55) +
```
Kashargul: where does that 55 come from?

code/modules/mob/living/carbon/human/say.dm
NovaDusklight: Thanks, I'll change it.

NovaDusklight: Edited to one comment after out discussion to keep this clean.
How should the SK variable be renamed and scoped in this case?

tgui/packages/tgui/interfaces/ShadekinConfig.tsx
NovaDusklight: I misinterpreted how this was handling the UI. It should remain 0.
I've fixed this now.

code/modules/mob/living/carbon/human/say.dm
NovaDusklight: For context, there are 39 instances of:
var/datum/component/shadekin/SK

Would it not be best I use the same practice as the other 38 instances, so that when we change the variable name later, it can be done in bulk easier?

tgui/packages/tgui/interfaces/ShadekinConfig.tsx
```diff
@@ -38,14 +39,15 @@ export const ShadekinConfig = (props) => {
...
+    hide_voice_in_phase,
   } = data;
 
   const isSubtle =
     flicker_time < 5 || flicker_break_chance < 5 || flicker_distance < 5;
 
   const windowHeight =
     (isSubtle ? 220 : 190) +
-    (extended_kin ? 95 : 0) +
+    (extended_kin ? 110 : 0) +
```
Kashargul: still in the wrong section. You are increasing the extended settings, but your code is not for the extended one. It adds to the base size above on both hands

NovaDusklight: I see I misunderstood what the extended_kin was for now.
I just tested with the height always added to make sure it looks right, without needing resizing now.

code/__defines/dcs/signals.dm
```diff
@@ -740,6 +740,15 @@
...
 #define COMSIG_HUMAN_DNA_FINALIZED "human_dna_finished"
+///from /mob/living/carbon/human/GetVoice(): (list/voice_data) - voice_data[1] contains the voice name
+#define COMSIG_HUMAN_GET_VOICE "human_get_voice"
+	#define COMPONENT_VOICE_CHANGED (1<<0)
+///from /mob/living/carbon/human/GetAltName(): (list/name_data) - name_data[1] contains the alt name
+#define COMSIG_HUMAN_GET_ALT_NAME "human_get_alt_name"
+	#define COMPONENT_ALT_NAME_CHANGED (1<<0)
+///from /mob/living/carbon/human/get_visible_name(): (list/name_data) - name_data[1] contains the visible name
+#define COMSIG_HUMAN_GET_VISIBLE_NAME "human_get_visible_name"
+	#define COMPONENT_VISIBLE_NAME_CHANGED (1<<0)
```
Cameron-The-Raven: Can you put these in a separate file? Signals.dm is being deleted in favor of individualized files in https://github.com/VOREStation/VOREStation/pull/18741 and doing such will make decon easier.

NovaDusklight: Sure, I'll take a look at that soon. It's 3AM right now, so not sure if now, or when I wake up xD

NovaDusklight: I've moved them over to signals_mob_main.dm, and re-tested

PR #18823, opened 2025-11-24
More surgery tweaks and fixes

Cameron-The-Raven:
## About The Pull Request
- Makes frankensurgery possible on boneless entities
- Adds defines to make it easier to track what stage a surgery step has set the limb to be
- Made it so brain surgery no longer applies a permanent debuff (this discouraged bringing people back alive and encouraged just resleeving them)
- Adjusted a few steps to have min&max duration for surgery to = each other
- Using hemostats now clamps arteries, causing internal bleeding to be slowed by 10x in that limb until the internal bleeding is properly fixed.
- Removes a hardref cortical stacks did by having a hardref to the owner's mind.
- Removes a few duplicate spread_germs_to_organ() procs done in end_step(). This is already called in start_step()
- Makes reform always heal someone up from death instead of possibly having them remain dead if they took mass damage.
- REAGENT_ID_MYELAMINE now makes it so blood loss is slowed down by an additional 30 (on top of the default 40) when taken for internal blood loss, meaning treating it will slow it down further until the medication completely closes the wound.
- Tourettes will not have a 50x more likely chance to trigger if you have a slow pulse or if your heart has stopped.
- You can no longer accidentally do the bone repair surgery if the bone if fixed (Or if they have no bones!)
## Changelog
:cl: Diana
qol: Hemostats now clamp internal bleeding, causing 10x slower bloodloss during surgery.
fix: Brain surgery can be done on boneless creatures.
code: Cleaned up surgery code and made it more easily readable.
spellcheck: Removed a few unneeded spaces in text.
qol: Brain surgery no longer permanently debuffs the person it's performed on.
qol: Made a few surgery steps have the same min and max time duration.
fix: Reform vore now will always bring the person at least to alive status.
fix: Tourettes will no longer trigger 50x more likely if your pulse is slowed or stopped.
fix: You can no longer accidentally do bone repair surgery when the bone is already fixed
/:cl:


code/modules/surgery/encased.dm
```diff
@@ -255,7 +255,7 @@
 	if(!ishuman(target))
 		return
 	var/obj/item/organ/external/affected = target.get_organ(target_zone)
-	return ..() && affected && affected.open >= 2 && affected.open < 3
+	return ..() && affected && affected.open >= FLESH_RETRACTED && affected.open < 3
```
Cameron-The-Raven:
```suggestion
	return ..() && affected && affected.open >= FLESH_RETRACTED && affected.open < BONE_RETRACTED
```

PR #18824, opened 2025-11-24
No shadekin knockdown on shuttle move

Willburd:
## About The Pull Request
Does not remove shadekin being gibbed under shuttles landing. Stops shadekin being thrown or knocked down by shuttles they are on moving if they are phased. This seems to be more of a physical interaction. This can be closed if throwing/knocking phased shadekin during shuttle movement is intended?

## Changelog
Shadekin in phase will not be knocked down by shuttle movement.

:cl: Will
fix: Phased shadekin are no longer knocked down or thrown when a shuttle they are inside of moves.
/:cl:

### DIFF
--- a/code/modules/shuttles/shuttle.dm
+++ b/code/modules/shuttles/shuttle.dm
@@ -330,6 +330,8 @@
 					TA.ChangeTurf(get_base_turf_by_area(TA), 1, 1)
 		if(knockdown)
 			for(var/mob/living/M in A)
+				if(M.is_incorporeal())
+					continue
 				if(M.buckled)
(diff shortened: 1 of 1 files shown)


Kashargul:
Almost the question, why does the shuttle even transport them?

Willburd:
> Almost the question, why does the shuttle even transport them?

At least from my own players on Outpost the shuttle not taking them would be undesirable behavior, currently it only knocks them down and shows a message they were. This just hides that allowing them to tag along with shuttles to get to other Z levels. 

I would prefer this personally over shadekin getting a teleport to zlevel or mob ability. I think this PR's behavior would also be accepted by virgo staff if asked?

PR #18832, opened 2025-11-27
Makes ambulant blood not a genetrait [GBP IGNORE]

Cameron-The-Raven:
## About The Pull Request
This was being used to changeling test, If a changeling DIDN'T have it, you could instakill them. If they DID have it, it meant they got to live longer while they were tested, ultimately doing the opposite of what the trait entails. This happened on numerous shifts and has gone from 'negative' to generally 'positive' trait.
## Changelog
:cl: Diana
balance: Ambulant blood no longer is a gene trait.
/:cl:


Cameron-The-Raven:
This gets 0 use on Virgo (and chomp) and only affects the combat downstream. (I spoke to them and they're good with this) so mergin

Issue #17303, opened 2025-03-10
[MAPPING] Virgo StairB


Sixthberry:
### Brief description of the bug

Stairs everywhere!

### What you expected to happen

A normal day.

### What actually happened

Stairs!

![Image](https://github.com/user-attachments/assets/49ea21c2-9b04-43a8-8ea3-090688ad0717)
![Image](https://github.com/user-attachments/assets/d2a5c787-08ff-455a-9dcd-651f5be73a3f)

### Reproduction steps

```bash
Just go up to the Midpoint?
```

### Code Revision

Commit: c9eb8e0654898baecfa681ad1a011b3b4b3fdcce

PR #18844, opened 2025-11-29
Floorbreaker Landmark

KillianKirilenko:
## About The Pull Request

Adds a mapping landmark that, at late_init, will break the tile under it. Removes the need for `/broken` subtypes of existing flooring. If it somehow ends up on a non-simulated and/or non-floor turf it logs an error with its X/Y instead.

No player-facing changes, no changes to existing maps.

## Changelog

:cl:
qol: (mapping) broken floors can now be done via a landmark rather than a subtype of that specific floor
/:cl:


Kashargul:
That shouldn't be done though. This adds init overhead. Subtyping does not.

Kashargul:
After past experiences with such helpers... and for example tether having an entire Z levels of such effects... I really don't feel like we should bring such back.

Kashargul:
Solution without subtyping or new vars:

```
if(broken == -1)
    broken = rand(0, flooring.has_damage_range)
```

can be tied to turf init and edited by mappers go that

KillianKirilenko:
> That shouldn't be done though. This adds init overhead. Subtyping does not.

Except the current broken subtypes do the exact same thing. All I've done is offload what they do to a marker object.

The init overhead should be minimal so long as they're not overused. If they're overused, mappers are doing something wrong.

> After past experiences with such helpers... and for example tether having an entire Z levels of such effects... I really don't feel like we should bring such back.

"an entire Z level"? that's hyperbole and you know it, Kash. C'mon.

> Solution without subtyping or new vars:
> 
> ```
> if(broken == -1)
>     broken = rand(0, flooring.has_damage_range)
> ```
> 
> can be tied to turf init and edited by mappers go that

I don't really feel like that's an ideal solution either, as it becomes difficult to tell which tiles have been mapedited when scrolling around unless they are also mapedited to a different iconstate.

Kashargul:
<https://github.com/VOREStation/VOREStation/issues/17303>

I present you the entire Z-Level of fall trigger effects... exposed by a change on effect visibility not that long ago

KillianKirilenko:
> #17303
> 
> I present you the entire Z-Level of fall trigger effects... exposed by a change on effect visibility not that long ago

Alright, I'll give you that one, but I feel it's a _little_ different as those are there for the entire round whereas these are one-and-done deals that get deleted after init is done. They're more like floor decals.

code/game/objects/effects/floorbreak.dm
```diff
@@ -0,0 +1,16 @@
+/obj/effect/landmark/floorbreak
+	name = "Floor Breaker"
+	icon_state = "floorbreaker"
+	delete_me = FALSE //just to be sure
+
+/obj/effect/landmark/floorbreak/Initialize(mapload)
+	. = ..()
```
Kashargul, line 7: After testing this, this will hard del horribly, or simply not at all... becasue QDEL_HINT_LETMELIVE... if you check the landmark code, it gets added to global landamrks

Which is also a source of null space langmarks later on.. do we have any other of those landmarks, because any use of them for things like that with delete_me set to false would explain why people end in nullspace

code/game/objects/effects/landmarks.dm
```diff
@@ -65,11 +65,9 @@
 		if("bluespacerift")
 			GLOB.endgame_exits += loc
 			delete_me = TRUE
-		//VOREStation Add Start
 		if("vinestart")
 			GLOB.vinestart += loc
 			delete_me = TRUE
-		//VORE Station Add End
 
 	if(delete_me)
```
Kashargul, line 72: The code below should be adjusted again, delete_me landmarks should actually delete

```
/obj/effect/landmark/Destroy(var/force = FALSE)
	if(delete_me || force)
		GLOB.landmarks_list -= src
		return ..()
	return QDEL_HINT_LETMELIVE
```

code/game/objects/effects/floorbreak.dm
```diff
@@ -0,0 +1,16 @@
...
+	return INITIALIZE_HINT_LATELOAD
+
+/obj/effect/landmark/floorbreak/LateInitialize()
+	if(istype(src.loc, /turf/simulated/floor))
+		var/turf/simulated/floor/our_turf = src.loc
+		our_turf.break_tile()
+	else
+		log_world("Floor Breaker at [src.loc.x] / [src.loc.y] was somehow placed in a non-turf location, or placed on an unsimulated or non-floor turf (e.g. wall, open space).")
+	qdel(src)
```
Kashargul, line 16: delete_me needs to be true before a qdel can be called ever; it makes no sense for this to be a landmark. It should just be an obj/effect/floorbreak if at all

code/game/objects/effects/landmarks.dm
KillianKirilenko: Resolved via upstream merge.

code/game/objects/effects/floorbreak.dm
KillianKirilenko: Noted. I'll experiment with this later.

KillianKirilenko: Seems to work fine after conversion.

code/game/objects/effects/floorbreak.dm
```diff
@@ -0,0 +1,16 @@
...
+/obj/effect/floorbreak/Initialize(mapload)
+	. = ..()
+	return INITIALIZE_HINT_LATELOAD
+
+/obj/effect/floorbreak/LateInitialize()
+	if(istype(src.loc, /turf/simulated/floor))
+		var/turf/simulated/floor/our_turf = src.loc
+		our_turf.break_tile()
+	else
+		log_world("Floor Breaker at [src.loc.x] / [src.loc.y] was somehow placed in a non-turf location, or placed on an unsimulated turf, non-floor turf, or other invalid location (e.g. wall, open space, inside a container).")
```
Kashargul: this check should be put into init and return init hint qdel after logging, only return lateload if we really do something, rest seems fine

KillianKirilenko: Done.

code/game/objects/effects/floorbreak.dm
```diff
@@ -0,0 +1,17 @@
...
+	name = "Floor Breaker"
+	icon = 'icons/mob/screen1.dmi'
+	icon_state = "floorbreaker"
+
+/obj/effect/floorbreak/Initialize(mapload)
+	. = ..()
+	if(!istype(src.loc, /turf/simulated/floor))
+		log_world("Floor Breaker at X: [src.loc.x], Y: [src.loc.y] was somehow placed in a non-turf location, or placed on an unsimulated turf, non-floor turf, or other invalid location (e.g. wall, open space, inside a container).")
+		qdel(src)
+		return
```
Kashargul: return INITIALIZE_HINT_QDEL

qdel is illegal to call in Init itself

PR #18849, opened 2025-11-29
Mech construction fix and unittest

Willburd:
## About The Pull Request
Turns out byond breaks a new X() call if the path being provided is a string. Why are mech construction paths a string? I DUNNO! fixing it now and unit testing the strings away. We have path type for a reason, and the only reason it's not changed is likely just legacy. 

Also calling new typehere() in a signal call appears to not work, may be a weird macro thing, so i'm required to store the result in a var and pass it instead a line prior. I did the same thing for slimes and they worked, but I'm going to make those the same for consistency.

## Changelog
Mechs use a path instead of a string in their construction
Mech construction unittest
Fixed mechs vanishing on construction

:cl: Will
fix: mechs no longer vanish when attempting to construct them
fix: slimecores not being produced during grinding
code: Unittest for mech construction datums
/:cl:

### DIFF
--- a/code/modules/xenobio/items/weapons_vr.dm
+++ b/code/modules/xenobio/items/weapons_vr.dm
@@ -112,7 +112,7 @@
 		while(S.cores)
 			playsound(src, 'sound/machines/juicer.ogg', 25, 1)
 			if(do_after(user, 15, target = src))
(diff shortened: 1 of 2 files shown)


Willburd:
I hate mech code

PR #18853, opened 2025-11-30
Patch fix, Neurotoxins, and Paralysis

Cameron-The-Raven:
## About The Pull Request

Big points:
1. The way paralysis (And fakedeath) works has been redone.
Paralysis no longer immediately KO's you and stops you from hearing/seeing. Instead, it paralyzes you (duh).
This means if you are paralyzed, you can still look around (but won't turn to face said things), still hear, and still see, but you are stuck in place!
You can not emote or speak (you are paralyzed) BUT you can still subtle! Meaning if you use paralysis chems in a scene, you can still communicate/post with your captor.

2. Patches have been fixed and patch chemicals no longer kill you.
2.5 If a chemical is not designed for patches, it now applies to the blood at 25% efficiency.

3. Lichpowder is no longer a toxin, but now a normal reagent.
Zombie powder is still a toxin so it will do slowly increasing toxins.
Lichpowder no longer suffocates you.
Zombie powder now does suffocation at a reduced rate.

4. More things respect the FAKEDEATH status flag.
A lot of things didn't really care about fakedeath. Now if you examine someone, it won't show the 'SSD' message, but rather just the normal 'they aren't breathing' and 'they appear to be sleeping' messages that a corpse shows.
Health analyzers no longer break when you scan a person with FAKEDEATH flag.
PDA health analyzers respect the FAKEDEATH flag now.
Individuals under the FAKEDEATH flag will no longer have status indicators above them, preventing them from being immediately outted.
Fake death no longer paralyzes, but rather simply makes you appear dead. Whatever is CAUSING the fake death should apply an applicable (see: paralysis/weaken/stun/sleep/etc) status effect if more implementations are made, but this isn't essential.

5. Adds beakers of lich powder and zombie powder for easy debugging.
Lich powder now has a MUCH lower chance of doing a visible message and giving you away.

6. Adds a new chemical reaction for an alternative way to make zombiepowder
This reaction requires list(REAGENT_ID_POTASSIUMCHLOROPHORIDE = 5, REAGENT_ID_STOXIN = 5, REAGENT_ID_CHLORALHYDRATE = 5, REAGENT_ID_MERCURY = 5)
to create 2 units of zombiepowder.
(The idea behind it was 'slam a bunch of neurotoxins, cardiac inhibitors, and respiratory inhibitors to make this')
This makes a -relatively- labor intensive method to create zombie powder that doesn't involve carp toxin without involving distilling in the mix (which is reserved for the much nicer Lich powder).

## Changelog
:cl: Diana
qol: Paralysis no longer KOs you. Instead, you are able to see and hear everything going on around you, just not interact, move, or anything of that manner! Additionally, you can not use 'me' while paralyzed, but can however use 'subtle!'
fix: Medical patches no longer apply to the blood and result in topical medications killing you.
qol: Medications now apply at an efficiency of 25% into the bloodstream if a patch is not a topical medication.
qol: Fakedeath doesn't immediately out you via raising toxins damage under scanners.
qol: Lichpowder no longer suffocates you to death.
qol: Zombiepowder now does a much lower suffocation damage.
add: Adds a new way to make zombiepowder which uses 5u potassiumchlorophoride, 5u sleep toxin, 5u chloral, and 5u mercury for 2u of zombie powder.
fix: Fakedeath now applies to PDA scanners as well
fix: Medications that are dermal can now be used via patches again.
fix: Status flags will no longer show above individuals that are mimicking death.
qol: Medications that are not dermal now are absorbed into the blood at 25% efficiency.
code: Gets rid of a unneeded proc in custom_emote
fix: Fakdeath will no longer break medical scanners
/:cl:


Cameron-The-Raven:
dreamseeker_2025-12-02_07-02-59
image 2nd needs fixed

PR #18858, opened 2025-12-01
Turf Cascade Subsystem

Willburd:
## About The Pull Request
Originally the bluespace cascade spawned in process flagged turfs that expanded themselves. This is horribly laggy. Moved it to a subsystem dedicated for it.

## Changelog
Moves bluespace cascade to a subsystem, it's generic so you could make a version for the nuke that is nuclear hellfire instead of bluespace instead.

:cl: Will
refactor: Bluespace cascade moved to subsystem
/:cl:

### DIFF
--- a/code/controllers/subsystems/turf_cascade.dm
+++ b/code/controllers/subsystems/turf_cascade.dm
@@ -33,7 +33,7 @@ SUBSYSTEM_DEF(turf_cascade)
 			return
 		last_group_time = world.time
 
-		if(!length(currentrun) && length(remaining_turf) && !turf_iterations)
+		if(!length(currentrun) && length(remaining_turf) && turf_iterations <= 0)
 			// Create a random list of tiles to expand with instead of doing it in order
 			var/subtractive_rand_max = conversion_rate * (1 - (conversion_probability / 100))
 			var/i = 10 // Always do at least a handful of the oldest, to avoid spots that linger unfilled


code/controllers/subsystems/turf_cascade.dm
```diff
@@ -0,0 +1,70 @@
+#define DEFAULT_CONVERSION_RATE 75
+#define DEFAULT_CONVERSION_DELAY 0.25 SECONDS
+
+SUBSYSTEM_DEF(turf_cascade)
+	name = "Turf Cascade"
+	wait = DEFAULT_CONVERSION_DELAY
+	flags = SS_NO_INIT
+
+	dependencies = list(
+		/datum/controller/subsystem/mobs
```
Kashargul: why dependencies when it's a SS NO INIT?

Willburd: I just assumed it needed them. will remove.

Willburd:
Needs a tracy and a compare to the speed it originally went at to avoid major player facing changes, but those are just argument tweaks.

Willburd:
Just confirming that this is more or less done, it just needs a tracy profiling and a comparison to original behavior of the supermatter turf spreading, to know if it needs to be sped up or slowed down.

Willburd:
Tried to match this to the original rate of growth that this had before, it should be ready now. Better than every single turf being shoved up SSobj's processing hole.

Cameron-The-Raven:
Gives massive CPU gains and doesn't cause massive lag. Tracy is extremely positive looking compared to the old version.

PR #18860, opened 2025-12-03
nerfs pulled punches

zeskorion:
## About The Pull Request
Pulled punches have been reduced in effectiveness, to no longer be better at downing someone than actually punching them without holding back.

## Changelog

:cl:
balance: Pulled punches are now affected by species brutemod, and don't have an inherent damage bonus
balance: Pulled punches no longer use species unarmed bonus. You're holding back, not hitting with full strength
fix: xenochimerae cannot pull their special unarmed attack when feral
/:cl:


zeskorion:
open to whatever balance changes need to be made. I mostly did this PR for the purposes of immersion- I don't like pulling punches being more effective than going all out, if you want to actually put someone on the floor.

Cameron-The-Raven:
I'd suggest halving the user.species.unarmed_bonus by 1/2 on pulled punches instead of removing it entirely. 

You might be pulling your punches, but if you're boxing and you have 'inhuman strength' (as the trait is called) it'd make sense that you'd have an innate advantage over your opponent that isn't genetically built like a brick shithouse.

If you go to box with Mike Tyson and he's pulling his punches, it's still going to hurt pretty bad.

zeskorion:
> I'd suggest halving the user.species.unarmed_bonus by 1/2 on pulled punches instead of removing it entirely.
> 
> You might be pulling your punches, but if you're boxing and you have 'inhuman strength' (as the trait is called) it'd make sense that you'd have an innate advantage over your opponent that isn't genetically built like a brick shithouse.
> 
> If you go to box with Mike Tyson and he's pulling his punches, it's still going to hurt pretty bad.

yes, but at the point that you're hitting harder with pulled punches, shouldn't it just deal real damage? My gripe is kinda that, at a certain point, it should just be dealing real damage- pulling punches should you hit with *less* force, rather than magically turning your punch damage into special Nonlethal Damage That slows You At Ten Points. I *could* change the code to make it instead work like that, but I thought this was a solution that preserves the nonlethal sparring nature of pulled punches.  If I boxed with mike tyson, I'd have real bruises, and not magical woundless bruises.

plus, a number of people have stated a desire for pulled punches to be roughly symmetrical for this reason. it makes sparring a lot more fair.

PR #18861, opened 2025-12-03
Gun circuitry balance tweaks

Cameron-The-Raven:
## About The Pull Request
Adjusts the gun circuit to make it less insane.

You can still do some INSANE THINGS with it (like wearing clothing full of guns) but you can no longer shove 7 laser cannons into one circuit, shove 7 of those into a bluespace duffle, connect them via EPV2 circuits, shove 40 of those into a crate, put 5 of those crates on the cargo tug, and then click someone and have them eat 9800 laser cannon shots in one click (784000 damage)

I was witness to this monstrosity in action and I'm killing this foul beast where it stands.

## Adjustments:
Gun size now affects the complexity of the gun circuit. Tiny guns keeps complexity at 30, small 60, normal 90, large 120, bulky 150. 
(The only circuit that really gets used for gun circuits are the ones with 240 max complexity, so this makes it so you can no longer fit 7 laser cannons into a single circuit.)

Circuits will no longer fire the gun if it's:
1. In a storage item (backpack, webbing, pouches, etc)
2. In a container (closets, crates, etc)
3. On the same tile as a container (Prevents shoving 40 of them into a crate and then unleashing Dr Eggman's death beam on them)

Also does some general cleanup of the firing mechanism code to use some early returns.
## Changelog
:cl: Diana
balance: Gun circuits will no longer allow fire while obscured inside of items such as crates, lockers, or bags.
balance: Gun circuits will now have their complexity increased based on how large the gun you place inside it is.
code: Cleans up the gun circuit code a little to use early returns.
/:cl:

### DIFF
--- a/code/modules/integrated_electronics/subtypes/manipulation.dm
+++ b/code/modules/integrated_electronics/subtypes/manipulation.dm
@@ -7,7 +7,9 @@
 	extended_desc = "The firing mechanism can slot in most ranged weapons, ballistic and energy.  \
 	The first and second inputs need to be numbers.  They are coordinates for the gun to fire at, relative to the machine itself.  \
 	The 'fire' activator will cause the mechanism to attempt to fire the weapon at the coordinates, if possible.  Note that the \
-	normal limitations to firearms, such as ammunition requirements and firing delays, still hold true if fired by the mechanism."
+	normal limitations to firearms, such as ammunition requirements and firing delays, still hold true if fired by the mechanism. \
+	Additionally, the complexity of the circuit increases based on the weapon's size. A tiny gun will have 30 complexity, a small 60, medium 90, large 120, and bulky 150. \
+	It can likewise not shoot while obscured (inside a container, such as a closet or backpack) or if there is a closet or container on the same tile as it."
 	complexity = 30 // Increased due to clear balance necessity.
 	w_class = ITEMSIZE_NORMAL
 	size = 3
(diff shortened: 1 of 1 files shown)


TeleTubby2:
Unfortunate, this kinda thing was really only enabled by being able to bulk produce circuits via copypasta, would take like 4 hours to do otherwise lol

Cameron-The-Raven:
> Unfortunate, this kinda thing was really only enabled by being able to bulk produce circuits via copypasta, would take like 4 hours to do otherwise lol

I'd rather them pop up now and get shown off (and abused) so they can be properly adjusted.

This is still by no means 'balanced' as you can still wear circuit helm/chest/boots/gloves and put guns in them, but at least then you're having to swap armor for guns...I'll have to keep an eye on it to see if it's still OP.

I really DON'T want to implement having guns in circuitry use their 'one handed accuracy' and make it essentially RNG if you 1shot things or not, but if I get oneshot again by someone being a funny guy and putting 50 laser cannons in a circuit I'm going to be annoyed and have to do some further balancing.

TeleTubby2:
I feel like if people are powergaming it that's more of an administrative issue, but I've got zero idea of the downstream kinda environment so it's probably an actual issue there.

Cameron-The-Raven:
> I feel like if people are powergaming it that's more of an administrative issue, but I've got zero idea of the downstream kinda environment so it's probably an actual issue there.

It's one of those things that the game facilitates it. It's not really on the _player_ for using tools that are available to them. 

The game _allows_ you to freely do the above (In the more used instance I've seen, it was putting 7 laser cannons into a single circuit and using that as a god-gun, which  does 560 damage w/ 50% armor pen aka dead)

This is the same situation as when someone found out how to hack the script terminal in game and PDA bombed everyone 10x times all at once. 

It's funny when done once, but then you tell them not to do it again and patch it afterwards to a (more) balanced state.

PR #18869, opened 2025-12-06
Improves seasonal turf code slightly

TheCaramelion:
## About The Pull Request

Nothing player-facing.

Changes the seasonal turf code to use a proc instead of having all their logic on Initialize, making them easier to override.

## Changelog
:cl: Guti
code: Seasonal turfs are easier to override now
/:cl:


TheCaramelion:
Fought them enough, I'm porting biomes instead

PR #18885, opened 2025-12-11
Movement adjustments.

Cameron-The-Raven:
## Health Deficiency / Halloss
Makes 'health deficiency' no longer hardset numbers but rather a percentile. (Slowdown still begins when you have taken 40% or more damage)

This means that characters that have lower max HP will require less injury to 'feel it', while characters with higher max HP will require more injury to 'feel it'.

Species trauma mod (pain sensitivity) still applies here and someone with 2x trauma mod will begin to feel the effects of their injuries at 20% instead of 40% and slowdown will be 2x more severe (slowdown of 4 while at 50% injury instead of at 100%)

When it comes to slowdown, halloss no longer applies a halloss/10 slowdown. Now instead uses the above health deficiency code.

35 halloss will do a 1.4 slowdown (bringing move delay from 3 from to 4.4, which is EXTREMELY noticeable)

70 halloss (2 taser shots) will do a slowdown of 2.8. This is essentially diona speed.

Even if a target is 'harder to take down' such as having a higher max HP cap, they will still be slowed by pain.

Painkillers still reduce traumatic_shock like normal. Painkillers will decrease damage slowdown like normal.
Painkillers are only 33% effective against halloss AND will never lower the slowdown from halloss below 1 move delay (equivalent to 25 halloss) meaning a single taser shot will still be EXTREMELY effective at slowing someone down.

Caps injury slowdown for the health percent at 4 (occurs when at 0% health, or 100% injured. Also applies with 100 halloss damage.)

## Shock

Shock stage now operates linearly instead of applying a massive slowdown once you hit a magic number.
Once hitting a shock_stage of 10, your 'adrenaline' starts to wear off and you begin to suffer increasing slowdown. This means the longer you're in shock, the slower you will become, from a min of 0.5 (just entered shock) to a max of 3.0 (been in shock for 2 minutes OR suffered a major, traumatic injury such as the loss of a limb)

Additionally, if you have narcotics in your system (just oxycodone atm) you won't immediately receive 61 shock_stage upon entering soft-crit.

## Aim Adjustments

REMOVES slowdown penalty from aiming. Aim mode is -already- detrimental (it will make you always miss projectile weapons on a moving target) and applying a further debuff to the user (it applied a slowdown of FIVE, equivalent to 50 pre-change halloss) made very little sense.

## Misc 
Makes mRun mutation ('no slowdown mutation') actually remove slowdown after the slowdown calculation is all performed.

Stops doubling of limbloss slowdown. Stance_damage already took care of slowdown from limbloss, but it was duplicated, meaning lost limbs were causing 2x the slowdown they should have.

Cleans up the unusual running trait some and gets rid of a lot of unneeded checks it was performing.

Makes unusual running a TRAIT()
Fixes force_max_speed and haste modifier not applying HUMAN_LOWEST_SLOWDOWN

Makes species slowdown only be calculated if NOT in space (This will cover most cases of 'the species is heavy and therefore slower' while some edge cases such as diona's 'slower because of physical movement disability' are able to move faster)

pain_mod is used for pain calculations now (as intended) instead of trauma_mod (these are completely different systems)

## Changelog
:cl: Diana
qol: Heavily reworks human movement code, making things such as halloss affected by species pain tolerance, making injury slowdown based on health % instead of a hardcoded 40 damage, making traumatic shock gradually apply slowdown instead of all at once.
qol: Removes 5 slowdown penalty when using AIM MODE
fix: force_max_speed will now properly apply the HUMAN MAX SPEED instead of allowing things to slow you down.
fix: Modifiers with haste now apply HUMAN MAX SPEED instead of allowing things to slow you down.
code: Makes Unusual Movement a TRAIT() instead of a species var.
balance: Oxycodone now fully ignores slowdown from halloss and injury. (It still applies 1 slowdown and multiplies all slowdown you're taking by 1.25x however)
balance: The max slowdown from injury / halloss is now 4.
fix: pain_mod now affects pain.
/:cl:


Cameron-The-Raven:
testing done, bugs ironed out now that I've had a second to actually sit down and test & make adjustments.

PR #18901, opened 2025-12-15
Flavored Donk Pocket Microwave Fix

Lilly-Lira:
## About The Pull Request

Fixes flavored donk pockets from losing their flavor when microwaved.

## Changelog

:cl:
fix: Flavored donk pockets retain flavor after microwaving
/:cl:

### DIFF
--- a/code/modules/food/recipes_microwave.dm
+++ b/code/modules/food/recipes_microwave.dm
@@ -67,7 +67,7 @@ I said no!
 	)
 	result = /obj/item/reagent_containers/food/snacks/donkpocket //SPECIAL
 
-/datum/recipe/donkpocket/warm/make_food(var/obj/container as obj)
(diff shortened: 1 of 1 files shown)


code/modules/food/recipes_microwave.dm
```diff
@@ -67,6 +67,17 @@ I said no!
 	)
 	result = /obj/item/reagent_containers/food/snacks/donkpocket //SPECIAL
 
+/datum/recipe/donkpocket/warm/make_food(var/obj/container as obj)
```
Kashargul: why obj as obj? Makes no sense. This explicitly is an object container... remove the as part

Lilly-Lira: I mirrored the same declarations as the parent, but agree it's not necessary, so I've removed it!  Changing the parent is out of scope for this PR so I've left the var/obj/container as obj unchanged in the parent make_food

PR #18919, opened 2025-12-18
Laying down is random once again

ReoDaProtovali:
## About The Pull Request

When you lay down, the direction you lay down in is random, again. You can still toggle the direction you want to be laying in using the flip resting direction verb if you want to lay in a specific direction
## Changelog
:cl:
qol: Resting is random again, getting up and laying back down wont always result in you laying in the same direction every time
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human.dm
+++ b/code/modules/mob/living/carbon/human/human.dm
@@ -1852,6 +1852,31 @@
 	rest_dir = !rest_dir
 	update_transform(TRUE)
 
+// These are here mostly for people who wish to use macros for funny things, so they can lay in specific directions.
+// They're hidden so they dont clog up the verb panel normally.
+/mob/living/carbon/human/verb/lay_right()
+	set name = "Rest Right"
+	set hidden = TRUE
+
+	rest_dir = FALSE
+	resting = !resting
+	to_chat(src, span_notice("You are now [resting ? "resting" : "getting up"]."))
+
+/mob/living/carbon/human/verb/lay_left()
+	set name = "Rest Left"
+	set hidden = TRUE
+
+	rest_dir = TRUE
+	resting = !resting
+	to_chat(src, span_notice("You are now [resting ? "resting" : "getting up"]."))
+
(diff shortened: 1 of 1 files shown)


Cameron-The-Raven:
Having seen people with macros that rely on the 'laying down' being locked to one direction or another, I can see this interfering with their macros and making their 'emotes' impossible because of the RNG this adds.

A toggle would probably be best to enable/disable random direction facing.

ReoDaProtovali:
this is the functionality it had before the ability to manually toggle the direction was added... Though, arguably having to *toggle* randomness for the benefit of a few macros is pretty questionable in my opinion. If it's something purely for macros, we might as well readd the verbs for laying down in specific directions. ...Probably as hidden verbs, so it doesnt clutter the menu. That's my thoughts, anyhow.

Cameron-The-Raven:
> this is the functionality it had before the ability to manually toggle the direction was added... Though, arguably having to _toggle_ randomness for the benefit of a few macros is pretty questionable in my opinion. If it's something purely for macros, we might as well readd the verbs for laying down in specific directions. ...Probably as hidden verbs, so it doesnt clutter the menu. That's my thoughts, anyhow.

Adding hidden verbs to lay-left or lay-right could work, yeah

Kashargul:
Heavily against the verb bloat, verbs are a performance horror as they are

Not even to mention the amount of copy pasted code

ReoDaProtovali:
even just to exist? I could instead add a single verb that has a parameter that you can set with a macro, maybe...

Kashargul:
> even just to exist? I could instead add a single verb that has a parameter that you can set with a macro, maybe...

I would prefer a character preference to jsut set it to be static or random

ReoDaProtovali:
that's a lot more work for something that's arguably only being done for macro users (and still doesnt properly fix that issue unless there's a verb to toggle stuff)

ShadowLarkens:
Thank you for the contribution! However, we've decided that we prefer resting's current mechanics.

PR #18920, opened 2025-12-19
Turret Processing Fix

Willburd:
https://github.com/user-attachments/assets/df4d2665-4668-4169-a0bd-c25fdfa26ccd

(First turret is a standard turret, second is a turret with the maximum possible firerate.)

## About The Pull Request
Turret firerates were entirely dependent and limited by SSmachine's processing tick rate. Meaning all of their firerate and retaliation code wasn't actually using their real fire rates. This moves turrets to fast processing, and uses a world time delay to avoid excessive vision checks. Also added some sounds to turrets when they enter and leave retaliation mode, as it is much more dangerous now that it functions.

## Changelog
Removes multiple sleep and spawn calls in turrets, and moves them to timers.
Moves turrets to fast processing, and adds it's own delay code
Turret retaliation code actually functions now, gave it sounds to make it clear that its in the mode.

:cl: Will
fix: Turrets are no longer limited to 2 second tickrate when in combat
fix: Turret retaliation behavior is actually functional, and makes sounds when they enter and leave it.
/:cl:


code/game/machinery/portable_turret.dm
```diff
@@ -47,6 +51,9 @@
 	req_one_access = list(ACCESS_SECURITY, ACCESS_HEADS)
 	blocks_emissive = EMISSIVE_BLOCK_UNIQUE
 
+	speed_process = TRUE // Otherwise we'll only tick every 2 seconds
```
Kashargul: can we please not default them??? Change to speed processing if needed. Else keep it on the normal one

Kashargul:
Merging master in when we restore caches can break after icon changes....

PR #18922, opened 2025-12-19
Fixes synths sometimes having permanant disfigurement and an inconsistancy with self repair

ReoDaProtovali:
## About The Pull Request

If you or someone else repairs you without using the surgical method and your limb was disfigured, that method of healing would not fix the disfigurement, resulting in having fully repaired bodyparts that were still considered disfigured. This was primarily an issue with the head, which makes you unidentifiable. This PR aims to solve that by clearing disfigurement when the limb is fully healed. If you STILL somehow end up healthy but disfigured, you can heal the limb with this method anyways to fix the disfigurement. It also makes it so synths can heal themself on the same limb they're holding something with, because it's tedious and the same restrictions do not apply to organic mobs applying topical like brute/burn kits, quality of life.
disfigured_protovali

## Changelog
:cl:
qol: removed the same-limb restriction on synth limb self repairs
fix: fixed synths being able to get disfigured at full health by being non-surgically repaired.
/:cl:


TeleTubby2:
Can't speak for the QOL stuff (asides from the edge case of now being able to repair with one arm left, IMO should be applied to the organic fix stuff rather than removed globally, but that's not my call) but I'm pretty sure the disfigurement remaining is intentional rather than a bug.

Organics have the same thing for burn (which is about as easy to fix, not counting infections) as it models the point where your face is melted, which just fixing tissue damage won't resolve. In the case of robotics as an analog, It's safe to presume it has the same intention behind it.

you might be able to do a bubba'd field fix to get your car back to snuff after a fender bender and have it work perfectly fine, but it'll look like ass and have some nasty dents until you either get bodywork done or replace a few panels. In the case of robots, I'd imagine that'd be work that'd be pretty difficult to do coherently on yourself

ReoDaProtovali:
it's probably not intentional because it goes away in the equivalent healing method done via surgery. You can do that if you're fully healed. So in theory, to fix a fully healed, disfigured limb, you'd have to DAMAGE it again, just to fix it surgically. and if that is intentional, then it's a change for the better anyways, because that's a bad intention.

Synthetics, unlike organics do not have a dedicated "fix disfigurement" surgery, and its' bundled with the "fix repair" surgery. it also fixes it every time you attempt to repair it that way. I made it so you have to fully heal the part to get the disfigurement fix. Arguably, if you're "Bubba field fixing" yourself, you should only beable to heal extreme damage, and not minor damage, but it's the other way around. ~~Not that I'd ever want that, god forbid I have to go to science for every 1-4 damage I get from a rat.~~

Cameron-The-Raven:
> does anyone even read commit details on PRs?

Yes


Also, I don't see any problem with removing the 'one limb' restriction...It was honestly kind of dumb because you can do surgery on your own hand, heal your own hand, but synths can't heal theirself with one hand. 

Polaris was extremely afraid of FBPs being strong and that kind of stuff shows, here. But we have proteans so any FBP balance complaints are genuinely out the window until FBPs reach protean stage of strength.


In terms of the disfigurement fixing, I don't see any issue with them being able to fix disfigurement.

ReoDaProtovali:
...how is this balance?

Willburd:
I'm all for removing the restriction on self repair hand. Pretty much none of carbon code cares about what hand is used, it feels unneeded and purely to spite synths as a form of "balance" from polaris. Either all of carbon code gets the same treatment for surgery and firstaid or fbps get the check removed in my opinion.

As for disfigurement, either add a surgery for it or do this. One or the other. I'm not even sure there's a way to fix disfigurement on synths currently.

ReoDaProtovali:
I said this before, but the normal surgical repair for synths will fix disfigurement, it's not specifically mentioned and it's purely a bonus effect ontop of what it already does. As a result of that, it's able to get to full health through normal repair and still be disfigured. this change is done so if you get to full health through non-invasive repair, it'll fix the disfigurement anyways, since you're already topped-up.

ReoDaProtovali:
of note: just like normal repair before this PR, invasive surgery cant be done if you're at full health

PR #18926, opened 2025-12-21
Makes bio augments not infectable by default

Cameron-The-Raven:
## About The Pull Request
Makes bio augments not infectable by default.

The only bio augment that IS infectable is the locomotion implant.
## Changelog
:cl: Diana
fix: Medical implant will no longer become infected
/:cl:


Randulf0:
common chimera W

PR #18928, opened 2025-12-21
Makes R&D artifacts not just 'destroy them for the rocks'

Cameron-The-Raven:
## About The Pull Request
Makes it so that archeological artifacts (anything dug out of rocks) have an inherent value when deconstructed in the destructive analyzed. This value is between 20-60 (between 20 to 60 seconds worth of points)

This makes it so archeology is no longer stuck between the decision of "Do I destroy this artifact so I can deconstruct the 'broken rock' or do I excavate it and can't deconstruct it even if it's useful?"
## Changelog
:cl: Diana
qol: Archeological finds can now be deconstructed for points instead of having to destroy the artifact before you know what it is.
/:cl:


ShadowLarkens:
Love the idea, don't necessarily love the /obj/item vars, is there any way we can avoid them?

Cameron-The-Raven:
> Love the idea, don't necessarily love the /obj/item vars, is there any way we can avoid them?

Will look into possibly making it an element

PR #18935, opened 2025-12-24
Vore test runtime fix

Willburd:
## About The Pull Request
Doesn't appear to apply to upstream, but if the first map loaded is unsuitable for unit testing, and the turf scan looking for a valid turf encounters a door. It will runtime. As zas does not place zones in door turfs. Fixes a downstream runtime where our underground uses planet atmos, and pois that spawn in it runtime when the turfscan reaches the doors in them.

documentation of a normal door ingame with no zone:
prdoor

## Changelog
Adds a zone nullcheck when locating a valid turf to place the test mob on.

:cl: Willburd
fix: runtime during vore unit tests on some downstream maps
/:cl:

### DIFF
--- a/code/modules/unit_tests/vore_tests.dm
+++ b/code/modules/unit_tests/vore_tests.dm
@@ -1,6 +1,8 @@
 /datum/unit_test/proc/create_test_human(turf/loc = null)
 	if(!loc)
 		for(var/turf/simulated/floor/tiled/T in world)
+			if(!T.zone)
+				continue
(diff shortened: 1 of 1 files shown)


Willburd:
Currently appears to be related to a strange zas issue on the map itself. DNM please.

Willburd:
Updated with new info after investigation. PR appears to be a valid issue, just very isolated to specific conditions of the map itself.

PR #18943, opened 2025-12-26
optimizes transfer

Kashargul:
## Changelog
:cl:
code: moves all forceMoves for vore code to nom_atom
fix: limits belly entered messages and transfer lookups
/:cl:


code/modules/vore/eating/belly_obj_vr.dm
```diff
@@ -516,13 +516,20 @@
...
-
 // Called whenever an atom enters this belly
 /obj/belly/Entered(atom/movable/thing, atom/OldLoc)
 	. = ..()
 	if(!owner)
 		thing.forceMove(get_turf(src))
 		return
+	if(length(contents) > 180)
+		to_chat(owner, span_userdanger("Your belly [src] contains more than 180 items, keep the count below the limit of 200 or get removed from the round."))
+	else if(length(contents) > 200)
```
Cameron-The-Raven:
```suggestion
	if(length(contents) > BELLY_CONTENT_LIMIT * 0.9)
		to_chat(owner, span_userdanger("Your belly [src] contains more than [BELLY_CONTENT_LIMIT * 0.9] items, keep the count below the limit of [BELLY_CONTENT_LIMIT] or get removed from the round."))
	else if(length(contents) > BELLY_CONTENT_LIMIT)
```

code/modules/vore/eating/belly_obj_vr.dm
```diff
@@ -516,13 +516,20 @@
...
+		to_chat(owner, span_userdanger("You've ingested over 200 items into your belly [src] and got kicked for trying to lag the server."))
+		log_and_message_admins("tried to ingest more than than the sane limit of 200 items.", owner)
```
Cameron-The-Raven:
```suggestion
		to_chat(owner, span_userdanger("You've ingested over [BELLY_CONTENT_LIMIT] items into your belly [src] and got kicked for trying to lag the server."))
		log_and_message_admins("tried to ingest more than than the sane limit of [BELLY_CONTENT_LIMIT] items.", owner)
```

code/modules/vore/eating/belly_obj_vr.dm
```diff
@@ -516,13 +516,20 @@
...
 
-
 // Called whenever an atom enters this belly
 /obj/belly/Entered(atom/movable/thing, atom/OldLoc)
 	. = ..()
 	if(!owner)
 		thing.forceMove(get_turf(src))
 		return
+	if(length(contents) > BELLY_CONTENT_LIMIT - 20)
+		to_chat(owner, span_userdanger("Your belly [src] contains more than 180 items, keep the count below the limit of 200 or get removed from the round."))
```
Cameron-The-Raven:
```suggestion
	if(length(contents) > BELLY_CONTENT_LIMIT * 0.9)
		to_chat(owner, span_userdanger("Your belly [src] contains more than [BELLY_CONTENT_LIMIT * 0.9] items, keep the count below the limit of [BELLY_CONTENT_LIMIT] or get removed from the round."))
```

ReoDaProtovali:
200 is a rather small hard limit. Especially if it's going to be throwing ban threats at players for even getting near it. Regardless. The issue is transfers, not having a large volume of items. If anything, there should be a limit to how many items you can transfer (manually or not) at once to mitigate server lag. It'll also be a lot less intrusive than "you've doing vore too much. Stop or eat a ban."

Kashargul:
> 200 is a rather small hard limit. Especially if it's going to be throwing ban threats at players for even getting near it. Regardless. The issue is transfers, not having a large volume of items. If anything, there should be a limit to how many items you can transfer (manually or not) at once to mitigate server lag. It'll also be a lot less intrusive than "you've doing vore too much. Stop or eat a ban."

200 is more than enough and already too high if you turbo mode digest those. Earlier considerations even suggested merely 100 or less.

ReoDaProtovali:
Under the assumption items are always being digested at maxium speed. This is a really poor solution to a problem that mostly arises from processing the extreme volume of items at a time, opposed to setting a limit to how many items could be affected at once (be it digestion, transfers or other such things.) I do think a limit of 200 (or even 100) on how many entries could be processed at once would be completely reasonable. A limit of 200 things in a belly is not. I've hit ~300 completely naturally as a swoopie and that was while digesting. At the very least, id argue a hard limit of 500 would be more reasonable, but im against the idea of a hard limit in bellies in general (atleast for the number of things you can put in them.) Especially when certain things like items dropped by prey results in them literally falling out of the belly with this system if they'd otherwise surpass the cap.

The scare factor in approaching the limit is also still very not cool, I think making players afraid of doing vore and reaching a large volume of items possibly unintentionally is very bad.

Kashargul:
> Under the assumption items are always being digested at maxium speed. This is a really poor solution to a problem that mostly arises from processing the extreme volume of items at a time, opposed to setting a limit to how many items could be affected at once (be it digestion, transfers or other such things.) I do think a limit of 200 (or even 100) on how many entries could be processed at once would be completely reasonable. A limit of 200 things in a belly is not. I've hit ~300 completely naturally as a swoopie and that was while digesting. At the very least, id argue a hard limit of 500 would be more reasonable, but im against the idea of a hard limit in bellies in general (atleast for the number of things you can put in them.) Especially when certain things like items dropped by prey results in them literally falling out of the belly with this system if they'd otherwise surpass the cap.
> 
> The scare factor in approaching the limit is also still very not cool, I think making players afraid of doing vore and reaching a large volume of items possibly unintentionally is very bad.

It's known that you like to cause issues. You are free to rewrite the entire processing code. But having 500 items ingested is not any normal gameplay situation. That's more power gaming by that and nothing that we want to encourage anyway.

Cameron-The-Raven (approved):

PR #18946, opened 2025-12-27
Resleever minifix

Cameron-The-Raven:
## About The Pull Request
Makes it so that if outside forces (i.e ID passing resleever) gives you the resleever prompt but you accidentally close it (via typing in chat when the popup comes up) it allows you to manually click it to resleeve.
## Changelog
:cl: Diana
qol: You can manually resleeve with the autosleever if you're given the prompt and accidentally close it.
/:cl:


Kashargul:
An accidental close would be a null return. You are doing this even on intentional rejections. Intended?

Cameron-The-Raven:
> An accidental close would be a null return. You are doing this even on intentional rejections. Intended?

Yep, it's meant to allow for them to re-decide on if they want to realeeve or not. The accidental close is "You are typing and when you press space, it's hovered over the "NO" button, meaning you automatically press NO" due to how the TGUI hijacks your screen no matter what you're doing

PR #18948, opened 2025-12-27
Allows non-contagious viruses to be picked up with blood

TheCaramelion:
## About The Pull Request

Makes it so non-contagious viruses are stored in the blood, but **not** spreadable by it
Adds info desc to the extrapolator to tell the user about changing modes

## Changelog
:cl: Guti
qol: Non-Contagious viruses can be picked up with a sample of blood
qol: Added extra info description to the extrapolator
/:cl:


TheCaramelion:
Mold burger enjoyers rejoice - Your Food Poisoning will no longer be invisible to others.

PR #18960, opened 2025-12-31
Adds the weather anomaly

TheCaramelion:
## About The Pull Request

Adds weather anomaly, with different kinds of weathers.

Rain. Cleans everything and wets tiles.
Storm. Cleans everything, wets tiles, small chance of lightning.
Blood Rain. Makes an absolute of a mess of the areas it is in.
Acid Rain. Melts items and people caught under it.
Hail. Pelts you.
Ash Storm. Temperature damage, makes floors dirty.

Once the anomaly spawns, it will alert everyone on the affected clouds on the zone, then start raining after 40ish~ seconds, giving time to move things/yourself, or get an umbrella and enjoy the indoors rain.

Anomalous weather datum can be changed to rain any reagents, and weather reagents may be collected from the weather anomaly.

## Changelog
:cl: Guti
add: Added weather anomaly
qol: Smoke from neutralized anomalies is now much short-lived
fix: Fixed spawned anomalies from telesci not having cores
/:cl:

### DIFF
--- a/code/game/objects/effects/anomalies/anomalies_weather_themes.dm
+++ b/code/game/objects/effects/anomalies/anomalies_weather_themes.dm
@@ -36,7 +36,7 @@
 	if(iswall(to_affect))
 		return
 
-	for(var/atom/thing as anything in to_affect)
+	for(var/atom/thing in to_affect)
 		if(isliving(thing))
 			affect_mob(thing)
 			continue
--- a/code/modules/clothing/suits/reactive_armour.dm
+++ b/code/modules/clothing/suits/reactive_armour.dm
@@ -338,7 +338,7 @@
 	shock_turf_windup(owner.loc)
 
 /obj/item/clothing/suit/armor/reactive/weather/proc/shock_turf_windup(turf/target)
-	addtimer(CALLBACK(GLOBAL_PROC_REF(lightning_strike), target))
+	addtimer(CALLBACK(GLOBAL_PROC_REF(lightning_strike), target), 1 SECOND)
 
 /obj/item/clothing/suit/armor/reactive/stealth
 	name = "reactive stealth armor"


SatinIsle:
Generally think this is very cool.
Uncertain about digestive rain, rather than having just acid rain with the same effect. Having it apply stomach acid specifically seems like it could lead to uncomfy pref situations.

code/game/objects/effects/anomalies/anomalies_weather_themes.dm
```diff
@@ -0,0 +1,188 @@
...
+		weather_colour = reagent_holder.get_color()
+
+	if(weather_colour)
+		visuals.color = weather_colour
+
+/datum/anomalous_weather/proc/affect_turf(turf/to_affect)
+	if(iswall(to_affect))
+		return
+
+	for(var/atom/thing as anything in to_affect)
```
Cameron-The-Raven: Is the 'as anything' needed here?

code/modules/clothing/suits/reactive_armour.dm
```diff
@@ -296,6 +297,49 @@
...
+	playsound(src, 'sound/effects/lightningbolt.ogg', 33, TRUE)
+
+	new /obj/effect/effect/smoke/bad(loc)
+
+	if(!isopenturf(owner.loc))
+		return
+	shock_turf_windup(owner.loc)
+
+/obj/item/clothing/suit/armor/reactive/weather/proc/shock_turf_windup(turf/target)
+	addtimer(CALLBACK(GLOBAL_PROC_REF(lightning_strike), target))
```
Cameron-The-Raven: Why do we add a timer instead of just calling it?

TheCaramelion: Oops I forgot to actually give it the delay

PR #18967, opened 2026-01-01
Spontaneous Prey Bellies

SatinIsle:
## About The Pull Request

This adds the ability to choose which bellies spontaneous prey will end up in. If left null, they will go into the preds currently selected belly as per usual.

As per request, this is direction based on the preds current direction, allowing players to choose different bellies for when they are facing towards, away from or left/right of their prey when the spont vore happens.

## Changelog
:cl:
add: This adds the ability to choose which bellies spontaneous prey will end up in. If left null, they will go into the preds currently selected belly as per usual. Allows for 4 directions.
fix: some typescript fixes (by kash)
/:cl:


Kashargul:
@Cameron-The-Raven we have to check if the direction logic is working or needs an edit

Cameron-The-Raven:
Made a fix on this to fix throw vore, as it was broken and not using the new directional belly system.

Cameron-The-Raven:
Deconflicted it as well.

PR #18974, opened 2026-01-03
Adds new symptoms for virology [WIP]

Cameron-The-Raven:
## About The Pull Request
WIP PR. Putting this here just for ease of access and a reminder.

### Anaerobic Resuscitation
- Brain damage over time. Isn't lethal unless the resistance is high enough, in which case it will kill them eventually.
- Has some various side effects to go along with the brain damage that depends on the stage. Various things from dropping your hands, becoming confused, 

### Bone Dysplasia
- At max stage, the host's bones become 50% stronger.
- If at max stage and cured, the host's bones revert back to their original strength, with every one of them fracturing in the process.

### Cytopathic Reanimation
- Stealthy symptom that causes the disease to be able to spread and process in the dead.
- Does nothing until the following criteria is met:
1. Has reached max stage.
2. The host is dead

Once those criteria are met, the corpse will slowly begin to heal external damage and bloodloss while taking brain damage, up to a point. When the corpse's health is healed below hardcrit's worth of damage, it'll revive with a dose of zombiepowder in its system. 

After they've been revived, they'll occasionally emote things such as drooling, twitching, choking, or (worst of all) randomly collapsing.

## TODO: 

Add a lot more symptoms. 

Probably make the Cytopathic Reanimation symptom do some fucky stuff with your organs, internal and externally. Make it so you have to have certain stats to keep it from becoming a zombie virus.
## Changelog
:cl: Diana
add: Adds some new symptoms for virology! 
/:cl:


code/datums/diseases/advance/symptoms/necroa.dm
```diff
@@ -0,0 +1,131 @@
...
+Bonus
+	Allows spread and processing in the dead.
+	Allows resurrection of the dead.
+
+//////////////////////////////////////
+*/
+
+/datum/symptom/necroa
+	name = "Cytopathic Reanimation"
+	desc = "Allows for reanimation of dead tissue and corpses. The virus can infect and spread through corpses."
```
zeskorion, line 21: the second ability seems a bit redundant with necrotic metabolism, or whatever it's called here. I suggest being very, very careful with this symptom- it'll make certain buffs to viro gameplay, such as trifectas, a lot harder, simply because *it* exists now

code/datums/diseases/advance/symptoms/necroa.dm
```diff
@@ -0,0 +1,131 @@
...
+			limb.status = 0
+
+	//Big checks to see if there's a reason we CAN'T revive.
+	if(!check_for_valid_revival(infectee))
+		return
+	//If we pass our checks, let's go ahead and revive!
+	revive(infectee)
+
+///Revive proc shamelessly stolen from /datum/modifier/redspace_corruption
+/datum/symptom/necroa/proc/revive(mob/living/carbon/human/M)
```
zeskorion, line 96: I Think this beneficial effect should be a threshold, perhaps on necrotic metabolism?

code/datums/diseases/advance/symptoms/bone_dysplasia.dm
```diff
@@ -0,0 +1,62 @@
...
+		return
+	if(A.stage < 5)
+		return
+	if(!triggered)
+		triggered = TRUE
+		to_chat(infectee, span_notice("You feel a strange sensation in your bones as they harden and thicken."))
+		for(var/obj/item/organ/external/organ in infectee.organs)
+			if(organ.cannot_break)
+				continue
+			organ.min_broken_damage *= 1.5
```
zeskorion, line 62: does this continue ramping until they become entirely unbreakable? As with other symptoms, i recommend gating  any huge beneficial effects beyond a threshold

zeskorion (commented):
Virology is what I started my dev stuff with, so I have a *lot* of opinions, but I'm gonna try not to get too deep in da weeds here.

I suggest most highly beneficial effects be hidden behind thresholds, and viruses otherwise act somewhat benign. Otherwise, you can simply mindlessly combine most symptoms with positive effects in one virus,

code/datums/diseases/advance/symptoms/necroa.dm
Cameron-The-Raven: Yeah, I want to have necrotic metabolism still be it's own thing ('this virus processes and spreads by the dead') with no downsides.

For this one, I plan on it having the current upsides and some **really** strong downsides that require thresholds to move 'remove' those downsides.
I.e: 
If revived and resistance isn't above X amount, random organs will begin to reject (Gets rid of a downside)
If stealth is above X amount, the zombie powder is replaced with lich powder (beneficial to having high stealth)
If revived and resistance isn't above X amount, a random organ is chosen to die() (Gets rid of another downside, requires a higher resistance)

code/datums/diseases/advance/symptoms/bone_dysplasia.dm
Cameron-The-Raven: Nah, at the current moment it's a one time thing. It essentially gives the 'dense bones' trait. The 'triggered' var keeps it from continually proccing itself.

I could make it so it scales based on resistance as well, which could possibly make it so if the disease has a _negative_ resistance, it actually weakens the person's bones. If resistance is >=5, then it does the 'bones break when healed' thing.

code/datums/diseases/advance/symptoms/necroa.dm
Cameron-The-Raven: Could make it a threshold, yeah. I'm open to suggestions on what the thresholds should be, as I haven't gotten to sit down and actually work out numbers and min/max that stats can get up to with what combinations.

code/datums/diseases/advance/symptoms/bone_dysplasia.dm
zeskorion: Thresholds is generally better than fully dynamic scaling, as it gives something to aim for- but it's a decent idea. Just hardcap the values.

code/datums/diseases/advance/symptoms/necroa.dm
zeskorion: I would really want to further tailor virus stats in the code, from what i know they're a bit iffy rn.

code/datums/diseases/advance/symptoms/necroa.dm
zeskorion: Interesting to see how this works out. I had a similar symptom on beeviro, called Hemetophagy. Required blood be around to revive the host- but it also needed the virus to *have* process dead. Perhaps this could be a niche heal, or even a negative symptom, that *requires* processing on dead to work?

I'm spouting ideas for this symptom, mind you, not critiques. I'm not invested in virocode right now enough to analyze its place in the meta

Cameron-The-Raven: Hmm, yeah, that's a good idea. I could easily remove the innate 'dead processing' from it so if you want to have it process in corpses and do it's special 'dead only' effects, you _have_ to combine it with the dead processing symptom.

It does make numbers and thresholds harder though since it means you're going to guaranteed have 2/6 of your symptom slots taken up w/ this and dead processing.

It'd essentially make it so:
W/o dead processing:
Alive - nothing
Dead - it does nothing
Died and revived - they begin to have all the super negative side effects happening (unless they've been nullified with resistance)

I'd have to check on that last one to see if viruses linger after being revived.

W/ dead processing: 
Alive - nothing
Dead - Begin to do the gradual healing over time. (If resistance = X, it will also revive them.)
Died and revived - Same as above.


I'll have to work out numbers (as we only have 6 symptoms max, 1 disease max) and the side effects/downsides of it. Because this is meant to be a pretty **nasty** symptom that if you're alive is completely harmless but when someone dies it essentially becomes a zombie virus. Sure, you get the upside that you can keep coming back alive, but you have the downside that unless you have a _bunch_ of medical care (or an extremely tailored virus) you're going to die again.

I might make it so that if you revive it always guarantees you have 95 (out of 100) brain damage, but that also has the issue of 'you'll immediately suffocate to death' because of how our brain-blood code works.

zeskorion: I think the real potential of this, and a lot of other symptoms, would be if the 3-virus thing I did back in beeviro was done. It really opens up viro gameplay for *fun*. 

Otherwise, i'll defer any choices of implementation to you, ultimately.

Cameron-The-Raven: Yeah viro atm is kinda a dead department with a good part of that reason being because of the 1 virus cap we currently have. You can make _ONE_ good virus and...that's it. No mixing and matching, just 6 symptoms max. It really kills creativity.

Pretty much because of that 1 virus cap, any symptoms are heavily hampered to their usefulness because you can't have: 
"Virus 1: Process dead, revives if dead, 4 misc symptoms to increase resistance" 
"Virus 2: Get rid of brain damage, convert the lingering damage to toxins and have the virus heal it"
"Virus 3: Process dead, symptom to regrow limbs, filler symptoms to allow regrow limbs to be max strength"

I might just make this PR make it so the cap is 3 viruses like how it used to be before it was changed.1

Because atm it also has the exploit that if you give yourself 'the sniffles' off just a virus that makes you cough, you're immune to any other viruses that come your way as well.

zeskorion: How i did it with beeviro was that you get 3 viruses. One must be of a severity 0 or lower. One must be between 1 and 4 severity. One must be of 5 severity or higher

It adds very interesting gameplay, to make a virus that neutralizes the lethal symptoms itself, so all 3 are stable simultaneously- without being super easy to powergame.

PR #18983, opened 2026-01-04
Fixes dark maw not eating captured prey

TheCaramelion:
## About The Pull Request

Adds an extra check to override if either prey/pred should be incorporeal on the vore helpers

## Changelog
:cl: Guti
fix: Fixed Dark Maw not eating captured prey, if prefs aligned
/:cl:

### DIFF
--- a/code/_helpers/vore.dm
+++ b/code/_helpers/vore.dm
@@ -4,7 +4,7 @@
 
 /// Most basic check of them all.
 /// Checks if PRED can eat PREY.
-/proc/can_vore(mob/living/pred, mob/living/prey)
+/proc/can_vore(mob/living/pred, mob/living/prey, allow_incorporeal = FALSE)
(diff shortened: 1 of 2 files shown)


code/_helpers/vore.dm
```diff
@@ -14,7 +14,8 @@
 	if(!is_vore_predator(pred))
 		return FALSE
 	if(prey.is_incorporeal() || pred.is_incorporeal())
-		return FALSE
+		if(!allow_incorporeal)
```
Kashargul, line 17: that's bypassing two way?

TheCaramelion: Could be done just for the pred, currently the only case that needs to skip it is the dark maw

Kashargul: please adjust, unless you want it to be funny kin traps too...

PR #18996, opened 2026-01-07
Mech Maintenance Quickfix

KillianKirilenko:
## About The Pull Request

Quick fix for mech maintenance causing client lockups and possible server issues.

## Changelog

:cl:
fix: fixed mech maintenance mode causing issues for clients and server whilst trying to move
/:cl:


Kashargul:
There's 2 more messages in that proc that could cause issues

PR #19000, opened 2026-01-09
Intent based struggles

Kashargul:
## Changelog
:cl:
qol: prey struggles are now intent based
/:cl:


Rick4242:
So does that mean struggling while on green intent will no longer act as an escape attempt? Or is there some other way this works?

Kashargul:
> So does that mean struggling while on green intent will no longer act as an escape attempt? Or is there some other way this works?

Quite the case

PR #19002, opened 2026-01-09
Virology Update #4

TheCaramelion:
## About The Pull Request

New year, new broken virology PR.

The most important change of them all. **You can have up to 3 viruses now**, as long as you can play with the slots.
Slot 1 is for diseases with 0 or less severity
Slot 2 is for diseases with 1 to 4 severity
Slot 3 is for diseases with 5 or more severity
If a virus goes to the same slot that a virus is already occupying, they will fight for the spot. The attacker transmission against the defender's resistance*2

Another big change is the cure. They're all scrambled up now. Orange Juice will no longer be a heal-it-all for viruses with low transmission, instead it will be picked from a range of options ranged from **water** to **does that chem even exist??**. Because we love making Chemists work in this house.

Oh and injecting yourself with a virus will infect you, 100% of the time. No silly check for your clothes if you inject it.

Apart from that, a couple of viruses have been slightly tweaked, and Narcolepsy has been added.

## Changelog
:cl: Guti
balance: You can now have up to 3 viruses in your body.
balance: Viruses will have a varied range of reagents for their cures now.
balance: Removed nutriment-based virus protection.
balance: Cure chance lowered to 5-10
balance: Headache symptom stun reduced from 15 to 5
balance: Incremented the power of Tissue Hydration
balance: Reduced the severity of Weight Loss symptom to 1
balance: Spyndrome's stealth changed to -1, resistance to 3
add: Added Narcolepsy symptom
fix: Fixed admin-created viruses having the developer warning name
fix: Injecting diseased blood will no longer check for clothing immunity
code: Cleaned up a few old symptom files
code: Added comparators for virus resistance and symptom names
spellcheck: Corrected some spelling mistakes across various symptoms
/:cl:

### DIFF
--- a/code/datums/diseases/_disease.dm
+++ b/code/datums/diseases/_disease.dm
@@ -125,7 +125,7 @@ GLOBAL_LIST_INIT(diseases, subtypesof(/datum/disease))
 	if(affected_mob.stat == DEAD && !global_flag_check(virus_modifiers, SPREAD_DEAD) && !force_spread)
 		return
 
-	if(affected_mob.reagents.has_reagent(REAGENT_ID_SPACEACILLIN) || (affected_mob.nutrition > 300 && prob(affected_mob.nutrition/50)))
+	if(affected_mob.reagents.has_reagent(REAGENT_ID_SPACEACILLIN))
 		return
 
 	var/spread_range = 2


TheCaramelion:
Goodbye my sweet GBP :(

SatinIsle:
Does this mean we can use viruses like pokemon to battle each other

TheCaramelion:
> Does this mean we can use viruses like pokemon to battle each other

Yes but the battles are so small that no one will be able to tell :(

PR #19005, opened 2026-01-10
remote view look upwards

Willburd:
dreamseeker_RY8YR19ePB

## About The Pull Request
Updates the look upward verb to give a remote view one z level above you (if possible). Lets you see in slightly smaller radius than normal. Also updates remote view to allow a view size being passed for any remote view subtype.

## Changelog
Lookup now starts remotely viewing the turf above you.
Remoteview arguments updated to pass in a view size(not just on zooming items)
Remoteview setting datum can now specify if a view type hides most of the hud on zoom

:cl: Will
qol: Look upward verb now lets you look upward one z level
code: remote views with zoom hude or different sizes are easier to start
/:cl:


Kashargul:
Needs testing:

Can people be interacted with from below?
Can people shoot through ceilings?

Willburd:
> Needs testing:
> 
> Can people be interacted with from below? Can people shoot through ceilings?

No and no, adjacency expects Z levels to be the same. Guns also don't shoot up Z levels. Me and diana messed around with it for a bit, at least the basic interactions are all blocked. I'm not sure if there are any weird cases with specific items, but that would be a failure of the item to do proper checks for range. Range also expects the same z level in most cases like orange() or view().

PR #19012, opened 2026-01-11
Votes throws an alert when called

TheCaramelion:
## About The Pull Request

When any kind of vote is called, it will also throw an alert at the top right of the user's corner, just like the hunger/cold/admin PM ones.

## Changelog
:cl: Guti
qol: Votes now throw an alert when active
/:cl:

### DIFF
--- a/code/modules/vote/vote_datum.dm
+++ b/code/modules/vote/vote_datum.dm
@@ -135,8 +135,6 @@
 	if(remaining() == 0)
 		var/result = calculate_result()
 		handle_result(result)
-		for(var/mob/mob in GLOB.player_list)
-			mob.clear_alert("vote")
 		qdel(src)
 
 /datum/vote/Destroy(force)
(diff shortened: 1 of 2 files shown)


code/modules/vote/vote_datum.dm
```diff
@@ -121,6 +121,8 @@
 	to_chat(world, span_lightpurple("Type <b>vote</b> or click <a href='byond://?src=\ref[src];[HrefToken()];vote=open'>here</a> to place your vote. \
 		You have [time/10] seconds to vote."))
 	world << sound('sound/ambience/alarm4.ogg', repeat = 0, wait = 0, volume = 50, channel = 3)
+	for(var/mob/mob in GLOB.player_list)
```
Kashargul, line 124: why can't you just attach it to the datum?

TheCaramelion: Unsure if I'm able to; throw_alert requires the typepath and initializing one attached to the datum throws a runtime

ShadowLarkens: yeah I'm pretty sure this is the best way to do it

PR #19018, opened 2026-01-13
The Great Slobber Update

ShadowLarkens:
## About The Pull Request
Adds two new features that interact with my new drippy wetstack system.

### First: New Trait: Major Slobberer

Basically it just makes the `Lick` verb mildly wetten your target (assuming no protection bubble)

### Second: New Belly Addon: Makes Prey Wet
Constantly wettens your prey per belly cycle. They will continue to drip for a bit when let out. 

## Changelog

:cl:
add: Major Slobberer trait, adds wet stacks on lick and makes people drippy.
add: Makes Prey Wet belly addon, adds wet stacks to prey in a belly and makes them drippy. Basically irrelevant unless you spit them back out.
/:cl:

### DIFF
--- a/tgui/packages/tgui/interfaces/VorePanel/constants.ts
+++ b/tgui/packages/tgui/interfaces/VorePanel/constants.ts
@@ -85,8 +85,9 @@ export const modeToTooltip = {
   'Slow Body Digestion': 'Continues to digest a body after the prey has died.',
   'Muffle Items': ' Muffles noise from items inside the vorgan.',
   'TURBO MODE': 'Heavily increases tick speed of the vorgan (6x).',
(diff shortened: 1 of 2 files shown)


Kashargul:
UI side tooltip is missing

tgui/packages/tgui/interfaces/VorePanel/constants.ts
```diff
@@ -86,6 +86,7 @@ export const modeToTooltip = {
   'Muffle Items': ' Muffles noise from items inside the vorgan.',
   'TURBO MODE': 'Heavily increases tick speed of the vorgan (6x).',
   'Absorbed Devour': 'Allows absorbed prey to devour other prey.',
```
Kashargul: WHile having a PR touching that... that key might need a fix

PR #19020, opened 2026-01-13
asset loading TM

Kashargul:
By moving the asset gen to invoke async, we no longer mess up the susbsystem in tick checks... but we might have caused multithreading for icon generation... Maybe linux really isn't too happy about parallel file access from multiple threads. this is a TM PR only to test the theory

Reminder, undo `Configuration/EventScripts/PreCompile.sh` after the tests, delete the `rust-g` folder to have it redownload


Kashargul:
Live has been reset to the old rust g config

PR #19038, opened 2026-01-16
Allows stripping people by resizing

SatinIsle:
## About The Pull Request

Added a new preference that allows you to drop all of your items and possibly underwear upon being suddenly resized by more than 30%.

This also adds a variable to resize() that determines whether a specific method can trigger the stripping.
It is currently enabled on size guns and bluespace crackers only, I want to add it on a case by case basis.

## Changelog
:cl:
add: Added a new preference that allows you to drop all of your items and possibly underwear upon being suddenly resized by more than 30%.
fix: Fixed temperature damage toggle saving.
fix: Bluespace collars changing into shock collar sprites on toggling power.
/:cl:

### DIFF
--- a/code/modules/vore/eating/belly_obj.dm
+++ b/code/modules/vore/eating/belly_obj.dm
@@ -296,6 +296,7 @@
 	"digest_tox",
 	"digest_clone",
 	"bellytemperature",
+	"temperature_damage",
 	"immutable",
 	"can_taste",
 	"escapable",
@@ -1325,6 +1326,7 @@
 	dupe.digest_tox = digest_tox
 	dupe.digest_clone = digest_clone
 	dupe.bellytemperature = bellytemperature
+	dupe.temperature_damage = temperature_damage
 	dupe.immutable = immutable
 	dupe.can_taste = can_taste
 	dupe.escapable = escapable


SatinIsle:
I will need to check a few items like rig suits to check they behave okay

code/__defines/vore_prefs.dm
```diff
@@ -12,6 +12,7 @@
 	target.noisy = source.noisy;														\
 	target.show_vore_fx = source.show_vore_fx;											\
 	target.selective_preference = source.selective_preference;							\
+	target.size_strip_preference = source.size_strip_preference;									\
```
Kashargul: formatting again is off

code/__defines/size.dm
```diff
@@ -1,3 +1,8 @@
 
 #define MAX_VOICE_FREQ 70000
 #define MIN_VOICE_FREQ 15000
+
+//Size change stripping options
+#define SIZESTRIP_NONE 				"Drop Nothing"								//Default, no dropping items on size change
```
Kashargul: why are those strings... can we please not. Use integers

code/modules/vore/eating/vorepanel_vr.dm
```diff
@@ -656,6 +656,14 @@
 				host.client.prefs_vr.selective_preference = host.selective_preference
 			unsaved_changes = TRUE
 			return TRUE
+		if("switch_strip_mode_pref")
+			host.size_strip_preference = tgui_input_list(ui.user, "What would you prefer happen to you when your size is changed by over 30%?","Selective Bellymode", list(SIZESTRIP_NONE, SIZESTRIP_ITEMS, SIZESTRIP_ALL))
```
Kashargul: should be handled UI side as well, so that we don't need to worry about strings in save files...

code/__defines/vore_prefs.dm
SatinIsle: Oops, I originally gave the variable a poor name, then regex editted it to something better and forgot this bit

code/__defines/size.dm
SatinIsle: Okay, made these into numbers and made it so the UI works with them.

code/modules/vore/eating/vorepanel_vr.dm
ShadowLarkens: still needs to be addressed @SatinIsle

SatinIsle: > still needs to be addressed @SatinIsle

Do you have any advice about this?
I made changes so that the save files no longer use strings (it's now stored as a number)

ShadowLarkens: I believe Kasha wants this to be a dropdown in the UI, see these examples for how to set up one:

https://github.com/VOREStation/VOREStation/blob/master/tgui/packages/tgui/interfaces/VorePanel/VoreUserGeneralTabs/VoreUserPreferencesSpont.tsx#L57-L68
https://github.com/VOREStation/VOREStation/blob/master/code/modules/vore/eating/vorepanel_vr.dm#L1422-L1431

SatinIsle: I'll try and set this up when I have time, just might take a bit for me to wrap my head around!

Kashargul: I can take a look sometime next week just been busy with some runtime hunting and trying to fix the rurst g race condition

SatinIsle: > I can take a look sometime next week just been busy with some runtime hunting and trying to fix the rurst g race condition

No worries, I think I have got it working now, took me some time to wrap my head around it!

SatinIsle:
Some pics to show how the UI looks ingame, it fits pretty neatly with everything else:
image
image
image

code/modules/vore/eating/vorepanel_vr.dm
Kashargul: Ye, might be good to move the other one as well .. the selective one

tgui/packages/tgui/interfaces/VorePanel/constants.ts
```diff
@@ -39,6 +39,19 @@ export const reagentToColor = {
   Ethanol: undefined,
 } as const;
 
+export const stripModeToColor = {
```
Kashargul: likely should be an object Record<number, {displayText:string, color: string}>

code/modules/clothing/accessories/accessory_vr.dm
```diff
@@ -362,6 +362,10 @@
 	return data
 
 /obj/item/clothing/accessory/collar/shock/bluespace/tgui_act(action, list/params, datum/tgui/ui, datum/tgui_state/state)
+	if(action == "power") //No multiple sprites for these
```
Kashargul: wtf is that code .... please just return TRUE... that even should just go below into the switch action

PR #19058, opened 2026-01-18
Viro Fixes

TheCaramelion:
## About The Pull Request

Fixes being able to contract the same disease up to three times, and viruses not getting deleted when the mob is.

## Changelog
:cl: Guti
fix: Fixed being able to contract the same disease multiple times.
fix: Fixed viruses not removing their mob reference when the mob was deleted.
/:cl:

### DIFF
--- a/code/modules/mob/mob.dm
+++ b/code/modules/mob/mob.dm
@@ -26,8 +26,7 @@
 	QDEL_NULL(hud_used)
 	for(var/key in alerts) //clear out alerts
 		clear_alert(key)
-	if(LAZYLEN(viruses))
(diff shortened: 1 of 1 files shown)


code/modules/mob/mob.dm
```diff
@@ -26,6 +26,8 @@
 	QDEL_NULL(hud_used)
 	for(var/key in alerts) //clear out alerts
 		clear_alert(key)
+	if(LAZYLEN(viruses))
```
Kashargul: Why? QDEL_LIST_NULL is self safe?

PR #19081, opened 2026-01-23
Virology Patch

TheCaramelion:
## About The Pull Request

Bunch of fixes after an... Eventful shift.

## Changelog
:cl: Guti
fix: Fixed mlemington activating the mlem spree at any time
fix: Fixed narcolepsy missing a description
fix: Fixed Pandemic UI not showing the severity of a virus
fix: Fixed viruses changing cures when taking it out of a mob
fix: Fixed the PanDEMIC showing the ID for the reagent instead of the name as the cure
spellcheck: Removed "the" when Biometallic Replication symptom picked any organ
code: Added IMMUTABLE flag for viruses, to not allow them to change
/:cl:

### DIFF
--- a/code/datums/diseases/advance/advance.dm
+++ b/code/datums/diseases/advance/advance.dm
@@ -294,6 +294,8 @@ GLOBAL_LIST_INIT(advance_cures, list(
 		var/res = clamp(resistance - (length(symptoms) / 2), 1, length(GLOB.advance_cures))
 		cures = list(pick(GLOB.advance_cures[res]))
 		var/datum/reagent/cure_reag = SSchemistry.chemical_reagents[cures[1]]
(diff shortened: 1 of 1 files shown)


code/datums/diseases/advance/advance.dm
```diff
@@ -164,7 +171,13 @@ GLOBAL_LIST_INIT(advance_cures, list(
 /datum/disease/advance/proc/Refresh(new_name = FALSE, archive = FALSE)
 	GenerateProperties()
 	AssignProperties()
-	id = null
+	if(s_processing && symptoms && symptoms.len)
```
ShadowLarkens:
```suggestion
	if(s_processing && length(symptoms))
```

PR #19091, opened 2026-01-25
bunch of global vars

Kashargul:
## Changelog
:cl:
code: converts a bunch of variables to GLOB
fix: a few runtimes
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/station_special_abilities.dm
+++ b/code/modules/mob/living/carbon/human/species/station/station_special_abilities.dm
@@ -1301,7 +1301,6 @@
 	set desc = "you can lay Eggs"
 	set category = "Abilities.General"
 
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/carbon/human/species/station/station_special_abilities.dm
```diff
@@ -1296,16 +1296,12 @@
...
-//egglaying
-var/eggs = 0
-
-
 /mob/living/proc/mobegglaying()
 	set name = "Egg laying"
 	set desc = "you can lay Eggs"
 	set category = "Abilities.General"
 
+	var/eggs = 0
```
ShadowLarkens: i think this is meant to be ON /mob/living so that they can... produce eggs and then lay them

PR #19098, opened 2026-01-26
some more globals

Kashargul:
## Changelog
:cl:
code: converts the next set of globals to GLOBs
/:cl:

### DIFF
--- a/code/modules/asset_cache/assets/preferences.dm
+++ b/code/modules/asset_cache/assets/preferences.dm
@@ -1,4 +1,5 @@
 /datum/asset/simple/preferences
+	var/preview_icons = 'icons/mob/human_races/preview.dmi'
 	assets = list(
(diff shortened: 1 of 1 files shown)


code/modules/asset_cache/assets/preferences.dm
```diff
@@ -67,3 +67,7 @@
 			preference_data[preference_entry.savefile_key] = data
 
 	return preference_data
+
+// Not actually used; just forces this into the RSC for TGUI.
+/datum/prefrence_icons
+	var/preview_icons = 'icons/mob/human_races/preview.dmi'
```
ShadowLarkens: this should probably just go on `/datum/asset/simple/preferences` as a var

ShadowLarkens (changes requested):

code/modules/asset_cache/assets/preferences.dm
Kashargul: Ye, might be easier

PR #19099, opened 2026-01-27
port ci update [No GBP]

Kashargul:
Port of https://github.com/tgstation/tgstation/pull/94989


LemonInTheDark:
been having some issues, may be wise to hold off

Kashargul:
> been having some issues, may be wise to hold off

Thanks for letting us know. Will take a look if time allows. Is it not firing at all?

Might be an issue with too shallow histories... going to try to just use fetch-depth 0

PR #19103, opened 2026-01-27
Make showers not process while inactive + a little QoL

ReoDaProtovali:
## About The Pull Request

Gives shower code a shower so it doesnt stink.
Showers are a bit nicer to use, and dont process when they're not being used !!!!

## Changelog
:cl:
qol: Showers now instantly run a wash call on things that go under them while active
code: Removed a fair few unnecessary variables on showers and replaced it with better code.
code: Showers can have their output reagent overridden, either by adminbus or subtypes.
code: Showers no longer run process() calls while inactive. This should free up a fair few SSmachines process calls that did literally nothing
/:cl:

### DIFF
--- a/code/game/objects/structures/watercloset.dm
+++ b/code/game/objects/structures/watercloset.dm
@@ -180,7 +180,7 @@
 /obj/machinery/shower/Destroy()
 	QDEL_NULL(soundloop)
 	QDEL_NULL(reagents)
-	STOP_MACHINE_PROCESSING()
+	STOP_MACHINE_PROCESSING(src)
 	return ..()
 
 //add heat controls? when emagged, you can freeze to death in it?


ShadowLarkens:
the shower will eventually run of water since it no longer replenishes

Willburd:
> the shower will eventually run of water since it no longer replenishes

A possible solution might just be to refill reagents during process, and only return PROCESS_KILL if the reagent reserve is full?

ReoDaProtovali:
it wont run out of reagents because it does not use reagents anymore. there's a parameter for copy, which makes it just do the reaction without subtracting reagents from the source

Willburd:
> it wont run out of reagents because it does not use reagents anymore. there's a parameter for copy, which makes it just do the reaction without subtracting reagents from the source

Looking at the copy argument for reagent splash, that should be just fine then. All it does is prevent remove_reagent() from being called and doesn't call handle_reactions() during trans_to_holder().

PR #19115, opened 2026-01-30
starts_with var fixes for #19112

Willburd:
Corrected maps and new prefabs for #19112


Willburd:
wrong target, ignore me.

PR #19120, opened 2026-01-31
Conversion of realname and nickname from bay

NovaDusklight:
## About The Pull Request

This PR attempts to fully convert the player name and nickname from bay to TG.
Everything should work as normal otherwise.

I've tested these are best I can, and everything appears normal.
Review of the implementation would be appreciated, as this is my first attempt at this sort of PR.

## Changelog

:cl: Nova
refactor: converted name and nickname prefs from bay to TG
/:cl:


code/modules/client/preference_setup/general/01_basic.dm
```diff
@@ -36,29 +32,36 @@
 /datum/category_item/player_setup_item/general/basic/sanitize_character()
 	pref.biological_gender  = sanitize_inlist(pref.biological_gender, get_genders(), pick(get_genders()))
 	pref.identifying_gender = (pref.identifying_gender in all_genders_define_list) ? pref.identifying_gender : pref.biological_gender
-	pref.real_name		= sanitize_name(pref.real_name, pref.species, is_FBP())
-	if(!pref.real_name)
-		pref.real_name      = random_name(pref.identifying_gender, pref.species)
-	pref.nickname		= sanitize_name(pref.nickname)
+	// Sanitize real_name - is this step necessary anymore?
+	var/current_name = pref.read_preference(/datum/preference/text/human/real_name)
```
ShadowLarkens: this should be handled in the /datum/preference https://github.com/tgstation/tgstation/blob/6e75bd2911526ff3f1b1fc38163aa905b06cead8/code/modules/client/preferences/names.dm#L52

ShadowLarkens (changes requested):

code/modules/client/preference_setup/general/01_basic.dm
NovaDusklight: I've just tested handling an implementation that more tries to follow how TG is doing it in the example shown.
Could you please give this a look and let me know if this is quite what you meant?
I was not entirely sure where to put the file, but I did separate the names logic handling, as TG's implementation looked a lot cleaner that way.

NovaDusklight: I didn't want to bloat preferences.dm, so separated it similarly to TG, but I remain uncertain about where I put the file. Whether to keep it there, or keep it in preference setup.

code/modules/client/preferences/names.dm
```diff
@@ -0,0 +1,92 @@
...
+	var/maximum_value_length = MAX_NAME_LEN
+	var/allow_numbers = FALSE
+	var/default_species = SPECIES_HUMAN
+
+/datum/preference/name/is_valid(value)
+	if(!istext(value))
+		return FALSE
+	if(length(value) > maximum_value_length)
+		return FALSE
+	var/sanitized = sanitize_name(value, default_species, allow_numbers)
```
ShadowLarkens: I don't think this'll work because is_valid blocks writes to the savefile

NovaDusklight: I didn't know that was checked first. You're right. the pref_deserialize and pref_serialize should handle it, but I'll change this portion as it's just wrong.

code/modules/client/preferences/names.dm
```diff
@@ -0,0 +1,89 @@
...
+		return FALSE
+	return TRUE
+
+/datum/preference/name/pref_deserialize(input, datum/preferences/preferences)
+	if(!istext(input))
+		return create_default_value()
+	var/species = default_species
+	if(preferences?.species)
+		species = preferences.species
+	var/sanitized = sanitize_name(input, species, allow_numbers)
```
ShadowLarkens, line 32: `allow_numbers` here is always FALSE.. you'll have to figure out some equivalent to `is_FBP`? this was the main point i got stuck on and stopped with my own port

NovaDusklight: Just tested and confirmed you're right. I don't know what to do about this yet, but I'll try and figure it out.

NovaDusklight: I got a fix working, and also found another bug I fixed while at it.
Your idea of adding an fbp check was basically 90% of the work.
I also added a secondary check in copy_to_mob for now, as I found that you could add numbers to your name as a FBP, then change your slot to be fully organic, and join the round with those numbers still there. Loading the slot would fix this issue, but they could just repeat the process.

ShadowLarkens (approved):
great job~!

Kashargul:
was this thoroughly tested? Those changes need to work or we get a highway to hell

@Cameron-The-Raven double check please, seems to work fine for me

NovaDusklight:
As thoroughly as I could test it. ID cards. Slot names, VR, ghost join, and
most systems that call for  a name.

I'd appreciate if another person could test too though.

It's worked perfectly for me for all systems i tested, but I think another
tester confirming would be good to be on the safe side

PR #19123, opened 2026-02-01
Allows toggling of reagent scans in deep scan devices

SatinIsle:
## About The Pull Request

Adds a verb to deep scan devices that allows them to turn off the sediment scan info, as this pastes a lot of text.

I think usr is okay to use in verbs, right?

## Changelog
:cl:
qol: Adds a verb to deep scan devices that allows them to turn off the sediment scan info, as this pastes a lot of text.
/:cl:

### DIFF
--- a/code/modules/mining/drilling/scanner.dm
+++ b/code/modules/mining/drilling/scanner.dm
@@ -28,12 +28,8 @@
 	set category = "Object"
 	set src in view(1)
 
-	if(sediment_scan)
-		to_chat(usr, span_notice("\The [src] will no longer scan for reagents."))
-		sediment_scan = FALSE
-		return
-	to_chat(usr, span_notice("\The [src] will now scan for reagents."))
-	sediment_scan = TRUE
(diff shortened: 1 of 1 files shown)


code/modules/mining/drilling/scanner.dm
```diff
@@ -22,6 +23,18 @@
...
 
+/obj/item/mining_scanner/verb/toggle_sediment_scan()
+	set name = "Toggle Sediment Scan"
+	set category = "Object"
+	set src in view(1)
+
+	if(sediment_scan)
+		to_chat(usr, span_notice("\The [src] will no longer scan for reagents."))
+		sediment_scan = FALSE
+	else
```
Kashargul: just return instead of if / else

KillianKirilenko: alternately something like
```c
to_chat(usr, span_notice("\The [src] will [sediment_scan ? "no longer" : "now"] scan for reagents."))
sediment_scan = !sediment_scan
```
ought to work nicely, if I remember how to use an in-line if

SatinIsle: That looks good to me, I forgot that was a thing we could do

PR #19126, opened 2026-02-02
Reduce Clumsy Gene's Obnoxious RNG

Willburd:
## About The Pull Request
Reduces the clumsy rng fail chance from 50% to 15% on all checks, and moves the percent chance to a define for easy editing if downstream servers disagree with this change. Clumsy is kinda unplayable in a roleplay focused server, having it be a quirky thing that sometimes happens has a better vibe than "i stab myself in the eye 50% of the time i click something with a screwdriver." Also adds a define to separate failures and harmful failures, so they can have different chances.

## Changelog
Moved clumsy rng chance to a define instead of hardcoded numbers on each check
Added a define for potentially lethal clumsy interactions, set to 10%.
Changed clumsy rng to 15% instead of 50% by default.
Corrected clumsy trait to actually apply the clumsy mutation, was non functional.

:cl: Will
fix: clumsy trait/gene not being applied to characters
balance: Clumsy gene failure chance reduced from 50% to 15%
balance: Clumsy gene failures that are potentially lethal moved from 50% to 10%
code: Clumsy failure chance moved to a define
code: Added separate define for harmful clumsy chance
/:cl:

### DIFF
--- a/code/__defines/dna.dm
+++ b/code/__defines/dna.dm
@@ -145,5 +145,5 @@
 // Misc
 #define TK_DENIED_MESSAGE span_warning("You are too distracted to focus your telekinesis.")
 #define TK_OUTRANGED_MESSAGE span_notice("Your telekinesis won't reach that far.")
-#define CLUMSY_FAIL_CHANCE(U) ((CLUMSY in U.mutations) && prob(10))
-#define CLUMSY_HARM_CHANCE(U) ((CLUMSY in U.mutations) && prob(5))
+#define CLUMSY_FAIL_CHANCE(U) ((CLUMSY in U.mutations) && prob(15))
+#define CLUMSY_HARM_CHANCE(U) ((CLUMSY in U.mutations) && prob(10))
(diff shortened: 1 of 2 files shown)


code/__defines/dna.dm
```diff
@@ -145,3 +145,5 @@
 // Misc
 #define TK_DENIED_MESSAGE span_warning("You are too distracted to focus your telekinesis.")
 #define TK_OUTRANGED_MESSAGE span_notice("Your telekinesis won't reach that far.")
+#define CLUMSY_FAIL_CHANCE 10
```
Kashargul: if((CLUMSY in user.mutations) && prob(CLUMSY_HARM_CHANCE))

that is a common pattern, can't we move this fully to a define instead?

if(CLUMSY_HARD_CHANCE(user))
if(CLUMSY_FAIL_CHANCE(user))

#define CLUMSY_HARM_CHANCE(user) ((CLUMSY in user.mutations) && prob(5))
#define CLUMSY_FAIL_CHANCE(user) ((CLUMSY in user.mutations) && prob(10))

Willburd: Sure, i'll rework the pr or maybe just move it to a helper proc for possibly more complex logic in the future?

Kashargul: Hmm, I think a general define might work there for now. If we really need a helper later on, finding usages with a define is fast

Willburd: Moved logic to defines

Willburd:
Upped the chances a little. I still want it to have an impact on play, just not at 50%.
non lethal 10% -> 15%
lethal 5% -> 10%

PR #19128, opened 2026-02-02
Allows CPR and defibs to work on vox

Cameron-The-Raven:
## About The Pull Request
Vox are a rarely used species and currently if they die in game, there's no way to actually bring them back (They're unable to be sleeved and have the no_defib flag).

This allows them to be revived via CPR and defib, but still prevents them from being sleeved.

## Changelog

:cl: Diana
qol: Allows vox to be revived via defibs & CPR
/:cl:


SatinIsle:
Good change, I wonder why it was like that in the first place. I'd guess some ancient legacy balancing or something?

I don't think vox are plants though, just bird people things? I'd guess the no sleeving thing is because of whitelists.

PR #19138, opened 2026-02-04
Fixes microwave reagents, oopsies

MeepleMuncher:
## Changelog
:cl:
fix: Fixed microwaves not being able to accept reagents
/:cl:

### DIFF
--- a/code/modules/food/kitchen/microwave.dm
+++ b/code/modules/food/kitchen/microwave.dm
@@ -52,24 +52,35 @@
 ********************/
 
 /obj/machinery/microwave/RefreshParts()
-	efficiency = 0
-	advanced_microwave = FALSE
-	name = "microwave"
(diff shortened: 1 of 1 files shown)


code/modules/food/kitchen/microwave.dm
```diff
@@ -137,6 +148,7 @@
 	if(handle_dirty(O, user)) return TRUE
 	if(handle_deconstruction(O, user)) return TRUE
 	if(try_insert_item(O, user)) return TRUE
+	if(try_insert_reagent(O, user)) return FALSE
```
Kashargul, line 151: why FALSE?

MeepleMuncher: it has to return false in order for afterattack to proc! which is what actually allows for fluids to be poured from glasses and beakers and such...

PR #19141, opened 2026-02-05
Fix chem subsystem reverse lookup

Willburd:
fixchem

## About The Pull Request
Reverse lookup in chem subsystem was only reading the first reagent in the recipes requirement list (and not the catalysts either). Meaning reagent documentation that should state they are a component of a recipe were not logging what they could produce properly. For example bicardine was not listing it was part of vermicetol as bicardine was not the first reagent in vermicetol's recipe.

## Changelog
Fixes SSchemistry init to properly log the production reagents for reverse lookup lists, and not just the first reagent.

:cl: Will
fix: SSchemistry reverse lookup data is correctly init. Meaning Bingle and the chem analyzer now properly list all reagents a chemical can actually produce, and not just the first recipe.
fix: Daiquiri is now possible to drinkmix and is no longer blocked by fire punch (add limejuice first)
/:cl:

### DIFF
--- a/code/_helpers/global_lists.dm
+++ b/code/_helpers/global_lists.dm
@@ -62,7 +62,7 @@ GLOBAL_LIST_EMPTY(ear_styles_list)	// Stores /datum/sprite_accessory/ears indexe
 GLOBAL_LIST_EMPTY(tail_styles_list)	// Stores /datum/sprite_accessory/tail indexed by type
 GLOBAL_LIST_EMPTY(wing_styles_list)	// Stores /datum/sprite_accessory/wing indexed by type
 
-GLOBAL_LIST_INIT(custom_species_bases, new) // Species that can be used for a Custom Species icon base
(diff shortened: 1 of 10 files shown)


code/controllers/subsystems/chemistry.dm
```diff
@@ -39,24 +39,32 @@ SUBSYSTEM_DEF(chemistry)
...
-		if(D.required_reagents && D.required_reagents.len)
-			var/reagent_id = D.required_reagents[1]
+
+		var/list/scan_list = list()
+		if(D.required_reagents?.len)
+			scan_list += D.required_reagents
+		if(D.catalysts?.len)
+			scan_list += D.catalysts
+
+		for(var/i in 1 to scan_list.len)
```
Kashargul, line 49: length(list)

Willburd: modified

PR #19143, opened 2026-02-05
Nerfs the vac pack

zeskorion:
## About The Pull Request
the vac pack attachment now obeys container capacity restrictions
the vac pack attachment now obeys trash eater restrictions
the vac pack attachment has a ~~double~~ 150% attack cooldown, meaning that it is slower to vacuum multiple tiles in succession. This makes it more of a sidegrade to other janitorial equipment, specialized in versatility and compactness, rather than a one-size fits all solution to any mess in existence

## Changelog
:cl: Wide Eyes
balance: vac pack now cleans half as fast
fix: vac pack no longer bypasses trash eating restrictions
fix: vac pack no longer bypasses container size restrictions
/:cl:

### DIFF
--- a/code/game/objects/items/devices/vacpack.dm
+++ b/code/game/objects/items/devices/vacpack.dm
@@ -120,7 +120,10 @@
 			return
 	if(istype(target,/obj/structure/window) || istype(target,/obj/structure/grille))
 		target = get_turf(target) // Windows can be clicked to clean their turf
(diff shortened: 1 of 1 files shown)


Aroliacue:
I would recommend removing the click-delay change. Otherwise, good PR.

zeskorion:
> I would recommend removing the click-delay change. Otherwise, good PR.

I feel it's somewhat necessary. The vacpack is a do-it-all with janitorial stuff

PR #19144, opened 2026-02-05
Makes translocators able to properly use any device cell

ReoDaProtovali:
## About The Pull Request

For some reason, translocators are set to check for the typepath of the cell they initialize with, including subtypes. This meant higher tier translocators could only use specific cell types, since they initialize with better cells (normal starts with an adv cell, alien starts with void cell, ect) while the basic translocator could use any device cell as it uses the base `/obj/item/cell/device` cell, which includes all subtypes of device cells. 

This Pr makes it use a static istype() check instead of being dynamic to the translocator's cell type, so that the cell it requires is not tied to the cell it spawns with.
## Changelog
:cl:
fix: all tiers of translocators can use any power device power cell
/:cl:

### DIFF
--- a/code/game/objects/items/devices/translocator_vr.dm
+++ b/code/game/objects/items/devices/translocator_vr.dm
@@ -174,7 +174,7 @@ This device records all warnings given and teleport events for admin review in c
 		rebuild_radial_images()
 
 /obj/item/perfect_tele/attackby(obj/W, mob/user)
-	if(istype(W,cell_type) && !power_source)
+	if(istype(W, /obj/item/cell/device) && !power_source)
 		power_source = W
 		power_source.update_icon() //Why doesn't a cell do this already? :|
 		user.unEquip(power_source)


code/game/objects/items/devices/translocator_vr.dm
```diff
@@ -174,7 +174,7 @@ This device records all warnings given and teleport events for admin review in c
 		rebuild_radial_images()
 
 /obj/item/perfect_tele/attackby(obj/W, mob/user)
-	if(istype(W,cell_type) && !power_source)
+	if(istype(W, /obj/item/cell/device) && !power_source)
```
Kashargul, line 177: /obj/item/cell/device/weapon

why should it fit the large cells?

ReoDaProtovali: why shouldn't it? the mini-translocator fits all cells, and starts with the small cells. There's no reason the better translocator shouldn't fit larger/smaller cells. Besides, cells are usually restricted based on class, not power capacity

Unless you're misunderstanding the type, `/obj/item/cell/device/weapon` is the higher capacity device power cell, tier 2. Normal, big cells are just `/obj/item/cell`

Kashargul: Because it puts the usages up quite massively. Before that, one was heavily charge limited. So that's more going to need a balance discussion.

ReoDaProtovali: It's not particularly charge limited if you can swap cells. It just makes it inconsistent with everything else that uses cells if translocators are only ever supposed to use the same type of cell they start with.

At that point, they might as well all be made to have integrated, irremovable cells and you just charge the device itself

ReoDaProtovali: Regardless, get back to me on what the intent is, because this function doesnt work as intended regardless.

Kashargul: Might be best to move them to integrated cells only

ReoDaProtovali: I feel like most of the community will disagree with that, especially since some of them specifically are, but if that's what the staffmind think is best I cant really dispute it

ReoDaProtovali: Also I dont want a GBP hit for balance so I might let someone else do it

PR #19168, opened 2026-02-10
moves all subsystems to length

Kashargul:
## Changelog
:cl:
code: considering length is faster than .len, this moves subsystems where possible to length
/:cl:

### DIFF
--- a/code/controllers/subsystems/job.dm
+++ b/code/controllers/subsystems/job.dm
@@ -14,16 +14,16 @@ SUBSYSTEM_DEF(job)
 
 
 /datum/controller/subsystem/job/Initialize()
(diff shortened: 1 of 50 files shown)


code/controllers/subsystems/job.dm
```diff
@@ -80,13 +80,9 @@ SUBSYSTEM_DEF(job)
 	return dept_datums
 
 /datum/controller/subsystem/job/proc/get_job(rank)
-	if(!occupations.len)
```
ShadowLarkens, line 83: are you sure these won't break anything?

ShadowLarkens: #19191

PR #19169, opened 2026-02-10
ghost block perf

Kashargul:
Highly experimental to increase ghost block performance...

@Cameron-The-Raven maybe you can test this with Will, you two said it caused issues on z changes,

🆑 
fix: high performance cost of large ghost vision blocked areas
/🆑


Kashargul:
Ye no... that makes things much worse even

PR #19176, opened 2026-02-11
"Innland" Redgate maptweaks

KillianKirilenko:
## About The Pull Request

Makes some changes to the "innland" redgate map, as the map felt a little lacking outside of just... having a few teppi around in the corners of the map.

The main addition is a set of four bedrooms tucked in the northwest corner of the map, wrapped in unbreakable walls and linked to the main building via a portal stairwell. The lighting will probably look pretty jank around the transition but there's not much I can do about that. Bedrooms are fully-featured 4x4 spaces with a personal locker, bin, table, chairs, double-bed (with curtains), light switch, lockable door, and cabinet with candles+matches.

Other changes include a pathway that leads from the gate straight into the building itself, with a door on the north side (opposite the original door) for easier access. The Booze-O-Mat and Kitchen Cabinet had their access locks removed so anyone can take over bartending/cooking there. A light was added to the freezer room since that was otherwise unlit, and a couple of small bulbs have been added to the hall outside the 'management' office to make it easier to navigate that space (and find the new stairwell). Finally a wall cabinet by the kitchen door has plenty of candles and matches for adding some more soft lighting, as the main room is very very dark.

## Changelog

:cl:
qol: innland redgate; kitchen cabinets and dispensers now have no access restrictions
qol: innland redgate; added a couple of extra lights and some candles/matches
maptweak: innland redgate; added path leading directly from gate, added door to north side of building
maptweak: innland redgate; added four bedrooms linked via portal 'stairway'; moved tcomm unit into janitor closet
/:cl:

### DIFF
--- a/maps/redgate/innland.dmm
+++ b/maps/redgate/innland.dmm
@@ -198,7 +198,6 @@
 /area/redgate/structure/powered/bed1)
 "aI" = (
 /obj/machinery/button/remote/airlock{
-	desc = "A remote control switch for the medbay foyer.";
 	dir = 4;
 	id = "bedroom1";
 	name = "Door Control";
@@ -224,7 +223,6 @@
 /area/redgate/wilds)
 "aO" = (
 /obj/machinery/button/remote/airlock{
-	desc = "A remote control switch for the medbay foyer.";
 	dir = 8;
 	id = "bedroom2";
 	name = "Door Control";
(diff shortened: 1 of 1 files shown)


KillianKirilenko:
Bedrooms in NW corner;
image

Main building;
image

PR #19178, opened 2026-02-12
Mechfab Fixes

KillianKirilenko:
## About The Pull Request
Tweaks the mechfab to avoid a situation where it could be deadlocked and require the output tile to be deconstructed in order to continue.

Gently tested, seems to work as intended and doesn't spam the chat with the warning message.

## Changelog
:cl:
qol: mech part fabricator now has a clearer message when you try to change its output direction whilst it's printing
fix: mech part fabricator now completely refuses to start printing a part if its output direction is invalid; ongoing queues should be frozen until the output is reoriented
/:cl:

### DIFF
--- a/code/modules/research/tg/machinery/mech_fabricator.dm
+++ b/code/modules/research/tg/machinery/mech_fabricator.dm
@@ -240,7 +240,7 @@
 		return FALSE
 
 	var/turf/exit = get_step(src, drop_direction)
-	if (exit.density)
+	if(exit.density)
 		if(verbose)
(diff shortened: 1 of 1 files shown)


code/modules/research/tg/machinery/mech_fabricator.dm
```diff
@@ -239,6 +239,12 @@
 	if(!D || length(D.reagents_list))
 		return FALSE
 
+	var/turf/exit = get_step(src, drop_direction)
+	if (exit.density)
```
ShadowLarkens:
```suggestion
	if(exit.density)
```

PR #19202, opened 2026-02-18
Move autolathe design datums to techweb datums

Willburd:
## About The Pull Request
Moves old autolathe datums to techweb datums, along with multiple other changes. Basically ports almost all of the TG autolathe. Including automatic directional printing and hacked techweb designs. In the future techweb disks could be created as loot with unique designs not part of the techweb that can be inserted into an autolathe(or protolathe) to print. The rework unfortunately comes at the cost of stripping out ore silo support, without major departure from what TG has. I'm sure i'll find out quick if that's demanded to return or not.

Fixes many instances of missing ammunition in the lathe. If clutter is the concern, these could be moved to techweb disks and given to security instead of being default lathe entries. However this would be a much larger change and should be done in it's own PR.

## Changelog
Ported TG autolathe (mostly)
all OG autolathe recipes datums have been migrated to techweb designs
adds a wall of missing recipes that our codebase can make but just didn't have
autolathe states with all INITIAL tagged autolathe designs and all hacked autolathe designs
support for using techweb disks to bring non-standard autolathe designs to a autolathe 
support for HACKED techweb designs that only appear as printables if the machine is hacked by wirecutting
protolathe and circuit imprinter wire hacking added (you could previously open the panel but nothing appeared)
There are no hacked circuit imprinter designs, but the option for the future is there.
adds many unprintable standard ammunitions that were either never brought to the lathe or disabled

:cl: Will, Kash
refactor: replaced autolathe datums with techweb datums
code: many new validity checks for techweb unit test added, multiple changes made to fix them.
code: added unit test that checks for circuits missing from the imprinter
balance: autolathe uses it's own isolated material storage
balance: many circuits tossed into the engivend replaced with an autolathe and imprinter circuit instead, as they are all printable now
fix: techweb entries now respect machine hacked state regarding hack only designs
add: protolathe and circuit imprinter wire hacking
add: a literal mountain of lathe and circuit designs that were missing
qol: circuit boards that cannot be printed on station will show a unique message on examine.
qol: techweb and lathe menus now use scaled icons for larger images instead of huge buttons
fix: injector maker no longer uses an AFRG board during construction
/:cl:

### DIFF
--- a/code/modules/research/tg/designs/boards/telecomms.dm
+++ b/code/modules/research/tg/designs/boards/telecomms.dm
@@ -76,6 +76,7 @@
 
 // Telecomm parts
 /datum/design_techweb/subspace
+	desc = "Complex components used to construct and repair telecomms machinery."
 	build_type = PROTOLATHE
 	category = list(
 		RND_CATEGORY_MACHINE + RND_SUBCATEGORY_MACHINE_TELECOMMS
--- a/code/modules/unit_tests/techwebs.dm
+++ b/code/modules/unit_tests/techwebs.dm
@@ -117,10 +117,12 @@
 		if(design.id == DESIGN_ID_IGNORE)
 			continue
 
+		/* Later
 		// Must have desc
 		if(!design.desc)
 			TEST_NOTICE(src, "TECHWEB DESIGN - [design.type] did not have a description.")
 			failed = TRUE
+		*/
 
 		// Must all be accessible by science
 		if(!(design.departmental_flags & DEPARTMENT_BITFLAG_SCIENCE))


Willburd:
I can't think of anything else I need to do in this PR. Players will inevitably run into whatever explodes. It's working for me so far, and mostly reflects prior behavior with a lot of techweb fixes.



Woe be downstream deconflicting.

Willburd:
added some requested material weapons and replaced shotgun shells with their ammo boxes (unlikely the weapons that use mags, shotguns prefer ammoboxes for loading)

code/modules/research/tg/designs/modsuit_designs.dm
```diff
@@ -24,48 +28,70 @@
...
+	category = list(
+		RND_CATEGORY_MODSUIT_MODULES
+	)
 
 /datum/design_techweb/mechfab/modsuit/zero_controller
 	name = "Zero Suit Control Module"
 	desc = "A very bare bones and stylish suit designed for pilots with advanced holographic interfaces for the task."
 	id = "zero_rig_module"
 	materials = list(MAT_PLASTEEL = 12000, MAT_GOLD = 3000, MAT_GRAPHITE = 3000, MAT_OSMIUM = 1000, MAT_PLASTIC = 5000)
 	build_path =/obj/item/rig/zero
```
Cameron-The-Raven:
```suggestion
	build_path = /obj/item/rig/zero
```

code/modules/research/tg/designs/weapon_designs.dm
```diff
@@ -953,14 +911,141 @@
...
+	build_path = /obj/item/material/sword/battleaxe
+	category = list(
+		RND_CATEGORY_HACKED,
+		RND_CATEGORY_INITIAL,
+		RND_CATEGORY_WEAPONS + RND_SUBCATEGORY_WEAPONS_MELEE
+	)
+	departmental_flags = DEPARTMENT_BITFLAG_SCIENCE | DEPARTMENT_BITFLAG_SECURITY
+*/
+
+/datum/design_techweb/gladius
```
Cameron-The-Raven, line 1011: all the material/sword weapons have a force of 42, may want to keep an eye on these...Not a fault of the autolathe, but is more a problem of material code being stinky.

code/modules/research/tg/designs/weapon_designs.dm
```diff
@@ -953,14 +911,141 @@
...
+	materials = list(MAT_STEEL = 7400)
+	build_path = /obj/item/material/sword/sabre
+	category = list(
+		RND_CATEGORY_HACKED,
+		RND_CATEGORY_INITIAL,
+		RND_CATEGORY_WEAPONS + RND_SUBCATEGORY_WEAPONS_MELEE
+	)
+	departmental_flags = DEPARTMENT_BITFLAG_SCIENCE | DEPARTMENT_BITFLAG_SECURITY
+
+/datum/design_techweb/longsword
```
Cameron-The-Raven, line 1067: meanwhile, with another great example of material code, the longsword (two handed weapon) only does 18

code/modules/research/tg/designs/illegal_designs.dm
Cameron-The-Raven, line 1: Should we add RND_CATEGORY_HACKED to all of these in this file?

code/modules/research/tg/designs/weapon_designs.dm
Willburd: Do we want to disable these and make them downstream or removed?

code/modules/research/tg/designs/illegal_designs.dm
Willburd: donezo... do we want a display when researching if a design is hack required or not? Might be good to ask kash on.

Willburd:
Is there anything else we should add to the techweb while we are here?

code/modules/research/tg/designs/illegal_designs.dm
Kashargul: Hmmm... might be interesting to have a small hint on the UI, nothing too massive though... something like a small icon

code/modules/research/tg/designs/weapon_designs.dm
Kashargul: Shouldn't be printed easily likely.... we really need to get to a gun and weapon recode

Willburd: Disabled, only the ones originally present from the original designs are enabled now.

code/modules/research/tg/designs/illegal_designs.dm
Willburd: This is low priority to the full function of the lathe. It can always be added later. Up to you.

Willburd:
ready once again

Kashargul (approved):

PR #19218, opened 2026-02-22
Disable afk prey pref for dead people

TheGreatKitsune:
## About The Pull Request
Dying in game will cause the AFK prey pref to prevent noms. To allow medborgs and other such interactions to still work in this case, and since bypassing the pref maliciously by using this would be a serious issue for a moderator to look at anyways, the AFK prey pref will be disabled if dead.
## Changelog
:cl: Cerami
fix: Disables the AFK prey preference from functioning while someone is dead
/:cl:

### DIFF
--- a/code/_helpers/vore.dm
+++ b/code/_helpers/vore.dm
@@ -20,7 +20,7 @@
 		return FALSE
 	if(!pred.can_be_afk_pred && (!pred.client || pred.away_from_keyboard))
 		return FALSE
-	if(!prey.can_be_afk_prey && (!prey.client || prey.away_from_keyboard))
+	if(!prey.is_dead() && !prey.can_be_afk_prey && (!prey.client || prey.away_from_keyboard))
 		return FALSE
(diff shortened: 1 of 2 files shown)


Kashargul:
@SatinIsle

SatinIsle:
Hmmm, I think this is probably okay? Does AFK trigger on dead bodies that have been sat around for a while or something?

TheGreatKitsune:
AFK triggers if the ghost isn't in the body. (Sets client to empty which the AFK code checks for to see if DC'd)

SatinIsle:
Then yeah, I'd say AFK pref isn't really intended to prevent these interactions.

TheGreatKitsune:
Yeah. Not intuitive to the user that leaving their body as a ghost triggers the AFK prey code right now, you can't change your prefs to allow it once dead, and also if someone abuses this to eat someone that is AFK that's probably grounds for either a serious warning or a straight ban so.

PR #19242, opened 2026-02-28
Tether Stardriver

KillianKirilenko:
## About The Pull Request

With the Exploration department having been shuttered some time ago, and with few expeditions launched recently, NT is looking to cut some unnecessary costs. As a result, the giant, maintenance-heavy Excursion Shuttle has been replaced with the smaller and less intensive NTV Stardriver, the same model as the SD's NTV Starstuff shuttle. The hangar has been ~~refitted into a 'drop hangar' with the shuttle connected by a small docking arm;~~ restored to its original configuration (mostly).

image

Despite the downsizing its capacity is still roughly equal to that of the much larger Excursion Shuttle. The smaller footprint also means it'll have an easier time landing at various POIs and away-mission sites (whale, etc.) and it also just makes sense from a standardization standpoint as well.

Unrelatedly, maintenance behind gateway prep has also been redesigned and reshaped to allow for the space there to be used as a new hideout/building niche;

image

## Changelog
:cl:
maptweak: replaced the tether excursion shuttle with a smaller carboncopy of the starstuff
maptweak: reworked maintenance behind the gateway prep area to make a new project niche with a view of space
/:cl:


KillianKirilenko:
It just seemed like a cool bit of flavour. I'm restoring the hangar anyway, but mostly because I *cannot* be bothered to figure out why it's spitting active edge errors at me.

KillianKirilenko:
Wretched machine. Back to the drawing board with this one, again.

PR #19257, opened 2026-03-06
Gender identity component

Willburd:
## About The Pull Request
Moves gender identity var out of humans and into a component, allowing pronouns to be used on any mob as a player desires.

Byond does not natively support genders other than neuter, male, female, or plural. Changing to genders like herm (something only humans can do currently due to a custom gender handler.) cause it to explode. The goal of this component is moving that human code into a reusable component that can be passed to other mobs during transformations.

Technically this also makes it much easier to add genders, but that is out of the scope of this PR.

## Changelog
Adds gender_identity component.
Adds many lines of support for new gender component.

:cl: Will
refactor: Gender identity moved to a component
refactor: cleans up multiple places where pronouns were solved manually instead of using the helper procs
add: borgs and pais now use the gender identity of their slot when spawned.
/:cl:


Willburd:
aauugghhh so many places where the pronoun helpers are not used, manually checking male/female.

PR #19258, opened 2026-03-06
Default spawnpoint

TheGreatKitsune:
## About The Pull Request
Actually sets the default spawn type to one that all maps have and that works for both carbons and silicons. Also adds a unit test to make sure that spawn point is defined for new maps. (I would have done Arrivals Shuttle but minitest doesn't have it so Gateway it is)
## Changelog
:cl: Cerami
fix: Specify the default spawn type.
/:cl:


Kashargul:
Having looked through the code yesterday evening, might be able to actually try the shuttle as we don't map test mini test.

At the same time, it might fall back to gateway if that's unavailable. Best to test out on minitest.

PR #19267, opened 2026-03-09
Flammable Items, Abstractness, and hypnosis

Cameron-The-Raven:
## About The Pull Request
Implements a new burning subsystem, burning component, and flammability from TG.

Ports over ABSTRACTness from TG (These are items that don't really exist and we don't treat them as they exist for certain interactions. Currently only cared about in fire/destructive analyzer)

Flammable items will now catch on fire, smoke, slowly be degraded, then eventually perish under the heat. The 'atom_destruction' proc can be overridden per-object to allow for them to have special interactions when they are destroyed.

The 'take_damage' proc now will deal damage to the object it's called on properly, eventually resulting in that item having 'atom_destruction' called, resulting in it spilling any contents it may have out (barring ABSTRACT contents) and being destroyed entirely.

Picking up burning objects will BURN you! 
Wearing fireproof gloves, a firesuit, or anything else that protects > BURNING_ITEM_MINIMUM_TEMPERATURE and has hand coverage will allow you to pick up burning items (and extinguish them!)

Examining objects will inform you of their various resistances, such as fireproof items informing you of such when you examine them.

Adds in object flags (NYI, but now here).

Also adds hypnotext because the examining system required it.
dreamseeker_nK7qrz9QfZ
## Changelog
:cl: Diana
add: Objects can now take damage. The most prevalent source of this is FIRE! Objects can now catch on fire and melt away!
qol: Objects will make sounds when they take damage now.
qol: Examining objects will inform you of their various properties. Examining a fireproof, bombproof, or flammable item will inform you of such.
code: Items can now be labeled as ABSTRACT. This means the game will ignore them in certain situations.
code: Instead of things having their own specialty /deconstruct proc, they all now share a /atom_deconstruct proc
/:cl:


NovaDusklight:
I now feel legally required to do something with the font.

Willburd:
Does this remove /obj/effect/abstract in favor of the trait or just applies it to the type? This type should probably be removed so that all locations that respect it can be changed to check trait.

Cameron-The-Raven:
adjusting abstract effects to something will have to be done in the future, due to the fact that ABSTRACT is an item_flag and not an obj_flag. Outside of scope of this PR (just added abstract as it's part of the flags that porting flammability brought over) however

PR #19270, opened 2026-03-09
Radiation Refactor

Cameron-The-Raven:
## About The Pull Request
Ports over TG's radiation backend with some edits.

### MAJOR NOTES:

Radiation is now MASSIVELY more optimized and does not result in the garbage SS being flooded with useless 'radiation datums' and instead controls radiation via a subsystem.

Radiation no longer spawns a 'radiation effect' for every turf that is affected by radiation.

Radiation now works in 'pulses'. This means a radiation source will always be emitting radiation FROM the radiation source instead of leaving a trail of garbage radiation sources behind it.

Radiation now will be hampered by insulating materials. Lead, concrete, and rad_collectors are the greatest insulators. R_walls are the second greatest insulator. Things like phoron-glass is the third greatest insulator (along with walls). Glass and grills are very weak insulators.

Radiation armor actually MATTERS now. The greater the armor, the more it'll protect against radiation. This does not lower the chance of you getting rads altogether, but massively lowers the amount of rads you get if you are affected by rads.

Radiationn -generally- has a timer when you step into an area affected by radiation. For most things this timer is 3 seconds. For some infrequent things (pacman generators for example) it does not have a timer, just a base chance.

## Radiation Mechanics

Radiation now works by radioactive entities sending out radioactive PULSES. These can be triggered via signals (uranium floors, walls, and doors perform this) or via sending out signals during its process() loop (Supermatter does this).

Getting hit by a singular pulse from a radiation emitting object -generally- doesn't immediately irradiate you. You have 3 seconds (or until the next pulse) to vacate the area and you'll be unharmed. Linger any longer and every pulse will start rolling the dice to see if you get irradiated.

You are not warned if you're in a radioactive area, so geiger-counters now have a use. Geiger counters are talked about more below.

#### Entities that can be hit are:

any mob/living
any obj/item/geiger
any rad_collector (These are special as they skip the 'line calculation' to save on CPU). This means that if a rad collector is within range of a rad source, it gets full radiation as if it was right next to the radiation source without taking range or insulation into account. 

### During this line, it takes in account:
1. Distance (The further you are from the radioactive source, the lower the radiation you receive will be and the lower your chance of getting irradiated will be)
2. Objects between the entity and the source. (The more radiation insulating objects between them, the lower the chance of getting irradiated and the lower the radiation you receive)

Number 2 has a caveat that it -generally- looks for the most radiation insulating tile. This means if you put down a lead wall and multiple phoron glass windows on that tile, it'll be more absorbing than if you put multiple lead walls in a row across multiple tiles or multiple windows in a row on multiple tiles. This additionally means rwall+blast door OR lead wall + blast door is the 'best' way to insulate things mapwise without doing some janky game mechanics.

## Radiation Resistance 

Radiation resistance now actually matters;
Radiation resistance is now calculated per-limb and takes all your highest rad insulation into account.
What this means is that if you are wearing 4 pieces of clothing covering your chest with each piece giving 10, 50, 80, 90 radiation resistance, it'll give you a 90% chance of your chest being excluded from the radiation calculation.
This is done for every limb.

This means if you're wearing full body protection that gives 90% rad resistance, each limb has a 90% of not being affected, reducing the amount of radiation you get from that pulse by how many limbs are protected. 

If you have 10 limbs and 9 roll to be protected and you are hit with a rad pulse of 50 rads, you'll only obtain 5 rads. If you have 10 limbs and 0 are protected, you'll obtain 50 rads.

## Geiger Counters
dreamseeker_gydZ45eULB

Geiger counters now inform you via colors of how a radioactive source is affecting that tile.

Green = No radioactive source nearby, or no radiation pulses have occurred in some time.

Yellow = The tile is in range of a radioactive source, but there is ample insulation between the tile and the turf to prevent becoming irradiated.

Orange = The tile is in range of a radioactive source. There is insulation between the tile and source, but just barely.

Red = The tile is in range of a radioactive source. There is not enough insulation between the tile and source and anything on that tile is in risk of being irradiated. (Note: Even if there is a 0.1% chance of being irradiated, it will still show red.)

Geiger counters can additionally scan living entities to see how irradiated they are.

## Showers
Showers now remove 5 rads per tick that you are under them. This is only 20% as effective as being in the cryopod.
This does not clean chronic radiation, so if you're making repeated trips into Chernobyl, you'll still need to visit medical to have those removed.

## Reagents
Reagents can now add traits that can be checked via the HAS_TRAIT system.

Prussian Blue changed from a reagent check to adding a 'RADIATION_EFFECT_IMMUNE' trait.

Screwdriver - The drink the bartender can make - will now have the same effect as prussian blue. (This was locked to engineer livers only on TG, but given we have less of a 'gamey' aspect, the effect works on everyone equally)

This means that the bartender has some actual use outside of being a completely social job.

## MISC

NOTE: Because of how the OLD and the NEW radiation work, power generation by rad_collectors will be affected and have been tweaked. 
The singularity should be unaffected (should give some more power generation if anything) since it mostly worked on just giving rad collectors direct power.
The SM was examined at 500 and 1000 power. 1000 power gave ~500kW for one collector with a pressurized phoron canister so the ratios were carried over as much as possible.

## Changelog
:cl: Diana
add: Screwdriver (The Drink) stops radiation effects while it is in your system.
add: Uranium walls, doors, and coins now emit slight amounts of radiation.
qol: Radiation protective items now have a tag upon examine that informs of their radiation protective status.
refactor: Massively improves radiation performance by moving to pulses instead of 1000s of deleted radiation_sources a second. 
code: Radiation reworked. Please report on Discord or Github if power generation feels too low from radioactive (SM) sources.
/:cl:


AzzyDreemurr13:
I wonder if this will lead to an increase in radiation on Talon to the engineering compartment....

Cameron-The-Raven:
> I wonder if this will lead to an increase in radiation on Talon to the engineering compartment....

Shouldn't as long as the engine is properly surrounded by radiation blocking walls.

AzzyDreemurr13:
But they block a certain amount, I just hope it doesn't raise the radiation too much when the ship's nuclear engine is running.

Cameron-The-Raven:
> But they block a certain amount, I just hope it doesn't raise the radiation too much when the ship's nuclear engine is running.

Nah, it doesn't block amounts anymore but thresholds. I.e if you have a moderate radiation source, anything That blocks moderate or higher will block it. If you have an extreme rad source, anything that blocks extreme will block it, etc

AzzyDreemurr13:
Mmm, I see, again the main thing is that the classic walls that were used previously could block this moderate and high level, so that later you don’t end up in a situation: how to get to the engineering department and at least turn on the engine, not to mention access to protection suit

Cameron-The-Raven:
> Mmm, I see, again the main thing is that the classic walls that were used previously could block this moderate and high level, so that later you don’t end up in a situation: how to get to the engineering department and at least turn on the engine, not to mention access to protection suit

If the engine isn't running, it's not going to be making radiation and you'll be fine to enter engineering without an issue.

Most rad sources have a sort of time limit to 'how long can I be around it before I start to get irradiated'. For most things this is 3 seconds. (Some rad sources have no time limit, however) and even then you're not guaranteed to get irradiated if you're in the zone. 

This chance goes down further the more space from you and the radiation source, along with the more 'blocking' items between you and the rad source. The radiation you obtain also goes down the further you are from the rad source and the more things between you and the source.

This means that for most things, unless you are directly hugging it with no protection you'll be safe. Radiation is still quite the threat if you let it accumulate, but a quick dip into Chernobyl won't leave any long-lasting effects.

Additionally, prussian blue still prevents radiation effects (but not its accumulation), meaning you can use meds to stop it as well in the  event of you having no rad suits and having to enter an irradiated zone.

AzzyDreemurr13:
> Additionally, prussian blue still prevents radiation effects (but not its accumulation), meaning you can use meds to stop it as well in the event of you having no rad suits and having to enter an irradiated zone.

Just look Talon map And look where the reactor is located, in the current reality, when a nuclear engine is constantly on, the walls contain the radiation and it stays inside this small room. But since you're upgrading the radiation system, I'm afraid that this little room will be insufficient and will cover sections that have never been touched.

Cameron-The-Raven:
> > Additionally, prussian blue still prevents radiation effects (but not its accumulation), meaning you can use meds to stop it as well in the event of you having no rad suits and having to enter an irradiated zone.
> 
> Just look Talon map And look where the reactor is located, in the current reality, when a nuclear engine is constantly on, the walls contain the radiation and it stays inside this small room. But since you're upgrading the radiation system, I'm afraid that this little room will be insufficient and will cover sections that have never been touched.

The talon might need to be adjusted, but for reference, geiger counters that are red = dangerous (in risk of getting irradiated) while geiger counters that are orange = safe (radiation source present but there is enough insulation between the source and that position):

For reference, all of these are with the SM at ~1000 power. Higher power = more penetration.

dreamseeker_Du2WhKNVfv

With blast doors shut: 
dreamseeker_vkZmd4ZL1E

With blast doors shut + extra insulation (lead walls): 
dreamseeker_UFNGpyG4T2

Double reinforced design: 
dreamseeker_F9mIbTy0nk

Single reinforced design, rad collectors surrounding it (A side room was made as well to simulate room separation): 
dreamseeker_pOSgwxVTKL

For the last photo, that one 'red' marker in the room on the left only has a 0.8% chance of being irradiated every time the SM pulses with a strength of 4 rads, which is almost harmless.

The red marker in the side room doorway (directly left of the door) has a 2.28% chance of being irradiated every pulse for a strength of 10.5 rads

AzzyDreemurr13:
> The talon might need to be adjusted, but for reference, geiger counters that are red = dangerous (in risk of getting irradiated) while geiger counters that are orange = safe (

Teleport to Talon and start the nuclear reactor at 200 power.
It uses an improved type of uranium generator (As a game mechanic) but the sprite of a nuclear reactor. 
Without your changes, the reactor is usually yellow inside, but it's extremely rare for the scanner to show orange for a few seconds. 

But outside it's always green.

code/controllers/subsystems/radiation.dm
```diff
@@ -1,148 +1,186 @@
...
-		// Okay, now ray trace to find resistence!
-		var/turf/origin = source.source_turf
-		var/working = source.rad_power
-		while(origin != T)
-			origin = get_step_towards(origin, T) //Raytracing
-			if(!resistance_cache[origin]) //Only get the resistance if we don't already know it.
-				origin.calc_rad_resistance()
-
-			if(origin.cached_rad_resistance)
-				if(CONFIG_GET(flag/radiation_resistance_calc_mode) == RAD_RESIST_CALC_DIV)
```
Kashargul, line 96: do we no longer use those configs? if so, fully axe

code/controllers/subsystems/radiation.dm
```diff
@@ -1,148 +1,186 @@
...
-					working = round((working / (origin.cached_rad_resistance * CONFIG_GET(number/radiation_resistance_multiplier))), 0.01)
-				else if(CONFIG_GET(flag/radiation_resistance_calc_mode) == RAD_RESIST_CALC_SUB)
-					working = round((working - (origin.cached_rad_resistance * CONFIG_GET(number/radiation_resistance_multiplier))), 0.01)
-
-			if(working <= CONFIG_GET(number/radiation_lower_limit)) // Too far from this source
```
Kashargul, line 101: do we no longer use those configs? if so, fully axe

code/controllers/subsystems/radiation.dm
```diff
@@ -1,148 +1,186 @@
...
+			var/rad_resistance = clothing.armor["rad"]
+			if(prob(rad_resistance))
+				protected_limbs++
 				break
 
-		// Accumulate radiation from all sources in range, not just the biggest.
-		// Shouldn't really ever have practical uses, but standing in a room literally made from uranium is more dangerous than standing next to a single uranium vase
-		. += working / (dist ** 2)
-
-	if(. <= CONFIG_GET(number/radiation_lower_limit))
```
Kashargul, line 109: do we no longer use those configs? if so, fully axe

Cameron-The-Raven:
> > The talon might need to be adjusted, but for reference, geiger counters that are red = dangerous (in risk of getting irradiated) while geiger counters that are orange = safe (
> 
> Teleport to Talon and start the nuclear reactor at 200 power. It uses an improved type of uranium generator (As a game mechanic) but the sprite of a nuclear reactor. Without your changes, the reactor is usually yellow inside, but it's extremely rare for the scanner to show orange for a few seconds.
> 
> But outside it's always green.

Ah, yeah, just checked. Just a resprited PACMAN generator.

dreamseeker_YCnsfzD2Ih

Its radiation is self contained to its room (it has a 2 range as well for the pulses)

The generator has a 2% chance when it uses fuel to cause a radiation pulse. At 50 sheets, it does a pulse of 500 strength (this is multiplied by the chance, which is 10%, which brings the 'rads you would get hit with' down to 50. 

It has a 10% chance of irradiating you for 50 rads if you're next to it when one of these pulses happen. That would cause very very minor damage but nothing too major.

code/controllers/subsystems/radiation.dm
Cameron-The-Raven: deleted

AzzyDreemurr13:
Mmmm, maybe it's worth adding a new warning note, since newbies often get confused with the ship and don't launch it with full fuel of 50 units. So that later there would be no shock as to where they got the radiation through the walls.

Cameron-The-Raven:
> Mmmm, maybe it's worth adding a new warning note, since newbies often get confused with the ship and don't launch it with full fuel of 50 units. So that later there would be no shock as to where they got the radiation through the walls.

There shouldn't be any confusion there, since you have to be right up against it to get irradiated. It doesn't go through the walls due to the radiation source being too weak.

AzzyDreemurr13:
By the way, does the new radiation affect synthetics in any way?

Cameron-The-Raven:
> By the way, does the new radiation affect synthetics in any way?

Radiation effects are unchanged (barring an admin toggle for special rad glow effect), this is mostly a backend change to the radiation subsystem which affects how radiation is spread out.

AzzyDreemurr13:
> > Mmmm, maybe it's worth adding a new warning note, since newbies often get confused with the ship and don't launch it with full fuel of 50 units. So that later there would be no shock as to where they got the radiation through the walls.
> 
> There shouldn't be any confusion there, since you have to be right up against it to get irradiated. It doesn't go through the walls due to the radiation source being too weak.

Ah, phew, then everything's even better. I was most afraid of the radiation coming through the wall, especially when I saw that everything was red in your update.

Randulf0:
This seems to have broken radiation wave events.
Both the exterior and interior events produce no rads at all.

Kashargul:
> This seems to have broken radiation wave events. Both the exterior and interior events produce no rads at all.

PR is closed, make an issue and link to the PR

PR #19281, opened 2026-03-13
Freezing Airlocks

Willburd:
## About The Pull Request
Upported feature from yawnwider. external airlocks will freeze over and require breaking the ice off of them to use. To avoid headaches upstream, airlocks will only freeze under very specific circumstances:

A) The airlock must be /external or /glass_external subtype (can_freeze() proc can be overridden to allow or forbid this)
B) The airlock must be on a planet (or shuttle)
C) The airlock must be exposed to an outside planet-side turf that can experience weather
D) The temperature of that outdoor turf must be lower than -15C

If all of these conditions are met the door has a chance to freeze. Airlocks automatically melt if the temperature is above 0C, or if the conditions before are unmet. Originally the freezing airlocks were a subtype, but being able to just deconstruct the airlock and reconstruct it has obvious issues. It is limited to planet surface exterior airlocks with super cold weather to avoid becoming an issue upstream.

## Changelog
Ports yawnwider freezing external airlocks.

:cl: Will
add: External airlocks on planets will freeze in extremely cold weather.
/:cl:

### DIFF
--- a/code/game/machinery/doors/airlock.dm
+++ b/code/game/machinery/doors/airlock.dm
@@ -1233,7 +1233,7 @@ About the new airlock wires panel:
 
 /// Most airlocks don't freeze, subtypes set this
 /obj/machinery/door/airlock/proc/can_freeze()
-	SHOULD_BE_PURE(TRUE)
+	SHOULD_BE_PURE(TRUE) // Don't put logic here, just return if the airlock can freeze or not.
 	PROTECTED_PROC(TRUE)
 	return FALSE
 


Aroliacue:
Awesome feature, love this. Yawnwider has some cool stuff

Willburd:
> Awesome feature, love this. Yawnwider has some cool stuff

Working on an upport of their map to archive on chomp, I was told to bring most of the features upstream as I do. So most of their stuff is going to get up here with some polishing to pass modern code review.

PR #19282, opened 2026-03-13
Slime Research Unittest

Willburd:
## About The Pull Request
I missed a slime, so making a unit test so it's impossible to miss a slime ever again.

## Changelog
Adds a unit test to make sure all xenobio slimes have research point values assigned.

:cl: Will
code: Slime research point unittest
fix: light pink slime cores lacked a destructive analyzer point value
/:cl:


code/modules/unit_tests/slime_tests.dm
```diff
@@ -0,0 +1,12 @@
+/datum/unit_test/all_slimes_must_have_research_values
+
+/datum/unit_test/all_slimes_must_have_research_values/Run()
+	var/failed = FALSE
+
+	for(var/slime in subtypesof(/obj/item/slime_extract))
+		if(!(slime in SSresearch.techweb_point_items))
```
Kashargul, line 7: Why no TEST_ASSERT?

Willburd: Because having a test that iterates through entire lists fail and drop out of the test for each one (with multiple potential failures) is bad in my opinion. I've done all my list based tests in this format. I do not want to do a single CI run for each missing path, and would rather it give all missing entries and fail so they can all be fixed at once. TEST_ASSERT drops the entire test on first fail and only reports one issue.

Issue #19187, opened 2026-02-14
[UI] Copied Loadout Slots Are Linked


KillianKirilenko:
### Brief description of the bug

If you use the Copy Slot feature, then the original and copied-to loadout slots will be "linked" until you save, swap character, and load back to the original character. Any changes, especially to colour customization or selectors (such as for uniforms), will be carried over between the slots.

### Reproduction steps

```bash
1. Make a new character
2. Set up Loadout Slot 1
3. Copy it to Loadout Slot 2 or 3 using the [Copy Slot] button
4. Make changes to Slot 2 and/or 3
5. Swap back to Slot 1 to see your changes have applied to it too
```

### Server Revision Info

Round ID: 1499
BYOND Version: 516.1677
Server revision compiled on: 2026-02-13T03:16:58+00:00
Master commit: [5d1d996c009e1a67c49080bd827b9cb26246e74e](https://github.com/VOREStation/VOREStation/commit/5d1d996c009e1a67c49080bd827b9cb26246e74e)
The following pull requests are currently test merged:
[#17936: 'Toggle on Tether' by ItsSelis at commit c8bfb52ece](https://github.com/VOREStation/VOREStation/pull/17936)

Local commit: f60175c3594a0e595c3d29db8720b913d3a4fb7f
TGS version: 6.19.2
DMAPI version: 7.4.0

PR #19296, opened 2026-03-16
Fix Loadout items not properly initializing their own metadata on copy

ARGUS-Memory:
## About The Pull Request

copy_loadout used check_list_copy on the outer gear list only, leaving the inner tweak metadata lists as shared references. Modifications to gear tweaks in the copied slot would write into the shared inner lists, affecting the original slot until a save/reload cycle broke the reference.

Fix: deep copy both levels, matching the pattern used for body_markings in 03_body.dm.
## Changelog
:cl:
fix: Copy loadout now correctly makes unique metadata entries for gear. So that changing the color on one pair of glasses doesnt affect the ones in your other loadout.
/:cl:


ARGUS-Memory:
Fixes #19187

PR #19313, opened 2026-03-19
reimplements attack and player logs

Kashargul:
> [!CAUTION]
> Database user requires DROP permissions to use TRUNCATE, make sure to only enable it for the erro_dialog table

## Changelog
:cl:
admin: reimplements attack and player logs
code: rids of the old db connection hook... the dbcore subsystem needs to handle it
fix: player logs per mob verbs
/:cl:


Kashargul:
Might have to see how high the trade off is when using TG's logging... but I don't quite want Drop rights back on the db

PR #19314, opened 2026-03-19
modularize synx.dm into synx/ subfolder

ARGUS-Memory:
Split old Synx.dm into modules for easier debugging and maintenance, also split of child variants into their own files.

Turn the main Synx.dm into a top level style includer for all other synx code
## Changelog
:cl:
code: Reorganized Synx. DM as a master file with module children
/:cl:


ARGUS-Memory:
Mapping tools are not ready for this

PR #19325, opened 2026-03-22
escape shuttle to subsystem

Kashargul:
## Changelog
:cl:
refactor: moves escape shuttle to a subsystem
/:cl:


Cameron-The-Raven:
the starspawner was a server crash poison

PR #19326, opened 2026-03-22
Allows treating infected organs

Cameron-The-Raven:
## About The Pull Request
Allows treating infected organs if they have a germ level. This surgery already sets the germ level to 0 when completed, but you can't do it on organs that you want to heal the germs of.

It's much less efficient than using the bioregenerator (it uses resources) but it's an option, especially if doing ghetto surgery
## Changelog
:cl: Diana 
qol: Treat Organ surgery's detection logic now includes infected organs
/:cl:


Cameron-The-Raven:
github wtf

PR #19330, opened 2026-03-22
remove hacked requirement for chem sprayer design

Willburd:
## About The Pull Request
I think this was a copypasta issue. Close this if it is intended.

## Changelog
Removes the lathe hacking requirement on the chem sprayer.

:cl: Will
balance: chem sprayer does not require protolathe hacking to print
/:cl:


Cameron-The-Raven:
hmmm, technically it has traditionally been an antag tool (pacid sprayer) but...Given we don't have antags it has the potential benefit of now (with touch reagent code) being a medical device, or being a device the janitor can ask for. Seems fine  to be a normal item.

PR #19364, opened 2026-03-31
Add a TG pref to not be radio announced

NovaDusklight:
## About The Pull Request

Adds a new preference directly below your spawn point selection to disable radio announcements for your arrival and departure.
EDIT: This only silences arrival if you are off duty, or visiting. Will not suppress working roles joining the round.
This will silence departures for all roles for vorny privacy reasons, but can be changed easily.

image

## Changelog

:cl: Nova
qol: Allow silencing arrival radio messages if off-duty
qol: Allow silencing departure radio messages.
/:cl:


TheCaramelion:
Honestly I'm not really a fan of station jobs coming in silently, removes that interaction bit when someone comes in to work

NovaDusklight:
> Honestly I'm not really a fan of station jobs coming in silently, removes that interaction bit when someone comes in to work

Edit: Rewriting this after further thought. 
On-duty roles are now only silenced when they depart, so your issue with it should be resolved. I agree with why you had issue with it.
Though I'd appreciate if you gave feedback more constructively, instead of rejecting it due to one aspect that can be changed

Kashargul:
> > Honestly I'm not really a fan of station jobs coming in silently, removes that interaction bit when someone comes in to work
> 
> Edit: Rewriting this after further thought. On-duty roles are now only silenced when they depart, so your issue with it should be resolved. I agree with why you had issue with it. Though I'd appreciate if you gave feedback more constructively, instead of rejecting it due to one aspect that can be changed

Departement shouldn't be silenced either... one should know who's there when being crew

NovaDusklight:
> Departement shouldn't be silenced either... one should know who's there when being crew

That makes sense. So would you say that this would be fine, if it only allowed silencing for visiting roles?
I'm more than happy to make that change, I just want confirmation on what should/should not be silenced.

NovaDusklight:
@Kashargul I changed the PR with the feedback mind. Now it works as such:

Visitor: The only role that can have a silent arrival.
Other roles: Arrival is always announced. Departure is silenced, similar to silent portals on station. (We already can leave a round silently as any role, so I think this is fine?)

code/game/machinery/cryopod.dm
```diff
@@ -539,6 +539,9 @@
 		if(istype(to_despawn, /mob/living/dominated_brain))
 			depart_announce = FALSE
 
+		if(to_despawn.client?.prefs?.read_preference(/datum/preference/toggle/living/radio_announce) == FALSE)
+			depart_announce = FALSE
```
Cameron-The-Raven, line 543: Problem with this: When the user ghosts and despawn_occupant is called sometime later, client = null, meaning this fails to fire and the announcement is made.

NovaDusklight:
I didn't think this would catch that. I'll look into it as soon as I can

lbnesquik:
Visitors and interns should be fine, IMO. As long as you are still in the manifest.

PR #19368, opened 2026-04-01
expands misc gripper functionality

zeskorion:
## About The Pull Request
Grippers can now hold papers, pens, and leashes
leashes now work when used on, or by, non-carbons.

## Changelog
:cl:
add: Leashes can now work when used by, or on, non-carbons
qol: misc gripper can hold a few more utility items. Pens, paper, leash
/:cl:


Kashargul:
Has that been tested with the different mob AI holders to not cause issues? Especially with active path finding?

zeskorion:
> Has that been tested with the different mob AI holders to not cause issues? Especially with active path finding?

tested several, no issues. Adding two failsafe checks to make sure leashed mobs always have minds, though

code/game/objects/items/leash.dm
```diff
@@ -32,16 +32,23 @@
 	w_class = ITEMSIZE_SMALL
 	var/mob/living/leash_pet = null //Variable to store our pet later
 	var/mob/living/leash_master = null //And our master too
+	var/doubleclick = FALSE //This is awful, but necessary for borgs to use leashes. Grippers call both attack() and attackby() for their functionality, attaching and detaching the leash, so this var essentially disables using the leash in very quick succession
```
Kashargul: why? Just return the right usage like most items do when usable by borgs, this looks horrible.

code/game/objects/items/leash.dm
```diff
@@ -62,24 +72,29 @@
...
+		if(humantarget.handcuffed)
+			leashtime = 5
 
 	C.visible_message(span_danger("\The [user] is attempting to put the leash on \the [C]!"), span_danger("\The [user] tries to put a leash on you"))
 	add_attack_logs(user,C,"Leashed (attempt)")
 	if(!do_after(user, leashtime, C)) //do_mob adds a progress bar, but then we also check to see if they have a collar
 		return
-	if(tgui_alert(C, "Would you like to be leased by [user]? You can OOC escape to escape", "Become Leashed",list("No","Yes")) != "Yes")
-		return
+	/*if(tgui_alert(C, "Would you like to be leased by [user]? You can OOC escape to escape", "Become Leashed",list("No","Yes")) != "Yes")
```
Kashargul: why is that removed?

Kashargul (dismissed):

code/game/objects/items/leash.dm
zeskorion: Change made for testing. I'll fix when i wake

code/game/objects/items/leash.dm
zeskorion: As stated jn the comment: due to how grippers are coded, this will be used twice in quick succession; they first call item.attack(target), and then target.attackby(item). I have no idea *why*, and i dont want to even get *into* removing those. The end effect is that borgs first attach the leash, and then, due to this functionality, immediately attempt to unattach it. This effectively prevents that from occuring. I'm scared that *changing* that jank will screw grippers up on some core level that i cannot begin to describe

Kashargul: Check how other items handle their return on gripper usage to cancel the attack chain

zeskorion: > Check how other items handle their return on gripper usage to cancel the attack chain

Sure. It might just not be a factor with afterattack(), which most things use.  I'll enter the code mines tomorrow

zeskorion: > Check how other items handle their return on gripper usage to cancel the attack chain

we've established other items do not actually handle this issue. NEeds further looking into, i'll make a github issue. For now, i've lengthened the unleash time to make working around it feasible

code/game/objects/items/leash.dm
zeskorion: fixed

PR #19384, opened 2026-04-07
Always Randomize Name Hotfix

Willburd:
## About The Pull Request
Fixes a stackover flow caused by "Always randomize name" preference. When the new name is saved to the file it'll eventually try to update the display mob. This then calls copy_to(), which reads the always randomized name pref, and makes a new random name that is saved to the file. Repeating the prior logic of updating the display mob. This happens until we finally overflow and crash out. 

This pr does not fix the core logic issue, it just stops the server exploding. Someone smarter than me at pref code needs to fix this properly. Sorry.

## Changelog
Adds a hotfix that prevents real_name preference save updating the preview mob (thus avoiding an inf recursion)

:cl: Will
fix: always randomize name preference temporarily hot-fixed
/:cl:


Willburd:
Closing as someone else has offered to fix this properly, to avoid accidental merge.

PR #19391, opened 2026-04-08
Gun Init incorrectly assumes firemode properties

Willburd:
## About The Pull Request
This mostly applies to some downstream guns, and a single gun here. Guns with firemodes different from single shot, or that start with a different firemode than the first mode will have incorrect firing properties, while saying their firemode does other things.

## Changelog
Applies the current firemode datum to the gun during init if a valid firemode can be selected.

:cl: Will
fix: Guns with nonstandard firemode selection properly set firemode properties during init
/:cl:


code/modules/projectiles/gun.dm
```diff
@@ -144,6 +144,10 @@
 		verbs -= /obj/item/gun/verb/give_dna
 		verbs -= /obj/item/gun/verb/allow_dna
 
+	if(length(firemodes) && sel_mode <= firemodes.len)
```
Kashargul: why? length(....) is null save

just do sel_mode <= length(firemodes)

Willburd: oh huh. i never considered that.

PR #19427, opened 2026-04-20
Camera circuit addition

NovaDusklight:
## About The Pull Request

Adds a video camera circuit to circuitry.

This works by having one input circuit which can be synced to multiple video exporters. 
These video output circuits have a passive power draw, and MUST be in an assembly to work.
Uses a slightly altered version of the UI from the Camera Console.

Circuits all have fairly high complexity, but not too high.
Logic circuits complexity = 1
Camera circuit complexity = 5, 6, or 8. Depending on the circuit.

Functions similarly to the spy cameras, but requiring circuitry assemblies for the camera and viewer.

## Changelog

:cl: Nova
add: Added working video circuitry
/:cl:


Kashargul (commented):
Hmm ok, that's partially a copy of the console... would need a full rework of those UIs

PR #19428, opened 2026-04-21
Fixes disposal bins not unlinking properly when being deconstructed and causing a hard del

ReoDaProtovali:
## About The Pull Request

I guess I just forgot to make them unlink in the past, lol. I also added an additional unlink attempt on destroy, just to be safe.
Speedmerge pls?
## Changelog
:cl:
fix: Fixes a hard del caused by disposal bins not trying to unlink on destruction, resulting in uncleared refs.
/:cl:


code/modules/recycling/disposal_machines.dm
```diff
@@ -606,6 +607,7 @@
 	for(var/atom/movable/AM in src)
 		AM.forceMove(T)
 	//..() //*cough
+	SEND_SIGNAL(src, COMSIG_DISPOSAL_UNLINK) //unlinks in destroy, too. Multi-sending shouldnt be too bad.
```
Kashargul, line 617: this fires twice here and in Destroy?

ReoDaProtovali: sending it again is not an issue, it will just do nothing, I wanted to be safe, but I could remove the second one if it's a concern

ReoDaProtovali: this is also the case for outlets. Outlets had an unlink signal but I added a second one in destroy just to be doubly sure that they'll unlink. In theory, in the future, i'll try to port the TG thing where disposal pipes are cached inside of nullspace when deconstructed instead of deleted and made anew every time, which would then necessitate separate delinkage. But as of now, it'll just provide a second check that does nothing 99% of the time, until something qdels it without calling deconstruct (I imagine admin delete does this?)

ReoDaProtovali:
it has been a month, this is still causing harddels in live.

PR #19435, opened 2026-04-25
Cargo Material Purchasing

Willburd:
pointsink

## About The Pull Request
I was requested to upport this. I was concerned about mining gameplay, but mining can also mine whatever they want with this and sell it to buy what the station is asking for without being screwed over by mining ore rng. Adds supply packs for all major materials at 50 sheet stacks for an almost price-gouging price to avoid invalidating mining.

## Changelog
Tweaks many material prices
Adds material prices to mats missing them
Adds unit test to enforce material prices
Fixed a bug with non-material stacks runtiming the supply shuttle when sold.

:cl: Will
add: Cargo can now order most materials at a markup and only in full stacks
balance: Many material stacks have had their sell prices changed
fix: Attempting to sell non-material stacks on the supply shuttle no longer runtime
/:cl:

### DIFF
--- a/code/ATMOSPHERICS/components/binary_devices/algae_generator_vr.dm
+++ b/code/ATMOSPHERICS/components/binary_devices/algae_generator_vr.dm
@@ -292,6 +292,7 @@
 	hardness = 10
 	sheet_singular_name = "sheet"
 	sheet_plural_name = "sheets"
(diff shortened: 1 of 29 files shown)


Willburd:
I'd need to know what changes are desired over that downvote. I was requested to do this by Kash.

PR #19444, opened 2026-05-02
Some bugfixes

Cameron-The-Raven:
## About The Pull Request
Fixes a bug with accessories accidentally searching for nearby users.
Gives emitters and field generators EMPPROOF. Not used ATM, but if either one of them has things added, they should be EMP proof.
Fixes partslathe so it doesn't have every part.
Fixes a bug where the field gen would break and never be able to be turned off/restarted if a window or other dense object bumped into it. Instead, it just qdel's the object.
Fixes attempting to drag-drop non-item entities into disposals.
Fixes a borg module runtime
Fixes horsemask, now makes you properly neigh
Makes thieves gloves not show a interaction when using it
Fixes thieves gloves from having a 1 second do_after for EVERY click interaction
## Changelog
:cl: Diana
fix: Fixes a bug with accessories searching for nearby users.
fix: Fixes a bug where partslathe had every part.
fix: Fixes a bug where windows would delete the shield gen containment field
fix: Attempting to drop non-item entities (unintended) into disposal will no longer runtime.
fix: Fixes a borg runtime
fix: Horsemask now properly makes you neigh. Also adds an easter egg...This mask isn't ever used, so...
fix: Fixes some doafters
/:cl:

### DIFF
--- a/code/game/machinery/partslathe_vr.dm
+++ b/code/game/machinery/partslathe_vr.dm
@@ -344,10 +344,10 @@
 /obj/machinery/partslathe/proc/update_recipe_list()
 	if(!partslathe_recipies)
 		partslathe_recipies = list()
-		var/list/paths = subtypesof(/obj/item/stock_parts)
+		var/list/paths = subtypesof(/obj/item/stock_parts) - subtypesof(/obj/item/stock_parts/subspace)
 		for(var/type in paths)
 			var/obj/item/stock_parts/I = new type()
-			if(!I.matter)
+			if(!I.matter || I.rating > 1)
 				qdel(I)
 				continue // Ignore parts we can't build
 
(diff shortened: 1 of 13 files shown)


code/game/machinery/partslathe_vr.dm
```diff
@@ -344,10 +344,10 @@
 /obj/machinery/partslathe/proc/update_recipe_list()
 	if(!partslathe_recipies)
 		partslathe_recipies = list()
-		var/list/paths = subtypesof(/obj/item/stock_parts)
+		var/list/paths = subtypesof(/obj/item/stock_parts) - subtypesof(/obj/item/stock_parts/subspace)
```
Kashargul: are you sure? I would say types of for the latter

Cameron-The-Raven: /subspace isn't a path that actually exists, it's just a non-existent path

Kashargul: there are no non existing paths, it still exists, byond creates all parents. So why do that work twice by having it excluded and then skipped

Cameron-The-Raven: adjusted

PR #19455, opened 2026-05-10
Changes RMS material cost to require silver

tacoguy7765093:
## About The Pull Request
For a tool that's powerful and able to potentially generate every rare material given enough time once you get a stable power source, having a minor cheap material cost feels fair. Once a engine is setup, its also likely a basic mining haul should have come in.
## Changelog
:cl:
balance: RMS material cost changed.
/:cl:


CMDRYouth:
I highly recommend changing this to require a more rare material, such as Verdantium.  There are already discussions about recycling coins and finding maintenance spawn materials to circumvent this change entirely.

Verdantium at least fully requires mining and coins of it do not spawn.  To my knowledge, the odds of sheets of it spawning elsewhere are also low.

PR #19468, opened 2026-05-16
Cameranet to SScameras draft

Kashargul:
## About The Pull Request
WIP Draft to move cameranet to SScameras
## Changelog
:cl:
code: cameranet to SScameras
/:cl:


Cameron-The-Raven:
should be good to TM, there's a weird edge bug where the very edges of the map have a thin layer of visibility (the transition tiles) but it's very minor.

Cameron-The-Raven:
current bugs:
Being between camera blockers at the map border will result in the edge camera blocker not showing up until you enter it.
dreamseeker_5RFdr1Fguy

Because of how our space has an 'edge' var that works with the skybox, the static shows up offset instead of in its right spot. This can be seen working on walls, but not on space turf.
dreamseeker_962en4TuCX
dreamseeker_0hCcuUQTeA

Vs TG:
dreamseeker_1rWIvF4qeo

Kashargul:
We need to dig some to see how to best resolve those. I really would like to get away from the old system.

PR #19478, opened 2026-05-20
Atmos Pipe Improvements

Willburd:
pipeA
pipeB
pipeC
pumpfix
allconnect

## About The Pull Request
There are multiple issues with atmo pipes that this resolved, improves, or adds functionality for. Includes the univeral pipe adaptor also having multiple connections on the same side at once. Allowing for many pipes to split off from the same pipe with a single adaptor.

## Changelog
-standard, high pressure, and volume pumps can be built on any pipe layer
-multiple pumps can be built on a turf if they are not already on a turf on that pipe layer
-vents and scrubber support aux and fuel layers
-pressure tanks can be constructed and deconstructed
-canister connectors for scrubber and supply pipes added
-fixed major issues involving atmo machines that accept multiple pipe layers where the connection would be improperly done if multiple compatible pipes were present
-aux, fuel, scrubber, and supply pipes no longer reset their color (and ignore RPD color) when deconstructed
-added missing volume pump starts on prefab
-universal pipe adaptors can have multiple connections on the same side, and will flow into each other

:cl: Will, Reo
add: multiple atmo pumps can be constructed on the same turf
add: atmo pumps can be constructed on all pipe layers
add: canister connectors can be constructed for all pipe layers
add: vents and scrubbers support aux and fuel pipe layers
add: pressure tanks can be constructed and destructed, and now support all pipe layers
add: rpd and pipe dispenser can produce pressure tanks
add: all connections into a universal pipe adaptor function, and are no longer limited to one per side when building the network
fix: bad recipes removed from the rpd and pipe dispenser (use the correct layer select for them instead)
fix: atmo machines that support multiple pipe layer connections now properly link to pipes nearby if more than one compatible type of pipe is present in the linking direction
fix: aux, fuel, scrubber, and supply pipes no longer reset their color (and ignore RPD color) when deconstructed
fix: volume pumps no longer have broken map icons
fix: added missing volume pump on prefab
/:cl:


Willburd:
tested with an SM setup and looking over SD for any weirdness. Everything seems fine.

Willburd:
Needs more validation on the universal pipe adaptor connecting and disconnecting properly in all scenarios.

PR #19509, opened 2026-06-06
Trasheat backend

Cameron-The-Raven:
## About The Pull Request
Adds a recursive check to trasheating for when eating storage items.

Condenses all the /storage/etc into just /storage for the whitelist. (This is where the recursive check comes into play).

Adds TTS devices to the vore whitelist.

Adds Bluespace Harpoon and CE belt to blacklist.

Makes adminbus trasheat not ignore the blacklist.

Allows you to eat PDAs with it's owner in your stomach.

Changes the name of adminbus trasheat var.

===NOT ENABLED===:
Adds a MIDDLE GROUND to trasheat. This checks:

Is it on the whitelist? Proceed like normal (Allow it)
Is it on the blacklist? Proceed like normal (Block it)

Is it NOT whitelisted OR blacklisted? Do the below:
Is the item's force > 30? (Block it)
Is the item >= ITEMSIZE_LARGE? (Block it)

This criteria could be expanded, but this should allow for more items to be eaten via trasheat without a continually expanding whitelist.

Eventually, the whitelist should be cut down, but this is a more 'experimental' PR.

======
## Changelog
:cl: Diana
fix: You can now eat PDAs with the owner in your stomach.
fix: Adminbus no longer bypasses blacklist.
fix: Circuit box, CE belt, bluespace harpoon, and locus are now blacklisted for object vore.
code: All /storage items are able to be eaten now. Whitelist still applies.
code: Trasheat now recursively searches through items that things are holding.
/:cl:


code/game/objects/trash_eating.dm
```diff
@@ -8,19 +8,60 @@
...
+
+	//If it's not whitelisted, check criteria.
+	if(force > 30 || throwforce > 30) //Swords, etc.
+		to_chat(user, span_warning("The [src] is too powerful to eat."))
+		return FALSE
+
+	if(w_class >= ITEMSIZE_LARGE)
+		to_chat(user, span_warning("The [src] is too large to eat."))
+		return FALSE
+
```
Verkister, line 50: No return value for "none of above"?

Verkister: Also looking at the other checks calling this proc, they seem to be expecting opposite results, blocking trues and allowing falses. (recursive search and the trash eater check)

Cameron-The-Raven: Ye I need to invert the logic, I wrote this before I went to bed w/o testing which is why I left it as a draft.

PR #19519, opened 2026-06-11
Rascal's Pass Map Swap

Lilly-Lira:
Test merge PR for setting the map to Rascal's Pass.  Don't actually merge this.


vorestation.dme
```diff
@@ -5050,7 +5050,6 @@
 #include "maps\expedition_vr\beach\submaps\mountains_areas.dm"
 #include "maps\expedition_vr\virgo5\v5_features.dm"
 #include "maps\groundbase\groundbase_areas.dm"
-#include "maps\groundbase\groundbase_poi_maps.dm"
```
Kashargul, line 5053: Pois are always to be included in the main include. The map datum of groundbase itself needs adjustments.

Lilly-Lira: Fixed!

PR #19526, opened 2026-06-17
Looc sound

Very-Soft:
## About The Pull Request

Adds a sound effect to LOOC messages. Only plays if you can hear the LOOC message normally, does not play for RLOOC. Also adds a preference toggle for the sound effect.

## Changelog
:cl:
add: Sound effect to LOOC messages
add: Preference for LOOC sound

/:cl:


code/modules/client/verbs/ooc.dm
```diff
@@ -176,6 +176,8 @@
 			admin_stuff += "/([key])"
 
 		to_chat(target, span_looc(create_text_tag("looc", "LOOC:", target) + " <EM>[display_name][admin_stuff]:</EM> " + span_message("[msg]")))
+		if(target.prefs?.read_preference(/datum/preference/toggle/looc_sounds))
+			target << sound('sound/talksounds/looc_sound.ogg', volume = 50)	//Only sound for actually local targets, not for RLOOC because that would be annoying
```
Kashargul: SEND_SOUND(target, sound(...))

Kashargul (changes requested):

code/modules/client/verbs/ooc.dm
Very-Soft: That doesn't seem to allow for setting the volume.

Kashargul: it should

Kashargul (approved):

PR #19528, opened 2026-06-18
Aghost Logging

Very-Soft:
Adjusts the placement of the re-enter body admin log for aghosting so that it reports the key correctly.

Per directive after consulting Joan:
Adds an admin log to the opposite end of aghosting (when you aghost) as it seemed to be missing.
Makes both a regular admin log instead of log_and_message_admins, as there doesn't seem to be much point in messaging the admins every time someone aghosts.
## Changelog
:cl:
fix: Aghost admin logs record the associated key.
change: Aghost admin logs no longer message admins
add: Aghost admin log on aghost

/:cl:

### DIFF
--- a/code/modules/admin/admin_verbs.dm
+++ b/code/modules/admin/admin_verbs.dm
@@ -57,7 +57,7 @@ ADMIN_VERB(admin_ghost, R_HOLDER, "Aghost", "Ghost out of your body with the opt
 		to_chat(user, span_filter_system(span_warning("Error: Aghost: Can't admin-ghost whilst in the lobby. Join or Observe first.")))
 	else
 		//ghostize
-		log_admin("[key_name(mob)] has aghosted.")
+		log_admin("[key_name(user)] has aghosted.")
 		var/mob/body = mob
 		var/mob/observer/dead/ghost
 		if(build_mode)
--- a/code/modules/admin/admin_verbs.dm
+++ b/code/modules/admin/admin_verbs.dm
@@ -57,7 +57,7 @@ ADMIN_VERB(admin_ghost, R_HOLDER, "Aghost", "Ghost out of your body with the opt
 		to_chat(user, span_filter_system(span_warning("Error: Aghost: Can't admin-ghost whilst in the lobby. Join or Observe first.")))
 	else
 		//ghostize
-		log_admin("[key_name(src)] has aghosted.")
+		log_admin("[key_name(mob)] has aghosted.")
 		var/mob/body = mob
 		var/mob/observer/dead/ghost
 		if(build_mode)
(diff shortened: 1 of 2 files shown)


code/modules/admin/admin_verbs.dm
```diff
@@ -57,6 +57,7 @@ ADMIN_VERB(admin_ghost, R_HOLDER, "Aghost", "Ghost out of your body with the opt
 		to_chat(user, span_filter_system(span_warning("Error: Aghost: Can't admin-ghost whilst in the lobby. Join or Observe first.")))
 	else
 		//ghostize
+		log_admin("has aghosted.")
```
Kashargul: log_and_message with string, user as vars. Only log like this won't work

log_admin itself would need to have a keyname(thingy) inside the string otherwise it's lacking context

Kashargul (changes requested):

code/modules/admin/admin_verbs.dm
Very-Soft: Ah I can update it, I couldn't actually test the log_admin function correctly as admin logging is currently non-functional

Kashargul: Logging is fully functional. It's inside the admin log json file

Very-Soft: The tool for reading admin logs in the game is not functional

image

I reported this to selis earlier tonight.

Kashargul: You need to use the modern log viewer

code/modules/admin/admin_verbs.dm
```diff
@@ -57,6 +57,7 @@ ADMIN_VERB(admin_ghost, R_HOLDER, "Aghost", "Ghost out of your body with the opt
 		to_chat(user, span_filter_system(span_warning("Error: Aghost: Can't admin-ghost whilst in the lobby. Join or Observe first.")))
 	else
 		//ghostize
+		log_admin("[key_name(src)] has aghosted.")
```
Kashargul: src is invalid for datum events, user

This Pr nonetheless will need some discussion as usually aghosting should be announced.

log_and_message_admins("text", user) would be the right format for that. Message spam is prevented by just using the admin chat filters and making a specific admin tab.

code/modules/admin/admin_verbs.dm
```diff
@@ -57,6 +57,7 @@ ADMIN_VERB(admin_ghost, R_HOLDER, "Aghost", "Ghost out of your body with the opt
 		to_chat(user, span_filter_system(span_warning("Error: Aghost: Can't admin-ghost whilst in the lobby. Join or Observe first.")))
 	else
 		//ghostize
+		log_admin("[key_name(mob)] has aghosted.")
```
Kashargul: I would explicitly pass user as it's the client which is usually the first thing that get's checked in the key_name proc and always will return the right mob internally.

Kashargul (approved):

PR #19541, opened 2026-06-23
Borgs can now nibble kibble (again)

WolvesAndOwls:
## About The Pull Request
Borgs and now nibble kibble as they once did
## Changelog
:cl:
fix: Borg-o's are now borg consumable again
/:cl:


Willburd:
Nature is healing.

code/game/objects/items/trash_vr.dm
```diff
@@ -25,6 +25,20 @@
 	desc = "Contains every type of scrap material your robot puppy needs to grow big and strong."
 	icon = 'icons/mob/dogborg_vr.dmi'
 	icon_state = "kibble"
+	var/bitecount = 0
+	var/bites = 5
+
+/obj/item/trash/rkibble/attack_ai(mob/user as mob)
```
Cameron-The-Raven:
```suggestion
/obj/item/trash/rkibble/attack_ai(mob/user)
```

Also, does this mean that the AI can attack it at range?

WolvesAndOwls: attack_robot() just calls attack_ai(). There is a check that makes sure the user is a robot, but it WILL proc from a distance. I forgot to check that ;-;

WolvesAndOwls: image
All good now.

Cameron-The-Raven:
Making edits.

Making it give charge, use attack_robot (Since borgs call attack_robot, which calls attack_ai if there isn't a specific attack_robot interaction on the item, while AI never calls attack_robot, only attack_ai), along with swapping the distance to return early.

Likewise, converted it to a balloon alert which is less chat spam and looks nicer.

PR #19556, opened 2026-06-28
Refactors throw vore code. Again.

Cameron-The-Raven:
## About The Pull Request
Refactors throw code:

Eat Trash is no longer a verb-proc, instead is a proc that calls a helper proc after doing some checks.

Throw vore now does the proper recursive checks.

Also adds a fix for healbellies. 

## Changelog
:cl:
refactor: Trasheat code is a bit easier to follow on the backend, now.
fix: Throwing trash now properly respects the recursive search check.
code: Optimizes healbelly code so that handle_belly_update isn't called upwards of 26 times a cycle, only called one time a cycle now.
/:cl:

### DIFF
--- a/code/game/objects/trash_eating.dm
+++ b/code/game/objects/trash_eating.dm
@@ -95,15 +95,16 @@
 
 
 /mob/living/proc/eat_trash_proc(obj/item/item_to_eat, thrown = FALSE)
-	if(!item_to_eat.check_item_devourability(src))
+	if(!item_to_eat.check_item_devourability(src) && !thrown)
 		to_chat(src, span_warning("You can not eat this item."))
 		return FALSE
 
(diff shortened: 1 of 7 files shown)


code/game/objects/trash_eating.dm
```diff
@@ -9,19 +9,25 @@
...
-		to_chat(user, span_warning("You can't eat that so casually!"))
-		return FALSE
+		message_to_send = "You can't eat that so casually!"
 
 	if(hidden_uplink)
-		to_chat(user, span_warning("You really should not be eating this."))
+		message_to_send = "You really should not be eating this."
 		message_admins("[key_name(user)] has attempted to ingest an uplink item. ([user ? ADMIN_JMP(user) : "null"])")
+
+	if(message_to_send)
```
Kashargul: && please

PR #19559, opened 2026-06-30
GPS Rename

Cameron-The-Raven:
## About The Pull Request
Allows you to rename the GPS in your GPS watch.
## Changelog
:cl: Diana
qol: The survival watch GPS can now be renamed.
/:cl:

### DIFF
--- a/code/modules/clothing/accessories/watches.dm
+++ b/code/modules/clothing/accessories/watches.dm
@@ -35,7 +35,7 @@
 /obj/item/clothing/accessory/watch/survival
 	name = "survival watch"
 	desc = "An overengineered wristwatch that purports to be both space and water proof, and includes a compass, micro GPS beacon, and temperature and pressure sensors. The beacon is off by default, and can only transmit its location: it cannot scan for other signals."
-	description_fluff = "Hold ALT whilst left-clicking on the survival watch to toggle the status of its micro-beacon."
+	description_fluff = "Hold ALT whilst left-clicking on the survival watch to toggle the status of its micro-beacon. Ctrl-click in help intent to rename the micro-beacon."
 	icon_state = "wristwatch_survival"
 
 	var/obj/item/gps/gps = null
@@ -71,3 +71,15 @@
 	if(Adjacent(user))
 		gps.tracking = !gps.tracking
 		to_chat(user,span_notice("You turn the micro beacon [gps.tracking ? "on" : "off"]."))
+
+/obj/item/clothing/accessory/watch/survival/item_ctrl_click(mob/user)
+	. = ..()
+	if(!Adjacent(user) || (!user.a_intent == I_HELP))
+		return
+	var/new_name = tgui_input_text(user, "What would you like to name the GPS?", "Rename GPS", gps.gps_tag, 11)
+	new_name = sanitize(new_name, 11)
+	if(!new_name || !Adjacent(user))
+		return FALSE
+	gps.gps_tag = uppertext(new_name)
+	name = "global positioning system ([gps.gps_tag])"
+	to_chat(user,span_notice("You rename the GPS to [uppertext(new_name)]."))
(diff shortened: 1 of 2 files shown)


KillianKirilenko:
1) The watch GPS was never meant to be renamable. It is an **<ins>emergency</ins>** beacon and should be treated as such.
2) The basic 1pt webbing has lower storage capacity than all the items in the webbing selector, which is why the selector costs more, and why the basic webbing isn't just in the selector to begin with.

Cameron-The-Raven:
> 1) The watch GPS was never meant to be renamable. It is an **<ins>emergency</ins>** beacon and should be treated as such.
> 2) The basic 1pt webbing has lower storage capacity than all the items in the webbing selector, which is why the selector costs more, and why the basic webbing isn't just in the selector to begin with.

The thing with the watch beacon is that with multiple people using it, it just has the GPS filled with with the survival watch GPS beacons that are unnamed. 

The intent may have been an emergency beacon, but if someone is getting their shit kicked in chances are they aren't going to have time to remove it and toggle it on. If the beacon only lasted for X amount of time and automatically activated if your HP dropped below X amount, or it did the 'SOS AT XYZ LOCATION' announcement that the red button on the ships did, sure. But in it's current state, nobody is using it as an emergency beacon.

Most departments have multiple GPS readily available at round start and creating one only requires glass+steel.

Ive gotten multiple requests to allow renaming the survival watch because of it, because for people using the GPS there's just watch spam and no personalization.

As for the webbing, It seems like the file I was looking at (that has the drop pouches selectable) was an archived file, whoops.

KillianKirilenko:
Are that many people really wearing them, with beacons enabled, that it's a problem?

I figured the number of people just wearing *regular* watches could be counted on one hand, honestly, nevermind the survival ones.

Cameron-The-Raven:
> Are that many people really wearing them, with beacons enabled, that it's a problem?
> 
> I figured the number of people just wearing *regular* watches could be counted on one hand, honestly, nevermind the survival ones.

Yeah, downstream (Chomp) has the Exploration department and outsiders that use them. I can't speak for the number of how many since that's not tracked and I don't keep too close an eye on Chomp.

Our combat-oriented downstream (Outpost 21) also uses them a decent amount, where the "spend time enabling it" will get you killed since TTK is generally in seconds.

I know at least 2 people immediately off the top of my head (not including myself) from OP21 that uses them, with there more than likely being others that I haven't noticed, esp. since all watches look the same on the GPS. The server always has ~15-20 players during the rounds, so it's a decent portion of the player count.

code/modules/clothing/accessories/watches.dm
```diff
@@ -71,3 +71,15 @@
 	if(Adjacent(user))
 		gps.tracking = !gps.tracking
 		to_chat(user,span_notice("You turn the micro beacon [gps.tracking ? "on" : "off"]."))
+
+/obj/item/clothing/accessory/watch/survival/item_ctrl_click(mob/user)
+	. = ..()
+	if(!Adjacent(user) || (!user.a_intent == I_HELP))
+		return
+	var/new_name = tgui_input_text(user, "What would you like to name the GPS?", "Rename GPS", gps.gps_tag, 11)
+	new_name = sanitize(new_name, 11)
```
Kashargul: you are sanitizing twice, tgui input text intrnally sanitizes unless you removed that part

code/modules/clothing/accessories/watches.dm
```diff
@@ -71,3 +71,14 @@
...
+
+/obj/item/clothing/accessory/watch/survival/item_ctrl_click(mob/user)
+	. = ..()
+	if(!Adjacent(user) || (!user.a_intent == I_HELP))
+		return
+	var/new_name = tgui_input_text(user, "What would you like to name the GPS?", "Rename GPS", gps.gps_tag, 11)
+	if(!new_name || !Adjacent(user))
+		return FALSE
+	gps.gps_tag = uppertext(new_name)
+	name = "global positioning system ([gps.gps_tag])"
```
Cameron-The-Raven:
```suggestion
	gps.name = "global positioning system ([gps.gps_tag])"
```

PR #19575, opened 2026-07-07
Tesla Corona Analyzers

Poojawa:
introduces a means for Engineering to contribute to techweb point gain, and to reward safe high powered setups and to actually make the other various subtypes worth messing with.

Defaults are 10% efficiency with a 50 Kilowatt minimum to generate point values. Values are defines for future balancing since I only vaguely recall the average Tesla engine set up's output. This means each strike's base power value needs to be a minimum 500 Kilowatts in order to register and generate points for Techweb

coronas

corona animation

Additionally I took the liberty to edit all the coils to give them more visually obvious raised feet when unanchored and an open panel state for the maintenance hatch, as the only difference was... you saw slightly more of the coil colors.

:cl: Poojawa
add: Added Corona Analyzers to Tesla coil subtypes, able to assist R&D point generation
image: modified tesla coil sprites to be more visually obvious if anchored or the panel is open
refactor: cleaned up unneeded circuitry paths and grammar
/:cl:


Poojawa:
what the actual fuck, I didn't even touch any of the maps.

code/modules/power/tesla/coil.dm
```diff
@@ -272,7 +309,7 @@
 	if(Adjacent(user))
 		. += span_info("This tesla coil will amplify any power it receives by [round((((amp_eff) * 100) - 100), 0.1)]% of the original power when relaying it.")
 		. += span_info("Every jump the tesla makes reduces the effectiveness of the amplifier by 10%, meaning at 10 jumps, it stops increasing power.")
-		. += span_danger("This tesla coil will NOT produce produce energy.")
+		. += span_danger("This tesla coil will NOT produce energy.")
```
Poojawa, line 312: Hydroponics generator when

code/modules/power/tesla/coil.dm
```diff
@@ -420,6 +451,16 @@
 	connect_to_network()
 	update_icon()
 
+/obj/machinery/power/tesla_coil/research/pre_mapped
+	anchored = TRUE
+
+/obj/machinery/power/tesla_coil/research/pre_mapped/Initialize(mapload)
```
Kashargul: why ? this calls parent.

/obj/machinery/power/tesla_coil/research/Initialize(mapload)
	. = ..()
	connect_to_network()
	if(!linked_techweb)
		CONNECT_TO_RND_SERVER_ROUNDSTART(linked_techweb, src)
	update_icon()

This entire part is redundant

Poojawa: All the other ones call it too just to run connect_to_network()

and it only cares if it doesn't have a link, so pretty harmless

Kashargul: Well, if it calls parent, there's no need to call it again. It just adds useless init time. Double inits should be avoided.

Willburd:
> what the actual fuck, I didn't even touch any of the maps.

Stellar delight decided to have drainbamage and spawn a poi with an active edge somewhere. Always a fun time. Active edges fail CI, mostly to force mappers to try and fix things.

Poojawa:
Yeah I'm unfortunately installing VSC to more easily find and purge the active edges. They displease me.

code/modules/power/tesla/coil.dm
Poojawa: ideally power machines should have a LateInitialize that handles connect to network and could slot in the techweb connection there too. ~

Kashargul: They likely have but ye it makes no sense to dupe init code., If the parent handles it, the child should not repeat it.

code/modules/power/tesla/coil.dm
```diff
@@ -207,17 +209,57 @@
...
+	icon_state = "rpcoil0"
+	icontype = "rpcoil"
+	power_loss = CORONA_LOSS_FACTOR					// default 10 aka a 10% efficiency for powernet input.
+	zap_range = 1									// power leakage allows for a very short range jump but inherits the default efficiency for said zap.
+	var/power_judgement = CORONA_POWER_JUDGEMENT	// avoids cheese by setting a minimum threshold for power generation
+	var/last_shock_value = 0						// used for examine
+	var/sufficent_power = FALSE						// lets people see if they're putting enough juice in
+	var/datum/techweb/linked_techweb				// R&D server to contribute points to
+
+/obj/machinery/power/tesla_coil/research/LateInitialize()
```
Kashargul: I looked through, why would this even need late init? Normal init can handle tech web fine as is?

code/modules/power/tesla/coil.dm
```diff
@@ -50,10 +50,13 @@
 /obj/machinery/power/tesla_coil/Initialize(mapload)
 	. = ..()
 	set_wires(new /datum/wires/tesla_coil(src))
+	default_apply_parts()
 
-/obj/machinery/power/tesla_coil/Initialize(mapload)
+/obj/machinery/power/tesla_coil/LateInitialize()
```
Kashargul: I looked through, why would this even need late init? Normal init can handle tech web fine as is?

code/modules/power/tesla/coil.dm
Poojawa: It doesn't hurt to wait a bit just in case. It's what that verb us actually used for.

https://github.com/VOREStation/VOREStation/blob/e6a43a72487dc2915ac45fa3b29b04cf9f353abe/code/__defines/research.dm#L85-L87

Kashargul: You'll need to explicitly return init hint lateload in the objects initialize call then. initialize(mapload)
..()
stuff
return INITIALIZE_HINT_LATELOAD


so that the subsystem properly queries it up

code/modules/power/tesla/coil.dm
```diff
@@ -50,10 +50,13 @@
...
-
-/obj/machinery/power/tesla_coil/Initialize(mapload)
-	. = ..()
 	default_apply_parts()
+	return INITIALIZE_HINT_LATELOAD
+
+/obj/machinery/power/tesla_coil/LateInitialize()
+	if(anchored)
+		connect_to_network()
+	update_icon()
```
Kashargul, line 56: That entire part should stay on init, makes no sense to move to late init other than increasing init time and memory usage

code/modules/power/tesla/coil.dm
```diff
@@ -207,17 +209,63 @@
...
+	var/datum/techweb/linked_techweb				// R&D server to contribute points to
+
+/obj/machinery/power/tesla_coil/research/Initialize(mapload)
+	. = ..()
+	return INITIALIZE_HINT_LATELOAD
+
+/obj/machinery/power/tesla_coil/research/LateInitialize()
+	if(anchored)
+		connect_to_network()
+	update_icon()
```
Kashargul: this entire section from the if anchored to the update icon can be deleted then

code/modules/power/tesla/coil.dm
```diff
@@ -207,17 +209,63 @@
...
+	if(!linked_techweb)
```
Kashargul, line 231: This and the line below is fine to stay on late init

Kashargul (changes requested):

Kashargul (approved):

Cameron-The-Raven:
Looking and testing now.

A typical (t3 power PA) tesla will have ~20 tesla balls. meaning ~20 tesla strikes.
In testing, a normal setup w/ t1 parts (minitest standard, but with the collectors replaced with amplifiers) provided 173KW, 1730KW, and 870KW to the three corona analyzers in the following photo: 
dreamseeker_Q26SGXlPuj

With no connected powernet (no knots) it allows all three to generate points with the current 50KW define.

Removing the amplifiers and just placing down the research coils resulted in the following zap power: 
Code_aQvhrswAXc
dreamseeker_OkvnVySPlV


dreamseeker_cnacicMWEH
With a default t1 tesla: 
47 18: 1738260 power
47 20: 869130 power
47 24: 496645 power
47 26: 744968 power
49 18: 869130 power
49 26: 496645 power
51 18: 869130 power
51 26: 496645 power
53 18: 620807 power
53 26: 744968 power
55 19: 620807 power
55 21: 496645 power
55 24: 869130 power
55 26: 496645 power

Will need to do further testing.

Poojawa:
If you feel that's too low of a bar for point gen, it's easy enough to bump up the difficulty. 

I wanted to balance out the minimums without going overboard, and tune from there without being unfair. 

Minimum to 500kW, bump power loss back to 20, which would mean only 5% of the strike counts, and it'd still need to clear the threshold. since it does tap over a MW without too much effort, we really want to make sure you can't just swap 'em all over and hope to win. lol

Ideally the balance should be where you need to bounce a bolt through a power chain outwards away from the power plant itself. Since grounding rods that close really ought to be taking priority for the shocks and almost never actually permitting any power. But that was for TG tesla, which didn't have the various modes - and you had to disassemble the whole coil to flick it like the gas heater/cooler!

Cameron-The-Raven:
Yeah, in its current state you can have the primary tesla bolts do their thing (hit the closest coils, go down a path of amplifiers, then end at a collector) and then have all the 'excess' bolts that aren't grabbed by the coils be sucked up by the research coils that are just sitting around.

In terms of grounding rods, grounding rods have de-priority. It goes:
Coils -> Grounding Rods -> People -> Machinery

Coils can only take one shock per tick, while grounding rods absorb all the unused bolts. See: 
dreamseeker_myWPyrHrEc

Example of just tossing a few research coils into an already pre-established setup w/o any amps: 
dreamseeker_AqwdmX0cwz

And an example of some amp chains->collector 
image

The amp->chain collector results in ~280kW to 1.5MW depending on RNG.

Highest I got was 1.738mW on one of the setup lines there.

So I'd say 500kW seems about right

Poojawa:
Love the patches of space vomit too. lmao. 

Made sure the power loss factor applies, as the conditional just... allowed it to ignore the loss factor if it wasn't hooked up. 

Make a powernet with a SMES node on said powernet a hard requirement to generate points.

Now that the limiter is in, we probably will end up needing to kick it down again, probably 150kW or so since that should be within the 1.8MW casual peak.

code/modules/power/tesla/coil.dm
```diff
@@ -207,17 +206,62 @@
...
+	if(power_produced >= power_judgement)//is our power strike worthy?
+		sufficent_power = TRUE
+	else
+		sufficent_power = FALSE
+	if(powernet)// must have a hard connection to the powergrid
+		for(var/obj/machinery/power/smes/S in powernet.nodes)	//and a SMES unit for the full power grid experience. No running it in your basement or without proper wiring
+			if(S && istype(linked_techweb) && sufficent_power)
+				linked_techweb.add_point_type(TECHWEB_POINT_TYPE_GENERIC, 1)
+	playsound(src, 'sound/effects/lightningshock.ogg', 100, 1, extrarange = 5)
+	tesla_zap(src, zap_range, power_produced, current_jumps = current_jumps)
```
Cameron-The-Raven:
```suggestion
	if(powernet && sufficient_power && linked_techweb)// must have a hard connection to the powergrid
		for(var/obj/machinery/power/smes/S in powernet.nodes)	//and a SMES unit for the full power grid experience. No running it in your basement or without proper wiring
			linked_techweb.add_point_type(TECHWEB_POINT_TYPE_GENERIC, 1)
			break
	playsound(src, 'sound/effects/lightningshock.ogg', 100, 1, extrarange = 5)
	tesla_zap(src, zap_range, power_produced, current_jumps = current_jumps)
```

Moved some of those up from the for() loop to the check itself

Add a break here so you can't cheese it by putting multiple SMES down and attached to the grid.

Also I don't think we need to do an istype for the linked techweb. If the var somehow gets set to a non-techweb datum/obj/etc there's bigger problems, so we can just check for if it is.

code/modules/power/tesla/coil.dm
```diff
@@ -207,17 +206,62 @@
...
+
+/obj/machinery/power/tesla_coil/research/examine(mob/user)
+	. = ..()
+	. += span_notice("This corona analyzer has [sufficent_power ? "sufficent power output to produce additional research data points" : "insufficent power output for electrical corona analysis"].")
+	. += span_info("You will need more power than you think you do.")
```
Cameron-The-Raven:
```suggestion
	. += span_info("Will require a strike of at least [power_judgement / 1000]KW to produce research data points. Last strike was: [round(last_shock_value / 1000)]KW.")
```

Should work and give more info to engineers that check it...Might need to double check to make sure the round function works.

Cameron-The-Raven:
Adjusted it a little bit. Since it has to be hardwired (and will always have the CORONA_LOSS_FACTOR divisor) the 500kW made it so it'd require 5MW to hit it (Very much doable, but attempting to do that with T1 parts is impossible, which is when you'd want to get this going for science to research stuff faster) ...I lowered it to 100kW, so it'll now require 1MW.

Also did some cleanup of just using power_produced >= power_judgement instead of having a true/false var.

Poojawa:
Works for me!

PR #19577, opened 2026-07-07
Power related cleanup and tweaks

Poojawa:
Technically atomizing some things from my layered cable PR from Rogue Star 

SMES units given an open hatch state, cleaned up the "hybrid"'s redundant code as well. 

Initially was going to up port the Talon potato glow, but that'd already been done, As the originator of the code I decided to cut out the rogue star comments that were added. 

Porta Gens have a little EMP status on their examines now since how long is 10 minutes anyway in the spessmans.

No gameplay changes of note. 

:cl: Poojawa
qol: Multitool flagged items will display more information about the current cable's powernet status
fix: fixed cable power readouts from displaying negative Watts, 0ing out instead
image: SMES units now have an open panel state when the maintenance hatch is open
refactor: Utilized KILOWATTS, MEGAWATTS in power code on a lazy basis to improve readability.
/:cl:


code/_helpers/unsorted.dm
```diff
@@ -455,6 +455,8 @@ Turf and target are seperate in case you want to teleport some distance from a t
 
 // Format a power value in W, kW, MW, or GW.
 /proc/DisplayPower(powerused)
+	if(powerused <= 0)	//no negative numbers, please.
```
Kashargul, line 458: When can we even get negative numbers? Are they a normal condition or are we just covering up an issue?

Poojawa: without this cap, and with too high of a draw, it does do `there's -134450 W in the network` which is very silly. 

With the multitool reading changes, you can see that you have xW available, whatever is the load, and what excess is in the network on that particular check. Basically the same information you can gleam from the power monitoring console.

code/modules/power/port_gen.dm
```diff
@@ -85,15 +88,18 @@
...
 			if(prob(10)) stat &= BROKEN
-			duration = 300
+			base_duration = 3 MINUTES
 
 	stat |= EMPED
-	if(duration)
-		spawn(duration)
+	emp_duration = world.time + base_duration
+	if(emp_duration)
+		spawn(emp_duration)
```
Cameron-The-Raven:
```suggestion
		spawn(duration)
```
using emp_duration here will result in too long of a spawn.
Should be swapped from spawn to a timer in the future

code/modules/power/port_gen.dm
```diff
@@ -70,12 +71,14 @@
 			. += span_notice("The generator is on.")
 		else
 			. += span_notice("The generator is off.")
+		if(stat & EMPED)
+			. += span_warning("The generator's readout estimates [emp_duration - world.time] seconds remaining until it can restart.")
```
Kashargul: Does this even display seconds? looks more like 10ths. Byond internal time to seconds is * 0.1

Poojawa: It gives a block of numbers that may or may not be accurate, but it provides some idea without an exact idea

Cameron-The-Raven:
```suggestion
			. += span_warning("The generator's readout estimates [round((emp_duration - world.time) * 0.1)] seconds remaining until it can restart.")
```

Kashargul (approved):

Cameron-The-Raven (approved):

PR #19578, opened 2026-07-07
Chem Master QOL and fixes

Poojawa:
Up-ports Rogue star PRs 309, 359, and 487 addressing issues with the Chem Master

Chem master's buffer slashed to 300 units (Bluespace beaker) but gifted the ability to actually keep reagents in the machine to eject a beaker.

Overfilling mechanics implemented to prevent wasting reagents since they'd be just deleted otherwise.

:cl: Poojawa
qol: Chem Masters now permit a choice between beaker ejections, clearing or keeping the buffer inside the machine. Buffer has been changed to 300 (one bluespace beaker) as a result.
fix: Chem Masters now check for for potential overfilling, resolving issues where reagents would be deleted if the beaker was too full to accept the full amount.
fix: Prevents eggs and siffruit from being 'cracked into' the Chem Master or Chem Dispensor, fixing a null runtime that completely breaks the machines
/:cl:


tgui/packages/tgui/interfaces/ChemMaster/ChemMasterBeaker.tsx
```diff
@@ -21,10 +21,23 @@ export const ChemMasterBeaker = (props: {
...
     >
-      Eject and Clear Buffer
+      Eject and Keep Buffer
     </Button.Confirm>
   ) : (
     <Button icon="eject" disabled={!beaker} onClick={() => act('eject')}>
+      Eject and Keep Buffer
+    </Button>
+  );
+  const headerButton2 = bufferNonEmpty ? (
```
Kashargul: unused const?

Should just be combined above, because like this it won't show. There's no need to split it out and repeat the bufferempty logic

tgui/packages/tgui/interfaces/ChemMaster/ChemMasterBeaker.tsx
Kashargul: include the added buttons here, use <></> fragments to have 2 buttons without a parent component

Kashargul (changes requested):

tgui/packages/tgui/interfaces/ChemMaster/ChemMasterBeaker.tsx
```diff
@@ -10,12 +10,13 @@ import type { reagent } from './types';
...
   beaker: BooleanLike;
   beakerReagents: reagent[];
+  bufferReagents: reagent[];
   bufferNonEmpty: BooleanLike;
 }) => {
   const { act } = useBackend();
   const { beaker, beakerReagents, bufferNonEmpty } = props;
 
-  const headerButton = bufferNonEmpty ? (
+  const headerButton = (<> {bufferNonEmpty ? (
```
Kashargul, line 18: closer but proper styling would be

`bufferNonEmpty ? (<Button>...</Button><Button>...</Button> ) : (<Button>...</Button><Button>...</Button>)`

Repeating conditions aka the bufferNonEmpty shouldn't be done.

if it complaints that it lacks a parent, those buttons are put into a fragment group

`<><Button></Button><Button></Button></>`
Biome will state if the element needs a parent though

Poojawa: I meant to undo this particular commit because it was messy, but fixed now. Honk.

Poojawa:
it all works as I intend it to ingame now. Both eject buttons function, the buffer not only preserves information when you close the UI, but copy/pasta of the BeakerContents for a renaming into BufferContents ensures that that particular section is now independent as it should be. This is ready to go

tgui/packages/tgui/interfaces/ChemMaster/ChemMasterBuffer.tsx
```diff
@@ -2,17 +2,21 @@ import { useBackend } from 'tgui/backend';
...
 
-import { BeakerContents } from '../common/BeakerContents';
+import { BufferContents } from '../common/BufferContents';
 import { modalOpen } from '../common/ComplexModal';
 import { transferAmounts } from './constants';
-import type { reagent } from './types';
+import type { Data, reagent } from './types';
 
 export const ChemMasterBuffer = (props: {
+  beaker: BooleanLike;
```
Kashargul: why is beaker passed as props and used from data? Should just be used from props in that case

tgui/packages/tgui/interfaces/ChemMaster/ChemMasterBuffer.tsx
```diff
@@ -2,17 +2,21 @@ import { useBackend } from 'tgui/backend';
...
   mode: BooleanLike;
+  beakerReagents: reagent[];
   bufferReagents: reagent[];
 }) => {
   const { act } = useBackend();
+  const { data } = useBackend<Data>();
   const { mode, bufferReagents = [] } = props;
```
Kashargul, line 15: beaker would go here if passe by props

tgui/packages/tgui/interfaces/ChemMaster/ChemMasterBuffer.tsx
```diff
@@ -91,8 +95,69 @@ export const ChemMasterBuffer = (props: {
             </Stack>
           )}
         />
+      ) : bufferReagents.length > 0 ? (
+        <BufferContents
```
Kashargul: This looks like a large duplication of the above code which could be shortened down massively

tgui/packages/tgui/interfaces/common/BufferContents.tsx
```diff
@@ -0,0 +1,28 @@
+import { Box, Stack } from 'tgui-core/components';
```
Kashargul: This file is a heavy copy paste of the beaker one. Could be unified easily to reduce bundle size

Kashargul (commented):
Whilst functionally, that web code should be adjusted. Not sure if you want to do that yourself. But people often complained about bundle size in the past due to it making tgui less responsive and here are quite a few easily optimizable things

tgui/packages/tgui/interfaces/ChemMaster/ChemMasterBuffer.tsx
Poojawa: It is a 1:1 copy with minimal changes, and it works. soooo...

Poojawa:
I don't really think I'm at all equipped to handle the next stage of tgui optimizations...

Kashargul:
> I don't really think I'm at all equipped to handle the next stage of tgui optimizations...

If you want I can take that over and explain the diffs

Poojawa:
I really kinda don't want this PR to be spiraled out of the overall scope. The changes here are pretty trivial in the grand scheme of things. 

It doesn't make sense to word the copy pasta as a notable contributor to the overall bundle load and performance of other users. Surely something like the Character menu's UI's reliance on fetch-generating fancy little button renders of markings has a more significant impact than a bit of text that just... renders more text?

it does what it needs to do, and I can't imagine it gets any measurably more performant with a bit more squishing down on this particular UI.

Kashargul:
> I really kinda don't want this PR to be spiraled out of the overall scope. The changes here are pretty trivial in the grand scheme of things.
> 
> It doesn't make sense to word the copy pasta as a notable contributor to the overall bundle load and performance of other users. Surely something like the Character menu's UI's reliance on fetch-generating fancy little button renders of markings has a more significant impact than a bit of text that just... renders more text?
> 
> it does what it needs to do, and I can't imagine it gets any measurably more performant with a bit more squishing down on this particular UI.

Functionality is one thing, passing maintainer review another. If there're things pointed out and dismissed as "but it works" that won't fulfill the review passing.

Poojawa:
Oh so we're pulling rank like this now. Alright. 

Whatever. if you have to make sure you're getting commit credits in my PRs, then I'll just open them all up to edits.

Poojawa:
New day, reset of stress. 

realized that the buffer doesn't need any information about the beaker. Because it doesn't care about if the beaker is in the machine at all, just that there's reagents in there. 

In the same way, the beaker side doesn't need to worry about anything buffer related except that there's something in it for the eject buttons. 

Combined both into ChemicalContents,tsx because that seemed like a much better way of keeping their roles separate, but avoiding extra stuff like formatUnits from being exported. Updated the other menus that called for it.

Kashargul (approved):

PR #19583, opened 2026-07-07
Shock and Trauma adjustments & additions

Cameron-The-Raven:
## About The Pull Request
Gets rid of trauma_mod and replaces it wholesale with pain_mod, as they on paper sounded like they did the same thing but were vastly different, leading to confusion about what they did.

### Additions:
Traumatic shock now causes your HR to go to >120BPM if you're significantly deep in traumatic shock.

Neural Hypersensitivity kept in name but does its effect properly (100% increased pain from all sources). Point cost bumped up to 5 points to match the pain intolerance traits.

Inverse of Neural Hypersensitivity added: Pain Tolerance, Extreme. This provides a 50% pain resistance for 5 points.

### Balance:
Makes pain_mod now affect the traumatic shock calculation.
Traumatic shock will no longer 100% perma-weaken you at it's late stage, instead has a chance to weaken you.
Traumatic shock no longer does a 40 second weaken w/ 5% chances of bringing it back to 40 seconds each 2 seconds (that has the % prob called multiple times), instead goes from a 6-10 second weaken with the former probs.
Blurry eyes and stuttering now happens when either extremely injured or significantly in shock.

GRIT: Grit **did not function** as the game stated. While it was noted to 'reduce pain by 15%' this was at best _technically_ correct as it increased the pain threshold you start to accumulate trauma from 80 'trauma units' to 92 'trauma units', which would at worst be 6 extra damage and at best 12 extra damage. This did not reduce pain slowdown or anything of that sort.

Pain Tolerance, Minor does the same functionality as grit, along with its proper 'pain reducing' functionality.

### Fixes:
Achieving a HR > thready will give the thready readout instead of a blank readout.
Removes some unneeded unicode.
Clamps the HR calculation to keep from going to values that are out of the HR range.
Shock stages no longer 'stack up' and all run their % chances at the same time. This means less chat-spam and no more perma-weakens.
## Changelog
:cl: Diana
del: Grit removed. "Pain Tolerance, minor" covers everything grit was supposed to do, now.
add: Adds "Pain Tolerance, Extreme" which is an inverse of Neural Hypersensitivity. Provides 50% reduced pain.
add: Being in shock will now increase your heart rate.
balance: Pain modifier now affects traumatic shock
balance: Trauma no longer weakens you for 40 seconds repeatedly. Now only weakens for 6-10 seconds when it rolls.
balance: Blurry eyes happen when significantly damaged or getting deep in shock.
balance: Being deep in trauma no longer perma-weakens you, but now has a significant chance to keep you weakened.
fix: Shock stage no longer piles up the probabilities when in shock, instead only does the one appropriate for its stage. This reduces chatspam.
fix: Your HR can no longer go so high it reads nothing.
code: trauma_mod rolled into pain_mod
/:cl:


code/modules/mob/living/carbon/human/life.dm
```diff
@@ -1864,56 +1864,85 @@
...
+
+
+
+
+
+
+
+
+
+
```
Cameron-The-Raven:
```suggestion
```

Kashargul (approved):

PR #19584, opened 2026-07-08
Makes material costs use sheets amount define

TheCaramelion:
## About The Pull Request

Makes every material cost of objects use the defined sheet's amount. Nothing player facing.

## Changelog
:cl: Guti
code: Changed all material costs to use sheets define.
/:cl:


Kashargul:
You might need to check for the newly added things

PR #19586, opened 2026-07-08
Construction Menu Sorting

Poojawa:
up-ports Rogue Star 268 but with some extra additions like the various species of wood. renames steel_vr to the only material actually in said file, "plastitanium". 

Also includes the examine contextual hint to the frame parts, because these things are pretty vague otherwise. 

:cl: Poojawa
qol: Sorted the material construction menus
refactor: optimized the wood species recipe removals
qol: Frame parts give a contextual hint to usage on examine.
/:cl:


Poojawa:
how does renaming just a file break the material datum inside of it for Unit tests

Kashargul:
I need to take a closer look later, but there're multiple tests failing and even global init crashes with runtimes in the master controller

some material was broken from a quick look

Kashargul (approved):

PR #19569, opened 2026-07-04
Food Synthesizers

Poojawa:
Initial port, need to figure out what madness you people have done to tgui... and sprite sheets.

At this moment, it compiles but I haven't tested it yet.  Too eepy. Will rebuild tgui once I take a second lookover and see what parts of the UI needs updating. 

finishes this PR finally. 
#13020

Info from here: https://github.com/TS-Rogue-Star/Rogue-Star/pull/206

Significant changes was reducing Nutripaste from 30 power to 6, as this codebase doesn't have the nutriment bloat that RS has. 
Also placed it on tech webs, but haven't tested 

:cl: Poojawa
add: Added working Food Synthesizers, can print a variety of meals and even (opt in) crew cookies
/:cl:

PR #19594, opened 2026-07-09
enables food synth in techwebs

Kashargul:
Enables the food synthesizer from https://github.com/VOREStation/VOREStation/pull/19569 in the techwebs


Lilly-Lira:
Github bot gone rogue

Poojawa:
There shouldn't be any harm to adding these to tech web, as you'll still need to request the canister from Cargonia to supply 'em.

Willburd:
> There shouldn't be any harm to adding these to tech web, as you'll still need to request the canister from Cargonia to supply 'em.

I see no harm myself if they are cross department dependent to actually use, and don't come with any prefilled when constructed.

PR #19598, opened 2026-07-10
Removes some unsued vars [No GBP]

Cameron-The-Raven:
## About The Pull Request
Removes two unused vars from IDs


Poojawa:
would be an interesting antag treat if the IDs were greyscaled and you could fudge the colors/ 'photo' aspect with a telecrystal tool, like proper passport fraud in movies

PR #19619, opened 2026-07-17
rm verdigris

Kashargul:
More to take a look at the TG rust-g cave gen.... it would replace verdigris, but it's ~ 13% slower... so we should compare verdigris and this.

Needs a close comparisons of random generated mining caves, asteroids, buildings etc... to see if it's worth the swap or to keep our in repo map gen


code/modules/random_map/automata/automata.dm
```diff
@@ -1,24 +1,23 @@
...
 
 // Automata-specific procs and processing.
 /datum/random_map/automata/seed_map()
-	return // Do not seed, we use Verdigris for this now
+	return // Do not seed, we use rust-g for this now
 
 /datum/random_map/automata/generate_map()
-	map = verdigris_generate_automata(limit_x, limit_y, iterations, initial_wall_cell)
-	return
+	var/map_string = rustg_cave_system_generator_generate("[limit_x]", "[limit_y]", "\[]", "25", "1.5", "1", "30", "1", "15", "[initial_wall_cell]", "[iterations]", "6", "4", "1")
```
Kashargul, line 11: Tg uses map strings, our old system still uses lists

Willburd:
Working fine with non-standard map sizes. (tested outpost's map at 400x160). Was tested a while ago, but i figured I'd leave the report if it is forgotten.

Kashargul:
> Working fine with non-standard map sizes. (tested outpost's map at 400x160). Was tested a while ago, but i figured I'd leave the report if it is forgotten.

What are the thoughts about the new cave routing? As we add ~13% costs on that part

Willburd:
> > Working fine with non-standard map sizes. (tested outpost's map at 400x160). Was tested a while ago, but i figured I'd leave the report if it is forgotten.
> 
> What are the thoughts about the new cave routing? As we add ~13% costs on that part

No objection to the changes, at worst i will modify params if i can. It can hollow out some of our asteroids a bit, but even the old algo would do that sometimes. I know this is needed for SSmapping future work so I have no plans on standing in the way of it for everyone's sake because of cave shape differences.

code/modules/random_map/automata/automata.dm
Willburd: Would moving fully to strings in the mappingSS fix the 13% performance issue from the cavegen code changes? I'm wondering if it's stringification/reversing it overhead for the size of strings/list being made for each map.

Kashargul:
> > > Working fine with non-standard map sizes. (tested outpost's map at 400x160). Was tested a while ago, but i figured I'd leave the report if it is forgotten.
> > 
> > 
> > What are the thoughts about the new cave routing? As we add ~13% costs on that part
> 
> No objection to the changes, at worst i will modify params if i can. It can hollow out some of our asteroids a bit, but even the old algo would do that sometimes. I know this is needed for SSmapping future work so I have no plans on standing in the way of it for everyone's sake because of cave shape differences.

Moving fully to strings would bring back a large share of the lost performance, yes

PR #19624, opened 2026-07-20
fix multiple tests leaking dom into each other

Kashargul:
## Changelog
:cl:
fix: tgui test runs leaking their doms into each other
/:cl:


KillianKirilenko:
Nobody wants leaky doms everywhere, it's true.

PR #19630, opened 2026-07-24
Update Remoteviews to 1685

Willburd:
## About The Pull Request
516.1685 fixes a longstanding byond bug that remote views needed an ugly hack to work around. This removes the hack and fixes another issue with vore eggs blackscreening the player. The latter is a temporary fix, as holders need a better refactor. They do many strange things, and are reliant on a process() loop to fix their state when they break. This should be modernized to signals and fix the many race conditions they have using enter/enter.

## Changelog
-Removes ugly remoteview nested mob hack
-Fixes black screen when escaping a vore egg

:cl: Will
refactor: Updated remote view code for byond 516.1685
fix: blackscreen when escaping vore egg
/:cl:


Willburd:
Fully validated, the hack is no longer needed for any of the prior issue cases.

PR #19637, opened 2026-07-30
Code Cleanup for Food Synthesizers

Poojawa:
Did get a report about infinite food via construction system abuse, this is now fixed.

Cleaned up some overlay code with mutable_appearance magics now that I'm hyperwired on them. Mimicked finish_production for crew meals, added additional sanity checks to them too. and we're using client side matrix transforms now! woot!

As a reminder, you can order refills through cargo's vendor refill section.

:cl:Poojawa
fix: fixed infinite food supplies via construction abuse of the Food Synthesizer
code: Hardened some edge case issues involved with the reagent cartridge. Byond needs its references!
/:cl:


code/modules/food/kitchen/cooking_machines/foodsynthesizer.dm
```diff
@@ -57,9 +57,23 @@
...
+	var/obj/item/reagent_containers/synthdispcart/existing_cart = locate() in contents
+	if(existing_cart)
+		cart = existing_cart
+	//Map load spawn a cart
+	else if(mapload)
+		cart = new /obj/item/reagent_containers/synthdispcart(src)
+
+	//failsafe just in case.
+	if(!istype(cart))
+		cart = null
```
Kashargul: That code makes little sense.

Cart was defaulted to null in the define above, so it never should be a path and always be null or a real object

Kashargul (changes requested):

code/modules/food/kitchen/cooking_machines/foodsynthesizer.dm
Poojawa: I figure as much, leftover from actually figuring out the order of operations to prevent multiple from spawning.

Poojawa:
@Kashargul Is there any update for this?

Kashargul:
> @Kashargul Is there any update for this?

Hopefully will have some time end of the week or next week

Kashargul (approved):

PR #19639, opened 2026-07-31
Secured Armory Gunlockers

Poojawa:
Initial creation and info here: https://github.com/TS-Rogue-Star/Rogue-Star/pull/207

I promise this is the last time that I'm creating a new UI from bastardized elements. At least until I start aggressively touching clothing vending machines to make custom clothes. heh heh heh. 

New changes: 

- Removed hardcoded 4 slot limit and implementation, these can be map edited or a child subtype with different positioning. UI should adapt to how the guns themselves are arranged, maintaining the ability to make chonkier ones. 
- Rack Slots should arrange nicely on a grid formation for gun overlays, based on the number of rows, 4slots|1row, 2|2, 1|4, etc. 
- Initial test to utilize SSassets for the UI gun icons. This is future proofing for custom guns with attachments and the sort.
- proc/scatter_contents added to closets, for use with ex_act to throw objects using the throw_at_random() handling. Instead of the guns being auto-deleted by someone using a blasting charge on the doors, it'll instead either bust the lock on the doors, or throw the contents all over. This includes sev = 1 explosions.
- You can open them without access requirements if the security alert level is red or delta

Might add a section for ammo storage? Maybe? Could also add ammo charging to subtypes of 'em.

:cl:Poojawa
add: New and Improved armory gun lockers! Begone terrible armory racks!
/:cl:


code/game/objects/structures/crates_lockers/closets/secure/guncabinet.dm
```diff
@@ -59,3 +79,435 @@
...
+				"depleted" = TRUE,
+				"iconKey" = null
+			))
+			continue
+
+		// Sync gun state before reading ammo and icon_state
+		G.update_icon()
+		var/current_charge = get_gun_ammo(G)
+		var/max_charge = get_max_gun_ammo(G)
+		slots += list(list(
```
Kashargul: UNTYPED_LIST_ADD

code/game/objects/structures/crates_lockers/closets/secure/guncabinet.dm
```diff
@@ -59,3 +79,435 @@
...
+		"open" = opened,
+		"max_gun_rows" = max_gun_rows,
+		"slots_per_row" = slots_per_row
+		)
+	var/list/slots = list()
+	for(var/i in 1 to length(rack_slots))
+		var/obj/item/gun/G = rack_slots[i]
+
+		if(!G)
+			slots += list(list(
```
Kashargul: UNTYPED_LIST_ADD

code/game/objects/structures/crates_lockers/closets/secure/guncabinet.dm
```diff
@@ -59,3 +79,435 @@
...
+			ui.user.put_in_hands(occupant)
+			to_chat(ui.user, span_notice("You take [occupant.name] from slot [slot_idx]."))
+			rack_slots[slot_idx] = null
+			SStgui.update_uis(src)
+		else
+			var/obj/item/gun/held_gun = ui.user.get_active_held_item()
+			if(istype(held_gun) && check_weapon(held_gun, ui.user))
+				if(ui.user.unEquip(held_gun, src))
+					rack_slots[slot_idx] = held_gun
+					SStgui.update_uis(src)
```
Kashargul: why? We are inside tgui_act, returning TRUE requests and update

code/game/objects/structures/crates_lockers/closets/secure/guncabinet.dm
```diff
@@ -59,3 +79,435 @@
...
+		if(!opened)
+			to_chat(ui.user, span_notice("The cabinet doors are closed."))
+			return TRUE
+
+		var/obj/item/gun/occupant = rack_slots[slot_idx]
+		if(occupant)
+			ui.user.put_in_hands(occupant)
+			to_chat(ui.user, span_notice("You take [occupant.name] from slot [slot_idx]."))
+			rack_slots[slot_idx] = null
+			SStgui.update_uis(src)
```
Kashargul: why?

code/game/objects/structures/crates_lockers/closets/secure/guncabinet.dm
```diff
@@ -59,3 +79,435 @@
...
+	return 0
+
+/obj/structure/closet/secure_closet/guncabinet/fancy/ex_act(severity)
+	if(severity >= 2)	//study enough to survive door knockers
+		if(!opened)
+			welded = TRUE	//simulate being cut open for ease of repair logic
+			locked = FALSE
+			doorstatus = CABINET_BROKEN
+			update_icon()
+			SStgui.update_uis(src)
```
Kashargul: if this alters state, the ui state handler does this

code/game/objects/structures/crates_lockers/closets/secure/guncabinet.dm
```diff
@@ -59,3 +79,435 @@
...
+			visible_message(span_warning("The door blow open!"))
+		else
+			visible_message(span_warning("The force of the blast scatters [src]'s weapons everywhere!"))
+			scatter_contents(max_range = 3, throw_speed = 2)
+			SStgui.update_uis(src)
```
Kashargul: as above

code/game/objects/structures/crates_lockers/closets/secure/guncabinet.dm
```diff
@@ -59,3 +79,435 @@
...
+			span_notice("You start [anchored ? "un" : ""]securing [src] [anchored ? "from" : "to"] the floor."))
+		if(do_after(src, user, 2 SECONDS * Wench.toolspeed))
+			anchored = !anchored
+			to_chat(user, span_notice("You [anchored ? "" : "un"]secured [src]!"))
+		return
+
+	if(I.has_tool_quality(IS_WELDER))
+		var/obj/item/weldingtool/Wald = I.get_welder()
+		if(!opened && locked)
+			if(!do_after(user, src, 3 SECONDS * Wald.toolspeed))
```
Kashargul: missing do_after target

code/game/objects/structures/crates_lockers/closets/secure/guncabinet.dm
```diff
@@ -59,3 +79,435 @@
...
+				return
+			welded = TRUE
+			locked = FALSE
+			doorstatus = CABINET_BROKEN
+			update_icon()
+			user.visible_message(span_warning("[user] cuts through [src]'s lock with [I]!"))
+			return
+
+		if(!opened && doorstatus == CABINET_REPAIR)
+			if(!do_after(user, src, 3 SECONDS * Wald.toolspeed))
```
Kashargul: as above

code/game/objects/structures/crates_lockers/closets/secure/guncabinet.dm
```diff
@@ -59,3 +79,435 @@
...
+		return
+
+	if(welded && doorstatus == CABINET_BROKEN && istype(I, repair_material))
+		var/obj/item/stack/M = I
+		if(M.get_amount() < 2)
+			to_chat(user, span_warning("You need at least two sheets to repair the doors."))
+			return
+
+		to_chat(user, span_notice("You begin applying materials to repair [src]..."))
+		if(do_after(user, 3 SECONDS, target = src) && doorstatus == CABINET_BROKEN)
```
Kashargul: as above

tgui/packages/tgui/interfaces/GunLocker/GunLocker.tsx
```diff
@@ -0,0 +1,170 @@
+import { Box, Button, ProgressBar, Section, Stack } from 'tgui-core/components';
+import { useBackend } from 'tgui/backend';
+import type { BooleanLike } from 'tgui-core/react';
+
+export type SlotData = {
```
Kashargul: no export needed if we don't split the file for the types

tgui/packages/tgui/interfaces/GunLocker/GunLocker.tsx
```diff
@@ -0,0 +1,170 @@
...
+  slots_per_row: number;
+  slots: SlotData[];
+};
+
+export type GunLockerStaticData = {
+  icon_urls: Record<string, string>;
+};
+
+export const GunLocker = (props) => {
+  const { act, data, staticData: rawStatic } = useBackend<GunLockerData>();
```
Kashargul: what is that staticData handling here? Is this an attempt to badly introduce a state?

tgui/packages/tgui/interfaces/GunLocker/GunLocker.tsx
```diff
@@ -0,0 +1,170 @@
...
+        <Stack vertical>
+          {rows.map((rowSlots, rowIndex) => (
+            <Stack.Item key={rowIndex} mb={1}>
+              <Stack>
+                {rowSlots.map((slot) => {
+                  const iconUrl = slot.iconKey ? icon_urls[slot.iconKey] : null;
+                  const hasWeapon = Boolean(slot.ref);
+
+                  return (
+                    <Stack.Item key={slot.index} grow basis={0} mx={0.5}>
```
Kashargul: code below seems heavily AI coded and non tgui standard explain please

style is used for usual inline tgui props which are preferred as they won't pose issues on major tgui version changes

The tgui Image element isn't used, instead the default hml one

Kashargul (changes requested):

code/game/objects/structures/crates_lockers/closets/secure/guncabinet.dm
```diff
@@ -59,3 +79,435 @@
...
+					open(ui.user)
+				return TRUE
+
+		if("lock")
+			if(!opened && !emagged)
+				locked = !locked
+				to_chat(ui.user, span_notice("You [locked ? "lock" : "unlock"] [src]."))
+				return TRUE
+
+	if(findtext(action, "rackslot"))
```
ShadowLarkens: what?? use a parameter
```tsx
act("rackslot", { idx: gun.idx });
```
```dm
var/slot_idx = params["idx"]
```

code/game/objects/structures/crates_lockers/closets/secure/guncabinet.dm
ShadowLarkens: this should be an `update_static_data_for_all_viewers` yeah
unless you ditch that stuff for the appearance refs I suggested

code/game/objects/structures/crates_lockers/closets/secure/guncabinet.dm
ShadowLarkens: should be an `update_static_data_for_all_viewers`

code/game/objects/structures/crates_lockers/closets/secure/guncabinet.dm
ShadowLarkens: yeah these should just be in ui_data, autoupdate is turned on

code/game/objects/structures/crates_lockers/closets/secure/guncabinet.dm
```diff
@@ -59,3 +79,435 @@
...
+	var/slots_per_row = DEFAULT_RACK_SLOTS
+	//Use these offsets if you add a new type of gunlocker that has a different point for guns to be overlayed onto
+	var/gun_sprite_offsetx = 0
+	var/gun_sprite_offsety = 0
+	//Use these if you're adjusting how close or wide the slots are
+	var/gun_sprite_spacingx = GUNCABINET_SPACER_X
+	var/gun_sprite_spacingy = GUNCABINET_SPACER_Y
+
+	/// Slot array storing references to placed weapons
+	var/list/obj/item/gun/rack_slots = list()
```
ShadowLarkens: you should just use `contents` and pass `REF(i)` back and forth....

code/game/objects/structures/crates_lockers/closets/secure/guncabinet.dm
```diff
@@ -59,3 +79,435 @@
...
+/obj/structure/closet/secure_closet/guncabinet/fancy/update_icon()
+	cut_overlays()
+
+	var/total_slots = length(rack_slots)
+	if(!total_slots)
+		return
+
+	// Add weapon sprites in fixed slots
+	var/slots_per_row = ceil(total_slots / max_gun_rows)
+	for(var/i in 1 to total_slots)
```
ShadowLarkens: contents order is stable so it shouldn't be a problem to just iterate `for(var/obj/item/gun/G in src)`

code/game/objects/structures/crates_lockers/closets/secure/guncabinet.dm
```diff
@@ -59,3 +79,435 @@
...
+	guninfo.Cut()
+	return ..()
+
+/obj/structure/closet/secure_closet/guncabinet/fancy/tgui_interact(mob/user, datum/tgui/ui)
+	ui = SStgui.try_update_ui(user, src, ui)
+	if(!ui)
+		ui = new(user, src, "GunLocker")
+		ui.open()
+
+/obj/structure/closet/secure_closet/guncabinet/fancy/tgui_static_data(mob/user)
```
Kashargul: considering performance, this really should use a spritesheet containing all guns and their states then use the css class from there

code/game/objects/structures/crates_lockers/closets/secure/guncabinet.dm
```diff
@@ -59,3 +79,435 @@
...
+	if(opened && istype(I, /obj/item/gun))
+		tgui_interact(user)
+		return
+
+	if(I.has_tool_quality(TOOL_WRENCH))
+		var/obj/item/tool/wrench/Wench = I
+		user.visible_message(
+			"[user] begins [anchored ? "un" : ""]securing [src] [anchored ? "from" : "to"] the floor.",
+			span_notice("You start [anchored ? "un" : ""]securing [src] [anchored ? "from" : "to"] the floor."))
+		if(do_after(src, user, 2 SECONDS * Wench.toolspeed))
```
Kashargul: here as well

code/game/objects/structures/crates_lockers/closets/secure/guncabinet.dm
```diff
@@ -59,3 +79,435 @@
...
+		//utilize getFlatIcon to squish all the various overlays down, for weapon attachments and the sort if any.
+		if (!SSassets.cache[asset_name])
+			var/icon/flat_icon = getFlatIcon(G)
+			SSassets.transport.register_asset(asset_name, flat_icon)
+
+		if(user?.client)
+			SSassets.transport.send_assets(user.client, asset_name)
+		icon_urls[icon_key] = SSassets.transport.get_asset_url(asset_name)
+
+	data["icon_urls"] = icon_urls
```
ShadowLarkens: so 516 has a neat thing, if a client is aware of an appearance they can get it in the browser with REF(x) or REF(x.appearance)
you can make sure that clients are aware of the appearances of all guns in a locker by doing something along these lines
```dm
/obj/structure/closet/secure_closet/guncabinet/fancy
	var/atom/movable/overlay/invisible_appearance_holder

/obj/structure/closet/secure_closet/guncabinet/fancy/Initialize()
	. = ..()
	invisible_appearance_holder = new /atom/movable/overlay(src)
	invisible_appearance_holder.invisibility = INVISIBILITY_MAXIMUM
	vis_contents += invisible_appearance_holder

/obj/structure/closet/secure_closet/guncabinet/fancy/Entered(atom/movable/AM)
	. = ..()
	invisible_appearance_holder.vis_contents += AM

/obj/structure/closet/secure_closet/guncabinet/fancy/Exited(atom/movable/AM)
	. = ..()
	invisible_appearance_holder.vis_contents -= AM
```

and then in your UI data you can just pass `"ref" = REF(gun)` and in the UI use this
```tsx
/**
 * Waits until two XMLHttpRequests have loaded at iconSrc before calling cb().
 * @param iconSrc
 * @param cb
 */
function getTwice(iconSrc: string, cb: () => void) {
  const xhr = new XMLHttpRequest();
  // Block effect until we load
  xhr.open('GET', iconSrc + '?preload');
  xhr.send();
  xhr.onload = () => {
    const xhr = new XMLHttpRequest();
    // Block effect until we load
    xhr.open('GET', iconSrc + '?preload2');
    xhr.send();
    xhr.onload = cb;
  };
}

export const AppearanceDisplay = (props: { iconSrc: string }) => {
  const { iconSrc } = props;
  const [icon, setIcon] = useState<string>();

  // This forces two XMLHttpRequests to go through
  // before we try and render the icon for real.
  // Basically just makes sure BYOND knows we really want this icon instead of possibly getting back a transparent png.
  useEffect(() => {
    getTwice(iconSrc, () => {
      setIcon(iconSrc);
    });
  }, [iconSrc]);

  if (icon) {
    return (
      <Image fixErrors src={icon} ml={-1} mt={-1} height="64px" width="64px" />
    );
  } else {
    return <Icon name="spinner" size={2.2} spin color="gray" />;
  }
};
```
```
<AppearanceDisplay iconSrc={gun.ref} />
```

ShadowLarkens: *this should probably be an actual tgui component, ask Kash about it

code/modules/projectiles/gunlocker_helpers.dm
```diff
@@ -0,0 +1,308 @@
+//Gun classifications for locker sanity
+#define GUN_SIDEARM	0	//one handed sidearms, pistols and the sort.
+#define GUN_LONGARM	1	//shotguns, rifles, etc.
+#define GUN_HEAVY	2	//Stuff too large to sprite in the lockers.
```
ShadowLarkens, line 4: if these are used in more than one file they should be in the __DEFINES folder, not redefined per file

Poojawa:
I do keep saying that I still have no idea what I'm doing with tgui stuff. Kitbashing only works so well and half the time stuff seems to either be tgui_data or ui_data ... and I still have no idea what counts as 'standard tgui' when it's completely different between codebases. I can't even function match because of the subtle differences between 'em. 

Been trying to simplify what I had hardcoded before with this same system when porting from RS. Tried to pre-emptively shake out the old inferno code for typescript but it's like, I cannot visualize how this stuff works. So that's still on me of course. 

With that out of the way, 

The actual structural in game sprite doesn't care (nor should it) about whatever gun icon. If it's in the locker, it's been assigned a locker specific sprite that's been positioned relative to the slot. Don't need to care about vis_contents unless that's specifically something that is UI dependent as well.

The fancy tgui however, I wanted to experiment with handling custom guns with variable overlays. it *could* be removed but utilizing SSassets seemed like a performant way of getting the dozens of potentials + cache that state. After the discussion about sprite sheets in relation to the character creator assets, I decided while it would be nice, it would require potentially indexing ***every*** gun with ***every*** state. Blame Outpost21 for having this in mind. :P 

That leaves just a raw getFlatIcon run and no real cache (Old behavior), or dabbling in SSassets for content delivery to maybe reduce server costs? I dunno! just trying to keep accounting for making is as lightweight as possible on grandpa Byond.

code/modules/projectiles/gunlocker_helpers.dm
Poojawa: These are only utilized Here and in the lockers, modular friendly!

tgui/packages/tgui/interfaces/GunLocker/GunLocker.tsx
Poojawa: this was vibe machine translated from the brutalized old inferno code into what I can only assume was typescript, 

I'll try to break it down into the similar formatting of the original code inspiration, being Janicart code.

Poojawa: I actually have no idea what exactly is actually wrong with it. if I'm being completely honest. Like, other things use style?? what part of it is bad? is it missing the window component and that's why it's bad?

I cannot figure out what exactly is wrong with any of it at this point, or why it's bad. there's zero documentation on your expectations and I don't know how to resolve it. Suggestions are welcome.

Poojawa:
Everything's working now.

tgui/packages/tgui/interfaces/GunLocker/GunLocker.tsx
Kashargul: Let's give a short summary:

Window: It contains the hooks required for proper rendering and layout. Skipping it breaks several core functionalities, so it should always be used as the root component.

Scrollables: Wrapping the entire window in a scrollable is an outdated pattern. The recommended approach is to follow the 100% height layout model using stack fill, stack items (with grow if a child should fill remaining space), and section fill. This is especially important with the upcoming TGUI Next release, where UIs are expected to follow these modern layout standards. In practice, scrollables should only be added to the specific sections that actually require scrolling.

Props vs. Style: Prefer component props over inline styles whenever possible. Layout and theming props are designed to work with TGUI's internal CSS and remain compatible with future changes. For example, <Stack width="100%">respects any layout calculations performed by TGUI, whereas style={{ width: '100%' }} bypasses those mechanisms and may lead to inconsistent behavior.

Img vs. Image: Img is native, but especially when people didn't yet have received the assets (slow connection) it doesn't use any of the internal re render options Image usage which therefore is a lot more robust for TGUI usage.

I don't have the time currently to go into all details. But you can find most of that on the tgui-core repo... 

As said before, if you get stuck with TGUI, asking is often better than trying to AI code it. AI is build for general web dev. Whilst TGUI uses React, its internals differ and AI often fails on exactly those things. It defaults to common web code without specifics a game UI needs compared to a website.

code/game/objects/structures/crates_lockers/closets/secure/guncabinet.dm
Kashargul: After discussing with other devs, the approach Shadow recommended is the best here. It will only pass the reference to our UI and we render the object natively without needing any flat icon building.

Poojawa: vis_contents has almost no utility from what I glean from other examples in the codebase to the desired effect.

I'm not trying to render planes or special effects, it's just overlay building... 




The UI product is 100% purely just for the UI icon,

the structure overlays use a completely different accounted for icon and icon_state + offsets, no additional systems on there at all.

Kashargul: You miss the point, 516 added a way where tgui can directly get the icon from the game itself through the object reference. So instead of doing

Build asset, register, send asset, render asset

what 516 added allows the following
send object reference to tgui, tgui renders it directly from the byond local game data

Reference -> local game data

no icon building needed, no assets needed. As long the object is in "sight", which here the vis content workaround tricks the client into, tgui can render the actual object like the game does. (We just hijack the byond internal, visible object)

You skip all the icon building, asset caching and sending for the same result.

Lower server cpu load
Lower network load
Lower delay between image being available

Poojawa: Ah, okay. A bit more sense put that way, pfpf. I'll fiddle and see.

code/game/objects/structures/crates_lockers/closets/secure/guncabinet.dm
Poojawa: Somethin' funky happens with that, Probably alpha layer nonsense from byond itself.

Poojawa:
Also since this has been sitting

Currently the UI doesn't get a full color image of the gun, since it's invisible and therefor gets a black overlay squishing down. 

Everything else has been cleaned up and working. I've just been fiddling with usability at this point

Kashargul:
> Also since this has been sitting
> 
> 
> 
> Currently the UI doesn't get a full color image of the gun, since it's invisible and therefor gets a black overlay squishing down. 
> 
> 
> 
> Everything else has been cleaned up and working. I've just been fiddlingwith usability at this point

I'll take a look when time allows.

Poojawa:
This still needs a review for this experimental system I was asked to add.

PR #19641, opened 2026-07-31
Tilt code

WolvesAndOwls:
## About The Pull Request
This PR adds "Pixel tilt" functionality to living mobs. Currently, the keybinding is shift+East/West.
## Changelog
Adds the ability to rotate sprites for flavor reasons like pixel shifting.
:cl:
add: Pixel tilting for living mobs
/:cl:


WolvesAndOwls:
I would like feedback on this change. I like pixel tilting, but I'm worried about the keybinding being shift+east and west.

Cameron-The-Raven:
Will need add it so it resets / add restrictions on when it can happen. 

Ex: Buckled, status effect applied such as weakened/paralyzed, moving into resting, STAT changing from 0 to 1/2, moving undoing it (like how moving undoes pixel shifting)
Will have to go through and find all said cases, but those are the first few that immediately come to mind.

WolvesAndOwls:
> Will need add it so it resets / add restrictions on when it can happen.
> 
> Ex: Buckled, status effect applied such as weakened/paralyzed, moving into resting, STAT changing from 0 to 1/2, moving undoing it (like how moving undoes pixel shifting) Will have to go through and find all said cases, but those are the first few that immediately come to mind.

Alright, I put it through the ropes. If all else fails, resting will reset the tilt factor, but I was unable to find an edge case that wasn't covered (I'm sure there is one.)

code/modules/mob/mob.dm
```diff
@@ -1189,6 +1189,28 @@
 		pixel_x++
 		is_shifted = TRUE
 
+/mob/verb/tiltcounterclock()
+	set hidden = TRUE
+	if(world.time <= allowtilttime) //This is a sloppy "wait" fix to work around tilting durring the get-up animation
+		return FALSE
+	if(!canface() || !canmove || lying || stunned || sleeping || resting)
```
Kashargul: incapacitated()

WolvesAndOwls: Done and verified working

WolvesAndOwls:
All good!

Cameron-The-Raven:
With a combination of flips and resting I managed to get myself permanently tilted.
dreamseeker_YBEMc2cwXk

dreamseeker_lhqKiYHAzm
dreamseeker_0AqMTxlI8q


It seems like it can be reproduced with just resting?

Steps: Tilt yourself. Rest. Unrest. Move. You're now permanently tilted.

code/modules/mob/mob.dm
```diff
@@ -1189,6 +1189,28 @@
 		pixel_x++
 		is_shifted = TRUE
 
+/mob/verb/tiltcounterclock()
+	set hidden = TRUE
+	if(world.time <= allowtilttime) //This is a sloppy "wait" fix to work around tilting durring the get-up animation
+		return FALSE
+	if(!canface() || incapacitated(63))
```
Cameron-The-Raven: Please use the proper flags here instead of a hardcoded number

code/modules/mob/mob.dm
```diff
@@ -1189,6 +1189,28 @@
...
+	if(how_tilted >= -maximum_tilt)
+		transform = turn(transform, -5)
+		how_tilted = how_tilted - 5
+		is_shifted = TRUE
+
+/mob/verb/tiltclock()
+	set hidden = TRUE
+	if(world.time <= allowtilttime) //This is a sloppy "wait" fix to work around tilting durring the get-up animation
+		return FALSE
+	if(!canface() || incapacitated(63))
```
Cameron-The-Raven: Ditto

code/modules/mob/living/living_movement.dm
```diff
@@ -270,6 +270,8 @@ default behaviour is:
 		pixel_y = default_pixel_y
 		layer = initial(layer)
 		plane = initial(plane)
+		transform = turn(transform, -how_tilted)
```
Cameron-The-Raven: May want to just call update_transform() here? In testing it seems like it works fine, but I didn't extensively test with it.

Cameron-The-Raven (changes requested):

code/modules/mob/mob.dm
WolvesAndOwls: Fixed. Now using INCAPACITATED_ALL, I didn't see it before

code/modules/mob/mob.dm
WolvesAndOwls: Same here.

code/modules/mob/living/living_movement.dm
WolvesAndOwls: Swapped and seems to work fine! Less accident prone as well.

WolvesAndOwls:
> With a combination of flips and resting I managed to get myself permanently tilted.
> 
> It seems like it can be reproduced with just resting?
> 
> Steps: Tilt yourself. Rest. Unrest. Move. You're now permanently tilted.

Fixed. I wasn't clearing the tilt UNTIL the mob moved. If the mob never moved, the tilt var never reset and rising from resting would keep you upright with an erroneous tilt var. After you move, the var would assume that your tilt was "normal" and mess up your tilt. While this was resetable by resting and unresting, it shouldn't be a problem anymore.

Cameron-The-Raven (approved):

PR #19643, opened 2026-08-02
Usr Wire Fix

Willburd:
## About The Pull Request
Several instances of usr are present in wire pulsing/cutting/mending. This usr value can be in an invalid state if not called from a verb, such as emps by mobs. I have corrected the few instances where this isn't actually an issue, but cases where shock() is called remain unsolvable.

I'm not sure what to do here, as removing the shock is bad for gameplay. However i'd need to modify all wire procs to pass user, and then nullcheck it for each shock(if shock already doesn't do so.) In this case we significantly deviate from TG wire code. TG has this same code for air alarms, and appears to have removed or replaced shock() on wire modification in most other places.  ( as seen here on TG's repo https://github.com/tgstation/tgstation/blob/197496bca0ee32b9883b09df09aca6a970ff46a7/code/datums/wires/airalarm.dm#L59 ) 

Shock() behavior is not the same as electrifying the device, as electrified devices are often per-object handling on touching them to electrify the user doing so AFTER it has been cut. Shock() behavior instantly zaps the user performing the wire cut. Replacing one with the other seems unsatisfactory. 

This issue is exclusively with shock() as all other usr related issues were just tgui static data updating, which could be replaced by "update all users" which TG also does.

This PR needs discussion tag, 

Locations in code with issues are marked in this PR.

## Changelog
Replaced usr from wire interaction code.

:cl: Will
code: replaced usr from wire interaction behaviors
/:cl:

### DIFF
--- a/code/modules/power/power.dm
+++ b/code/modules/power/power.dm
@@ -320,6 +320,8 @@
 
 	if (!cell && !PN)
 		return 0
+	if(!isliving(M))
+		return 0
 	var/PN_damage = 0
 	var/cell_damage = 0
 	if (PN)
@@ -333,8 +335,6 @@
 	else
 		power_source = cell
 		shock_damage = cell_damage
-	if(!isliving(M))
-		return 0
 	var/drained_hp = M.electrocute_act(shock_damage, source, siemens_coeff) //zzzzzzap!
 	var/drained_energy = drained_hp*20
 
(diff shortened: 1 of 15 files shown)


code/modules/power/power.dm
```diff
@@ -333,6 +333,8 @@
 	else
 		power_source = cell
 		shock_damage = cell_damage
+	if(!isliving(M))
```
Willburd: Why so late in the proc if we are going to drop out after doing all of this logic?

Cameron-The-Raven:
Did some sanity.
/shock checks for isliving
rig/shock checks for isliving,
electrocute_mob checks for isliving (later after all the spark calc stuff but before the damage )

Cameron-The-Raven:
fixing sev 1 bug, discussion on permanent fix will be had later.

PR #19647, opened 2026-08-03
Egg

Very-Soft:
## About The Pull Request

Adds an easter egg to the game

## Changelog
:cl:
add: Added a rare easter egg to the game
/:cl:


JoanRisu:
This was a request by me. Assuming it passes the checks and meets current coding practices by the team, I'd like this added as soon as possible.

code/modules/awaymissions/egg_event.dm
```diff
@@ -0,0 +1,96 @@
...
+	E.expire()
+
+/area/egg_tree_zone
+	name = "Unregistered"
+	icon = 'icons/effects/effects.dmi'
+	icon_state = "glitch"
+	dynamic_lighting = FALSE
+
+/area/egg_tree_zone/man
+	enter_message = "<span class='cult'>(Well.  There is a man here.  He looks at you like you're a bit lost.)</span>"
```
Kashargul: Legacy span, use span macros

code/modules/awaymissions/egg_event.dm
```diff
@@ -0,0 +1,96 @@
...
+	if("[M.name] - [M.ckey]" in visitors)
+		return
+	. = ..()
+	if(isliving(M))
+		var/mob/living/L = M
+		var/obj/item/reagent_containers/food/snacks/egg/E = new(get_turf(L))
+		if(L.put_in_any_hand_if_possible(E, TRUE))
+			to_chat(L ,span_cult("(The man smiles and hands you what you've been looking for.)"))
+			to_chat(L ,span_boldannounce("* You got an egg."))
+			visitors |= "[L.name] - [L.ckey]"
```
Kashargul: why do we use |= here when we already check that it is inside the list further up? +=

Kashargul (changes requested):

code/modules/awaymissions/egg_event.dm
```diff
@@ -0,0 +1,96 @@
...
+	if(rand(1,1000) != 66)
+		return FALSE
+	if(!isturf(teleport_destination))
+		return FALSE
+	if(!isliving(to_teleport))
+		return FALSE
+	if(!to_teleport.ckey)
+		return FALSE
+	var/turf/egg_destination
+	for(var/obj/effect/landmark/egg_teleport/E in world)
```
Kashargul: do not iterate through world.... please use a proper global list for those effects in GLOB.landmarks_list

Kashargul (dismissed):

code/modules/awaymissions/egg_event.dm
Very-Soft: did this because I don't think u can use macros when setting a variable on the object. Just moved this into the entered proc with the other messages instead

code/modules/awaymissions/egg_event.dm
Very-Soft: was just being double extra sure people only entered the list once, changed it to +=

code/modules/awaymissions/egg_event.dm
Very-Soft: k

code/modules/awaymissions/egg_event.dm
Kashargul: macros are part of the early compile task, so it should make no issue

KillianKirilenko:
...is this some kind of obscure reference almost nobody is going to get, or is otherwise going to be horribly outdated within about six months?

code/modules/awaymissions/egg_event.dm
Kashargul: See here 

https://github.com/VOREStation/VOREStation/blob/7646be36291efb69d7769b129179691c30aa390d/code/modules/mob/_modifiers/feysight.dm#L6

also might want to axe the double space

Very-Soft: The double space was intentional but I will follow your instruction.

JoanRisu: Double space shouldn't be a problem if its an intentional for the presentation, no?

JoanRisu:
Obscure reference.

code/modules/awaymissions/egg_event.dm
Kashargul: > Double space shouldn't be a problem if its an intentional for the presentation, no?

My issue is that it’s a poor formatting style because, a year or two from now, nobody will know why they’re there. Someone will just see double spaces and “fix” them without realizing they were intentional.

Lilly-Lira: @Very-Soft Can you add a code comment that the spacing is intentional?  I believe we should be good after that.

Cameron-The-Raven (commented):
Everything seems good now codewise.

A 6.6% chance per ladder climb seems a bit high, but we'll see.

Very-Soft:
> Everything seems good now codewise.
> 
> A 6.6% chance per ladder climb seems a bit high, but we'll see.

the chance is 1 in 1000, as it is only looking for 1 number

Cameron-The-Raven:
> > Everything seems good now codewise.
> > 
> > A 6.6% chance per ladder climb seems a bit high, but we'll see.
> 
> the chance is 1 in 1000, as it is only looking for 1 number

Oh! I completely missed that it was a != Number.

Yeah, this should be adequately rare.

PR #19658, opened 2026-08-11
Mining machine construction

Willburd:
## About The Pull Request
Allows the construction and disassembly of the mining processor, stacking machine, unloader, and mint. As well as directional settings for them. Stacking machines now drop their contents on destruction. The mining processor does not for now, as it would allow a points exploit where you deconstruct and put the ore back through it.

## Changelog
Adds circuitboards for mining processor, stacking machine, mint
Adds construction steps for mining machines
Fixes hardref clearing in mining machine Destroys
Ore machines use conveyor detection instead of input plates during map init
Moves mining machines to their own global list for faster lookups
Mining machines and their consoles link based on nearest unlinked console/machine distance
Stacking machine releases all sheets on destruction
Direction of mining machines can be set by dragging them in a direction, altclick to clear
Ore machines now use bump() to handle incoming ores and sheets, making them directionally agnostic.
Mining machine circuitboards added to base mining techweb node

:cl: Will
add: Mining processer, stacking machine, and mint can now be constructed
add: Mining machines can have their direction set by dragging them in a direction, and alt-clicking to reset it.
add: Mining machines and consoles link to the nearest one when constructed, and show unlinked status
refactor: General cleanup and hardref fixes for mining machines
fix: Destroyed or deconstructed stacking machines drop their sheet contents
/:cl:


Willburd:
Testing requested, I'm not 100% familiar with all the quirks of mining machinery.

code/modules/mining/machinery/mineral.dm
```diff
@@ -0,0 +1,71 @@
...
+	GLOB.mineral_machines -= src
+	. = ..()
+
+/obj/machinery/mineral/proc/has_link()
+	return FALSE
+
+/obj/machinery/mineral/proc/find_nearest_linkable(filtering_type)
+	var/nearest_distance = INFINITY
+	var/nearest_machine = null
+	for(var/obj/machinery/mineral/checking in GLOB.mineral_machines)
```
Kashargul: @Cameron-The-Raven tracy the global list vs a range please, curious about it.

code/modules/mining/machinery/machine_unloading.dm
```diff
@@ -33,72 +41,88 @@
...
-					else if(ore == ORE_VOPAL)
-						ore_chunk.icon_state = "ore_void_opal"
-					else if(ore == ORE_BAUXITE)
-						ore_chunk.icon_state = "ore_bauxite"
-					else if(ore == ORE_PAINITE)
-						ore_chunk.icon_state = "ore_painite"
-					else
-						ore_chunk.icon_state = "boulder[rand(1,4)]"
+	var/turf/output = get_step(src,output_dir)
+	var/turf/input = get_step(src,reverse_direction(output_dir))
```
Poojawa, line 64: This could be both a multitool function as well as being aware of conveyor belt direction, rather than hard coded in/out. adjacency check to avoid teleporting mischief, 

If a conveyor is directional inwards, hint that it could be an input. if a conveyor direction is outwards from the machine, well that's probably an output vector. Instantly fixes the corner regression.

code/modules/mining/machinery/machine_unloading.dm
```diff
@@ -33,72 +41,88 @@
...
+				else if(ore == ORE_COPPER)
+					ore_chunk.icon_state = "ore_copper"
+				else if(ore == ORE_TIN)
+					ore_chunk.icon_state = "ore_tin"
+				else if(ore == ORE_VOPAL)
+					ore_chunk.icon_state = "ore_void_opal"
+				else if(ore == ORE_BAUXITE)
+					ore_chunk.icon_state = "ore_bauxite"
+				else if(ore == ORE_PAINITE)
+					ore_chunk.icon_state = "ore_painite"
```
Poojawa:
```
if(ore)
	ore_chunk.icon_state = ore.icon_state
else
	ore_chunk.icon_state = "boulder[rand(1,4)]"
```

These are all identical to the ore item path anyway, sooo...

code/modules/mining/machinery/mineral.dm
Poojawa: This really should just be area range check like APCs/SMES do for their terminal connections. 

If it's ~~magic~~ bluespace station connection then it should just be checking station z levels for rather than trolling through all machines everywhere until it finds something. Headache protection if someone decides to make an independent mining POI like tg's golem ship.

Willburd: > This really should just be area range check like APCs/SMES do for their terminal connections.
> 
> If it's ~magic~ bluespace station connection then it should just be checking station z levels for rather than trolling through all machines everywhere until it finds something. Headache protection if someone decides to make an independent mining POI like tg's golem ship.

Area check is prone to area border issues, if players build them on the border of an area it won't link despite seemingly being in range due to the areas differing. The check is on it's own global list, or possibly in range() check, waiting for diana on that tracy performance check though to see which has better performance.

code/modules/mining/machinery/machine_unloading.dm
Willburd: > This could be both a multitool function as well as being aware of conveyor belt direction, rather than hard coded in/out. adjacency check to avoid teleporting mischief,
> 
> If a conveyor is directional inwards, hint that it could be an input. if a conveyor direction is outwards from the machine, well that's probably an output vector. Instantly fixes the corner regression.

I've also been thinking about moving the logic to bump, as that would also accept input from all directions while still using protolathe style drag direction output. Which was the goal of this PR + constructable.

code/modules/mining/machinery/machine_unloading.dm
Willburd: > ```
> if(ore)
> 	ore_chunk.icon_state = ore.icon_state
> else
> 	ore_chunk.icon_state = "boulder[rand(1,4)]"
> ```
> 
> These are all identical to the ore item path anyway, sooo...

ore var in this context is not a reference to an ore object, it's just a string id. It doesn't have icon_state, that's not how oreboxes store ores. The ore objects themselves are destroyed on collection, and it just counts up in an assoc list using the ore's id. Which is what's used here for the if. It doesn't look like there's too nice of a way to get the icon_state from the id. 

I'll need to use a lookup.

Willburd: Managed to make it a lot less awful then that ugly if-else.

code/modules/mining/machinery/machine_unloading.dm
Willburd: Additionally, I would also like to obliterate the input panels too, they're not even the right object type. The issue with that is it would spread into other objects that have been using them, and require map edits to all maps. I'm going to try the bump logic first, and do another PR in the future to kill input panels.

Willburd: Ores on conveyors uses bump logic now. Directly unloading oreboxes into the processor still uses the reverse direction, but none of the processing setups use this on any of our maps, they use unloaders onto conveyors. Conveyors will work fine now without any map edits.

Willburd: Actually, I'm going to try doing the smart directional check detecting conveyors as well as you recommend.

Willburd: Using conveyor detection now.

code/modules/mining/machinery/mineral.dm
Willburd: Moved to range check as per tracy.

Willburd:
Awaiting final review, testing seems to show this ready with the corner regression fixed by using bump code.

PR #19611, opened 2026-07-14
[READY] Taursuits and Tail socks [IDB IGNORE]

Poojawa:
Implements a VASTLY expanded Taursuit coverage for the majority of non-exotic species in the taur line up. This also follows #19596 for sorting into what counts as exotic. I got a lot!

fa45dc4ee74f2a1b2d5f290be3821bd2

540615ed651fa5c8d1af4a2f2c0ae84e

Also implements a tail sock system, which permits ALL sprite_accessory/tails (leggies too I guess) to be sheathed on a sprite - no more requirements of having to have single species ENTIRE sprites just to add on a funny oval shape for a tail. I'll be yeeting all of the species-specifics into human mobs. Don't need 'em! Backpack logic remains confoundingly annoying for north facement. But if you use a satchel you won't notice it. I'll probably look into making taur-fitted saddled bags and shoes later. There's a few edge case rigsuits that really need them to look nice. 

As shown, they also support wag states, so you can wag in space!

2cf2d1c6be8839a13bdfb3c7dc763fef

Yes, you can set the suits to the old mode as well, if you want. It's a toggle on the suit item itself. You can toggle just the tailsock off, or the entire suit sprite to default. Clip masks were improved to an extent. 

Taur versions are just the tail and the fat version to cover up their belly. Using icon_sprite_tag will append _sock and _w_sock for the regular and wag versions. This has to be manually set for tail/taur types that do not utilize a custom color (or have a special sock feature, like a clear portion for a tail maw). Body types that don't really make sense for plating (Nagas, mermaids, tree roots, etc) are aggressively clipped and socked by default. 

Tauric suits have two master sprite sheets, all handcrafted by me. Yes, there's a lot of them. Yes it was very crazy of me to do so. There is a readme file that explains how it works, along with my template files and the sprite sheets. It doesn't cover every single costume, but I tried to get as much as I could smacked in that you'd easily get on station other than vanity. Easy enough to add more later though!

Mob spritesheets have all be reset to the default mob, except for Teshari and Altavan or w/e they are. The files remain so if someone wants to custom make a species' aspect like unathi's big shoulder pauldrons. But you don't need to add tail pixels to the suits anymore! None of them!

## YOU DO NOT NEED TO SPRITE TAILS ONTO SUIT ANYMORE AAAAAAAA

An example of the "Master list" sprites of long/standard body length. Most taurs fit into these.

77ada2f55b6495960e39a8dcb028d3a2

:cl:Poojawa
[...]

PR #19661, opened 2026-08-12
Revert "[READY] Taursuits and Tail socks [IDB IGNORE]"

Kashargul:
Reverts VOREStation/VOREStation#19611


Cameron-The-Raven:
Notes before full readd:
The tauric halves on clothing should be readded / fixed. There was a massive drop in artistic quality on suits that were previously sprited. See:

Previously:
image


Now:
image

Poojawa:
I disagree on the "Massive drop in quality". some aspects of it could be improved, but the old, wolf only, sprites were essentially this same base sprite and then smearing a noise filter on the sprite. The excessive noise makes recoloring extremely tedious to carry out for the 30 some department coats - all of which were blurred, smeared, and overall awful to deal with.

PR #19663, opened 2026-08-14
Revert "Revert "[READY] Taursuits and Tail socks [IDB IGNORE]""

Kashargul:
Reverts VOREStation/VOREStation#19661

Push it through a proper TM process. The issues and fixes should be properly tested before being pushed and run on master.


Poojawa:
Uhh

Why did this get closed when the hotfix didn't even get reviewed or tested??

PR #19672, opened 2026-08-18
Micro in Food check

Poojawa:
Addresses https://github.com/CHOMPStation2/CHOMPStation2/issues/12968

:cl: Poojawa
fix: attempting to eat/feed food with a micro inside without the eater both food vore and Spont pred enabled will give an alert instead of silently failing or ejecting the prey onto the floor. You really should just be examining what you eat beforehand though. 
/:cl:

### DIFF
--- a/code/modules/food/food/snacks.dm
+++ b/code/modules/food/food/snacks.dm
@@ -129,7 +129,7 @@
 
 		//micro in food check if someone couldn't be bothered to examine their food.
 		if((!eater.food_vore || !eater.can_be_drop_pred) && food_inserted_micros && food_inserted_micros.len)
-			to_chat(eater, span_danger("Ewww, [user == eater ? "You can't" : "\The [eater] can't"] consume that, there's a bug in this!") )
+			to_chat(eater, span_vdanger("Ewww, [user == eater ? "You can't" : "\The [eater] can't"] consume that, there's a bug in this!") )
 			return ITEM_INTERACT_FAILURE
 
 		var/swallow_whole = FALSE
--- a/code/modules/food/food/snacks.dm
+++ b/code/modules/food/food/snacks.dm
@@ -126,6 +126,12 @@
 			if(liquid_belly_check())
 				to_chat(user, span_infoplain("[user == eater ? "You can't" : "\The [eater] can't"] consume that, it contains something produced from a belly!"))
 				return ITEM_INTERACT_FAILURE
+
+		//micro in food check if someone couldn't be bothered to examine their food.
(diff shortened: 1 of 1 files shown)


code/modules/food/food/snacks.dm
```diff
@@ -126,6 +126,12 @@
 			if(liquid_belly_check())
 				to_chat(user, span_infoplain("[user == eater ? "You can't" : "\The [eater] can't"] consume that, it contains something produced from a belly!"))
 				return ITEM_INTERACT_FAILURE
+
+		//micro in food check if someone couldn't be bothered to examine their food.
+		if((!eater.food_vore || !eater.can_be_drop_pred) && food_inserted_micros && food_inserted_micros.len)
```
TheCaramelion, line 131: There's a helper proc in `code/_helpers/vore.dm` to check if both pred & prey are compatible, `can_food_vore()`

TheCaramelion (changes requested):

code/modules/food/food/snacks.dm
Poojawa: I'm aware, however all it does is dump the micro(s) onto the floor and doesn't give a message to the consumer. As per the reported issue, this provides a means of avoiding spoiling a 'surprise' for opt-in participants.

TheCaramelion: The helper returns TRUE/FALSE after checking prefs, nothing else

code/modules/food/food/snacks.dm
```diff
@@ -126,6 +126,12 @@
...
+			to_chat(eater, span_danger("Ewww, [user == eater ? "You can't" : "\The [eater] can't"] consume that, there's a bug in this!") )
```
Kashargul: best use vdanger as it's vore related.

code/modules/food/food/snacks.dm
Kashargul: ok.. why are we null checking and len checking? just use length(thing) at that...

if it's a known lazy list use explicitly LAZYLEN

Kashargul (changes requested):

code/modules/food/food/snacks.dm
Poojawa: Every single instance of this list check does this.

Kashargul: horrible... should just be replaced

ood_inserted_micros && food_inserted_micros.len -> LAZYLEN(food_inserted_micros)

Poojawa:
Cleaned up every interaction related to that particular list to use LAZYADD and LAZYREMOVE as well as the LAZYLEN checks. 

from null it begins, to null it returns.

PR #19686, opened 2026-08-30
Fix cyborgs bleeding oil when hit by the hound tongue

lbnesquik:
## About The Pull Request
This PR fixes this issue from downstream.
https://github.com/CHOMPStation2/CHOMPStation2/issues/12958

If something has no force, no bleeding. Plushes and other things like that probably would be doing it too.
## Changelog
:cl:

fix: fix cyborg licks making other cyborgs bleed
/:cl:


Kashargul:
Obsolete and done on an outdated branch

lbnesquik:
Only obsolete because I took a while :(

PR #19690, opened 2026-09-01
Buffs RUST radiation generation

Cameron-The-Raven:
## About The Pull Request
Buffs the RUST radiation generation. The previous levels did absolutely nothing.
When I did the rad rework, I had 0 RUSTs to go off of and no testing with the RUST, so a 'ballpark number' was used and found to be much too low.

This increases the range to scale on energy (like the SM), the power to scale based on phoron temperature (like it did previously) and increases the threshold from MEDIUM insulation to EXTREME insulation (Requires rad collectors to block it)

Likewise, when the RUST explodes, it does a 14 tile wide radiation blast with massive amounts of rads based on the reactants within it.
## Changelog
:cl: Diana
qol: Changes the RUST generator to produce more radiation and radiate further. Rad collectors (at 1013kpa) now collect about as much as a mag-trap

/:cl:

### DIFF
--- a/code/modules/power/fusion/core/core_field.dm
+++ b/code/modules/power/fusion/core/core_field.dm
@@ -544,13 +544,13 @@
 	return
 
 //Reaction radiation is fairly buggy and there's at least three procs dealing with radiation here, this is to ensure constant radiation output.
-/obj/effect/fusion_em_field/proc/radiation_scale()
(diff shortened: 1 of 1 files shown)


Cameron-The-Raven:
Virgo doesn't use the RUST. Going to merge here and TM on Chomp, since it's used there.

PR #19707, opened 2026-09-05
Correct the vent crawl error text

lbnesquik:
## About The Pull Request
When you tried to ventcrawl, it checks whether you are next to a vent, and whether it was welded or not. But, the error text was the same for both issues, and didn't mention welded vents at all. This PR corrects it.

I did test it with proteans and xenos and it worked fine.
## Changelog
:cl:
fix: fixed ventcrawl using an incorrect error message for welded vents. 
/:cl:


code/modules/ventcrawl/ventcrawl.dm
```diff
@@ -182,7 +182,13 @@
 		for(var/obj/machinery/atmospherics/machine in range(1,src))
 			if(is_type_in_list(machine, GLOB.ventcrawl_machinery))
 				vent_found = machine
-			if(!vent_found || !vent_found.can_crawl_through())
+			// if(!vent_found || !vent_found.can_crawl_through())
+			// 	vent_found = null
+			if(!vent_found.can_crawl_through())
```
Kashargul, line 187: This will runtime to hell and back. You are literally accessing null.var

first null check... then access an object variable

code/modules/ventcrawl/ventcrawl.dm
```diff
@@ -182,7 +182,13 @@
...
+				vent_found = null
+				to_chat(src, span_danger("You can't crawl through that welded vent!"))
+				return
+			if(!vent_found)
```
Kashargul: this needs to be above the object var access and just should be a continue... why set it to null if it was null?

Kashargul (dismissed):

code/modules/ventcrawl/ventcrawl.dm
lbnesquik: Whoops, you are right.

lbnesquik:
Removed the commented text as well.

code/modules/ventcrawl/ventcrawl.dm
```diff
@@ -182,8 +182,12 @@
 		for(var/obj/machinery/atmospherics/machine in range(1,src))
 			if(is_type_in_list(machine, GLOB.ventcrawl_machinery))
 				vent_found = machine
-			if(!vent_found || !vent_found.can_crawl_through())
+			if(!vent_found)
```
Kashargul, line 185: still makes little sense, if null, make null

if(!vent_found) 
    continue

lbnesquik: What about this instead? 
```
if(!vent_found)
   return
```

Kashargul: > What about this instead?
> 
> ```
> if(!vent_found)
>    return
> ```

can we safely return? In general likely yes as we only care for the first machine

lbnesquik: Yes, we do it if the vent isn't next to us.

Kashargul: Might need to be a continue here, because no we only ever check the first atmospherics we find.

lbnesquik: Added it.

PR #19711, opened 2026-09-06
Allows the wrenching of smartfridges

lbnesquik:
## About The Pull Request
It is currently impossible to move smartfridges because they think you are trying to put the tool inside them. 

Fixes this issue over that downstream : 
https://github.com/CHOMPStation2/CHOMPStation2/issues/4171

## Changelog
:cl:
qol: Made it possible to wrench smartfridges.
/:cl:

### DIFF
--- a/code/modules/food/kitchen/smartfridge/smartfridge.dm
+++ b/code/modules/food/kitchen/smartfridge/smartfridge.dm
@@ -121,7 +121,13 @@
 		update_icon()
 		return
 
-	if(wrenchable && default_unfasten_wrench(user, O, 20))
+	if(wrenchable)
+		if(O.has_tool_quality(TOOL_WRENCH))
+			if(allowed(user))
+				default_unfasten_wrench(user, O, 20)
+			else
+				to_chat(user, span_warning("\The [src] smartly denies you access to wrench it."))
+			return
 		return
 
 	if(O.has_tool_quality(TOOL_CROWBAR))
--- a/code/modules/food/kitchen/smartfridge/smartfridge.dm
+++ b/code/modules/food/kitchen/smartfridge/smartfridge.dm
@@ -20,7 +20,7 @@
 	var/locked = 0
 	var/scan_id = 1
 	var/is_secure = 0
-	var/wrenchable = 0
+	var/wrenchable = 1
 	var/persistent = null // Path of persistence datum used to track contents
 	circuit = /obj/item/circuitboard/smartfridge //This one is meant to be uncraftable, however.
 
--- a/code/modules/food/kitchen/smartfridge/smartfridge.dm
+++ b/code/modules/food/kitchen/smartfridge/smartfridge.dm
@@ -20,7 +20,11 @@
 	var/locked = 0
 	var/scan_id = 1
 	var/is_secure = 0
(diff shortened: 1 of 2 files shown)


code/modules/food/kitchen/smartfridge/smartfridge.dm
```diff
@@ -20,7 +20,7 @@
 	var/locked = 0
 	var/scan_id = 1
 	var/is_secure = 0
-	var/wrenchable = 0
+	var/wrenchable = 1
```
Kashargul: I'm not sure this is a good idea to do for the mapped in ones.

Poojawa:
```suggestion
	var/shoot_inventory = FALSE
	var/locked = FALSE
	var/scan_id = TRUE
	var/is_secure = FALSE
	var/wrenchable = TRUE
```

Might as well also make these not magic number too.

lbnesquik: Well, you sometimes might want to move the bar vendor. Maybe it can check for access like for decon if you are worried about people taking some specific ones.

Kashargul: My worries are persistent ones being hauled around. Those should certainly remain in place.

lbnesquik:
Implemented Poojawa recommendation. Deleted the 'wrenchable' option from the drying racks given we default to wrenchable already. And, copied the access check for the deconstruction for wrenching, so you can't run away with smartfridges you don't have access to. 

Locked stays as numbers based because emags set that to -1.

code/modules/food/kitchen/smartfridge/smartfridge.dm
lbnesquik: We could just slap wrenchable = false on them?

Kashargul: putting it to staff discussion

code/modules/food/kitchen/smartfridge/smartfridge.dm
```diff
@@ -175,7 +181,7 @@
 
 /obj/machinery/smartfridge/secure/emag_act(remaining_charges, mob/user)
 	if(!emagged)
-		emagged = 1
+		emagged = TRUE
 		locked = -1
```
Poojawa, line 189:
```suggestion
		locked = FALSE
```

It's not like there's a mythical third state for a binary locked or unlocked??

Probably an unintended misattribution to eating an emag charge five+ years ago

lbnesquik: I wager the idea is that it can't be toggled back once it is emagged? I could make another variable for it. Emagged and so on.

code/modules/food/kitchen/smartfridge/smartfridge.dm
Kashargul: Please ensure that persistent ones can't be moved

lbnesquik: I added a specific check for that, so we should be all good!

Kashargul (approved):

PR #19713, opened 2026-09-07
Remove non-functional xenochimera version of hardfeet trait

Willburd:
## About The Pull Request
Removes non-functional xenochimera version of the hardfeet trait. Both traits cost 0 points, and you can already take normal hardfeet as xenochimera. This trait does not work, as hardfeet is a neutral subtype, and xenochi version assumes it will inherit, but is a positive subtype? Either way the trait is useless and costs nothing, and you can always get the normal trait. So i'm just removing it outright as code bloat.

## Changelog
Removes xenochimera version of hardfeet trait.

:cl: Will
del: Removes non-functional xenochimera version of hardfeet trait
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/traits/positive.dm
+++ b/code/modules/mob/living/carbon/human/species/station/traits/positive.dm
@@ -482,15 +482,6 @@
 	category = 0
 	custom_only = FALSE
 
-/datum/trait/positive/hardfeet/xenochimera
-	sort = TRAIT_SORT_SPECIES
-	allowed_species = list(SPECIES_XENOCHIMERA)
-	name = "Xenochimera: Hard Feet"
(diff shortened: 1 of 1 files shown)


Kashargul:
Please test what happens if any save file was using it.

Willburd:
> Please test what happens if any save file was using it.

I've commented out traits on outpost before, this should do  the same where the trait is cleaned from the slot during trait sanity checks. This trait was non-functional and cost 0 points, so it shouldn't affect character points much. I can do a check again to be safe. I will need a bit as I am busy today. 

I am unsure if you want a migration from the xeno-positive to generic-neutral trait, as it would be messing with player traits even more than just a simple sanity cleanup

Cameron-The-Raven (approved):

Cameron-The-Raven:
All it does is remove the trait from the character loadout as soon as the file is loaded (as it's an invalid trait)

PR #19728, opened 2026-09-13
Reworks Geiger Counters to Behave Like Before the Radiation Rework

Willburd:
becnoid

## About The Pull Request
Redoes handling of geiger counters to behave like they did previously. Only reporting the radiation you would receive if you are standing on their turf. This is based off of analyzing an older repo's rad implementation, and comparing it to how radiation is applied to mobs. 

The underlying radiation rework is unchanged, some parts were condensed into procs or defines for reuse in other places, but instead of showing the breakdown of the radiation source vs the insulation and distance. It now behaves like it did before the rework of simply showing the active amount of rads (in Bq) that you receive standing at a location that is being irradiated. 

Chance based radiation exposure is still present, but the core issue is that players did not respond well to the "indirect" data being shown. Resulting in constant confusion and a repeated request to return to just showing the amount of rads at the turf the counter is on.

Also adds a radiation debugging tool, because testing purely with known objects and wondering why one behaves wildly different from another by code-digging every time is awful. I also made it clear that there are different types of radiation penetration through walls. Because threshold code is silly. I marked these as alpha, beta, and gamma for simplicity. alpha cannot pass a single wall, beta can pass a reinforced wall, and gamma does not really care about much. There is technically also neutrino, but that's only if you max out radiation threshold entirely, nothing does this by default, so it's mostly an engineering brag goal.

## Changelog
Returns geiger counter to showing the measured effect on a mob per the turf they are standing on, instead of a comparative analysis of the radiation source's emissions and wall insulation.
Fixes an issue where geiger counters in your hand wouldn't show the actual turf's radiation, and instead apparently a rad count modified by the storage depth of being inside your mob.
Geiger counters now show the type of radiation being emitted. As not all radiation emitters emit radiation that can pass through walls (some can't even pass standard walls! no matter how high rads are!)

:cl: Will
add: Geiger counters now show the type of radiation being emitted.
refactor: Restored geiger counter to pre-radiation update behavior
fix: geiger counters in hands show the current turf's radiation, not a reduced radiation from being inside your mob
fix: radiation sources that qdel before processing their radiation pulse would lock up radiation subsystem
code: Radiation debug object
/:cl:


Willburd:
Turns out sev-1 issues needs to be fixed as well with rad controller lockup.

Willburd:
pretty sure this is entirely ready now with no other additions. Unless there are virgo tweaks.

ItsSelis:
@Cameron-The-Raven you keep forgetting to apply the milestone

ItsSelis (approved):

PR #19729, opened 2026-09-14
SDMM update fix

Willburd:
## About The Pull Request
SDMM updated, so now SHOULD_NOT_SLEEP() checks for some recursive cases not caught before.

## Changelog
Fixes a SDMM should not sleep error on compile

:cl: Will
code: Fixes SDMM should_not_sleep issue on compile
/:cl:


Willburd:
Not sure if this is the best solution, Invoke async is an option, but seems to be frowned on.

